PR #66, opened 2022-02-24
Introduce ConnectClerk

### LINKED ISSUE #54: `WithClerkState.data` prop should be optional
- [x] Review the documentation: https://docs.clerk.dev/
- [x] Search for existing issues: https://github.com/clerkinc/javascript/issues
- [x] Go through package changelog files.
- [x] Provide the Frontend API key from your application dashboard.
    - `clerk.rjzug.yeali.lcl.dev`

## Package + Version

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/clerk-remix`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] other:

### Version:

```
0.1.0-alpha.0
```

### Browser/OS

`Chrome latest`

## Description

Currently the type, `WithClerkState`, associated with being used with `useLoaderData` has a type that implies that the `data` prop is guaranteed: https://github.com/clerkinc/javascript/blob/next/packages/remix/src/ssr/types.ts#L5-L8.

However, when trying out the package with my project, `data` was undefined bc an `interstitial` was returned from `getAuthData` in `rootAuthLoader`: https://github.com/clerkinc/javascript/blob/next/packages/remix/src/ssr/rootAuthLoader.ts#L28-L30

So, this leads me to the following questions/suggestions, in order of preference:
1. What is interstitial? Ideally, `getAuthData` would return the user either logged in or out and then I would have my data returned.
2. If interstitial is kept, could we return `data` so that I always have access to it?
3. If I cannot keep data, then it should be optional on `WithClerkState`

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

This PR introduces small but important changes to the setup process our users need to follow in order to integrate Clerk in a Remix project.

Before this PR, the recommended `root.tsx` file would look like:

```tsx
import { ClerkProvider } from "@clerk/remix";
import { rootAuthLoader, WithClerkState } from "@clerk/remix/ssr.server";

export const loader: LoaderFunction = (args) => rootAuthLoader(args, ({ auth }) => {
  const { userId } = auth;
  const posts: Posts[] = database.getPostsByUser(userId);
  return posts;
});

export default function App() {
  const { data: posts, clerkState } = useLoaderData<WithClerkState<Posts[]>>();
  
  return (
    <ClerkProvider frontendApi={"clerk.renewing.hermit-14.lcl.dev"} clerkState={clerkState}>
      <html lang="en">...</html>
    </ClerkProvider>
  );
}

```
This approach however, has 2 major drawbacks: 
1. The user needs to explicitly pass `clerkState` to the `ClerkProvider`
2. In the specific edge cases where the interstitial page needs to be used, the user defined data returned from `useLoaderData` would be undefined. Since we cannot throw the interstitial from the loader directly, we use `ClerkProvider` instead. However, If the user used the undefined `data` before `ClerkProvider` handlers the interstitial, the app could break. Many thanks to @[dvargas92495](https://github.com/dvargas92495) who also reported this in #54 

In order to solve both issues, we're introducing a `ConnectClerk` HOC that can be used as such:

```tsx
export const loader: LoaderFunction = (args) => rootAuthLoader(args, ({ auth }) => {
  const { userId } = auth;
  const posts: Posts[] = database.getPostsByUser(userId);
  return { posts };
});

function App() {
  const { posts } = useLoaderData<{ posts: Posts[] }>();
  return (
    <html lang='en'>
      ....
    </html>
  );
}

export default ConnectClerk(App, { frontendApi: '...' });
```

Notice that the user no longer needs to manually use ClerkProvider or manually pass clerkState to ClerkProvider. Also, now we can handle the interstitial state before any user-defined code runs.

Lastly, `rootAuthLoader` now **requires** either a Response or an object to be returned from the callback, if one is provided . This greatly simplifies how `useLoaderData` is used in the root App component. We updated the types accordingly so we now throw but on compile and on runtime if this rule is not respected:

image
image

Fixes #54

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1737,7 +1737,6 @@
       "version": "7.17.0",
       "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.17.0.tgz",
       "integrity": "sha512-etcO/ohMNaNA2UBdaXBBSX/3aEzFMRrVfaPv8Ptc0k+cWpWW0QFiGZ2XnVqQZI1Cf734LbPGmqBKWESfW4x/dQ==",
-      "dev": true,
       "dependencies": {
         "regenerator-runtime": "^0.13.4"
       },
@@ -5481,11 +5480,25 @@
         "url": "https://opencollective.com/popperjs"
       }
     },
+    "node_modules/@remix-run/react": {
+      "version": "1.2.2",
+      "resolved": "https://registry.npmjs.org/@remix-run/react/-/react-1.2.2.tgz",
+      "integrity": "sha512-A+g/Y4QnT4ldZovjbP4sNJ29ZZt+oDP152pN8oYOHoQuqqrlc+CnLSMKCN1D+EljBfJx66553TwMIS5kCwdNzg==",
+      "peer": true,
+      "dependencies": {
+        "history": "^5.2.0",
+        "react-router-dom": "^6.2.1"
+      },
+      "peerDependencies": {
+        "react": ">=16.8",
+        "react-dom": ">=16.8"
+      }
+    },
     "node_modules/@remix-run/server-runtime": {
       "version": "1.2.1",
       "resolved": "https://registry.npmjs.org/@remix-run/server-runtime/-/server-runtime-1.2.1.tgz",
       "integrity": "sha512-wn01V+O7y4dHQ059QGeXOpbG+d3IrqYnil9YDsJi4kielEFRDHDAFbmrMC+oF5qRYTuVRR3i6h2CSCyV2ICfeQ==",
-      "dev": true,
+      "peer": true,
       "dependencies": {
         "@types/cookie": "^0.4.0",
         "cookie": "^0.4.1",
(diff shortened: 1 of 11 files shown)


colinclerk (approved):
Lovely!  We'll update the docs to use `ConnectClerk()` instead of `<ClerkProvider>` directly

SokratisVidros (approved):
Much better. 🎖️ Very nicely done @nikosdouvlis

dvargas92495:
this works for me!

PR #77, opened 2022-03-04
Remix integration improvements

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Up until now, we used `throw json()` to stop a loader from inside `getAuth` without having to change the loader api and introduce a loader wrapper (which is not recommended by Remix). However, depending on the app's structure it's possible that a nested `CatchBoundary` will catch this error before `ClerkProvider` renders the interstitial.

Remix runs all loaders in [parallel and waits for all of them to either return or throw](https://github.com/remix-run/remix/blob/5b8a0ce0aa0201aa2402fc41405ffbe89605963b/packages/remix-server-runtime/server.ts#L302). Once all the loaders have run, Remix evaluates the returned values in order, stopping when [the first error or redirect is encountered](https://github.com/remix-run/remix/blob/5b8a0ce0aa0201aa2402fc41405ffbe89605963b/packages/remix-server-runtime/server.ts#L342).  If an error is found, Remix try to match it with a `CatchBoundary`, and if a `CatchBoundary` exists it gets rendered that instead of the normal react tree (that includes our `ClerkProvider`). Notice though, that if Remix finds a redirect before an error, they just follow the redirect and render nothing at all.

The above means that:
- We still need to throw from getAuth to stop the loader early during interstitial, otherwise we risk a returned redirect or an actual error depending on the user’s code inside nested loaders.
- During interstitial, we need to throw from inside `rootAuthLoader`  to prevent Remix from evaluated any nested redirects/errors and catch boundaries

In order to support all the edge cases described so far, we will use the root `CatchBoundary` to render the interstitial.
```
//root.tsx
export const loader: LoaderFunction = args => rootAuthLoader(args);

function App() {
  return (
    <html lang='en'> ... </html>
  );
}

export default ConnectClerk(App);

export const CatchBoundary = ConnectClerkCatchBoundary();
```

Going with this approach also means that if they allow us to return arbitrary responses from the loaders in the future,  we can simply remove  `export const CatchBoundary = ConnectClerkBoundary();` from our docs - no action needed from our users and no breaking changes.

This PR includes the following changes:
- Introduces `ConnectClerkCatchBoundary`. This HOC must be exported from the root.tsx page. If a user needs a custom boundary, they can simply wrap their boundary with the HOC. eg:
```
export const CatchBoundary = ConnectClerkCatchBoundary(() => {
  const caught = useCatch();
  return (
    <html>
           ...
    </html>
  );
});
```

- Removes the `frontendApi` option from `ConnectClerk`. Instead of hardcoding it, the users are now able to use the value directly from the env, by creating a `.env` file. The CLERK_FRONTEND_API value will then become accessible during SSR and CSR. Eg: 
```
CLERK_FRONTEND_API=clerk.***.***-14.lcl.dev
CLERK_API_KEY=test_jP61SgV*******6IVy07li3APw
```
For anyone that wants to override the env value, we added a `frontendApi` option in the `rootAuthLoader` options. Eg:
```
export const loader: LoaderFunction = args => rootAuthLoader(args, {frontendApi: '...'});
```

- Instead of fetching the interstitial, we now compute it locally, requiring 1 less BAPI request (~90-300ms)

### DIFF
--- a/packages/remix/src/client/Interstitial.tsx
+++ b/packages/remix/src/client/Interstitial.tsx
@@ -1,5 +1,47 @@
 import React from 'react';
 
-export function Interstitial({ html }: { html: string }) {
-  return <html dangerouslySetInnerHTML={{ __html: html }} />;
+function isStaging(frontendApi: string): boolean {
+  return (
+    frontendApi.endsWith('.lclstage.dev') ||
+    frontendApi.endsWith('.stgstage.dev') ||
+    frontendApi.endsWith('.clerkstage.dev')
+  );
+}
+
+const getScriptUrl = (frontendApi: string, libVersion: string) => {
+  const major = libVersion.includes('alpha') ? 'next' : isStaging(frontendApi) ? 'staging' : libVersion.split('.')[0];
+  return `https://${frontendApi}/npm/@clerk/clerk-js@${major}/dist/clerk.browser.js`;
+};
+
+const createInterstitialHTMLString = (frontendApi: string, libVersion: string) => {
+  return `
+    <head>
+        <meta charset="UTF-8" />
+    </head>
+    <body>
+        <script>
+            window.startClerk = async () => {
+                const Clerk = window.Clerk;
+                try {
+                    await Clerk.load();
+                    window.location.reload();
+                } catch (err) {
+                    console.error('Clerk: ', err);
+                }
+            };
+            (() => {
+                const script = document.createElement('script');
+                script.setAttribute('data-clerk-frontend-api', '${frontendApi}');
+                script.async = true;
+                script.src = '${getScriptUrl(frontendApi, libVersion)}';
+                script.addEventListener('load', startClerk);
+                document.body.appendChild(script);
+            })();
+        </script>
+    </body>
+`;
+};
+
+export function Interstitial({ frontendApi, version }: { frontendApi: string; version: string }) {
+  return <html dangerouslySetInnerHTML={{ __html: createInterstitialHTMLString(frontendApi, version) }} />;
 }
--- a/packages/remix/src/errors.ts
+++ b/packages/remix/src/errors.ts
@@ -65,3 +65,7 @@ export const loader: LoaderFunction = args => rootAuthLoader(args, ({ auth }) =>
     return { data: posts };
 })
 `);
+
+export const getAuthInterstitialErrorRendered = createErrorMessage(
+  `This state shouldn't be possible. Please reach out to support@clerk.dev so we can help, or use the links below:`,
+);
--- a/packages/remix/src/errors.ts
+++ b/packages/remix/src/errors.ts
@@ -69,3 +69,9 @@ export const loader: LoaderFunction = args => rootAuthLoader(args, ({ auth }) =>
 export const getAuthInterstitialErrorRendered = createErrorMessage(
   `This state shouldn't be possible. Please reach out to support@clerk.dev so we can help, or use the links below:`,
 );
+
+export const noFrontendApiError = createErrorMessage(`
+The CLERK_FRONTEND_API1 environment variable must be set before using Clerk.
+During development, grab the Frontend Api value from the Clerk dashboard, create an .env file and set the CLERK_FRONTEND_API key.
+For production apps, please consult the Remix documentation on environment variables.
+`);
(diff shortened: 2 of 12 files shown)


packages/remix/src/client/Interstitial.tsx
```diff
@@ -1,5 +1,47 @@
 import React from 'react';
 
-export function Interstitial({ html }: { html: string }) {
-  return <html dangerouslySetInnerHTML={{ __html: html }} />;
+function isStaging(frontendApi: string): boolean {
```
SokratisVidros, line 3: 🙃 We should move this logic to a helper in ClerkJS

packages/remix/src/client/Interstitial.tsx
```diff
@@ -1,5 +1,47 @@
...
+                try {
+                    await Clerk.load({authVersion: 2});
+                    window.location.reload();
+                } catch (err) {
+                    console.error('Clerk: ', err);
+                }
+            };
+            (() => {
+                const script = document.createElement('script');
+                script.setAttribute('data-clerk-frontend-api', 'clerk.renewing.hermit-14.lcl.dev');
```
SokratisVidros, line 34: ⛔ The frontend API should be interpolated in the ClerkJS loading script.

packages/remix/src/client/Interstitial.tsx
```diff
@@ -1,5 +1,47 @@
...
+  return `
+    <head>
+        <meta charset="UTF-8" />
+    </head>
+    <body>
+        <script>
+            window.startClerk = async () => {
+                const Clerk = window.Clerk;
+                try {
+                    await Clerk.load({authVersion: 2});
```
SokratisVidros, line 26: 🙃 

```suggestion
                    await Clerk.load();
```

packages/remix/src/errors.ts
```diff
@@ -65,3 +65,13 @@ export const loader: LoaderFunction = args => rootAuthLoader(args, ({ auth }) =>
     return { data: posts };
 })
 `);
+
+export const getAuthInterstitialErrorRendered = createErrorMessage(
+  `This state shouldn't be possible. Please reach out to support@clerk.dev so we can help, or use the links below:`,
+);
+
+export const noFrontendApiError = createErrorMessage(`
+The CLERK_FRONTEND_API1 environment variable must be set before using Clerk.
```
SokratisVidros, line 74: 🙃 

```suggestion
The CLERK_FRONTEND_API environment variable must be set before using Clerk.
```

packages/remix/src/errors.ts
```diff
@@ -65,3 +65,13 @@ export const loader: LoaderFunction = args => rootAuthLoader(args, ({ auth }) =>
...
+During development, grab the Frontend Api value from the Clerk dashboard, create an .env file and set the CLERK_FRONTEND_API key.
```
SokratisVidros, line 75: 🙃 

```suggestion
During development, grab the Frontend Api value from the Clerk Dashboard, create an .env file and set the CLERK_FRONTEND_API key.
```

SokratisVidros (changes requested):

packages/remix/src/client/Interstitial.tsx
nikosdouvlis: Thanks for catching that

nikosdouvlis:
> Everything in the PR looks great, thank you!
> 
> For consistency, can we add the ability to pass `apiKey?` directly to `RootAuthLoaderOptions`?

Thanks for the suggestion! We could, but it's not trivial as it requires changes to `clerk-sdk-node`. Right now, the SDK initialises before the `rootAuthLoader` runs. I created a ticket so we can tackle this separately.

PR #87, opened 2022-03-08
fix(backend-core): More robust cross-origin check for dev/prod

igneel64:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Revise the cross-origin check based on request headers.
Please take a look at the code of the process and tests.

### DIFF
--- a/packages/backend-core/src/Base.ts
+++ b/packages/backend-core/src/Base.ts
@@ -4,7 +4,7 @@ import { JWTExpiredError } from './api/utils/Errors';
 import { parse } from './util/base64url';
 import { isDevelopmentOrStaging, isProduction } from './util/clerkApiKey';
 import { checkClaims, isExpired } from './util/jwt';
-import { checkCrossOriginReferrer } from './util/request';
+import { checkCrossOrigin } from './util/request';
 import { JWT, JWTPayload } from './util/types';
 
 export const API_KEY = process.env.CLERK_API_KEY || '';
@@ -266,9 +266,16 @@ export class Base {
     }
 
     // In cross-origin requests the use of Authorization header is mandatory
-    // TODO: The origin header can be present in same-origin POST, PUT, PATCH requests etc.
-    // Is this check enough for API endpoint authorization ?
-    if (origin) {
+    if (
+      origin &&
+      checkCrossOrigin({
+        originURL: new URL(origin),
+        host,
+        forwardedHost,
+        forwardedPort,
+        forwardedProto,
+      })
+    ) {
       return { status: AuthStatus.SignedOut };
     }
 
@@ -284,8 +291,8 @@ export class Base {
     if (
       isDevelopmentOrStaging(API_KEY) &&
       referrer &&
-      checkCrossOriginReferrer({
-        referrerURL: new URL(referrer),
+      checkCrossOrigin({
+        originURL: new URL(referrer),
         host,
         forwardedHost,
         forwardedPort,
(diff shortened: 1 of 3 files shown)


packages/backend-core/src/Base.ts
```diff
@@ -263,7 +266,9 @@ export class Base {
     }
 
     // In cross-origin requests the use of Authorization header is mandatory
-    if (isCrossOrigin) {
+    // TODO: The origin header can be present in same-origin POST, PUT, PATCH requests etc.
```
igneel64, line 269: @colinclerk @SokratisVidros 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin#description 

It seems that we need to do the full check as `checkCrossOriginReferrer` to address the API endpoint use case.
1. Frontend requests a same-origin API resource from the server
2. Authorization is in the cookie and not authorization header.
3. Origin might be included as stated in the source.
4. We would return 401 (_?_)

packages/backend-core/src/__tests__/utils/request.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    expect(checkCrossOriginReferrer({ referrerURL, host })).toEqual(false);
+  });
+
+  test('is CO with mixed default security ports', () => {
+    const referrerURL = new URL('https://localhost:80');
+    const host = new URL('http://localhost:443').host;
+    expect(checkCrossOriginReferrer({ referrerURL, host })).toEqual(true);
+  });
+
+  test.skip('is CO when HTTPS to HTTP', () => {
```
igneel64, line 22: @colinclerk @SokratisVidros 
I think we cannot safely cover this case (_and vice versa_) as is.

packages/backend-core/src/Base.ts
igneel64: So is it ok if we keep the `checkCrossOriginReferrer` there for now still ?
_(if we do, we will change the name to `checkCrossOrigin`)_

packages/backend-core/src/__tests__/utils/request.test.ts
colinclerk: I think you're right, so this is best effort.

One thing I forgot is that we also try to detect `crossOriginReferrer` with the check immediately above this:
```js
    if (isDevelopmentOrStaging(API_KEY) && !clientUat) {
      return {
        status: AuthStatus.Interstitial,
        interstitial: await fetchInterstitial(),
      };
    }
```

This works because `clientUat` has `SameSite=Strict` in development, so if the page navigation is Cross**Site** then this cookie won't exist and we'll throw an interstitial.

The `checkCrossOriginReferrer` we're working on tries to capture the requests that are Cross**Origin** but not Cross**Site**.  For example, imagine a developer running their own version of `accounts` on `localhost:4000` then redirecting the user to `localhost:3000`.

It makes sense to me that Referrer-Policy can prevent this detection, but I think our options are:
1. Leave the relatively narrow error cases
2. Throw an interstitial when there's no `Referer`, but this breaks `view-source`

igneel64: I would say we are ok with this. _(at least until the next revision)_

packages/backend-core/src/Base.ts
colinclerk: yes

colinclerk (approved):

PR #97, opened 2022-03-11
feat(clerk-js,clerk-react): Replace signOutOne with signOut(options)

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Replaces signOutOne with signOut(options). 

Currently, we have 2 different signOut methods on Clerk: `signOut()` and `signOutOne` as shown below:
```
function TestComponent() {
  const clerk = useClerk();

  const signOutAllUsers = () => {
    return clerk.signOut(() => navigate('...'));
  };

  const signOutCurrent = () => {
    return clerk.signOutOne(() => navigate('...'));
  };
}
```

Proposed change: Combine the 2 methods into a single `signOut` that accepts an optional `callback` and an optional `options`  obj

- If no sessionId is provided, the client will be destroyed and the user will be signed out
- If a sessionId is provided and
  - its the only active session, the client will be destroyed and the user will be signed out
  - its the current active session, the session will be removed and the user will be signed out
  - its included in the client.sessions array but its not the current session, it will be removed but the current session will remain active

Example usage: 

### Sign out from all (multi + single session apps):
```
function TestComponent() {
  const { signOut } = useAuth();

  const signOutAll = () => {
    return signOut(() => navigate('...'));
  };
}
```

### Sign out from current (multi-session app): 
```
function TestComponent() {
  const { signOut, sessionId } = useAuth();

  const signOutCurrent = () => {
    return signOut(() => navigate(), { sessionId });
  };
}
```

### Removing an arbitrary session (multisession app):
```
function TestComponent() {
  const { signOut } = useAuth();
  const { sessions } = useSessionList();

  // single or multi session
  const signOutOtherSession = () => {
    return signOut(() => navigate('...'), { sessionId: sessions[3] });
  };
}

```

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -17,6 +17,7 @@ import type {
   Resources,
   SignInProps,
   SignInResource,
+  SignOut,
   SignOutCallback,
   SignUpProps,
   SignUpResource,
@@ -25,6 +26,7 @@ import type {
   UserProfileProps,
   UserResource,
 } from '@clerk/types';
+import { SignOutOptions } from '@clerk/types/src';
 import { AuthenticationService } from 'core/services';
 import { ERROR_CODES } from 'ui/common/constants';
 import {
@@ -137,22 +139,30 @@ export default class Clerk implements ClerkInterface {
     this.#isReady = true;
   };
 
-  public signOutOne = async (
-    signOutCallback?: SignOutCallback,
-  ): Promise<void> => {
-    if (this.#environment?.authConfig.singleSessionMode) {
-      return this.signOut(signOutCallback);
-    }
-    await this.session?.remove();
-    return this.setSession(null, ignoreEventValue(signOutCallback));
-  };
-
-  public signOut = async (signOutCallback?: SignOutCallback): Promise<void> => {
+  public signOut: SignOut = async (
+    callbackOrOptions?: SignOutCallback | SignOutOptions,
+    options?: SignOutOptions,
+  ) => {
     if (!this.client) {
       return;
     }
+    const cb =
+      typeof callbackOrOptions === 'function' ? callbackOrOptions : undefined;
+    const opts =
+      callbackOrOptions && typeof callbackOrOptions === 'object'
+        ? callbackOrOptions
+        : options || {};
+
+    if (opts.sessionId) {
+      const session = this.client.sessions.find(s => s.id === opts.sessionId);
+      await session?.remove();
+      if (this.session?.id === opts.sessionId) {
+        return this.setSession(null, ignoreEventValue(cb));
+      }
+    }
+
     await this.client.destroy();
-    return this.setSession(null, ignoreEventValue(signOutCallback));
+    return this.setSession(null, ignoreEventValue(cb));
   };
 
   public openSignIn = (props?: SignInProps): void => {
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -143,7 +143,7 @@ export default class Clerk implements ClerkInterface {
     callbackOrOptions?: SignOutCallback | SignOutOptions,
     options?: SignOutOptions,
   ) => {
-    if (!this.client) {
+    if (!this.client || !this.session) {
       return;
     }
     const cb =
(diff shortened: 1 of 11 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -137,22 +139,31 @@ export default class Clerk implements ClerkInterface {
     this.#isReady = true;
   };
 
-  public signOutOne = async (
-    signOutCallback?: SignOutCallback,
-  ): Promise<void> => {
-    if (this.#environment?.authConfig.singleSessionMode) {
-      return this.signOut(signOutCallback);
+  public signOut: SignOut = async (
```
SokratisVidros, line 142: 🔧 Since we are now combining the methods for V3 let's unit test them as well.

packages/clerk-js/src/ui/contexts/CoreClerkContext.tsx
```diff
@@ -2,43 +2,13 @@ import type { LoadedClerk } from '@clerk/types';
 import React from 'react';
 import { assertContextExists } from 'ui/contexts/utils';
 
-const excludedProps = ['user', 'session', 'client'] as const;
+export type CoreClerkProps = LoadedClerk;
 
-type ExcludedProp = typeof excludedProps[number];
```
SokratisVidros, line 7: 💯

packages/react/src/hooks/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+    isomorphicClerk.addOnLoaded(resolve);
+  });
+};
+
+/**
+ * @internal
+ */
+export const createGetToken = (isomorphicClerk: IsomorphicClerk) => async (options: any) => {
+  await clerkLoaded(isomorphicClerk);
```
SokratisVidros, line 19: ❓Can you please elaborate why such utils are required for `useAuth` since `<ClerkLoaded/>` is used? Is this for SSR?

packages/types/src/clerk.ts
```diff
@@ -13,8 +12,22 @@ export type UnsubscribeCallback = () => void;
 export type BeforeEmitCallback = (
   session: ActiveSessionResource | null,
 ) => void | Promise<any>;
+
 export type SignOutCallback = () => void | Promise<any>;
 
+export type SignOutOptions = {
```
SokratisVidros, line 18: 🙃 It would be great to align naming for method argument types. In same cases we use `FunctionXParams` in others `FunctionXOptions`. I prefer the first. WDYT?

SokratisVidros (changes requested):

packages/clerk-js/src/core/clerk.ts
nikosdouvlis: Thanks Sokrati, will cover everything with unit tests once we agree on the final api

packages/types/src/clerk.ts
nikosdouvlis: Completely agree - and we should try to align every other method  as well (already did that in #99 for many methods)

In this particular case, `signOut()` accepts 2 arguments: an optional `SignOutCallback` function and an optional `SignOutOptions` object so it follows this convention. `SignOutOptions` describes the 2nd argument, not the param list as a whole

packages/react/src/hooks/utils.ts
nikosdouvlis: I'm not exactly sure what you mean by `ClerkLoaded`. Although `ClerkLoaded` can be used, it's certainly not required for `useAuth` to work.

For SSR-enabled apps, the userId and sessionId are always populated. However, during CSR and before clerkjs has loaded, calling signOut or getToken wouldn't behave as expected since the actual behavior is on window.Clerk and not IsomorphicClerk.

This particular hook returns values that can either be initialised during SSR or can be awaited (like the 2 methods). I introduced this `await clerkLoaded` internal helper to allow calling signOut/getToken without the need for isLoaded: both these methods transparently await for clerkjs to load before resolving 

This functionality can be moved to isomorphicClerk in the future and be used in other apis as well

SokratisVidros: Thanks for explaining

packages/clerk-js/src/core/clerk.ts
SokratisVidros: Cool, let's wrap this up tomorrow since the API is set now.

nikosdouvlis: Done!

PR #99, opened 2022-03-13
feat(types): Refactor SignIn and SignUp types

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

In this PR we aligned the signatures of all SignUp, SignIn, EmailAddress, PhoneNumber methods. All methods now accept an `options` object as param as this allows us to add/deprecated arguments in the future without the need for complicated Typescript overloads and value parsing. As an added bonus, using named arguments improves readability for non-TS repos. Some apis became slightly more verbose but given the excellent IDE support for TS types, we don't expect this to be a real issue.

On the type level, we removed old deprecated types and aligned the names of the existing. Most of the types have been refactored and are now simpler to read and actually stricter.

Types that were common between methods are now split into separate type aliases allowing us to make very specific signature adjustments without introducing breaking changes in the future,eg: 

```
// old
  create: (params: SignUpParams) => Promise<SignUpResource>;
  update: (params: SignUpParams) => Promise<SignUpResource>;

// new
  create: (params: SignUpCreateParams) => Promise<SignUpResource>;
  update: (params: SignUpUpdateParams) => Promise<SignUpResource>;
```

You can find more examples about the naming patterns below. Always open to new suggestions :) I hope to create a more detailed "best practises" guide in Notion soon.

#### Example 1: A function that accepts a "bucket" of params, most of our methods follow this pattern
```tsx
// Example function:
await signUp.prepareVerification({ strategy: 'email_link', redirect_url: '...' )

// The above function would be typed as: 
type PrepareVerificationParams = { strategy: EmailLinkStrategy; redirect_url?: string;  }
type PrepareVerificationReturn = Promise<SignUpResource>;
type PrepareVerification: (params: PrepareVerificationParams) => PrepareVerificationReturn;
```

#### Example 2: 1st argument is a callback, 2nd argument is an optional config options object. Some specific Clerk apis follow this pattern
```tsx
// Example function:
await setSession( () => {}, options );

// The above function would be typed as: 
type SetSessionCallback = () => Promise<void>;
type SetSessionOptions = { configOption: ... }
type SetSessionReturn = Promise<void>
type SetSession = (callback: SetSessionCallback, options?: SetSessionOptions ) => SetSessionReturn;
```
We don't always need to create named types, but when we do let's be consistent :)


packages/types/src/factors.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export type PasswordFactor = {
+  strategy: PasswordStrategy;
+};
+
+export type OauthFactor = {
+  strategy: OAuthStrategy;
+};
+
+export type EmailCodeConfig = Omit<EmailCodeFactor, 'safe_identifier'>;
```
SokratisVidros, line 42: ❓ Is this the `EmailCodeConfig` or is it `EmailCodeParams`?

packages/types/src/oauth.ts
```diff
@@ -131,23 +118,19 @@ export function getOAuthProviderData({ provider, strategy }: getOAuthProviderDat
 }
 
 export function sortedOAuthProviders(sortingArray: OAuthStrategy[]) {
-  return OAUTH_PROVIDERS
```
SokratisVidros, line 134: 🙃 Team let's align our linters to avoid such changes in the future.

packages/types/src/providers.ts
```diff
@@ -0,0 +1,36 @@
+export type FacebookOauthProvider = 'facebook';
```
SokratisVidros, line 1: 🔧 @nikosdouvlis Let's split these in to files, one called `Oauth` and another called `web3` so that it's more accurate.

SokratisVidros (changes requested):

packages/types/src/providers.ts
nikosdouvlis: Done, thanks

packages/types/src/factors.ts
nikosdouvlis: Right now the mapping is:
- signIn.firstFactors  -> `EmailCodeFactor | EmailLinkFactor | ...`, used to describe the factor
- signIn.prepareFirstFactor -> `EmailLinkConfig | EmailCodeConfig | ...`, used to select a factor
- signIn.attemptFirstFactor -> `EmailCodeAttempt | EmailLinkAttempt | ...`, used to attempt a factor

But I'm not 100% sure about the `*Config` naming for the 2nd case. Would `EmailCodeFactorSelector` make more sense? Any naming suggestions are welcome :)

SokratisVidros (approved):
🥇

packages/clerk-js/src/core/resources/PhoneNumber.ts
```diff
@@ -4,6 +4,7 @@ import type {
   PhoneNumberResource,
   VerificationResource,
 } from '@clerk/types';
+import { AttemptPhoneNumberVerificationParams, SetReservedForSecondFactorParams } from '@clerk/types/src';
```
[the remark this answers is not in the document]
nikosdouvlis: Woops, webstorm. Fixed

packages/types/src/factors.ts
```diff
@@ -0,0 +1,71 @@
...
+  strategy: EmailLinkStrategy;
+  email_address_id: string;
+  safe_identifier: string;
+};
+
+export type PhoneCodeFactor = {
+  strategy: PhoneCodeStrategy;
+  phone_number_id: string;
+  safe_identifier: string;
+  default?: boolean;
```
gkats, line 26: :question: Where is this `default` property used?

packages/types/src/strategies.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type PasswordStrategy = 'password';
+export type PhoneCodeStrategy = 'phone_code';
+export type EmailCodeStrategy = 'email_code';
+export type EmailLinkStrategy = 'email_link';
+export type TicketStrategy = 'ticket';
+export type OrganizationTicketStrategy = 'ticket';
+
+export type OAuthStrategy = `oauth_${OAuthProvider}`;
+export type Web3Strategy = `web3_${Web3Provider}_signature`;
```
gkats, line 12: :upside_down_face: Perhaps it's better if we add the provider to the type name here as well? `Web3MetamaskStrategy` instead of simply `Web3Strategy`. What happens if we get another web3 provider in the future?

gkats (approved):

PR #105, opened 2022-03-16
Navigation improvements 

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

The default `navigate` function provided by Remix works in 5 phases:
1. calls the route loader
2. replaces the current url
3. waits for the loader to return route data
4. unmounts the old tree
5. finally, it mounts the new page

In our case, SignIn remained mounted up until step 5. Because we couldn't reliably await the navigation, withRedirectToHome in some cases would trigger before the new page is mounted causing a redirect to the home page instead of respecting the redirectUrl prop.

We used the useTransition remix hook to create a custom awaitable navigate function for use with Remix applications. The hook stores all pending navigations and resolves them when the state changes to idle (loading -> idle). 

I'm also considering adding a small delay for extra safety - keep in mind that this delay will only applied to navigations triggered by clerk (after sign in, after sign out etc) so it shouldn't be noticeable at all. 

This PR also includes a beforeUnload fix - we now use listen for the native `beforeUnload` event as well as the custom `clerk:beforeUnload` event (which exists to cover inconsistencies between browsers and only fires when windowNavigate is called). Ideally we don't want to listen for the native `beforeUnload` event always as this disabled the BF cache - to prevent this, we created a `beforeUnloadTracker` which starts tracking just before a potential navigation is going to happen and stops tracking right after. 

PS: We use `beforeUnload` in one more [place](https://github.com/clerkinc/javascript/blob/8b00a66c18924a2a99afdb3539bd6663d68eb673/packages/clerk-js/src/utils/safeLock.ts#L6) but my tests show that this does not affect BF at all, however I need to investigate further to make sure.

Additional context:
- https://developers.google.com/web/updates/2018/07/page-lifecycle-api#the-beforeunload-event
- https://developers.google.com/web/updates/2018/07/page-lifecycle-api#managing-cross-browsers-differences
- https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#usage_notes

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -1,5 +1,6 @@
 import { LocalStorageBroadcastChannel } from '@clerk/shared/utils/localStorageBroadcastChannel';
 import { noop } from '@clerk/shared/utils/noop';
+import { inClientSide } from '@clerk/shared/utils/ssr';
 import type {
   ActiveSessionResource,
   AuthenticateWithMetamaskParams,
@@ -31,7 +32,8 @@ import { AuthenticationService } from 'core/services';
 import { ERROR_CODES } from 'ui/common/constants';
 import {
   appendAsQueryParams,
-  CLERK_BEFORE_UNLOAD_EVENT,
+  createBeforeUnloadTracker,
+  createPageLifecycle,
   ignoreEventValue,
   isAccountsHostedPages,
   isDevOrStagingUrl,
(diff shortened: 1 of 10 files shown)


SokratisVidros:
@nikosdouvlis Does this fix apply only to remix or are there scenarios that we need to fix in @clerk/clerk-react as well?

igneel64 (approved):

colinclerk:
> @nikosdouvlis Does this fix apply only to remix or are there scenarios that we need to fix in @clerk/clerk-react as well?

Before this PR, we only caught unloads when the external navigation was triggered by our own `windowNavigate` function (which would be the case for anyone using our components).

The update here enables us to properly catch the unload event when developers do this manually:
```
setSession("sess_foo", () => {
  window.location.href = "/bar"
});
```

This portion of the PR is not Remix-specific

colinclerk (approved):
lgtm!

PR #115, opened 2022-03-17
feat(clerk-js): Move auth state to request in withServerSideAuth

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Changes the `rootAuthLoader` api from
```tsx
withServerSideAuth(context => {
    const {sessionId, getToken, userId} = context.auth;
    const { user } = context;
    return { props: {} };
  },
  { loadUser: true, loadSession: true },
);
```

to
```tsx
withServerSideAuth(context => {
    const { auth, user, session } = context.req;
    const {sessionId, getToken, userId} = auth;
    return { props: {} };
  },
  { loadUser: true, loadSession: true },
);
```

This change intends to make accessing auth state/ user/ session consistent across all environments and packages.


packages/nextjs/src/middleware/utils/injectAuthIntoRequest.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { AuthData, ContextWithAuth } from '../types';
+
+/**
+ * Inject `auth`, `user` and `session` properties into ctx.request
+ * @internal
+ */
+export function injectAuthIntoRequest(ctx: GetServerSidePropsContext, authData: AuthData): ContextWithAuth {
+  const { user, session, userId, sessionId, getToken } = authData;
+  (ctx.req as any).auth = { userId, sessionId, getToken };
```
nikosdouvlis, line 11: @igneel64 @SokratisVidros 
Instead of making a copy of `ctx` and `req`, I decided to mutate the original and inject the auth state directly - both are class instances and not plain objects so this should be safer. unless I'm wrong, most express middlewares mutate the original req directly so I'm following the same practise here.  Can you think of any scenarios where this could be considered problematic?

igneel64: Not at all, only typings but I don't see any danger on that now.

igneel64 (approved):

PR #118, opened 2022-03-17
fix(clerk-js): Redirect user to signUp url when there is an error with external account sign up

npetridis:
## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description
When attempting to sign up with sso while we have set an allowlist that forbids the specific sso email to sign up then after getting the relative error we are redirected to [signIn page](https://github.com/clerkinc/javascript/blob/0bc7abed4ae686441f87a41ded4012cfa427b44f/packages/clerk-js/src/core/clerk.ts#L539) which doesn't have access to the signUp state to show the proper error.
The fix, after an unsuccessful sso sign up attempt, [redirects to the signUp page](https://github.com/clerkinc/javascript/blob/0bc7abed4ae686441f87a41ded4012cfa427b44f/packages/clerk-js/src/core/clerk.ts#L535-L537) so that the proper error is can be displayed.

https://www.notion.so/clerkdev/SSO-sign-up-w-Allowlist-doesn-t-display-error-58267ed1448d466b811d4d92510e2e7f

### DIFF
--- a/packages/clerk-js/src/utils/url.test.ts
+++ b/packages/clerk-js/src/utils/url.test.ts
@@ -1,4 +1,4 @@
-import { SignUpResource } from '../../../types/dist';
+import { SignUpResource } from '@clerk/types';
 import {
   appendAsQueryParams,
   buildURL,
--- a/packages/clerk-js/src/utils/url.test.ts
+++ b/packages/clerk-js/src/utils/url.test.ts
@@ -1,7 +1,9 @@
+import { SignUpResource } from '../../../types/dist';
 import {
   appendAsQueryParams,
   buildURL,
   getAllETLDs,
+  hasExternalAccountSignUpError,
   isAccountsHostedPages,
   isDevOrStagingUrl,
   trimTrailingSlash,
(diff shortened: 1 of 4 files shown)


packages/clerk-js/src/utils/url.test.ts
```diff
@@ -1,7 +1,9 @@
+import { SignUpResource } from '../../../types/dist';
```
nikosdouvlis, line 1: 🔧 
```suggestion
import { SignUpResource } from '@clerk/types';
```

packages/clerk-js/src/utils/url.ts
```diff
@@ -1,5 +1,6 @@
 import { camelToSnake, isIPV4Address } from '@clerk/shared/utils/string';
 import { loadScript } from 'utils';
+import { SignUpResource } from '../../../types/dist';
```
nikosdouvlis, line 3: 🔧 
```suggestion
import { SignUpResource } from '@clerk/types;
```

SokratisVidros (approved):

PR #120, opened 2022-03-18
Pre commit formatting pub docs

igneel64:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [x] `build/tooling/chore`

## Description

## Chores
- Introduce pre-commit formating on staged files _(let's not solve formating stuff again)_.
- Make the `next` _(or any preid)_ publishing more correct.

We can decide if we at some point want to pass all the js files through the formatter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "husky": "^7.0.4",
     "lerna": "^4.0.0",
     "prettier": "^2.5.1",
+    "pretty-quick": "^3.1.3",
     "rimraf": "^3.0.2",
     "turbo": "^1.1.5",
     "typescript": "^4.6.2"
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx pretty-quick --staged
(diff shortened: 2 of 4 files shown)


gkats (approved):

package.json
```diff
@@ -28,6 +28,7 @@
     "husky": "^7.0.4",
     "lerna": "^4.0.0",
     "prettier": "^2.5.1",
+    "pretty-quick": "^3.1.3",
```
nikosdouvlis, line 31: ❓ Just curious, how come we didn't we go with the standard `lint-staged` util instead? I **think** we might want to run both `eslint --fix` (import ordering etc) and `prettier`  pre-commit

igneel64: @nikosdouvlis 
`pretty-quick` was just selected for the easy integration with husky. _(After the upgrade, there is no need for a lint_staged package.json attribute)_
We can add eslint but I would consider it out of scope for now since it will release dragons.

nikosdouvlis (approved):

PR #131, opened 2022-03-21
feat(clerk-sdk-node): Enable CLERK_JWT_KEY usage from clerk-sdk-node

igneel64:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

- Simple change to allow sdk-node usage of the `CLERK_JWT_KEY` when added from the environment.
- Fix correct apiKey on JWKsClient

FYI on a follow up PR we will make `jwtKey` injectable in the Base contructor making it much more comprehensive.

### DIFF
--- a/packages/sdk-node/src/Clerk.ts
+++ b/packages/sdk-node/src/Clerk.ts
@@ -126,7 +126,7 @@ export default class Clerk extends ClerkBackendAPI {
     this._jwksClient = jwks({
       jwksUri: `${serverApiUrl}/${apiVersion}/jwks`,
       requestHeaders: {
-        Authorization: `Bearer ${defaultApiKey}`,
+        Authorization: `Bearer ${apiKey}`,
       },
       timeout: 5000,
       cache: true,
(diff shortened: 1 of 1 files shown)


packages/sdk-node/src/Clerk.ts
```diff
@@ -158,11 +158,12 @@ export default class Clerk extends ClerkBackendAPI {
 
     /** Base initialization */
 
+    // TODO: More comprehensive base initialization
```
SokratisVidros, line 161: ❓ @igneel64 What do you mean by this comment?

SokratisVidros (approved):

packages/sdk-node/src/Clerk.ts
igneel64: This is added on purpose _(commitment)_ and will be addressed in a PR coming up in a bit. 

Currently the Base contructor is position based `new Base(attr1, attr2, etc..)` which does not make it comprehensive enough on which attribute is which without intellisense.
Also kind of in the same feeling, it will make `jwtKey` injectable.

SokratisVidros: Makes sense.

PR #135, opened 2022-03-23
Expose Avatar With Uploader Component

dvargas92495:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description
Exposes the Avatar With Uploader component to be used outside of the default User Profile.

I'm sure there's probably a more thought out process as to when and how new components are exposed to the API, but I figured I would put up the PR to communicate a feature request. I wanted to use your Profile Image component without having to search `npm` for one (which, I did not find a good one). My approach for the PR was basically command+f-ing `UserProfile` and adapting `AvatarWithUploader` where it made sense.

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -4,6 +4,7 @@ import { inClientSide } from '@clerk/shared/utils/ssr';
 import type {
   ActiveSessionResource,
   AuthenticateWithMetamaskParams,
+  AvatarWithUploaderProps,
   BeforeEmitCallback,
   Clerk as ClerkInterface,
   ClerkOptions,
@@ -238,6 +239,23 @@ export default class Clerk implements ClerkInterface {
   };
 
   public unmountUserButton = (node: HTMLDivElement): void => {
+    this.assertComponentsReady(this.#components);
+    this.#components.unmountComponent({
+      node,
+    });
+  };
+
+  public mountAvatarWithUploader = (node: HTMLDivElement, props?: AvatarWithUploaderProps): void => {
+    this.assertComponentsReady(this.#components);
+    this.#components.mountComponent({
+      name: 'AvatarWithUploader',
+      node,
+      nodeClassName: 'cl-avatar-with-uploader',
+      props,
+    });
+  };
+
+  public unmountAvatarWithUploader = (node: HTMLDivElement): void => {
     this.assertComponentsReady(this.#components);
     this.#components.unmountComponent({
       node,
--- a/packages/clerk-js/src/ui/Components.tsx
+++ b/packages/clerk-js/src/ui/Components.tsx
@@ -28,6 +28,7 @@ import { SignUp, SignUpModal } from './signUp';
 import type { AvailableComponentCtx, AvailableComponentProps } from './types';
 import { UserButton } from './userButton';
 import { UserProfile } from './userProfile';
+import { avatarWithUploader } from "./userProfile/account/avatarWithUploader"
 
 import './styles/clerk.scss';
 
@@ -36,6 +37,7 @@ const AvailableComponents = {
   SignUp,
   UserButton,
   UserProfile,
+  AvatarWithUploader: avatarWithUploader,
 };
 
 type AvailableComponentNames = keyof typeof AvailableComponents;
--- a/packages/react/src/components/uiComponents.tsx
+++ b/packages/react/src/components/uiComponents.tsx
@@ -1,4 +1,4 @@
-import { SignInProps, SignUpProps, UserButtonProps, UserProfileProps } from '@clerk/types';
+import { SignInProps, SignUpProps, UserButtonProps, UserProfileProps, AvatarWithUploaderProps } from '@clerk/types';
 import React from 'react';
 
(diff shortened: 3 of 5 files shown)


dvargas92495:
I can't request review so going to tag @nikosdouvlis and @SokratisVidros for review

SokratisVidros:
Hey @dvargas92495! Thanks a lot for the contribution!

This PR created a very interesting internal debate so I'd like to provide all the necessary context.

Overall we all like the direction of the PR. This is how we envisage ClerkJS in couple of months from now. Our vision is to provide a lot of components and help frontend developers build fast.

Currently ClerkJS offers 4 UI components and numerous Control components that have been already embraced by the community. As you can imagine, every component has an expected maintenance overhead. Moreover, backwards compatibility is a 0-day requirement. 

At this point all the building blocks for four top level Clerk Components are treated as internal components that we can change as much as we have to. Furthermore, in 2-3 weeks we will start a big project to revamp Clerk Components and make them more granular and enable better CSS customizations.

So, we are a bit skeptical whether it's the right time to expose internal components.

I'd like to hear your thoughts as well.

SokratisVidros:
That's an interesting approach. 

If we stick to a dedicated path such as `import {XComponent} from "@clerk/clerk-react/internals` we can definitely give it a shot. 

Couple of things to keep in mind:
- Developers should be advised to used a fixed Clerk version in their package.json if they use any internal or experimental components.
- ClerkJS hotloading need to be tweaked for fixed `@clerk/clerk-react` versions. Since the internal components exist in ClerkJS but are rendered via `@clerk/clerk-react`, making ClerkJS hotloading smarter will prevent potential incompatibilities when a new ClerkJS version is released.

nikosdouvlis:
This is such an interesting discussion! I really like the `experimental` suggestion - this is something that we have discussed internally in the past and I definitely think we should at least explore this path.

An experimental module, such as `import { useAuthStateMachine } from '@clerk/react/experimental` will let us experiment with new APIs without the  overhead of managing a separate `canary` release tag - that applies both for us and for our users.

Making the community part of the API design process is really important for us. I feel that the reduced release overhead and the ability to introduce breaking changes for experimental APIs will do just that: it will create a tighter feedback loop between Clerk and the community.

dvargas92495:
> Developers should be advised to used a fixed Clerk version in their package.json

Absolutely agree. In some sense, my current approach is pasting a fixed version of what I copied

> ClerkJS hotloading

Any reason why you guys opt for this approach instead of the traditional importing of components? This would make another one of my eventual requests easier, which is true SSR of Clerk components

> will let us experiment with new APIs without the overhead of managing a separate `canary` release tag

In my previous company, managing a "canary" or "staging" branch was an absolute nightmare. Since then, I've loved the approach of pushing straight to live. As you said you can get a tighter feedback loop from users.

So it sounds like we're on board! should I iterate on this PR now or wait for you guys to setup the external/internal module to establish the desired pattern?

SokratisVidros:
@dvargas92495 The main reason we picked ClerkJS hot-loading is the ability to ship bug fixes and feature updates to all our customers without requiring them to bump their NPM modules and redeploy. 

Our hot-loading followings [semantic versioning](https://semver.org/). That is minor and patch fixes are always hotloading. Breaking changes are always part of a major release. For example our latest `@clerk/clerk-react@3` hot loads all `@clerk/clerk-js@3` versions.

Moreover, `@clerk/clerk-js` is still a fat package, served a cached by CDNs. We believe it's better to add minimal overhead to the application bundle and load it on runtime as we plan to split it into chunks soon.

As for the `experimental` path, let me talk to the team and I will get back to you in 2-3 days.

SokratisVidros:
PTAL https://stackblitz.com/edit/clerk-react-no-hotloading

dvargas92495:
Oh that's sweet and super simple to use, thanks for sharing!

dvargas92495:
Looks like the Remix Clerk Provider is missing the `Clerk` prop from its parameter props: https://github.com/clerkinc/javascript/blob/main/packages/remix/src/client/RemixClerkProvider.tsx#L10-L13

React: 
https://github.com/clerkinc/javascript/blob/main/packages/react/src/contexts/ClerkProvider.tsx#L13

dvargas92495:
And then trying this approach within Clerk Remix results in:

`ReferenceError: self is not defined`

Not sure what could be going on here, could be some weird exporting issue

SokratisVidros:
@dvargas92495 great catch. We will push a fix ASAP.

SokratisVidros:
@dvargas92495 FYI: https://github.com/clerkinc/javascript/pull/166

dvargas92495:
Going to close this PR - The component changes I've heard about should start achieving this!

PR #145, opened 2022-03-24
feat(types,clerk-js,backend-core,clerk-react): Replace thrown error w…

nikosdouvlis:
…ith null return in getToken

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

After discussing internally, we all agreed that our main target should be TS cobases so our API design should be optimised for that.

So instead of throwing an error if getToken were to be called without an active session, we change the getToken return value to Promise<string | null>

### DIFF
--- a/packages/backend-core/src/util/createGetToken.ts
+++ b/packages/backend-core/src/util/createGetToken.ts
@@ -20,14 +20,14 @@ type CreateGetToken = (params: {
  */
 export const createGetToken: CreateGetToken = params => {
   const { cookieToken, fetcher, headerToken, sessionId } = params || {};
-  return (options: ServerGetTokenOptions = {}) => {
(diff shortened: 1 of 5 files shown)


igneel64 (approved):

SokratisVidros (commented):
+1

PR #151, opened 2022-03-28
fix(backend-core): Throw on missing JWT key

igneel64:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

When an error was thrown in the Networkless verification flow, due to a missing `CLERK_JWT_KEY` environment variable and no injected `jwtKey` available, the error was hidden behind a signed out state and without any error message.

1. Add an understandable message on what is going wrong.
2. Throw as it was originally designed.

/* Consolidated a bit the logic
/* More description on the ticket

https://www.notion.so/clerkdev/Throw-error-on-missing-JWT-key-fe68711c48cd4c41a855ac1b311a3f80


packages/backend-core/src/types/core.ts
```diff
@@ -0,0 +1,59 @@
+import { JWTPayload } from './jwt';
+
+export enum AuthStatus {
+  SignedIn = 'Signed in',
```
SokratisVidros, line 4: 🙃 Does the status have to be human friendly?

packages/backend-core/src/types/core.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export type AuthState = {
+  status: AuthStatus;
+  session?: Session;
+  interstitial?: string;
+  sessionClaims?: JWTPayload;
+};
+
+export type AuthStateParams = {
```
SokratisVidros, line 32: 💯

packages/backend-core/src/util/errors.ts
```diff
@@ -0,0 +1,13 @@
+export class JWTExpiredError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'JWTExpiredError';
```
SokratisVidros, line 4: 🔧 When extending JS errors this: `Object.setPrototypeOf(this, JWTExpiredError.prototype);` is usually required.

SokratisVidros (approved):

npetridis (approved):
Good job with the cleanup 🏅

packages/backend-core/src/util/errors.ts
igneel64: @SokratisVidros 
Great reminder!
This is a critical mishap if the target of our compilation was ES5. Currently we are using the latest target for Node 12 which is ES2019 _(we should upgrade btw to ES2020)._ In this case we are safe.

PR #158, opened 2022-03-29
Prevent re-initialization of Clerk singletons

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

This PR prevents Clerk from breaking if `IsomorphicClerk.loadClerkJS` is called more than once after Clerk loaded and initialised.

`ClerkProvider` needs to be a top-level component component, wrapping most of the user's app. It should only mount *once* during the lifetime of the app, however this is something we cannot control - it is possible under certain conditions that 3rd party frameworks (eg NextJS, Remix, etc) will trigger a complete unmount-remount of the whole root component, causing effects in `ClerkProvider` to run again.

In this particular case, `ClerkProvider` was creating a new instance of `IsomorphicClerk` that was in turn loading and replacing the window.Clerk singleton, causing it to lose state.

There are a number of changes that we could do to further improve this flow. Another potential cause of bugs is the [module-level initialisation of window.Clerk](https://github.com/clerkinc/javascript/blob/276ef8e7a332f8604ef9956af6386db14187a68a/packages/clerk-js/src/index.browser.ts#L12) - I will share more thoughts on a notion page so we can discuss further.

### DIFF
--- a/packages/react/src/utils/scriptLoader.ts
+++ b/packages/react/src/utils/scriptLoader.ts
@@ -11,34 +11,45 @@ const FAILED_TO_LOAD_ERROR = 'Clerk: Failed to load Clerk';
 const MISSING_PROVIDER_ERROR = 'Clerk: Missing provider';
 const MISSING_BODY_ERROR = 'Clerk: Missing <body> element.';
 
-function isStaging(frontendApi: string): boolean {
+const UNSTABLE_RELEASE_TAGS = ['staging'];
+
+const extractNonStableTag = (packageVersion: string) => {
+  const tag = packageVersion.match(/-(.*)\./)?.[1];
+  return tag && UNSTABLE_RELEASE_TAGS.includes(tag) ? tag : undefined;
+};
+
+const extractMajorVersion = (packageVersion: string) => {
+  return packageVersion.split('.')[0];
+};
+
+const forceStagingReleaseForClerkFapi = (frontendApi: string): boolean => {
   return (
     frontendApi.endsWith('.lclstage.dev') ||
     frontendApi.endsWith('.stgstage.dev') ||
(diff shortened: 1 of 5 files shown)


SokratisVidros (approved):

npetridis (approved):

packages/react/src/utils/scriptLoader.ts
```diff
@@ -11,34 +11,45 @@ const FAILED_TO_LOAD_ERROR = 'Clerk: Failed to load Clerk';
 const MISSING_PROVIDER_ERROR = 'Clerk: Missing provider';
 const MISSING_BODY_ERROR = 'Clerk: Missing <body> element.';
 
-function isStaging(frontendApi: string): boolean {
+const NON_STABLE_RELEASE_TAGS = ['staging'];
```
SokratisVidros, line 14: 🙃 Could be just `UNSTABLE_RELEASE_TAGS`

SokratisVidros (approved):

igneel64 (approved):

packages/react/src/utils/scriptLoader.ts
nikosdouvlis: makes sense :) done

PR #163, opened 2022-04-01
feat(backend-core): Verification errors and errorReason attribute

igneel64:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend-core`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Addition of the `Auth-Result` response header on signed-out and interstitial states.
_Feedback expected!_

Notable changes:
- Check the `mapErrorReasonToResponse` which drives which `verifySessionToken` errors lead to signed out states vs interstitial.

https://www.notion.so/clerkdev/Add-observability-of-getAuthState-decisions-38752519809d417c9a25ca40b6d6d4e8


packages/backend-core/src/util/errors.ts
```diff
@@ -1,13 +1,60 @@
...
-    super(message);
-    this.name = 'MissingJWTVerificationKeyError';
+/**
+ * The JWT verification process does not need to know about the type of token (header, cookie etc.)
+ * To translate the generic errors to mapped AuthErrorReason header responses we map them based on the token type.
+ *
+ * Here we also steer the flow based on the verification error.
+ * `shouldSignout` set as true will lead to a signout, otherwise we throw the interstitial
+ *
+ * TODO: Can be probably done better with template literals
```
igneel64, line 19: FYI due to the late hours, my mind was not in capable state to add some smarts to reduct this switch.

colinclerk: lgtm - sometimes the unreduced is more readable

colinclerk (commented):
I think setting `Auth-Result` for `withServerSideAuth` was missed.

I'm incredibly excited to have this in place for future debugging. Really great stuff - thanks Peter!

colinclerk (commented):
Noticed a disparity.

Currently `InlineKeyInvalid` is defined but never thrown.

This was intended to be used if `importKeyFunction` throws during `verifySessionToken`:
```
return this.importKeyFunction(jwk, algorithm);
```

packages/backend-core/src/types/errors.ts
```diff
@@ -0,0 +1,45 @@
+/* Expected values for the 'Auth-Result' response header */
+export enum AuthErrorReason {
+  KeyMissing = 'key-missing',
+  InlineKeyMissing = 'inline-key-missing',
+  InlineKeyInvalid = 'inline-key-invalid',
+  HeaderExpired = 'header-expired',
+  HeaderEarly = 'header-early',
+  HeaderInvalid = 'header-invalid',
+  HeaderInvalidIssuer = 'header-invalid-issuer',
+  HeaderKeyMismatch = 'header-key-mismatch',
```
colinclerk, line 10: It looks like you introduced "HeaderUnverified" instead of using "HeaderKeyMismatch"

I suggested the name "HeaderKeyMismatch" and "CookieKeyMismatch" because when verification fails, we don't know which mistake the developer made. We've seen both of:
1. They're using a dev instance Frontend API but a prod instance JWT Key
2. They're using a prod instance Frontend API but a dev instance JWT Key

The final possibility is that we're dealing with an attacker.

I thought "mismatch" was a polite way to describe the possible developer error.

I still lean toward "KeyMismatch" but happy to go either way.  We should remove the one we don't choose from the `AuthErrorReason` enum

packages/backend-core/src/types/errors.ts
```diff
@@ -0,0 +1,45 @@
...
+  CookieMissing = 'cookie-missing',
+  CookieExpired = 'cookie-expired',
+  CookieEarly = 'cookie-early',
+  CookieInvalid = 'cookie-invalid',
+  CookieInvalidIssuer = 'cookie-invalid-issuer',
+  CookieKeyMismatch = 'cookie-key-mismatch',
+  CookieOutDated = 'cookie-outdated',
+  CookieUnauthorizedParty = 'cookie-unauthorized-party',
+  CookieUnverified = 'cookie-unverified',
+  OriginUnauthorized = 'origin-unauthorized',
```
colinclerk, line 28: This is unused (probably in favor of the UnauthorizedParty variants)

packages/backend-core/src/Base.ts
```diff
@@ -68,10 +70,17 @@ export class Base {
...
-      !jwtKey && this.loadCryptoKeyFunction
-        ? await this.loadCryptoKeyFunction(token)
-        : await this.loadCryptoKey(jwtKey || process.env.CLERK_JWT_KEY);
+
+    let availableKey: CryptoKey;
+    if (!jwtKey && this.loadCryptoKeyFunction) {
+      try {
+        availableKey = await this.loadCryptoKeyFunction(token);
+      } catch (_) {
+        throw new TokenVerificationError(TokenVerificationErrorReason.PublicKeyFetchError);
```
colinclerk, line 79: Is it possible to end up here if `CLERK_API_KEY` is not set?

If so, can we differentiate between `KeyMissing` and `PublicKeyFetchError`?

igneel64:
@colinclerk 

- `withServerSideAuth` is working in this implementation with the addition on the `getAuthData` function. https://github.com/clerkinc/javascript/pull/163/files#diff-e544579dd5d60c6235edd5d097bdc3441fb8930f578080cf03cef6b11da5961bR35

- The one missing is `withEdgeMiddlewareAuth` which we cannot pre-modify the response afaik, because we just passthrough for now. We can enrich the request object though for the route or next middleware to use.

> Currently InlineKeyInvalid is defined but never thrown.
Fixed, didn't get the definition right away.

> I still lean toward "KeyMismatch" but happy to go either way. We should remove the one we don't choose from the AuthErrorReason enum
I now get it, I see your point. No strong opinion on that, but in the case the the `verifySignature` function fails, I would prefer for us to directly associate it with verification and have the comment of mismatch on our side.

> Is it possible to end up here if CLERK_API_KEY is not set?
We have left this case on the clients of backend-core e.g. as `clerk-sdk-node` and `nextjs`. In both of them we throw either at runtime _(sdk-node instance type)_ or Next.js at build time.

WDYT ?

colinclerk:
Ah, sorry - the "middleware" in the filename confused me.

It looks like we *can* modify response headers from Next.js edge middleware. The example here sets a hello=world cookie:
https://nextjs.org/docs/api-reference/next/server#public-methods

If I understand right, this would set a header:
```
export function middleware(request: NextRequest) {
  let response = NextResponse.next()
  response.headers.set("Auth-Result", "foo")
  return response
}
```

We can do this in a separate PR, though

> I would prefer for the code to directly associate it with verification and have the comment of mismatch on our sid

Makes sense. Can we do `HeaderVerificationFailed` and `CookieVerificationFailed`? Then delete the unused.

>We have left this case on the clients of backend-core e.g. as clerk-sdk-node and nextjs. In both of them we throw either at runtime (sdk-node instance type) or Next.js and default sdk-node at build time.

Build time should definitely be okay. If we ever hear a complaint about observability of the runtime error, we can revisit.

Thank you!

igneel64:
> Can we do HeaderVerificationFailed and CookieVerificationFailed?
Done

> It looks like we can modify response headers from Next.js edge middleware

Well it is a bit more complicated than that as the middleware does not work as we would normally expect. Laying down what I have tested before:
1. Returning the `response` will not allow any other code to run and just serve the modified response with the page content.
```ts
export default function withEdgeMiddlewareAuth(function(){
   console.log("If NextResponse.next() has been returned from withEdgeMiddlewareAuth, I will never run.");
  
  // For the sake of the example this is just a passthrough
  return NextResponse.next();
});
```

2. For the interstitial, we can add the header just with one line. For the signed out state, it means we would not allow the code inside the handler function to run (point 1), but the header will be visible to the response.

colinclerk:
Perfect - we can take a deeper look at middleware another time.  Thank you Peter!

colinclerk (approved):

SokratisVidros (approved):
💯 Nicely done folks. 

A future step is to generate an errors page explaining what these errors code mean to assist our users while debugging.

PR #164, opened 2022-04-02
Adding a no cache cache control header to solve infinite loop issues

dvargas92495:
## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

     - Is it possible to have the package allow Node 14? AWS Lambda's latest version is Node 14 so that's what I run locally.

When Clerk returns an interstitial and the developer has caching implemented, it will continually redirect bc cached versions will continue to load the interstitial. Adding a "no cache" Cache Control header solved it on my end, which basically tells my CDN to not keep the interstitial in the cache. I think this could be a reasonable default, or at the very least, a parameter we could pass into `rootAuthLoader`

### DIFF
--- a/packages/remix/src/ssr/rootAuthLoader.ts
+++ b/packages/remix/src/ssr/rootAuthLoader.ts
@@ -33,7 +33,10 @@ export const rootAuthLoader: RootAuthLoader = async (
   const { authData, showInterstitial } = await getAuthData(args.request, opts);
 
   if (showInterstitial) {
-    throw json(wrapClerkState({ __clerk_ssr_interstitial: showInterstitial, __frontendApi: frontendApi }));
+    throw json(
+      wrapClerkState({ __clerk_ssr_interstitial: showInterstitial, __frontendApi: frontendApi }), 
+      {headers: {"Cache-Control": "no-cache"}}
+    );
   }
 
   if (!callback) {
--- a/packages/remix/src/ssr/rootAuthLoader.ts
+++ b/packages/remix/src/ssr/rootAuthLoader.ts
@@ -35,7 +35,7 @@ export const rootAuthLoader: RootAuthLoader = async (
   if (showInterstitial) {
     throw json(
       wrapClerkState({ __clerk_ssr_interstitial: showInterstitial, __frontendApi: frontendApi }), 
-      {headers: {"Cache-Control": "no-cache"}}
+      { status: 401 }
     );
   }
 
--- a/packages/remix/src/ssr/getAuth.ts
+++ b/packages/remix/src/ssr/getAuth.ts
@@ -15,7 +15,7 @@ export async function getAuth(argsOrReq: Request | LoaderFunctionArgs): GetAuthR
   const { authData, showInterstitial } = await getAuthData(request);
 
   if (showInterstitial || !authData) {
-    throw json(EMPTY_INTERSTITIAL_RESPONSE);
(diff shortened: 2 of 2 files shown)


dvargas92495:
ccing @nikosdouvlis for review

nikosdouvlis:
Very interesting. @dvargas92495 Thank you very much for the effort you put into this. Let me also cc @colinclerk  and @igneel64 so we can discuss.

colinclerk:
Makes sense to me!

@dvargas92495 to confirm, you’re seeing the interstitial come through with a 401 and it’s still being cached by default?

We chose 401 instead of 200 in part expecting it wouldn’t be cached - I don’t see a problem with adding the header to further emphasize it shouldn’t be cached.

dvargas92495:
It's using Remix's `json` method, which is a `Response` object that returns a 200. I think returning a 401 instead should also work

SokratisVidros (approved):

packages/remix/src/ssr/rootAuthLoader.ts
```diff
@@ -33,7 +33,10 @@ export const rootAuthLoader: RootAuthLoader = async (
   const { authData, showInterstitial } = await getAuthData(args.request, opts);
 
   if (showInterstitial) {
-    throw json(wrapClerkState({ __clerk_ssr_interstitial: showInterstitial, __frontendApi: frontendApi }));
+    throw json(
+      wrapClerkState({ __clerk_ssr_interstitial: showInterstitial, __frontendApi: frontendApi }), 
+      {headers: {"Cache-Control": "no-cache"}}
```
nikosdouvlis, line 38: As discussed and in order to be consistent across similar implementations (Nextjs), instead of adding a `Cache-Control` header explicitly, shall we change the status of this response to `401`?

```tsx
    throw json(wrapClerkState({ __clerk_ssr_interstitial: showInterstitial, __frontendApi: frontendApi }), {
      status: 401,
    });
```

The same change should be made to: https://github.com/clerkinc/javascript/blob/acb818ce77b036690845763a45471ebf5229de49/packages/remix/src/ssr/getAuth.ts#L18

I'd be happy to do the change myself, but I don't want to hijack your PR :)

nikosdouvlis (commented):
Let me provide some more context in case it's helpful. We went with `200` responses during the first beta implementation of `@clerk/remix`; any thrown `401` would conflict with nested catch/error boundaries depending on each application's structure.

We always try to provide the best DX possible, so we wanted to keep the initial. To minimize the initial setup and make the integration as transparent as possible, we were reluctant to refresh expired tokens using catch boundaries. However, edge cases similar to the infinite loop you experienced led us to the `ClerkCatchBoundary` solution. 

After the `ClerkCatchBoundary` change, the interstitial response status should've been changed to `401`.

igneel64 (approved):

nikosdouvlis:
This is now merged, @dvargas92495 thanks for the contribution :)

PR #172, opened 2022-04-05
Add support for Android WebOTP

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

(draft PR, some minor cleanup is needed)
This PR adds support for WebOTP on Android devices.

In order for WebOTP to work, the SMS message [must follow a very specific format](https://wicg.github.io/web-otp/#intro-server) and it looks like our does follow this format. However, after testing the behavior was at best inconsistent - sometimes the popup wasn't getting triggered at all, other times the popup did open but the WebOTP promise never resolved (it was rejecting after a 4-min timeout).

After many tries, I've found that this message will always fail:
```
606174 is your Clerk Staging verification code and it will expire in 10 minutes. Do not share this with anyone.

@accounts.clerkstage.dev #606174
```

but these two will always work: 
```
614261 is your Clerk Staging verification code. It will expire in 10 minutes

@accounts.clerkstage.dev #614261
```
```
614261 is your Clerk Staging verification code; It will expire in 10 minutes.

@accounts.clerkstage.dev #614261
```

It looks like that if the text before the last line contains more than 1 dot (.), the WebOTP will open but the promise will never resolve. The [spec says that webOTP supports iframes](https://wicg.github.io/web-otp/#sctn-iframe-guidance) but chromium is limited to webOTP calls [from cross-origin iframes that have no more than one unique origin in its ancestor chain](https://web.dev/web-otp/#:~:text=At%20the%20moment%20Chrome%20only%20supports%20WebOTP%20API%20calls%20from%20cross%2Dorigin%20iframes%20that%20have%20no%20more%20than%20one%20unique%20origin%20in%20its%20ancestor%20chain.%20In%20the%20following%20scenarios%3A)... It could be that Chromium uses (.) to parse origin ancestor chain - I will try to go through chromium's repo to validate this but I didn't have a chance to do it yet (this is just a guess, I could be very wrong).

All in all, it looks like this PR can get WebOTP to work consistently across Android devices that support it, if we slightly change our SMS message.

I might be missing something, so any feedback is welcome :)

Quick demo:
Screenshot 2022-04-06 at 12 26 19 AM
Screenshot 2022-04-06 at 12 26 27 AM

### DIFF
--- a/packages/shared/components/oneTimeCodeInput/OneTimeCodeInput.tsx
+++ b/packages/shared/components/oneTimeCodeInput/OneTimeCodeInput.tsx
@@ -7,6 +7,7 @@ import { Input } from '../input';
 import { Spinner } from '../spinner';
 // @ts-ignore
 import styles from './OneTimeCodeInput.module.scss';
+import { useWebOTP } from './useWebOTP';
 
 const BLANK_CHAR = '';
 const VERIFY_SUCCESS_ACTION_DELAY = 500;
@@ -46,6 +47,7 @@ export function OneTimeCodeInput({
   resendCodeButtonClassName,
   ...rest
 }: OneTimeCodeInputProps): JSX.Element {
+  useWebOTP({ onOtpReceived: handleWebOtp });
   const inputValues = Array.from({ length }, (_, i) => value[i] || BLANK_CHAR);
   const [activeInputIndex, setActiveInputIndex] = React.useState(0);
   const [canResendCode, setCanResendCode] = React.useState(true);


packages/shared/components/oneTimeCodeInput/useWebOTP.ts
```diff
@@ -0,0 +1,57 @@
...
+  const abortControllerRef = React.useRef<AbortController>();
+  const isMountedRef = React.useRef<boolean>(true);
+
+  React.useEffect(() => {
+    if (!inClientSide() || !('OTPCredential' in window)) {
+      return;
+    }
+
+    abortControllerRef.current = new AbortController();
+    console.log('this starts running');
```
SokratisVidros, line 33: 🙃  Please remove

packages/shared/components/oneTimeCodeInput/useWebOTP.ts
```diff
@@ -0,0 +1,57 @@
...
+    void navigator.credentials
+      .get({
+        otp: { transport: ['sms'] },
+        signal: abortControllerRef.current.signal,
+      })
+      .then(otp => {
+        console.log('credential code', otp, otp?.code);
```
SokratisVidros, line 40: 🙃 Ditto

packages/shared/components/oneTimeCodeInput/useWebOTP.ts
```diff
@@ -0,0 +1,57 @@
...
+        }
+      })
+      .catch(e => {
+        console.log('webotp exception: ', e);
+      });
+
+    // return cleanup;
+  }, []);
+
+  // const cleanup = () => {
```
SokratisVidros, line 52: ❓ Should the cleanup be reverted?

SokratisVidros (commented):
@nikosdouvlis Changing the message requires an API change. If not mistaken the `Do not share this with anyone` part needs to be removed. Shall we proceed as such?

PR #174, opened 2022-04-07
feat(remix): Add Auth-Result response header

igneel64:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Add the `Auth-Result` response header on Remix.

### DIFF
--- a/packages/remix/src/ssr/rootAuthLoader.ts
+++ b/packages/remix/src/ssr/rootAuthLoader.ts
@@ -30,17 +30,20 @@ export const rootAuthLoader: RootAuthLoader = async (
   const frontendApi = process.env.CLERK_FRONTEND_API || opts.frontendApi;
   assertFrontendApi(frontendApi);
 
-  const { authData, showInterstitial } = await getAuthData(args.request, opts);
+  const { authData, showInterstitial, errorReason } = await getAuthData(args.request, opts);
 
   if (showInterstitial) {
-    throw json(
-      wrapClerkState({ __clerk_ssr_interstitial: showInterstitial, __frontendApi: frontendApi }), 
-      { status: 401 }
-    );
+    throw json(wrapClerkState({ __clerk_ssr_interstitial: showInterstitial, __frontendApi: frontendApi }), {
+      status: 401,
+      headers: { 'Auth-Result': errorReason || '' },
+    });
   }
 
   if (!callback) {
-    return { ...wrapClerkState({ __clerk_ssr_state: authData, __frontendApi: frontendApi }) };
+    return json(
+      { ...wrapClerkState({ __clerk_ssr_state: authData, __frontendApi: frontendApi }) },
+      { headers: { 'Auth-Result': errorReason || '' } },
+    );
   }
 
   const callbackResult = await callback?.(injectAuthIntoRequest(args, sanitizeAuthData(authData!)));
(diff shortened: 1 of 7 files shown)


packages/remix/src/ssr/rootAuthLoader.ts
```diff
@@ -54,5 +57,8 @@ export const rootAuthLoader: RootAuthLoader = async (
     throw new Error(invalidRootLoaderCallbackResponseReturn);
   }
 
-  return { ...callbackResult, ...wrapClerkState({ __clerk_ssr_state: authData, __frontendApi: frontendApi }) };
+  return json(
+    { ...callbackResult, ...wrapClerkState({ __clerk_ssr_state: authData, __frontendApi: frontendApi }) },
+    { headers: { 'Auth-Result': errorReason || '' } },
```
SokratisVidros, line 62: ❓ https://clerkinc.slack.com/archives/C01KTTYKWBU/p1649399218490019

packages/remix/src/ssr/utils.ts
```diff
@@ -75,3 +66,46 @@ export function assertObject(val: any, error?: string): asserts val is Record<st
...
+
+/**
+ * @internal
+ */
+export const throwInterstitialJsonResponse = (opts: { frontendApi: string; errorReason: string | undefined }) => {
+  throw json(
+    wrapWithClerkState({
+      __clerk_ssr_interstitial: true,
+      __frontendApi: opts.frontendApi,
+      __lastAuthResult: opts.errorReason,
```
igneel64, line 78: ❓ 
What do you think about changing it to `__errorReason` to reduce the cognitive load on going through the flow ?

packages/remix/src/ssr/rootAuthLoader.ts
```diff
@@ -30,20 +36,17 @@ export const rootAuthLoader: RootAuthLoader = async (
...
 
-  const { authData, showInterstitial } = await getAuthData(args.request, opts);
+  const { authData, showInterstitial, errorReason } = await getAuthData(args.request, opts);
 
   if (showInterstitial) {
-    throw json(
-      wrapClerkState({ __clerk_ssr_interstitial: showInterstitial, __frontendApi: frontendApi }), 
-      { status: 401 }
-    );
+    throw throwInterstitialJsonResponse({ frontendApi, errorReason });
```
igneel64, line 42: ❓ 
Is the `throw` needed here since `throwInterstitialJsonResponse` throws on its own ?

nikosdouvlis: You're right, it's not **needed** per se, but I used it to further emphasize that we need to **throw** in order for the catch boundary to work - otherwise the SSR flow will break. It also helps TS infer that the execution stops there if the if is true, without using `return`

I was worried that wrapping the `throwInterstitialJsonResponse` into another function that uses a try/catch block during a future refactor could potentially break the flow.

Afaik immediately throwing a thrown error works as expected - but let me know if you have any objections

packages/remix/src/ssr/utils.ts
nikosdouvlis: Valid point! I think we should go with something like `__lastAuthResult` or similar to be consistent with what we do in NextJS

PR #176, opened 2022-04-11
feat(types,clerk-js): Allow users to delete external accounts

chanioxaris:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Allow users to delete a connected external account

### DIFF
--- a/packages/clerk-js/src/core/resources/ExternalAccount.test.ts
+++ b/packages/clerk-js/src/core/resources/ExternalAccount.test.ts
@@ -1,0 +1,25 @@
+import { BaseResource, ExternalAccount } from 'core/resources/internal';
+
+describe('External account', () => {
+  it('destroy', async () => {
+    const targetId = 'test_id';
+
+    const deletedObjectJSON = {
+      object: 'external_account',
+      id: targetId,
+      deleted: true,
+    };
+
+    // @ts-ignore
+    BaseResource._fetch = jest.fn().mockReturnValue(Promise.resolve({ response: deletedObjectJSON }));
+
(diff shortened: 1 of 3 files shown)


packages/clerk-js/src/core/resources/User.ts
```diff
@@ -132,6 +132,10 @@ export class User extends BaseResource implements UserResource {
     return new ExternalAccount(json, this.path() + '/external_accounts');
   };
 
+  deleteExternalAccount = async ({ externalAccountId }: { externalAccountId: string }): Promise<void> => {
```
SokratisVidros, line 135: 🔧 @chanioxaris Since we don't have a `deleteEmailAddress` or `deletePhoneNumber` method on the user object, it's fine to omit it for external accounts too. Instead, users should be able to do `user.externalAccounts[0].destroy()`.

SokratisVidros (approved):
🎖️

packages/clerk-js/src/core/resources/User.ts
chanioxaris: Yes you are right, no need to introduce this wrapper on the user resource. Has been removed

PR #177, opened 2022-04-11
feat(clerk-js): OAuth with ticket

gkats:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Modified the sign up flow in the SignUpStart component when a ticket is present in the query parameters, but the sign up cannot be completed. 

We check if there's any social login method available and render it. When the user then clicks on the social login provider button, we'll check if there's an existing sign up object and opt to update it instead of creating a new one.

### DIFF
--- a/packages/clerk-js/src/core/resources/SignUp.ts
+++ b/packages/clerk-js/src/core/resources/SignUp.ts
@@ -167,7 +167,9 @@ export class SignUp extends BaseResource implements SignUpResource {
 
   public authenticateWithRedirect = async (params: AuthenticateWithRedirectParams): Promise<void> => {
     const { redirectUrl, redirectUrlComplete, strategy } = params || {};
-    const { verifications } = await this.create({
+    const authenticateFn = (args: SignUpCreateParams | SignUpUpdateParams) =>
+      this.id ? this.update(args) : this.create(args);
+    const { verifications } = await authenticateFn({
       strategy,
       redirectUrl,
       actionCompleteRedirectUrl: redirectUrlComplete,
--- a/packages/clerk-js/src/ui/signUp/SignUpStart.tsx
+++ b/packages/clerk-js/src/ui/signUp/SignUpStart.tsx
@@ -38,6 +38,7 @@ function _SignUpStart(): JSX.Element {
   const [emailOrPhoneActive, setEmailOrPhoneActive] = React.useState<ActiveIdentifier>('emailAddress');
   const signUp = useCoreSignUp();
   const [isLoading, setIsLoading] = React.useState(false);
+  const [isMissingRequirements, setMissingRequirements] = React.useState(false);
   const formFields = {
     firstName: useFieldState('first_name', ''),
     lastName: useFieldState('last_name', ''),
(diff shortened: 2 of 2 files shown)


packages/clerk-js/src/ui/signUp/SignUpStart.tsx
```diff
@@ -58,6 +59,9 @@ function _SignUpStart(): JSX.Element {
   const fields = determineFirstPartyFields(environment, hasTicket);
   const oauthOptions = userSettings.socialProviderStrategies;
   const web3Options = userSettings.web3FirstFactors;
+  const isEmailAddressUsedForIdentification = !!userSettings.enabledFirstFactorIdentifiers.find(
```
igneel64, line 62: 🙃 
For a boolean return you can use `Array.prototype.some`

packages/clerk-js/src/ui/signUp/SignUpStart.tsx
```diff
@@ -252,7 +258,7 @@ function _SignUpStart(): JSX.Element {
   ) : null;
 
   const shouldShowEmailAddressField =
```
igneel64, line 260: 🙃 
At a later point or now where we have the feature fresh, I would extract this check to another file or at the end of this one and comment on the cases which this extensive boolean check applies to.

igneel64 (approved):

SokratisVidros (approved):

packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -167,7 +167,8 @@ export class SignUp extends BaseResource implements SignUpResource {
 
   public authenticateWithRedirect = async (params: AuthenticateWithRedirectParams): Promise<void> => {
     const { redirectUrl, redirectUrlComplete, strategy } = params || {};
-    const { verifications } = await this.create({
+    const authenticateFn = this.id ? this.update : this.create;
```
nikosdouvlis, line 170: 🙃 
I'd write this as:
```suggestion
    const authenticateFn = (...args) => this.id ? this.update(...args) : this.create(...args);
```
`authenticateFn` does not run with the same context (`this`) as `this.update` and `this.create` do. 

Note that in this particular case, it's totally fine because `create` and `update` are declared as arrow functions so they do retain their context even if called without it (eg: `this.create` vs `create()`) but if the declaration of these function changes in a future refactor, this could break

nikosdouvlis (approved):
Looks good!

igneel64 (approved):

PR #186, opened 2022-04-13
feat(clerk-js): Make sign up flow resumable

yourtallness:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

When a `sign_up` has a status of `missing_requirements`, the callback handling will redirect to a new `/sign-up/continue` route that renders a new `SignUpContinue` component which is similar to `SignUpStart`, with the difference that it will try to resume an existing SU instead of creating a new one.

## Related issue

https://www.notion.so/clerkdev/a853f869f69c414b85e6d5290a2c4172?v=531db68c98394e709bafd5b5e1df4add&p=312cec8edb5b431b9b65b6a6e14b8d50

## Screenshot

image


packages/clerk-js/src/ui/signUp/SignUpStart.tsx
```diff
@@ -153,7 +159,9 @@ function _SignUpStart(): JSX.Element {
 
     try {
       setError(undefined);
-      const res = await signUp.create(buildRequest(reqFields));
+      const req = buildRequest(reqFields);
+      const promise = continueExisting ? signUp.update(req) : signUp.create(req);
```
SokratisVidros, line 163: The component should be smart enough to pick between create or update if signUp has an id. I'd also suggest to introduce a signUp.save method on the resource model that will handle this conditional.

packages/clerk-js/src/ui/common/withRedirectToHome.tsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import { useCoreSession, useEnvironment } from 'ui/contexts';
 import { useNavigate } from 'ui/hooks';
 
-export function withRedirectToHome<P extends SignInProps | SignUpProps>(
+export function withRedirectToHome<P extends SignInProps | SignUpProps | { continueExisting?: boolean }>(
```
SokratisVidros, line 6: @yourtallness please check my comment below about removing `{ continueExisting?: boolean }` prop.

packages/clerk-js/src/ui/signUp/SignUpStart.tsx
yourtallness: You can have a previous sign up and this component with create a new one anyway (as it works now).

So currently it's not a matter of having an id.

Do we want to change this?

packages/clerk-js/src/ui/common/withRedirectToHome.tsx
yourtallness: Which comment are you referring to?

SokratisVidros: https://github.com/clerkinc/javascript/pull/186#discussion_r849576512

packages/clerk-js/src/ui/signUp/SignUpStart.tsx
SokratisVidros: Hmmm, you are right. We always create a new sign up. When will the user land on this component with an existing sign up? Won't the `<SignUp/>` redirect them to `<SignUpContinue/>`?

SokratisVidros: Correct. I believe it's more clear this way.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -451,6 +451,8 @@ export default class Clerk implements ClerkInterface {
       params.afterSignUpUrl || params.redirectUrl || displayConfig.afterSignUpUrl,
     );
 
+    const navigateToContinueSignUp = makeNavigate(displayConfig.signUpUrl + '#/continue');
```
nikosdouvlis, line 454: One more hard link joins the 'routes-we-need-to-refactor' gang 😄 

PS: Not related to this PR

packages/clerk-js/src/core/clerk.ts
```diff
@@ -500,6 +504,10 @@ export default class Clerk implements ClerkInterface {
       }
     }
 
+    if (su.status == 'missing_requirements') {
```
nikosdouvlis, line 507: Is this scenario possible? Let's take a potential user "A" for example:

- If A does not exist and we start from si, this is going to be handled by line 478
- If A does not exist and we start from su, this is going to be handled by the component (no transfer will take place)
- If  A exists and we start from si, we cannot end up in a  su `missing_requirements` state*
- If A` exists and we start from su, we cannot end up in a  su `missing_requirements` state*

(*) This could be possible if the customer updates the instance settings and adds more su requirements after A signed up. In this case, if A tries to sign in, would we prompt them to enter the additional required fields would we simply complete the sign in and let the customer handle the extra data?

yourtallness: This scenario occurs when the oauth flow completes (for a sign-up) & there are missing requirements (e.g. the fname & lname are mandatory but the provider doesn't return them, or the provider does not return a verified email).

packages/clerk-js/src/ui/common/authForms/Header.tsx
```diff
@@ -37,6 +40,7 @@ export function Header({
         />
       )}
       {error && <Error style={errorStyle}>{error}</Error>}
+      {info && <Info>{info}</Info>}
```
SokratisVidros, line 43: 🔧 Let's introduce an Alert component with two variants/flavors for error and info. It's a more scalable approach as we will add more use cases to our components.

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
```diff
@@ -0,0 +1,350 @@
...
+  const signUp = useCoreSignUp();
+
+  // Redirect to sign-up if there is no persisted sign-up
+  if (!signUp.id) {
+    void navigate(displayConfig.signUpUrl);
+    return null;
+  }
+
+  // Pre-populate fields from existing sign-up object
+  const formFields = {
```
SokratisVidros, line 44: 🔧 A dedicated hook to prepare the data for the `<SignUpContinue/>` will be helpful as it will enable us to do more targeted unit testing. 

Moreover, it would also help us to move all the field creation logic to the SignUp resource so that the React components can be more dummy.

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
```diff
@@ -0,0 +1,350 @@
...
+  //       //
+  //       // We will revise this strategy as part of the Clerk DX epic.
+  //       void (await signUp.create({}));
+  //     }
+  //   }
+  //
+  //   void handleOauthError();
+  // });
+
+  const handleChangeActive = (type: ActiveIdentifier) => (e: React.MouseEvent) => {
```
SokratisVidros, line 130: ⛔  There is a lot of code duplication between the `<SignUpStart/>` and the `<SignUpContinue/>` components that  increases the maintenance overhead. Let's leverage composition and make them both depend on a new, common `<SignUpForm/>` component.

SokratisVidros (changes requested):

packages/clerk-js/src/core/clerk.ts
```diff
@@ -451,6 +451,9 @@ export default class Clerk implements ClerkInterface {
       params.afterSignUpUrl || params.redirectUrl || displayConfig.afterSignUpUrl,
     );
 
+    const navigateToContinueSignUp = makeNavigate(displayConfig.signUpUrl + '/continue');
```
colinclerk, line 454: This doesn't look like it will work if the sign up component is mounted with hash routing.

packages/clerk-js/src/ui/common/authForms/Header.tsx
yourtallness: Introduced `Alert` component.

Do we need to keep the `Error` component around for backwards compatibility?

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
yourtallness: I have moved the field determination logic to utils for the time being.

Don't know if SignUp would be the proper place, but maybe a form helper would be convenient.

packages/clerk-js/src/core/clerk.ts
yourtallness: @SokratisVidros I've left testing CRA testing for last, but will of course not merge unless it works there as well.

colinclerk: Fwiw... you can test hash based routing in any framework. Just drop the component without `routing` and `path` props.

This shouldn't work because it will navigate to `/sign-up/continue`, which routers treat as separate route than `/sign-up`.

If we're redirecting to a nested path and we don't know whether it's `hash` or `path` mounted, we should always use `hash`. Then, If it turns out that <SignUp/> is path-mounted, <SignUp/> itself can strip the hash away.

`navigateToFactorTwo` looks like it always uses hash routing, which is what I'd expect.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -523,16 +532,26 @@ export default class Clerk implements ClerkInterface {
...
     } catch (err) {
       if (isError(err, ERROR_CODES.FORM_IDENTIFIER_NOT_FOUND)) {
         signInOrSignUp = await this.client.signUp.authenticateWithMetamask();
+
+        if (
+          signInOrSignUp.verifications.web3Wallet.status === 'verified' &&
+          signInOrSignUp.status === 'missing_requirements'
+        ) {
+          // @ts-ignore
+          return this.navigate(signUpUrl + '/continue');
```
SokratisVidros, line 553: 🔧 Let's use `navigateToContinueSignUp`

packages/clerk-js/src/ui/common/alert/Alert.tsx
```diff
@@ -0,0 +1,24 @@
...
+export type AlertType = 'info' | 'error';
+
+export type AlertProps = {
+  type: AlertType;
+  children: React.ReactNode;
+  style?: {};
+} & React.HTMLAttributes<HTMLDivElement>;
+
+// Renders global alerts across components, will be replaced by notification snackbars.
+export const Alert: React.FC<AlertProps> = ({ type, children, style }) => {
```
SokratisVidros, line 12: Because of https://spin.atomicobject.com/2022/01/04/think-twice-react-fc/ let's do the following:

```suggestion
export const function Alert({ type, children, style }: AlertProps { ... }
```

packages/clerk-js/src/ui/common/authForms/Header.tsx
```diff
@@ -36,12 +35,21 @@ export function Header({
           handleClick={handleBack}
         />
       )}
-      {error && <Error style={errorStyle}>{error}</Error>}
+
+      {alerts.length > 0 && (
```
SokratisVidros, line 39: ❓ Do we show multiple alerts at the same time or is it always one? If no, let's just stick to a single Alert for now to make the API easier.

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
```diff
@@ -0,0 +1,161 @@
...
+        )}
+
+        {showWeb3Providers && (
+          <SignUpWeb3
+            web3Options={web3Options}
+            setError={setError}
+          />
+        )}
+
+        {showFormFields(userSettings) && (
```
SokratisVidros, line 139: 💯 This is going towards the right direction.

packages/clerk-js/src/ui/signUp/utils.ts
```diff
@@ -45,5 +70,75 @@ export function determineFirstPartyFields(environment: EnvironmentResource, hasT
...
 
+  // Enable fields based on current form state
+
+  const shouldShowEmailAddressField =
+    (isEmailAddressUsedForIdentification && hasTicket && hasEmail) ||
+    fields.emailAddress ||
+    (fields.emailOrPhone && activeCommIdentifierType === 'emailAddress');
+
+  if (shouldShowEmailAddressField) {
+    fields.emailAddress ||= 'required'; // Is this accurate?
```
SokratisVidros, line 81: 🔧 These are remnants of the old auth config modelling. Currently we have the user_settings object and it's better to build a modelling on top of it without using hardcoded enumerations.

packages/clerk-js/src/ui/signUp/utils.ts
```diff
@@ -45,5 +70,75 @@ export function determineFirstPartyFields(environment: EnvironmentResource, hasT
...
+
+  // If continuing with an existing sign-up, show only fields absolutely necessary to minimize fiction
+
+  if (signUp) {
+    const hasVerifiedEmail = signUp.verifications?.emailAddress?.status == 'verified';
+    const hasVerifiedPhone = signUp.verifications?.phoneNumber?.status == 'verified';
+    const hasVerifiedExternalAccount = signUp.verifications?.externalAccount?.status == 'verified';
+
+    if (hasVerifiedEmail) {
+      delete fields.emailAddress;
```
SokratisVidros, line 99: 🔧 Creating the superset of fields and the removing the non-missing and optional ones seems a bit counter intuitive. Isn't it simpler to do the opposite and build the final fields object as the business logic flows?

SokratisVidros (changes requested):

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
```diff
@@ -0,0 +1,161 @@
...
+    } catch (err) {
+      handleError(err, fieldsToSubmit, setError);
+    }
+  };
+
+  const completeSignUpFlow = (su: SignUpResource) => {
+    if (su.status === 'complete') {
+      return setSession(su.createdSessionId, navigateAfterSignUp);
+    } else if (su.emailAddress && su.verifications.emailAddress.status !== 'verified') {
+      return navigate(displayConfig.signUpUrl + '/verify-email-address');
```
colinclerk, line 103: This looks broken for hash routing

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
```diff
@@ -0,0 +1,161 @@
...
+    } else if (su.phoneNumber && su.verifications.phoneNumber.status !== 'verified') {
+      return navigate(displayConfig.signUpUrl + '/verify-phone-number');
```
colinclerk, line 105: This looks broken for hash routing

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
SokratisVidros: @yourtallness -> https://github.com/clerkinc/javascript/blob/main/packages/clerk-js/src/ui/signUp/SignUpStart.tsx#L171

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
SokratisVidros: @yourtallness -> https://github.com/clerkinc/javascript/blob/main/packages/clerk-js/src/ui/signUp/SignUpStart.tsx#L171

packages/clerk-js/src/core/clerk.ts
nikosdouvlis: I think this has the same hash routing issue Colin mentioned in a previous comment.

nikosdouvlis: ❓ Why do we need ts-ignore here?

packages/clerk-js/src/ui/common/alert/Alert.tsx
nikosdouvlis: (Somewhat unrelated to this PR, but still an interesting read)
React18 removes the implicit `children` prop from `React.FC` so in theory, we can use it in more places now. However, there are still valid reasons to avoid it: 

https://github.com/facebook/create-react-app/pull/8177#issue-537764080

packages/clerk-js/src/core/clerk.ts
igneel64: @nikosdouvlis 
The return type of navigate is not `void` . We can go with:
`void this.navigate(signUpUrl + '/continue');` or just not return

packages/clerk-js/src/ui/common/authForms/Header.tsx
nikosdouvlis: 🔧  Instead of passing an array of alerts, we can simply pass a component that renders multiple alerts if needed. I'd change HeaderProps to accept a `alert?: React.ReactNode;` prop instead. That'd simplify the common use cases like:  https://github.com/clerkinc/javascript/pull/186/files#diff-f552e096d1e5ff1ceb1055fa13af0712224e06969c8e494cac5aad867ecefde3R102

In case multiple alerts are necessary (
https://github.com/clerkinc/javascript/pull/186/files#diff-a365df1e8e318373fc6e01f7e89b201851d990bacd9724017b5c98bdd892b6f5R117), I'd wrap them in a fragment, eg: 
```
        alerts={<>
            {error && <Alert type='error'>{error}</Alert>}
            <Alert type='info'>Please fill in the following information to complete your sign-up</Alert>
        </>}
```

packages/clerk-js/src/ui/styles/_alert.scss
```diff
@@ -1,13 +1,19 @@
-.error {
+.alert {
```
nikosdouvlis, line 1: 🔧  In theory, this is a breaking change since the `cl-error` class was publicly targettable.

We need to make sure that the `cl-error` class still exists as a targetable class on Alert component

Am I missing something? cc @SokratisVidros  @igneel64

SokratisVidros: Good catch @nikosdouvlis

igneel64: Anything in these styles can be considered public in a way so yes!

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
igneel64, line 105: 🙃 
+ gives me these errors:
image

packages/clerk-js/src/ui/signUp/utils.ts
```diff
@@ -45,5 +70,75 @@ export function determineFirstPartyFields(environment: EnvironmentResource, hasT
...
+        delete fields[k];
+      }
+    });
+  }
+
   return fields;
 }
+
+export function showFormFields(userSettings: UserSettingsResource): boolean {
+  const { socialProviderStrategies, web3FirstFactors } = userSettings;
```
igneel64, line 137: ❓ 
Do we want `web3FirstFactors` as unused ?

yourtallness: Fixing

packages/clerk-js/src/ui/styles/_alert.scss
yourtallness: @nikosdouvlis Only the `cl-error` class or the `Error` component as well?

packages/clerk-js/src/ui/signUp/utils.ts
yourtallness: I didn't want to re-write the field determination logic for this feature, so I took a conservative approach.

What I essentially did was:

- Determine fields to show, as before
- If PSU-mode, strip anything that's redundant for completing the sign-up

I would recommend a re-write in a separate PR.

packages/clerk-js/src/ui/common/authForms/Header.tsx
yourtallness: @nikosdouvlis Will adapt as per your recommendation.

packages/clerk-js/src/ui/common/alert/Alert.tsx
yourtallness: Converting to function

packages/clerk-js/src/core/clerk.ts
yourtallness: Btw, I originally routed to `#/continue` but this does not work when we're already in the `SignUp` component.

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
yourtallness: And `signUpUrl` is supposed to be a string, go figure.

packages/clerk-js/src/ui/common/authForms/Header.tsx
yourtallness: Fixed

packages/clerk-js/src/ui/signUp/utils.ts
yourtallness: The field determination function was given a complete overhaul, please take a look.

packages/clerk-js/src/ui/signUp/utils.ts
yourtallness: The function was rewritten to be more readable, each field is generated in separate methods.

SokratisVidros (approved):

PR #191, opened 2022-04-13
feat(clerk-js): Organization slugs

gkats:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Added support for organization slugs. Added slug as a parameter in `clerk.createOrganization()`.

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -548,8 +548,8 @@ export default class Clerk implements ClerkInterface {
     }
   };
 
-  public createOrganization = async ({ name }: CreateOrganizationParams): Promise<OrganizationResource> => {
-    return await Organization.create(name);
+  public createOrganization = async ({ name, slug }: CreateOrganizationParams): Promise<OrganizationResource> => {
+    return Organization.create({ name, slug });
   };
 
   public getOrganizationMemberships = async (): Promise<OrganizationMembership[]> => {
--- a/packages/clerk-js/src/core/resources/Organization.ts
+++ b/packages/clerk-js/src/core/resources/Organization.ts
@@ -1,4 +1,5 @@
 import type {
+  CreateOrganizationParams,
   GetMembershipsParams,
   MembershipRole,
   OrganizationInvitationJSON,
@@ -16,6 +17,7 @@ export class Organization extends BaseResource implements OrganizationResource {
 
   id!: string;
   name!: string;
+  slug!: string;
   publicMetadata: Record<string, unknown> = {};
   createdAt!: Date;
   updatedAt!: Date;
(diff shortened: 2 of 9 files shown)


packages/clerk-js/src/core/resources/Organization.ts
```diff
@@ -24,12 +26,12 @@ export class Organization extends BaseResource implements OrganizationResource {
     this.fromJSON(data);
   }
 
-  static async create(name: string): Promise<OrganizationResource> {
+  static async create({ name, slug }: CreateOrganizationParams): Promise<OrganizationResource> {
```
nikosdouvlis, line 29: ❓ I'm not entirely familiar with the organisations API... is this a public method intended to be used by the users? If yes, then this would be a breaking change and should probably be handled using overloads. If no, please ignore :)

gkats: I saw that this is only used by `clerk.createOrganization` but was reluctant to make the change as well.

@igneel64 What's your take regarding backwards compatibility here?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -548,8 +548,8 @@ export default class Clerk implements ClerkInterface {
     }
   };
 
-  public createOrganization = async ({ name }: CreateOrganizationParams): Promise<OrganizationResource> => {
-    return await Organization.create(name);
+  public createOrganization = async ({ name, slug }: CreateOrganizationParams): Promise<OrganizationResource> => {
+    return await Organization.create({ name, slug });
```
SokratisVidros, line 552: 🙃 
```suggestion
    return Organization.create({ name, slug });
```

packages/clerk-js/src/core/resources/Organization.ts
SokratisVidros: We can overload it for now and teach it to accept a string (name) or an object. 

For future reference, **we must always use objects for resource methods even if we have just one parameter**

SokratisVidros (changes requested):

packages/clerk-js/src/core/resources/Organization.ts
igneel64: The change would be considered breaking but if we go with the suggestion of:

> We can overload it for now and teach it to accept a string (name) or an object.

I think we are good.

nikosdouvlis: If we go with overloads, please add a deprecation warning for the non-object form so we can remove it in v4

igneel64 (approved):

npetridis (approved):

packages/clerk-js/src/core/resources/Organization.ts
gkats: Could you please take a look at this commit? https://github.com/clerkinc/javascript/pull/191/commits/0fea8df722f99673c20ba8b48952c4574dc9c501

Is this what you guys had in mind regarding deprecation?

nikosdouvlis (approved):

gkats:
@nikosdouvlis Awesome feedback, hope this is better. https://github.com/clerkinc/javascript/pull/191/commits/8b89c5154f91615b2eea6f4219a797b85df117c1

nikosdouvlis (approved):

PR #197, opened 2022-04-14
Cleanup interstitial logic

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend-core`
- [x] `@clerk/clerk-sdk-node`
- [x] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

- Split the `getAuthState` logic into 2 distinct scenarios, `handleRequestWithTokenInHeader` and `handleRequestWithTokenInCookie` that can be handled separately
- For the token in cookie scenario, the complex interstitial logic has been extracted into `InterstitialRules`. These rules run sequentially and in the same order as before
- To further simplify `getAuthState`, it no longer fetches the interstitial. This responsibility is moved back to the caller

### DIFF
--- a/packages/backend-core/src/Base.ts
+++ b/packages/backend-core/src/Base.ts
@@ -8,14 +8,23 @@ import {
   BuildAuthenticatedStateOptions,
   JWT,
   JWTPayload,
+  TokenType,
   TokenVerificationErrorReason,
   VerifySessionTokenOptions,
 } from './types';
 import { parse } from './util/base64url';
-import { isDevelopmentOrStaging, isProduction } from './util/clerkApiKey';
 import { mapErrorReasonResponse, TokenVerificationError } from './util/errors';
+import {
+  crossOriginRequestWithoutCors,
+  hasClientUatButCookieIsMissingInProd,
+  isNormalSignedOutState,
+  nonBrowserRequestInDevRule,
+  potentialFirstLoadInDevWhenUATMissing,
+  potentialFirstRequestOnProductionEnvironment,
+  potentialRequestAfterSignInOrOurFromClerkHostedUiInDev,
+  runInterstitialRules,
+} from './util/interstitialRules';
 import { checkClaims, isExpired } from './util/jwt';
-import { checkCrossOrigin } from './util/request';
 
 export const API_KEY = process.env.CLERK_API_KEY || '';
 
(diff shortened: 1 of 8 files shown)


packages/backend-core/src/Base.ts
```diff
@@ -88,14 +90,13 @@ export class Base {
...
-   * @param {string} token Clerk JWT verification token
+   * @param jwt Clerk JWT verification token
    * Modify the RSA public key from the Clerk PEM supplied and return a contructed CryptoKey.
    * You will find that at your application dashboard (https://dashboard.clerk.dev) under Settings ->  API keys
    *
    */
-  loadCryptoKey = async (key?: string): Promise<CryptoKey> => {
-    if (!key) {
+  loadCryptoKey = async (jwt?: string): Promise<CryptoKey> => {
+    if (!jwt) {
```
igneel64, line 99: This is actually a key here, how did the change came about ?

igneel64:
I personally don't prefer the nested logic versus the line by line as it was before. It makes the flowchart easier to follow in my view.
I am not sure how this came about but I would like to know how other people feel about it.

packages/backend-core/src/Base.ts
nikosdouvlis: Probably from the comment - but it should be `key`!

packages/backend-core/src/util/interstitialRules.ts
```diff
@@ -0,0 +1,85 @@
...
+// prevent interstitial rendering
+export const nonBrowserRequestInDevRule: InterstitialRule = (params, key) => {
+  if (isDevelopmentOrStaging(key) && !params.userAgent?.startsWith('Mozilla/')) {
+    return { status: AuthStatus.SignedOut, errorReason: AuthErrorReason.HeaderMissingNonBrowser };
+  }
+  return undefined;
+};
+
+// Is this correct? In cross-origin requests the use of Authorization header is mandatory
+export const crossOriginRequestWithoutCors: InterstitialRule = params => {
```
colinclerk, line 31:
```suggestion
export const crossOriginRequestWithoutHeader: InterstitialRule = params => {
```

colinclerk (approved):
I didn't review the refactor very significantly, just the names of the rules.

In general I like having the verbose rule names listed here, though I commented a lot on them to add extra clarity.  Let me know if there's any confusion.

This is a little bit of a scary change. We'd benefit from having an external test suite that we could use to run tests against this SDK and others written for other languages, so we know they're always behaving consistently.

SokratisVidros:
Team what's the status of this?

nikosdouvlis:
Closing this one for now, will tackle it as it's own task during the true SSR project.
Some of the safest changes have been extracted to a separate PR we can merge now: https://github.com/clerkinc/javascript/pull/233

Ticket: https://www.notion.so/clerkdev/Refactor-interstitial-logic-856b68a9f76845a3b0e5ddd1c4f3a727

PR #213, opened 2022-04-27
fix(backend-core): Fix case of buildAuthState erroneous return value

igneel64:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

- When `buildAuthState` returned an erroneous AuthState status _(`!== SignedIn`)._ the response was not returned as it should be to the user.

### DIFF
--- a/packages/backend-core/src/Base.ts
+++ b/packages/backend-core/src/Base.ts
@@ -295,6 +295,8 @@ export class Base {
         interstitial: await fetchInterstitial(),
         errorReason: AuthErrorReason.CookieOutDated,
       };
+    } else if (!authenticatedState.sessionClaims) {
+      return authenticatedState;
     }
 
     return {


colinclerk (commented):
I believe this Unknown case can be removed, since the three if statements cover every case?

```
    return {
      status: AuthStatus.Interstitial,
      interstitial: await fetchInterstitial(),
      errorReason: AuthErrorReason.Unknown,
    };
```

I'm a little surprised typescript: isn't catch the code as impossible to reach though.  Am I missing something?

igneel64:
@colinclerk If there was an else statement, this would be marked as unreachable. Now it does not.
I would propose that we don't remove it but either `throw` or `signed-out` WDYT ? _I would recommend being more defensive in the case we add something on `buildAuthState`_

nikosdouvlis (approved):

PR #224, opened 2022-05-05
set intersticial script tag as crossOrigin='anonymous'

nicksrandall:
This should fix issues when apps are using https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy

| NOTE: This change may need to happen in other places but this is one that I need. 

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/packages/remix/src/client/Interstitial.tsx
+++ b/packages/remix/src/client/Interstitial.tsx
@@ -60,6 +60,7 @@ const createInterstitialHTMLString = (frontendApi: string, libVersion: string, d
                 script.setAttribute('data-clerk-frontend-api', '${frontendApi}');
                 script.async = true;
                 script.src = '${getScriptUrl(frontendApi, libVersion)}';
+                script.crossOrigin = '';
                 script.addEventListener('load', startClerk);
                 document.body.appendChild(script);
             })();


packages/remix/src/client/Interstitial.tsx
```diff
@@ -60,6 +60,7 @@ const createInterstitialHTMLString = (frontendApi: string, libVersion: string, d
                 script.setAttribute('data-clerk-frontend-api', '${frontendApi}');
                 script.async = true;
                 script.src = '${getScriptUrl(frontendApi, libVersion)}';
+                script.crossOrigin = '';
```
SokratisVidros, line 63: 🙃 Since setting the attribute name to an empty value, like crossorigin or crossorigin="", is the same as anonymous let's write `anonymous` to ease the future reader.

SokratisVidros (approved):
👏

nikosdouvlis (approved):

PR #226, opened 2022-05-08
fix(clerk-js): Hide signup input form when an oauth exists but we have no auth factor

npetridis:
## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Hide signup input form when an oauth exists but we have no auth factor
https://www.notion.so/clerkdev/Hide-name-in-SignUp-when-OAuth-is-the-only-factor-dd5341fba63c4237a7f66672d234e97a

### DIFF
--- a/packages/clerk-js/src/ui/signUp/SignUpStart.tsx
+++ b/packages/clerk-js/src/ui/signUp/SignUpStart.tsx
@@ -302,12 +302,7 @@ function _SignUpStart(): JSX.Element {
       </Control>
     ) : null;
 
-  const atLeastOneAuthFactor =
-    userSettings.attributes.email_address.used_for_first_factor ||
-    userSettings.attributes.phone_number.used_for_first_factor ||
-    (userSettings.attributes.password.required && userSettings.attributes.username.required);
-
-  const showFormFields = atLeastOneAuthFactor || !oauthOptions.length;
+  const showFormFields = userSettings.hasValidAuthFactor || (!oauthOptions.length && !web3Options.length);
 
   return (
     <>
--- a/packages/clerk-js/src/ui/signUp/SignUpStart.tsx
+++ b/packages/clerk-js/src/ui/signUp/SignUpStart.tsx
@@ -302,7 +302,12 @@ function _SignUpStart(): JSX.Element {
       </Control>
     ) : null;
 
-  const atLeastOneFormField = nameField || usernameField || emailAddressField || phoneNumberField || passwordField;
+  const atLeastOneAuthFactor =
+    userSettings.attributes.email_address.used_for_first_factor ||
+    userSettings.attributes.phone_number.used_for_first_factor ||
+    (userSettings.attributes.password.required && userSettings.attributes.username.required);
+
+  const showFormFields = atLeastOneAuthFactor || !oauthOptions.length;
 
   return (
     <>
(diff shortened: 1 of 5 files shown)


packages/clerk-js/src/ui/signUp/SignUpStart.tsx
```diff
@@ -302,7 +302,12 @@ function _SignUpStart(): JSX.Element {
       </Control>
     ) : null;
 
-  const atLeastOneFormField = nameField || usernameField || emailAddressField || phoneNumberField || passwordField;
+  const atLeastOneAuthFactor =
+    userSettings.attributes.email_address.used_for_first_factor ||
+    userSettings.attributes.phone_number.used_for_first_factor ||
+    (userSettings.attributes.password.required && userSettings.attributes.username.required);
+
+  const showFormFields = atLeastOneAuthFactor || !oauthOptions.length;
```
SokratisVidros, line 310: 🔧 @npetridis we should add another check for Web3 options. They behave in the exact same way as oauth options.

packages/clerk-js/src/ui/signUp/SignUpStart.tsx
```diff
@@ -302,7 +302,12 @@ function _SignUpStart(): JSX.Element {
       </Control>
     ) : null;
 
-  const atLeastOneFormField = nameField || usernameField || emailAddressField || phoneNumberField || passwordField;
+  const atLeastOneAuthFactor =
```
nikosdouvlis, line 305: ❓ Would it make sense to extract this as a `hasValidAuthFactor` method on the `UserSettings` class?

Eg: 
```tsx
  get hasValidAuthFactor() {
    return this.attributes.email_address.used_for_first_factor ||
    this.attributes.phone_number.used_for_first_factor ||
    (this.attributes.password.required && this.attributes.username.required);
  }

```

packages/clerk-js/src/ui/signUp/SignUpStart.tsx
```diff
@@ -302,7 +302,7 @@ function _SignUpStart(): JSX.Element {
       </Control>
     ) : null;
 
-  const atLeastOneFormField = nameField || usernameField || emailAddressField || phoneNumberField || passwordField;
+  const showFormFields = userSettings.hasValidAuthFactor || (!oauthOptions.length && !oauthOptions.length);
```
nikosdouvlis, line 305: 🔧 
I think this should be checking for web3 options, the `(!oauthOptions.length && !oauthOptions.length)` check is probably wrong

nikosdouvlis (changes requested):

packages/clerk-js/src/ui/signUp/SignUpStart.tsx
npetridis: wrong copy/paste when backtesting the fix :P

nikosdouvlis (approved):

SokratisVidros (approved):

PR #239, opened 2022-05-13
Safe url hash build

SokratisVidros:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Navigating to two factor after a sign up transfer flow was broken when the sign in URL contained querystring parameters such as redirect_url or after_sign_up_url.

The culprit was the string concatenation was used to build the final factor-two url as we were appending the #/factor-two string after the query strings.

This PR fixes the issue by leveraging the new, smarter buildURL utility.


packages/clerk-js/src/utils/path.ts
```diff
@@ -0,0 +1,11 @@
+const SEPARATOR = '/';
+const MUTLIPLE_SEPARATOR_REGEX = new RegExp(SEPARATOR + '{1,}', 'g');
```
agis, line 2: 🤓 typo `MULTIPLE`

SokratisVidros: Oops, fixed!

nikosdouvlis (approved):
:partyparrot:

PR #244, opened 2022-05-17
feat(types): Include new organization role `guest_member`

alex-ntousias:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

This PR includes the new organization role `guest_member`.

### DIFF
--- a/packages/backend-core/src/api/resources/Enums.ts
+++ b/packages/backend-core/src/api/resources/Enums.ts
@@ -24,7 +24,7 @@ export type OAuthStrategy = `oauth_${OAuthProvider}`;
 
 export type OrganizationInvitationStatus = 'pending' | 'accepted' | 'revoked';
 
-export type OrganizationMembershipRole = 'basic_member' | 'admin';
+export type OrganizationMembershipRole = 'basic_member' | 'guest_member' | 'admin';
 
 export type SignInIdentifier = 'username' | 'email_address' | 'phone_number' | 'web3_wallet' | OAuthStrategy;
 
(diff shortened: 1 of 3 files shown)


gkats (approved):
Do we want to make this available to the NodeJS SDK as well?

If yes, we should add it here: https://github.com/clerkinc/javascript/blob/a4439a08c16e3e424a9a0d0f7aa62ac59fea55b4/packages/backend-core/src/api/resources/Enums.ts#L27

gkats:
I've just noticed that we've exposed two different constants for the membership role. It's also here: https://github.com/clerkinc/javascript/blob/a4439a08c16e3e424a9a0d0f7aa62ac59fea55b4/packages/backend-core/src/types/jwt.ts#L49

alex-ntousias:
> I believe we should refactor the one in jwt.ts to use the enum value.

@gkats It feels weird to have a package called `types` which would depend on a package called `resources`. 
I would expect it to be the other way around, i.e. have all the types declared in `Enums` to be in the `types` package instead.

Anyway, I'm modifying these two types to include the extra role but I will avoid removing the JWT type at this point. Let's take this offline and have another PR to tackle it.

PR #245, opened 2022-05-17
feat(backend-core): Get organization by ID or slug

gkats:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend-core`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Added a method on the organizations API to retrieve an organization by either ID or slug. 

The signature is `organizations.getOrganization({ organizationIdOrSlug: "acme-inc" })`.

### DIFF
--- a/packages/backend-core/API.md
+++ b/packages/backend-core/API.md
@@ -21,6 +21,7 @@ Reference of the methods supported in the Clerk Backend API wrapper. [API refere
 - [Organization operations](#organization-operations)
   - [getOrganizationList()](#getorganizationlist)
   - [createOrganization(params)](#createorganizationparams)
+  - [getOrganization(params)](#getorganizationparams)
   - [updateOrganization(organizationId, params)](#updateorganizationorganizationid-params)
   - [updateOrganizationMetadata(organizationId, params)](#updateorganizationmetadataorganizationid-params)
   - [deleteOrganization(organizationId)](#deleteorganizationorganizationid)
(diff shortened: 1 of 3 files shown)


packages/backend-core/src/api/collection/OrganizationApi.ts
```diff
@@ -25,6 +25,10 @@ type OrganizationMetadataRequestBody = {
   privateMetadata?: string;
 };
 
+type GetOrganizationParams = {
+  organizationIdOrSlug: string;
```
nikosdouvlis, line 29: ❓ I'm wondering if instead of `organizationIdOrSlug` it'd make more sense to split the type into:

```ts
type GetOrganizationParams = { id: string } | { slug: string };
```

So we'd have: 
```ts
getOrganization({ id: '' })
// and 
getOrganization({ slug: '' })
```

Which reads slightly better and is more explicit. However, If we expect that the developer does not know the kind of string they'll be passing when calling `getOrganization` in advance, then `organizationIdOrSlug` makes total sense.

gkats: @nikosdouvlis Thanks for the suggestion! Updated the code accordingly.

igneel64 (approved):

packages/backend-core/API.md
```diff
@@ -195,6 +196,26 @@ const organization = await clerkAPI.organizations.createOrganization({
...
+Fetch an organization whose ID or slug matches the one provided in the parameters.
+
+The instance that the organization belongs to is determined by the API key you've provided when configuring the API. You can either set the `CLERK_API_KEY` environment variable, or provide the `apiKey` property explicitly when configuring the API client.
+
+Available parameters:
+
+- _organizationIdOrSlug_ The ID or slug of the organization.
+
+```ts
+const organizationBySlug = await clerkAPI.organizations.getOrganization({
```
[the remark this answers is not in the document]
gkats: I'm the worst. Updated and thanks for catching that!

PR #248, opened 2022-05-17
fix(clerk-js): Fix password field display logic, fix margin between o…

yourtallness:
…auth & web3 providers

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Implements:

- Show password field if `enabled` && `required`
- Since `enabled` is decoupled from whether we actually show a field, expose a `show` flag in the form builder utils instead
- Fix margin when both oauth & web3 providers are shown


gkats (approved):
Looks good to me. 

I like the fact that we renamed the `showDisabled` property to `disabled`, but I'm not so sure about the `enabled`/`required` renaming to `show`.

SokratisVidros (approved):
Overall it looks OK however I'd avoid the extra `show` flag and I'd try to generate an array of all the fields that needs to be present that should include a required flag. If a field is not in an array it won't be rendered. That way, this is more dynamic and can scale better with new fields in the future.

packages/clerk-js/src/ui/signUp/SignUpStart.tsx
```diff
@@ -102,7 +102,8 @@ function _SignUpStart(): JSX.Element {
   };
 
   React.useLayoutEffect(() => {
-    if (Object.values(fields).filter(f => f.enabled && !f.required).length) {
+    // Don't proceed with token flow if there are still optional fields to fill in
+    if (Object.values(fields).filter(f => f.show && !f.required).length) {
```
igneel64, line 106: 🙃 
I think we can change this to `Array.prototype.some`

igneel64 (approved):

packages/clerk-js/src/ui/signUp/SignUpStart.tsx
yourtallness: Will do ✅

yourtallness:
@gkats @SokratisVidros Will eliminate `show` & replace it will presence check within the form builder object.

PR #252, opened 2022-05-19
Introduce v4 module

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

- IntroduceD the v4 module, allowing us to develop the new components without the overhead of managing a separate release branch. You shouldn't see any differences when running the `npm run build` and `npm run test` commands. To make sure that the old build process remains the same, I reenabled the `bundle watch` command.
- Introduced a new `npm run dev:v4` command that uses webpack + esbuild + HMR to build and serve clerkjs at `https://js.lclclerk.com/npm/clerk.browser.js`
- Refactored most of the problematic absolute imports to relative. I made the fewest changes that were required. More work needs to be done on that end, let's aim to remove the path aliases all together once the components are released

Note: The basic build system with `emotion` will be introduced with the next PR

### Fun stuff
With the switch to esbuild for the dev server, we gained some noticeable perf. benefits, namely: 
Time for first build: ~7.5 -> ~1sec
HMR update: ~400ms -> ~50ms

This only affects the dev mode for now, but If we notice any issues I'll switch it back to babel/ts-loader.


packages/clerk-js/package.json
```diff
@@ -76,45 +77,47 @@
...
+    "webpack-dev-server": "^4.9.0",
     "webpack-merge": "^5.8.0"
   },
   "browserslist": "last 2 versions, ios_saf > 12, Safari > 12, > 1%, not dead, not ie > 0",
   "bundlewatch": {
     "files": [
       {
         "path": "./dist/clerk.js",
-        "maxSize": "148kB"
+        "maxSize": "162kB"
```
SokratisVidros, line 116: ❓ Does this mean that the new bundle is bigger till we remove hot-loading?

SokratisVidros (approved):
🚀

packages/clerk-js/package.json
nikosdouvlis: The prod bundle is not affected form these changes. Bundlewatch was disabled these values hadn't been updated for a couple of months now.

The bundle size on main is at 159kB, same as this branch

PR #253, opened 2022-05-20
Introduce themable styled system

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description


packages/clerk-js/src/v4/styledSystem/theme.ts
```diff
@@ -0,0 +1,25 @@
+import { colors } from './colors';
+import { createInternalTheme } from './createInternalTheme';
+import { shadows } from './shadows';
+import { radii, sizes, space } from './sizes';
+import { fonts, fontSizes, fontWeights, letterSpacings, lineHeights } from './typography';
+
+export const baseTheme = createInternalTheme({
```
nikosdouvlis, line 7: The `createInternalTheme` function accepts a `theme` of type `T`. This generic type is not restricted on purpose: this makes the `baseTheme` strictly typed after the `createInternalTheme` call while still allowing us to add theme tokens without changing types while developing.

It'd be a good idea to narrow down the `T` type once we finalise the tokens for the internal base theme for extra safety

packages/clerk-js/src/v4/styledSystem/sizes.ts
```diff
@@ -0,0 +1,54 @@
+const space = {
+  px: '1px',
+  '0x5': '0.125rem',
```
nikosdouvlis, line 3: I dynamically prefix all keys with `$` so we can freely choose key names without weird accessing patterns, even if they start with a number:
```ts
paddingBottom: theme.sizes['2xl']
```

For keys containing decimal values, I suggest we follow the `0x5` (0,5) convention because intellisense just works with this format: 
```ts
// this is ok
paddingBottom: theme.sizes.$0x5
```

packages/clerk-js/src/v4/styledSystem/css.ts
```diff
@@ -0,0 +1,35 @@
+import { applyCustomCssUtilities, WithCustomCssUtilities } from './customCssUtilities';
```
nikosdouvlis, line 1: Don't bother with this util too much. I'm including it for now, but for 99% of the use cases the `createStyles` would be a better fit. This was useful before `createStyles` was introduced but it'll be removed soon if we have no use for it

igneel64 (approved):

npetridis (approved):

packages/clerk-js/src/v4/styledSystem/css.ts
nikosdouvlis: `css` has been removed

PR #258, opened 2022-05-24
Introduce more v4 primitives

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

This PR introduces more primitives, improvements to the `createVariants` algorithm (allows deep merging of pseudo-selector blocks) and lastrly, introduces `createCssVariables` method. This method makes defining locally-scoped variables inside different variants easy.

In the following example, we have a `colorScheme` variant that sets the color of the  `accentColor` CSS variable. Notice that the `type` variant does not need to defined the colors again - you can simply use the CSS variable we defined earlier.

```ts
const { accentColor } = createCssVariables('accentColor');
const { applyVariants, filterProps } = createVariants(theme => ({
  variants: {
    colorScheme: {
      primary: {
        [accentColor]: theme.colors.$primary500,
      },
      danger: {
        [accentColor]: theme.colors.$danger500,
      },
    },
    type: {
      solid: {
        backgroundColor: accentColor,
      },
      outline: {
        border: '1px solid',
        borderColor: accentColor,
    },
  },
}));
```
In order to avoid conflicts with any user-defined CSS variables or nested styles, the above code will generate unique identifiers for each variable by adding a hash to the variable's name: 
image

### DIFF
--- a/packages/clerk-js/src/v4/primitives/Heading.tsx
+++ b/packages/clerk-js/src/v4/primitives/Heading.tsx
@@ -1,10 +1,10 @@
 import React from 'react';
 
 import { createVariants, PrimitiveProps, StyleVariants } from '../styledSystem';
-import { Box } from './Box';
 
 const { applyVariants } = createVariants(theme => ({
   base: {
+    boxSizing: 'border-box',
     // TODO: Should this be $text?
     color: theme.colors.$black,
   },
@@ -26,9 +26,10 @@ const { applyVariants } = createVariants(theme => ({
 export type HeadingProps = PrimitiveProps<'div'> & StyleVariants<typeof applyVariants> & { as?: 'h1' };
 
 export const Heading = React.forwardRef<HTMLDivElement, HeadingProps>((props, ref) => {
+  const { as: As = 'h1', ...rest } = props;
   return (
-    <Box
-      {...props}
+    <As
+      {...rest}
       css={applyVariants(props)}
       ref={ref}
     />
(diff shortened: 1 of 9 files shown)


packages/clerk-js/src/v4/primitives/Heading.tsx
```diff
@@ -0,0 +1,36 @@
...
+  defaultVariants: {
+    as: 'h1',
+  },
+}));
+
+export type HeadingProps = PrimitiveProps<'div'> & StyleVariants<typeof applyVariants> & { as?: 'h1' };
+
+export const Heading = React.forwardRef<HTMLDivElement, HeadingProps>((props, ref) => {
+  return (
+    <Box
```
igneel64, line 30: 🙃  
This should be `As` right?

nikosdouvlis: The `as` prop is getting handled by the underlying `Box`. But your comment made me realize that a default fallback value for the `as` prop is missing. Fixing :)

nikosdouvlis: Done

igneel64 (approved):

PR #267, opened 2022-05-27
New `setActive` method to set the active session and organization

alex-ntousias:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

This PR introduces a new method called `setActive`. 
This basically replaces the older `setSession` which is now deprecated.

In addition to setting the active session, it can also be used to set the active organization for the session.
Its signature is:
```
export type SetActiveParams = {
  session?: ActiveSessionResource | string | null;
  organization?: string | null;
  beforeEmit?: BeforeEmitCallback;
};

setActive: (SetActiveParams) => Promise<void>;
```

The session parameter can accept either an `ActiveSessionResource` or the id of such a session. It also accepts `null` in which case the current session will be removed.

The organization parameter accepts the id that we want to set as active in the session.
If no session parameter is passed, the currently active session will be used. Otherwise, the organization will be set as active in the passed session.
Please note that passing a `null` organization id means that the currently active organization will be removed as active.

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -27,7 +27,7 @@ import type {
   UserProfileProps,
   UserResource,
 } from '@clerk/types';
-import { SignOutOptions } from '@clerk/types/src';
+import { SetActiveParams, SignOutOptions } from '@clerk/types/src';
 
 import packageJSON from '../../package.json';
 import type Components from '../ui';
@@ -250,19 +250,27 @@ export default class Clerk implements ClerkInterface {
     });
   };
 
-  public setSession = async (
-    session: ActiveSessionResource | string | null,
-    beforeEmit?: BeforeEmitCallback,
-  ): Promise<void> => {
+  public setActive = async ({ session, beforeEmit }: SetActiveParams): Promise<void> => {
     if (!this.client) {
-      throw new Error('setSession is being called before the client is loaded. Wait for init.');
+      throw new Error('setActive is being called before the client is loaded. Wait for init.');
+    }
+
+    if (session === undefined) {
+      if (!this.session) {
+        throw new Error(
+          'setActive should either be called with a session param or there should be already an active session.',
+        );
+      }
+      session = this.session;
     }
 
     if (typeof session === 'string') {
       session = (this.client.sessions.find(x => x.id === session) as ActiveSessionResource) || null;
     }
 
-    this.#authService?.setAuthCookiesFromSession(session);
+    if (session) {
+      this.#authService?.setAuthCookiesFromSession(session);
+    }
 
     // If this.session exists, then signout was triggered by the current tab
     // and should emit. Other tabs should not emit the same event again
@@ -271,7 +279,7 @@ export default class Clerk implements ClerkInterface {
     }
 
     //1. setLastActiveSession to passed usersession (add a param)
-    if (typeof document != 'undefined' && document.hasFocus()) {
+    if (session && typeof document != 'undefined' && document.hasFocus()) {
       await this.#touchLastActiveSession(session);
     }
 
@@ -280,7 +288,7 @@ export default class Clerk implements ClerkInterface {
     //   When undefined, neither SignedIn nor SignedOut renders, which avoids flickers or
     //   automatic reloading when reloading shouldn't be happening.
     const beforeUnloadTracker = createBeforeUnloadTracker();
-    if (beforeEmit) {
+    if (session && beforeEmit) {
       beforeUnloadTracker.startTracking();
       this.session = undefined;
       this.organization = undefined;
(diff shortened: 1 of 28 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -271,7 +294,7 @@ export default class Clerk implements ClerkInterface {
     }
 
     //1. setLastActiveSession to passed usersession (add a param)
-    if (typeof document != 'undefined' && document.hasFocus()) {
+    if (session && typeof document != 'undefined' && document.hasFocus()) {
       await this.#touchLastActiveSession(session);
```
alex-ntousias, line 298: In this line, we're touching the given session and we're also updating the active organization id with the one that is currently in `session.lastActiveOrganizationId`.

I'm not a huge fan of this. The main reason is that the method's signature hides the above complexity.
If you guys also agree, we should probably change it to something more suitable. 
However, I would prefer if it's done in a separate PR.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -250,19 +256,36 @@ export default class Clerk implements ClerkInterface {
     });
   };
 
-  public setSession = async (
-    session: ActiveSessionResource | string | null,
-    beforeEmit?: BeforeEmitCallback,
-  ): Promise<void> => {
+  public setActive = async ({ session, organization, beforeEmit }: SetActiveParams): Promise<void> => {
```
alex-ntousias, line 259: I believe this method could probably be written much simpler if we re-arrange the code to have some early returns.
However, given how critical this is, I hesitated to do it.

igneel64: @alex-ntousias 
I agree and the thing is there so many sideEffect functions which makes it daunting.

packages/clerk-js/src/core/clerk.ts
igneel64: Should we add a comment that this happens here ?

igneel64 (approved):

packages/clerk-js/src/core/clerk.ts
alex-ntousias: Will do 👍

packages/clerk-js/src/core/clerk.ts
alex-ntousias: cc: @SokratisVidros @nikosdouvlis @raptisj ☝️

packages/clerk-js/src/core/clerk.ts
```diff
@@ -305,6 +330,14 @@ export default class Clerk implements ClerkInterface {
     this.#resetComponentsState();
   };
 
+  /** @deprecated Use `setActive` instead */
```
SokratisVidros, line 333: Folks, `setSession` is a very critical function for ClerkJS. I'd suggest that we don't mark it as deprecated yet to avoid confusion. 

Moreover, as Alex said before, `setActive` requires necessary refactorings. So, we'd rather do all these at once.

We will deprecate it when we go to ClerkJS V4 in June.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -305,6 +330,14 @@ export default class Clerk implements ClerkInterface {
...
+  public setSession = async (
```
SokratisVidros, line 334: 💯

packages/types/src/clerk.ts
```diff
@@ -312,6 +320,28 @@ export type RedirectOptions = {
...
+   * The session resource or session id (string version) to be set as active.
+   * If `null`, the current session is deleted.
+   */
+  session?: ActiveSessionResource | string | null;
+
+  /**
+   * The organization id to be set as active in the current session.
+   * If `null`, the currently active organization is removed as active.
+   */
+  organization?: string | null;
```
SokratisVidros, line 334: @alex-ntousias This should accept an `OrganizationResource | string | null` as in `session` argument;

packages/clerk-js/src/core/clerk.ts
SokratisVidros: I agree. We will do that separately. There is a relevant [epic](https://www.notion.so/clerkdev/Revamp-ClerkJS-setSessionz-9def3ea6a35b464baa69ab9f83bc5c91).

packages/types/src/clerk.ts
alex-ntousias: @SokratisVidros Initially I've had it like that. 
After a chat with Peter though, it became clear that it's better to just have it as id. It simplifies the code by avoiding certain if statements, and makes sure that the organization resource actually exists in the user's memberships (because we have to iterate through them to find it).

The downside is that the caller, if he has an organization resource, he will have to pass `organization.id`, which didn't seem like a big deal.
WDYT?

packages/clerk-js/src/core/clerk.ts
alex-ntousias: Alright, makes sense. I'll change it.

packages/types/src/clerk.ts
SokratisVidros: I see. It's a good DX practise to accept both in the JS/TS world. Let's do it.

PR #268, opened 2022-05-29
(v4) Appearance object - Classname generation

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

- Use the `ElementsConfig` type to strictly type all public types as well as all internal helpers in order to avoid accidental breaking changes
- Introduce the `descriptors.*` and `descriptors.*.setId` APIs used to bind a `customizable` component to the correct prop of the `appearance` object
- Propagate `appearance` prop changes from ClerkProvider to clerkjs (reactive propagation)
- Apply predefined stable clerk classnames (`cl-socialButtons-buttonIcon`)
- Apply predefined stable id classnames (`cl-id-google`)
- Apply predefined stable state classnames (`cl-loading`)
- Parse and apply custom CSS rules from the appearance object
- Read and append custom classnames from the appearance object
- Support selectors with ids
- Support selectors with states
- Support selectors with id + states
- Support all the above as nested selectors, eg: `appearance = { elements: { ... }, signIn: { elements: { ... } }. signUp: { elements: {...} }` 

Quick preview: 
image


packages/types/src/appearance.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export type ElementObjectKey<K extends string> = K extends `${infer Parent}-${infer Rest}`
+  ? `${Parent}${Capitalize<Rest>}`
+  : K;
+
+/**
+ * A map that describes the possible combinations we need to generate
+ * for each unique base element
+ */
+export type ElementsConfig = {
```
nikosdouvlis, line 52: Since you were interested, here's how the type generation happens cc: @colinclerk 
This file generates the public `Appearance` type using the `ElementsConfig` internal type as the single source of truth.

The generation of the actual runtime values our internal elements use happens here: packages/clerk-js/src/v4/customizables/elementDescriptors.ts 

When created, every `ElementDescriptor` object is bound to the `ElementsConfig` type. They are used to connect an element to a specific key of the appearance object. Since both the public `Appearance` type and the internal `ElementDescriptor` objects are using the `ElementsConfig` type, changing the one without updating the other would break - introducing accidental breaking changes to the customisation API should be pretty hard now

packages/clerk-js/src/v4/customizables/elementDescriptors.ts
```diff
@@ -0,0 +1,87 @@
...
+export const CLASS_PREFIX = 'cl-';
+export const ID_CLASS_PREFIX = 'cl-id-';
+export const OBJ_KEY_DELIMITER = '__';
+
+/**
+ * This object is strictly typed using the ElementsConfig type
+ * and is used as the single source of truth to generate the
+ * descriptor map
+ */
+export const APPEARANCE_KEYS: Readonly<Array<keyof ElementsConfig>> = [
```
nikosdouvlis, line 12: This needs to be kept in sync with the `ElementConfig` type manually. This is by design - `ElementsConfig` is the single source of truth, so `APPEARANCE_KEYS` can't change unless we make a change to `ElementsConfig` as well

packages/clerk-js/src/v4/Components.tsx
```diff
@@ -63,24 +56,18 @@ function assertDOMElement(element: HTMLElement): asserts element {
...
-    nodes: new Map(),
-  };
-
   static render(clerk: Clerk, environment: EnvironmentResource, options: ClerkOptions): Components {
     /**  Merge theme retrieved from the network with user supplied theme options. */
-    injectTheme(
-      environment.displayConfig.theme,
-      deepCamelToSnake(options.theme || {}) as DeepPartial<DisplayThemeJSON>,
-    );
+    // TODO: Remove all helpers
```
SokratisVidros, line 61: @nikosdouvlis I'd suggest we do the network/prop merging in Clerk.ts to keep the Components renderer clean from external dependencies.

SokratisVidros: This is part of a future PR of course.

packages/clerk-js/src/v4/customizables/AppearanceProvider.tsx
```diff
@@ -0,0 +1,26 @@
...
+};
+
+const [AppearanceContext, useAppearance] = makeContextAndHook<AppearanceContextValue>('AppearanceContext');
+
+type AppearanceProviderProps = React.PropsWithChildren<{
+  appearance: Appearance | undefined;
+}>;
+
+const AppearanceProvider = (props: AppearanceProviderProps) => {
+  console.log('AppearanceProvider', props.appearance);
```
SokratisVidros, line 18: 🙃 Please remove.

packages/clerk-js/src/v4/customizables/classGeneration.ts
```diff
@@ -0,0 +1,147 @@
...
+) => {
+  return elemId ? cn + ' ' + elemDescriptor.getTargettableIdClassname(elemId) : cn;
+};
+
+const addClerkTargettableStateClass = (cn: string, state: ElementState | undefined) => {
+  return state ? cn + STATE_PROP_TO_CLASSNAME[state] : cn;
+};
+
+const addEmojiSeparator = (cn: string) => {
+  return cn + ' 🔒️';
```
SokratisVidros, line 53: 💯

packages/clerk-js/src/v4/elements/FlowCard.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { generateFlowMetadataClassname } from '../customizables/classGeneration';
+import { CardStateProvider, FlowMetadata, FlowMetadataProvider } from './contexts';
+
+type FlowCardRootProps = React.PropsWithChildren<FlowMetadata>;
+
+const FlowCardRoot = (props: FlowCardRootProps): JSX.Element => {
+  const { children, flow, page, ...rest } = props;
+
+  return (
+    <FlowMetadataProvider
```
SokratisVidros, line 13: ❓ Can you please elaborate on this element? Does it block the UI from changing until a condition is met? It seems to me that we need a Wizard element.

packages/clerk-js/src/v4/elements/utils.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
SokratisVidros, line 1: 🙃 This can live outside v4 if that's not causing any other troubles.

packages/clerk-js/src/v4/signIn/SignInStart.tsx
```diff
@@ -149,16 +152,64 @@ export function _SignInStart(): JSX.Element {
...
+        <Header.Root>
+          <Header.Title>Sign in</Header.Title>
+          <Header.Subtitle>to continue to [Appname]</Header.Subtitle>
+        </Header.Root>
+        <Flex
+          direction={'col'}
+          elementDescriptor={descriptors.authOptions}
+        >
+          <SocialButtons.Root />
+          <hr style={{ width: '100%' }} />
```
SokratisVidros, line 174: 🔧 This should be also an element.

packages/types/src/appearance.ts
SokratisVidros: Very nice!

SokratisVidros (approved):
🎖️ 💅

packages/clerk-js/src/v4/elements/FlowCard.tsx
nikosdouvlis: (this component has been changed and now also renders the InternalThemeProvider)

This one provides the necessary flow/page metadata, aka which flow/page the nested customisables should use as their relative position inside the `appearance` object.

It also renders an `InternalThemeProvider` that's responsible to provide the base `theme` for our styled system (used by emotion). This also needs to be handled on a per-flow basis, as the appearance object also supports nesting `variables` objects

PR #272, opened 2022-06-01
Fixes on `setActive`

alex-ntousias:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

This PR includes some fixes on bugs and omissions that were found after testing the newly introduced `setActive`.
In particular, it includes the following:
* Removed some conditions where we were checking whether the `session` param was given in `setActive`. We don't need them because we use the current session if no session was passed. Also, they were wrong because they were preventing execution when the passed `session` was `null`.
* Add the organization in the resources that are being emitted to the listeners.
* Avoid deprecation of `setSession` just yet, given how critical it is. In fact, we reverted the implementation of `setSession` to avoid calling `setActive` internally until we're sure that everything is working as expected.
* Modified the `organization` param of `setActive` to accept an organization resource in addition to string and null that we’ve had.


packages/clerk-js/src/core/clerk.ts
```diff
@@ -296,7 +301,7 @@ export default class Clerk implements ClerkInterface {
     //1. setLastActiveSession to passed usersession (add a param).
     //   Note that this will also update the session's active organization
     //   id.
-    if (session && typeof document != 'undefined' && document.hasFocus()) {
+    if (typeof document != 'undefined' && document.hasFocus()) {
```
igneel64, line 304: ❓ 
Why was the check removed here ?

igneel64: Scratch it, saw `#touchLastActiveSession` handles falsy values.

packages/clerk-js/src/ui/contexts/CoreClerkContextWrapper.tsx
```diff
@@ -23,6 +23,7 @@ export function CoreClerkContextWrapper(props: CoreClerkContextWrapperProps): JS
     client: clerk.client,
     session: clerk.session,
     user: clerk.user,
+    organization: clerk.organization,
```
igneel64, line 26: This here just adds the state on the Wrapper. No effect currently but will be initially used in `useOrganization` with the addition of the ctx.

packages/clerk-js/src/core/clerk.ts
alex-ntousias: Yep, I actually removed these checks to align it with the `setSession` implementation. The only difference is basically that in the `setActive` the session can `undefined`. But if it is, we use the current session, so at this point we have the same state as `setSession`, so no need for the extra check.

igneel64 (approved):

PR #273, opened 2022-06-01
feat(clerk-js): Support multiple unverified identifiers during Sign up flow

chanioxaris:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

During a signUp, might there be more than one identifier that need verification (e.g email & phone). Support the above flow, by teaching our accounts components to properly redirect to the corresponding screens

### DIFF
--- a/packages/clerk-js/src/ui/common/verifyMagicLink/VerifyMagicLink.tsx
+++ b/packages/clerk-js/src/ui/common/verifyMagicLink/VerifyMagicLink.tsx
@@ -2,36 +2,42 @@ import { isMagicLinkError, MagicLinkErrorCode } from 'core/resources/internal';
 import React, { useEffect, useState } from 'react';
 import type { MagicLinkVerificationStatusModalProps } from 'ui/common';
 import { MagicLinkVerificationStatusModal } from 'ui/common';
-import { useCoreClerk } from 'ui/contexts';
+import { useCoreClerk, useCoreSignUp } from 'ui/contexts';
 import { useNavigate } from 'ui/hooks';
+import { completeSignUpFlow } from 'ui/signUp/util';
 import type { VerificationStatus } from 'utils/getClerkQueryParam';
 
 type VerifyMagicLinkProps = Pick<MagicLinkVerificationStatusModalProps, 'successHeader'> & {
   redirectUrlComplete?: string;
   redirectUrl?: string;
+  verifyEmailPath?: string;
+  verifyPhonePath?: string;
 };
 
 export function VerifyMagicLink({
   successHeader,
   redirectUrlComplete,
   redirectUrl,
+  verifyEmailPath,
+  verifyPhonePath,
 }: VerifyMagicLinkProps): JSX.Element | null {
   const { handleMagicLinkVerification } = useCoreClerk();
   const { navigate } = useNavigate();
-
+  const signUp = useCoreSignUp();
   const [verificationStatus, setVerificationStatus] = useState<VerificationStatus>('loading');
 
   useEffect(() => {
     const verify = async () => {
       try {
-        await handleMagicLinkVerification(
-          {
-            redirectUrlComplete,
-            redirectUrl,
-          },
-          navigate,
-        );
+        await handleMagicLinkVerification({ redirectUrlComplete, redirectUrl }, navigate);
         setVerificationStatus('verified');
+
+        return completeSignUpFlow({
+          signUp,
+          verifyEmailPath,
+          verifyPhonePath,
+          navigate,
+        });
       } catch (err) {
         let status: VerificationStatus = 'failed';
         if (isMagicLinkError(err) && err.code === MagicLinkErrorCode.Expired) {
--- a/packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
+++ b/packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
@@ -1,4 +1,3 @@
-import { SignUpResource } from '@clerk/types';
 import React from 'react';
 import type { FieldState } from 'ui/common';
 import {
@@ -15,6 +14,7 @@ import { Body, Header } from 'ui/common/authForms';
 import { useCoreClerk, useCoreSignUp, useEnvironment, useSignUpContext } from 'ui/contexts';
 import { useNavigate } from 'ui/hooks';
 import { SignUpForm } from 'ui/signUp/SignUpForm';
+import { completeSignUpFlow } from 'ui/signUp/util';
 
 import { SignInLink } from './SignInLink';
 import {
@@ -32,7 +32,7 @@ function _SignUpContinue(): JSX.Element | null {
   const { navigate } = useNavigate();
   const { displayConfig, userSettings } = useEnvironment();
   const { attributes } = userSettings;
-  const { setActive } = useCoreClerk();
+  const { setSession } = useCoreClerk();
   const { navigateAfterSignUp } = useSignUpContext();
   const [activeCommIdentifierType, setActiveCommIdentifierType] = React.useState<ActiveIdentifier>(
     getInitialActiveIdentifier(attributes, userSettings.signUp.progressive),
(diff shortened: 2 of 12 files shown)


packages/clerk-js/src/ui/common/verifyMagicLink/VerifyMagicLink.tsx
```diff
@@ -18,20 +19,21 @@ export function VerifyMagicLink({
...
-            redirectUrl,
-          },
-          navigate,
-        );
+        await handleMagicLinkVerification({ redirectUrlComplete, redirectUrl }, navigate);
         setVerificationStatus('verified');
+
+        return CompleteSignUpFlow({
+          signUp,
+          signUpUrl: displayConfig.signUpUrl,
```
SokratisVidros, line 34: ⛔ @chanioxaris urls should always be fetched from the useCoreSignUp or useCoreSignIn contexts as that's where we compute their final value based on props and Dashboard configuration. In cases where a component such as VerifyMagicLink is shared across multiple ClerkJS components, make the signUpUrl a component prop and pass its value based on the SignUp Or SignIn context.

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
```diff
@@ -100,24 +100,15 @@ function _SignUpContinue(): JSX.Element | null {
...
-  };
 
-  const completeSignUpFlow = (su: SignUpResource) => {
-    if (su.status === 'complete') {
-      return setActive({
-        session: su.createdSessionId,
-        beforeEmit: navigateAfterSignUp,
+      return CompleteSignUpFlow({
+        signUp: res,
+        signUpUrl: displayConfig.signUpUrl,
```
SokratisVidros, line 106: ⛔ Ditto: about signUpUrl

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
```diff
@@ -100,24 +100,15 @@ function _SignUpContinue(): JSX.Element | null {
...
+        handleComplete: () => setActive({ session: res.createdSessionId, beforeEmit: navigateAfterSignUp }),
```
SokratisVidros, line 107: 🙃 Let's keep useSession for now. We will change it to setActive in a separate PR. Is this correct @igneel64 and @alex-ntousias ?

packages/clerk-js/src/ui/signUp/SignUpStart.tsx
```diff
@@ -88,7 +89,18 @@ function _SignUpStart(): JSX.Element {
...
         formState.emailAddress.setValue(res.emailAddress || '');
-        void completeSignUpFlow(res);
+
+        // In case we are in a Ticket flow and the sign up is not complete yet, update the state
+        // to render properly the SignUp form with other available options to complete it (e.g. OAuth)
+        if (res.status === 'missing_requirements') {
+          setMissingRequirementsWithTicket(true);
+        }
+
+        void CompleteSignUpFlow({
```
SokratisVidros, line 99: 🙃 
```suggestion
        return CompleteSignUpFlow({
```

packages/clerk-js/src/ui/signUp/SignUpStart.tsx
```diff
@@ -88,7 +89,18 @@ function _SignUpStart(): JSX.Element {
       .create(signUpParams)
       .then(res => {
```
SokratisVidros, line 90:
```suggestion
      .then(signUp => {
```

packages/clerk-js/src/ui/signUp/util.ts
```diff
@@ -0,0 +1,27 @@
+import { SignUpResource } from '@clerk/types';
+
+type CompleteSignUpFlowProps = {
+  signUp: SignUpResource;
+  signUpUrl?: string;
+  navigate: (to: string) => Promise<void>;
+  handleComplete?: () => Promise<void>;
+};
+
+export const CompleteSignUpFlow = ({
```
SokratisVidros, line 10: 🙃 Since this is a utility function you can just name it `completeSignUpFlow`. Components are usually titleized.

SokratisVidros (changes requested):

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
igneel64: Yeap, as discussed yesterday, we are moving `setActive` to experimental currently. We will be changing it to `setSession`.

packages/clerk-js/src/ui/signUp/util.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (signUp.status === 'complete') {
+    return handleComplete && handleComplete();
+  } else if (signUp.status === 'missing_requirements') {
+    if (signUp.unverifiedFields?.includes('email_address')) {
+      return navigate(signUpUrl ? signUpUrl + '/verify-email-address' : 'verify-email-address');
+    }
+    if (signUp.unverifiedFields?.includes('phone_number')) {
+      return navigate(signUpUrl ? signUpUrl + '/verify-phone-number' : 'verify-phone-number');
+    }
+  }
```
igneel64, line 25: ❓ 
If we pass the signUpUrl from contexes like `useCoreSignUp` , will the interpolation work as expected ?
Reminder these URLs go through https://github.com/clerkinc/javascript/blob/main/packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx#L60-L67

cc: @SokratisVidros @nikosdouvlis

packages/clerk-js/src/ui/signUp/SignUpContinue.tsx
chanioxaris: Restored back to setSession instead of setActive e7c2ed56b2edb6fc49fd121a4345e644582fab9d

packages/clerk-js/src/ui/signUp/util.ts
SokratisVidros: @igneel64 it should work as expected if we use the `buildURL` function and not just concatenate strings. Up until now, no concatenation was required as we were just doing `navigate('verify-email-address');`.

packages/clerk-js/src/ui/signUp/SignUpVerifyEmailAddressWithMagicLink.tsx
```diff
@@ -57,15 +59,11 @@ export function SignUpVerifyEmailAddressWithMagicLink(): JSX.Element {
...
-  };
-
-  const completeSignUpFlow = async (su: SignUpResource) => {
-    if (su.status === 'complete') {
-      return setActive({
-        session: su.createdSessionId,
-        beforeEmit: navigateAfterSignUp,
+      await completeSignUpFlow({
+        signUp: su,
+        signUpUrl: displayConfig.signUpUrl,
```
SokratisVidros, line 64: @chanioxaris To alleviate the confusion around URLs for now, can you please remind me why do we need to pass signUpUrl in the `completeSignUpFlow` helper? 

Can we do `navigate('verify-email-address');` and `navigate('verify-phone-number')` in every flow?

chanioxaris: The `navigate('verify-email-address')` will only work if we are on the `sign-up` path. It won't work for example if we are at `sign-up/continue` -> `sign-up/continue/verify-email-address` as that's not the correct path

SokratisVidros: I can see two options at this point:

chanioxaris: Discussed offline and decided to go with option B for the moment, as a routing refactoring will come up within the next weeks

SokratisVidros (approved):

PR #285, opened 2022-06-08
feat(clerk-js,clerk-react,types): Introduce useOrganization, useOrganizationList v2

igneel64:
…izationList v2

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description


packages/react/src/hooks/useOrganization.ts
```diff
@@ -0,0 +1,106 @@
...
+  );
+
+  const { data: membershipList } = useSWR(
+    membershipListParams
+      ? composeOrganizationResourcesUpdateKey(organization, lastOrganizationMember, 'memberships')
+      : null,
+    currentOrganizationMemberships,
+  );
+
+  return {
```
colinclerk, line 82: Can we block until `invitationList` and `membershipList` are loaded?

I believe this allows `isLoaded: true` while `invitationList` and `membershipList` are still loading.

colinclerk (changes requested):
I don't think we should move forward with this strategy for reactivity.  We don't have a good abstraction for managing `lastOrganizationInvitation` and `lastOrganizationMembership` and it's evident from the repeated usage of `__unstable__`.

Instead, can we add something to BaseResource._fetch that triggers revalidations after any POST completes?

igneel64:
Discussed in length offline, the functionality presented is fully reactive and the API fits the use-cases as specified.
The fixes on the mechanism, since internal, would be addressed at https://www.notion.so/clerkdev/Improve-reactivity-mechanism-of-clerk-js-clients-0ed49f1cb005473ead6245d41276f26e

nikosdouvlis (approved):

PR #302, opened 2022-07-01
feat(edge,nextjs): Compatibility with API middleware experimental edge

igneel64:
Compatibility with API middleware experimental edge runtime.

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

With the release of Next.js 12.2, many changes came forth on the experimental edge middleware support as it now becomes GA.

The changes on 12.2 affect us in the following ways:
1. Middleware files _(now at a different location)_ cannot return a response.
2. API routes can experimentaly use the `edge` runtime.
3. Breaking API changes for edge runtimes.

In this PR:
1. We address the breaking changes on our `@clerk/nextjs/edge-middleware` and `@clerk/edge/vercel-edge` functions.
2. We allow for seamless change in API experimental edge runtime in the following fashion:
```ts
// Normal middleware
import { withAuth } from "@clerk/nextjs/api";

export default withAuth(handler);

// Edge middleware. Works exactly as is without the need for the developer to change import paths
import { withAuth } from "@clerk/nextjs/api";

export default withAuth(handler);

export const config = {
 runtime: 'experimental-edge'
}
```

## Important considerations for (2)
1. There is no complete parity between the node-sdk exports and the clerk-edge exports. E.g. how would we handle `requireAuth` which is not valid for edge middleware ? 
2. Typings are still being worked on if we want them to work seamlessly, if possible.

Personally I am really skeptical if (2) this  is worth it at this point. The developer would need to change his handlers either way, as `edge` runtime handlers have different APIs than `nodejs`. I think we are bearing unneeded complexity for minimal increase in DX, if any at all, as confusion may also incur.

### DIFF
--- a/packages/edge/src/vercel-edge/index.ts
+++ b/packages/edge/src/vercel-edge/index.ts
@@ -76,12 +76,12 @@ export function withEdgeMiddlewareAuth(
 ): any {
   return async function clerkAuth(req: NextRequest, event: NextFetchEvent) {
     const { loadUser, loadSession, jwtKey, authorizedParties } = options;
-    const cookieToken = req.cookies['__session'];
+    const cookieToken = req.cookies.get('__session');
     const headerToken = req.headers.get('authorization');
     const { status, interstitial, sessionClaims, errorReason } = await vercelEdgeBase.getAuthState({
       cookieToken,
       headerToken,
-      clientUat: req.cookies['__client_uat'],
+      clientUat: req.cookies.get('__client_uat'),
       origin: req.headers.get('origin'),
       host: req.headers.get('host') as string,
       userAgent: req.headers.get('user-agent'),
--- a/packages/nextjs/api.js
+++ b/packages/nextjs/api.js
@@ -1,1 +1,9 @@
-module.exports = require('./dist/api');
+let exportLib;
+if (process.env.NEXT_RUNTIME === 'edge') {
+  exportLib = require('./dist/edge-middleware');
+  exportLib.withAuth = exportLib.withEdgeMiddlewareAuth;
+} else {
+  exportLib = require('./dist/api');
+}
+
+module.exports = exportLib;
--- a/package-lock.json
+++ b/package-lock.json
@@ -5705,7 +5705,23 @@
     "node_modules/@next/env": {
       "version": "12.0.10",
       "resolved": "https://registry.npmjs.org/@next/env/-/env-12.0.10.tgz",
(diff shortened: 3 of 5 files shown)


colinclerk (commented):
To avoid a breaking change, can we keep  `withEdgeMiddlewareAuth` the same for any devs who want to remain on 12.0-12.1?

Then, can we make a new function in edge-middleware to enable this?  Importing from "edge-api" would support the new way to retrieve cookie:
```
exportLib = require('./dist/edge-api');
```

Then, I believe this would support 12.07+.

igneel64:
@colinclerk 
If you check the last commit, I introduced a compatibility handler for cookies, since it is the only breaking change for us, which I tested on 12.2 and > 12.2.
WDYT ?

colinclerk:
Cookie compat looks good. I believe we can revert the dep back to `"next": "^12.0.7"`

Besides that, let's get `requireAuth` working on 12.2 experimental-edge and this should be good to release. (Only `requireAuth`, not `requireServerSideAuth` or `requireEdgeMiddlewareAuth`)

packages/edge/src/vercel-edge/index.ts
```diff
@@ -94,13 +119,14 @@ export function withEdgeMiddlewareAuth(
...
-      return new NextResponse(interstitial, {
-        headers: { 'Content-Type': 'text/html', 'Auth-Result': errorReason || '' },
-        status: 401,
-      });
+      return interstitialResponse(interstitial as string, errorReason);
     }
 
     if (status === AuthStatus.SignedOut) {
+      if (options.strict) {
+        return signedOutResponse();
```
colinclerk, line 127: I think we still want to pass errorReason here and set the Auth-Result error (we set it for interstitial and signed out states)

igneel64: Sure, good catch!

packages/nextjs/api.js
```diff
@@ -1 +1,10 @@
-module.exports = require('./dist/api');
+let exportLib;
+if (process.env.NEXT_RUNTIME === 'edge') {
```
nikosdouvlis, line 2: ✅

nikosdouvlis (approved):

PR #308, opened 2022-07-08
[WIP] Attempt to resolve hooks error

colinclerk:
Customer reported a hooks error on `useOrganization` in Discord. It is not clear why this made it through tests, but we do indeed appear to have conditional hooks here.

This PR is currently untested. We must ensure the organizations-demo still works before merging/publishing.

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/packages/react/src/hooks/useOrganization.ts
+++ b/packages/react/src/hooks/useOrganization.ts
@@ -62,14 +62,14 @@ export const useOrganization: UseOrganization = ({
   }
 
   const { data: invitationList, isValidating: isInvitationsLoading } = useSWR(
-    (isomorphicClerk.loaded && invitationListParams)
+    (isomorphicClerk.loaded && invitationListParams && session && organization)
       ? composeOrganizationResourcesUpdateKey(organization, lastOrganizationInvitation, 'invitations')
       : null,
     pendingInvitations,
   );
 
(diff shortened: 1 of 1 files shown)


nikosdouvlis:
> Is this a weird typescript thing? I can't figure out the scenario where this will actually be null or undefined, since `shouldFetch` should prevent the call.

Looking at this SWR type:
```tsx
    <Data = any, Error = any, SWRKey extends Key = null>(key: SWRKey, fetcher: Fetcher<Data, SWRKey> | null): SWRResponse<Data, Error>;
```

the `data` prop that the hook returns is typed based on the return value of the fetcher function. The `() => {}` returns void , so TS infers that there is a branch in the code for which the fetcher returns `void`, thus the error

PR #310, opened 2022-07-12
fix(clerk-js): Fix navigation when an error occurs during progressive sign up flow

chanioxaris:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

This commit fixes a logical error in our redirect logic during an OAuth flow and progressive sign up. If the external account was verified and were still missing requirements for sign up to complete, we redirected to the sign-up continue url. Although the right thing to do, is to first check if there is an error on the external account verification. Image a scenario where an instance uses sign-up restrictions for emails. The user tries to sign-up using an OAuth account whose email is not listed on the allowlist. We should redirect them back on the sign-up url and preview the error, instead of the sign up continue path

https://www.notion.so/clerkdev/PSU-asks-for-email-after-OAuth-flow-although-it-s-already-provided-when-Allowlists-are-on-e8129bd8b7ca488abc5f07ef5b382a93

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -628,12 +628,12 @@ export default class Clerk implements ClerkInterface {
       }
     }
 
-    if (su.externalAccountStatus === 'verified' && su.status == 'missing_requirements') {
-      return navigateToContinueSignUp();
-    }
-
     if (hasExternalAccountSignUpError(signUp)) {
       return navigateToSignUp();
+    }
+
+    if (su.externalAccountStatus === 'verified' && su.status === 'missing_requirements') {
+      return navigateToContinueSignUp();
     }
 
     return navigateToSignIn();
--- a/packages/clerk-js/src/core/clerk.test.ts
+++ b/packages/clerk-js/src/core/clerk.test.ts
@@ -829,6 +829,54 @@ describe('Clerk singleton', () => {
       });
     });
 
+    it('redirects user to signUp url if there is an external account signup error even if missing requirements', async () => {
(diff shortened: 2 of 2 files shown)


agis (approved):
💯

SokratisVidros (approved):
💯

packages/clerk-js/src/core/clerk.ts
```diff
@@ -628,14 +628,14 @@ export default class Clerk implements ClerkInterface {
...
 
-    if (su.externalAccountStatus === 'verified' && su.status == 'missing_requirements') {
-      return navigateToContinueSignUp();
-    }
-
     if (hasExternalAccountSignUpError(signUp)) {
       return navigateToSignUp();
     }
 
+    if (su.externalAccountStatus === 'verified' && su.status == 'missing_requirements') {
```
igneel64, line 635: 🔧 
`==` should be replaced with `===`

igneel64 (approved):

packages/clerk-js/src/core/clerk.ts
chanioxaris: Nice catch, updated

PR #319, opened 2022-08-02
feat(nextjs,edge,backend-core): Add getAuth and runClerkMiddleware pr…

igneel64:
Draft proposal

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

## Integration strategy proposal for Next.js 12.2 compatible APIs

This PR adds 2 new methods for use in the new `middleware` structure of Next.js > 12.2.
- `runClerkMiddleware` should be run on the middleware file
- `getAuth` can be run on both `gssp` and `Node.js API routes`
- API calls work for`gssp`, `middleware` and Node.js API routes from the same import.

```ts
// gssp
import { getAuth, clerkApi } from "@clerk/nextjs/server";

export default function Home() {
  return ();
}

export async function getServerSideProps(ctx) {
  const auth = getAuth(ctx.res);

  const sessions = await clerkApi.sessions.getSessionList();
  console.log("gssp sessions: ", sessions);
  return { props: {} };
}

// middleware
import { NextRequest } from "next/server";
import { runClerkMiddleware, clerkApi } from "@clerk/nextjs/server";

export async function middleware(request: NextRequest) {
  const { auth, response } = await runClerkMiddleware(request);
  const sessions = await clerkApi.sessions.getSessionList();
  console.log("sessions middleware", sessions);
  return response;
}
```

The proposal uses a header(Clerk-Auth-Middleware) which is passed from the middleware file down to `gssp` and API routes. There `getAuth` retrieves the values from the decoded key  etc., adds the appropriate `getToken` version and removes it from the client. In the worst case scenario where the client does not use `getAuth`, the header retrieved on the client is the decoded JWT.

## Issues/12.2 stuff
- Intersitital rewrite does not receive the `Authorization` header properly. Using `NextResponse.rewrite` does calls a proxy request on the Next.js server but does not seem to add any additional headers on the request, it only adds them on the final response. See more [here](https://github.com/vercel/next.js/blob/9a1b756defbdc735a136504aae7fa1a6212eefb4/packages/next/server/next-server.ts#L502). The only place I found which can add data on the original request is the query string which persists for the final rewrite request. **This is a blocker for interstitial.**
- As it [seems from GH](https://github.com/vercel/next.js/discussions/38650) the `req` object is immutable from the middleware so the `getAuth` method cannot retrieve data from the `req` but from the `res` object.
- The `experimental-edge` API routes seem to run before the middleware file which seems unintuitive to me. Using experimental-edge API together with `middleware` seems to be having these issues.

### WIP
- Typings.
- Testing more cases.
- Add options on `getAuth` which were present on the wrappers such as `authorizedParties` etc.
- Check if we can throw at some point when we detect no removal of `Clerk-Auth-Middleware`

Based on: https://www.notion.so/clerkdev/Next-js-12-2-New-integration-strategy-move-auth-to-middleware-ts-0293c1f53a134dc08ed9b4cfee7b06fd

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5702,12 +5702,6 @@
       "integrity": "sha512-W6CLUJ2eBMw3Rec70qrsEW0jOm/3twwJv21mrmj2yORiaVmVYGS4sSS5yUwvQc1ZlDLYGPnClVWmUUMagKNsfA==",
       "peer": true
     },
-    "node_modules/@next/env": {
-      "version": "12.0.10",
-      "resolved": "https://registry.npmjs.org/@next/env/-/env-12.0.10.tgz",
-      "integrity": "sha512-mQVj0K6wQ5WEk/sL9SZ+mJXJUaG7el8CpZ6io1uFe9GgNTSC7EgUyNGqM6IQovIFc5ukF4O/hqsdh3S/DCgT2g==",
-      "dev": true
-    },
     "node_modules/@next/swc-android-arm-eabi": {
       "version": "12.2.0",
       "resolved": "https://registry.npmjs.org/@next/swc-android-arm-eabi/-/swc-android-arm-eabi-12.2.0.tgz",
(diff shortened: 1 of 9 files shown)


colinclerk:
Can you build an example repo showing the ordering problem for middleware vs edge api routes?  That's very surprising.

I posted a comment on the issue of retrieving middleware data elsewhere, surprising there's an issue there.

Can continue the interstitial solution discussion in Slack.

Thanks!

Issue #316, opened 2022-07-27
Property 'env' is missing in type is missing in Next API Request with auth prop


calebfaruki:
## Package + Version

- `@clerk/nextjs@3.8.1`
- `@clerk/types@2.20.0`
- `@clerk/clerk-sdk-node@3.9.0`

## Description

```
Argument of type '(req: WithAuthProp<NextApiRequest>, res: NextApiResponse) => void' is not assignable to parameter of type 'Middleware'.
  Types of parameters 'req' and 'req' are incompatible.
    Type 'WithAuthProp<Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>>' is not assignable to type 'WithAuthProp<NextApiRequest>'.
      Property 'env' is missing in type 'Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>> & { auth: { sessionId: string | null; userId: string | null; getToken: ServerGetToken; claims: ClerkJWTClaims | null; }; }' but required in type 'NextApiRequest'.ts(2345)
```

Here is the code:

```
// pages/api/get-authenticated-user.ts
import { withAuth, WithAuthProp } from '@clerk/nextjs/api';
import type { NextApiRequest, NextApiResponse } from 'next';

export default withAuth(
  (req: WithAuthProp<NextApiRequest>, res: NextApiResponse) => {
    const { sessionId } = req.auth;

    if (!sessionId) {
      return res.status(401).json({ id: null });
    }

    return res.status(200).json({ id: sessionId });
  }
);
```

SokratisVidros:
Hi there. Thanks for reporting. Which version of nextjs are you using?

erhise:
Can confirm this one => using `@clerk/nextjs@3.8.1` `next@12.2.3`

drasch:
I'm running next 12.1.6

I too am running into this issue with Typescript. `withAuth` is now looking for a function with `express` `Request` and `Response` instead of `NextApiRequest` and `NextApiResponse`


This is what I had to change as a workaround for typescript. 


```
+import type { Request, Response } from 'express' 


-  async (req: WithAuthProp<NextApiRequest>, res: NextApiResponse) => {                      
+  async (req: WithAuthProp<Request>, res: Response) => {                                    
```

Unfortunately for some other libraries I had to use `req as any as NextApiRequest` because the types were incorrect.

Renetnielsen:
+1 for this issue

Renetnielsen:
[...]

PR #320, opened 2022-08-02
Fix typings for backend middlewares and helpers

SokratisVidros:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

#316


packages/nextjs/src/api/index.ts
```diff
@@ -1,6 +1,65 @@
...
+  WithAuthProp,
+} from '@clerk/clerk-sdk-node';
+import type { NextApiRequest, NextApiResponse } from 'next';
 
-export { default } from '@clerk/clerk-sdk-node';
+export default Clerk;
 export * from '@clerk/clerk-sdk-node';
+
+// https://nextjs.org/docs/api-routes/api-middlewares#connectexpress-middleware-support
+function runMiddleware(req: NextApiRequest, res: NextApiResponse, fn: ClerkMiddleware) {
```
igneel64, line 14: 🙃 
Should this go to a `utils` file since it is kind of a detail to the usage of `withAuth` etc ?

packages/nextjs/src/api/index.ts
```diff
@@ -1,6 +1,65 @@
...
+      await runMiddleware(req, res, Clerk.expressRequireAuth(options));
+      return handler(req, res);
+    } catch (error) {
+      // @ts-ignore
+      const errorData = error.data || { error: error.message };
+      // @ts-ignore
+      res.statusCode = error.statusCode || 401;
+      /**
+       * Res.json is available in express-like environments.
+       * Res.send is available in express-like but also Fastify.
```
igneel64, line 59: ❓  
This will be a breaking change for Fastify clients right ? _The way we used it up until now can be seen [here](https://github.com/clerkinc/clerk-fastify-react-prisma-starter/blob/main/packages/server/src/auth/clerkHandler.ts)._

packages/sdk-node/examples/next/package.json
```diff
@@ -1,17 +1,24 @@
 {
-  "name": "next",
-  "version": "0.2.0",
+  "name": "my-app",
```
igneel64, line 2: ❓ 
Should this example be moved on the `nextjs` package or be removed entirely ?
Also if we decide to keep this, don't we need to have the `@clerk/nextjs` package in the dependencies ? _(The way it was done previously was using the locally compiled package.)_

packages/sdk-node/src/Clerk.ts
```diff
@@ -50,21 +50,38 @@ export type RequireAuthProp<T> = T & {
...
+export type MiddlewareOptionsInLooseMode = {
   onError?: (error: ClerkAPIResponseError) => unknown;
   authorizedParties?: string[];
   jwtKey?: string;
   strict?: boolean;
 };
 
-// eslint-disable-next-line @typescript-eslint/no-empty-function
-const noop: NextFunction = () => {};
+export type MiddlewareOptionsInStrictMode = MiddlewareOptionsInLooseMode & {
```
igneel64, line 74: ❓ 
Do we need to export this ?

packages/sdk-node/src/Clerk.ts
```diff
@@ -309,7 +324,7 @@ export default class Clerk extends ClerkBackendAPI {
           sessionId: null,
           getToken: createSignedOutState().getToken,
           claims: null,
-        };
+        } as WithAuthProp<Request>;
```
igneel64, line 327: ^^ ditto

packages/nextjs/src/api/index.ts
```diff
@@ -1,6 +1,65 @@
...
+  });
+}
+
+type NextApiHandlerWithAuth<T = any> = (
+  req: WithAuthProp<NextApiRequest>,
+  res: NextApiResponse<T>,
+) => void | Promise<void>;
+
+// Set the session on the request and then call provided handler
+export function withAuth(handler: NextApiHandlerWithAuth, options?: ClerkMiddlewareOptions): NextApiHandlerWithAuth {
```
igneel64, line 32: 🙃 
We can also put `withAuth` and `requireAuth` in a separate file and keep the index just for imports/exports. WDYT ?

packages/nextjs/src/api/index.ts
SokratisVidros: It's a very small function specific to Next.js so I'd rather keep it here for now.

packages/nextjs/src/api/index.ts
SokratisVidros: Good idea. Will do.

packages/sdk-node/src/Clerk.ts
```diff
@@ -50,21 +50,38 @@ export type RequireAuthProp<T> = T & {
...
-export type MiddlewareOptions = {
+type MiddlewareRequireAuthProp = (
+  req: RequireAuthProp<Request>,
+  res: Response,
+  next: NextFunction
+) => Promise<void>;
+
+export type Middleware = MiddlewareWithAuthProp | MiddlewareRequireAuthProp;
+
+export type MiddlewareOptionsInLooseMode = {
```
nikosdouvlis, line 67: ❓ Same question as below, is this a public or private type?

packages/nextjs/src/api/index.ts
SokratisVidros: Correct. Fastify clients will have to update to a new method coming from clerk-sdk-node

packages/sdk-node/src/Clerk.ts
SokratisVidros: Nope. Good catch.

packages/sdk-node/examples/next/package.json
SokratisVidros: Yes. Moving it.

packages/sdk-node/src/Clerk.ts
SokratisVidros: It's private. This is now fixed.

packages/nextjs/src/api/index.ts
SokratisVidros: Done.

packages/nextjs/examples/next/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@clerk/clerk-sdk-node": "file:../../dist",
```
igneel64, line 12: ❗

igneel64 (approved):

PR #321, opened 2022-08-02
Yourtallness/totp user profile

yourtallness:
## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

* User Profile MFA section will list TOTP if enabled
* If enabled, it can be disabled
* If not enabled, user can opt to add an MFA method
* When adding an MFA method, one can choose between TOTP & phone_code if available
* If only one choice is available, the choice screen is skipped & the user it directly taken to the applicable component
* TOTP card allows user to either scan a QR code or copy the secret / QR URI to the clipboard
* `useClipboard` hook implementation was copied from Chakra
* QR-code-as-SVG rendering by [qrcode.react](https://www.npmjs.com/package/qrcode.react) package

## TODO

NEEDS TESTS!!!

Default flag for phones needs tweaking to avoid confusing UX.


package.json
```diff
@@ -52,5 +52,9 @@
     "lerna": "lerna",
     "nuke": "rm -r node_modules; for d in packages/*/node_modules; do echo $d; rm -r $d; done",
     "yalc:all": " for d in packages/*/; do echo $d; cd $d; yalc push --replace; cd '../../'; done"
+  },
+  "dependencies": {
```
igneel64, line 56: 🔧 
This is not the right place for any package dependencies. This should be moved to the `clerk-js` pacakge.

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
```diff
@@ -1,129 +1,85 @@
...
-import { AddPhone, VerifyPhone } from './PhonePage';
-import { SuccessPage } from './SuccessPage';
-import { AddBlockButton } from './UserProfileBlockButtons';
+import { TileButton } from './TileButton';
+import { getSecondFactors } from './utils';
 
 export const MfaPage = withCardStateProvider(() => {
-  const title = 'Add authentication step';
+  const card = useCardState();
+  const { attributes } = useEnvironment().userSettings;
```
igneel64, line 18: 🙃 
Super minor but when we choose desctructuring, we do many levels in a single destructuring statement e.g.:
`const { userSettings: { attributes } } = useEnvironment();`

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
```diff
@@ -1,129 +1,85 @@
...
+  const user = useCoreUser();
+  const title = 'Add multifactor authentication';
+  const [selectedMethod, setSelectedMethod] = React.useState<VerificationStrategy | undefined>(undefined);
```
igneel64, line 21: 🙃 
I think this is the same as:
`  const [selectedMethod, setSelectedMethod] = React.useState<VerificationStrategy>();`

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
```diff
@@ -1,129 +1,85 @@
...
 
-  const ref = React.useRef<PhoneNumberResource | undefined>();
-  const wizard = useWizard({ defaultStep: 2 });
+  // Calculate second factors on first use only
+  const secondFactors = React.useMemo<string[]>(() => {
```
igneel64, line 24: 🙃 
This is uneeded `<string[]>` .

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
```diff
@@ -1,129 +1,85 @@
...
-  const enableMfa = (phone: PhoneNumberResource) => {
-    if (phone.verification.status !== 'verified') {
-      return onUnverifiedPhoneClick(phone);
-    }
+    return <ContentPage.Root headerTitle={title} />;
+  }
+
+  // If only phone_code is available, just render that
+  // Otherwise check if selected
+  if ((secondFactors.length == 1 && secondFactors[0] == 'phone_code') || selectedMethod === 'phone_code') {
```
igneel64, line 43: ^^ ditto

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
```diff
@@ -1,129 +1,85 @@
...
-      .setReservedForSecondFactor({ reserved: true })
-      .then(() => {
-        card.setIdle();
-        onSuccess();
-      })
-      .catch(err => handleError(err, [], card.setError));
-  };
+  // If only totp is available, just render that
+  // Otherwise check if selected
+  if ((secondFactors.length == 1 && secondFactors[0] == 'totp') || selectedMethod === 'totp') {
```
igneel64, line 49: ^^ ditto

packages/clerk-js/src/v4/UserProfile/MfaPhoneCodePage.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { NavigateToFlowStartButton } from './NavigateToFlowStartButton';
+import { ContentPage } from './Page';
+import { AddPhone, VerifyPhone } from './PhonePage';
+import { SuccessPage } from './SuccessPage';
+import { AddBlockButton } from './UserProfileBlockButtons';
+
+export const MfaPhoneCodePage = withCardStateProvider(() => {
+  const title = 'Add multifactor authentication';
+
+  const ref = React.useRef<PhoneNumberResource | undefined>();
```
igneel64, line 19: 🙃 
`const ref = React.useRef<PhoneNumberResource>();` is the same without using `| undefined` explicitly

packages/clerk-js/src/v4/UserProfile/MfaPhoneCodePage.tsx
```diff
@@ -0,0 +1,129 @@
...
+          : 'There are no available phone numbers to register for SMS code two-step authentication.'}
+      </Text>
+      <Col gap={2}>
+        {availableMethods.map(phone => {
+          const formattedPhone = stringToFormattedPhoneString(phone.phoneNumber);
+          const flag = getFlagEmojiFromCountryIso(parsePhoneString(phone.phoneNumber).iso);
+
+          return (
+            <ArrowBlockButton
+              // elementDescriptor={descriptors.socialButtonsButtonBlock}
```
igneel64, line 93: ❓ 
Are these comments here by mistake ?

packages/clerk-js/src/v4/UserProfile/MfaTOTPPage.tsx
```diff
@@ -0,0 +1,192 @@
...
+});
+
+type AddAuthenticatorAppProps = {
+  title: string;
+  onContinue: () => void;
+};
+
+type DisplayFormat = 'qr' | 'uri';
+
+const AddAuthenticatorApp = (props: AddAuthenticatorAppProps) => {
```
igneel64, line 48: 🙃 
I would put both `AddAuthenticatorApp` and `VerifyTOTP` on different files to reduce the length a bit.

packages/clerk-js/src/v4/UserProfile/RemoveResourcePage.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useReducer } from 'react';
```
igneel64, line 1: 🔧 
Unused.

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
yourtallness: Copy pasta, fixing

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
yourtallness: Fixing

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
yourtallness: Removing

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
```diff
@@ -1,129 +1,85 @@
...
-};
+    return sfs;
+  }, []);
 
-const AddMfa = (props: AddMfaProps) => {
-  const { onSuccess, title, onAddPhoneClick, onUnverifiedPhoneClick } = props;
-  const card = useCardState();
-  const user = useCoreUser();
-  const availableMethods = user.phoneNumbers.filter(p => !p.reservedForSecondFactor);
+  if (secondFactors.length == 0) {
```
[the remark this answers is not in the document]
yourtallness: Fixing

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
yourtallness: Fixing

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
yourtallness: Fixing

packages/clerk-js/src/v4/UserProfile/MfaPhoneCodePage.tsx
yourtallness: Fixing

packages/clerk-js/src/v4/UserProfile/RemoveResourcePage.tsx
yourtallness: Removing

packages/clerk-js/src/v4/UserProfile/MfaPhoneCodePage.tsx
yourtallness: This pre-existed, I just moved it to a different file

cc @nikosdouvlis

packages/clerk-js/src/v4/UserProfile/RemoveResourcePage.tsx
```diff
@@ -112,6 +112,20 @@ export const RemoveMfaPage = () => {
...
+export const RemoveMfaTOTPPage = () => {
+  const user = useCoreUser();
+
+  return (
+    <RemoveResourcePage
+      title='Remove multifactor authentication'
+      messageLine1='Verification codes from this authenticator will no longer be required when signing in.'
+      messageLine2='Your account may not be as secure. Are you sure you want to continue?'
+      successMessage='Multifactor authentication via authenticator application has been removed.'
+      deleteResource={() => Promise.resolve(user.disableTOTP())}
```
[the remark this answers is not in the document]
yourtallness: Reducing :-D

packages/clerk-js/package.json
```diff
@@ -44,10 +44,12 @@
     "@popperjs/core": "^2.4.4",
     "browser-tabs-lock": "^1.2.15",
     "classnames": "^2.3.1",
+    "copy-to-clipboard": "^3.3.2",
```
SokratisVidros, line 47: ❓ How does this affect the bundle size? Is it a big package to add?

nikosdouvlis: This lib should add about ~1kb to the final bundle, so it's okish 

Another alternative would be to extract the logic we need from that repo and make a custom hook instead, but this can be done as an optimization in the future

gkats: Not suggesting that we change anything in this PR, but:

- It looks like the [method](https://github.com/sudodoki/copy-to-clipboard/blob/193826f9859923d4e83c35093f3c436be00a162b/index.js#L79) that's used by the copy-to-clipboard package is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand#browser_compatibility). 
- There is a newer [clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) which is currently the recommended way to "copy to clipboard".
- Looks like [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#browser_compatibility) won't be an issue.

packages/clerk-js/src/v4/UserProfile/MfaPhoneCodePage.tsx
nikosdouvlis: Don't worry about them for not, I've already applies most elementDescriptor on the next branch directly. I 'll do a final pass today to apply the rest now that UP is done

packages/clerk-js/src/v4/UserProfile/AddAuthenticatorApp.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const card = useCardState();
+  const [totp, setTOTP] = React.useState<TOTPResource | undefined>(undefined);
+  const [displayFormat, setDisplayFormat] = React.useState<DisplayFormat>('qr');
+
+  React.useEffect(() => {
+    void user
+      .createTOTP()
+      .then((totp: TOTPResource) => setTOTP(totp))
+      .catch(err => handleError(err, [], card.setError));
+  }, [user]);
```
nikosdouvlis, line 32: This useEffect is not idempotent so we'll need to refactor some of the logic here during the true SSR project. Can we add a `// TODO: React18` comment, so it's easier to find in the future?

packages/clerk-js/src/v4/UserProfile/AddAuthenticatorApp.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  if (!totp && !card.error) {
```
nikosdouvlis, line 34: Did we reach a verdict regarding the position of the spinner? I feel the UX is better if the spinner is within the card, even if an error occurs afterward 

We can totally do this on a different PR

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
```diff
@@ -1,129 +1,88 @@
...
+        <Grid
+          columns={3}
+          gap={2}
         >
-          Add a phone number
-        </AddBlockButton>
+          <TileButton
+            icon={AuthenticatorApp}
+            onClick={() => setSelectedMethod('totp')}
+            sx={{ height: '96px' }}
```
nikosdouvlis, line 68: ❓ Why are we adding the `height` via `sx`? Can we set it inside TileButton directly? 

Also let's change the above to
```suggestion
            sx={t => ({ height: t.sizes.$24 })}
```

so it can be themed correctly when the user passes in  a custom `appearance.variables` config

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
```diff
@@ -1,129 +1,88 @@
...
+          >
+            Authenticator application
+          </TileButton>
+
+          <TileButton
+            icon={Mobile}
+            onClick={() => setSelectedMethod('phone_code')}
+            sx={{ height: '96px' }}
```
nikosdouvlis, line 76: Ditto

packages/clerk-js/src/v4/UserProfile/AddAuthenticatorApp.tsx
yourtallness: Adding

packages/clerk-js/src/v4/UserProfile/AddAuthenticatorApp.tsx
yourtallness: No strong opinion, will move inside card

packages/clerk-js/src/v4/UserProfile/MfaSection.tsx
```diff
@@ -1,82 +1,49 @@
...
+
   const mfaPhones = user.phoneNumbers
     .filter(ph => ph.verification.status === 'verified')
     .filter(ph => ph.reservedForSecondFactor)
     .sort(defaultFirst);
 
   return (
     <ProfileSection
-      title='Two-step authentication'
+      title='Multifactor authentication'
```
nikosdouvlis, line 30: ❓ Did we change the copy here again? 🙃 Some designs state multifactor, others have Two-step. Creating a slack thread so we can finalise this because I remember changing the copy here at least 2 times already!

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
yourtallness: Moving to `TileButton`

packages/clerk-js/src/v4/icons/index.ts
```diff
@@ -6,6 +6,8 @@
  */
 export { default as ArrowLeftIcon } from './arrow-left.svg';
 export { default as ArrowRightIcon } from './arrow-right.svg';
+export { default as AuthenticatorApp } from './authenticator-app.svg';
```
nikosdouvlis, line 9: We already have an icon named AuthApp that's the same, shall we use that instead? This one looks greyish so I guess it has a hardcoded color set within its definition

packages/clerk-js/src/v4/UserProfile/MfaPage.tsx
yourtallness: Moved to `TileButton`

packages/clerk-js/src/v4/UserProfile/MfaSection.tsx
yourtallness: I got this from Figma

Agreed, we should use a single term here

packages/clerk-js/src/v4/UserProfile/MfaTOTPPage.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { useWizard, Wizard } from '../common';
+import { withCardStateProvider } from '../elements';
+import { AddAuthenticatorApp } from './AddAuthenticatorApp';
+import { SuccessPage } from './SuccessPage';
+import { VerifyTOTP } from './VerifyTOTP';
+
+export const MfaTOTPPage = withCardStateProvider(() => {
+  const title = 'Add authenticator app';
+  const wizard = useWizard({});
```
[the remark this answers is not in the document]
yourtallness: Removing

packages/clerk-js/src/v4/icons/index.ts
yourtallness: Replacing with `AuthApp`

packages/clerk-js/package.json
yourtallness: Will handle in a subsequent PR

Just re-used the same `copy-to-clipboard` dep Chakra uses in the dashboard.

PR #343, opened 2022-08-24
fix(clerk-js,types,clerk-sdk-node,clerk-react,nextjs,remix,edge,clerk…

nikosdouvlis:
…-expo,backend-core): Fix dependencies

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/packages/clerk-js/src/ui/elements/TimerButton.tsx
+++ b/packages/clerk-js/src/ui/elements/TimerButton.tsx
@@ -27,7 +27,7 @@ export const TimerButton = (props: TimerButtonProps) => {
 
   const disable = () => {
     setRemainingSeconds(throttleTimeInSec);
-    intervalIdRef.current = setInterval(() => {
+    intervalIdRef.current = window.setInterval(() => {
       setRemainingSeconds(seconds => {
         if (seconds === 1) {
           clearInterval(intervalIdRef.current);
--- a/packages/types/src/resource.ts
+++ b/packages/types/src/resource.ts
@@ -3,7 +3,7 @@ export type ClerkResourceReloadParams = {
 };
 
 export interface ClerkResource {
-  readonly id?: string;
+  readonly id?: string | undefined;
   pathRoot: string;
   reload(p?: ClerkResourceReloadParams): Promise<this>;
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -23,10 +23,7 @@
     "pretty": true,
     "sourceMap": true,
     "strict": true,
-    "types": ["node"],
-    "paths": {
-      "@clerk/*": ["*/src"]
-    },
+    //    "types": ["node"],
     "target": "ES2019",
     "noErrorTruncation": true
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,8 +12,6 @@
         "@commitlint/cli": "^16.0.2",
         "@commitlint/config-conventional": "^16.0.0",
         "@commitlint/config-lerna-scopes": "^16.0.0",
-        "@types/node": "14.14.33",
-        "@types/web": "^0.0.55",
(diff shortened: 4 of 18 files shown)


tsconfig.json
```diff
@@ -23,10 +23,7 @@
     "pretty": true,
     "sourceMap": true,
     "strict": true,
-    "types": ["node"],
-    "paths": {
-      "@clerk/*": ["*/src"]
-    },
+    //    "types": ["node"],
```
gkats, line 26: :upside_down_face: Let's remove altogether?

packages/types/src/resource.ts
```diff
@@ -3,7 +3,7 @@ export type ClerkResourceReloadParams = {
 };
 
 export interface ClerkResource {
-  readonly id?: string;
+  readonly id?: string | undefined;
```
gkats, line 6: :upside_down_face: If the `undefined` part was added to accomodate the `UserSettingsResource`, I would expect that simply making it optional would suffice :/ But I guess you checked it and the linter complains somewhere?

packages/clerk-js/src/ui/elements/TimerButton.tsx
```diff
@@ -27,7 +27,7 @@ export const TimerButton = (props: TimerButtonProps) => {
 
   const disable = () => {
     setRemainingSeconds(throttleTimeInSec);
-    intervalIdRef.current = setInterval(() => {
+    intervalIdRef.current = window.setInterval(() => {
```
gkats, line 30: :question: I assume simply accessing `window` is safe in every context.

gkats (approved):

SokratisVidros (approved):
💯

SokratisVidros:
@nikosdouvlis Please add to the description all the Github issues that this PR fixes.

packages/clerk-js/src/ui/elements/TimerButton.tsx
nikosdouvlis: Hey @gkats , thanks for checking!

It should be safe in this case, since the `disable` function is invoked within the `useEffect` callback (client-side only) or from a user interaction (clicking the button). In both cases, `window` should always exist.

packages/types/src/resource.ts
nikosdouvlis: Exactly! 

This fixes an issue where a Clerk package that has a direct dependency on `@clerk/types` (like `backend-core`) is installed within a TS application that includes the following options in its `tsconfig.json` file:

```
{
  "compilerOptions": {
    ...
    "exactOptionalPropertyTypes": true,
    "skipLibCheck": false,
  },
}
```

Having `skipLibCheck` set to false would make `tsc` type check the app and all the imported libs, and without the explicit `| undefined;` in our type, the `exactOptionalPropertyTypes` would cause an error.

More info: https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes

tsconfig.json
nikosdouvlis: ✅

packages/types/src/resource.ts
gkats: Thanks for the explanation!

PR #346, opened 2022-08-25
Organization membership metadata

gkats:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend-core`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

We've recently added public and private metadata fields for organization memberships. Both metadata fields can be  accessed from the Backend API, but public metadata can only be read from the Frontend API.

This PR adds support for the organization metadata fields by:
- Adding `publicMetadata` in the `OrganizationMembershipResource` type.
- Adding `publicMetadata` in the `OrganizationMembership` resource in clerkJS.
- Adding support for updating an organization membership's public and private metadata in backend-core and exposing it to our Node SDK.

The new backend-core method for the organizations API is `organizations.updateOrganizationMembershipMetadata()`.

### DIFF
--- a/packages/backend-core/src/api/endpoints/OrganizationApi.ts
+++ b/packages/backend-core/src/api/endpoints/OrganizationApi.ts
@@ -5,7 +5,7 @@ import { AbstractAPI } from './AbstractApi';
 
 const basePath = '/organizations';
 
-type OrganizationMetadataParams = {
+type MetadataParams = {
   publicMetadata?: Record<string, unknown>;
   privateMetadata?: Record<string, unknown>;
 };
@@ -20,7 +20,7 @@ type CreateParams = {
   slug?: string;
   /* The User id for the user creating the organization. The user will become an administrator for the organization. */
   createdBy: string;
-} & OrganizationMetadataParams;
+} & MetadataParams;
 
 type GetOrganizationParams = { organizationId: string } | { slug: string };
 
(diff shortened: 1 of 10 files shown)


packages/backend-core/src/api/endpoints/OrganizationApi.ts
```diff
@@ -160,6 +165,19 @@ export class OrganizationAPI extends AbstractAPI {
     });
   }
 
+  public async updateOrganizationMembershipMetadata(params: UpdateOrganizationMembershipMetadataParams) {
+    const { organizationId, userId, publicMetadata, privateMetadata } = params;
```
nikosdouvlis, line 169: ❓Just asking, If this method is called without providing one of the `publicMetadata` or `privateMetadata` props, eg when a user only wants to update on the of two, the prop not provided will end up as `undefined` here.

Can this lead to unintentionally deleting the corresponding metadata value on the BE?

nikosdouvlis (approved):
Had a question, the rest look good!

packages/backend-core/src/api/endpoints/OrganizationApi.ts
gkats: @nikosdouvlis Thanks for raising the concern. The BE API will omit undefined values (as if they are not passed at all). It will work as expected.

PR #357, opened 2022-09-02
feat(nextjs,edge,backend-core): Expose clerkApi, getAuth & runClerkMi…

yourtallness:
…ddleware in @clerk/nextjs/server

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [X] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [X] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [X] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Draft PR combining @igneel64 draft & @colinclerk's POC

# Related issue(s)

https://github.com/clerkinc/javascript/issues/299

Aims:

- support for next.js 12.2+ middlewares
- challenge: request is now immutable, req.auth injection is not an option any more
- challenge: middleware can no longer render a response (e.g. the intestitial), req needs to be rewritten to interstitial endpoint
- new approach for getting auth data: run middleware first, call getAuth in API route to consume auth data (no more function wrappers or withAuth going forward)
- additive only, don't remove any existing functionality

### DIFF
--- a/packages/backend-core/src/types/core.ts
+++ b/packages/backend-core/src/types/core.ts
@@ -30,7 +30,7 @@ export type TokenType = 'cookie' | 'header';
 export type AuthState = {
   status: AuthStatus;
   session?: Session;
-  /* Interstitial is returned as null when the client will handle the interstitial logic */
+  /* Interstitial is returned as null when the interstitial endpoint will be rewritten to instead of being rendered directly */
   interstitial?: string | null;
   sessionClaims?: ClerkJWTClaims;
   /* Error reason for signed-out and interstitial states. Would probably be set on the `Auth-Result` response header. */
--- a/packages/backend-core/src/util/createGetToken.ts
+++ b/packages/backend-core/src/util/createGetToken.ts
@@ -31,7 +31,7 @@ export const createGetToken: CreateGetToken = params => {
   };
 };
 
-const signedOutGetToken = createGetToken({
+export const signedOutGetToken = createGetToken({
   sessionId: undefined,
   cookieToken: undefined,
   headerToken: undefined,
--- a/packages/nextjs/src/server/types.ts
+++ b/packages/nextjs/src/server/types.ts
@@ -1,4 +1,4 @@
-import { AuthErrorReason, ClerkBackendAPI, Organization, Session, User } from '@clerk/backend-core';
+import { AuthErrorReason, ClerkBackendAPI } from '@clerk/backend-core';
 import { ClerkJWTClaims, ServerGetToken } from '@clerk/types';
 import { IncomingMessage } from 'http';
 import { NextApiRequest } from 'next';
@@ -18,20 +18,16 @@ export const NEXT_REWRITE_HEADER = 'x-middleware-rewrite';
 export const NEXT_RESUME_HEADER = 'x-middleware-next';
 export const NEXT_REDIRECT_HEADER = 'Location';
 
-// TODO consolidate AuthData declarations across repo
 export type AuthData = {
   sessionId: string | null;
-  session: Session | undefined | null;
   userId: string | null;
-  user: User | undefined | null;
   orgId: string | null | undefined;
-  organization: Organization | undefined | null;
   getToken: ServerGetToken;
   claims: ClerkJWTClaims | null;
 };
 
 enum AuthResultExt {
-  StandardIn = 'standard-in',
+  StandardSignedIn = 'standard-signed-in',
 }
 
 export const AuthResult = { ...AuthResultExt, ...AuthErrorReason };
--- a/packages/nextjs/src/server/utils/getAuth.ts
+++ b/packages/nextjs/src/server/utils/getAuth.ts
@@ -1,4 +1,4 @@
-import { createSignedOutState } from '@clerk/backend-core';
+import { signedOutGetToken } from '@clerk/backend-core';
 
 import type { RequestLike } from '../types';
 import { AuthData, AuthResult, SESSION_COOKIE_NAME, SessionsApi } from '../types';
@@ -14,11 +14,11 @@ export function createGetAuth(sessions: SessionsApi) {
     const authResult = getAuthResultFromRequest(req);
 
     if (!authResult) {
-      throw 'You need to use "withClerkMiddleware" in your Next.js middleware.js file. See https://nextjs.org/docs/advanced-features/middleware.';
+      throw 'You need to use "withClerkMiddleware" in your Next.js middleware.js file. See https://clerk.dev/docs/quickstarts/get-started-with-nextjs.';
     }
 
     // Signed in case
-    if (authResult === AuthResult.StandardIn) {
+    if (authResult === AuthResult.StandardSignedIn) {
       // Get the token from header or cookie
 
       const headerToken = getHeader(req, 'authorization')?.replace('Bearer ', '');
@@ -37,6 +37,12 @@ export function createGetAuth(sessions: SessionsApi) {
     }
 
     // Signed out case assumed
-    return createSignedOutState();
+    return {
+      sessionId: null,
+      userId: null,
+      orgId: null,
+      getToken: signedOutGetToken,
+      claims: null,
+    };
   };
 }
--- a/packages/nextjs/src/server/utils/requestResponseUtils.ts
+++ b/packages/nextjs/src/server/utils/requestResponseUtils.ts
@@ -24,15 +24,21 @@ function getQueryParam(req: RequestLike, name: string): string | null | undefine
 
   // Check if the request contains a parsed query object
   // NextApiRequest does, but the IncomingMessage in the GetServerSidePropsContext case does not
+
+  let queryParam: string | null | undefined;
+
   if ('query' in req) {
-    return req.query[name] as string | undefined;
+    queryParam = req.query[name] as string | undefined;
   }
 
   // Fall back to query string
-  const qs = (req.url || '').split('?')[1];
-  const searchParams = new URLSearchParams(qs);
+  if (!queryParam) {
+    const qs = (req.url || '').split('?')[1];
+    const searchParams = new URLSearchParams(qs);
+    queryParam = searchParams.get(name);
+  }
 
-  return searchParams.get(name);
+  return queryParam;
 }
 
 export function getHeader(req: RequestLike, name: string): string | null | undefined {
--- a/packages/nextjs/src/server/utils/withClerkMiddleware.ts
+++ b/packages/nextjs/src/server/utils/withClerkMiddleware.ts
@@ -17,26 +17,30 @@ const DEFAULT_API_URL = process.env.CLERK_API_URL || 'https://api.clerk.dev';
 const DEFAULT_API_VERSION = process.env.CLERK_API_VERSION || 'v1';
 
 // Using public interstitial endpoint for time being
-const INTERSTITIAL_URL = `${DEFAULT_API_URL}/${DEFAULT_API_VERSION}/public/interstitial?frontendApi=${process.env.NEXT_PUBLIC_FRONTEND_API}`;
+const INTERSTITIAL_URL = `${DEFAULT_API_URL}/${DEFAULT_API_VERSION}/public/interstitial?frontendApi=${process.env.NEXT_PUBLIC_CLERK_FRONTEND_API}`;
 
 type WithAuthOptions = {
   jwtKey?: string;
   authorizedParties?: string[];
+  ignoredPaths?: string[];
 };
 
 export function withClerkMiddleware(handler: NextMiddleware, opts: WithAuthOptions = {}): NextMiddleware {
   return async (req: NextRequest, event: NextFetchEvent) => {
     const { headers, cookies } = req;
-    const { jwtKey, authorizedParties } = opts;
+    const { jwtKey, authorizedParties, ignoredPaths } = opts;
+    const pathname = req.nextUrl.pathname;
 
-    // Skip if this is an internal next request
-    if (req.nextUrl.pathname.startsWith('/_next/')) {
+    // Skip if:
+    // - it is an internal next request
+    // - it is an ignored path
+    if (pathname.startsWith('/_next/') || (ignoredPaths || []).find(p => new RegExp(p).test(pathname))) {
       return NextResponse.next();
     }
 
     // throw an error if the request already includes an auth result, because that means it has been spoofed
     if (getAuthResultFromRequest(req)) {
-      throw 'Auth violation detected';
+      throw 'withClerkMiddleware: Auth violation detected';
     }
 
     // get auth state, check if we need to return an interstitial
@@ -68,10 +72,10 @@ export function withClerkMiddleware(handler: NextMiddleware, opts: WithAuthOptio
 
     // Signed in / out case
 
-    const authResult = status === AuthStatus.SignedIn ? AuthResult.StandardIn : errorReason;
+    const authResult = status === AuthStatus.SignedIn ? AuthResult.StandardSignedIn : errorReason;
 
     if (!authResult) {
-      throw 'Auth result could not be determined';
+      throw 'withClerkMiddleware: Auth result could not be determined';
     }
 
     setAuthResultOnRequest({ req, authResult });
--- a/packages/backend-core/src/Base.ts
+++ b/packages/backend-core/src/Base.ts
@@ -269,7 +269,7 @@ export class Base {
     // }
 
     if (clientUat === '0') {
-      return { status: AuthStatus.SignedOut, errorReason: AuthErrorReason.StandardOut };
+      return { status: AuthStatus.SignedOut, errorReason: AuthErrorReason.StandardSignedOut };
     }
 
     if (isProductionKey && clientUat && !cookieToken) {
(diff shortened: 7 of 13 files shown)


packages/nextjs/src/server/types.ts
```diff
@@ -0,0 +1,8 @@
+import { AuthStatus, ClerkBackendAPI } from '@clerk/backend-core';
+
+export const AuthStatusMapping = {
+  [AuthStatus.SignedIn]: 'i',
```
nikosdouvlis, line 4: 🙃 I know that this is irrelevant, but maybe turn this into an 1 and 0 or even an actual enum starting from 1?

packages/nextjs/src/server/utils/getAuth.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    if (!clerkParam) {
+      throw 'You need to use "runClerkMiddleware" on your Next.js middleware file.';
+    }
+
+    // Signed out case
+    if (clerkParam === AuthStatusMapping[AuthStatus.SignedOut]) {
+      return {
+        // TODO what to return here?
+        sessionClaims: null,
```
nikosdouvlis, line 23: Addressing the TODO here, do we simply want to return the signed out state as in: https://github.com/clerkinc/javascript/blob/0889bde9bc7f9e1a5d4c1e706c49212e1f7b36f4/packages/backend-core/src/util/createGetToken.ts#L41

or am I missing something?

packages/nextjs/src/server/utils/getAuth.ts
```diff
@@ -0,0 +1,50 @@
...
+      return {
+        ...authData,
+        getToken: createGetToken({
+          headerToken: authData.headerToken,
+          cookieToken: authData.cookieToken,
+          sessionId: authData.sessionId,
+          fetcher: (...args) => sessionsApi.getToken(...args),
+        }),
+        headerToken: headerToken, // TODO return this?
+        cookieToken: cookieToken, // TODO return this?
```
nikosdouvlis, line 46: I think we shouldn't increase the public API surface unless a customer explicitly asks for these two - we can always expose them going forwards

packages/nextjs/src/server/utils/getAuth.ts
yourtallness: You're probably right

packages/nextjs/src/server/utils/getAuth.ts
yourtallness: @nikosdouvlis These fields already exist, the question is whether to populate them

packages/nextjs/server.js
```diff
@@ -0,0 +1,16 @@
+// TODO add instance / constructor export for clerkApi
+
+let exportLib = {};
+
+if (process.env.NEXT_RUNTIME === 'edge') {
```
SokratisVidros, line 5: ❓ Is this variable available on Vercel Paas?

packages/nextjs/src/server/types.ts
SokratisVidros: Let's add something more descriptive and less hacky.

I've suggested using the actual JWT claims as an encoded string at it would help us and customers to debug easily.

We will also avoid doing [this](https://github.com/clerkinc/javascript/pull/357/files#diff-e0202f43b73d4d25fa9b0e4e88d07dbbc974c8c4381f1c8a2663bc3288679addR35) twice.

packages/nextjs/src/server/utils/getAuth.ts
```diff
@@ -0,0 +1,50 @@
...
+export function createGetAuth(sessionsApi: SessionsApi) {
+  return function (req: NextRequest) {
+    // When the __clerk query param is set, we trust that middleware already ran
+    // Then, we don't have to run any re-verify the JWT here, we
+    // can just strip out the claims manually.
+
+    const clerkParam = req.nextUrl.searchParams.get('__clerk');
+
+    if (!clerkParam) {
+      throw 'You need to use "runClerkMiddleware" on your Next.js middleware file.';
```
SokratisVidros, line 16:
```suggestion
      throw 'You need to use "runClerkMiddleware" on your Next.js middleware.js file.';
```

Let's also add a link to Next.js middleware docs.

packages/nextjs/src/server/utils/runClerkMiddleware.ts
```diff
@@ -0,0 +1,141 @@
...
+const DEFAULT_API_VERSION = process.env.CLERK_API_VERSION || 'v1';
+
+const INTERSTITIAL_URL = `${DEFAULT_API_URL}/${DEFAULT_API_VERSION}/internal/interstitial`;
+
+type RunClerkMiddlewareOptions = {
+  jwtKey?: string;
+  authorizedParties?: string[];
+};
+
+// TODO Extend NextResponse instead?
```
SokratisVidros, line 19: +1 on the extension

packages/nextjs/src/server/utils/runClerkMiddleware.ts
```diff
@@ -0,0 +1,141 @@
...
+  redirect: () => NextResponse;
+};
+
+// TODO middleware will need to discern error (interstitial case) from signed out case
+export type RunClerkMiddlewareReturn = {
+  auth?: AuthData; // TODO need a more limited type without user & organization, but possibly with isSignedIn?
+  response?: ClerkNextResponse;
+};
+
+export function createRunClerkMiddleware(base: Base, sessionsApi: SessionsApi) {
```
SokratisVidros, line 33: 🙃 Consider using an object here as more options might come soon

packages/nextjs/server.js
colinclerk: Yes, it should be set on any PaaS running Next.js:
https://nextjs.org/docs/advanced-features/react-18/switchable-runtime#global-runtime-option

From past research, we believe it's available in both getServerSideProps and API routes, and regardless of whether the runtime is set globally or on a per-API route basis.

SokratisVidros: Thanks for checking this.

packages/nextjs/src/server/types.ts
yourtallness: This no longer applies, so resolving.

packages/nextjs/src/server/utils/runClerkMiddleware.ts
yourtallness: This no longer applies, resolving.

packages/nextjs/src/server/utils/runClerkMiddleware.ts
yourtallness: Not applicable any more after latest changes

packages/nextjs/src/server/utils/getAuth.ts
yourtallness: Added

packages/nextjs/src/server/utils/withClerkMiddleware.ts
```diff
@@ -0,0 +1,132 @@
...
+  NEXT_REWRITE_HEADER,
+  SESSION_COOKIE_NAME,
+} from '../types';
+import { getAuthResultFromRequest, setAuthResultOnRequest } from './requestResponseUtils';
+
+const DEFAULT_API_URL = process.env.CLERK_API_URL || 'https://api.clerk.dev';
+const DEFAULT_API_VERSION = process.env.CLERK_API_VERSION || 'v1';
+
+// Using public interstitial endpoint for time being
+const INTERSTITIAL_URL = `${DEFAULT_API_URL}/${DEFAULT_API_VERSION}/public/interstitial?frontendApi=${process.env.NEXT_PUBLIC_FRONTEND_API}`;
```
nikosdouvlis, line 20: ❓ Should this be `NEXT_PUBLIC_CLERK_FRONTEND_API` ?

yourtallness: Fixed

packages/backend-core/src/types/core.ts
```diff
@@ -30,7 +30,8 @@ export type TokenType = 'cookie' | 'header';
 export type AuthState = {
   status: AuthStatus;
   session?: Session;
-  interstitial?: string;
+  /* Interstitial is returned as null when the client will handle the interstitial logic */
```
SokratisVidros, line 33: ❓ What does `the client` refer-to? Are we talking about npm packages using `@clerk/backend-core`?

packages/backend-core/src/util/createGetToken.ts
```diff
@@ -43,6 +43,7 @@ export const createSignedOutState = () => {
     sessionId: null,
     session: null,
     userId: null,
+    orgId: null,
```
SokratisVidros, line 46: 👍

packages/nextjs/src/server/types.ts
```diff
@@ -0,0 +1,40 @@
...
+export const NEXT_REWRITE_HEADER = 'x-middleware-rewrite';
+export const NEXT_RESUME_HEADER = 'x-middleware-next';
+export const NEXT_REDIRECT_HEADER = 'Location';
+
+// TODO consolidate AuthData declarations across repo
+export type AuthData = {
+  sessionId: string | null;
+  session: Session | undefined | null;
+  userId: string | null;
+  user: User | undefined | null;
```
SokratisVidros, line 26: ❓ Since we don't load the actual resources in the context of the middleware, which is the scenario that we get a `session`, a `user`, or an `organization` object in the `AuthData`?

packages/nextjs/src/server/types.ts
```diff
@@ -0,0 +1,40 @@
...
+  orgId: string | null | undefined;
+  organization: Organization | undefined | null;
+  getToken: ServerGetToken;
+  claims: ClerkJWTClaims | null;
+};
+
+enum AuthResultExt {
+  StandardIn = 'standard-in',
```
SokratisVidros, line 34: 🔧 @yourtallness Let's proceed with your proposal to make this more human readable.

packages/nextjs/src/server/utils/getAuth.ts
```diff
@@ -0,0 +1,42 @@
...
+export function createGetAuth(sessions: SessionsApi) {
+  return function getAuth(req: RequestLike): AuthData {
+    // When the auth result is set, we trust that the middleware has already run
+    // Then, we don't have to re-verify the JWT here,
+    // we can just strip out the claims manually.
+
+    const authResult = getAuthResultFromRequest(req);
+
+    if (!authResult) {
+      throw 'You need to use "withClerkMiddleware" in your Next.js middleware.js file. See https://nextjs.org/docs/advanced-features/middleware.';
```
SokratisVidros, line 17: 🔧 We should update this string to point to our Next.js docs.

packages/nextjs/src/server/utils/requestResponseUtils.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+function getQueryParam(req: RequestLike, name: string): string | null | undefined {
+  if (req instanceof NextRequest) {
+    return req.nextUrl.searchParams.get(name);
+  }
+
+  // Check if the request contains a parsed query object
+  // NextApiRequest does, but the IncomingMessage in the GetServerSidePropsContext case does not
+  if ('query' in req) {
```
SokratisVidros, line 27: 🔧 This clause should check if `req.query[name]` is present. If it's not, shouldn't we try to fallback to `req.url` instead of returning `undefined`?

packages/nextjs/src/server/utils/withClerkMiddleware.ts
```diff
@@ -0,0 +1,132 @@
...
+  authorizedParties?: string[];
+};
+
+export function withClerkMiddleware(handler: NextMiddleware, opts: WithAuthOptions = {}): NextMiddleware {
+  return async (req: NextRequest, event: NextFetchEvent) => {
+    const { headers, cookies } = req;
+    const { jwtKey, authorizedParties } = opts;
+
+    // Skip if this is an internal next request
+    if (req.nextUrl.pathname.startsWith('/_next/')) {
```
SokratisVidros, line 33: 💡 It would be great to have an array of path allowlists in `WithAuthOptions` so that if we miss a `Next.js` internal path in the current or future version of the framework, users can still use this library until we patch it.

This is a very common pattern in middleware. We did the same in Ruby last week.

For example:

```ts
type WithAuthOptions = {
  jwtKey?: string;
  authorizedParties?: string[];
  ingoredPaths?: string[];
};

...

if ((req.nextUrl.pathname.startsWith('/_next/') || ingoredPaths.find(p => new Regexp(p).test(req.nextUrl.pathname))) {
  return NextResponse.next()
}
```

packages/nextjs/src/server/utils/withClerkMiddleware.ts
```diff
@@ -0,0 +1,132 @@
...
+      return NextResponse.next();
+    }
+
+    // throw an error if the request already includes an auth result, because that means it has been spoofed
+    if (getAuthResultFromRequest(req)) {
+      throw 'Auth violation detected';
```
SokratisVidros, line 39: 🙃 

```suggestion
      throw 'withClerkMiddleware: Auth violation detected';
```

packages/nextjs/src/server/utils/withClerkMiddleware.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (status === AuthStatus.Interstitial) {
+      return NextResponse.rewrite(INTERSTITIAL_URL);
+    }
+
+    // Signed in / out case
+
+    const authResult = status === AuthStatus.SignedIn ? AuthResult.StandardIn : errorReason;
+
+    if (!authResult) {
+      throw 'Auth result could not be determined';
```
SokratisVidros, line 74: 🙃 
```suggestion
      throw 'withClerkMiddleware: Auth result could not be determined';
```

packages/backend-core/src/types/core.ts
yourtallness: This refers to the backend client code, will clarify.

yourtallness: Actually this was taken from Peter's PR and no longer applies.

yourtallness: Changing to:

`/* Interstitial is returned as null when the interstitial endpoint will be rewritten to instead of being rendered directly */`

packages/nextjs/src/server/types.ts
yourtallness: We don't but it's confusing to have so many different AuthData declarations so I'm using the superset here.

Otherwise we'd have to generate different signed out data for this case as well.

I don't think the complexity is worth it.

packages/nextjs/src/server/types.ts
yourtallness: Renaming to `standard-out` & `standard-out`

packages/nextjs/src/server/utils/getAuth.ts
yourtallness: Done

packages/nextjs/src/server/utils/withClerkMiddleware.ts
yourtallness: Done

packages/nextjs/src/server/utils/withClerkMiddleware.ts
yourtallness: ✅

packages/nextjs/src/server/types.ts
SokratisVidros: `Standard-signed-in`?

packages/nextjs/src/server/types.ts
SokratisVidros: I am skeptical whether this will create expectations to the end users about getting the full objects. Maybe we can just Pick the rest.

packages/nextjs/src/server/utils/requestResponseUtils.ts
yourtallness: The way `next` currently works the `query` is always up-to-date, when it exists.

The `req.url` case is mainly for the GSSP case where no `req.query` exists.

packages/nextjs/src/server/types.ts
yourtallness: Yes, that 😅

packages/nextjs/src/server/types.ts
yourtallness: It would be nice if we had AR-like `association.loaded?` semantics.

packages/nextjs/src/server/utils/requestResponseUtils.ts
yourtallness: That being said, I'll add the fallback there as well.

packages/nextjs/src/server/utils/withClerkMiddleware.ts
yourtallness: Adding

packages/nextjs/src/server/types.ts
SokratisVidros: Isn't that a bit over-engineering given that we have the undefined type?

SokratisVidros (approved):

packages/nextjs/src/server/types.ts
yourtallness: Maybe if we had the following notion:
- `null` = known not to exist
- `undefined` = not populated, you'll need to load it

yourtallness: Anyway, narrowed down to only the fields we are setting, for now.

PR #387, opened 2022-09-22
feat(types,clerk-js): Introduce BackupCodeResource and user.createBackupCode() oepration

chanioxaris:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

- Introduce the new BackupCode resource and the relevant types
- Introduce a new `user.createBackupCode()` method, that generates a set for backup codes for a user

https://www.notion.so/clerkdev/Introduce-new-FAPI-me-backup-codes-endpoints-31f0832bb7544d4d90ba27758cac2a54

### DIFF
--- a/packages/clerk-js/src/core/resources/User.ts
+++ b/packages/clerk-js/src/core/resources/User.ts
@@ -1,4 +1,6 @@
 import type {
+  BackupCodeJSON,
+  BackupCodeResource,
   CreateEmailAddressParams,
   CreatePhoneNumberParams,
   CreateWeb3WalletParams,
@@ -23,6 +25,7 @@ import type {
 
 import { unixEpochToDate } from '../../utils/date';
 import { normalizeUnsafeMetadata } from '../../utils/resourceParams';
+import { BackupCode } from './BackupCode';
 import {
   BaseResource,
   DeletedObject,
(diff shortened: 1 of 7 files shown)


packages/clerk-js/src/core/resources/User.ts
```diff
@@ -177,6 +180,17 @@ export class User extends BaseResource implements UserResource {
     return new DeletedObject(json);
   };
 
+  createBackupCode = async (): Promise<BackupCodeResource> => {
+    const json = (
+      await BaseResource._fetch<BackupCodeJSON>({
+        path: this.path() + '/backup_codes/',
+        method: 'POST',
+      })
+    )?.response as unknown as BackupCodeJSON;
```
nikosdouvlis, line 189: This is not an issue with this specific PR, but the type casting/ checking for the `verifyTOTP`, `disableTOTP` and `createBackupCode` is definitely brittle and since it involves network requests, the actual runtime value might be different that what we expect at this point

I did a very quick pass and I think that instead of the manual type casts,  I'd rather use an assertion function to at least throw/ try to recover at that point. What do you think?

This can definitely be done in a separate PR as its not related with the changes made by Haris :)

cc @SokratisVidros

nikosdouvlis (approved):
Everything looks good!

packages/clerk-js/src/core/resources/User.ts
SokratisVidros: +1

PR #388, opened 2022-09-23
feat(clerk-js): Add impersonation banner if session.actor is defined

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Display a non-removable floating banner if the user is being impersonated.

https://user-images.githubusercontent.com/73396808/191917793-e5cecddb-028c-445d-822d-0c2e45e8b057.mov


packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <Text
+          colorScheme='neutral'
+          variant='regularMedium'
+        >
+          Signed in as {id}
+        </Text>
+        <Link
+          variant='regularMedium'
+          sx={t => ({
+            alignSelf: 'flex-start',
```
nikosdouvlis, line 66: ❓  Why do we need this? I'd expect that this Link will be aligned to the left by default

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,138 @@
...
+            color: t.colors.$primary500,
+            ':hover': {
+              cursor: 'pointer',
+            },
+          })}
+          onClick={async () => {
+            await signOut();
+            close();
```
nikosdouvlis, line 74: ❓ I'm actually a [little surprised that this is working](https://developer.mozilla.org/en-US/docs/Web/API/Window/close). Can you please provide a little more context?

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,138 @@
...
+  );
+};
+
+export const ImpersonationFab: React.FC = () => {
+  const session = useCoreSession();
+  const { parsedInternalTheme } = useAppearance();
+  const actor = session?.actor;
+
+  //essentials for calcs
+  const eyePadding = '20px';
```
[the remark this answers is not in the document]
desiprisg: I needed them in this component to do the correct calculations for responsiveness in mobile.

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
desiprisg: Think it works just like document (which is the same as window.document). Will change to be explicit though.

packages/clerk-js/src/ui/foundations/transitions.ts
```diff
@@ -1,4 +1,5 @@
 const transitionDuration = Object.freeze({
+  slowest: '400ms',
```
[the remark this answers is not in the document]
desiprisg: Sure, it is quite fast though.

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const eyeWidth = '24px';
+  const eyeHeight = '24px';
+  const textPaddingLeft = parsedInternalTheme.sizes.$5;
+  const textPaddingRight = parsedInternalTheme.sizes.$8;
+
+  return (
+    <>
+      {actor && (
```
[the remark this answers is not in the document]
desiprisg: Missed a bang but agree!

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,138 @@
...
+            sx={t => ({
+              position: 'fixed',
+              height: `calc(${eyeWidth} + 2 * ${eyePadding})`,
+              width: 'auto',
+              overflow: 'hidden',
+              top: '109px',
+              right: '23px',
+              zIndex: 999999,
+              boxShadow: t.shadows.$boxShadow2,
+              borderRadius: '32px',
```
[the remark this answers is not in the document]
desiprisg: Should we use the modal zIndex here?

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,138 @@
...
+type EyeCircleProps = {
+  width: string;
+  height: string;
+  padding: string;
+};
+
+const EyeCircle: React.FC<EyeCircleProps> = ({ width, height, padding }) => {
+  return (
+    <Flex
+      id='impersonationEye'
```
SokratisVidros, line 17: ❓ Is this ID necessary for the appearance object? If not do we really need it?

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,138 @@
...
+type FabContentProps = { id: string; paddingLeft: string; paddingRight: string };
+
+const FabContent: React.FC<FabContentProps> = ({ id, paddingLeft, paddingRight }) => {
+  const { signOut } = useCoreClerk();
+
+  return (
+    <Col
+      id='impersonationText'
+      sx={{
+        whiteSpace: 'nowrap',
```
SokratisVidros, line 46: 🔧 I'd also suggest to add `text-overflow: ellipsis;` and maybe `overflow: hidden;` so that the UI doesn't break when rendering the dynamic user identifier

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,138 @@
...
+        sx={{
+          paddingLeft,
+          paddingRight,
+        }}
+      >
+        <Text
+          colorScheme='neutral'
+          variant='regularMedium'
+        >
+          Signed in as {id}
```
SokratisVidros, line 61: 🔧 Rendering the user id is not very user-friendly. We should render the user name or identifier following the same algorithm we have in the UserButton https://github.com/clerkinc/javascript/blob/2bbb13022e4a9d3cbd62f06ad2b06b229f05583e/packages/clerk-js/src/ui/UserButton/UserButton.tsx#L49)

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,138 @@
...
+          <Flex
+            align='center'
+            sx={t => ({
+              position: 'fixed',
+              height: `calc(${eyeWidth} + 2 * ${eyePadding})`,
+              width: 'auto',
+              overflow: 'hidden',
+              top: '109px',
+              right: '23px',
+              zIndex: 999999,
```
SokratisVidros, line 109: https://www.geeksforgeeks.org/what-is-maximum-minimum-value-for-z-index-property-in-css/#:~:text=The%20maximum%20and%20minimum%20value,from%20%E2%88%922147483648%20to%20%2B2147483647.

SokratisVidros (changes requested):
Very nicely done @desiprisg

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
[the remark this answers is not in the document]
nikosdouvlis: You're right, fixed ✔️

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
nikosdouvlis: Hey George, sorry I wasn't clear enough. As far as I know, calling `close()` that way would result in the following console warning and at the same time, that tab *will* remain open
```
Scripts may close only the windows that were opened by them.
```

You can find more details here: https://developer.mozilla.org/en-US/docs/Web/API/Window/close

desiprisg: You're right. This happens when the tab was not opened by a script. We can add a redirect in that case.

nikosdouvlis: The tab can only be closed if it was opened with `window.open()` by the caller tab, so I'm pretty sure this is not going to work for our use case.

We could just sign the user out  - the host app should already have a 
```
<SignedOut><RedirectToSignIn></SignedOut>
``` 
(or similar) mechanism in place handling signed-out users.

Wdyt @SokratisVidros ?

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
desiprisg: ✔️ Fixed

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
desiprisg: Done ✔️

nikosdouvlis:
Hey @desiprisg, as discussed privately on Slack, here's another sample implementation of the expanding FAB, with ellipsis working. It should also be responsive by default, without the need to explicitly measure `ch` units.

Here's the [link to the sandbox](https://stackblitz.com/edit/nextjs-6sluyp?file=pages/index.js).

Let me know if this helps. Keep in mind that some of the styles shown in the above example will become simpler by using the corresponding components (`<Flex/>` etc).

desiprisg:
@nikosdouvlis I did try this sometime, but I thought we wanted a max width on desktop and the ch units are good for this, but now that I think about it, it makes more sense to let it grow if the email/user id is that big. Only drawback is the animation will feel faster the more text there is. We could also adjust this though.

nikosdouvlis:
> @nikosdouvlis I did try this sometime, but I thought we wanted a max width on desktop and the ch units are good for this, but now that I think about it, it makes more sense to let it grow if the email/user id is that big.

Valid point about the possibility of a huge email address - let's restrict the width by setting `max-width` to something reasonable like `50vw` on larger screens while letting it extend all the way on mobiles.

> Only drawback is the animation will feel faster the more text there is. We could also adjust this though.
Love the attention to detail George! Its fine, don't worry about it 🙂

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
desiprisg: Need it for the eye animation.

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
desiprisg: ✔️ Added a new theme z-index for these kind of floating banners.

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const actor = session?.actor;
+  const isImpersonating = !!actor;
+
+  //essentials for calcs
+  const top = '109px';
+  const right = '23px';
+  const eyeWidth = parsedInternalTheme.sizes.$16;
+  const eyeHeight = eyeWidth;
+
+  if (!isImpersonating || !session.user) {
```
SokratisVidros, line 106: ❓ Can you please elaborate why do we need the `!session.user` check?

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,158 @@
...
+        <Flex
+          id='cl-impersonationText'
+          sx={t => ({
+            maxWidth: '0px',
+            transition: `max-width ${t.transitionDuration.$slowest} ease`,
+          })}
+        >
+          <FabContent
+            title={localizationKeys('impersonationFab.title', {
+              userEmailOrId: getFullName(session.user) || getIdentifier(session.user),
```
SokratisVidros, line 150: 🔧 Should we call this localization variable just identifier?

packages/clerk-js/src/ui/foundations/zIndices.ts
```diff
@@ -1,4 +1,5 @@
 export const zIndices = Object.freeze({
   navbar: '100',
+  fab: '9000',
```
SokratisVidros, line 3: 💯

SokratisVidros (approved):
👁️

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
desiprisg: We need to access the current user for the identifier. Made this check so typescript does not complain, even though we know it will be set.

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,158 @@
...
+        height,
+        transition: `transform ${t.transitionDuration.$slowest} ease`,
+        backgroundColor: t.colors.$danger500,
+        borderRadius: t.radii.$circle,
+      })}
+    >
+      <Icon
+        elementDescriptor={descriptors.impersonationFabIcon}
+        sx={{
+          margin: 'auto',
```
nikosdouvlis, line 42: ❓  Can we avoid using `margin` by adding the `center` prop on `Col` above?

That way we can also avoid adding the explicit `justify='center'` prop

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,158 @@
...
+        }}
+        icon={Eye}
+        size={undefined}
```
nikosdouvlis, line 45: ❓ Can you please elaborate?

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,158 @@
...
+        elementDescriptor={descriptors.impersonationFabActionLink}
+        sx={t => ({
+          alignSelf: 'flex-start',
+          color: t.colors.$primary500,
+          ':hover': {
+            cursor: 'pointer',
+          },
+        })}
+        localizationKey={signOutText}
+        onClick={async () => {
```
nikosdouvlis, line 86:
```suggestion
        onClick={signOut}
```

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -0,0 +1,158 @@
...
+          top,
+          right,
+          zIndex: t.zIndices.$fab,
+          boxShadow: t.shadows.$fabShadow,
+          borderRadius: t.radii.$halfHeight, //to match the circular eye perfectly
+          backgroundColor: t.colors.$white,
+          fontFamily: t.fonts.$main,
+          ':hover #cl-impersonationText': {
+            maxWidth: `calc(50vw - ${eyeWidth} - 2 * ${right})`,
+            [mqu.md]: {
```
nikosdouvlis, line 127: Did you have any luck simplifying these calculations using the example I shared?

packages/clerk-js/src/ui/customizables/elementDescriptors.ts
```diff
@@ -169,6 +169,12 @@ export const APPEARANCE_KEYS = containsAllElementsConfigKeys([
 
   'activeDeviceIcon',
 
+  'impersonationFab',
```
nikosdouvlis, line 172: ✅ nice

packages/clerk-js/src/ui/localization/defaultEnglishResource.ts
```diff
@@ -240,6 +240,10 @@ export const defaultResource: DeepRequired<LocalizationResource> = {
     action__signOutAll: 'Sign out of all accounts',
     action__addAccount: 'Add account',
   },
+  impersonationFab: {
+    title: 'Signed in as {{userEmailOrId}}',
```
nikosdouvlis, line 244:
```suggestion
    title: 'Signed in as {{identifier}}',
```

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
desiprisg: Hmm, this works but I'm getting type errors.

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
desiprisg: Did not want a default width/height because the svg was getting cropped. Its width is not equal to its height by design.

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
desiprisg: Good catch. Agree!

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
desiprisg: Pretty sure I'm doing the same thing now. Just set the max-width at 50vw for bigger screens so the animation is smoother. Is there something I missed in your example?

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
nikosdouvlis: @desiprisg Where did we land on this?

desiprisg: We just sign the user out.

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
desiprisg: So that only the text is clickable, and not the entirety of the parent's width.

nikosdouvlis (approved):
🚀

PR #394, opened 2022-09-26
feat(clerk-js): Add open prop in user button

raptisj:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Add an `defaultOpen` prop in `UserButton` component in order to have a defualt state.

```jsx 
     <UserButton defaultOpen /> 
}
```

### DIFF
--- a/packages/clerk-js/src/core/resources/DisplayConfig.ts
+++ b/packages/clerk-js/src/core/resources/DisplayConfig.ts
@@ -10,6 +10,7 @@ export class DisplayConfig extends BaseResource implements DisplayConfigResource
   afterSignOutUrl!: string;
   afterSignUpUrl!: string;
   afterSwitchSessionUrl!: string;
+  isOpen!: boolean;
   applicationName!: string;
   backendHost!: string;
   branded!: boolean;
@@ -47,6 +48,7 @@ export class DisplayConfig extends BaseResource implements DisplayConfigResource
     this.afterSignInUrl = data.after_sign_in_url;
     this.afterSignUpUrl = data.after_sign_up_url;
     this.afterSignOutUrl = data.after_sign_out_url;
+    this.isOpen = data.is_open;
     this.afterSignOutOneUrl = data.after_sign_out_one_url;
     this.afterSignOutAllUrl = data.after_sign_out_all_url;
     this.afterSwitchSessionUrl = data.after_switch_session_url;
--- a/packages/clerk-js/src/core/resources/DisplayConfig.ts
+++ b/packages/clerk-js/src/core/resources/DisplayConfig.ts
@@ -10,7 +10,6 @@ export class DisplayConfig extends BaseResource implements DisplayConfigResource
   afterSignOutUrl!: string;
   afterSignUpUrl!: string;
   afterSwitchSessionUrl!: string;
-  isOpen!: boolean;
   applicationName!: string;
   backendHost!: string;
   branded!: boolean;
@@ -48,7 +47,6 @@ export class DisplayConfig extends BaseResource implements DisplayConfigResource
     this.afterSignInUrl = data.after_sign_in_url;
     this.afterSignUpUrl = data.after_sign_up_url;
     this.afterSignOutUrl = data.after_sign_out_url;
-    this.isOpen = data.is_open;
     this.afterSignOutOneUrl = data.after_sign_out_one_url;
     this.afterSignOutAllUrl = data.after_sign_out_all_url;
     this.afterSwitchSessionUrl = data.after_switch_session_url;
--- a/packages/clerk-js/src/ui/UserButton/UserButton.tsx
+++ b/packages/clerk-js/src/ui/UserButton/UserButton.tsx
@@ -10,8 +10,9 @@ import { UserButtonPopover } from './UserButtonPopover';
 import { UserButtonTrigger } from './UserButtonTrigger';
 
 const _UserButton = () => {
+  const { isOpen: isButtonOpenFromOutside } = useUserButtonContext();
   const { floating, reference, styles, toggle, isOpen } = usePopover({
-    defaultOpen: false,
+    defaultOpen: isButtonOpenFromOutside,
     placement: 'bottom-end',
     offset: 8,
   });
@@ -20,7 +21,7 @@ const _UserButton = () => {
     <Flow.Root flow='userButton'>
       <Flex
         elementDescriptor={descriptors.userButtonBox}
-        isOpen={isOpen}
+        isOpen={isButtonOpenFromOutside || isOpen}
         align='center'
         gap={2}
       >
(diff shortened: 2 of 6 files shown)


packages/react/src/components/uiComponents.tsx
```diff
@@ -36,7 +36,7 @@ class Portal extends React.PureComponent<MountProps, {}> {
   private portalRef = React.createRef<HTMLDivElement>();
 
   componentDidUpdate(prevProps: Readonly<MountProps>) {
-    if (prevProps.props.appearance !== this.props.props.appearance) {
+    if (prevProps.props.appearance !== this.props.props.appearance || prevProps.props !== this.props.props) {
```
nikosdouvlis, line 39: ❓ Can this be simplified to this?
```suggestion
    if (prevProps.props !== this.props.props) {
```

packages/clerk-js/src/core/resources/DisplayConfig.ts
```diff
@@ -10,6 +10,7 @@ export class DisplayConfig extends BaseResource implements DisplayConfigResource
   afterSignOutUrl!: string;
   afterSignUpUrl!: string;
   afterSwitchSessionUrl!: string;
+  isOpen!: boolean;
```
nikosdouvlis, line 13: ❓ I'm unsure why we'd need `isOpen` on `DisplayConfig`. Can you please elaborate?

packages/clerk-js/src/ui/UserButton/UserButton.tsx
```diff
@@ -10,8 +10,9 @@ import { UserButtonPopover } from './UserButtonPopover';
 import { UserButtonTrigger } from './UserButtonTrigger';
 
 const _UserButton = () => {
+  const { isOpen: isButtonOpenFromOutside } = useUserButtonContext();
```
nikosdouvlis, line 13:
```suggestion
  const { isOpen: isOpenProp } = useUserButtonContext();
```

packages/clerk-js/src/ui/UserButton/UserButton.tsx
```diff
@@ -10,8 +10,9 @@ import { UserButtonPopover } from './UserButtonPopover';
...
   const { floating, reference, styles, toggle, isOpen } = usePopover({
-    defaultOpen: false,
+    defaultOpen: isButtonOpenFromOutside,
```
nikosdouvlis, line 15: ⚠️ The `defaultOpen` flag is only used for the **first** component mount, so changing the `isButtonOpenFromOutside` prop will not affect the behavior of the hook

packages/clerk-js/src/ui/hooks/usePopover.ts
```diff
@@ -18,13 +18,22 @@ type UsePopoverProps = {
...
+  const { update, reference, floating, strategy, x, y, context } = useFloating({
     open: isOpen,
     onOpenChange: setIsOpen,
     whileElementsMounted: props.autoUpdate === false ? undefined : autoUpdate,
     placement: props.placement || 'bottom-start',
     middleware: [offset(props.offset || 6), flip(), shift()],
   });
+
+  useEffect(() => {
+    setIsOpen(props.defaultOpen || false);
```
nikosdouvlis, line 30: If we end up using this useEffect, lets replace setIsOpen with one of the `toggle`, `open` or `close` functions defined at lines 39-41

packages/clerk-js/src/core/resources/DisplayConfig.ts
raptisj: @nikosdouvlis I think you are right on this one since the fields in `DisplayConfig` are mainly static and affected by API calls(ex. in instance)

packages/clerk-js/src/ui/UserButton/UserButton.tsx
raptisj: @nikosdouvlis `defaultOpen` does change since the usePopover gets rerenderd. That is how we setState within it

packages/react/src/components/uiComponents.tsx
nikosdouvlis: Hey John, sorry if I wasn't clear enough. This snippet is not showing a nested if, its showing a diff between the red/green parts.
So to rephrase my question, is the `if (prevProps.props !== this.props.props)`  enough?

packages/clerk-js/src/ui/UserButton/UserButton.tsx
```diff
@@ -10,29 +10,31 @@ import { UserButtonPopover } from './UserButtonPopover';
 import { UserButtonTrigger } from './UserButtonTrigger';
 
 const _UserButton = () => {
+  const { isOpen: isOpenProp } = useUserButtonContext();
   const { floating, reference, styles, toggle, isOpen } = usePopover({
     defaultOpen: false,
```
SokratisVidros, line 15: 🔧 I'd expect to pass the `isOpenProp` to `defaultOpen` in `usePopover` hook instead of introducing a new variable. That way we will avoid doing the `isOpenProp || isOpen` check throughout this component.

SokratisVidros (changes requested):

packages/clerk-js/src/ui/UserButton/UserButton.tsx
```diff
@@ -10,29 +10,31 @@ import { UserButtonPopover } from './UserButtonPopover';
...
     offset: 8,
+    isOpenProp,
   });
 
   return (
     <Flow.Root flow='userButton'>
       <Flex
         elementDescriptor={descriptors.userButtonBox}
-        isOpen={isOpen}
+        isOpen={isOpenProp || isOpen}
```
nikosdouvlis, line 25: Hey @raptisj , if a user passes `isOpen={true}`, then the UserButton will stay open and won't respond to user input even after clicking it again, since `isOpenProp || isOpen` will actually be `true || isOpen` that, of course, always evaluates to true. Is that correct?

raptisj: @nikosdouvlis yes that is true. But what is the expected behavior if someone wants to control this as such `isOpen={true}`?

packages/clerk-js/src/ui/UserButton/UserButton.tsx
raptisj: @SokratisVidros That was the initial implementation but it changed. Default only handles the initial state and is not programmatically controlled. 
Something like this `isOpenProp || isOpen` exists because you have to choose how you want to handle the open/close state. Do you want the library to do it or let the user via prop?

SokratisVidros (approved):

nikosdouvlis (approved):
Looks good!

PR #417, opened 2022-10-04
fix(clerk-js): Replace gravatar with Boring avatar

raptisj:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Replace gravatar with Boring avatar package. 
- Also, improve how we handle initials or the lack of them.

With the removal of gravatar the flow will be the following:
1. if a user has an image, show the image
2. if a user doesn’t have an image, show first/last name initials with Boring avatar
3. if a user doesn’t have an image nor first/last name initials, show only Boring avatar

Screen Shot 2022-10-07 at 11 50 51 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,5 +49,8 @@
     "lerna": "lerna",
     "nuke": "rm -r node_modules; for d in packages/*/node_modules; do echo $d; rm -r $d; done",
     "yalc:all": "for d in packages/*/; do echo $d; cd $d; yalc push --replace; cd '../../'; done"
+  },
+  "dependencies": {
+    "boring-avatars": "^1.7.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -49,8 +49,5 @@
     "lerna": "lerna",
     "nuke": "rm -r node_modules; for d in packages/*/node_modules; do echo $d; rm -r $d; done",
     "yalc:all": "for d in packages/*/; do echo $d; cd $d; yalc push --replace; cd '../../'; done"
-  },
-  "dependencies": {
-    "boring-avatars": "^1.7.0"
   }
 }
--- a/packages/clerk-js/src/ui/common/BoringAvatar.tsx
+++ b/packages/clerk-js/src/ui/common/BoringAvatar.tsx
@@ -1,0 +1,144 @@
+/**
+    This module was extracted from the boring-avatars library (https://github.com/boringdesigners/boring-avatars)
+    MIT License, Copyright (c) 2021 boringdesigners
+ */
+
+const ELEMENTS = 3;
+const SIZE = 80;
+
+export const hashCode = (name: string) => {
+  let hash = 0;
+  for (let i = 0; i < name.length; i++) {
+    const character = name.charCodeAt(i);
+    hash = (hash << 5) - hash + character;
+    hash = hash & hash; // Convert to 32bit integer
+  }
+  return Math.abs(hash);
+};
+
+export const getDigit = (number: number, ntn: number) => {
+  return Math.floor((number / Math.pow(10, ntn)) % 10);
+};
+
+export const getUnit = (number: number, range: number, index?: number) => {
+  const value = number % range;
+
+  if (index && getDigit(number, index) % 2 === 0) {
+    return -value;
+  } else {
+    return value;
+  }
+};
+
+export const getRandomColor = (number: number, colors: string[], range: number) => {
+  return colors[number % range];
+};
+
+function generateColors(name: string, colors: string[]) {
+  const numFromName = hashCode(name);
+  const range = colors && colors.length;
+
+  const elementsProperties = Array.from({ length: ELEMENTS }, (_, i) => ({
+    color: getRandomColor(numFromName + i, colors, range),
+    translateX: getUnit(numFromName * (i + 1), SIZE / 10, 1),
+    translateY: getUnit(numFromName * (i + 1), SIZE / 10, 2),
+    scale: 1.2 + getUnit(numFromName * (i + 1), SIZE / 20) / 10,
+    rotate: getUnit(numFromName * (i + 1), 360, 1),
+  }));
+
+  return elementsProperties;
+}
+
+const AvatarMarble = props => {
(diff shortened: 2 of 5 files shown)


packages/clerk-js/src/ui/elements/Avatar.tsx
```diff
@@ -93,21 +91,37 @@ export const Avatar = (props: AvatarProps) => {
...
+      sx={{ position: 'relative' }}
+      justify='center'
+      align='center'
     >
-      {initials}
-    </Text>
+      <BoringAvatar
+        size={props.size?.(parsedInternalTheme)}
+        name={`${props.firstName} ${props.lastName}`}
+        colors={['#CEEBD1', '#B6DEB9', '#B1CCB4', '#AEBFAF', '#A6ADA7']}
```
nikosdouvlis, line 105: ⚠️ These values should be coming from the theme, otherwise, they won't be customizable by the end user.
Let's discuss this because its possible that we'd need to provide `variables.colorAvatar` or similar API for that.

package.json
```diff
@@ -49,5 +49,8 @@
     "lerna": "lerna",
     "nuke": "rm -r node_modules; for d in packages/*/node_modules; do echo $d; rm -r $d; done",
     "yalc:all": "for d in packages/*/; do echo $d; cd $d; yalc push --replace; cd '../../'; done"
+  },
+  "dependencies": {
+    "boring-avatars": "^1.7.0"
```
nikosdouvlis, line 54: 🛑 This should be a dependency of `clerk-js`, but right now it's in the root-level `package.json`

packages/clerk-js/src/ui/elements/Avatar.tsx
SokratisVidros: We hardcoded them on purpose, for now, to test what it looks like. We will get back to this.

packages/clerk-js/src/ui/elements/Avatar.tsx
```diff
@@ -51,14 +49,14 @@ export const Avatar = (props: AvatarProps) => {
...
-    initials && (!avatarExists || error) ? (
+    !avatarExists || error ? (
       <InitialsAvatarFallback {...props} />
     ) : (
       <Image
         elementDescriptor={imageElementDescriptor || descriptors.avatarImage}
         alt={fullName}
         title={fullName}
-        src={src || GRAVATAR_DEFAULT_AVATAR}
+        src={src || (profileImageUrl as string)}
```
nikosdouvlis, line 59: ❓ Why do we need the explicit `as string` type assertion here? Let's see if there's a way to avoid it by correctly narrowing its type before that point.

nikosdouvlis:
It looks like adding `boring-avatars` increased the bundle size by ~3kb. Is there any way to extract only the module we need from that repo?

I know this looks like nitpicking, but saving a couple of KBs here and there comes a long way ✌🏻

packages/clerk-js/src/ui/elements/Avatar.tsx
raptisj: @nikosdouvlis this acted as a fallback but I guess it should be just `src={src || ''}`

packages/clerk-js/src/ui/common/BoringAvatar.tsx
```diff
@@ -0,0 +1,144 @@
+/**
```
nikosdouvlis, line 1: Great!

packages/clerk-js/src/ui/elements/Avatar.tsx
```diff
@@ -51,14 +49,14 @@ export const Avatar = (props: AvatarProps) => {
...
-    initials && (!avatarExists || error) ? (
+    !avatarExists || error ? (
       <InitialsAvatarFallback {...props} />
     ) : (
       <Image
         elementDescriptor={imageElementDescriptor || descriptors.avatarImage}
         alt={fullName}
         title={fullName}
-        src={src || GRAVATAR_DEFAULT_AVATAR}
+        src={src || ''}
```
nikosdouvlis, line 59: ❓ Don't we need a default fallback in case something goes extremely wrong?

raptisj: @nikosdouvlis in theory we will never reach this state. @SokratisVidros should we keep this gravatar default for such an extreme case?

SokratisVidros: Let's remove it.

packages/clerk-js/src/ui/common/BoringAvatar.tsx
```diff
@@ -0,0 +1,142 @@
...
+    This module was extracted from the boring-avatars library (https://github.com/boringdesigners/boring-avatars)
+    MIT License, Copyright (c) 2021 boringdesigners
```
SokratisVidros, line 3: 💯

SokratisVidros (approved):

packages/clerk-js/src/ui/elements/Avatar.tsx
```diff
@@ -91,23 +89,41 @@ export const Avatar = (props: AvatarProps) => {
   );
 };
 
+const avatarColors = ['#6C47FF', '#5BC5EF', '#FBD486', '#BEF972', '#FF8F8F'];
```
nikosdouvlis, line 92: Let's merge this one and figure out a way to make the avatar colors customizable via `appearance` with a new PR.

nikosdouvlis (approved):
This is looking great, good job keeping the bundle size unaffected :)

PR #420, opened 2022-10-05
refactor(clerk-js): Create generic Select component

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Create a generic Select component that will be used in PhoneInput, UserProfile and the new Org components.
```
<Select 
  option={selectedOption}
  options{options} 
  onChange={(option) => {setValue(option.customField)}
  optionBuilder={(option) => (<p>{option.text}</p>)}
  noResultsMessage="Not Found"
  placeholder="Search.."
  comparator={(term, option) => option.searchTerm.toLowerCase().includes(term.toLowerCase())}
>
  <SelectButton>
    {selectedOption.text}
  </SelectButton>
  <SelectOptionList/>
</Select>
```

Example of closed select:
Screenshot 2022-10-13 at 11 57 03
Example of open select:
Screenshot 2022-10-13 at 11 56 14
>

<!-- Fixes # (issue number) --


desiprisg:
As it stands, the component is slow when pressing down on the arrow keys to navigate (with large lists). Will find a fix asap.

packages/clerk-js/src/ui/elements/Select.tsx
```diff
@@ -0,0 +1,316 @@
...
+    ...rest
+  } = props;
+  const { floating, reference, styles, toggle: togglePopover, isOpen } = usePopover({ autoUpdate: false });
+  const [selectedIndex, setSelectedIndex] = useState(defaultIndex);
+  const selectedItemRef = React.useRef<HTMLDivElement>(null);
+  const { filteredItems, searchInputProps } = useSearchInput({
+    items: options,
+    comparator: (term, item) => {
+      if (typeof item === 'object') {
+        return item.searchTerm.toLowerCase().includes(term.toLowerCase());
```
SokratisVidros, line 69: ❓ What's the type of item? It's safer to check for `item.searchTerm` instead of checking if it's an object.

packages/clerk-js/src/ui/elements/Select.tsx
```diff
@@ -0,0 +1,316 @@
...
+      } else {
```
SokratisVidros, line 70: 🙃 No need to the `else`, just `return true` would work.

packages/clerk-js/src/ui/elements/Select.tsx
```diff
@@ -0,0 +1,316 @@
...
+    },
+  });
+  const [selectedOption, setSelectedOption] = useState(filteredItems[selectedIndex]);
+
+  const scrollToItemOnSelectedIndexChange = () => {
+    if (!isOpen) {
+      setSelectedIndex(0);
+      return;
+    }
+    selectedItemRef.current?.scrollIntoView({ block: 'nearest' });
```
SokratisVidros, line 82: 💯

packages/clerk-js/src/ui/elements/Select.tsx
```diff
@@ -0,0 +1,316 @@
...
+
+const SelectOptionBuilder = React.memo(React.forwardRef(_SelectOptionBuilder)) as <T extends DefaultOption>(
+  props: SelectOptionBuilderProps<T> & { ref?: React.ForwardedRef<HTMLDivElement> },
+) => ReturnType<typeof _SelectOptionBuilder>;
+
+const SelectSearchbar = (props: PropsOfComponent<typeof InputWithIcon>) => {
+  const { sx, ...rest } = props;
+  React.useEffect(() => {
+    // @ts-expect-error
+    return () => props.onChange({ target: { value: '' } });
```
SokratisVidros, line 222: ❓ Can you please elaborate on this line? Why do we trigger the `onChange` on unmount?

packages/clerk-js/src/ui/elements/Select.tsx
```diff
@@ -0,0 +1,316 @@
...
+        colorScheme='neutral'
+        variant='ghost'
+        textVariant='smallMedium'
+        isDisabled={props.isDisabled}
+        sx={[
+          theme => ({
+            paddingLeft: theme.space.$3x5,
+            paddingRight: theme.space.$3x5,
+            backgroundColor: theme.colors.$blackAlpha50,
+            position: 'absolute',
```
SokratisVidros, line 290: ❓ Are we sure we need absolute positioning for the trigger button? I'd expect the opposite.

packages/clerk-js/src/ui/elements/Select.tsx
desiprisg: If it's an object(and not a string or number) it will have the searchTerm property based on the types.

packages/clerk-js/src/ui/elements/Select.tsx
desiprisg: You're right. That got away. This needs to be handled by the PhoneInput.

packages/clerk-js/src/ui/elements/Select.tsx
desiprisg: We clear the search state provided by the `useSearchInput` hook to avoid persisting the search state when reopening. Do we not want this? It is the previous behaviour.

desiprisg:
Will make some changes make the component a bit more flexible!

packages/clerk-js/src/ui/elements/Select.tsx
```diff
@@ -0,0 +1,338 @@
...
+        width: '100%',
+        paddingLeft: theme.space.$2,
+        paddingRight: theme.space.$2,
+        ...(isSelected && { backgroundColor: theme.colors.$blackAlpha200 }),
+        '&:hover': {
+          backgroundColor: theme.colors.$blackAlpha200,
+        },
+      })}
+    >
+      {['string', 'number'].includes(typeof option) && <Text>{option}</Text>}
```
nikosdouvlis, line 50: Do we need this check here? What else do you expect to receive here?

desiprisg: This is the default builder. In order to use the default builder, the Options must be numbers or strings. We cannot render an object there. In that case, a builder will need to be provided.

nikosdouvlis (approved):

packages/clerk-js/src/ui/elements/Select.tsx
desiprisg: ✔️

PR #429, opened 2022-10-10
fix(clerk-js): Add customization for avatar gradient from appearance …

raptisj:
…provider

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Add customization for avatar gradient from Appearance provider.

### DIFF
--- a/packages/clerk-js/src/ui/elements/Avatar.tsx
+++ b/packages/clerk-js/src/ui/elements/Avatar.tsx
@@ -53,7 +53,7 @@ export const Avatar = (props: AvatarProps) => {
       <InitialsAvatarFallback {...props} />
     ) : (
       <Image
-        elementDescriptor={imageElementDescriptor || descriptors.avatarImage}
+        elementDescriptor={[imageElementDescriptor, descriptors.avatarImage]}
         alt={fullName}
         title={fullName}
         src={src || ''}
@@ -66,7 +66,7 @@ export const Avatar = (props: AvatarProps) => {
   // TODO: Revise size handling. Do we need to be this dynamic or should we use the theme instead?
   return (
     <Flex
-      elementDescriptor={boxElementDescriptor || descriptors.avatar}
+      elementDescriptor={[boxElementDescriptor, descriptors.avatarBox]}
       sx={[
         theme => ({
           flexShrink: 0,
@@ -95,7 +95,6 @@ function InitialsAvatarFallback(props: AvatarProps) {
 
   return (
     <Flex
-      elementDescriptor={descriptors.avatar}
       sx={{ position: 'relative' }}
       justify='center'
       align='center'
--- a/packages/types/src/appearance.ts
+++ b/packages/types/src/appearance.ts
@@ -206,7 +206,7 @@ export type ElementsConfig = {
   formFieldInputCopyToClipboardButton: WithOptions<never, never, never>;
   formFieldInputCopyToClipboardIcon: WithOptions<never, never, never>;
 
-  avatar: WithOptions<never, never, never>;
+  avatarBox: WithOptions<never, never, never>;
   avatarImage: WithOptions<never, never, never>;
 
   // TODO: We can remove "Popover" from these:
--- a/packages/clerk-js/src/ui/customizables/elementDescriptors.ts
+++ b/packages/clerk-js/src/ui/customizables/elementDescriptors.ts
@@ -89,7 +89,7 @@ export const APPEARANCE_KEYS = containsAllElementsConfigKeys([
   'formFieldInputCopyToClipboardButton',
   'formFieldInputCopyToClipboardIcon',
 
-  'avatar',
+  'avatarBox',
   'avatarImage',
 
   'userButtonBox',


packages/types/src/appearance.ts
```diff
@@ -361,6 +361,12 @@ export type Variables = {
    * The background color for the card container.
    */
   colorBackground?: CssColor;
+
+  /*
+   * An array of colors to create the avatar gradient.
+   * This will take place when a user doesn't provide an avatar image.
+   */
+  colorAvatarGradient?: CssColor[];
```
nikosdouvlis, line 369:
```suggestion
  /*
   * If a user hasn't provided an avatar image, a colorful gradient avatar will be shown as fallback.
   * To customise the gradient,pass an array of CSS colors.
   * @default ['#6C47FF', '#5BC5EF', '#FBD486', '#BEF972', '#FF8F8F'];
   */
  colorAvatarGradient?: CssColor[];
```

Also, let's move it right under `colorInputBackground`

packages/clerk-js/src/ui/elements/Avatar.tsx
```diff
@@ -104,7 +106,7 @@ function InitialsAvatarFallback(props: AvatarProps) {
       <BoringAvatar
         size={Number(props.size?.(parsedInternalTheme))}
         name={`${props.firstName} ${props.lastName}`}
-        colors={avatarColors}
+        colors={avatarColors as string[]}
```
nikosdouvlis, line 109: In general, let's try to avoid explicit type assertions (`as string[]`) as much as possible. When you use a type assertion, you lose all benefits of the TS inference from that point on. For example, the above line will only fail during **runtime** if avatarColors takes an unexpected value since TS will not be able to throw an error in advance.

packages/clerk-js/src/ui/common/BoringAvatar.tsx
```diff
@@ -49,7 +49,15 @@ function generateColors(name: string, colors: string[]) {
   return elementsProperties;
 }
 
-const AvatarMarble = props => {
+type AvatarProps = {
```
nikosdouvlis, line 52: Nice ✅

packages/clerk-js/src/ui/foundations/colors.ts
```diff
@@ -37,6 +37,7 @@ export const colors = Object.freeze({
   ...whiteAlpha,
   colorBackground: 'white',
   colorInputBackground: 'white',
+  colorAvatarGradient: [],
```
nikosdouvlis, line 40:
```suggestion
  colorAvatarGradient: ['#6C47FF', '#5BC5EF', '#FBD486', '#BEF972', '#FF8F8F'],
```

Let's move the default avatar colors inside the theme and let `parseAppearance` handle merging the default with the user-provided values.

You're going to need a couple of more changes, writing them in comments as well

packages/clerk-js/src/ui/customizables/parseVariables.ts
```diff
@@ -19,6 +20,7 @@ export const createColorScales = (theme: Theme) => {
     colorInputText: toHSLA(variables.colorInputText),
     colorBackground: toHSLA(variables.colorBackground),
     colorInputBackground: toHSLA(variables.colorInputBackground),
+    colorAvatarGradient: (variables.colorAvatarGradient as string[])?.map(color => toHSLA(color)),
```
nikosdouvlis, line 23:
```suggestion
    colorAvatarGradient: variables.colorAvatarGradient ? variables.colorAvatarGradient.map(toHSLA) : undefined,
```

Note 1: I removed the ternary to make the `undefined` value more visible. This is a rather complex part of the codebase so let's prefer clarity over shortcuts 

Note 2: If `colorAvatarGradient` gets `undefined` as a value, `parseAppearance` will use the default value from the internal theme as a fallback in the next step

packages/clerk-js/src/ui/elements/Avatar.tsx
nikosdouvlis: If you implement the changes the other comments suggest, I think we can simply:
```suggestion
        colors={parsedInternalTheme.colors.$colorAvatarGradient}
```

nikosdouvlis (approved):

packages/clerk-js/src/ui/elements/Avatar.tsx
```diff
@@ -95,6 +95,7 @@ function InitialsAvatarFallback(props: AvatarProps) {
 
   return (
     <Flex
+      elementDescriptor={descriptors.avatar}
```
nikosdouvlis, line 98: Do we need this? 

It looks like the parent Flex already has an `avatar` descriptor

raptisj: @nikosdouvlis yes we need this. There is a different `<Flex />`  element in each case.

nikosdouvlis: I pushed a change, @raptisj please review the last commit.

PR #430, opened 2022-10-11
feat(clerk-js): Introduce Menu element

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Example:
```
      <Menu>
        <MenuTrigger>
          <Button sx={{ width: '100px' }}>Open</Button>
        </MenuTrigger>
        <MenuList>
          <MenuItem>Action A</MenuItem>
          <MenuItem>Action B</MenuItem>
          <MenuItem>Very Long Action C</MenuItem>
          <MenuItem>Action D</MenuItem>
          <MenuItem>Action E</MenuItem>
        </MenuList>
      </Menu>
```

Screenshot 2022-10-11 at 23 56 26

### DIFF
--- a/packages/clerk-js/src/ui/elements/Menu.tsx
+++ b/packages/clerk-js/src/ui/elements/Menu.tsx
@@ -44,7 +44,10 @@ export const MenuTrigger = (props: MenuTriggerProps) => {
 
   return cloneElement(children, {
     ref: reference,
-    onClick: toggle,
+    onClick: (e: React.MouseEvent) => {
+      children.props?.onClick?.(e);
+      toggle();
+    },
   });
 };
 
--- a/packages/clerk-js/src/ui/elements/Menu.tsx
+++ b/packages/clerk-js/src/ui/elements/Menu.tsx
@@ -1,0 +1,175 @@
+import React, { PropsWithChildren, useEffect, useRef } from 'react';
+
+import { createContextAndHook } from '../../ui/utils';
+import { Button, Col } from '../customizables';
+import { usePopover } from '../hooks';
+import { animations, PropsOfComponent } from '../styledSystem';
+import { colors } from '../utils/colors';
+import { Portal } from './Portal';
+
+type UsePopoverReturn = ReturnType<typeof usePopover>;
+
+type MenuState = {
+  popoverCtx: UsePopoverReturn;
+};
+
+const [MenuStateCtx, useMenuState] = createContextAndHook<MenuState>('MenuState');
+
+type MenuProps = PropsWithChildren<Record<never, never>>;
+
+export const Menu = (props: MenuProps) => {
+  const popoverCtx = usePopover({
+    placement: 'right-start',
+    offset: 8,
+  });
+
+  return (
+    <MenuStateCtx.Provider
+      value={{ value: { popoverCtx } }}
+      {...props}
+    />
+  );
+};
+
+type MenuButtonProps = PropsOfComponent<typeof Button>;
+
+export const MenuButton = (props: MenuButtonProps) => {
+  const { popoverCtx } = useMenuState();
+  const { reference, toggle } = popoverCtx;
+
+  return (
+    <Button
+      ref={reference}
+      onClick={toggle}
+      {...props}
+    />
+  );
+};
+
+const findMenuItem = (el: Element, siblingType: 'prev' | 'next', options?: { countSelf?: boolean }) => {
+  let tagName = options?.countSelf ? el.tagName : '';
+  let sibling: Element | null = el;
+  while (sibling && tagName !== 'BUTTON') {
+    sibling = sibling[siblingType === 'prev' ? 'previousElementSibling' : 'nextElementSibling'];
+    tagName = sibling?.tagName ?? '';
+  }
+  return sibling;
+};
+
+type MenuListProps = PropsOfComponent<typeof Col>;
+
+export const MenuList = (props: MenuListProps) => {
+  const { sx, ...rest } = props;
+  const { popoverCtx } = useMenuState();
+  const { floating, styles, isOpen } = popoverCtx;
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    const current = containerRef.current;
+    floating(current);
+    current?.focus();
+  }, [isOpen]);
+
+  const onKeyDown = (e: React.KeyboardEvent<HTMLDivElement>) => {
+    const current = containerRef.current;
+    if (!current) {
+      return;
+    }
+
+    if (current !== document.activeElement) {
+      return;
+    }
+
+    if (e.key === 'ArrowDown') {
+      return (findMenuItem(current.children[0], 'next', { countSelf: true }) as HTMLElement)?.focus();
+    }
+  };
+
+  if (!isOpen) {
+    return null;
+  }
+
+  return (
+    <Portal>
+      <Col
+        ref={containerRef}
+        role='menu'
+        onKeyDown={onKeyDown}
+        tabIndex={0}
+        sx={[
+          theme => ({
+            backgroundColor: colors.makeSolid(theme.colors.$colorBackground),
+            border: theme.borders.$normal,
+            outline: 'none',
+            borderRadius: theme.radii.$lg,
+            borderColor: theme.colors.$blackAlpha200,
+            paddingTop: theme.space.$2,
+            paddingBottom: theme.space.$2,
+            overflow: 'hidden',
+            top: `calc(100% + ${theme.space.$2})`,
+            animation: `${animations.dropdownSlideInScaleAndFade} ${theme.transitionDuration.$slower} ${theme.transitionTiming.$slowBezier}`,
+            transformOrigin: 'top center',
+            boxShadow: theme.shadows.$boxShadow1,
+            zIndex: theme.zIndices.$dropdown,
+          }),
+          sx,
+        ]}
+        style={styles}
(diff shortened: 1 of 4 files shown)


packages/clerk-js/src/ui/elements/Menu.tsx
```diff
@@ -0,0 +1,175 @@
...
+    if (!current) {
+      return;
+    }
+
+    if (current !== document.activeElement) {
+      return;
+    }
+
+    if (e.key === 'ArrowDown') {
+      return (findMenuItem(current.children[0], 'next', { countSelf: true }) as HTMLElement)?.focus();
```
nikosdouvlis, line 84: Lot's of things happening here, can we break this statement in multiple lines so its a bit easier to follow along?

nikosdouvlis: The `current.children[0]` looks like a low-level way to target the first child and using it makes me _slightly_ uncomfortable.

Can we get the `children` from the props and use that instead?

packages/clerk-js/src/ui/elements/Menu.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const { toggle } = popoverCtx;
+  const buttonRef = useRef<HTMLButtonElement>(null);
+
+  const onKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>) => {
+    const current = buttonRef.current;
+    if (!current) {
+      return;
+    }
+
+    const keyMap: Record<string, 'prev' | 'next'> = {
```
nikosdouvlis, line 138: Is there a reason for this very generic `keyMap` instead of an early return if `e.key` is not ArrowUp or ArrowDown?

If you think we'll need to detect more events in the future, consider using a switch on `e.key` instead

packages/clerk-js/src/ui/elements/Menu.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+export const Menu = (props: MenuProps) => {
+  const popoverCtx = usePopover({
+    placement: 'right-start',
+    offset: 8,
+  });
+
+  return (
+    <MenuStateCtx.Provider
+      value={{ value: { popoverCtx } }}
```
nikosdouvlis, line 28: `{ value: { popoverCtx } }` will get create a new object reference on each render, causing all other listeners of that context to rerender as a result.

Lets use a `useMemo` as we do in other places as well.

packages/clerk-js/src/ui/elements/Menu.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+export const MenuList = (props: MenuListProps) => {
+  const { sx, ...rest } = props;
+  const { popoverCtx } = useMenuState();
+  const { floating, styles, isOpen } = popoverCtx;
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    const current = containerRef.current;
+    floating(current);
```
nikosdouvlis, line 69: I'd expect `floating` to be passed to the `ref` prop but I guess this is a way to use `floating` while still having a custom reference (`containerRef`) you can use for `.focus`. Can you confirm that this is indeed the case?

packages/clerk-js/src/ui/elements/Menu.tsx
```diff
@@ -0,0 +1,175 @@
...
+  return (
+    <Button
+      ref={reference}
+      onClick={toggle}
+      {...props}
+    />
+  );
+};
+
+const findMenuItem = (el: Element, siblingType: 'prev' | 'next', options?: { countSelf?: boolean }) => {
```
nikosdouvlis, line 49: I'm a bit unsure about the intention of the `countSelf` param here. Can you elaborate?

packages/clerk-js/src/ui/elements/Menu.tsx
```diff
@@ -0,0 +1,175 @@
...
+  let tagName = options?.countSelf ? el.tagName : '';
+  let sibling: Element | null = el;
+  while (sibling && tagName !== 'BUTTON') {
```
nikosdouvlis, line 52: ❓ Are you sure that every browser returns the tagName in uppercase? Shall we normalise tagName to uppercase as well before the comparison just to be safe?

packages/clerk-js/src/ui/elements/Menu.tsx
```diff
@@ -0,0 +1,175 @@
+import React, { PropsWithChildren, useEffect, useRef } from 'react';
+
+import { createContextAndHook } from '../../ui/utils';
+import { Button, Col } from '../customizables';
+import { usePopover } from '../hooks';
+import { animations, PropsOfComponent } from '../styledSystem';
+import { colors } from '../utils/colors';
+import { Portal } from './Portal';
+
+type UsePopoverReturn = ReturnType<typeof usePopover>;
```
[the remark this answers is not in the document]
desiprisg: Alright!

packages/clerk-js/src/ui/elements/Menu.tsx
desiprisg: Yes, that's the case.

packages/clerk-js/src/ui/elements/Menu.tsx
desiprisg: When using the function at the `MenuList`, we want the first `button` found, including the element passed to the function, so in that case, `countSelf` would be `true`.

packages/clerk-js/src/ui/elements/Menu.tsx
desiprisg: Sure!

packages/clerk-js/src/ui/elements/Menu.tsx
desiprisg: Can you give me an example? The function receives an `Element`.

packages/clerk-js/src/ui/elements/Menu.tsx
desiprisg: Will change it. It looks better.

packages/clerk-js/src/ui/elements/Menu.tsx
desiprisg: ✔️

nikosdouvlis:
This is looking real nice. One last comment regarding the API. What do you think about changing MenuButton to a MenuTrigger that passes the floating ref and the onClick props to its first child,[ similar to what mantine does](https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/Menu/MenuTarget/MenuTarget.tsx)? 

That way, we'll be able to pass a preexisting button or a ButtonIcon short of component without having to override the styles of MenuButton

```
      <Menu> 
        <MenuTrigger>
           <Button>Open</Button>
        <MenuTrigger/>
        <MenuList>
          <MenuItem>Action A</MenuItem>
          <MenuItem>Action B</MenuItem>
          <MenuItem>Very Long Action C</MenuItem>
          <MenuItem>Action D</MenuItem>
          <MenuItem>Action E</MenuItem>
        </MenuList>
      </Menu>
```

packages/clerk-js/src/ui/elements/Menu.tsx
raptisj: @desiprisg @nikosdouvlis I wonder, couldn't this be simplified using the index of the child as a source of truth? 🙃

example: https://github.com/clerkinc/javascript/commit/08e0eef0dba2ecfbfb3e4ca9663618758161c802

desiprisg: @raptisj Do you mean instead of passing the element itself in `findMenuItem`?

raptisj: @desiprisg what you said or a different approach, like passing the index to each child so it is aware of it. Just and idea

```  
const childrenWithProps = getValidChildren(children).map((child, index) =>
    React.cloneElement(child, {
      tabIndex: index,
    }),
  );
```

desiprisg: I think if we go that way, it's quite tricky to manage to have no re-renders of the children when navigating with the keyboard. That's why I didn't follow that route.

raptisj: @desiprisg ok got that🙃 

Although since changes are happening in the <Menu />(`isOpen` for example) the children will rerender so you can't really avoid that.

desiprisg: Yes, the children will render only when opening or closing the menu, not when navigating through the items with the keyboard. What I'm avoiding in this case, is the re-render of all the children when pressing `ArrowUp` and `ArrowDown`. If the children were aware of their index and I went that route, I'd need to update the context with the newly selected index each time (so the child can check if his index is the selected one), thus causing all the items to re-render.

nikosdouvlis (approved):

PR #448, opened 2022-10-19
fix(clerk-js): Make Select element themable

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/packages/clerk-js/src/ui/elements/Select.tsx
+++ b/packages/clerk-js/src/ui/elements/Select.tsx
@@ -3,7 +3,7 @@ import React, { isValidElement, PropsWithChildren, useState } from 'react';
 import { usePopover, useSearchInput } from '../../ui/hooks';
 import { Button, Flex, Icon, Text } from '../customizables';
 import { Caret, MagnifyingGlass } from '../icons';
-import { animations, PropsOfComponent } from '../styledSystem';
+import { animations, common, PropsOfComponent, ThemableCssProp } from '../styledSystem';
 import { colors, createContextAndHook } from '../utils';
 import { InputWithIcon } from './InputWithIcon';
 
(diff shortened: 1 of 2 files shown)


packages/clerk-js/src/ui/elements/Select.tsx
```diff
@@ -309,16 +311,13 @@ export const SelectButton = (props: PropsOfComponent<typeof Button>) => {
     <Button
       ref={reference}
       colorScheme='neutral'
-      variant='ghost'
+      variant='select'
```
nikosdouvlis, line 314: ❓ Do we need a variant for that? It looks like we're only using it in a single place. Do you think we can move the corresponding styles inside the Select component directly for now?

desiprisg: Not right now. I guess we can use the `ghost` variant which uses the same properties from `vars` and override the other properties in the `SelectButton` for now.

packages/clerk-js/src/ui/primitives/Button.tsx
```diff
@@ -8,7 +8,7 @@ import { Spinner } from './Spinner';
 
 const vars = createCssVariables('accent', 'accentDark', 'accentDarker', 'accentLighter', 'accentLightest', 'border');
 
-const { applyVariants, filterProps } = createVariants(theme => {
+const { applyVariants, filterProps } = createVariants((theme, props) => {
```
nikosdouvlis, line 11: Do we use these props?

nikosdouvlis (approved):

packages/clerk-js/src/ui/primitives/Button.tsx
desiprisg: Right, we don't anymore. That got away. Thanks.

PR #451, opened 2022-10-19
chore(clerk-js,clerk-react,types): Share hook and context implementation between `clerk-js` and `clerk-react`

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

This PR merges and extracts all contexts and hooks related to how Clerk provides access to its internal state (all the use* and the useCore* hooks) to the `clerk/shared` package.

This can be considered a first step to moving all state management logic to a completely separate and reusable package that both `clerk-react` and `clerk-js` can have as a dependency in the future. Short term, this allows us to reuse all the organization hooks **within clerk-js** exactly like a 3rd party dev would.

Some caveats however:
- The implementation is shared but the actual code still gets bundled twice: once in `clerk-js` (UMD package) and once in `clerk-react`. This can be avoided when we remove hotloading, as discussed in the past.
- Both packages use the concept of the "clerk singleton instance", hidden behind the `Clerk` interface. However, the actual implementation between `clerk-js` and `clerk-react` varies slightly. I've further aligned them by adding a `loaded` public getter on `core/Clerk` but more work is needed to make sure that both implementations satisfy `Clerk`. Again, this is something that removing hotloading will solve by default.
- The clerk-js hooks are still named `useCore*` to avoid a messy PR with renames throughout the codebase. We'll rename them in a separate PR
- In theory, we still need to check `isLoaded` within `clerk-js` (which will always be loaded), otherwise, TS won't be able to correctly narrow the types down from `Resource | undefined` to just `Resource`. Let's discuss alternatives


SokratisVidros (approved):
Nicely done! Kudos for taking the time to do the same for all user hooks as well.

packages/clerk-js/src/ui/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -227,3 +227,5 @@ export const ImpersonationFab = () => {
     </Portal>
   );
 };
+
+export const ImpersonationFab = withCoreSessionSwitchGuard(_ImpersonationFab);
```
desiprisg, line 231: 💯

packages/shared/src/hooks/useOrganization.tsx
```diff
@@ -0,0 +1,131 @@
...
+      invitationList: undefined;
+      membershipList: undefined;
+      membership: undefined;
+    }
+  | {
+      isLoaded: true;
+      organization: OrganizationResource;
+      invitationList: undefined;
+      membershipList: undefined;
+      membership: undefined;
```
desiprisg, line 30: ❓Why are these undefined in this case?

PR #453, opened 2022-10-21
fix(clerk-js): Prioritize focus of tabs when keyboard navigating

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/packages/clerk-js/src/ui/elements/Tabs.tsx
+++ b/packages/clerk-js/src/ui/elements/Tabs.tsx
@@ -27,7 +27,7 @@ type TabsProps = PropsOfComponent<typeof Col> & {
 export const Tabs = (props: TabsProps) => {
   const { defaultIndex = 0, children, sx, ...rest } = props;
   const [selectedIndex, setSelectedIndex] = React.useState(defaultIndex);
-  const [focusedIndex, setFocusedIndex] = React.useState(defaultIndex);
+  const [focusedIndex, setFocusedIndex] = React.useState(-1);
 
   return (
     <TabsContextProvider value={{ selectedIndex, setSelectedIndex, focusedIndex, setFocusedIndex }}>
@@ -90,7 +90,7 @@ export const Tab = (props: TabProps) => {
   }
 
   const { setSelectedIndex, selectedIndex, focusedIndex, setFocusedIndex } = useTabsContext();
-  const buttonRef = React.useRef<HTMLButtonElement | any>(null);
+  const buttonRef = React.useRef<HTMLButtonElement>(null);
   const isActive = tabIndex === selectedIndex;
   const isFocused = tabIndex === focusedIndex;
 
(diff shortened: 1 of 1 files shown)


packages/clerk-js/src/ui/elements/Tabs.tsx
```diff
@@ -184,7 +185,6 @@ export const TabPanel = (props: TabPanelProps) => {
     <Flex
       id={`cl-tabpanel-${tabIndex}`}
       role='tabpanel'
-      tabIndex={0}
```
raptisj, line 187: @desiprisg accessibility wise you need tabindex here as well

https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-automatic

packages/clerk-js/src/ui/elements/Tabs.tsx
```diff
@@ -118,8 +118,9 @@ export const Tab = (props: TabProps) => {
     >
       <Button
         onClick={onClick}
-        focusRing={isFocused}
+        focusRing={isActive}
```
raptisj, line 121: @desiprisg ❓ isn't this going to have a focused state when activation has occurred with a mouse event as well? Do we want that?

packages/clerk-js/src/ui/elements/Tabs.tsx
desiprisg: You're right. We probably should remove the `focusRing` from the `div` though.

packages/clerk-js/src/ui/elements/Tabs.tsx
desiprisg: Yes it will, however that's not the case in our use case right now since we have fields with `autofocus` below. Top priority is to keep the focus on the tabs when keyboard navigating and allow other elements to take focus when clicking. Did not want to make a big refactor right now, just fix this one issue, but feel free to share any ideas. We should definitely rethink this.

packages/clerk-js/src/ui/elements/Tabs.tsx
nikosdouvlis: Hey peeps, is this one ready to get merged?

raptisj (approved):
@desiprisg @nikosdouvlis I think this is ready for merge. Since this is for internal use it can easily be tweaked in the future if needed.

Issue #201, opened 2022-04-18
getToken changes after each render


marjan-pluto:
- [X] Review the documentation: https://docs.clerk.dev/
- [X] Search for existing issues: https://github.com/clerkinc/javascript/issues
- [X] Go through package changelog files.
- [ ] Provide the Frontend API key from your application dashboard.

## Package + Version

- [ ] `@clerk/clerk-js`
- [X] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] other:

### Version:

```
^3.2.7
```

### Browser/OS

`Chrome latest`

## Description

We started migrating to v3 and noticed `getToken` from `useAuth` function identity is not stable and as a result changes after each re-render.

This causes the the app to go into infinite loop when `getToken` is used in the dependancy array of something like `useEffect`.
Usage Snippet:
<img width="620" alt="image" src="https://user-images.githubusercontent.com/97608659/163834327-9a5940a2-b318-4c21-8cfe-badebf2d0817.png">

A workaround is to remove it from the dependancy array, however that does not make eslint happy
<img width="618" alt="image" src="https://user-images.githubusercontent.com/97608659/163834611-34590203-bbce-46c1-bd85-153a382039b7.png">

It would be ideal if `getToken` can be made stable as it was in v2.

nikosdouvlis:
Hello @marjan-pluto ! Many thanks for reporting this one :)

`getToken` doesn't really need to be in the dependency array, but I can totally see why this is problematic. Your comment about eslint makes perfect sense.

I'll be working on this during the next few days and I'll ping you here when a fix has been deployed. Thanks again!

waldothedeveloper:
if `getToken` is not needed, update the docs folks until this is resolved...I was trying to `setState` following the example from the docs below and I was getting an infinite loop because of the `getToken`...
[https://clerk.dev/docs/integration/firebase#:~:text=29-,%7D%2C%20%5BgetToken%5D)%3B,-30](url)

dimkl:
@waldothedeveloper thank you for reporting it. We have updated our docs.

rjdellecese:
The problem described by this issue hasn't been resolved, has it? If not, it seems to me like this issue should remain open.

[...]

PR #455, opened 2022-10-22
fix(clerk-react): Added useAuth test and made getToken, and signOut stable

franciscop:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Following on this issue https://github.com/clerkinc/javascript/issues/201, made the functions returned by `useAuth()` stable (as per [the usual React terminology](https://shopify.engineering/master-reacts-stable-values)); meaning they will not change between unrelated re-renders. This allows you to properly depend on them without worries about re-renders, and when the underlying value _does change_ (isomorphicClerk), then they will update properly.

The logic inside `useAuth()` should be identical as before, just reorganized it a bit to be able to use a `useMemo` to make `auth` also stable.

PS, modified the types a bit, unfortunately I'm not _that_ familiar with TS but the default installation was giving me types issues (before any code change, just by importing the files into the test file). So some of the changes were fixes, and the one modifying `UseAuthReturn` was because I had trouble with getting the highly-dynamic variable to work otherwise.

#201

### DIFF
--- a/packages/react/src/contexts/ClerkContextProvider.tsx
+++ b/packages/react/src/contexts/ClerkContextProvider.tsx
@@ -146,9 +146,9 @@ function deriveState(
       lastOrganizationMember: null,
     };
   }
-  const userId: string | null | undefined = state.user ? state.user.id : state.user;
+  const userId: string | null | undefined = state.user ? state.user.id : (state.user as null | undefined);
   const user = state.user;
-  const sessionId: string | null | undefined = state.session ? state.session.id : state.session;
+  const sessionId: string | null | undefined = state.session ? state.session.id : (state.session as null | undefined);
   const session = state.session;
   const organization = state.organization;
   const lastOrganizationInvitation = state.lastOrganizationInvitation;
--- a/packages/react/src/hooks/useAuth.test.tsx
+++ b/packages/react/src/hooks/useAuth.test.tsx
@@ -1,0 +1,49 @@
+import { render, screen, userEvent } from '@clerk/shared/testUtils';
+import React, { useEffect, useState } from 'react';
+
+import { useAuth } from './useAuth';
+import { ClerkProvider } from '../contexts/ClerkProvider';
+
+const withFakeClerk = Component => {
+  return props => (
+    <ClerkProvider frontendApi='https://example.com'>
+      <Component {...props} />
+    </ClerkProvider>
+  );
+};
+
+describe('useAuth()', () => {
+  it('auth is stable between re-renders', async () => {
+    let renderCount = 0;
+    const MyDemo = () => {
+      const [count, setCount] = useState(0);
+      const auth = useAuth();
+      useEffect(() => {
+        renderCount++;
+      }, [auth]);
+      return <button onClick={() => setCount(i => i + 1)}>Hi</button>;
+    };
+    const App = withFakeClerk(MyDemo);
+    render(<App />);
+    const btn = screen.getByText('Hi');
+    userEvent.click(btn);
+    expect(renderCount).toBe(1);
+  });
+
+  it('getToken is stable between re-renders', async () => {
+    let renderCount = 0;
+    const MyDemo = () => {
+      const [count, setCount] = useState(0);
+      const { getToken } = useAuth();
+      useEffect(() => {
+        renderCount++;
+      }, [getToken]);
+      return <button onClick={() => setCount(i => i + 1)}>Hi</button>;
+    };
+    const App = withFakeClerk(MyDemo);
+    render(<App />);
+    const btn = screen.getByText('Hi');
+    userEvent.click(btn);
+    expect(renderCount).toBe(1);
+  });
+});
--- a/packages/react/src/hooks/useAuth.test.tsx
+++ b/packages/react/src/hooks/useAuth.test.tsx
@@ -1,49 +1,62 @@
-import { render, screen, userEvent } from '@clerk/shared/testUtils';
-import React, { useEffect, useState } from 'react';
+import { render } from '@clerk/shared/testUtils';
+import React, { useEffect } from 'react';
 
 import { useAuth } from './useAuth';
 import { ClerkProvider } from '../contexts/ClerkProvider';
 
-const withFakeClerk = Component => {
-  return props => (
-    <ClerkProvider frontendApi='https://example.com'>
-      <Component {...props} />
-    </ClerkProvider>
-  );
-};
+import { withClerk } from '../components/withClerk';
+
+jest.mock('../components/withClerk', () => ({
+  withClerk: (Component: any) => (props: any) => {
+    return (
+      <Component
+        {...props}
+        clerk={{}}
+      />
+    );
+  },
+}));
 
 describe('useAuth()', () => {
-  it('auth is stable between re-renders', async () => {
-    let renderCount = 0;
-    const MyDemo = () => {
-      const [count, setCount] = useState(0);
-      const auth = useAuth();
-      useEffect(() => {
-        renderCount++;
-      }, [auth]);
-      return <button onClick={() => setCount(i => i + 1)}>Hi</button>;
-    };
-    const App = withFakeClerk(MyDemo);
-    render(<App />);
-    const btn = screen.getByText('Hi');
-    userEvent.click(btn);
-    expect(renderCount).toBe(1);
-  });
-
   it('getToken is stable between re-renders', async () => {
     let renderCount = 0;
-    const MyDemo = () => {
-      const [count, setCount] = useState(0);
+    const MyDemo = ({ id }: { id: number }) => {
       const { getToken } = useAuth();
       useEffect(() => {
         renderCount++;
       }, [getToken]);
-      return <button onClick={() => setCount(i => i + 1)}>Hi</button>;
+      return <button>Hi {id}</button>;
     };
-    const App = withFakeClerk(MyDemo);
-    render(<App />);
-    const btn = screen.getByText('Hi');
-    userEvent.click(btn);
+    const App = withClerk(({ id }: { id: number; clerk }) => (
+      <ClerkProvider frontendApi='https://example.com'>
+        <MyDemo id={id} />
+      </ClerkProvider>
+    ));
+    expect(renderCount).toBe(0);
+    const { rerender } = render(<App id={0} />);
+    expect(renderCount).toBe(1);
+    rerender(<App id={1} />);
+    expect(renderCount).toBe(1);
+  });
+
+  it('signOut is stable between re-renders', async () => {
+    let renderCount = 0;
+    const MyDemo = ({ id }: { id: number }) => {
+      const { signOut } = useAuth();
+      useEffect(() => {
+        renderCount++;
+      }, [signOut]);
+      return <button>Hi {id}</button>;
+    };
+    const App = withClerk(({ id }: { id: number; clerk }) => (
+      <ClerkProvider frontendApi='https://example.com'>
+        <MyDemo id={id} />
+      </ClerkProvider>
+    ));
+    expect(renderCount).toBe(0);
+    const { rerender } = render(<App id={0} />);
+    expect(renderCount).toBe(1);
+    rerender(<App id={1} />);
     expect(renderCount).toBe(1);
   });
 });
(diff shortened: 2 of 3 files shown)


packages/react/src/hooks/useAuth.ts
```diff
@@ -1,38 +1,20 @@
...
-  | {
-      isLoaded: true;
-      isSignedIn: true;
-      userId: string;
-      sessionId: string;
-      actor: ActJWTClaim | null;
-      signOut: SignOut;
-      getToken: GetToken;
-    };
+type UseAuthReturn = {
```
SokratisVidros, line 9: Can you please elaborate on the reasoning for this change? This new type definition isn't the same as the previous one. The main reason is that by declaring a union of type objects we create a stricter `UseAuthReturn` type that allows only the declared combinations.

packages/react/src/hooks/useAuth.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <ClerkProvider frontendApi='https://example.com'>
+      <Component {...props} />
+    </ClerkProvider>
+  );
+};
+
+describe('useAuth()', () => {
+  it('auth is stable between re-renders', async () => {
+    let renderCount = 0;
+    const MyDemo = () => {
```
SokratisVidros, line 18: 🔧  An alternative and more expressive way to test this is to rerender the component via prop updates as shown at https://testing-library.com/docs/example-update-props/

packages/react/src/hooks/useAuth.ts
```diff
@@ -79,44 +61,56 @@ export const useAuth: UseAuth = () => {
...
-  const getToken: GetToken = createGetToken(isomorphicClerk);
-  const signOut: SignOut = createSignOut(isomorphicClerk);
+  const getToken: GetToken = useMemo(() => {
+    return createGetToken(isomorphicClerk);
+  }, [isomorphicClerk]);
+  const signOut: SignOut = useMemo(() => {
+    return createSignOut(isomorphicClerk);
+  }, [isomorphicClerk]);
+
+  // Make this stable as well!
```
SokratisVidros, line 71: For this first iteration of this fix, I'd suggest focusing only on stabilizing `getToken`. This change might have undesired side-effects for useAuth use cases.

SokratisVidros (changes requested):
Thanks for the taking a stab at it. I've just added my comments after the first review.

packages/react/src/hooks/useAuth.ts
franciscop: That makes sense, specially considering probably most people destruct anyway so the value of stabilizing Auth itself is not so high; the reason I bundled both of the changes together is that I had a lot of trouble setting the whole testing and everything up, so since I was here I stabilized everything :)

> "This change might have undesired side-effects for useAuth use cases."

For a large project with so many tests I'd normally expect that if I make changes that I believe don't change the API, and no tests break, it should be fine. That's IMHO one of the main goals of tests, to be confident that changes+refactoring don't break things. (though I understand maybe these hooks are under-tested?)

packages/react/src/hooks/useAuth.ts
franciscop: > PS, modified the types a bit, unfortunately I'm not that familiar with TS but the default installation was giving me types issues [...] the one modifying UseAuthReturn was because I had trouble with getting the highly-dynamic variable to work otherwise.

If I change just the scope of `getToken()` (+ `signOut()`) I believe I might be able to bring the original types back.

Side question: why is it so... messy? IMHO `isSignedIn` should be `true` or `false`, no chance for `undefined`. Same with `userId` or `sessionId`, why not make them either `string | null`? Right now the potential combinatorial explosion is scary and you _need_ the types to make sense of it, but if simplified well it'd be much better (again just my opinion/tip, this is your project after all).

packages/react/src/hooks/useAuth.test.tsx
franciscop: Wow I didn't know that one, that seems very interesting! I'll definitely add a similar thing to [my own React testing library, thanks for the tip!](https://react-test.dev/). Let me see if I can get it to work here as well, but seems like it should be easy.

franciscop:
Thanks for the feedback! I'm in my lunch break, I won't be able to commit anything until the weekend, so feel free to add any other comment and I'll try to fix them all then.

packages/react/src/hooks/useAuth.test.tsx
franciscop: Fixed as suggested, thanks! (also [added it to my testing library](https://react-test.dev/documentation#props) 🥳)

packages/react/src/hooks/useAuth.ts
franciscop: Fixed this as well! Now it's only memoizing (with `useCallback()`) the two functions, `signOut` and `getToken`, and not the whole return.

franciscop:
Fixed the given points! Unfortunately I still had to change the types into 3 parts and cast each return independently with `as {NAME}`, since otherwise with the default code I was getting this error which I couldn't figure out how to fix in any other way:

```js
src/hooks/useAuth.ts:80:14 - error TS2322: Type '() => { isLoaded: false; isSignedIn: any; sessionId: string; userId: string; actor: any; signOut: SignOut; getToken: GetToken; } | { isLoaded: true; isSignedIn: false; sessionId: string; userId: string; actor: any; signOut: SignOut; getToken: GetToken; } | { ...; }' is not assignable to type 'UseAuth'.
@clerk/clerk-react:test:       Type '{ isLoaded: false; isSignedIn: any; sessionId: string; userId: string; actor: any; signOut: SignOut; getToken: GetToken; } | { isLoaded: true; isSignedIn: false; sessionId: string; userId: string; actor: any; signOut: SignOut; getToken: GetToken; } | { ...; }' is not assignable to type 'UseAuthReturn'.
@clerk/clerk-react:test:         Type '{ isLoaded: false; isSignedIn: any; sessionId: string; userId: string; actor: any; signOut: SignOut; getToken: GetToken; }' is not assignable to type 'UseAuthReturn'.
@clerk/clerk-react:test:           Type '{ isLoaded: false; isSignedIn: any; sessionId: string; userId: string; actor: any; signOut: SignOut; getToken: GetToken; }' is not assignable to type '{ isLoaded: true; isSignedIn: true; userId: string; sessionId: string; actor: ActJWTClaim; signOut: SignOut; getToken: GetToken; }'.
@clerk/clerk-react:test:             Types of property 'isLoaded' are incompatible.
@clerk/clerk-react:test:               Type 'false' is not assignable to type 'true'.
@clerk/clerk-react:test: 
@clerk/clerk-react:test:     80 export const useAuth: UseAuth = () => {
```

This error was *not* shown when running the tests without `useAuth.test.tsx`, which makes me think it's something visible either only when running the tests, or because of the mocking the clerk instance is not being done properly. I believe these type changes should be totally retro-compatible, but please let me know if those are not acceptable.

packages/react/src/hooks/useAuth.ts
```diff
@@ -77,10 +80,10 @@ type UseAuth = () => UseAuthReturn;
  */
 export const useAuth: UseAuth = () => {
   const { sessionId, userId, actor } = useAuthContext();
-  const isomorphicClerk = useIsomorphicClerkContext();
+  const isomorphicClerk = useIsomorphicClerkContext() as unknown as IsomorphicClerk;
```
franciscop, line 83:
```suggestion
  const isomorphicClerk = useIsomorphicClerkContext();
```

packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -146,9 +146,9 @@ function deriveState(
       lastOrganizationMember: null,
     };
   }
-  const userId: string | null | undefined = state.user ? state.user.id : state.user;
+  const userId: string | null | undefined = state.user ? state.user.id : (state.user as null | undefined);
```
SokratisVidros, line 149: ❓ Is this type assertion required?

packages/react/src/hooks/useAuth.ts
```diff
@@ -1,40 +1,43 @@
...
-      actor: ActJWTClaim | null;
-      signOut: SignOut;
-      getToken: GetToken;
-    };
+type InitReturn = {
+  isLoaded: false;
+  isSignedIn: undefined;
+  userId: undefined;
+  sessionId: undefined;
+  actor: null;
```
SokratisVidros, line 15: 🔧 This should also be an undefined. It was a mistake to make it null in this scenario in the first place.

packages/react/src/hooks/useAuth.ts
```diff
@@ -1,40 +1,43 @@
...
-  | {
-      isLoaded: true;
-      isSignedIn: true;
-      userId: string;
-      sessionId: string;
-      actor: ActJWTClaim | null;
-      signOut: SignOut;
-      getToken: GetToken;
-    };
+type InitReturn = {
```
SokratisVidros, line 10: A better naming for this interface is the `UnknownState` as undefined means that we can't tell yet if the user is signed in or signed out.

packages/react/src/hooks/useAuth.ts
```diff
@@ -1,40 +1,43 @@
...
+  isSignedIn: true;
+  userId: string;
+  sessionId: string;
+  actor: ActJWTClaim | null;
+  signOut: SignOut;
+  getToken: GetToken;
+};
 
-type UseAuth = () => UseAuthReturn;
+type UseAuth = () => InitReturn | LoadedReturn | UserReturn;
```
SokratisVidros, line 40: Let's avoid creating these three different states and stick to the original type definition to avoid the extra assertions.

packages/react/src/contexts/ClerkContextProvider.tsx
franciscop: Without this, `npm test` will give me this issue:

```
src/contexts/ClerkContextProvider.tsx:149:9 - error TS2322: Type 'string | UserResource' is not assignable to type 'string'.
      Type 'UserResource' is not assignable to type 'string'.

    149   const userId: string | null | undefined = state.user ? state.user.id : state.user;
```

Alternatively these also work:

```ts
const userId: string | null | undefined = state.user && state.user.id;
```

packages/react/src/hooks/useAuth.ts
franciscop: If I use the original, TS doesn't seem to differentiate them properly and I get this error:

```
Type '() => { isLoaded: false; isSignedIn: any; sessionId: any; userId: any; actor: any; signOut: SignOut; getToken: GetToken; } | { isLoaded: true; isSignedIn: false; sessionId: string; userId: string; actor: any; signOut: SignOut; getToken: GetToken; } | { ...; }' is not assignable to type 'UseAuth'.
      Type '{ isLoaded: false; isSignedIn: any; sessionId: any; userId: any; actor: any; signOut: SignOut; getToken: GetToken; } | { isLoaded: true; isSignedIn: false; sessionId: string; userId: string; actor: any; signOut: SignOut; getToken: GetToken; } | { ...; }' is not assignable to type 'UseAuthReturn'.
        Type '{ isLoaded: true; isSignedIn: false; sessionId: string; userId: string; actor: any; signOut: SignOut; getToken: GetToken; }' is not assignable to type 'UseAuthReturn'.
          Type '{ isLoaded: true; isSignedIn: false; sessionId: string; userId: string; actor: any; signOut: SignOut; getToken: GetToken; }' is not assignable to type '{ isLoaded: true; isSignedIn: true; userId: string; sessionId: string; actor: ActJWTClaim; signOut: SignOut; getToken: GetToken; }'.
            Types of property 'isSignedIn' are incompatible.
              Type 'false' is not assignable to type 'true'.
```

marjan-pluto (approved):

dimkl:
@franciscop thank you for the contribution but we will close this PR in favor of : https://github.com/clerkinc/javascript/pull/980

PR #464, opened 2022-10-25
Phone number improvements

raptisj:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Refactor phone number input to change flags onPaste

An example where the last phone number is not valid
https://user-images.githubusercontent.com/23478420/197823787-41f50d1a-7c1f-45dd-84c9-2b3e8792529e.mov

### DIFF
--- a/packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
+++ b/packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
@@ -1,5 +1,6 @@
-import React, { SyntheticEvent } from 'react';
+import React from 'react';
 
+import { getLongestValidCountryCode } from '../../../ui/utils/phoneUtils';
 import { Flex, Input, Text } from '../../customizables';
 import { Select, SelectButton, SelectOptionList } from '../../elements';
 import { PropsOfComponent } from '../../styledSystem';
@@ -20,14 +21,9 @@ const countryOptions = [...IsoToCountryMap.values()].map(createSelectOption);
 
 type PhoneInputProps = PropsOfComponent<typeof Input>;
 
-interface ClipboardEvent<T = Element> extends SyntheticEvent<T> {
-  clipboardData: DataTransfer;
-}
-
 export const PhoneInput = (props: PhoneInputProps) => {
   const { onChange: onChangeProp, value, ...rest } = props;
   const phoneInputRef = React.useRef<HTMLInputElement>(null);
-  const [hasBeenPasted, setHasBeenPasted] = React.useState(false);
   const { setPhoneNumber, cleanPhoneNumber, formattedPhoneNumber, selectedIso, setSelectedIso } =
     useFormattedPhoneNumber({
       defaultPhone: value as string,
@@ -43,46 +39,28 @@ export const PhoneInput = (props: PhoneInputProps) => {
   const selectedCountryOption = React.useMemo(() => {
     return countryOptions.find(o => o.country.iso === selectedIso) || countryOptions[0];
   }, [selectedIso]);
-  const dynamicPadding = selectedCountryOption.country.code.length * 5; // this to calculate the padding of the input field depending the length of country code
+  const dynamicPadding = selectedCountryOption.country.code.length * 5; // this is to calculate the padding of the input field depending the length of country code
 
   React.useEffect(callOnChangeProp, [cleanPhoneNumber]);
 
-  const handlePaste = (e: ClipboardEvent<HTMLInputElement>) => {
+  const handlePaste = (e: React.ClipboardEvent<HTMLInputElement>) => {
+    e.preventDefault();
     const inputValue = e.clipboardData.getData('text');
-    const testPattern = /\+/g;
 
-    if (testPattern.test(inputValue)) {
-      setHasBeenPasted(true);
-
-      let result = '';
-      let selectedCountry = null;
-      for (let i = 5; i > 1; i--) {
-        const code = inputValue.slice(1, i);
-
-        result = inputValue.slice(i, inputValue.length);
-        selectedCountry = countryOptions.find(o => o.country.code === code);
-
-        if (selectedCountry) {
-          break;
-        }
-      }
+    if (inputValue.includes('+')) {
+      const { phoneNumberValue, selectedCountry } = getLongestValidCountryCode(inputValue);
 
       if (selectedCountry) {
-        setSelectedIso(selectedCountry?.country.iso);
-        setPhoneNumber(result, selectedCountry?.country.iso);
+        setSelectedIso(selectedCountry?.iso);
+        setPhoneNumber(phoneNumberValue, selectedCountry?.iso);
         return;
       }
+      setPhoneNumber(inputValue);
     }
-    setSelectedIso(selectedIso);
-    setPhoneNumber(inputValue);
   };
 
   const handlePhoneNumberChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    if (!hasBeenPasted) {
-      setPhoneNumber(e.target.value);
-    }
-
-    setHasBeenPasted(false);
+    setPhoneNumber(e.target.value);
   };
 
   return (
--- a/packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
+++ b/packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { SyntheticEvent } from 'react';
 
 import { Flex, Input, Text } from '../../customizables';
 import { Select, SelectButton, SelectOptionList } from '../../elements';
@@ -20,9 +20,14 @@ const countryOptions = [...IsoToCountryMap.values()].map(createSelectOption);
 
 type PhoneInputProps = PropsOfComponent<typeof Input>;
 
+interface ClipboardEvent<T = Element> extends SyntheticEvent<T> {
+  clipboardData: DataTransfer;
+}
+
 export const PhoneInput = (props: PhoneInputProps) => {
   const { onChange: onChangeProp, value, ...rest } = props;
   const phoneInputRef = React.useRef<HTMLInputElement>(null);
+  const [hasBeenPasted, setHasBeenPasted] = React.useState(false);
   const { setPhoneNumber, cleanPhoneNumber, formattedPhoneNumber, selectedIso, setSelectedIso } =
     useFormattedPhoneNumber({
       defaultPhone: value as string,
@@ -38,8 +43,47 @@ export const PhoneInput = (props: PhoneInputProps) => {
   const selectedCountryOption = React.useMemo(() => {
     return countryOptions.find(o => o.country.iso === selectedIso) || countryOptions[0];
   }, [selectedIso]);
+  const dynamicPadding = selectedCountryOption.country.code.length * 5; // this to calculate the padding of the input field depending the length of country code
 
   React.useEffect(callOnChangeProp, [cleanPhoneNumber]);
+
+  const handlePaste = (e: ClipboardEvent<HTMLInputElement>) => {
+    const inputValue = e.clipboardData.getData('text');
+    const testPattern = /\+/g;
+
+    if (testPattern.test(inputValue)) {
+      setHasBeenPasted(true);
+
+      let result = '';
+      let selectedCountry = null;
+      for (let i = 5; i > 1; i--) {
+        const code = inputValue.slice(1, i);
+
+        result = inputValue.slice(i, inputValue.length);
+        selectedCountry = countryOptions.find(o => o.country.code === code);
+
+        if (selectedCountry) {
+          break;
+        }
+      }
+
+      if (selectedCountry) {
+        setSelectedIso(selectedCountry?.country.iso);
+        setPhoneNumber(result, selectedCountry?.country.iso);
+        return;
+      }
+    }
+    setSelectedIso(selectedIso);
+    setPhoneNumber(inputValue);
+  };
+
+  const handlePhoneNumberChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    if (!hasBeenPasted) {
+      setPhoneNumber(e.target.value);
+    }
+
+    setHasBeenPasted(false);
+  };
 
   return (
     <Flex
@@ -86,6 +130,7 @@ export const PhoneInput = (props: PhoneInputProps) => {
             })}
           >
             <Flag iso={selectedIso} />
+            <Text sx={{ paddingLeft: '4px' }}>+{selectedCountryOption.country.code}</Text>
           </SelectButton>
         </Flex>
         <SelectOptionList
@@ -95,10 +140,11 @@ export const PhoneInput = (props: PhoneInputProps) => {
       </Select>
       <Input
         value={formattedPhoneNumber}
-        onChange={el => setPhoneNumber(el.target.value)}
+        onPaste={handlePaste}
+        onChange={handlePhoneNumberChange}
         maxLength={25}
         type='tel'
-        sx={theme => ({ paddingLeft: theme.space.$20 })}
+        sx={theme => ({ paddingLeft: `calc(${theme.space.$20} + ${dynamicPadding}px)` })}
         ref={phoneInputRef}
         {...rest}
       />
--- a/packages/clerk-js/src/ui/elements/PhoneInput/countryCodeData.ts
+++ b/packages/clerk-js/src/ui/elements/PhoneInput/countryCodeData.ts
@@ -7,7 +7,7 @@ export type PhonePattern = CountryCodeData[number][3] | '';
 type PositionalCountryType = [CountryName, CountryIso, DialingCode, PhonePattern?];
 
 const data = [
-  ['United States', 'us', '1', '(...) ...-....'],
+  ['United States', 'us', '1', '(...) ...-....', 100],
   ['United Kingdom', 'gb', '44', '.... ......'],
   ['India', 'in', '91', '.....-.....'],
   ['Canada', 'ca', '1', '(...) ...-....'],
@@ -541,13 +541,21 @@ export interface CountryEntry {
   iso: CountryIso;
   code: DialingCode;
   pattern: PhonePattern;
+  priority: number;
 }
 
-const createEntry = ([name, iso, code, pattern = '' as any]: CountryCodeData[number]): CountryEntry => ({
+const createEntry = ([
+  name,
+  iso,
+  code,
+  pattern = '' as any,
+  priority = 0 as any,
+]: CountryCodeData[number]): CountryEntry => ({
   name,
   iso,
   code,
   pattern,
+  priority,
 });
 
 export type IsoToCountryMapType = ReadonlyMap<CountryIso, CountryEntry>;
--- a/packages/clerk-js/src/ui/elements/PhoneInput/useFormattedPhoneNumber.ts
+++ b/packages/clerk-js/src/ui/elements/PhoneInput/useFormattedPhoneNumber.ts
@@ -31,8 +31,8 @@ export const useFormattedPhoneNumber = (props: UseFormattedPhoneNumberProps = {}
   }, [selectedIso, phoneNum]);
 
   const setPhoneNumber = React.useCallback(
-    (str: string, countryCode?: CountryIso) => {
-      setPhoneNum(format(str, countryCode || selectedIso));
+    (str: string, iso?: CountryIso) => {
+      setPhoneNum(format(str, iso || selectedIso));
     },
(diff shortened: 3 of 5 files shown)


packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
```diff
@@ -37,9 +42,48 @@ export const PhoneInput = (props: PhoneInputProps) => {
...
+  const dynamicPadding = selectedCountryOption.country.code.length * 5; // this to calculate the padding of the input field depending the length of country code
 
   React.useEffect(callOnChangeProp, [cleanPhoneNumber]);
 
+  const handlePaste = (e: ClipboardEvent<HTMLInputElement>) => {
+    const inputValue = e.clipboardData.getData('text');
+    const testPattern = /\+/g;
+
+    if (testPattern.test(inputValue)) {
+      setHasBeenPasted(true);
```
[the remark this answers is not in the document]
desiprisg: If I understand correctly, you use this state to avoid handling the change twice and thus duplicating the input, correct?

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
```diff
@@ -37,9 +42,48 @@ export const PhoneInput = (props: PhoneInputProps) => {
...
+      }
+
+      if (selectedCountry) {
+        setSelectedIso(selectedCountry?.country.iso);
+        setPhoneNumber(result, selectedCountry?.country.iso);
+        return;
+      }
+    }
+    setSelectedIso(selectedIso);
+    setPhoneNumber(inputValue);
```
desiprisg, line 76: 🚫 I'm pretty sure these two lines need to be removed since normal pastes(without the country code) are being duplicated. If removed, the `setPhoneNumber` will run in the `onChange` handler since the `hasBeenPasted` flag would be false. Not sure why the `setSelectedIso` was added here. Am I missing something?

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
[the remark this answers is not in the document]
raptisj: @nikosdouvlis if I paste a value it will also trigger the onChange resulting in an unexpected state change. For example, the +123123123123(US code) would result in (+1) in select, (231)231 2312312. The last two digits(12 in our case) was added from the onChange. Hope this makes it clearer.

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
raptisj: @desiprisg The `setPhoneNumber` is needed here if the phone number(with county code) does not correspond to a valid county code. So you just paste it as is and let the API inform the user that this is not valid.

The `setSelectedIso` seems to be redundant, you are right🙃

desiprisg: Still, when pasting something that does not satisfy the regular expression, we get a double paste right now!

Can you give me an example input to paste to see the above behaviour?

raptisj: @desiprisg why would we get a double-paste?

For example this is not valid `+699090909090`

desiprisg: We get a double-paste when pasting strings like `123`, strings that fail the test pattern! Does that not happen on your machine?

raptisj: @desiprisg oh I see what you mean. I forgot to move the `setPhoneNumber` within the test pattern check. Now it does not produce a double check

raptisj: @desiprisg 

https://user-images.githubusercontent.com/23478420/198993933-662495f4-13d0-486c-a0c6-778678c25899.mov

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
```diff
@@ -37,9 +42,48 @@ export const PhoneInput = (props: PhoneInputProps) => {
   const selectedCountryOption = React.useMemo(() => {
     return countryOptions.find(o => o.country.iso === selectedIso) || countryOptions[0];
   }, [selectedIso]);
+  const dynamicPadding = selectedCountryOption.country.code.length * 5; // this to calculate the padding of the input field depending the length of country code
```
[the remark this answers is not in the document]
raptisj: @nikosdouvlis this is not for the country code dropdown but for the input value. We want to avoid this. Currently it has a fixed padding which is not good for such cases

Screen Shot 2022-10-31 at 1 47 15 PM

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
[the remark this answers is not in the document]
SokratisVidros: Is it possible to handle the value parsing and setting in the onChange event or does that create a clunky experience?

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,26 @@ export function stringToFormattedPhoneString(str: string): string {
   const parsed = parsePhoneString(str);
   return `+${parsed.code} ${formatPhoneNumber(parsed.number, parsed.pattern, parsed.code)}`;
 }
+
+type CountryData = {
+  searchTerm: string;
+  country: CountryEntry;
+};
+
+export function getValidCountry(value: string, options: CountryData[]) {
```
SokratisVidros, line 125: 🙃 It's worth adding a comment for this best-effort algorithm to find the longest country code. Alternatively, you can rename this function to getLongestValidCountryCode to make it more descriptive.

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,26 @@ export function stringToFormattedPhoneString(str: string): string {
...
+  let result = '';
+  let selectedCountry = null;
+  for (let i = 5; i > 1; i--) {
+    const code = value.slice(1, i);
+
+    result = value.slice(i, value.length);
+    selectedCountry = options.find(o => o.country.code === code);
```
SokratisVidros, line 132:
```suggestion
    selectedCountryCode = options.find(o => o.country.code === code);
```

SokratisVidros (approved):

packages/clerk-js/src/ui/utils/phoneUtils.test.ts
```diff
@@ -198,4 +199,32 @@ describe('phoneUtils', () => {
       expect(res).toBe('lc');
     });
   });
+
+  describe('getLongestValidCountryCode(value, options)', () => {
+    it('finds valid country code and returns phone number value without country code', () => {
+      const countries = [
```
nikosdouvlis, line 205: Do we need to mock this? Can we use the actual country map instead for better coverage?

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
nikosdouvlis: @raptisj @desiprisg hey folks, is this one resolved?

packages/clerk-js/src/ui/utils/phoneUtils.test.ts
raptisj: @nikos regarding the first point, the function expects a list of countries as shown above in order to iterate over them, thus the mocking. For the latter point, we use a `find` array method for it will return the first it finds in the list. In the above example `us` comes before `do`.

We have to start with 4 digits and then go down because the opposite would result in returning only single digits. It would return `us` every time and never reach `bs` (see above)

nikosdouvlis: For the second issue, the current algo will probably return an unwanted result for all us numbers starting from, lets say, "12" (1 is the country code, 2 is the first phone digit).

raptisj: @nikosdouvlis I can confirm this works as expected `+12064563059`. returns `us`

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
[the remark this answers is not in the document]
nikosdouvlis: Can we remove the explicitly set state by using `e.preventDefault()` as shown [here](https://github.com/clerkinc/javascript/blob/2bbb13022e4a9d3cbd62f06ad2b06b229f05583e/packages/clerk-js/src/ui/elements/CodeControl.tsx#L114-L113)?

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
```diff
@@ -37,9 +43,36 @@ export const PhoneInput = (props: PhoneInputProps) => {
...
   React.useEffect(callOnChangeProp, [cleanPhoneNumber]);
 
+  const handlePaste = (e: ClipboardEvent<HTMLInputElement>) => {
+    const inputValue = e.clipboardData.getData('text');
+    const pattern = /\+/g;
+
+    if (pattern.test(inputValue)) {
+      setHasBeenPasted(true);
+
+      const { result, selectedCountryCode } = getLongestValidCountryCode(inputValue, countryOptions);
```
nikosdouvlis, line 57:
```suggestion
      const { result, selectedCountryCode } = getLongestValidCountryCode(inputValue);
```

We shouldn't use `countryOptions` here, because this is an array with options for the UI select element. This method purely deals with country data, so it should not be aware of ui-related concerns as this options array.

We should remove this param altogether, and instead simply use the country map within `phoneUtils`, as we do for the other utilities.

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,31 @@ export function stringToFormattedPhoneString(str: string): string {
   const parsed = parsePhoneString(str);
   return `+${parsed.code} ${formatPhoneNumber(parsed.number, parsed.pattern, parsed.code)}`;
 }
+
+type CountryData = {
+  searchTerm: string;
```
nikosdouvlis, line 121: It looks like this `searchTerm` was only added because you wanted to use `countyOptions` from the `PhoneInput.tsx`. As I commented above, lets not leak any UI concerns inside these helpers. Instead, lets remove CountryData altogether and use the country map directly here

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,31 @@ export function stringToFormattedPhoneString(str: string): string {
...
+const priorityCodes: string[] = ['1'];
+export function getLongestValidCountryCode(value: string, options: CountryData[]) {
+  let result = '';
+  let selectedCountryCode = null;
+  const numberValue = value.replace('+', '');
+
+  for (let i = 4; i > 0; i--) {
+    const code = numberValue.slice(0, i);
+    const priority = numberValue.slice(0, 1);
+    const hasPriority = priorityCodes.includes(priority);
```
nikosdouvlis, line 134: I think we should move this logic in the country [map itself](https://github.com/clerkinc/javascript/blob/2bbb13022e4a9d3cbd62f06ad2b06b229f05583e/packages/clerk-js/src/ui/elements/PhoneInput/countryCodeData.ts#L9) by adding a numeric priority as the fourth element in the countries that need it so the country map remains the single source of truth

You can then use the maps to check if a country has priority or not (and if both have, then prefer the country with the highest priority)

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,31 @@ export function stringToFormattedPhoneString(str: string): string {
...
+
+    result = numberValue.slice(hasPriority ? 1 : i, value.length);
+    selectedCountryCode = options.find(o => o.country.code === (hasPriority ? priority : code));
+
+    if (selectedCountryCode) {
```
nikosdouvlis, line 139: If a selected code is found, instead of using `break` you can simply return early 🙃

nikosdouvlis (changes requested):

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
desiprisg: Yes, seems fine in the vid. Sorry it took so long, I missed the notification. We can resolve this one. I'll make another pass anyway!

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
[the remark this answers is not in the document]
nikosdouvlis: @gkats any updates on this?

packages/clerk-js/src/ui/elements/PhoneInput/countryCodeData.ts
```diff
@@ -541,13 +541,21 @@ export interface CountryEntry {
...
+  priority: number;
 }
 
-const createEntry = ([name, iso, code, pattern = '' as any]: CountryCodeData[number]): CountryEntry => ({
+const createEntry = ([
+  name,
+  iso,
+  code,
+  pattern = '' as any,
+  priority = 0 as any,
```
nikosdouvlis, line 552:
```suggestion
  priority = 0,
```

packages/clerk-js/src/ui/elements/PhoneInput/useFormattedPhoneNumber.ts
```diff
@@ -31,8 +31,8 @@ export const useFormattedPhoneNumber = (props: UseFormattedPhoneNumberProps = {}
   }, [selectedIso, phoneNum]);
 
   const setPhoneNumber = React.useCallback(
-    (str: string) => {
-      setPhoneNum(format(str, selectedIso));
+    (str: string, countryCode?: CountryIso) => {
```
nikosdouvlis, line 34:
```suggestion
    (str: string, iso?: CountryIso) => {
```

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,34 @@ export function stringToFormattedPhoneString(str: string): string {
   const parsed = parsePhoneString(str);
   return `+${parsed.code} ${formatPhoneNumber(parsed.number, parsed.pattern, parsed.code)}`;
 }
+
+export const sortByPriority = (a: CountryEntry, b: CountryEntry) => {
```
nikosdouvlis, line 120: 🙃  nitpicky
```suggestion
export const byPriority = (a: CountryEntry, b: CountryEntry) => {
```

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,34 @@ export function stringToFormattedPhoneString(str: string): string {
...
+  }
+  if (a.priority < b.priority) {
+    return 1;
+  }
+  return 0;
+};
+
+export function getLongestValidCountryCode(value: string) {
+  let result = '';
+  let countries = null;
```
nikosdouvlis, line 132: Lets initialise this inside the for block directly, since its not used by anything else right now

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,34 @@ export function stringToFormattedPhoneString(str: string): string {
...
+  const numberValue = value.replace('+', '');
+  const countryGroup = [];
```
nikosdouvlis, line 134: I think we can rename this to `matchingCountries` 
```suggestion
  const countryGroup = [];
```

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,34 @@ export function stringToFormattedPhoneString(str: string): string {
...
+
+  for (let i = 4; i > 0; i--) {
+    const code = numberValue.slice(0, i);
+    countries = [...IsoToCountryMap.values()].filter(o => o.code === code);
+
+    if (countries.length) {
+      countryGroup.push(countries);
```
nikosdouvlis, line 141: 🙃 Another, more idiomatic way to write this would be
```suggestion
      countryGroup.push(...countries);
```
that way you also avoid using `flat()` below

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,34 @@ export function stringToFormattedPhoneString(str: string): string {
...
+    }
+  }
+
+  const selectedCountry = countryGroup.flat().sort(sortByPriority)[0];
+  result = numberValue.slice(selectedCountry?.code.length, value.length);
+
+  return { result, selectedCountry };
```
nikosdouvlis, line 148: Whats the difference between `result` and `selectedCountry`? Can we make result more semantic, otherwise we'd have to look at the functions implementation everytime

raptisj: @nikos this is the phone number value without the `+`. I will rename

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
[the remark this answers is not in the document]
desiprisg: Let's try and fix this before merging. It's more prominent when adding a phone number in `UserProfile`

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,34 @@ export function stringToFormattedPhoneString(str: string): string {
...
+  return 0;
+};
+
+export function getLongestValidCountryCode(value: string) {
+  let phoneNumberValue = '';
+  const numberValue = value.replace('+', '');
+  const matchingCountries = [];
+
+  for (let i = 4; i > 0; i--) {
+    let countries = null;
```
desiprisg, line 136: ❓Why do we need this line?

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,34 @@ export function stringToFormattedPhoneString(str: string): string {
...
+  }
+  return 0;
+};
+
+export function getLongestValidCountryCode(value: string) {
+  let phoneNumberValue = '';
+  const numberValue = value.replace('+', '');
+  const matchingCountries = [];
+
+  for (let i = 4; i > 0; i--) {
```
desiprisg, line 135: 🙃 Maybe turn the `4` into a `const` outside the function for clarification?

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,34 @@ export function stringToFormattedPhoneString(str: string): string {
...
 }
+
+export const byPriority = (a: CountryEntry, b: CountryEntry) => {
+  if (a.priority > b.priority) {
+    return -1;
+  }
+  if (a.priority < b.priority) {
+    return 1;
+  }
+  return 0;
```
desiprisg, line 127:
```suggestion
  return b.priority - a.priority;
```

packages/clerk-js/src/ui/utils/phoneUtils.ts
```diff
@@ -115,3 +116,34 @@ export function stringToFormattedPhoneString(str: string): string {
...
+
+export function getLongestValidCountryCode(value: string) {
+  let phoneNumberValue = '';
+  const numberValue = value.replace('+', '');
+  const matchingCountries = [];
+
+  for (let i = 4; i > 0; i--) {
+    let countries = null;
+    const code = numberValue.slice(0, i);
+    countries = [...IsoToCountryMap.values()].filter(o => o.code === code);
```
nikosdouvlis, line 138:
```suggestion
    const countries = [...IsoToCountryMap.values()].filter(o => o.code === code);
```

nikosdouvlis (approved):

PR #493, opened 2022-11-02
feat(clerk-js): Improve Popover behaviour

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

### Issue
When having nested popovers, clicking the second popover will result in all popovers being closed.

### Solution
This PR attempts to fix this issue by utilizing more of the floating-ui library and take advantage of its focus-trap, focus-return mechanisms and its ability to allow or block dismissal bubbling in each popover.

I have introduced a new `<Popover>` component which wraps the `children` with the appropriate components needed by the `floating-ui` library. The new component also receives an `isOpen` props and hides `children` accordingly. It also traps the focus into the Popover and returns the focus once it is closed. 

The `usePopover` hook now receives the `bubbles?: boolean` prop which determined if the Popover dismissal should stop or continue in the popover. This makes more in sense in modals than other Popovers.

The `usePopover` hook now subscribes the Popover to the `FloatingTree` context and returns a `nodeId` which has to be passed to the `Popover` component. This is necessary for the correct nested Popover behaviour.

In this PR, I have integrated these changes to every popover used in our codebase. If the popover does not need to be portalled, we pass `portal={false}` to the `<Popover/>` component.

### DIFF
--- a/packages/clerk-js/src/ui/elements/Popover.tsx
+++ b/packages/clerk-js/src/ui/elements/Popover.tsx
@@ -1,0 +1,54 @@
+import {
+  FloatingContext,
+  FloatingFocusManager,
+  FloatingNode,
+  FloatingPortal,
+  ReferenceType,
+} from '@floating-ui/react-dom-interactions';
+import React from 'react';
+import { PropsWithChildren } from 'react';
+
+type PopoverProps = PropsWithChildren<{
+  context: FloatingContext<ReferenceType>;
+  nodeId: string;
+  isOpen?: boolean;
+  initialFocus?: number | React.MutableRefObject<HTMLElement | null>;
+  order?: Array<'reference' | 'floating' | 'content'>;
+  portal?: boolean;
+}>;
+
+export const Popover = (props: PopoverProps) => {
+  const { context, initialFocus, order = ['reference', 'content'], nodeId, isOpen, portal = true, children } = props;
+
+  if (portal) {
+    return (
+      <FloatingNode id={nodeId}>
+        <FloatingPortal>
+          {isOpen && (
+            <FloatingFocusManager
+              context={context}
+              initialFocus={initialFocus}
+              order={order}
+            >
+              <>{children}</>
+            </FloatingFocusManager>
+          )}
+        </FloatingPortal>
+      </FloatingNode>
+    );
+  }
+
+  return (
+    <FloatingNode id={nodeId}>
+      {isOpen && (
+        <FloatingFocusManager
+          context={context}
+          initialFocus={initialFocus}
+          order={order}
+        >
+          <>{children}</>
+        </FloatingFocusManager>
+      )}
+    </FloatingNode>
+  );
+};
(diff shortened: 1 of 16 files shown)


nikosdouvlis:
**Temporarily** closing this one, will be re-opened when we handle https://www.notion.so/clerkdev/Modals-and-popovers-should-trap-focus-for-improved-accessibility-df47f40c1c8e4c569ccf461b3526ef07

packages/clerk-js/src/ui/elements/Popover.tsx
```diff
@@ -0,0 +1,54 @@
...
+              initialFocus={initialFocus}
+              order={order}
+            >
+              <>{children}</>
+            </FloatingFocusManager>
+          )}
+        </FloatingPortal>
+      </FloatingNode>
+    );
+  } else {
```
nikosdouvlis, line 39: 🙃 We don't need the else here

nikosdouvlis (approved):
Nice one

packages/clerk-js/src/ui/elements/contexts/FloatingTreeContext.tsx
```diff
@@ -0,0 +1,21 @@
...
+  return (props: T) => {
+    const parentId = useFloatingParentNodeId();
+    if (parentId == null) {
+      return (
+        <FloatingTree>
+          {/* @ts-expect-error */}
+          <Component {...props} />
+        </FloatingTree>
+      );
+    } else {
```
nikosdouvlis, line 14: 👁️

packages/clerk-js/src/ui/hooks/usePopover.ts
```diff
@@ -22,14 +23,36 @@ export type UsePopoverReturn = ReturnType<typeof usePopover>;
...
     whileElementsMounted: props.autoUpdate === false ? undefined : autoUpdate,
     placement: props.placement || 'bottom-start',
     middleware: [offset(props.offset || 6), flip(), shift()],
   });
 
+  useDismiss(context, { bubbles });
+
+  useEffect(() => {
+    const handleFocus = (e: FocusEvent) => {
+      if (e.relatedTarget === null) {
```
nikosdouvlis, line 40: ❓ Is there a reason we're explicitly checking for === null instead of !e.relatedTarget?

desiprisg: Don't think so. Will change it.

PR #500, opened 2022-11-03
fix(clerk-js): Disable scroll to top for OrganizationProfile navbar & correctly show active tab

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Clicking the navbar tabs causes the page to scroll to the active element's id. This is unnecessary in the OrganizationProfile. A new `scroll:boolean` property has been added to the `NavbarRoute` to determine if navigation to that route should cause scrolling to the element's id.

When navigating to a new route without clicking the navbar tabs, the active tab is not correctly recognized.


nikosdouvlis:
Im wondering if there's a simpler way to do this.

What would happen if we removed these ids?
- https://github.com/clerkinc/javascript/blob/38f24c059dd08110447f7b0ef8b246fefb031211/packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationMembers.tsx#L44
- https://github.com/clerkinc/javascript/blob/32d9a6f11fa8122fe0a768830416dc3e750542e6/packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationSettings.tsx#L23

desiprisg:
@nikosdouvlis You're right. I thought we needed those for the styling of the active tab as well, but it seems I misunderstood. Will simplify and we can merge.

nikosdouvlis (approved):
Nice!

PR #510, opened 2022-11-07
fix(clerk-js): Set navbar active id based on path

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Active navbar id is not correctly set when navigating between routes without clicking the navbar buttons. This PR makes sure that the active id is set on path changes.

Also reobserve navbar sections if the route changes, so that the active id changes based on which section is visible.

### DIFF
--- a/packages/clerk-js/src/ui/elements/Navbar.tsx
+++ b/packages/clerk-js/src/ui/elements/Navbar.tsx
@@ -71,47 +71,51 @@ export const NavBar = (props: NavBarProps) => {
     }
   };
 
-  useSafeLayoutEffect(function selectNavItemBasedOnVisibleSection() {
-    const mountObservers = () => {
-      const ids = routes.map(r => r.id);
-      const sectionElements = ids
-        .map(getSectionId)
-        .map(id => contentRef.current?.querySelector(id))
-        .filter(s => s);
-
-      if (sectionElements.length === 0) {
-        return false;
-      }
-
+  useSafeLayoutEffect(
+    function selectNavItemBasedOnVisibleSection() {
       const callback: IntersectionObserverCallback = entries => {
         for (const entry of entries) {
           const id = entry.target?.id?.split('section-')[1];
           if (entry.isIntersecting && id) {
-            return setActiveId(id as unknown as any);
+            return setActiveId(id);
           }
         }
       };
-
       const observer = new IntersectionObserver(callback, { root: contentRef.current, threshold: 1 });
-      sectionElements.forEach(section => section && observer.observe(section));
-      return true;
-    };
 
-    const intervalId = setInterval(() => {
-      if (mountObservers()) {
-        clearInterval(intervalId);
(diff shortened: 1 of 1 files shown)


packages/clerk-js/src/ui/elements/Navbar.tsx
```diff
@@ -69,40 +71,57 @@ export const NavBar = (props: NavBarProps) => {
...
+    [router],
+  );
+
+  useEffect(() => {
+    routes.every(route => {
+      const isRoot = router.currentPath === router.fullPath && route.path === '/';
+      const matchesPath = router.matches(route.path);
+      if (isRoot || matchesPath) {
+        setActiveId(route.id);
+        setActiveId(route.id);
```
anagstef, line 120: ❓ Why is this called twice here?

desiprisg: My bad! Must have left this when refactoring.

anagstef (approved):

packages/clerk-js/src/ui/elements/Navbar.tsx
```diff
@@ -69,40 +71,56 @@ export const NavBar = (props: NavBarProps) => {
...
+      return () => {
+        observer.disconnect();
+      };
+    },
+    [router],
+  );
+
+  useEffect(() => {
+    routes.every(route => {
+      const isRoot = router.currentPath === router.fullPath && route.path === '/';
```
nikosdouvlis, line 116: Lets say we do this. 
```suggestion
      const isRoot = route.path === '/';
```
Which scenarios break the above?

desiprisg: We are iterating through the navbar routes, so the `route.path` with be `'/'` at some point but we also need to make sure the current path is also the root. If we don't make both checks, the active id will be set as the root at all times, even if the `router.path` is not on the root.

nikosdouvlis: Ah I see. I misread `route` as `router` and the above snippet didn't make sense; sorry about that.

This is not related to this PR, however, it's not clearly apparent what the difference between these two is `router.currentPath === router.fullPath`. Noted so we can discuss later. Thanks!

nikosdouvlis (approved):

PR #512, opened 2022-11-07
fix(clerk-js): Delete user profile image

raptisj:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected
 
- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

Delete user profile image

### DIFF
--- a/packages/clerk-js/src/core/resources/User.ts
+++ b/packages/clerk-js/src/core/resources/User.ts
@@ -210,7 +210,7 @@ export class User extends BaseResource implements UserResource {
   setProfileImage = (params: SetProfileImageParams): Promise<ImageResource> => {
     const { file } = params || {};
 
-    if (!file) {
+    if (file === null) {
       return Image.delete(`${this.path()}/profile_image`);
     }
     return Image.create(`${this.path()}/profile_image`, {
--- a/packages/clerk-js/src/core/resources/User.ts
+++ b/packages/clerk-js/src/core/resources/User.ts
@@ -209,6 +209,10 @@ export class User extends BaseResource implements UserResource {
 
   setProfileImage = (params: SetProfileImageParams): Promise<ImageResource> => {
     const { file } = params || {};
+
+    if (!file) {
+      return Image.delete(`${this.path()}/profile_image`);
+    }
     return Image.create(`${this.path()}/profile_image`, {
       file,
     });
--- a/packages/clerk-js/src/ui/components/UserProfile/UserProfileAvatarUploader.tsx
+++ b/packages/clerk-js/src/ui/components/UserProfile/UserProfileAvatarUploader.tsx
@@ -11,6 +11,7 @@ export const UserProfileAvatarUploader = (
     <AvatarUploader
       {...rest}
       title={localizationKeys('userProfile.profilePage.imageFormTitle')}
+      hasImageUrl={!new RegExp('gravatar').test(user.profileImageUrl || '')}
       avatarPreview={
         <UserAvatar
           size={theme => theme.sizes.$11}
--- a/packages/clerk-js/src/ui/components/UserProfile/UserProfileAvatarUploader.tsx
+++ b/packages/clerk-js/src/ui/components/UserProfile/UserProfileAvatarUploader.tsx
@@ -11,7 +11,7 @@ export const UserProfileAvatarUploader = (
     <AvatarUploader
       {...rest}
       title={localizationKeys('userProfile.profilePage.imageFormTitle')}
-      hasImageUrl={!(user.profileImageUrl || '').includes('gravatar')}
+      hasDefaultImageUrl={(user.profileImageUrl || '').includes('gravatar')}
       avatarPreview={
         <UserAvatar
           size={theme => theme.sizes.$11}
--- a/packages/clerk-js/src/ui/components/UserProfile/UserProfileAvatarUploader.tsx
+++ b/packages/clerk-js/src/ui/components/UserProfile/UserProfileAvatarUploader.tsx
@@ -11,7 +11,7 @@ export const UserProfileAvatarUploader = (
     <AvatarUploader
       {...rest}
       title={localizationKeys('userProfile.profilePage.imageFormTitle')}
-      hasImageUrl={!new RegExp('gravatar').test(user.profileImageUrl || '')}
+      hasImageUrl={!(user.profileImageUrl || '').includes('gravatar')}
       avatarPreview={
         <UserAvatar
           size={theme => theme.sizes.$11}
--- a/packages/clerk-js/src/ui/elements/AvatarUploader.tsx
+++ b/packages/clerk-js/src/ui/elements/AvatarUploader.tsx
@@ -9,7 +9,7 @@ export type AvatarUploaderProps = {
   title: LocalizationKey;
   avatarPreview: React.ReactElement;
   onAvatarChange: (file: File | null) => Promise<unknown>;
-  hasImageUrl?: boolean;
+  hasDefaultImageUrl?: boolean;
 };
 
 export const fileToBase64 = (file: File): Promise<string> => {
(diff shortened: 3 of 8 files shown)


packages/clerk-js/src/core/resources/User.ts
```diff
@@ -209,6 +209,10 @@ export class User extends BaseResource implements UserResource {
 
   setProfileImage = (params: SetProfileImageParams): Promise<ImageResource> => {
     const { file } = params || {};
+
+    if (!file) {
```
gkats, line 213: :question: Is this check enough? Should we explicitly check that the `file` parameter is set to `null` before we trigger a DELETE request?

SokratisVidros (approved):

packages/clerk-js/src/ui/localization/defaultEnglishResource.ts
```diff
@@ -275,6 +275,7 @@ export const defaultResource: DeepRequired<LocalizationResource> = {
       title: 'Update profile',
       imageFormTitle: 'Profile image',
       imageFormSubtitle: 'Upload image',
+      destructiveActionSubtitle: 'Delete image',
```
gkats, line 278: :question: Should this say "Delete" or "Remove"?

SokratisVidros: Remove sounds better

packages/clerk-js/src/core/resources/User.ts
raptisj: @gkats what edge case do you have in mind?

packages/clerk-js/src/ui/localization/defaultEnglishResource.ts
raptisj: @SokratisVidros , @nikosdouvlis what do you think?

desiprisg: Since the image can't be retrieved, I'd say delete is better here.

packages/clerk-js/src/core/resources/User.ts
gkats: I don't have an edge case in mind. It's just an API choice.

```js
// No file in the parameters, is this intentional?
user.setProfileImage();

// vs

// File is passed and set to null
user.setProfileImage({ file: null });
```

I'm not sure how TS will behave with the above, but in plain JS, undefined and null will be different.

cc @SokratisVidros

desiprisg: Typescript should complain but not everyone will be using typescript, so maybe it should indeed be explicit.

nikosdouvlis: We need to be explicit about this, otherwise any improper call invocation would delete the image. Lets check for `=== null`

packages/clerk-js/src/core/resources/User.ts
```diff
@@ -209,6 +209,10 @@ export class User extends BaseResource implements UserResource {
 
   setProfileImage = (params: SetProfileImageParams): Promise<ImageResource> => {
     const { file } = params || {};
+
+    if (file === null) {
+      return Image.delete(`${this.path()}/profile_image`);
+    }
     return Image.create(`${this.path()}/profile_image`, {
```
nikosdouvlis, line 216: Not part of this PR, but its a little unfortunate that the user knows about the image endpoint 🙃

packages/clerk-js/src/ui/components/UserProfile/UserProfileAvatarUploader.tsx
```diff
@@ -11,6 +11,7 @@ export const UserProfileAvatarUploader = (
     <AvatarUploader
       {...rest}
       title={localizationKeys('userProfile.profilePage.imageFormTitle')}
+      hasImageUrl={!new RegExp('gravatar').test(user.profileImageUrl || '')}
```
nikosdouvlis, line 14: Would this be enough?
```suggestion
      hasImageUrl={(user.profileImageUrl || '').includes('gravatar')}
```

packages/clerk-js/src/ui/elements/AvatarUploader.tsx
```diff
@@ -8,7 +8,8 @@ import { FileDropArea } from './FileDropArea';
 export type AvatarUploaderProps = {
   title: LocalizationKey;
   avatarPreview: React.ReactElement;
-  onAvatarChange: (file: File) => Promise<unknown>;
+  onAvatarChange: (file: File | null) => Promise<unknown>;
+  hasImageUrl?: boolean;
```
nikosdouvlis, line 12: I think that the `hasImageUrl` prop doesn't really communicate its intent. What do you think about replacing this with an `onAvatarRemove?: () => Promise<unknown>` handler instead?

packages/clerk-js/src/ui/elements/AvatarUploader.tsx
```diff
@@ -70,6 +72,18 @@ export const AvatarUploader = (props: AvatarUploaderProps) => {
...
             />
+
+            {hasImageUrl && !showUpload && (
+              <Button
+                localizationKey={localizationKeys('userProfile.profilePage.destructiveActionSubtitle')}
+                isDisabled={card.isLoading}
+                sx={t => ({ color: t.colors.$danger500 })}
+                variant='link'
+                onClick={() => {
+                  void onAvatarChange(null);
```
nikosdouvlis, line 83: Lets also disable the Delete button while the request runs

packages/types/src/localization.ts
```diff
@@ -282,6 +282,7 @@ type _LocalizationResource = {
       title: LocalizationValue;
       imageFormTitle: LocalizationValue;
       imageFormSubtitle: LocalizationValue;
+      destructiveActionSubtitle: LocalizationValue;
```
nikosdouvlis, line 285: Do you think we can group the delete button key with the rest, by prefixing it with `imageForm`? eg: 
```suggestion
      imageFormDestructiveAction: LocalizationValue;
```

packages/clerk-js/src/core/resources/User.ts
SokratisVidros: ☝️

packages/clerk-js/src/ui/localization/defaultEnglishResource.ts
SokratisVidros: Let's do Remove

packages/clerk-js/src/core/resources/User.ts
SokratisVidros: At least it's nested 😌

packages/clerk-js/src/ui/elements/AvatarUploader.tsx
raptisj: @nikosdouvlis we want to check if avatar is the defualt gravatar one or a custom one. Maybe `hasDefaultImage`? I don't get how `onAvatarRemove` would work in this case

packages/types/src/localization.ts
raptisj: @nikosdouvlis probably `imageFormDestructiveActionSubtitle` is better since it matches the convention

packages/clerk-js/src/ui/components/UserProfile/UserProfileAvatarUploader.tsx
raptisj: @nikosdouvlis yes this will do 🙃

SokratisVidros (approved):

PR #517, opened 2022-11-09
NextJS 13 support

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

- Use react 18 across all packages
- Update the testing suite and testing dependencies to support react 18
- Bump nextjs to 13 in clerk/nextjs and playground
- Transfer the `clerk/nextjs/app-beta` module nextjs changes from the `next` branch to here so we can merge into main
- Add the org components to `clerk/nextjs/app-beta`
- Cleanup the request parsing logic in https://github.com/clerkinc/javascript/pull/517/commits/f5eae1959b7e9e8dae3974ce56d0394abda16f5a#diff-1284816a75126e44df1a317423de14d3dc320f5c092012f27cdd249cd44c9347L1 by introducing `isNextRequest` instead of using `try/catch` clauses for flow control
- Enable and use the experimental `app-dir` in the playground app

Next items are coming in a follow up PR


packages/clerk-js/package.json
```diff
@@ -66,21 +66,17 @@
...
     "@types/jsonwebtoken": "^8",
     "@types/qs": "^6.9.3",
-    "@types/react": "^17.0.39",
-    "@types/react-dom": "^17.0.11",
-    "@types/react-test-renderer": "^17",
+    "@types/react": "*",
+    "@types/react-dom": "*",
     "@types/sass": "^1",
-    "@types/testing-library__jest-dom": "^5",
-    "@types/testing-library__react-hooks": "^3",
```
anagstef, line 83: ❓ Are these deprecated?

anagstef (approved):

packages/clerk-js/package.json
nikosdouvlis: Not deprecated per se, however the latest testing-library version has been restructured to include re-export these lower-level libs and types, so I'm fairly confident we no longer need these.

In general, global `@types/*` can cause issues (conflicting type declarations or duplicate declarations) in monorepos when packages are using different versions or if they depend on packages that use conflicting `@type/`. For instance, we've fought build issues in the past caused by conflicts between different versions of node and dom types introduced by dependencies in the global space.

In general, the less global `@types` we use, the better

desiprisg (approved):

packages/clerk-js/package.json
```diff
@@ -118,7 +115,7 @@
     "files": [
       {
         "path": "./dist/clerk.browser.js",
-        "maxSize": "169kB"
+        "maxSize": "175kB"
```
SokratisVidros, line 118: ❓ Is the size increase caused by React 18?

SokratisVidros (commented):
@nikosdouvlis What about Gatsby? Does it need any package.json changes for React 18?

packages/clerk-js/package.json
nikosdouvlis: Yes, but its a +1kb change. I just added a small buffer to bundlewatch on top of the actual increase

nikosdouvlis:
> @nikosdouvlis What about Gatsby? Does it need any package.json changes for React 18?

Not right now, as Gatsby v4 does support react 18. However, we might need to explore gatsby 5 soon(ish)

PR #526, opened 2022-11-11
Auth redirect flow

raptisj:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

## [WIP] 
This PR introduces an improved way to handle redirects coming from outside providers.
This is particularly useful when an OAuth connect flow fails. Instead of redirecting back to the profile with a closed modal(due to page refresh) it will persist state from a URL param and bring the page to its previous state(e.x. Open modal in the connect Oauth providers page).

### DIFF
--- a/packages/clerk-js/src/ui/Components.tsx
+++ b/packages/clerk-js/src/ui/Components.tsx
@@ -137,11 +137,11 @@ const componentsControls = {} as ComponentControls;
 
 // TODO: move this elsewhere
 const componentNodes = Object.freeze({
-  signUp: 'signUpModal',
-  singIn: 'signInModal',
-  user: 'userProfileModal',
-  organization: 'organizationProfileModal',
-  createOrganization: 'createOrganizationModal',
+  SignUp: 'signUpModal',
+  SignIn: 'signInModal',
+  UserProfile: 'userProfileModal',
+  OrganizationProfile: 'organizationProfileModal',
+  CreateOrganization: 'createOrganizationModal',
 });
 
 const Components = (props: ComponentsProps) => {
@@ -158,19 +158,23 @@ const Components = (props: ComponentsProps) => {
   const { signInModal, signUpModal, userProfileModal, organizationProfileModal, createOrganizationModal, nodes } =
     state;
 
-  // TODO: rename these
-  const urlClerkState = getClerkQueryParam('__clerk_state') ?? '';
-  const decodedRedirectParams = urlClerkState ? JSON.parse(atob(urlClerkState)) : null;
+  const [urlState, setUrlState] = React.useState(null);
+
+  const readAndRemoveStateParam = () => {
+    const urlClerkState = getClerkQueryParam('__clerk_state') ?? '';
+    removeClerkQueryParam('__clerk_state');
+    return urlClerkState ? JSON.parse(atob(urlClerkState)) : null;
+  };
+
+  const decodedRedirectParams = readAndRemoveStateParam() ?? '';
 
   useSafeLayoutEffect(() => {
-    if (urlClerkState && decodedRedirectParams && decodedRedirectParams?.modal) {
+    if (decodedRedirectParams) {
+      setUrlState(decodedRedirectParams);
       setState(s => ({
         ...s,
-        [componentNodes[snakeToCamel(decodedRedirectParams.basePath)]]: {},
-        options: { ...s.options, startPath: decodedRedirectParams.fullPath },
+        [componentNodes[decodedRedirectParams.componentName]]: true,
       }));
-
-      removeClerkQueryParam('__clerk_state');
     }
 
     componentsControls.mountComponent = params => {
@@ -203,7 +207,8 @@ const Components = (props: ComponentsProps) => {
     };
 
     componentsControls.closeModal = name => {
-      setState(s => ({ ...s, [name + 'Modal']: null, options: { ...s.options, startPath: null } }));
+      setUrlState(null);
+      setState(s => ({ ...s, [name + 'Modal']: null }));
     };
 
     componentsControls.openModal = (name, props) => {
@@ -283,7 +288,8 @@ const Components = (props: ComponentsProps) => {
             <VirtualRouter
               preservedParams={PRESERVED_QUERYSTRING_PARAMS}
               onExternalNavigate={() => componentsControls.closeModal('userProfile')}
-              startPath={state.options?.startPath || '/user'}
+              // startPath={state.options?.startPath || '/user'}
+              startPath={`/user${urlState?.path || ''}`}
             >
               <UserProfileModal />
             </VirtualRouter>
--- a/packages/clerk-js/src/ui/Components.tsx
+++ b/packages/clerk-js/src/ui/Components.tsx
@@ -15,6 +15,7 @@ import { createRoot } from 'react-dom/client';
 
 import { PRESERVED_QUERYSTRING_PARAMS } from '../core/constants';
 import { clerkUIErrorDOMElementNotFound } from '../core/errors';
+import { buildURL, getClerkQueryParam, windowNavigate } from '../utils';
 import { CreateOrganization, CreateOrganizationModal } from './components/CreateOrganization';
 import { ImpersonationFab } from './components/ImpersonationFab';
 import { OrganizationProfile, OrganizationProfileModal } from './components/OrganizationProfile';
@@ -148,7 +149,60 @@ const Components = (props: ComponentsProps) => {
   const { signInModal, signUpModal, userProfileModal, organizationProfileModal, createOrganizationModal, nodes } =
     state;
 
+  const urlClerkState = getClerkQueryParam('__clerk_state') ?? '';
+
+  // TODO: this will be greatly refactored and moved elsewhere
+  const getState = () => {
+    const fullPath = urlClerkState?.split('__')[0] || '';
+    const basePath = fullPath?.split('/')[1] || '';
+    const isModal = Boolean(urlClerkState?.split('__')[1]?.split('=')[1]);
+    const url = new URL(window.location.href);
+    const params = new URLSearchParams(url.search);
+
+    // // Delete the foo parameter.
+    // params.delete('__clerk_state');
+    // window.history.pushState({}, '', url);``
+    // const po = new URL(window.location.href).searchParams.delete('__clerk_state');
+
+    console.log(params, 'params');
+    console.log(props, 'props');
+
+    // props.options.navigate(window.location.href)
+
+    return { isModal, fullPath, basePath };
+  };
+
+  const urlState = {
+    user: 'userProfileModal',
+  };
+
+  // const po = getState();
+  // console.log(window.location.href);
+
+  // useEffect(() => {
+  // const url = new URL(window.location.href);
+  // const searchParams = new URL(window.location.href).search;
+  // const params = new URLSearchParams(url.search);
+
+  // // Delete the foo parameter.
+  // // params.delete('__clerk_state');
+  // console.log(params, 'paramsa')
+  // console.log(searchParams, 'searchParams')
+  // console.log(url, 'url')
+  // windowNavigate(url);
+  // buildURL({ base: url, params }, { stringify: true });
+  // }, []);
+
   useSafeLayoutEffect(() => {
+    const { isModal, basePath } = getState();
+
+    if (urlClerkState && isModal) {
+      setState(s => ({ ...s, [urlState[basePath]]: {} }));
+
+      // Delete the foo parameter.
+      // params.delete('__clerk_state');
+    }
+
     componentsControls.mountComponent = params => {
       const { node, name, props, appearanceKey } = params;
       assertDOMElement(node);
@@ -259,7 +313,8 @@ const Components = (props: ComponentsProps) => {
             <VirtualRouter
               preservedParams={PRESERVED_QUERYSTRING_PARAMS}
               onExternalNavigate={() => componentsControls.closeModal('userProfile')}
-              startPath='/user'
+              // startPath={'/user'}
+              startPath={getState()?.fullPath || '/user'}
             >
               <UserProfileModal />
             </VirtualRouter>
--- a/packages/clerk-js/src/ui/Components.tsx
+++ b/packages/clerk-js/src/ui/Components.tsx
@@ -1,4 +1,4 @@
-import { createDeferredPromise } from '@clerk/shared';
+import { createDeferredPromise, snakeToCamel } from '@clerk/shared';
 import type {
   Appearance,
   Clerk,
@@ -15,7 +15,7 @@ import { createRoot } from 'react-dom/client';
 
 import { PRESERVED_QUERYSTRING_PARAMS } from '../core/constants';
 import { clerkUIErrorDOMElementNotFound } from '../core/errors';
-import { buildURL, getClerkQueryParam, windowNavigate } from '../utils';
+import { getClerkQueryParam, removeClerkQueryParam } from '../utils';
 import { CreateOrganization, CreateOrganizationModal } from './components/CreateOrganization';
 import { ImpersonationFab } from './components/ImpersonationFab';
 import { OrganizationProfile, OrganizationProfileModal } from './components/OrganizationProfile';
@@ -135,6 +135,15 @@ export type MountComponentRenderer = typeof mountComponentRenderer;
 
 const componentsControls = {} as ComponentControls;
 
+// TODO: move this elsewhere
+const componentNodes = Object.freeze({
+  signUp: 'signUpModal',
+  singIn: 'signInModal',
+  user: 'userProfileModal',
+  organization: 'organizationProfileModal',
+  createOrganization: 'createOrganizationModal',
+});
+
 const Components = (props: ComponentsProps) => {
   const [state, setState] = React.useState<ComponentsState>({
     appearance: props.options.appearance,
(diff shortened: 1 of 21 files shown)


packages/clerk-js/src/ui/Components.tsx
```diff
@@ -149,7 +160,21 @@ const Components = (props: ComponentsProps) => {
   const { signInModal, signUpModal, userProfileModal, organizationProfileModal, createOrganizationModal, nodes } =
     state;
 
+  // TODO: rename these
+  const urlClerkState = getClerkQueryParam('__clerk_state') ?? '';
```
SokratisVidros, line 164: 🙃 Let's convert the `__clerk_state` into a constant and re-use it across the code base.

packages/clerk-js/src/ui/Components.tsx
```diff
@@ -149,7 +160,21 @@ const Components = (props: ComponentsProps) => {
...
+  const decodedRedirectParams = urlClerkState ? JSON.parse(atob(urlClerkState)) : null;
```
SokratisVidros, line 165: 🔧 I'd suggest to move this to a separate utility helper e.g. `encodeBase64Json` and `decodeBase64Json`. Moreover, we must wrap this with a try/catch and cater for possible features and turn null on any error.

packages/clerk-js/src/ui/Components.tsx
```diff
@@ -149,7 +160,21 @@ const Components = (props: ComponentsProps) => {
...
+
   useSafeLayoutEffect(() => {
+    if (urlClerkState && decodedRedirectParams && decodedRedirectParams?.modal) {
```
SokratisVidros, line 168:
```suggestion
    if (decodedRedirectParams && decodedRedirectParams?.modal) {
```

packages/clerk-js/src/ui/Components.tsx
```diff
@@ -149,7 +160,21 @@ const Components = (props: ComponentsProps) => {
...
+      setState(s => ({
+        ...s,
+        [componentNodes[snakeToCamel(decodedRedirectParams.basePath)]]: {},
```
SokratisVidros, line 171: ❓ Component paths are dynamic. For example, the <UserProfile/> is located at `/user` by default but users can change it to anything. So the mapping between paths and components should be dynamic. 

Can we replace that basePath resolution with a property that says which component needs to be loaded on redirect? 

@nikosdouvlis WDTY?

packages/clerk-js/src/utils/getClerkQueryParam.ts
```diff
@@ -20,3 +22,10 @@ export function getClerkQueryParam<T extends ClerkQueryParam>(param: T): ClerkQu
   const val = new URL(window.location.href).searchParams.get(param);
   return val ? (val as ClerkQueryParamsToValuesMap[T]) : null;
 }
+
+export function removeClerkQueryParam<T extends ClerkQueryParam>(param: T) {
```
SokratisVidros, line 26: 🔧 Please add unit tests for this function.

packages/clerk-js/src/utils/getClerkQueryParam.ts
```diff
@@ -20,3 +22,10 @@ export function getClerkQueryParam<T extends ClerkQueryParam>(param: T): ClerkQu
...
+  new URL(window.location.href).searchParams.delete(param);
+  const strippedUrl = new URL(`${window.location.origin}${window.location.pathname}`);
+  window.history.pushState({}, '', strippedUrl);
```
SokratisVidros, line 29: 🔧 We need to make sure this solution will work across all routers and frameworks (e.g. Next.js, Remix, etc...). BTW, shouldn't we use replaceState instead of pushState?

SokratisVidros (changes requested):

packages/clerk-js/src/ui/Components.tsx
nikosdouvlis: A couple of things here.

1. As said before, reading + decoded + cleaning the state param should done with helper functions

2. We are only using this logic if the component was opened as a modal. So instead of always appending the state and checking the `modal` key, let's only append the state when it's needed in the first place. So if we have serialized state, we always perform the action dictated by the state (in this case, we open a modal)

3. We shouldn't have to do `options.startPath` . `options` is the public API of ClerkProvider and type changes to this are user-facing changes that we don't want 

4. I'm not sure what `componentNodes` represents or what `        [componentNodes[snakeToCamel(decodedRedirectParams.basePath)]]: {},` actually does... Can you please elaborate?

nikosdouvlis: Generally, it might be easier to append the `componentName` and the current path when the flow starts and just append it to the startPath, eg: 
```
              startPath={'/organization' + serializedState.path}
```
(this is just an example as closing the modal afterward should also clear the path so next time it opens, it opens as it normally would)

packages/clerk-js/src/ui/components/UserProfile/ConnectedAccountsPage.tsx
```diff
@@ -47,6 +47,15 @@ const AddConnectedAccount = () => {
   const user = useCoreUser();
   const { navigate } = useNavigate();
   const { strategies, strategyToDisplayData } = useEnabledThirdPartyProviders();
+  const { currentPath } = useRouter();
+  const isModal = useUserProfileContext().mode === 'modal';
+
+  const redirectParams = {
```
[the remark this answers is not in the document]
raptisj: @nikosdouvlis you are right, this is the case. So should we rename the url param from `__clerk_state` to `__clerk_modal_state`?

packages/clerk-js/src/utils/queryStateParams.ts
```diff
@@ -0,0 +1,38 @@
...
+type SerializeAndAppendModalStateProps = { url: string; isModal: boolean; currentPath: string; componentName: string };
+export const serializeAndAppendModalState = ({
+  url,
+  isModal,
+  currentPath,
+  componentName,
+}: SerializeAndAppendModalStateProps) => {
+  if (!isModal) {
+    return url;
+  }
```
desiprisg, line 26: This has me a bit confused. Why are we passing `isModal` here if we are just going to return the `url`? Shouldn't this be handled outside the function? Am I missing something?

raptisj: @desiprisg I refactored the function as you suggested.

packages/clerk-js/src/ui/Components.tsx
```diff
@@ -258,7 +276,7 @@ const Components = (props: ComponentsProps) => {
             <VirtualRouter
               preservedParams={PRESERVED_QUERYSTRING_PARAMS}
               onExternalNavigate={() => componentsControls.closeModal('userProfile')}
-              startPath='/user'
+              startPath={'/user' + urlStateParam?.path}
```
SokratisVidros, line 279: 🔧 Instead of string concatenation I'd suggest using the `buildUrl` helper that ensures proper & safer URL building.

packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationProfile.tsx
```diff
@@ -57,7 +57,7 @@ export const OrganizationProfileModal = (props: OrganizationProfileProps): JSX.E
   };
 
   return (
-    <Route path='organizationProfile'>
+    <Route path='organization'>
```
SokratisVidros, line 60: ❓ Is this change intentional? Won't this affect our current component routing?

packages/clerk-js/src/ui/elements/Accordion.tsx
```diff
@@ -25,9 +25,15 @@ export const AccordionItem = (props: AccordionItemProps) => {
...
 
+  React.useEffect(() => {
+    setIsOpen(defaultOpen);
+  }, [defaultOpen]);
+
   React.useLayoutEffect(() => {
-    if (scrollOnOpen && isOpen && contentRef.current) {
-      contentRef.current.scrollIntoView({ behavior: 'smooth', block: 'center' });
+    if (scrollOnOpen && isOpen) {
+      window.requestAnimationFrame(() => {
```
SokratisVidros, line 34: 🔧 Since `requestAnimationFrame` is used in a `useLayoutEffect` we should return a function with a cancelAnimationFrame to be 100% correct. PTAL https://css-tricks.com/using-requestanimationframe-with-react-hooks/

packages/clerk-js/src/utils/queryStateParams.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const decodeBase64Json = (value: string) => {
+  return JSON.parse(atob(value));
+};
+
+export const encodeBase64Json = (value: { path: string; componentName: string }) => {
+  return btoa(JSON.stringify(value));
+};
+
+export const readAndRemoveStateParam = () => {
```
SokratisVidros, line 11: ❓ Does this function also remove the __clerk_state param from the URL or just reads it?

packages/clerk-js/src/utils/queryStateParams.ts
```diff
@@ -0,0 +1,40 @@
...
+    path: currentPath.replace(regexPattern, '') || '',
+    componentName,
+  };
+
+  const encodedRedirectParams = encodeBase64Json(redirectParams);
+
+  const urlWithParams = new URL(url);
+  const searchParams = urlWithParams.searchParams;
+
+  searchParams.set('__clerk_state', encodedRedirectParams);
```
SokratisVidros, line 35: ⛔ Let's rename the param to `__clerk_modal_state` as we already have a `__clerk_status` param for authentication purposes and the name will cause confusion to the future reader.

packages/clerk-js/src/utils/queryStateParams.ts
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export const readAndRemoveStateParam = () => {
+  const urlClerkState = getClerkQueryParam('__clerk_state') ?? '';
+
+  return urlClerkState ? decodeBase64Json(urlClerkState) : null;
+};
+
+type SerializeAndAppendModalStateProps = { url: string; currentPath: string; componentName: string };
+export const serializeAndAppendModalState = ({
```
SokratisVidros, line 18:
```suggestion
export const appendModalState = ({
```

packages/clerk-js/src/utils/queryStateParams.ts
```diff
@@ -0,0 +1,40 @@
+import { getClerkQueryParam } from '../utils';
+
+export const decodeBase64Json = (value: string) => {
```
SokratisVidros, line 3: These utilities exist in `encoders.ts`. Please fetch them from there.

packages/clerk-js/src/ui/hooks/useReadParamState.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
```
SokratisVidros, line 1: 🔧 Please rename this to `useClerkModalState to dictate its usage

SokratisVidros (changes requested):

packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationProfile.tsx
raptisj: @SokratisVidros this corresponds to the `startPath` props passed in VirtualRouter(Components.tsx). This change was made for consistency(we have `sign-in` not `signIn`)

cc: @nikosdouvlis

packages/clerk-js/src/utils/queryStateParams.ts
```diff
@@ -0,0 +1,51 @@
...
+  return urlClerkState ? JSON.parse(atob(urlClerkState)) : null;
+};
+
+type SerializeAndAppendModalStateProps = {
+  url: string;
+  startPath?: string;
+  currentPath?: string;
+  componentName: string;
+};
+export const appendModalState = ({
```
anagstef, line 27: 🙃  We could add tests for this.

packages/clerk-js/src/utils/queryStateParams.ts
```diff
@@ -0,0 +1,51 @@
+import { encodeB64, getClerkQueryParam } from '../utils';
+
+export const buildUrl = ({ base, path }: { base: string; path: string | undefined }) => {
```
anagstef, line 3: 🔧 I see that we already have an exported `buildURL` from utils. We could rename this.

packages/clerk-js/src/ui/router/__tests__/VirtualRouter.test.tsx
```diff
@@ -37,7 +37,7 @@ const ShowPreserved = () => {
 const Tester = () => (
   <VirtualRouter
     preservedParams={['preserved']}
-    startPath='/'
+    startPath={'/' + ''}
```
anagstef, line 40: ❓ Is this for `tsc` I guess?

packages/clerk-js/src/ui/elements/NavigateToFlowStartButton.tsx
```diff
@@ -11,6 +11,10 @@ export const useNavigateToFlowStart = () => {
     if (to !== router.currentPath) {
       return router.navigate(to);
     }
+
+    if (router.urlStateParam?.path) {
```
anagstef, line 15: 🙃 We can also add some tests for this to be safe

anagstef (changes requested):
Looks good. Just the `buildUrl` naming needs to be refactored for me.

packages/clerk-js/src/utils/queryStateParams.ts
raptisj: @anagstef the util you are referring to is doing another thing. It can't be used for our case and I think it should not be refactored to do many things.

packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationProfile.tsx
SokratisVidros: @nikosdouvlis ☝️

packages/clerk-js/src/utils/queryStateParams.ts
raptisj: @anagstef added a test for this one

packages/clerk-js/src/utils/queryStateParams.ts
raptisj: @anagstef changed the naming here

anagstef (approved):

packages/clerk-js/src/utils/queryStateParams.ts
nikosdouvlis: This doesn't look like its used anywhere except for tests.
- Do we still need it?
- /utils exports an `encodeB64` maybe it also exports the corresponding decode method?
- If not, lets move it there and replace the JSON.parse inside readStateParam

packages/clerk-js/src/ui/components/CreateOrganization/CreateOrganization.tsx
```diff
@@ -47,7 +47,7 @@ export const CreateOrganizationModal = (props: CreateOrganizationProps): JSX.Ele
   };
 
   return (
-    <Route path='createOrganization'>
+    <Route path='create-organization'>
```
nikosdouvlis, line 50: Why do we need this change?

packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationProfile.tsx
nikosdouvlis: I think this is safe, but Id rather  not include it in this PR if its not 100% required as the routing logic is pretty complex

packages/clerk-js/src/ui/components/CreateOrganization/CreateOrganization.tsx
raptisj: @nikosdouvlis basically this is for convention reasons(see `sign-in`). Is this breaking? I don't think so

packages/clerk-js/src/utils/queryStateParams.ts
raptisj: @nikosdouvlis I removed the the helper that was not used

packages/clerk-js/src/utils/queryStateParams.ts
```diff
@@ -0,0 +1,48 @@
...
+    componentName,
+    startPath,
+  };
+
+  const encodedRedirectParams = encodeB64(JSON.stringify(redirectParams));
+
+  const urlWithParams = new URL(url);
+  const searchParams = urlWithParams.searchParams;
+
+  searchParams.set('__clerk_modal_state', encodedRedirectParams);
```
SokratisVidros, line 43: Lets move the `__clerk_modal_state` to `core/constants.ts`

SokratisVidros (approved):

SokratisVidros (approved):

nikosdouvlis (approved):

PR #572, opened 2022-11-28
feat(clerk-js,types): Terse paths parameters

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

After this PR, when the `routing` prop is not passed, and the `path` prop is passed, `routing` will default to `'path'`.

The types for this are also stricter so the path prop will throw a typescript error if passed without `routing='path'`

### DIFF
--- a/packages/clerk-js/src/ui/common/authPropHelpers.ts
+++ b/packages/clerk-js/src/ui/common/authPropHelpers.ts
@@ -86,5 +86,5 @@ export const normalizeRoutingOptions = ({
   routing?: RoutingStrategy;
   path?: string;
 }): RoutingOptions => {
-  return { routing: !!path && !routing ? 'path' : routing, path } as RoutingOptions;
+  return { routing: !!path && routing === undefined ? 'path' : routing, path } as RoutingOptions;
 };
--- a/packages/clerk-js/src/ui/common/authPropHelpers.ts
+++ b/packages/clerk-js/src/ui/common/authPropHelpers.ts
@@ -86,5 +86,5 @@ export const normalizeRoutingOptions = ({
   routing?: RoutingStrategy;
   path?: string;
 }): RoutingOptions => {
-  return { routing: !routing && path !== undefined ? 'path' : routing, path } as RoutingOptions;
+  return { routing: !!path && !routing ? 'path' : routing, path } as RoutingOptions;
 };
--- a/packages/clerk-js/src/ui/common/authPropHelpers.ts
+++ b/packages/clerk-js/src/ui/common/authPropHelpers.ts
@@ -1,5 +1,5 @@
 import { camelToSnake } from '@clerk/shared';
-import type { DisplayConfigResource } from '@clerk/types';
+import type { DisplayConfigResource, RoutingOptions, RoutingStrategy } from '@clerk/types';
 import type { ParsedQs } from 'qs';
 import qs from 'qs';
 
@@ -77,4 +77,14 @@ export const buildAuthQueryString = (data: BuildAuthQueryStringArgs): string | n
     }
   }
   return Object.keys(query).length === 0 ? null : qs.stringify(query);
+};
+
+export const normalizeRoutingOptions = ({
+  routing,
+  path,
+}: {
+  routing?: RoutingStrategy;
+  path?: string;
+}): RoutingOptions => {
+  return { routing: !routing && path !== undefined ? 'path' : routing, path } as RoutingOptions;
 };
--- a/packages/types/src/clerk.ts
+++ b/packages/types/src/clerk.ts
@@ -472,6 +472,26 @@ export interface Resources {
 
 export type RoutingStrategy = 'path' | 'hash' | 'virtual';
 
+export type RoutingOptions =
+  | {
+      /**
+       * Page routing strategy
+       */
+      routing?: Exclude<RoutingStrategy, 'path'>;
+      /**
+       * Root URL where the component is mounted on, eg: '/sign-in'
+       */
+      path?: never;
+    }
+  | {
+      routing: 'path';
+      path: string;
+    }
+  | {
+      routing?: never;
+      path: string;
+    };
+
 export type RedirectOptions = {
   /**
    * Full URL or path to navigate after successful sign in.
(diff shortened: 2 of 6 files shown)


SokratisVidros (approved):

packages/types/src/clerk.ts
```diff
@@ -472,6 +472,22 @@ export interface Resources {
 
 export type RoutingStrategy = 'path' | 'hash' | 'virtual';
 
+export type RoutingOptions =
```
nikosdouvlis, line 475: Let's not export this for now, as it is mostly used to DRY the other types internally. We can do it later if needed.

Ideally we wouldn't want to export `RoutingStrategy` either but since its a breaking change to do it now, let's leave it as is.

packages/clerk-js/src/ui/common/authPropHelpers.ts
```diff
@@ -78,3 +78,28 @@ export const buildAuthQueryString = (data: BuildAuthQueryStringArgs): string | n
...
+export const verifyRoutingOptions = ({
+  routing,
+  path,
+}: {
+  routing?: RoutingStrategy;
+  path?: string;
+}): RoutingOptions => {
+  if (!path) {
+    if (routing === 'path') {
+      throw new Error('Clerk: path routing requires passing the `path` prop');
```
nikosdouvlis, line 91: 🛑 Throwing the error here is great and you had the correct idea :)

Unfortunately, we cannot do this right now. Since we hotload `clerk-js` this is not only a breaking change, but it also has the potential to break live production apps if the user has been using these props incorrectly until now.

Imagine the following scenario:

```
<SignIn routing='hash' path='/sign-in' />
```

Based on the Portal impl [here](https://github.com/clerkinc/javascript/blob/b0d5e73ff718c43382cbf6f2ad23ae9627a8eb9a/packages/clerk-js/src/ui/portal/index.tsx#L1-L0), the above component will use the hash router even if the path prop exists. So in theory there can be a live app using these props in that manner.

I'm adding a [ticket](https://linear.app/clerk/issue/JS-56/throw-an-error-if-routing-component-params-are-used-incorrectly) for now

packages/clerk-js/src/ui/common/authPropHelpers.ts
```diff
@@ -78,3 +78,28 @@ export const buildAuthQueryString = (data: BuildAuthQueryStringArgs): string | n
   }
   return Object.keys(query).length === 0 ? null : qs.stringify(query);
 };
+
+export const verifyRoutingOptions = ({
```
nikosdouvlis, line 82: I think the general idea is fine! I've got two comments, so let's discuss.

1. The name is slightly misleading as we're _normalising_ the params and not just verifying them. Should we rename it accordingly?

2. You clearly have more context, but I think we only need the following check (pseudo code follows):

```tsx
props => {
  if (props.path && props.routing === undefined) {
     return { ...props, routing: 'path'} 
  }
 return props
}
```
Would that be enough?

nikosdouvlis (changes requested):

packages/clerk-js/src/ui/common/authPropHelpers.ts
desiprisg: Yeap, I'm aware that this can't throw an error. I was going to remove this once I solve another issue. We can do a `console.error` I guess or nothing at all.

nikosdouvlis: Let's do nothing at all for now, we will handle it properly when we work on the ticket I posted above

packages/clerk-js/src/ui/common/authPropHelpers.ts
desiprisg: Typescript wise, not really. We'll get some errors there. I'd prefer if we respect the types here to avoid any future mistakes.

packages/clerk-js/src/ui/common/authPropHelpers.ts
desiprisg: What if `routing='path'` but no `path` prop is passed?

packages/types/src/clerk.ts
```diff
@@ -472,6 +472,26 @@ export interface Resources {
...
+      path?: never;
+    }
+  | {
+      routing: 'path';
+      path: string;
+    }
+  | {
+      routing?: never;
+      path: string;
+    };
```
nikosdouvlis, line 493: Do you think this would be more readable? 

```
export type RoutingOptions =
  | { path: string }
  | { path: string; routing: Extract<RoutingStrategy, 'path'> }
  | { path?: never; routing: Extract<RoutingStrategy, 'hash' | 'virtual'> };
```

desiprisg: Wouldn't this work as well?
```
export type RoutingOptions =
  | { path: string; routing?: Extract<RoutingStrategy, 'path'> }
  | { path?: never; routing?: Extract<RoutingStrategy, 'hash' | 'virtual'> };
  ```

File: packages/clerk-js/src/ui/common/authPropHelpers.ts [hunk 5]
```diff
@@ -78,3 +78,13 @@ export const buildAuthQueryString = (data: BuildAuthQueryStringArgs): string | n
...
 };
+
+export const normalizeRoutingOptions = ({
+  routing,
+  path,
+}: {
+  routing?: RoutingStrategy;
+  path?: string;
+}): RoutingOptions => {
+  return { routing: !routing && path !== undefined ? 'path' : routing, path } as RoutingOptions;
```
nikosdouvlis, line 89: I think the check here should be:

```suggestion
  return { routing: !!path && routing === undefined ? 'path' : routing, path } as RoutingOptions;
```

nikosdouvlis (commented):
Good job, had a small last comment but this is looking great.

Can we also add a test for authPropHelpers? No need to test every helper, just the one we added

packages/clerk-js/src/ui/common/authPropHelpers.ts
desiprisg: `routing` can't really be null so we can just check `!routing` I guess.

nikosdouvlis: It shouldn't but it might be, lets test it explicitly here please because we want to avoid changing behavior for as many cases as possible

desiprisg: Done! We should change this in the next major version though.

PR #574, opened 2022-11-29
Use web workers to fire timeouts and intervals

nikosdouvlis:
## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

For more context, you can read this [introductory article](https://developer.chrome.com/blog/timer-throttling-in-chrome-88/) briefly explaining what timer throttling is.

Our authentication poller uses a usual `setInterval` timer to check if the JWT needs to be refreshed and update the `__session` cookie accordingly. Even though most browsers claim to fire the timers at least once per minute, even when throttled, our tests have shown that background tabs under certain conditions will throttle the timers more, and the once-per-minute claim is not guaranteed. 

If the authentication poller does not fire at the correct interval and given that by default, we're using a very short-lived token for security reasons (~1min TTL), the token can expire, and any background requests using that token will eventually fail.

To prevent this issue, we're introducing an abstraction over the web worker API. Web workers run on a different thread and are not subject to the same throttling restrictions. 

## What to test
- JWT refresh
- `__session` cookie update
- email link flow

## Other considerations
The [page lifecycle API](https://wicg.github.io/page-lifecycle/#worklets-workers) states that if a tab transitions to the `frozen` state, worker execution will also halt - no other background requests can be running at that time, though. The frozen state is not yet supported by most browsers, and although Chrome claims it supports it, it's not clear how and if it's actually used yet. If we end up facing issues in the future, we can use the `freeze` and `resume` events to handle the stale token accordingly.

### DIFF
--- a/packages/shared/src/utils/workerTimers/workerTimers.worker.ts
+++ b/packages/shared/src/utils/workerTimers/workerTimers.worker.ts
@@ -1,0 +1,39 @@
+import type { WorkerTimerEvent, WorkerTimerId, WorkerTimerResponseEvent } from './workerTimers.types';
+// This file is loaded as TEXT by esbuild during build
+// and fed into a blob so we can easily instantiate the web worker
+// look at the tsup.config.ts file for more details on the loader
+
+const respond = (p: WorkerTimerResponseEvent) => {
+  self.postMessage(p);
+};
+
+const workerToTabIds: Record<WorkerTimerId, WorkerTimerId> = {};
+
+self.addEventListener('message', e => {
+  const data = e.data as WorkerTimerEvent;
+
+  switch (data.type) {
+    case 'setTimeout':
+      workerToTabIds[data.id] = setTimeout(() => {
+        respond({ id: data.id });
+      }, data.ms) as unknown as WorkerTimerId;
+      break;
+    case 'clearTimeout':
+      if (workerToTabIds[data.id]) {
+        clearTimeout(workerToTabIds[data.id]);
+        delete workerToTabIds[data.id];
+      }
+      break;
+    case 'setInterval':
+      workerToTabIds[data.id] = setInterval(() => {
+        respond({ id: data.id });
+      }, data.ms) as unknown as WorkerTimerId;
+      break;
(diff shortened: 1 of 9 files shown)


packages/shared/tsup.config.ts
```diff
@@ -16,5 +18,21 @@ export default defineConfig(overrideOptions => {
...
+// Read transform and minify any files ending in .worker.ts
+// These files can be imported as modules and used as string when instantiating
+// a new web worker, without loading an external file during runtime
+export const WebWorkerMinifyPlugin: Plugin = {
+  name: 'WebWorkerMinifyPlugin',
+  setup(build) {
+    build.onLoad({ filter: /\.worker\.ts/ }, async args => {
+      const f = await readFile(args.path);
+      const js = await transform(f, { loader: 'ts', minify: true });
+      console.log(js.code);
```
SokratisVidros, line 34: 🙃 Please remove.

packages/shared/src/utils/workerTimers/workerTimers.worker.ts
```diff
@@ -0,0 +1,39 @@
+import type { WorkerTimerEvent, WorkerTimerId, WorkerTimerResponseEvent } from './workerTimers.types';
+// This file is loaded as TEXT by esbuild during build
```
SokratisVidros, line 2: 💯

SokratisVidros (approved):

desiprisg (approved):

anagstef (approved):

PR #580, opened 2022-11-29
feat(clerk-js): Add support for dates localization

anagstef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description

This PR introduces date related localization keys and modifiers, and the ability to add `string` parameters to localization modifiers. Until now we had only the modifier `titleize` which had no need for parameters (just the value to modify).

This PR adds the `timeString`, `weekday` and `numeric` modifiers and 6 new localization keys for relative dates.

#### `weekday:` 

Example usages: 
`{{ date | weekday('en-US','long') }}`
`{{ date | weekday('en-US') }}`
`{{ date | weekday }}`

This modifier is used upon a date value in one of the following types: `Date | string | number`.
Also, this modifier can receive 0-2 arguments:
- First argument is the `locale` as it is used by the `Date` object (i.e. `en-US`, `fr-FR`, etc.). If omitted, it falls back to `en-US`.
- Second argument is the `weekday`, getting the values `'long' | 'short' | 'narrow'`, as it is described [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat). If omitted, it falls back to `long`.

It returns a `string` with the weekday in the specified form: i.e. `Monday`.

#### `timeString:` 

Example usages: 
`{{ date | timeString('en-US') }}`
`{{ date | timeString }}`

This modifier is used upon a date value in one of the following types: `Date | string | number`.
Also, this modifier can receive 0-1 arguments:
- First argument is the `locale` as it is used by the `Date` object (i.e. `en-US`, `fr-FR`, etc.). If omitted, it falls back to `en-US`.

It returns a `string` with the 12-hour time value: i.e. `12:00 AM`.

#### `numeric:` 

Example usages: 
`{{ date | numeric('en-US') }}`
`{{ date | numeric }}`

This modifier is used upon a date value in one of the following types: `Date | string | number`.
Also, this modifier can receive 0-1 arguments:
- First argument is the `locale` as it is used by the `Date` object (i.e. `en-US`, `fr-FR`, etc.). If omitted, it falls back to `en-US`.

It returns a `string` with the calendar date value: i.e. `12/31/2022`.

#### `New localization strings for relative dates:`

We introduce 6 new localization keys for the relative dates:
```
dates.previous6Days
dates.lastDay
dates.sameDay
dates.nextDay
dates.next6Days
dates.numeric
```

The default english values are:
```
dates: {
  previous6Days: "Last {{ date | weekday('en-US','long') }} at {{ date | timeString('en-US') }}",
  lastDay: "Yesterday at {{ date | timeString('en-US') }}",
  sameDay: "Today at {{ date | timeString('en-US') }}",
  nextDay: "Tomorrow at {{ date | timeString('en-US') }}",
  next6Days: "{{ date | weekday('en-US','long') }} at {{ date | timeString('en-US') }}",
  numeric: "{{ date | numeric('en-US') }}"
}
```


nikosdouvlis:
I like where this is going 🙃

packages/types/src/localization.ts
```diff
@@ -518,4 +518,11 @@ type _LocalizationResource = {
     form_password_incorrect: LocalizationValue;
     not_allowed_access: LocalizationValue;
   };
+  relativeDates: {
```
nikosdouvlis, line 521: Let's rename this to `dates` so we can add more props in the future, if needed, under the same top-level key.

If we want to introduce more structure, we can also do `dates.relativeFormat` but I don't feel strongly about it.

These are my suggestions:

```tsx
  dates: {
     ...
  }
```

or 

```tsx
  dates: {
     relativeFormat: { 
         ....
      }
  }
```


Regarding the actual case names now, do you think we can remove the "at" suffix as we don't really enforce a specific format (eg "last {} at {}") - the user can pass any string they want 🙃

packages/shared/src/utils/date.ts
```diff
@@ -30,10 +39,16 @@ function normalizeDate(d: Date | string | number): Date {
...
+  relativeTo: Date | string | number;
+};
+
+export type RelativeDateCase = 'numeric' | 'previous6DaysAt' | 'lastDayAt' | 'sameDayAt' | 'nextDayAt' | 'next6DaysAt';
+
 /*
  * Follows date-fns format, see here:
  * https://date-fns.org/v2.21.1/docs/formatRelative
- * TODO: support localisation
  * | Distance to the base date | Result                    |
```
nikosdouvlis, line 52: Let's remove the table from the comment here as its no longer true

packages/clerk-js/src/ui/components/UserProfile/ActiveDevicesSection.tsx
```diff
@@ -154,11 +165,49 @@ const DeviceInfo = (props: { session: SessionWithActivitiesResource }) => {
...
         >
-          {formatRelative(props.session.lastActiveAt || new Date(), new Date())}
+          {getRelativeToNowDate(props.session.lastActiveAt, t)}
         </Text>
       </Col>
     </Flex>
   );
 };
+
+type RelativeDateKeys =
```
nikosdouvlis, line 176: We can avoid duplicating these by slightly changing getRelativeToNowDate, see comment below

packages/clerk-js/src/ui/components/UserProfile/ActiveDevicesSection.tsx
```diff
@@ -154,11 +165,49 @@ const DeviceInfo = (props: { session: SessionWithActivitiesResource }) => {
...
+  | 'relativeDates.previous6DaysAt'
+  | 'relativeDates.lastDayAt'
+  | 'relativeDates.sameDayAt'
+  | 'relativeDates.nextDayAt'
+  | 'relativeDates.next6DaysAt';
+
+const getRelativeToNowDate = (
+  date: Date,
+  t: (localizationKey: string | LocalizationKey | undefined) => string,
```
nikosdouvlis, line 185: Good thinking passing `t` to this one instead of creating another hook. Shall we do the exact opposite though as its more consistent with the rest of the codebase? 
Let's make `getRelativeToNowDate` return a `localizationKey` and do this instead: `
```tsx
t(relativeDateLocalizationKey({date: lastactiveat, relativeTo: Date.now() }))
```


Nitpicky: we can rename `getRelativeToNowDate` accordingly

packages/clerk-js/src/ui/components/UserProfile/ActiveDevicesSection.tsx
```diff
@@ -154,11 +165,49 @@ const DeviceInfo = (props: { session: SessionWithActivitiesResource }) => {
...
+): string => {
+  let localizationKey: RelativeDateKeys;
+  const relativeDate = formatRelative({ date: date || new Date(), relativeTo: new Date() });
+  if (!relativeDate) {
+    return '';
+  }
+  switch (relativeDate.relativeDateCase) {
+    case 'previous6DaysAt':
+      localizationKey = 'relativeDates.previous6DaysAt';
+      break;
```
nikosdouvlis, line 195:
```suggestion
    return localizationKeys('relativeDates.previous6DaysAt', { lastActiveAt: relativeDate.date })
```

packages/clerk-js/src/ui/localization/__tests__/applyTokensToString.test.ts
```diff
@@ -23,6 +26,25 @@ describe('applyTokensToString', function () {
     ],
     ['', ''],
     [undefined, ''],
+    ["Last {{ date | weekday('en-US','long') }} at {{ date | timeString('en-US') }}", 'Last Friday at 10:00 PM'],
```
nikosdouvlis, line 29: Nit picking here but I know you don't mind it 🙃 
Can we split the relative days to a separate `describe` block since they are way more complex than the rest?
Also, don't forget to add some "broken" cases, eg: 
- weekDay without params
- weekDay after a non-date token
- multiple modifiers with and without params, eg: `date | weekDay(...) | titleize`
-

packages/clerk-js/src/ui/localization/applyTokensToString.ts
```diff
@@ -1,8 +1,37 @@
+import { titleize } from '@clerk/shared';
+
 import { useCoreClerk, useEnvironment } from '../contexts';
-import { titleize } from '../shared';
+
+export const timeString = (val: Date | string | number, locale?: string): string => {
```
nikosdouvlis, line 5: Let's make a `localizationModifiers.ts` file, move the implementations there and export `MODIFIERS` from it.

packages/clerk-js/src/ui/localization/applyTokensToString.ts
```diff
@@ -1,8 +1,37 @@
...
+      minute: 'numeric',
+      hour12: true,
+    });
+  } catch (e) {
+    console.warn(e);
+    return '';
+  }
+};
+
+export const weekday = (
```
nikosdouvlis, line 18: Awesome. Do you think we should expose other `Intl` modifiers as well?

packages/clerk-js/src/ui/localization/applyTokensToString.ts
```diff
@@ -66,15 +95,36 @@ const parseTokensFromLocalizedString = (
...
     s = s.replace(`_++${token}++_`, value);
   });
   return s;
 };
 
 const assertKnownModifier = (s: any): s is Modifier => Object.prototype.hasOwnProperty.call(MODIFIERS, s);
+
+const getModifierWithParams = (modifierExpression: string): { modifierName: string; params: string[] } => {
+  const parts = modifierExpression
+    .split(/[(,)]/g)
```
nikosdouvlis, line 120: Lets also ignore whitespace here otherwise, `weekDay( 'en-US', 'long')` will break

packages/clerk-js/src/ui/localization/applyTokensToString.ts
```diff
@@ -66,15 +95,36 @@ const parseTokensFromLocalizedString = (
...
+      }
+    }, tokens[token]);
     s = s.replace(`_++${token}++_`, value);
   });
   return s;
 };
 
 const assertKnownModifier = (s: any): s is Modifier => Object.prototype.hasOwnProperty.call(MODIFIERS, s);
+
+const getModifierWithParams = (modifierExpression: string): { modifierName: string; params: string[] } => {
```
nikosdouvlis, line 118: ✅

packages/shared/src/utils/date.ts
```diff
@@ -30,10 +39,16 @@ function normalizeDate(d: Date | string | number): Date {
   }
 }
 
+export type DateFormatRelativeParams = {
```
nikosdouvlis, line 42:
```suggestion
type DateFormatRelativeParams = {
```

packages/shared/src/utils/date.ts
```diff
@@ -30,10 +39,16 @@ function normalizeDate(d: Date | string | number): Date {
...
+  date: Date | string | number;
+  relativeTo: Date | string | number;
+};
+
+export type RelativeDateCase = 'numeric' | 'previous6DaysAt' | 'lastDayAt' | 'sameDayAt' | 'nextDayAt' | 'next6DaysAt';
```
nikosdouvlis, line 47:
```suggestion
type RelativeDateCase = 'numeric' | 'previous6DaysAt' | 'lastDayAt' | 'sameDayAt' | 'nextDayAt' | 'next6DaysAt';
```

packages/clerk-js/src/ui/localization/applyTokensToString.ts
nikosdouvlis: We usually prefer to let TS infer the returned type instead of explicitly specifying it  for internal methdos
```suggestion
const getModifierWithParams = (modifierExpression: string) => {
```

packages/clerk-js/package.json
```diff
@@ -35,8 +35,8 @@
     "prepublishOnly": "npm run build",
     "postpublish": "node ./scripts/purge-cache.mjs",
     "start": "echo \"Noop\"",
-    "test": "jest",
-    "test:coverage": "jest --collectCoverage && open coverage/lcov-report/index.html"
+    "test": "TZ=UTC jest",
```
nikosdouvlis, line 38: Makes sense, but is there any other way to do this?

nikosdouvlis (commented):
I really like how composable the new API is, nice work

packages/clerk-js/src/ui/components/UserProfile/ActiveDevicesSection.tsx
anagstef: This is done like this, because there is a case where we have a numeric only value (i.e. `12/12/2022`), that does not need to pass through localization. This case is returned by `formatRelative` along with the other 5 ones. Should we extract this case before calling `t` ?

packages/clerk-js/src/ui/localization/applyTokensToString.ts
anagstef: Whitespace is getting ignored later on, `weekday( 'en-US', 'long')` works!

packages/clerk-js/package.json
anagstef: Yes, we can extend the `jsdom` environment to set `process.env.TZ` before setup. I will refactor.

nikosdouvlis (approved):
excellent stuff!

PR #586, opened 2022-11-30
feat(clerk-react): Add orgId and orgRole to useAuth

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `build/tooling/chore`

## Description


packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -47,10 +47,17 @@ export function ClerkContextProvider(props: ClerkContextProvider): JSX.Element |
   const clerkCtx = React.useMemo(() => ({ value: clerk }), [clerkLoaded]);
   const clientCtx = React.useMemo(() => ({ value: state.client }), [state.client]);
 
-  const { sessionId, session, userId, user } = derivedState;
+  const { sessionId, session, userId, user, organization } = derivedState;
   const actor = session?.actor || null;
-
-  const authCtx = React.useMemo(() => ({ value: { sessionId, userId, actor } }), [sessionId, userId, actor]);
+  const orgId = organization?.id;
```
nikosdouvlis, line 52: If an org is not selected, shouldn't this be `null`?

packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -47,10 +47,17 @@ export function ClerkContextProvider(props: ClerkContextProvider): JSX.Element |
...
+  const orgRole = session?.user.organizationMemberships.find(m => m.organization.id === orgId)?.role;
```
nikosdouvlis, line 53: Similarly, this should be `null` if an org is not selected

packages/react/src/hooks/useAuth.ts
```diff
@@ -31,6 +35,19 @@ type UseAuthReturn =
       userId: string;
       sessionId: string;
       actor: ActJWTClaim | null;
+      orgId: null;
+      orgRole: null;
+      signOut: SignOut;
+      getToken: GetToken;
+    }
+  | {
+      isLoaded: true;
```
nikosdouvlis, line 44: ✔️

packages/react/src/hooks/useAuth.ts
```diff
@@ -102,18 +122,36 @@ export const useAuth: UseAuth = () => {
...
+      orgId: null,
+      orgRole: null,
       signOut,
       getToken,
     };
   }
 
   if (!!sessionId && !!userId) {
+    if (!!orgId && !!orgRole) {
+      return {
```
nikosdouvlis, line 134: How come we're not using ternaries for `orgId` and `orgRole` instead of the nested if?  Is this the only way to satisfy the discriminated union?

packages/react/src/contexts/ClerkContextProvider.tsx
desiprisg: Oops. Yes. This mistake seems to be repeating itself. It's quite easy to miss!

packages/react/src/hooks/useAuth.ts
desiprisg: I think that was it, but I'll double check

desiprisg: Can't seem to make it work atm, but feel free to try it out

packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -49,8 +49,9 @@ export function ClerkContextProvider(props: ClerkContextProvider): JSX.Element |
 
   const { sessionId, session, userId, user, organization } = derivedState;
   const actor = session?.actor || null;
-  const orgId = organization?.id;
-  const orgRole = session?.user.organizationMemberships.find(m => m.organization.id === orgId)?.role;
+  const orgId = organization === null ? null : organization?.id;
+  const orgRole =
+    organization === null ? null : session?.user.organizationMemberships.find(m => m.organization.id === orgId)?.role;
```
SokratisVidros, line 54: 🔧 Let's make this drier and find the correct `organizationMembership` once before destructuring orgId and orgRole

SokratisVidros (changes requested):

SokratisVidros:
Do we also need to add `orgSlug`? It should be another available claim.

desiprisg:
Actually, we do not really need to differentiate between `null` and `undefined` for `orgId` and `orgRole` inside the `ClerkContextProvider` , since the `useAuth` implementation will set them accordingly depending on the session.

packages/react/src/contexts/ClerkContextProvider.tsx
desiprisg: Actually, we do not really need to differentiate between `null` and `undefined` for `orgId` and `orgRole` inside the `ClerkContextProvider`, since `useAuth` will set them accordingly depending on the `session`.

packages/react/src/contexts/ClerkContextProvider.tsx
desiprisg: I don't think it looks any better since we have to check for null everywhere, but done!

packages/react/src/hooks/useAuth.ts
nikosdouvlis: If type inference with ternaries does not work in this case, lets refactor the conditions to make this more readable: 


```
  if (!!sessionId && !!userId && !orgId) { 
       ...
  }

  if (!!sessionId && !!userId && !!orgId) { 
       ...
  }
```

desiprisg: Actually in the latter case we need to check for orgRole and slug as well for typescript to work, but done!

packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -47,10 +47,22 @@ export function ClerkContextProvider(props: ClerkContextProvider): JSX.Element |
   const clerkCtx = React.useMemo(() => ({ value: clerk }), [clerkLoaded]);
   const clientCtx = React.useMemo(() => ({ value: state.client }), [state.client]);
 
-  const { sessionId, session, userId, user } = derivedState;
+  const { sessionId, session, userId, user, organization } = derivedState;
```
SokratisVidros, line 50: ❓ Shouldn't we also have organizationId at this point?

packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -47,10 +47,22 @@ export function ClerkContextProvider(props: ClerkContextProvider): JSX.Element |
...
   const actor = session?.actor || null;
-
-  const authCtx = React.useMemo(() => ({ value: { sessionId, userId, actor } }), [sessionId, userId, actor]);
+  const organizationMembership =
+    organization === null
+      ? null
+      : session?.user.organizationMemberships.find(m => m.organization.id === organization?.id);
+  const orgId = organizationMembership === null ? null : organizationMembership?.organization.id;
```
SokratisVidros, line 56: I see the need for this but can we make it more dry somehow?

nikosdouvlis (approved):

packages/react/src/hooks/useAuth.ts
```diff
@@ -102,18 +126,39 @@ export const useAuth: UseAuth = () => {
...
       actor: null,
+      orgId: null,
+      orgRole: null,
+      orgSlug: null,
+      signOut,
+      getToken,
+    };
+  }
+
+  if (!!sessionId && !!userId && !!orgId && !!orgRole && orgSlug !== undefined) {
```
nikosdouvlis, line 137: What happens if I have an active org but orgSlug is undefined? I think an org can exist without a slug, so will we be hitting the `  throw new Error(invalidStateError);` error?

desiprisg: Should be null in that case. Pretty sure I checked this, but I can check again. We could just do `orgSlug || null` and not have it in the if statement as well though.

nikosdouvlis: I think this should be the way to go. Otherwise all orgs without slug will throw an error right now. The `|| null` fallback sounds reasonable

desiprisg: Done. Please check!

desiprisg: I can confirm that it would be ok either way since the slug is null in the response from the backend! I've also added null as a possible value for the slug in the `OrganizationResource` type since it was missing.

nikosdouvlis (approved):

PR #615, opened 2022-12-12
feat(clerk-js,types): Add footerLink elementIds

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

This PR adds some extra classes to allow hiding of specific footer actions.
Example (in the sign-in screen):
Screenshot 2022-12-12 at 14 22 00

Also fixing the casing for some elementIds that were recently introduced, and adding the type for a couple more, to get rid of some typescript errors.

### DIFF
--- a/packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationMembers.tsx
+++ b/packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationMembers.tsx
@@ -35,7 +35,7 @@ export const OrganizationMembers = withCardStateProvider(() => {
       <NavbarMenuButtonRow />
       <Col
         elementDescriptor={descriptors.profilePage}
-        elementId={descriptors.profilePage.setId('organization-members')}
+        elementId={descriptors.profilePage.setId('organizationMembers')}
         gap={8}
       >
         <Flex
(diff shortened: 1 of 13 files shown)


nikosdouvlis:
Every element within a page is rendered under a `Flow.Part` context. Do you think we could use that context to generate specific ids for all elements within a `Flow.Part`? 

Eg: https://github.com/clerkinc/javascript/blob/a1c03c4401ddcb4e1c90c99318f4e089bd54baf7/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx#L183

desiprisg:
@nikosdouvlis We could concatenate the root and the part, since the part repeats itself at times.

SokratisVidros (approved):

nikosdouvlis (approved):

nikosdouvlis:
> Are we good with the names here? Do we maybe want to invert them? Right now, In the `SignUp` page, the id is `signIn`, because the message displayed is about signing in. Same for the `SignIn` page.

I was confused by this so please invert them. In general, try to think about it like this: in the future, we can remove these ids and replace them with the FlowPart functionality (the POC from the PR we closed).

So SignUp would have a `footer` and a `footer__signUp`- the API might change but you get the idea.

Lets do the change and then close this one!

PR #627, opened 2022-12-22
Add a retry safeguard for network errors when doing a FAPI request

anagstef:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

This PR introduces a retry safeguard when the FAPI client does a request and it fails due to network errors. It retries 4 times, with exponential logic, only on `GET` requests to be extra safe.

However, it handles 4xx/5xx the same way it did before because these error codes are valid responses.

For 5xx errors, it is a good practice to retry if the request is `Idempotent`. So, we could also retry for `GET` requests with 5xx errors.

## Example:
Dec-22-2022 15-37-37


packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -31,7 +32,10 @@ export abstract class BaseResource {
       clerkMissingFapiClientInResources();
     }
 
-    const { payload, status, statusText } = await BaseResource.fapiClient.request<J>(requestInit);
+    const { payload, status, statusText } = await runWithExponentialBackOff(
```
SokratisVidros, line 35: ⛔ This should be added initially to GET requests only. Adding it to mutating endpoints can result in unexpected behavior.

SokratisVidros (changes requested):
Great idea, but please have a look at the comment.

packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -31,7 +32,10 @@ export abstract class BaseResource {
...
+      () => BaseResource.fapiClient.request<J>(requestInit),
+      { maxRetries: 4, firstDelay: 800 },
```
SokratisVidros, line 37: ❓ How does the first delay option work? Does it affect the first retry?

packages/clerk-js/src/core/resources/Base.ts
anagstef: Indeed! I thought it would be safe for all requests because this is for network errors, but maybe it succeeds and never reaches the client. Thank you!

packages/clerk-js/src/core/resources/Base.ts
anagstef: The first call happens instantly, if it fails, it retries after 800 ms (as `firstDelay` indicates), and then it continues with exponential logic that relates to `firstDelay`. Maybe `firstDelay` sounds a little confusing.

packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -32,10 +32,18 @@ export abstract class BaseResource {
...
-      () => BaseResource.fapiClient.request<J>(requestInit),
-      { maxRetries: 4, firstDelay: 800 },
-    );
+    let fapiResponse: FapiResponse<J>;
+
+    // retry only on GET requests for safety
+    if (requestInit.method === 'GET') {
+      fapiResponse = await runWithExponentialBackOff(() => BaseResource.fapiClient.request<J>(requestInit), {
+        maxRetries: 4,
+        firstDelay: 800,
```
SokratisVidros, line 41: This is too much. I'd set it to 500ms for now. What's the delay between exponential retries?

anagstef: Ok, I'll change it to 500ms, and then it will be 0, 500, 1000, 2000.

anagstef:
After some more digging here, we found that some of the errors may never bubble up to the user, which is helpful for when we retry and succeed the second time but may be harmful when something is wrong, and the user does not see any errors. 

For example, [here in SignInStart.tsx](https://github.com/clerkinc/javascript/blob/a3bd0cf40430c6432fa30eacc44acefc95868b6e/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx#L84) using the `attemptToRecoverFromSignInError` function we mute the error and user never gets an error message. Maybe we could log it on the FAPI side. This may be happening on other components too.

FYI: Regarding the original `TypeError: Failed to fetch` error, it may happen because of a CORS issue instead of timeout/offline. [See here.](https://stackoverflow.com/questions/49343024/getting-typeerror-failed-to-fetch-when-the-request-hasnt-actually-failed)

nikosdouvlis:
@SokratisVidros I think this is good to go, shall we merge?

SokratisVidros (approved):
💯

PR #637, opened 2022-12-28
chore(nextjs): Reuse ClerkProviderProps

nitroin:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature (nextjs/app-beta)
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/nextjs/app-beta`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

~This PR try to re-use `ClerkProviderProps` from `@clerk/clerk-react` instead of narrower (specific) type to allow feature parity, for example the ability to specify localizations.~

This PR enable localization opt-in in nextjs app folder by re-using `ClerkProviderProps` from `@clerk/clerk-react` instead of narrower (specific) type. Only tested props are added to a `NextAppBetaClerkProviderProps` to ensure DX.

### DIFF
--- a/packages/nextjs/src/app-beta/ClerkProvider.tsx
+++ b/packages/nextjs/src/app-beta/ClerkProvider.tsx
@@ -1,9 +1,19 @@
+// import type { ClerkProviderProps } from '@clerk/clerk-react';
 import React from 'react';
 
 import { initialState } from './auth';
+import type { NextAppBetaClerkProviderProps } from './client/ClerkProvider';
 import { ClerkProvider as ClerkProviderClient } from './client/ClerkProvider';
 
-export function ClerkProvider({ children }: { children: React.ReactNode }) {
+export function ClerkProvider({ children, ...restProps }: NextAppBetaClerkProviderProps) {
   const state = initialState()?.__clerk_ssr_state || { sessionId: null, orgId: null, userId: null };
-  return <ClerkProviderClient initialState={state}>{children}</ClerkProviderClient>;
+
+  return (
+    <ClerkProviderClient
+      {...restProps}
+      initialState={state as any}
+    >
+      {children}
+    </ClerkProviderClient>
+  );
 }
--- a/packages/nextjs/src/app-beta/client/ClerkProvider.tsx
+++ b/packages/nextjs/src/app-beta/client/ClerkProvider.tsx
@@ -1,5 +1,6 @@
 'use client';
 // !!! Note the import from react
+import type { ClerkProviderProps } from '@clerk/clerk-react';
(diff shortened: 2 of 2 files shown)


SokratisVidros (approved):
💯

packages/nextjs/src/app-beta/ClerkProvider.tsx
```diff
@@ -1,9 +1,18 @@
...
 import { ClerkProvider as ClerkProviderClient } from './client/ClerkProvider';
 
-export function ClerkProvider({ children }: { children: React.ReactNode }) {
+export function ClerkProvider({ children, ...restProps }: ClerkProviderProps) {
   const state = initialState()?.__clerk_ssr_state || { sessionId: null, orgId: null, userId: null };
-  return <ClerkProviderClient initialState={state}>{children}</ClerkProviderClient>;
+
+  return (
+    <ClerkProviderClient
+      {...restProps}
```
nikosdouvlis, line 12: Many thanks for the contribution! :)

This all looks good to me, but I'm not sure how the `appearance` prop will behave on a server-side component like this one. This prop accepts a value that might not be serializable (`appearance.elements` can also be a function). I'll run some tests to verify and get back to you ASAP

nitroin: That's make sense. Sorry to not having that investigated before, in that case maybe the component should still reuse `ClerkProviderProps` but "proxied" with `Omit<ClerkProviderProps,"not-supported-props">`?

nitroin: After this (maybe stale) discussion I've decided to narrow down the scope of this pull request.

This will allow to enable localization on server side components by reusing the same `ClerkProviderProps` but with a reduced api surface with `Pick<T,...>` - that way we can enable what we are sure to be able to serialize.

Does this looks good to you?

Obv: I will push some other commits and squash locally to fix pipeline/test issues.

nitroin:
I think it's better to close this PR, since this has been superseded by #786.

PR #671, opened 2023-01-13
fix(backend): Extract interstitial rules from tokens/request

dimkl:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

Purpose of this PR is to extract the interstitial rules to separate module and allow those roles to be unit tested and covered extensively!

Related to: https://github.com/clerkinc/javascript/pull/669

### DIFF
--- a/packages/backend/src/tokens/request.ts
+++ b/packages/backend/src/tokens/request.ts
@@ -78,7 +78,7 @@ export async function authenticateRequest(options: AuthenticateRequestOptions): 
 
   async function authenticateRequestWithTokenInCookie() {
     try {
-      const signedOutOrInterstitial = runInterstitialRules(options, [
+      const signedOutOrInterstitial = await runInterstitialRules(options, [
         nonBrowserRequestInDevRule,
         crossOriginRequestWithoutHeader,
         potentialFirstLoadInDevWhenUATMissing,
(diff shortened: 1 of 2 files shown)


nikosdouvlis (approved):
Excellent! FYI [this PR](https://github.com/clerkinc/javascript/pull/674) might introduce a couple of minor conflicts (I think we've moved a `try` from a try/catch clause higher up), but this looks pretty good to me :)

🙃 On a separate note, and since you asked me to be extremely nitpicky :) Since this is a refactor that, in theory can affect runtime behavior, do you think we should change the `chore` prefix to maybe `fix`? That way, a PATCH version bump will happen even if the change is merged/published by itself. Of course, this methodology is debatable, and I'm pretty unsure whether it's useful or not!

If we end up changing it to `fix`, we should also add the correct scope, eg `fix(backend): ...`

packages/backend/src/tokens/request.ts
```diff
@@ -56,114 +57,38 @@ export type AuthenticateRequestOptions = RequiredVerifyTokenOptions &
...
-      }
+      const state = await runInterstitialRules(options, [
+        nonBrowserRequestInDevRule,
+        crossOriginRequestWithoutHeader,
+        potentialFirstLoadInDevWhenUATMissing,
+        potentialRequestAfterSignInOrOurFromClerkHostedUiInDev,
+        potentialFirstRequestOnProductionEnvironment,
+        isNormalSignedOutState,
+        hasClientUatButCookieIsMissingInProd,
+        hasValidCookieToken,
```
SokratisVidros, line 83: ❓ Does this replace the check from lines 148-150?

SokratisVidros (approved):
😍

PR #672, opened 2023-01-14
fix(clerk-js): Account for user clock skew in TokenCache

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

getToken() makes the decision about returning from cache by comparing the exp claim on the JWT against the user's computer time. exp is generated on our server (GCP's clock), not the computer's time.

We're accounting for slow/fast clocks by measuring the elapsed time since the token was cached, instead of comparing GCP's clock with the user's clock.

### DIFF
--- a/packages/clerk-js/src/core/tokenCache.ts
+++ b/packages/clerk-js/src/core/tokenCache.ts
@@ -1,7 +1,5 @@
 import type { TokenResource } from '@clerk/types';
 
-import { clerkCoreErrorExpiredToken } from './errors';
-
 interface TokenCacheKeyJSON {
   audience?: string;
   tokenId: string;
@@ -11,10 +9,12 @@ interface TokenCacheEntry extends TokenCacheKeyJSON {
   tokenResolver: Promise<TokenResource>;
 }
 
+type Seconds = number;
+
 interface TokenCacheValue {
   entry: TokenCacheEntry;
-  expiresAt?: number;
-  expiresIn?: number;
+  createdAt: Seconds;
+  expiresIn?: Seconds;
 }
 
 interface TokenCache {
@@ -63,7 +63,9 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
     });
 
     const key = cacheKey.toKey();
-    const value: TokenCacheValue = { entry };
+
+    const createdAt = Math.floor(Date.now() / 1000);
+    const value: TokenCacheValue = { entry, createdAt };
 
     const deleteKey = () => {
       if (cache.get(key) === value) {
@@ -73,16 +75,11 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
 
     entry.tokenResolver
       .then(newToken => {
-        const nowSeconds = Math.floor(Date.now() / 1000);
         const expiresAt = newToken.jwt.claims.exp;
-        const expiresIn = expiresAt - nowSeconds;
-
-        if (expiresIn <= 0) {
-          clerkCoreErrorExpiredToken(expiresAt);
-        }
+        const issuedAt = newToken.jwt.claims.iat;
+        const expiresIn = expiresAt - issuedAt;
 
         // Mutate cached value and set expirations
-        value.expiresAt = expiresAt;
         value.expiresIn = expiresIn;
         setTimeout(deleteKey, expiresIn * 1000);
       })
@@ -102,9 +99,10 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
     }
 
     const nowSeconds = Math.floor(Date.now() / 1000);
-    const willExpireSoon = value.expiresAt && value.expiresAt - leeway < nowSeconds;
+    const elapsedSeconds = nowSeconds - value.createdAt;
+    const expiresSoon = value.expiresIn! - elapsedSeconds < leeway;
 
-    if (willExpireSoon) {
+    if (expiresSoon) {
       cache.delete(cacheKey.toKey());
       return;
     }
--- a/packages/clerk-js/src/core/tokenCache.ts
+++ b/packages/clerk-js/src/core/tokenCache.ts
@@ -27,6 +27,8 @@ interface TokenCache {
 const KEY_PREFIX = 'clerk';
 const DELIMITER = '::';
 const LEEWAY = 10;
+// This value should have the same value as the INTERVAL_IN_MS in AuthenticationPoller
+const SYNC_LEEWAY = 5;
 
 export class TokenCacheKey {
   static fromKey(key: string): TokenCacheKey {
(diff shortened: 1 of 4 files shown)


packages/clerk-js/src/core/tokenCache.ts
```diff
@@ -75,15 +75,15 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
...
-        const expiresIn = expiresAt - nowSeconds;
-
-        if (expiresIn <= 0) {
-          clerkCoreErrorExpiredToken(expiresAt);
-        }
+        const issuedAt = newToken.jwt.claims.iat;
+        const ttl = expiresAt - issuedAt;
+        // if user clock is running slow, expiresAt - nowSeconds will be bigger than the ttl
+        // if user clock is running , expiresAt - nowSeconds will be less than the ttl
+        const expiresIn = expiresAt - nowSeconds <= 0 ? ttl : Math.min(expiresAt - nowSeconds, ttl);
```
colinclerk, line 82: My instinct is that `expiresIn` should always be `ttl`, otherwise a clock running fast will refresh too frequently.

Also, it might be safer to use the time the promise was initiated instead of when it was resolved. wdyt?

anagstef (approved):

packages/clerk-js/src/core/tokenCache.ts
```diff
@@ -102,9 +99,10 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
     }
 
     const nowSeconds = Math.floor(Date.now() / 1000);
-    const willExpireSoon = value.expiresAt && value.expiresAt - leeway < nowSeconds;
```
SokratisVidros, line 105: @nikosdouvlis I feel that expiresAt and expiresIn variables were easier to read and follow in the algorithm. They also map better to the JWT lingo that drives the algorithm. 

What was the bug with the original algorithm?

SokratisVidros (changes requested):

packages/clerk-js/src/core/tokenCache.ts
nikosdouvlis: We used to compare the `expiresAt` that was generated on GCP with the user's own clock. That led to the clock skew issue that a user reported on intercom. Their clock was about 1minute behind GCP, so the `expiresAt - nowSeconds;` was always more than the token TTL.

According to your comment, we can define expiresAt as such:
```
value.expiresAt = nowSeconds  + ttl; 
```
We calculate the TTL and then we add it to the time the request for the new token was made. Does this make more sense to you?

dimkl: @SokratisVidros PTAL (review it per commit, it will make more sense)

packages/clerk-js/src/core/tokenCache.ts
```diff
@@ -63,7 +65,9 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
     });
 
     const key = cacheKey.toKey();
-    const value: TokenCacheValue = { entry };
+
+    const createdAt = Math.floor(Date.now() / 1000);
+    const value: TokenCacheValue = { entry, createdAt };
 
     const deleteKey = () => {
       if (cache.get(key) === value) {
```
dimkl, line 73: 🙃 i'm not a big fun of this check since it compares 2 `TokenCacheValue` objects and this equality operator is checking them not by value but by reference! I think we could something better here.

nikosdouvlis: Hey Dimitri, I might be wrong but I think this is meant to compare by reference, actually, but it's possible that we no longer need it. 

Keep in mind that before the setActive PR that we merged yesterday, the keys did not include the updateAt value. The `set` method schedules a delete call `expiresIn` seconds after caching the entry, while `get` might actually delete and cache an entry `leeway` seconds before the delete timeout is called.

The equality check in essence means that calling `deleteKey` will only delete the entry that scheduled the call and only that, even if a new entry with the same key was stored afterward.

LMK if this makes sense :) 

However, now that we do have the updatedAt in the key, let's think about whether we need to keep the check or not. I think i'd keep it, as the key is produced outside of that class so no one enforces that all entries will have different keys. Ofc, we can always refactor this logic and remove the check. LMK your thoughts

SokratisVidros (approved):
💯

packages/clerk-js/src/core/tokenCache.ts
```diff
@@ -73,16 +77,11 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
 
     entry.tokenResolver
       .then(newToken => {
-        const nowSeconds = Math.floor(Date.now() / 1000);
         const expiresAt = newToken.jwt.claims.exp;
-        const expiresIn = expiresAt - nowSeconds;
-
-        if (expiresIn <= 0) {
-          clerkCoreErrorExpiredToken(expiresAt);
```
nikosdouvlis, line 81: Lets remove this if we no longer use it from anywhere else :)

packages/clerk-js/src/core/tokenCache.ts
```diff
@@ -73,16 +77,11 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
...
-        }
+        const issuedAt = newToken.jwt.claims.iat;
+        const expiresIn = expiresAt - issuedAt;
```
nikosdouvlis, line 82:
```suggestion
        const expiresInSeconds = expiresAt - issuedAt;
```
or
```suggestion
        const expiresIn: Seconds = expiresAt - issuedAt;
```
or simply ignore as this is already fine :)

packages/clerk-js/src/core/tokenCache.ts
```diff
@@ -27,6 +27,8 @@ interface TokenCache {
 const KEY_PREFIX = 'clerk';
 const DELIMITER = '::';
 const LEEWAY = 10;
+// This value should have the same value as the INTERVAL_IN_MS in AuthenticationPoller
+const SYNC_LEEWAY = 5;
```
nikosdouvlis, line 31: 🙃 In the future we should either import this from the poller, or extracts all these constants in a separate file. We lose the benefits of code colocation, but I think the tradeoff is worth it

nikosdouvlis:
Excellent @dimkl

PR #682, opened 2023-01-17
fix(clerk-react): Add data-clerk-publishable-key attribute only when …

nikosdouvlis:
…PK is available

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/backend-core`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/edge`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/packages/react/src/types.ts
+++ b/packages/react/src/types.ts
@@ -21,16 +21,10 @@ export type IsomorphicClerkOptions = ClerkOptions & {
   clerkJSVariant?: 'headless' | '';
 } & (
     | {
-        /**
-         * @deprecated Use `publishableKey` instead
-         */
         frontendApi?: never;
         publishableKey: string;
       }
     | {
-        /**
-         * @deprecated Use `publishableKey` instead
-         */
         frontendApi: string;
         publishableKey?: never;
       }
--- a/packages/react/src/utils/scriptLoader.ts
+++ b/packages/react/src/utils/scriptLoader.ts
@@ -38,13 +38,8 @@ function getScriptSrc({ publishableKey, frontendApi, scriptUrl, scriptVariant = 
     return scriptUrl;
   }
 
-  let scriptHost: string;
-  if (publishableKey) {
-    const { frontendApi } = parsePublishableKey(publishableKey)!;
-    scriptHost = frontendApi;
-  } else if (frontendApi) {
-    scriptHost = frontendApi;
-  } else {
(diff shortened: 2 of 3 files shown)


packages/react/src/types.ts
```diff
@@ -21,16 +21,10 @@ export type IsomorphicClerkOptions = ClerkOptions & {
   clerkJSVariant?: 'headless' | '';
 } & (
     | {
-        /**
-         * @deprecated Use `publishableKey` instead
```
nikosdouvlis, line 25: Removed the deprecation warning as Kima is not enabled yet

packages/react/src/utils/scriptLoader.ts
```diff
@@ -83,8 +78,12 @@ export async function loadScript(params: LoadScriptParams): Promise<HTMLScriptEl
...
 
     const script = document.createElement('script');
-    script.setAttribute('data-clerk-publishable-key', publishableKey!);
-    script.setAttribute('data-clerk-frontend-api', frontendApi!);
+    if (publishableKey) {
+      script.setAttribute('data-clerk-publishable-key', publishableKey);
+    }
+    if (frontendApi) {
+      script.setAttribute('data-clerk-frontend-api', frontendApi);
+    }
```
SokratisVidros, line 86: 🔧 Let's throw on error if non of them is set at this point via the errorThrower.

SokratisVidros (approved):

colinclerk (approved):

packages/react/src/utils/scriptLoader.ts
nikosdouvlis: Thought of that as well, but it felt that this is too deep for an error as we already check a couple of times. Im going to add one as per your suggestion, better be on the safe side

nikosdouvlis: Done

agis (approved):

PR #692, opened 2023-01-19
Fix: Header token case of unknown 401

dimkl:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

We have introduced the new `Unknown` state in `RequestState` and the `isUnknown` helper method to support the case of header expired token that should result in returning an http 401 without body before running any other middleware.

The reason we introduced another state instead of reusing the `Interstitial` as i initially suggested in the original discussion is to separate this case and avoid any more confusion about interstitial. With introducing the new state we are closer to the flow diagram and it is more clear to implement and understand. This is also easier to rollout since it's a new state and it will not affect the current implementation.

### DIFF
--- a/packages/backend/src/tokens/authStatus.ts
+++ b/packages/backend/src/tokens/authStatus.ts
@@ -9,6 +9,7 @@ export enum AuthStatus {
   SignedIn = 'signed-in',
   SignedOut = 'signed-out',
   Interstitial = 'interstitial',
+  Unknown = 'unknown',
 }
 
 export type SignedInState = {
@@ -19,6 +20,7 @@ export type SignedInState = {
   publishableKey: string;
   isSignedIn: true;
   isInterstitial: false;
+  isUnknown: false;
   toAuth: () => SignedInAuthObject;
 };
 
@@ -30,15 +32,20 @@ export type SignedOutState = {
   publishableKey: string;
   isSignedIn: false;
   isInterstitial: false;
+  isUnknown: false;
   toAuth: () => SignedOutAuthObject;
 };
 
 export type InterstitialState = Omit<SignedOutState, 'isInterstitial' | 'status' | 'toAuth'> & {
   status: AuthStatus.Interstitial;
-  isSignedIn: false;
   isInterstitial: true;
-  reason: AuthReason;
   toAuth: () => null;
+};
+
+export type UnknownState = Omit<InterstitialState, 'status' | 'isInterstitial' | 'isUnknown'> & {
+  status: AuthStatus.Unknown;
+  isInterstitial: false;
+  isUnknown: true;
 };
 
 export enum AuthErrorReason {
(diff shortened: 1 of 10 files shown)


packages/sdk-node/src/clerkExpressWithAuth.ts
```diff
@@ -20,7 +20,7 @@ export const createClerkExpressWithAuth = (createOpts: CreateClerkExpressMiddlew
         options,
       );
       decorateResponseWithObservabilityHeaders(res, requestState);
-      if (requestState.isInterstitial) {
+      if (requestState.isInterstitial || requestState.isUnknown) {
```
nikosdouvlis, line 23: ✅ Great

packages/backend/src/tokens/authStatus.ts
```diff
@@ -30,17 +32,22 @@ export type SignedOutState = {
...
 
 export type InterstitialState = Omit<SignedOutState, 'isInterstitial' | 'status' | 'toAuth'> & {
   status: AuthStatus.Interstitial;
-  isSignedIn: false;
   isInterstitial: true;
-  reason: AuthReason;
   toAuth: () => null;
 };
 
+export type UnknownState = Omit<InterstitialState, 'status' | 'isInterstitial' | 'isUnknown'> & {
```
nikosdouvlis, line 45: Do you think it'd be a good idea to include a short comment about what this case is supposed to do? I'm talking about the "prevent user code from running" as discussed offline. However, I'm not sure whether it'd be better as a comment,  a separate doc or a combination of both.

nikosdouvlis (approved):
Adding a bit of context to this PR mainly for future reference. Another reason @dimkl decided to go with the new `Unknown` state is so that every leaf node of the diagram should actually be a leaf - splitting the interstitial in 2 different sub flows would increase the mental overhead as well as the code complexity, as for every `isInterstitial` check, we would need another nested condition. 

Pulling the condition a level higher and only when its needed makes total sense to me :)
Very nice!

SokratisVidros:
> Adding a bit of context to this PR mainly for future reference. Another reason @dimkl decided to go with the new `Unknown` state is so that every leaf node of the diagram should actually be a leaf - splitting the interstitial in 2 different sub flows would increase the mental overhead as well as the code complexity, as for every `isInterstitial` check, we would need another nested condition.
> 
> Pulling the condition a level higher and only when its needed makes total sense to me :) Very nice!

Good call. Let's also update our internal docs.

packages/backend/src/tokens/authStatus.ts
dimkl: Let's add it in the docs (as discussed offline).

PR #695, opened 2023-01-20
fix(clerk-react): Fix erroneously throwing an error when using the new Publishable Key

anagstef:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

This PR fixes the issue: 
When parsing the new Publishable Key, its Frontend API gets extracted and populated in the legacy `frontendApi` variable. This erroneously throws an error. Added an extra check to prevent this when PK is used.

### DIFF
--- a/packages/react/src/contexts/ClerkProvider.tsx
+++ b/packages/react/src/contexts/ClerkProvider.tsx
@@ -26,7 +26,7 @@ function ClerkProviderBase(props: ClerkProviderProps): JSX.Element {
       errorThrower.throwMissingFrontendApiOrPublishableKeyError();
     } else if (publishableKey && !isPublishableKey(publishableKey)) {
       errorThrower.throwInvalidPublishableKeyError({ key: publishableKey });
-    } else if (frontendApi && !isLegacyFrontendApiKey(frontendApi)) {
+    } else if (!publishableKey && frontendApi && !isLegacyFrontendApiKey(frontendApi)) {
       errorThrower.throwInvalidFrontendApiError({ key: frontendApi });
     }
   }


panteliselef (approved):

nikosdouvlis (commented):
This makes sense to me! GJ

Let's get a confirmation from @chanioxaris as well :)

nikosdouvlis (approved):

packages/react/src/contexts/ClerkProvider.tsx
```diff
@@ -26,7 +26,7 @@ function ClerkProviderBase(props: ClerkProviderProps): JSX.Element {
       errorThrower.throwMissingFrontendApiOrPublishableKeyError();
     } else if (publishableKey && !isPublishableKey(publishableKey)) {
       errorThrower.throwInvalidPublishableKeyError({ key: publishableKey });
-    } else if (frontendApi && !isLegacyFrontendApiKey(frontendApi)) {
+    } else if (!publishableKey && frontendApi && !isLegacyFrontendApiKey(frontendApi)) {
```
nikosdouvlis, line 29: ❓ At which point do we populate `frontendApi` here?

chanioxaris: Yes same, i am curious why the frontendApi will be populated here if someone pass a publishableKey

anagstef: This comes from backend actually. It has to do with SSR.

Here it is:
https://github.com/clerkinc/javascript/blob/20b7f0934a99923961e930a3e3519ace70cd9457/packages/backend/src/tokens/request.ts#L63

anagstef: The error we get comes from the server

chanioxaris: That makes sense, thanks

chanioxaris (approved):

PR #703, opened 2023-01-23
feat(backend): Use Frontend API to fetch JWKs for token verification when issuer provided

chanioxaris:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

Slightly modified the tokens.verifyToken() function into the below behavior:
- If a JWK KEY provided, use it to verify the provided token
- If a apiKey or secretKey provided, use the Backend API to fetch the JWKs and verify the provided token
- If an issuer of type string provided, use the Frontend API and well known endpoint to fetch the JWKs and verify the provided token

### DIFF
--- a/packages/backend/src/tokens/keys.ts
+++ b/packages/backend/src/tokens/keys.ts
@@ -114,8 +114,8 @@ export type LoadClerkJWKFromRemoteOptions = {
 export async function loadClerkJWKFromRemote({
   apiKey,
   secretKey,
-  apiUrl,
-  apiVersion,
+  apiUrl = API_URL,
+  apiVersion = API_VERSION,
   issuer,
   kid,
   jwksCacheTtlInMs = 1000 * 60 * 60, // 1 hour,
@@ -124,9 +124,9 @@ export async function loadClerkJWKFromRemote({
   const shouldRefreshCache = !getFromCache(kid) && reachedMaxCacheUpdatedAt();
   if (skipJwksCache || shouldRefreshCache) {
     let fetcher;
-    const key = secretKey || apiKey || '';
+    const key = secretKey || apiKey;
 
-    if (apiUrl) {
+    if (key) {
       fetcher = () => fetchJWKSFromBAPI(apiUrl, key, apiVersion);
     } else if (issuer) {
       fetcher = () => fetchJWKSFromFAPI(issuer);
@@ -181,7 +181,7 @@ async function fetchJWKSFromFAPI(issuer: string) {
   return response.json();
 }
 
-async function fetchJWKSFromBAPI(apiUrl: string = API_URL, key: string, apiVersion: string = API_VERSION) {
+async function fetchJWKSFromBAPI(apiUrl: string, key: string, apiVersion: string) {
   if (!key) {
     throw new TokenVerificationError({
       action: TokenVerificationErrorAction.SetClerkSecretKeyOrAPIKey,
--- a/packages/backend/src/tokens/verify.ts
+++ b/packages/backend/src/tokens/verify.ts
@@ -36,16 +36,12 @@ export async function verifyToken(token: string, options: VerifyTokenOptions): P
 
   if (jwtKey) {
     key = loadClerkJWKFromLocal(jwtKey);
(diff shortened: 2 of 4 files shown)


packages/backend/src/tokens/verify.ts
```diff
@@ -36,16 +36,12 @@ export async function verifyToken(token: string, options: VerifyTokenOptions): P
...
-    //
-    // Currently JWKS are fetched from Backend API without using the JWT issuer. This should change
-    // with the new Backend key format so that the API Key is not required for token verification and
-    // the jwks retrieval is driven from the current JWT header.
-    const opts = apiKey
-      ? ({ apiUrl, apiKey, apiVersion, kid, jwksCacheTtlInMs, skipJwksCache } as LoadClerkJWKFromRemoteOptions)
-      : ({ apiUrl, secretKey, apiVersion, kid, jwksCacheTtlInMs, skipJwksCache } as LoadClerkJWKFromRemoteOptions);
-    key = await loadClerkJWKFromRemote(opts);
+    // Fetch JWKS from Backend API using the key
+    key = await loadClerkJWKFromRemote({ apiKey, secretKey, apiUrl, apiVersion, kid, jwksCacheTtlInMs, skipJwksCache });
```
raptisj, line 44: @chanioxaris don’t we need to pass the issuer here as well?

packages/backend/src/tokens/keys.ts
```diff
@@ -121,9 +121,9 @@ export async function loadClerkJWKFromRemote({
   const shouldRefreshCache = !getFromCache(kid) && reachedMaxCacheUpdatedAt();
   if (skipJwksCache || shouldRefreshCache) {
     let fetcher;
-    const key = secretKey || apiKey || '';
+    const key = secretKey || apiKey;
 
-    if (apiUrl) {
+    if (key) {
```
SokratisVidros, line 126: 💯 Much better.

packages/backend/src/tokens/verify.ts
SokratisVidros: There is no need as we added it in a separate conditional clause

packages/backend/src/tokens/verify.ts
```diff
@@ -36,16 +36,12 @@ export async function verifyToken(token: string, options: VerifyTokenOptions): P
 
   if (jwtKey) {
     key = loadClerkJWKFromLocal(jwtKey);
+  } else if (typeof issuer === 'string') {
+    // Fetch JWKS from Frontend API if an issuer of type string has been provided
+    key = await loadClerkJWKFromRemote({ issuer, kid, jwksCacheTtlInMs, skipJwksCache });
```
SokratisVidros, line 41: @chanioxaris If not mistaken, the default issuer we currently pass is a function object, right? 

If the issuer is passed as a function during the verifyToken invocation, is it a function at this point or has the function executed and returned a string already?

packages/backend/src/tokens/verify.ts
raptisj: @SokratisVidros @chanioxaris sorry I still don't get it. How does the fetchJWKSFromFAPI get the issuer?

packages/backend/src/tokens/verify.ts
chanioxaris: Yes that's right the issuer can be a string or a function that returns a boolean value. We will only land in this statement if the provided issuer is of type string, so we can construct the well known url. If the issuer instead is a function we will fallback to other cases (Backend API, JWT key)

dimkl (approved):

PR #707, opened 2023-01-23
feat(clerk-js): Redirect to home when an unavailable component is mounted

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description


packages/clerk-js/src/ui/common/withRedirectToHome.tsx
```diff
@@ -1,35 +1,67 @@
...
-    const { authConfig, displayConfig } = useEnvironment();
-    const { singleSessionMode } = authConfig;
-    const ctx = useContext(CoreSessionContext);
-    const session = ctx?.value;
+    const clerk = useCoreClerk();
+    const environment = useEnvironment();
+
+    let shouldRedirect = false;
+    let warning = '';
+    switch (when) {
```
SokratisVidros, line 22: ⛔ We can make this method more dry by making each Component aware of whether it can be rendered. That is, let's introduce a `Component.shouldRender` static method that accepts (clerk, and environment) and calls that method here instead of doing a switch statement. That way, you will also avoid the second string parameter passed `withRedirectToHome` representing the component name.

packages/clerk-js/src/ui/components/SignIn/SignInFactorOne.tsx
```diff
@@ -117,4 +117,4 @@ export function _SignInFactorOne(): JSX.Element {
   }
 }
 
-export const SignInFactorOne = withRedirectToHome(withCardStateProvider(_SignInFactorOne));
+export const SignInFactorOne = withRedirectToHome(withCardStateProvider(_SignInFactorOne), 'signIn');
```
SokratisVidros, line 120: Refer to my previous comment to avoid passing `signIn`.

SokratisVidros (changes requested):

packages/clerk-js/src/ui/common/withRedirectToHome.tsx
```diff
@@ -1,35 +1,67 @@
...
-import { useEnvironment } from '../contexts';
-import { CoreSessionContext } from '../contexts/CoreSessionContext';
+import { canShowOrganizationProfile, canShowSignIn, canShowSignUp, canShowUserProfile } from '../../utils';
+import { useCoreClerk, useEnvironment } from '../contexts';
 import { useNavigate } from '../hooks';
 
 export function withRedirectToHome<P extends SignInProps | SignUpProps | { continueExisting?: boolean }>(
   Component: React.ComponentType<P>,
-  displayName?: string,
+  when: 'signIn' | 'signUp' | 'userProfile' | 'organizationProfile',
```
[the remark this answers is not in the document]
nikosdouvlis: If we still need this type after the refactorings discussed below, I'd look at the `appearance` or `Component.tsx` types because I'm almost sure we have this already

packages/clerk-js/src/ui/common/withRedirectToHome.tsx
nikosdouvlis: I'd suggest something along the lines of what Sokratis said above, as this `switch` statement here can be considered a code smell: when a new component is added, we'll have to find all similar `switch` statements and update them.

A different, more "FP-oriented" approach would be to pass a condition as an argument to the `withRedirectToHome` HOC, creating a more specific HOC as a result. eg;

```
const withRedirectToHomeUserGuard = (...args) => withRedirectToHome(condition, ...args)
const withRedirectToHomeOrganizationGuard = (...args) => withRedirectToHome(condition, ...args)
```

The names are terrible, I know, the rest is up to you :)

desiprisg: @SokratisVidros We would have to create a lot of `shouldRender` functions, since we use the HOC in most of the routes for `SignIn`, but not all. It does not wrap the `SignIn` but rather some of its route components.

desiprisg: Also, is there a reason we can't make this check in the `mount` methods? What's preventing us from doing that?

desiprisg: I did what @nikosdouvlis proposed since it is a simpler approach for now (because we use the HOC many times) and I think it scales well. Please take a look.

packages/clerk-js/src/ui/common/withRedirectToHome.tsx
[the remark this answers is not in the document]
desiprisg: The type introduced is that of the requirement that needs to be met, so I'm resolving this.

packages/clerk-js/src/utils/meetsRequirement.ts
```diff
@@ -0,0 +1,34 @@
+import { Clerk, ClerkOptions, EnvironmentResource } from '@clerk/types';
+
+export type Requirements = 'singleSession' | 'user' | 'organization';
+export type MeetsRequirement = (
+  clerk: Clerk,
+  environment?: EnvironmentResource | null,
+  options?: ClerkOptions,
+) => boolean;
+
+export const meetsRequirement: Record<Requirements, MeetsRequirement> = {
```
nikosdouvlis, line 10: I definitely like the idea! Couple of comments:

1. You want to add a namespace over a group of exports, rather than using an explicit type and the dictionary, you can just: 

```
const hello = () => {}
const there = () => {}
export const ClicheStarwarsMeme = { hello, there }

// somewhere else
import { ClicheStarwarsMeme } from './'
ClicheStarwarsMeme.hello()
```

2. The `meetsRequirement` namespace makes sense when reading in in the context of the calling function, but does it make sense as a module on its own? What does `import { meetsRequirements }` mean? I think its a little generic. That's why **I think** I preferred your previous approach with the standalone, composable helpers 

LMK what you think!

packages/clerk-js/src/ui/common/withRedirectToHome.tsx
```diff
@@ -1,39 +1,63 @@
...
 }
+
+export const withRedirectToHomeSingleSessionGuard = <P extends AvailableComponentProps>(Component: ComponentType<P>) =>
+  withRedirectToHome(
+    Component,
+    meetsRequirement.singleSession,
+    'Cannot render the component as a session already exists and single session mode is enabled. Redirecting to the home url.',
+  );
+
+export const withRedirectToHomeUserGuard = <P extends AvailableComponentProps>(Component: ComponentType<P>) =>
```
nikosdouvlis, line 51: nice 💯

packages/clerk-js/src/utils/clerkRequirements.ts
```diff
@@ -0,0 +1,22 @@
+import { Clerk, ClerkOptions, EnvironmentResource } from '@clerk/types';
+
+export type ClerkRequirement = (
```
nikosdouvlis, line 3: Note sure if that'd make more sense:

```suggestion
export type ComponentGuardCondition = (
```

packages/clerk-js/src/ui/types.ts
```diff
@@ -1,15 +1,31 @@
...
+  SignUpProps,
+  UserButtonProps,
+  UserProfileProps,
+  OrganizationSwitcherProps,
+  OrganizationProfileProps,
+  CreateOrganizationProps,
+};
 
-export type AvailableComponentProps = SignInProps | SignUpProps | UserProfileProps | UserButtonProps;
+export type AvailableComponentProps =
```
nikosdouvlis, line 21: GJ catching that

packages/clerk-js/src/core/clerk.ts
```diff
@@ -209,6 +212,9 @@ export default class Clerk implements ClerkInterface {
 
   public openSignIn = (props?: SignInProps): void => {
     this.assertComponentsReady(this.#componentControls);
+    if (sessionExistsAndSingleSessionModeEnabled(this, this.#environment) && this.#instanceType === 'development') {
```
nikosdouvlis, line 215: Maybe extract the shared condition into a concrete method
```suggestion
    if (sessionExistsAndSingleSessionModeEnabled(this, this.#environment) && this.#isDevInstance() {
```

nikosdouvlis (approved):

packages/clerk-js/src/utils/clerkRequirements.ts
desiprisg: Done

packages/clerk-js/src/core/clerk.ts
desiprisg: We already use this logic in 4-5 other places so I'd prefer not to refactor all of these in this PR.

desiprisg:
@SokratisVidros Can I get a review and the change requests lifted if all is ok with you?

packages/clerk-js/src/utils/componentGuardConditions.ts
```diff
@@ -0,0 +1,22 @@
+import { Clerk, ClerkOptions, EnvironmentResource } from '@clerk/types';
```
SokratisVidros, line 1: Nitpicky one: `componentGuards`

SokratisVidros (approved):
💯

PR #709, opened 2023-01-23
COR-51 fix(backend): Add proxyUrl prop

raptisj:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

Add proxyUrl prop and refactor verifyToken function so it can be called with the proxy url if it is present

### DIFF
--- a/packages/backend/src/tokens/interstitial.ts
+++ b/packages/backend/src/tokens/interstitial.ts
@@ -11,13 +11,14 @@ import { TokenVerificationError, TokenVerificationErrorAction, TokenVerification
 export type LoadInterstitialOptions = {
   apiUrl: string;
   frontendApi: string;
+  proxyUrl?: string;
   publishableKey: string;
   pkgVersion?: string;
   debugData?: any;
 };
 
 export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions, 'apiUrl'>) {
-  const { debugData, frontendApi, pkgVersion, publishableKey } = options;
+  const { debugData, frontendApi, pkgVersion, publishableKey, proxyUrl } = options;
   return `
     <head>
         <meta charset="UTF-8" />
@@ -26,6 +27,7 @@ export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions,
         <script>
             window.__clerk_frontend_api = '${frontendApi}';
             window.__clerk_debug = ${JSON.stringify(debugData || {})};
+            window.__clerk_proxy_url = '${proxyUrl}';
             window.startClerk = async () => {
                 function formRedirect(){
                     const form = '<form method="get" action="" name="redirect"></form>';
--- a/packages/backend/src/tokens/interstitial.ts
+++ b/packages/backend/src/tokens/interstitial.ts
@@ -100,7 +100,7 @@ export async function loadInterstitialFromBAPI(options: LoadInterstitialOptions)
 
 export function buildPublicInterstitialUrl(options: LoadInterstitialOptions) {
   options.frontendApi = parsePublishableKey(options.publishableKey)?.frontendApi || options.frontendApi || '';
-  const { apiUrl, frontendApi, pkgVersion, publishableKey } = options;
+  const { apiUrl, frontendApi, pkgVersion, publishableKey, proxyUrl } = options;
   const url = new URL(apiUrl);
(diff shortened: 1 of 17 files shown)


packages/backend/src/tokens/interstitialRule.ts
```diff
@@ -125,11 +125,11 @@ export async function runInterstitialRules<T>(opts: T, rules: InterstitialRule[]
   return signedOut<T>(opts, AuthErrorReason.UnexpectedError);
 }
 
-async function verifyRequestState(options: any, token: string) {
-  const issuer = (iss: string) => iss.startsWith('https://clerk.') || iss.includes('.clerk.accounts');
+async function verifyRequestState(options: any, token: string, issuer = '') {
```
panteliselef, line 128: ❓ Is this relevant to this PR ? Maybe I am missing some context here

packages/remix/src/ssr/types.ts
```diff
@@ -10,6 +10,7 @@ export type RootAuthLoaderOptions = {
   frontendApi?: string;
   publishableKey?: string;
   jwtKey?: string;
+  proxyUrl?: string;
```
panteliselef, line 13: ❓Doesn't this expose the proxyUrl to user when calling `getAuth` & `rootAuthLoader` and passing options ?

panteliselef (changes requested):

packages/backend/src/tokens/interstitialRule.ts
panteliselef: In the description you are mentioning "refactor verifyToken function"
But is this necessary for this PR ?

raptisj: @panteliselef I updated the description. Yes, this is part of this PR. We need to use the proxyUrl if present during verification since it makes a request to FAPI.

panteliselef: @raptisj go it

packages/remix/src/ssr/types.ts
raptisj: @panteliselef I omitted it in `rootAuthLoader`. In `getAuth` we pick certain values so we are ok

panteliselef (changes requested):
Looks good to me

packages/backend/src/tokens/interstitial.ts
```diff
@@ -67,6 +69,7 @@ export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions,
                     ? `script.setAttribute('data-clerk-publishable-key', '${publishableKey}');`
                     : `script.setAttribute('data-clerk-frontend-api', '${frontendApi}');`
                 }
+                script.setAttribute('data-clerk-proxy-url', '${proxyUrl}');
```
panteliselef, line 72: ❓ Should this line be added if proxyUrl is undefined ?

packages/backend/src/tokens/interstitialRule.ts
```diff
@@ -126,11 +126,11 @@ export async function runInterstitialRules<T>(opts: T, rules: InterstitialRule[]
...
```
SokratisVidros, line 129: ⛔ My recommendation is to avoid changing the signature of `verifyRequestState` method and pass the issuer as another key in the options. The options argument was introduced to cater for such cases.

SokratisVidros (changes requested):

packages/backend/src/tokens/interstitialRule.ts
raptisj: @SokratisVidros so you recommend removing the `issuer` param from `verifyRequestState` and using the `option.proxyurl` straight away right?

SokratisVidros: Yes

PR #713, opened 2023-01-24
fix(backend,types): Remove inline type imports

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

Inline type imports were introduced in TS 4.5, which, at the moment of writing, can be considered a recent release. Customers using older TS/babel versions reported build errors when trying to compile the new clerk/backend package. We decided to replace inline type imports with the older syntax as we have no apparent need for it right now.

### DIFF
--- a/packages/backend/src/tokens/factory.ts
+++ b/packages/backend/src/tokens/factory.ts
@@ -1,14 +1,14 @@
-import { type ApiClient } from '../api';
+import { ApiClient } from '../api';
 import { API_URL, API_VERSION } from '../constants';
 import {
-  type LoadInterstitialOptions,
   buildPublicInterstitialUrl,
   loadInterstitialFromBAPI,
   loadInterstitialFromLocal,
+  LoadInterstitialOptions,
 } from './interstitial';
 import {
-  type AuthenticateRequestOptions,
   authenticateRequest as authenticateRequestOriginal,
+  AuthenticateRequestOptions,
   debugRequestState,
 } from './request';
 
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -1,5 +1,5 @@
-import { type CreateBackendApiOptions, createBackendApiClient } from './api';
-import { type CreateAuthenticateRequestOptions, createAuthenticateRequest } from './tokens';
+import { createBackendApiClient, CreateBackendApiOptions } from './api';
+import { createAuthenticateRequest, CreateAuthenticateRequestOptions } from './tokens';
 
 export * from './api/resources';
 export * from './tokens';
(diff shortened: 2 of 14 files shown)


packages/backend/src/tokens/factory.ts
```diff
@@ -1,4 +1,4 @@
-import { type ApiClient } from '../api';
+import { ApiClient } from '../api';
 import { API_URL, API_VERSION } from '../constants';
 import {
   type LoadInterstitialOptions,
```
desiprisg, line 4: you missed one :)

nikosdouvlis: Thanks @desiprisg , I actually missed 4 :)

desiprisg: Can you build locally with an older ts version be 100% sure?

SokratisVidros (approved):

gkats (approved):

anagstef:
Maybe we should add a ts lint rule not to allow types import?

packages/backend/src/tokens/factory.ts
nikosdouvlis: Correct :) That's how I found the other 4

nikosdouvlis:
> Maybe we should add a ts lint rule not to allow types import?
> 
> https://typescript-eslint.io/rules/consistent-type-imports/

Good thinking :) TBH I'm not convinced that we don't want to use type imports. We are removing them for now to avoid further support confusion, but I think we can introduce them again in the future.

Moreover, that rule only checks the `import type { }` syntax, not the problematic `import { type }` one, so I'm not sure whether it's going to help or not. 

BTW, let's discuss offline if you have other eslint rule suggestion that could be useful

nikosdouvlis:
> > Maybe we should add a ts lint rule not to allow types import?
> > https://typescript-eslint.io/rules/consistent-type-imports/
> 
> Good thinking :) TBH I'm not convinced that we don't want to use type imports. We are removing them for now to avoid further support confusion, but I think we can introduce them again in the future.
> 
> Moreover, that rule only checks the `import type { }` syntax, not the problematic `import { type }` one, so I'm not sure whether it's going to help or not.
> 
> BTW, let's discuss offline if you have other eslint rule suggestion that could be useful

I was too quick to merge.
@anagstef corrected me offline as the suggested rule  can transform `import { type }` to `import type {}` for us, so lets add it :)

PR #719, opened 2023-01-24
Fix(clerk-js): setActive issue with cookie and session syncing

dimkl:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

This tickets tries to resolve cookie and session syncing issues found in setActive upon user and organization switching.

### Issue

Steps to reproduce:
1. user with multiple organizations signin
2. visit `/organization` page from tab A with organization A
3. open Application tab in Developer tools and find `__session` cookie
4. open a new tab B
5. go back to tab A
6. switch to organization B

Cookie `__session` will include id of organization A instead of organization B.

### Improvements 

Extra improvements in performance achieved with this PR:

1. org switch will not request `/tokens`
3. user switch will not request `/tokens`
5. poller will not request `/tokens` before 60 seconds (minus leeway) of inactivity (session cookie is being updated in every `/touch` request)

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -452,11 +452,9 @@ export default class Clerk implements ClerkInterface {
       return;
     }
 
-    this.session = session;
-    this.organization = organization === null ? null : this.#getLastActiveOrganizationFromSession();
-    this.#aliasUser();
-
+    this.#setSessionAndOrganization(session);
     this.#emit();
+
     this.#resetComponentsState();
   };
 
@@ -864,20 +862,17 @@ export default class Clerk implements ClerkInterface {
     if (!this.client) {
       // This is the first time client is being
       // set, so we also need to set session
-      this.session =
-        (this.#options.selectInitialSession
-          ? this.#options.selectInitialSession(newClient)
-          : this.#defaultSession(newClient)) || null;
-      this.organization = this.#getLastActiveOrganizationFromSession();
-      this.#aliasUser();
+      const session = this.#options.selectInitialSession
+        ? this.#options.selectInitialSession(newClient)
+        : this.#defaultSession(newClient);
+      this.#setSessionAndOrganization(session);
     }
     this.client = newClient;
 
     if (this.session) {
       const lastId = this.session.id;
-      this.session = newClient.activeSessions.find(x => x.id === lastId);
-      this.organization = this.#getLastActiveOrganizationFromSession();
-      this.#aliasUser();
+      const session = newClient.activeSessions.find(x => x.id === lastId);
+      this.#setSessionAndOrganization(session);
     }
 
     this.#emit();
@@ -1072,6 +1067,12 @@ export default class Clerk implements ClerkInterface {
     return (
       orgMemberships.map(om => om.organization).find(org => org.id === this.session?.lastActiveOrganizationId) || null
     );
+  };
+
+  #setSessionAndOrganization = (session?: ActiveSessionResource | null, organization?: OrganizationResource | null) => {
+    this.session = session || null;
+    this.organization = organization === null ? null : this.#getLastActiveOrganizationFromSession();
+    this.#aliasUser();
   };
 
   #aliasUser = () => {
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -392,48 +392,46 @@ export default class Clerk implements ClerkInterface {
       throw new Error('setActive is being called before the client is loaded. Wait for init.');
     }
 
-    if (session === undefined) {
-      if (!this.session) {
-        throw new Error(
-          'setActive should either be called with a session param or there should be already an active session.',
-        );
-      }
-      session = this.session;
+    if (session === undefined && !this.session) {
+      throw new Error(
+        'setActive should either be called with a session param or there should be already an active session.',
+      );
     }
 
     if (typeof session === 'string') {
       session = (this.client.sessions.find(x => x.id === session) as ActiveSessionResource) || null;
     }
 
+    let newSession = (session === undefined ? this.session : session) as ActiveSessionResource | null;
+
     // At this point, the `session` variable should contain either an `ActiveSessionResource`
     // or `null`.
     // We now want to set the last active organization id on that session (if it exists).
     // However, if the `organization` parameter is not given (i.e. `undefined`), we want
     // to keep the organization id that the session had.
-    if (session && organization !== undefined) {
-      if (organization === null) {
-        session.lastActiveOrganizationId = null;
-      } else if (typeof organization === 'string') {
-        session.lastActiveOrganizationId = organization;
-      } else {
-        session.lastActiveOrganizationId = organization.id;
-      }
+    const shouldSwitchOrganization = organization !== undefined;
+    if (newSession && shouldSwitchOrganization) {
+      const organizationId = typeof organization === 'string' ? organization : organization?.id;
+      newSession.lastActiveOrganizationId = organizationId || null;
     }
-
-    this.#authService?.setAuthCookiesFromSession(session);
 
     // If this.session exists, then signout was triggered by the current tab
     // and should emit. Other tabs should not emit the same event again
-    if (this.session && session === null) {
+    const shouldSignoutSession = this.session && newSession === null;
+    if (shouldSignoutSession) {
       this.#broadcastSignOutEvent();
     }
 
-    //1. setLastActiveSession to passed usersession (add a param).
+    //1. setLastActiveSession to passed user session (add a param).
     //   Note that this will also update the session's active organization
     //   id.
     if (inActiveBrowserTab()) {
-      await this.#touchLastActiveSession(session);
+      await this.#touchLastActiveSession(newSession);
+      // reload session from updated client
+      newSession = this.#getSessionFromClient(newSession?.id);
     }
+
+    this.#authService?.setAuthCookiesFromSession(newSession);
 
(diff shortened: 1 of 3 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -405,45 +403,39 @@ export default class Clerk implements ClerkInterface {
...
     //   Note that this will also update the session's active organization
     //   id.
     if (inActiveBrowserTab()) {
-      await this.#touchLastActiveSession(session);
+      await this.#touchLastActiveSession(newSession);
+      // reload session from updated client
+      newSession = this.#getSessionFromClient(newSession.id);
     }
 
+    this.#authService?.setAuthCookiesFromSession(newSession);
```
nikosdouvlis, line 428: This will fire even when we're not in the currently active tab. I'm curious whether this can result in an out-of-sync cookie caused by having multiple tabs open.

In the case of calling `setAuthCookiesFromSession` within `setActive`:
- a user is a member of 2 orgs, org1 and org2
- the user opens tab1 and switches to org1. Tab1 shows: org1, cookie shows: org1
- the user opens tab2 and switches to org2. Tab2 shows: org2, cookie shows: org2, tab1 shows: org1

However, after the 2nd switch, tab1 still shows org1 as the selected org, even if the cookie shows org2 because it was overwritten after the switch.


In the case of calling `setAuthCookiesFromSession` from the poller: 
I need to explore this further, but can a poller fired from a different tab cause the cookie to fall out of sync?

This comment intends to trigger a discussion in case we require more changes in the setActive/cookie syncing algorithm - not a code change for now.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1071,6 +1046,34 @@ export default class Clerk implements ClerkInterface {
...
+  };
+
+  #getLastActiveOrganizationFromSession = () => {
+    const orgMemberships = this.session?.user.organizationMemberships || [];
+    return (
+      orgMemberships.map(om => om.organization).find(org => org.id === this.session?.lastActiveOrganizationId) || null
+    );
+  };
+
+  #setSessionAndOrganization = (session?: ActiveSessionResource | null, organization?: OrganizationResource | null) => {
```
nikosdouvlis, line 1063: ❓ We never call this with an org as the 2nd argument, do we actually need the 2nd argument? Also, this seems to always be setting the user as well. Should that be reflected in the name somehow (either explicitly or by making the name more generic)?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1071,6 +1046,34 @@ export default class Clerk implements ClerkInterface {
     }
   };
 
+  #setIntermediateState = () => {
```
nikosdouvlis, line 1049: Nice to see this extracted in its own method :)

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -96,14 +92,22 @@ export class Session extends BaseResource implements SessionResource {
...
     }
   };
 
+  // If it's a session token, retrieve it with their session id, otherwise it's a jwt template token
+  // and retrieve it using the session id concatenated with the jwt template name.
+  // e.g. session id is 'sess_abc12345' and jwt template name is 'haris'
+  // The session token ID will be 'sess_abc12345' and the jwt template token ID will be 'sess_abc12345-haris'
+  #getCacheId(template?: string) {
+    return `${template ? `${this.id}-${template}` : this.id}-${this.updatedAt.getTime()}`;
+  }
```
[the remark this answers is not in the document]
nikosdouvlis: ✔️

packages/clerk-js/src/core/clerk.ts
```diff
@@ -387,13 +386,12 @@ export default class Clerk implements ClerkInterface {
...
-    if (session === undefined) {
-      if (!this.session) {
-        throw new Error(
-          'setActive should either be called with a session param or there should be already an active session.',
-        );
-      }
-      session = this.session;
+    let newSession = (session === undefined ? this.session : session) as ActiveSessionResource | null;
+
+    if (!newSession) {
```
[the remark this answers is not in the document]
dimkl: you are correct. i will fix it and add a test.

packages/clerk-js/src/core/clerk.ts
dimkl: you are correct! i will remove organization. Any suggestion about the name?

packages/clerk-js/src/core/clerk.ts
dimkl: syncing tabs session with global cookie value will always cause the cookie to have an incorrect value in the inactive tabs (if there are on different orgs or users). This is a problem of mutating a global resource based on a local resource.

packages/clerk-js/src/core/services/authentication/AuthenticationService.ts
```diff
@@ -28,6 +28,8 @@ export class AuthenticationService {
   };
 
   public setAuthCookiesFromSession(session: SessionResource | undefined | null): void {
+    // We shouldn't replace lastActiveToken with getToken to avoid retrieving tokens from FAPI
+    // when tokens are expiring since token will probably be updated from session.touch on setActive
     this.updateSessionCookie(session?.lastActiveToken);
```
dimkl, line 33: cc: @nikosdouvlis

packages/clerk-js/src/core/clerk.ts
```diff
@@ -393,63 +392,56 @@ export default class Clerk implements ClerkInterface {
...
+      );
     }
 
     if (typeof session === 'string') {
       session = (this.client.sessions.find(x => x.id === session) as ActiveSessionResource) || null;
     }
 
+    let newSession = (session === undefined ? this.session : session) as ActiveSessionResource | null;
+
     // At this point, the `session` variable should contain either an `ActiveSessionResource`
```
SokratisVidros, line 407: ❓ Does the comment still apply?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1077,6 +1052,34 @@ export default class Clerk implements ClerkInterface {
     }
   };
 
+  #setIntermediateState = () => {
```
SokratisVidros, line 1055: 💯 on the name

SokratisVidros (approved):

packages/clerk-js/src/core/clerk.ts
dimkl: yes. we can verify this with line `405`

packages/clerk-js/src/core/clerk.ts
dimkl: will discuss it offline in slack #team-js channel

packages/clerk-js/src/core/clerk.ts
dimkl: we discussed it offline and settled for `setAliases`

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -96,14 +92,22 @@ export class Session extends BaseResource implements SessionResource {
...
         tokenResolver: Promise.resolve(token),
       });
     }
   };
 
+  // If it's a session token, retrieve it with their session id, otherwise it's a jwt template token
+  // and retrieve it using the session id concatenated with the jwt template name.
+  // e.g. session id is 'sess_abc12345' and jwt template name is 'haris'
+  // The session token ID will be 'sess_abc12345' and the jwt template token ID will be 'sess_abc12345-haris'
+  #getCacheId(template?: string) {
```
SokratisVidros, line 107: 💯

packages/clerk-js/src/core/clerk.ts
SokratisVidros, line 1055: I was thinking about this a bit more and I'd like to coin the term `Transitive` state. 

`setTransitiveState`. In essence, this state helps us transition from a signed-out to a signed-in state and vice-versa.

WDYT?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1077,6 +1052,34 @@ export default class Clerk implements ClerkInterface {
...
+  };
+
+  #getLastActiveOrganizationFromSession = () => {
+    const orgMemberships = this.session?.user.organizationMemberships || [];
+    return (
+      orgMemberships.map(om => om.organization).find(org => org.id === this.session?.lastActiveOrganizationId) || null
+    );
+  };
+
+  #setAliases = (session?: ActiveSessionResource | null) => {
```
SokratisVidros, line 1069: Maybe `setClerkAccessors` or `setAccessors`?

packages/clerk-js/src/core/clerk.ts
nikosdouvlis: Transitive makes sense to me 👍🏻

packages/clerk-js/src/core/clerk.ts
nikosdouvlis: Ditto

nikosdouvlis (approved):
Nicely done @dimkl !

SokratisVidros:
Nice work team 👏

PR #720, opened 2023-01-25
feat(clerk-js): Specify image size and support retina displays

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/packages/clerk-js/src/ui/primitives/Image.tsx
+++ b/packages/clerk-js/src/ui/primitives/Image.tsx
@@ -1,14 +1,23 @@
 import React from 'react';
 
+import { generateSrc, generateSrcSet } from '../../utils';
 import type { PrimitiveProps, StateProps } from '../styledSystem';
 import { applyDataStateProps } from './applyDataStateProps';
 
-export type ImageProps = PrimitiveProps<'img'> & StateProps;
+export type ImageProps = PrimitiveProps<'img'> &
+  StateProps & {
+    size?: number;
+    xDescriptors?: number[];
+  };
 
 export const Image = React.forwardRef<HTMLImageElement, ImageProps>((props, ref) => {
+  const { src, size = 80, xDescriptors = [1, 2], ...rest } = props;
+
   return (
     <img
-      {...applyDataStateProps(props)}
+      srcSet={generateSrcSet({ src, width: size, xDescriptors })}
+      src={generateSrc({ src, width: size * 2 })}
+      {...applyDataStateProps(rest)}
       ref={ref}
     />
   );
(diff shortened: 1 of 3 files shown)


desiprisg:
This seems to be sufficient to solve our current issue, however please add here anything else we might want to support in the near future, so we can tackle it!

packages/clerk-js/src/ui/primitives/Image.tsx
```diff
@@ -1,14 +1,22 @@
...
+  };
 
 export const Image = React.forwardRef<HTMLImageElement, ImageProps>((props, ref) => {
+  const { src, size = 80, ...rest } = props;
+
   return (
     <img
-      {...applyDataStateProps(props)}
+      srcSet={generateSrcSet({ size, src, sizes: [1, 2] })}
+      src={generateSrcSet({ size, src, sizes: [2] })}
```
dimkl, line 18: ❓ could we make `sizes` more readable. I imagine that it is being used to also include retina displays but it's not very clear to mean as a parameter.
Would it make more sense to add a boolean as `includeRetina  = true` instead of the array and hide the implementation detail of adding multiple `width` query params inside `generateSrcSet` ?

dimkl (approved):

packages/clerk-js/src/ui/primitives/Image.tsx
desiprisg: Yes, you're right. I'll refactor it.

packages/clerk-js/src/ui/primitives/Image.tsx
```diff
@@ -1,14 +1,23 @@
...
+  };
 
 export const Image = React.forwardRef<HTMLImageElement, ImageProps>((props, ref) => {
+  const { src, size = 80, xDescriptors = [1, 2], ...rest } = props;
+
   return (
     <img
-      {...applyDataStateProps(props)}
+      srcSet={generateSrcSet({ src, width: size, xDescriptors })}
+      src={generateSrc({ src, width: size * 2 })}
```
nikosdouvlis, line 19: ❓ Do we always want to fetch an image twice the size of the img element by default?

nikosdouvlis (approved):
Nice @desiprisg !!

packages/clerk-js/src/ui/primitives/Image.tsx
desiprisg: If `srcset` is not supported, then we would want to support the retina displays. Is that not reason enough?

nikosdouvlis:
@desiprisg conflicts and merge!

PR #721, opened 2023-01-25
Refactoring set auth cookies from session

dimkl:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

- [x]  `npm test` runs as expected.

Use `session.getToken()` in `setAuthCookiesFromSession` (requires making the method async and all the other methods that use it).


packages/clerk-js/src/core/services/authentication/AuthenticationService.ts
```diff
@@ -18,17 +18,17 @@ export class AuthenticationService {
...
     this.clearLegacyAuthV1Cookies();
     this.startPollingForToken();
     this.refreshTokenOnVisibilityChange();
-  };
+  }
 
-  public setAuthCookiesFromSession(session: SessionResource | undefined | null): void {
-    this.updateSessionCookie(session?.lastActiveToken);
+  public async setAuthCookiesFromSession(session: SessionResource | undefined | null): Promise<void> {
+    this.updateSessionCookie(await session?.getToken());
```
nikosdouvlis, line 31: I'd like to share a bit of context. After discussing with Dimitris, we both decided that replacing `.lastActiveToken` with `getToken` makes sense because: 

1. its a more uniform way of accessing the token. The `lastActiveToken` should be considered an implementation detail of the `SessionResource` 
2. if for *any* reason, we have a valid active session that has `lastActiveToken` set to null (eg: FAPI did not include `lastActiveToken` in the response) its safer to do an extra request rather than deleted the `__session` cookie. We have not yet identified any such scenario, but we are aware of a discrepancy between the `response` and `client` props of a FAPI response: sessions in `client` include the `lastActiveToken` while sessions in `response` do not. This is just a safety measure just in case.

nikosdouvlis (approved):
Nicely done @dimkl

packages/clerk-js/src/core/services/authentication/AuthenticationService.ts
SokratisVidros: Team, what's the latest update on this?

dimkl:
By fixing the `last_active_token` in client fixtures, most of the tests were fixed. By also adding a `getToken()` mock in mocked sessions, the rest of the tests were also fixed.
I haven't introduced a test for this case . If we are okay with those change let's proceed with merging this.
@nikosdouvlis , @SokratisVidros WDYT?

PR #723, opened 2023-01-26
Introduce gatsby-plugin-clerk v4

anagstef:
## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [x] `gatsby-plugin-clerk`
- [x] `build/tooling/chore`

## Description

# gatsby-plugin-clerk v4.0.0

This PR introduces the major version 4 of the Clerk plugin for Gatsby (`gatsby-plugin-clerk`). It also makes the package publishable again.

It supports:
- **New Publishable/Secret keys**
- **Gatsby v5**
- **Server API routes** (via `gatsby-plugin-clerk/api`)
- **SSR** (via `gatsby-plugin-clerk/ssr`)

## Environment Variables

Gatsby supports private (hidden from the client) and public (accessible on the client side) environment variables. The convention for public environment variables is the prefix `GATSBY_`. All other variables are hidden, and accessible only on the server. [Read more here](https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/).

Clerk’s supported environment variables are:

- `CLERK_API_URL` (**optional**, defaults to `'https://api.clerk.dev'` if not provided)
- `CLERK_API_VERSION` (**optional**, defaults to `'v1'` if not provided)
- `CLERK_API_KEY` (deprecated legacy key)
- `GATSBY_CLERK_FRONTEND_API` (deprecated legacy key)
- `CLERK_SECRET_KEY` 
- `GATSBY_CLERK_PUBLISHABLE_KEY`

## Setting up a Gatsby project with Clerk

1. Create a new Gatsby 5 project
2. Install Clerk’s plugin:
```
npm i gatsby-plugin-clerk
```
3. Create a `.env.development` file in the root of the project, for the development instance.

> Note: `In development, Gatsby will load environment variables from a file named .env.development. For builds, it will load from .env.production.`

Add the env vars from the dashboard in the following format:

```
GATSBY_CLERK_PUBLISHABLE_KEY=your_publishable_key
CLERK_SECRET_KEY=your_secret_key
```

For apps with legacy keys:

```
GATSBY_CLERK_FRONTEND_API=your_frontend_api
CLERK_API_KEY=your_api_key
```

4. Inside your `gatsby-config.js` file:

a. Add the following snippet to be able to load environment variables:

```ts
require("dotenv").config({
  path: .env.${process.env.NODE_ENV}
})
```

b. Add the plugin without any options:

```ts
module.exports = {
  plugins: ['gatsby-plugin-clerk']
}
```

### Client-side

The usage on the client side stays almost the same as before. Client-side components now are imported directly from  `gatsby-plugin-clerk`. Example file `/pages/index.tsx`:

```ts
import React from 'react'
import {
  SignIn,
  SignedIn,
  SignedOut,
  UserButton
  } from 'gatsby-plugin-clerk'

export default function IndexPage() {
  return (
    <>
      <SignedIn>
        <UserButton />
      </SignedIn>
      <SignedOut>
        <SignIn />
      </SignedOut>
    </>
  )
}
```

### Server-Side Rendering (SSR)

User can access `withServerAuth` from `'gatsby-plugin-clerk/ssr'`. Example file `/pages/ssr.tsx`:

```ts
import * as React from 'react';
import { GetServerData } from 'gatsby';
import { withServerAuth } from 'gatsby-plugin-clerk/ssr';

export const getServerData: GetServerData<any> = withServerAuth(
  async props => {
    return { props: { data: '1', auth: props.auth } };
  },
  { loadUser: true },
);

function SSRPage({ serverData }: any) {
  return (
    <main>
      <h1>SSR Page with Clerk</h1>
      <pre>{JSON.stringify(serverData, null, 2)}</pre>
    </main>
  );
}

export default SSRPage;
```

### Server API routes

Importing `'gatsby-plugin-clerk/api'` gives acces to all the exports coming from `@clerk/sdk-node`. Example file `/api/hello.ts`:

```ts
import { clerkClient, withAuth } from 'gatsby-plugin-clerk/api';

interface ContactBody {
  message: string;
}

const handler = withAuth(async (req, res) => {
  const users = await clerkClient.users.getUserList();
  res.send({ title: `We have ${users.length} users`, message: req.body.message, auth: req.auth });
});

export default handler;
```

## Breaking changes

Updating `gatsby-plugin-clerk` from `v2.x` to `≥4.x` needs the following steps:

1. Move Clerk’s key(s) from the plugin’s options to the `.env.development` file and add the following snippet to `gatsby-config.js`, as described above:
```ts
require("dotenv").config({
  path: .env.${process.env.NODE_ENV}
})
```
2. Remove the `@clerk/clerk-react` package, which is now re-exported directly from the `gatsby-plugin-clerk` package.


anagstef:
❓ @nikosdouvlis Do we need to bump the `peerDependency` inside the plugin's `package.json` ?
```
  "peerDependencies": {
    "gatsby": "^4.24.8"
  }

File: packages/gatsby-plugin-clerk/src/ssr/authenticateRequest.ts [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+    isDevelopmentOrStaging(process.env.CLERK_SECRET_KEY || process.env.CLERK_API_KEY || '') &&
+    hostUrl.hostname === referrerUrl.hostname
+  ) {
+    return referrerUrl.host;
+  }
+
+  return forwardedHost;
+};
+
+function isDevelopmentOrStaging(apiKey: string): boolean {
```
panteliselef, line 50: 💡 Does it makes sense to add this to @clerk/shared ?

nikosdouvlis: It definately does @panteliselef !

However, it looks like clerk/shared is not tree shakable at the moment so I'm a a bit worried about adding it as a dependency of a package that can run in a server environment.

We do have a task to solve this issue for this cycle though, so lets add a TODO we can move it there right after :)

nikosdouvlis:
> ❓ @nikosdouvlis Do we need to bump the `peerDependency` inside the plugin's `package.json` ?
> 
> ```
>   "peerDependencies": {
>     "gatsby": "^4.24.8"
>   }
> ```

I think that the current implementation works nicely with any version >=2 but lets test first.

package.json
```diff
@@ -28,20 +28,19 @@
     "lerna": "^4.0.0",
     "prettier": "^2.6.0",
     "pretty-quick": "^3.1.3",
+    "react": "18.2.0",
```
nikosdouvlis, line 31: We need to make these peer deps

package.json
```diff
@@ -28,20 +28,19 @@
...
+    "react-dom": "18.2.0",
     "rimraf": "^3.0.2",
     "ts-jest": "^29.0.3",
     "tsup": "^6.5.0",
     "turbo": "^1.6.3",
-    "typescript": "4.9.4",
-    "next": "^13.0.2",
-    "react": "18.2.0",
-    "react-dom": "18.2.0"
+    "typescript": "4.9.4"
```
nikosdouvlis, line 37: We can use * here

package.json
```diff
@@ -28,20 +28,19 @@
...
   },
   "scripts": {
     "dev": "turbo dev --filter=@clerk/* --filter=!@clerk/expo",
     "dev:playground": "echo 'Not ready for use yet' && exit 1 && turbo dev --filter=@playground/*",
-    "build": "turbo build --filter=!gatsby-plugin-clerk --concurrency=${TURBO_CONCURRENCY:-2}",
-    "test": "turbo test --filter=!gatsby-plugin-clerk --concurrency=${TURBO_CONCURRENCY:-2}",
+    "build": "turbo build --concurrency=${TURBO_CONCURRENCY:-2}",
```
nikosdouvlis, line 42: ✔️

packages/types/src/ssr.ts
```diff
@@ -6,6 +6,9 @@ import { UserResource } from './user';
 export type ServerGetTokenOptions = { template?: string };
 export type ServerGetToken = (options?: ServerGetTokenOptions) => Promise<string | null>;
 
+/**
+ * @deprecated
```
nikosdouvlis, line 10: Lets remove this for now as we haven't discussed its replacement yet :)

nikosdouvlis (approved):
Nicely done @anagstef ! Thanks for the detailed description

package.json
nikosdouvlis: My bad, this is the root package.json. Ignore!

package.json
nikosdouvlis: My bad, this is the root package.json. Ignore this comment!

SokratisVidros (approved):
💯

PR #725, opened 2023-01-26
chore(nextjs): Invalidate next sdc/sbc router cache before setActive

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

See comment within the changes files

### DIFF
--- a/packages/clerk-js/src/core/clerk.test.ts
+++ b/packages/clerk-js/src/core/clerk.test.ts
@@ -1,8 +1,9 @@
-import { ActiveSessionResource, SignInJSON, SignUpJSON } from '@clerk/types';
+import type { ActiveSessionResource, SignInJSON, SignUpJSON } from '@clerk/types';
 import { waitFor } from '@testing-library/dom';
 
 import Clerk from './clerk';
-import { Client, DisplayConfig, Environment, MagicLinkErrorCode, SignIn, SignUp } from './resources/internal';
+import type { DisplayConfig } from './resources/internal';
+import { Client, Environment, MagicLinkErrorCode, SignIn, SignUp } from './resources/internal';
 
 const mockClientFetch = jest.fn();
 const mockEnvironmentFetch = jest.fn();
@@ -151,6 +152,35 @@ describe('Clerk singleton', () => {
       await waitFor(() => {
         expect(mockSession.touch).not.toHaveBeenCalled();
       });
+    });
+
+    it('calls __unstable__onBeforeSetActive before session.touch', async () => {
+      mockSession.touch.mockReturnValueOnce(Promise.resolve());
+      mockClientFetch.mockReturnValue(Promise.resolve({ activeSessions: [mockSession] }));
+
+      (window as any).__unstable__onBeforeSetActive = () => {
+        expect(mockSession.touch).not.toHaveBeenCalled();
+      };
+
+      const sut = new Clerk(frontendApi);
+      await sut.load();
+      await sut.setActive({ session: mockSession as any as ActiveSessionResource });
+      expect(mockSession.touch).toHaveBeenCalled();
+    });
+
+    it('calls __unstable__onAfterSetActive after beforeEmit and session.touch', async () => {
+      const beforeEmitMock = jest.fn();
+      mockSession.touch.mockReturnValueOnce(Promise.resolve());
+      mockClientFetch.mockReturnValue(Promise.resolve({ activeSessions: [mockSession] }));
+
+      (window as any).__unstable__onAfterSetActive = () => {
+        expect(mockSession.touch).toHaveBeenCalled();
+        expect(beforeEmitMock).toHaveBeenCalled();
+      };
+
+      const sut = new Clerk(frontendApi);
+      await sut.load();
+      await sut.setActive({ session: mockSession as any as ActiveSessionResource, beforeEmit: beforeEmitMock });
     });
   });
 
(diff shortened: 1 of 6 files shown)


packages/nextjs/src/client/invalidateNextRouterNon200ResponseCache.ts
```diff
@@ -0,0 +1,53 @@
...
+//   dataHref: string;
+//   json: Record<string, any> | null;
+//   response: Response;
+//   text: string;
+//   cacheKey: string;
+// }
+//
+// interface NextDataCache {
+//   [asPath: string]: Promise<FetchDataOutput>;
+// }
```
anagstef, line 11: ❓ Are these forgotten?

anagstef (approved):

anagstef:
LGTM!

packages/nextjs/src/client/invalidateNextRouterNon200ResponseCache.ts
nikosdouvlis: Nop, see the comment below

anagstef: Oh, my bad! nice!

packages/nextjs/src/client/index.tsx
```diff
@@ -16,22 +18,27 @@ type NextClerkProviderProps = {
...
+
   return (
     <ReactClerkProvider
       frontendApi={frontendApi || process.env.NEXT_PUBLIC_CLERK_FRONTEND_API || ''}
       publishableKey={publishableKey || process.env.NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY || ''}
       clerkJSUrl={clerkJSUrl || process.env.NEXT_PUBLIC_CLERK_JS}
       // @ts-expect-error
       proxyUrl={proxyUrl || process.env.NEXT_PUBLIC_CLERK_PROXY_URL}
-      navigate={to => push(to)}
+      navigate={navigate}
```
nikosdouvlis, line 41: I feel like we do need before/after setActive hooks, otherwise we risk ending up on the same flow if a setActive is called with a beforeEmit and the middleware somehow triggers afterwards

nikosdouvlis: If we're unsure about the api (I think that's the case) and since I don' want to add props accepting non-scaler values to ClerkProvider (/app requires the values to be serializable), Im thinking we can try the following API:

https://github.com/clerkinc/javascript/blob/2c0112908e2be131bd951ad875e4e48621779162/packages/clerk-js/src/core/fapiClient.ts#L74

cc @colinclerk

nikosdouvlis: Update: We're going to completely nuke the cache and move the logic in the beforeSetActive handler

colinclerk (approved):
This looks good to me for now. When we productize these hooks, I believe we will:
* Allow multiple callbacks
* Await them all

Good call on 1-by-1 deletion

nikosdouvlis:
> This looks good to me for now. When we productize these hooks, I believe we will:
> 
> * Allow multiple callbacks
> * Await them all
> 
> Good call on 1-by-1 deletion

We'll need to tweak the API a bit but lets see what possibilities these open up :)

anagstef (approved):
Looks good to me!

packages/clerk-js/src/core/clerk.test.ts
```diff
@@ -152,6 +153,35 @@ describe('Clerk singleton', () => {
         expect(mockSession.touch).not.toHaveBeenCalled();
       });
     });
+
+    it('calls __unstable__onBeforeSetActive before session.touch', async () => {
+      mockSession.touch.mockReturnValueOnce(Promise.resolve());
```
dimkl, line 158: 🙃 i think we should add a `expect.assertions(2)` to ensure that expectation in `__unstable__onBeforeSetActive` is called.

packages/clerk-js/src/core/clerk.test.ts
```diff
@@ -152,6 +153,35 @@ describe('Clerk singleton', () => {
...
+      };
+
+      const sut = new Clerk(frontendApi);
+      await sut.load();
+      await sut.setActive({ session: mockSession as any as ActiveSessionResource });
+      expect(mockSession.touch).toHaveBeenCalled();
+    });
+
+    it('calls __unstable__onAfterSetActive after beforeEmit and session.touch', async () => {
+      const beforeEmitMock = jest.fn();
```
dimkl, line 172: 🙃 i think we should add a `expect.assertions(2)` to ensure that expectation in `__unstable__onAfterSetActive` is called.

dimkl (commented):
❓ why did we have to set `__unstable__onBeforeSetActive`  and `__unstable__onAfterSetActive` on window and not on Clerk instance ?

PR #727, opened 2023-01-26
Support multi-domain authentication

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

As part of this PR we are introducing 2 new props for <ClerkProvider /> 
- isSatellite?: boolean → This prop will dictate whether or not we should perform Syncing with the primary app
- domain?: string → It should be the FAPI key of the satellite app e.g. clerk.satellite.dev 

Syncing with the primary app: 
- If isSatellite === true we should perform the syncing process to sync sessions with the primary app
  - Redirect to https://clerk.satellite.dev/v1/client/sync?redirect_url=https://clerk.satellite.dev?synced=true 
  - As part of the redirect_url we append ?synced=true as a way to determine that syncing was attempted
  
 In a following PR we will tackle the necessary changes for interstitial

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -196,12 +196,11 @@ export default class Clerk implements ClerkInterface {
     };
 
     if (this.#options.standardBrowser) {
-      await this.#loadInStandardBrowser();
+      this.#isReady = await this.#loadInStandardBrowser();
     } else {
       await this.#loadInNonStandardBrowser();
+      this.#isReady = true;
     }
-
-    this.#isReady = true;
   };
 
   public signOut: SignOut = async (callbackOrOptions?: SignOutCallback | SignOutOptions, options?: SignOutOptions) => {
@@ -954,33 +953,38 @@ export default class Clerk implements ClerkInterface {
     this.#componentControls?.updateProps(props);
   };
 
-  #syncWithPrimary = () => {
+  #redirectToSyncWithPrimary = () => {
     const q = new URLSearchParams({
       redirect_url: `${window.location.href}?synced=true`,
     });
     window.location.replace(new URL(`/v1/client/sync?${q.toString()}`, `https://${this.domain}`).toString());
   };
 
-  #loadInStandardBrowser = async (): Promise<void> => {
-    console.log('----- isSatellite -----', this.#options.isSatellite);
+  #stripSyncedQueryParam = (q: URLSearchParams) => {
+    q.delete('synced');
+    const queryString = q.toString() ? `?${q.toString()}` : '';
+    window.history.replaceState({}, '', `${window.location.pathname}${queryString}`);
+  };
+
+  #shouldSyncWithMain = () => {
+    const q = new URLSearchParams(window.location.search);
+    if (q.get('synced') !== 'true') {
+      this.#redirectToSyncWithPrimary();
+      return true;
+    }
+    this.#stripSyncedQueryParam(q);
+    return false;
+  };
+
+  #loadInStandardBrowser = async (): Promise<boolean> => {
     if (this.#options.isSatellite) {
-      const q = new URLSearchParams(window.location.search);
-      if (q.get('synced') !== 'true') {
-        this.#syncWithPrimary();
-        return;
-      } else if (q.get('synced') === 'true') {
-        q.delete('synced');
-        const queryString = q.toString() ? `?${q.toString()}` : '';
-        window.history.replaceState({}, '', `${window.location.pathname}${queryString}`);
+      if (this.#shouldSyncWithMain()) {
+        return false;
       }
     }
 
     this.#authService = new AuthenticationService(this);
     this.#pageLifecycle = createPageLifecycle();
-
-    console.log('----- loadInStandardBrowser -----', this.domain);
-    console.log('----- loadInStandardBrowser -----', this.frontendApi);
-    console.log('----- loadInStandardBrowser -----', window.location.host);
 
     this.#devBrowserHandler = createDevBrowserHandler({
       frontendApi: this.frontendApi,
@@ -1041,6 +1045,8 @@ export default class Clerk implements ClerkInterface {
         clerkErrorInitFailed();
       }
     }
+
+    return true;
   };
 
   #loadInNonStandardBrowser = async (): Promise<void> => {
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -93,14 +93,16 @@ import { AuthenticationService } from './services';
 
 export type ClerkCoreBroadcastChannelEvent = { type: 'signout' };
 
-type ClerkConstructorOptions = Pick<ClerkInterface, 'proxyUrl'>;
+type ClerkConstructorOptions = Pick<ClerkInterface, 'proxyUrl' | 'domain' | 'isSatellite'>;
 
 declare global {
   interface Window {
     Clerk?: Clerk;
     __clerk_frontend_api?: string;
     __clerk_publishable_key?: string;
-    __clerk_proxy_url?: Pick<ClerkInterface, 'proxyUrl'>['proxyUrl'];
+    __clerk_proxy_url?: ClerkConstructorOptions['proxyUrl'];
+    __clerk_domain?: ClerkConstructorOptions['domain'];
+    __clerk_is_satellite?: ClerkConstructorOptions['isSatellite'];
   }
 }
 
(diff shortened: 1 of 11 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -908,10 +914,31 @@ export default class Clerk implements ClerkInterface {
...
+    });
+    window.location.replace(new URL(`/v1/client/sync?${q.toString()}`, `https://${this.domain}`).toString());
+  };
+
   #loadInStandardBrowser = async (): Promise<void> => {
+    console.log('----- isSatellite -----', this.isSatellite);
+    if (this.isSatellite) {
+      const q = new URLSearchParams(window.location.search);
+      if (!q.has('synced')) {
+        // Navigate to main FAPI and try to sync cookies
```
gkats, line 929: :upside_down_face: Let's remove this comment.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -908,10 +914,31 @@ export default class Clerk implements ClerkInterface {
...
+      redirect_url: `${window.location.href}?synced=true`,
+    });
+    window.location.replace(new URL(`/v1/client/sync?${q.toString()}`, `https://${this.domain}`).toString());
+  };
+
   #loadInStandardBrowser = async (): Promise<void> => {
+    console.log('----- isSatellite -----', this.isSatellite);
+    if (this.isSatellite) {
+      const q = new URLSearchParams(window.location.search);
+      if (!q.has('synced')) {
```
gkats, line 928: :upside_down_face: We can also check that `synced` has a value of true (or "true"). 

It would also be nice if we could rewrite the URL to get rid of the `synced` parameter in the query string. 

This means that we'll end up in exactly the same URL as we started. Not sure if it's doable, we can try in a follow-up PR.

packages/nextjs/src/client/index.tsx
```diff
@@ -16,21 +16,33 @@ type NextClerkProviderProps = {
...
+    __clerk_ssr_state,
+    clerkJSUrl,
+    ...restProps
+  } = rest;
   const { push } = useRouter();
 
   ReactClerkProvider.displayName = 'ReactClerkProvider';
 
   return (
+    // @ts-expect-error
```
gkats, line 38: :upside_down_face: Better add this above each line that might raise the error? This way we'll know what needs to be changed in the future. Feel free to ignore.

packages/react/src/utils/scriptLoader.ts
```diff
@@ -95,6 +100,14 @@ export async function loadScript(params: LoadScriptParams): Promise<HTMLScriptEl
     if (proxyUrl) {
       script.setAttribute('data-clerk-proxy-url', proxyUrl);
     }
+    if (domain) {
+      script.setAttribute('data-clerk-domain', domain);
+    }
+
+    if (isSatellite) {
+      script.setAttribute('data-clerk-is-satellite', isSatellite ? 'true' : 'false');
```
gkats, line 108: :upside_down_face: 
```suggestion
      script.setAttribute('data-clerk-is-satellite', String(!!isSatellite));
```

gkats (commented):
:clap: Nicely done!

packages/react/src/utils/scriptLoader.ts
panteliselef: Now that I am thinking about it, maybe isSatellite should no be there. Instead I it should be included in the options of the `load` function. This is because isSatellite is required only after a `Clerk` instance already exists.

packages/clerk-js/src/core/clerk.ts
panteliselef: 👍

packages/clerk-js/src/core/clerk.ts
panteliselef: ✅

packages/types/src/key.ts
```diff
@@ -14,3 +14,14 @@ export type PublishableKeyOrFrontendApi =
       frontendApi: string;
       publishableKey?: never;
     };
+
+// export type SatelliteApp =
```
raptisj, line 18: @panteliselef this comment?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -948,7 +952,36 @@ export default class Clerk implements ClerkInterface {
     this.#componentControls?.updateProps(props);
   };
 
-  #loadInStandardBrowser = async (): Promise<void> => {
+  #redirectToSyncWithPrimary = () => {
```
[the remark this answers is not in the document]
gkats: We should use "primary" everywhere.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -948,7 +952,36 @@ export default class Clerk implements ClerkInterface {
...
+    window.location.replace(new URL(`/v1/client/sync?${q.toString()}`, `https://${this.domain}`).toString());
+  };
+
+  #stripSyncedQueryParam = (q: URLSearchParams) => {
+    q.delete('synced');
+    const queryString = q.toString() ? `?${q.toString()}` : '';
+    window.history.replaceState({}, '', `${window.location.pathname}${queryString}`);
+  };
+
+  #shouldSyncWithMain = () => {
```
gkats, line 968: :upside_down_face: The name of this function might be misleading. It actually performs the redirect.

I imagine "shouldSyncWithPrimary" means that I am a satellite and there's no `synced` query string parameter.

Then I can "syncWithPrimary".

packages/clerk-js/src/core/clerk.ts
```diff
@@ -948,7 +952,36 @@ export default class Clerk implements ClerkInterface {
...
+    this.#stripSyncedQueryParam(q);
+    return false;
+  };
+
+  #loadInStandardBrowser = async (): Promise<boolean> => {
+    if (this.#options.isSatellite) {
+      if (this.#shouldSyncWithMain()) {
+        return false;
+      }
+    }
```
gkats, line 983: :upside_down_face: This can be simplified further.

```js
if (this.shouldSyncWithPrimary()) {
  this.syncWithPrimary();
}
// do whatever else
  return
}

// shouldSyncWithPrimary()
return isSatellite && !syncTrueParamInTheQuery
```

packages/types/src/clerk.ts
```diff
@@ -460,6 +466,11 @@ export interface ClerkOptions {
   /** Optional support email for display in authentication screens */
   supportEmail?: string;
   touchSession?: boolean;
+  /**
+   * @experimental
+   * Indicates if the app is a Satellite app.
+   */
+  isSatellite?: boolean;
```
gkats, line 473: :upside_down_face: Let's remove the property descriptions until we finalize the API. Keep the "@experimental" annotation though. :+1:

packages/react/src/utils/scriptLoader.ts
```diff
@@ -69,13 +69,16 @@ export interface LoadScriptParams {
   frontendApi?: string;
   publishableKey?: string;
   proxyUrl?: string;
+
+  domain?: string;
   scriptUrl?: string;
+
```
gkats, line 75:
```suggestion
```

:laughing:

gkats (approved):
Looks good! :clap: :clap: 

I think the logic with the satellite check and redirection could be further simplified. Might be worth a try.

packages/clerk-js/src/core/clerk.ts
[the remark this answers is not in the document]
panteliselef: I agree, I will make the change

packages/types/src/key.ts
panteliselef: needs to be removed, correct types for developers should be a concern for later

packages/clerk-js/src/core/clerk.ts
```diff
@@ -949,7 +953,36 @@ export default class Clerk implements ClerkInterface {
...
+  #syncWithPrimary = () => {
+    const q = new URLSearchParams({
+      redirect_url: `${window.location.href}?synced=true`,
+    });
+    window.location.replace(new URL(`/v1/client/sync?${q.toString()}`, `https://${this.domain}`).toString());
+  };
+
+  #handleSyncedQueryParam = () => {
+    function stripSyncedQueryParam(q: URLSearchParams) {
+      q.delete('synced');
```
gkats, line 965: :upside_down_face: Wonder if we should name this parameter `__clerk_synced` to avoid any naming collisions.

gkats (approved):
:satellite:

packages/clerk-js/src/core/clerk.ts
```diff
@@ -949,7 +954,29 @@ export default class Clerk implements ClerkInterface {
     this.#componentControls?.updateProps(props);
   };
 
-  #loadInStandardBrowser = async (): Promise<void> => {
+  #syncWithPrimary = () => {
+    const q = new URLSearchParams({
+      redirect_url: `${window.location.href}?synced=true`,
```
gkats, line 959: :wrench: @panteliselef I think we missed this one!

panteliselef: Yes, already fixed :)

PR #729, opened 2023-01-27
Allow absolute and relative paths for ProxyUrl

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

This PR allows users to use the following values for the proxyUrl prop
- `https://my-clerk-app.dev/api/__clerk` (useful for Expo)
- `/api/__clerk` (useful for web)

The previous pattern of `my-clerk-app.dev/api/__clerk` is no longer valid

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -1,5 +1,12 @@
 import type { LocalStorageBroadcastChannel } from '@clerk/shared';
-import { inClientSide, isLegacyFrontendApiKey, noop, parsePublishableKey } from '@clerk/shared';
+import {
+  inClientSide,
+  isLegacyFrontendApiKey,
+  isValidProxyUrl,
+  noop,
+  parsePublishableKey,
+  proxyUrlToAbsoluteURL,
+} from '@clerk/shared';
 import type {
   ActiveSessionResource,
   AuthenticateWithMetamaskParams,
@@ -138,7 +145,12 @@ export default class Clerk implements ClerkInterface {
   public constructor(key: string, options?: unknown) {
     key = (key || '').trim();
 
-    this.proxyUrl = (options as ClerkConstructorOptions | undefined)?.proxyUrl;
+    const _unfilteredProxy = (options as ClerkConstructorOptions | undefined)?.proxyUrl;
+
+    if (!isValidProxyUrl(_unfilteredProxy)) {
+      errorThrower.throwInvalidProxyUrl({ url: _unfilteredProxy });
+    }
+    this.proxyUrl = proxyUrlToAbsoluteURL(this.proxyUrl);
 
     if (isLegacyFrontendApiKey(key)) {
       if (!validateFrontendApi(key)) {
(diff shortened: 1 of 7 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -137,6 +144,14 @@ export default class Clerk implements ClerkInterface {
 
     this.proxyUrl = (options as ClerkConstructorOptions | undefined)?.proxyUrl;
 
+    if (this.proxyUrl) {
+      if (isValidProxyUrl(this.proxyUrl)) {
```
gkats, line 148: :upside_down_face: Can't we have the `isValidProxyUrl` function handle undefined/null values and get rid of the double conditional here?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -137,6 +144,14 @@ export default class Clerk implements ClerkInterface {
...
+        this.proxyUrl = isProxyUrlRelative(this.proxyUrl) ? window.location.origin + this.proxyUrl : this.proxyUrl;
```
gkats, line 149: :wrench: Is it safe to join URL paths with string concatenation? Is there a chance that `window.location.origin` might have a trailing slash?

How about using `new URL(path, base).toString()` instead?

packages/shared/src/errors/thrower.ts
```diff
@@ -1,5 +1,6 @@
 const DefaultMessages = Object.freeze({
   InvalidFrontendApiErrorMessage: `The frontendApi passed to Clerk is invalid. You can get your Frontend API key at https://dashboard.clerk.dev/last-active?path=api-keys. (key={{key}})`,
+  InvalidProxyUrlErrorMessage: `The proxyUrl passed to Clerk is invalid. Proxy url should be an absolute URL or a relative path  with a leading '/'. (key={{url}})`,
```
gkats, line 3: :wrench: We should either use the uppercase word or lowercase for url and not mix "url" and "URL" in the same sentence.

packages/shared/src/utils/proxy.ts
```diff
@@ -0,0 +1,7 @@
+export function isValidProxyUrl(key: string) {
+  return key.startsWith('https://') || key.startsWith('/');
```
gkats, line 2: :upside_down_face: 
```suggestion
  return key.startsWith('https://') || isProxyUrlRelative(key);
```

gkats (approved):
Nice! Looks good overall, left a few comments for potential improvement.

packages/clerk-js/src/core/clerk.ts
panteliselef: I also dislike the nested conditionals here, but `proxyUrl` is an optional property, and I would like to throw an error if that property has set but it's format is wrong.
If I merge this logic, the the error will also throw if the proxyUrl is not set by the developer.

packages/shared/src/errors/thrower.ts
panteliselef:
```suggestion
  InvalidProxyUrlErrorMessage: `The proxyUrl passed to Clerk is invalid. The expected value for proxyUrl is an absolute URL or a relative path with a leading '/'. (key={{url}})`,
```

packages/shared/src/utils/proxy.ts
panteliselef: ✅

PR #730, opened 2023-01-27
AUTH-167 fix(clerk-js): Set show/hide password button type to button

yourtallness:
Pressing enter on the sign up form will trigger a click on the first button with type=submit within the form. While this isn't a problem on Chrome, on Safari the show/reveal password button catches the click shows/reveals the password instead of allowing the form to be submitted.
Setting type=button explicitly fixes the issue.

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

Note: button type was explicitly set to `undefined` to avoid reset behavior from UI frameworks.
Thus not sure if this can introduce a problem.

Also, this may need to be ported to `SimpleButton` as well.

### DIFF
--- a/packages/clerk-js/src/ui/primitives/Button.tsx
+++ b/packages/clerk-js/src/ui/primitives/Button.tsx
@@ -25,6 +25,7 @@ const { applyVariants, filterProps } = createVariants((theme, props: OwnProps) =
       ...common.disabled(theme),
       transitionProperty: theme.transitionProperty.$common,
       transitionDuration: theme.transitionDuration.$controls,
+      type: undefined,
     },
     variants: {
       textVariant: common.textVariants(theme),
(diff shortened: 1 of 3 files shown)


packages/clerk-js/src/ui/primitives/Button.tsx
```diff
@@ -164,10 +165,7 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>((props, ref) =>
   return (
     <button
       {...applyDataStateProps(rest)}
-      // Explicitly remove type=submit or type=button
-      // to prevent global css resets (eg tailwind) from affecting
-      // the default styles of our components
-      type={undefined}
+      type={rest.type}
```
nikosdouvlis, line 168: IIRC type needs to be explicitly set to `undefined` to avoid framework resets. `rest.type` should be undefined if not overridden by the caller, but we should double check

(I'd also leave the comment in place as we've already been bitten by this once)

nikosdouvlis: Also, I think its easy to test by mounting a SignIn/SignUp component in a next + tailwind project and checking whether the form Continue button is broken

anagstef (approved):

PR #731, opened 2023-01-27
fix(clerk-js): Do not modify src if it is data

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

When a new image is uploaded, we show the base64 encoded image. In that case, we don't want to append any search params.

### DIFF
--- a/packages/clerk-js/src/utils/url.ts
+++ b/packages/clerk-js/src/utils/url.ts
@@ -287,9 +287,30 @@ export function removeSearchParameterFromHash({
   return dummyUrlForHash.href.replace(DUMMY_URL_BASE, '');
 }
 
+export function isValidUrl(val?: string): val is string {
+  if (!val) {
+    return false;
+  }
+
+  try {
+    new URL(val);
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
+
+export function isDataUri(val?: string): val is string {
+  if (val?.startsWith('data:')) {
+    return true;
+  }
+
+  return false;
+}
+
 export const generateSrc = ({ src, width }: { src?: string; width: number }) => {
-  if (!src) {
-    return '';
+  if (!isValidUrl(src) || isDataUri(src)) {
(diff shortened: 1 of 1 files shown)


packages/clerk-js/src/utils/url.ts
```diff
@@ -292,6 +292,10 @@ export const generateSrc = ({ src, width }: { src?: string; width: number }) =>
     return '';
   }
 
+  if (src.substring(0, 5) === 'data:') {
```
dimkl, line 295: ❓ why not `if (src.startsWith('data:')) {` ?

dimkl (approved):
❓ Should we also add a test for this case?

packages/clerk-js/src/utils/url.ts
desiprisg: Right, :D

nikosdouvlis, line 295: Maybe lets make it more generic by introducing a util

```
function isValidUrl(val?: string): val is string {
	if (!val) {
    return false
  };

	try {
		new URL(val);
		return true;
	} catch (e) {
		return false;
	}
}

```

nikosdouvlis: Check if something like that already exists though because this snippet looks familiar

PR #741, opened 2023-01-30
fix(clerk-js): Fix sign out of all accounts when there is no active session

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

Bug:
It seems that when signing out of one of the sessions, and now being in the "Choose a user to continue with" screen, the signout button does nothing. This happens because there is no session at that point and the function returns.

Fix:
Instead of checking that a session exists, check the number of the sessions and return based on that.

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -202,7 +202,7 @@ export default class Clerk implements ClerkInterface {
     }
 
     const session = this.client.activeSessions.find(s => s.id === opts.sessionId);
-    const shouldSignOutCurrent = this.session?.id === session?.id;
+    const shouldSignOutCurrent = session?.id && this.session?.id === session.id;
     await session?.remove();
     if (shouldSignOutCurrent) {
       return this.setActive({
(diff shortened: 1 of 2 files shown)


dimkl (approved):

nikosdouvlis (approved):
Well done :)
@desiprisg can you add a short description explaining the bug and the fix and update the ticket accordingly?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -187,7 +187,7 @@ export default class Clerk implements ClerkInterface {
   };
 
   public signOut: SignOut = async (callbackOrOptions?: SignOutCallback | SignOutOptions, options?: SignOutOptions) => {
-    if (!this.client || !this.session) {
+    if (!this.client || this.client.sessions.length === 0) {
       return;
     }
     const cb = typeof callbackOrOptions === 'function' ? callbackOrOptions : undefined;
```
nikosdouvlis, line 193: If called without a specific sessionId, signOut needs to signout all sessions (in case of a multi-session app) and then destroy the client.

Do we need to change line 196 as well?

nikosdouvlis (approved):

PR #749, opened 2023-01-31
Include isSatellite and domain to interstitial and authenticateRequest

panteliselef:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `build/tooling/chore`

## Description

How to use in a Remix app 
```typescript
export const loader: LoaderFunction = args => {
  return rootAuthLoader(
    args,
    ({ request }) => {....},
    { loadUser: true, isSatellite: true, domain: 'clerk.satellite.dev' },
  );
};
```
and also 
```typescript
export default ClerkApp(App, {
  isSatellite: true,
  domain: 'clerk.satellite.dev',
});
```
⚠️ This is an experimental feature under active development
- You can use these properties and their values when this PR is merged
- You will receive type errors as `isSatellite` & `domain` are not available in types yet

### DIFF
--- a/packages/backend/src/tokens/interstitial.ts
+++ b/packages/backend/src/tokens/interstitial.ts
@@ -15,10 +15,20 @@ export type LoadInterstitialOptions = {
   publishableKey: string;
   pkgVersion?: string;
   debugData?: any;
+  isSatellite?: boolean;
+  domain?: string;
 };
 
-export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions, 'apiUrl'>) {
-  const { debugData, frontendApi, pkgVersion, publishableKey, proxyUrl } = options;
+export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions, 'apiUrl' | 'isSatellite' | 'domain' | 'proxyUrl'>) {
+  const {
+    debugData,
+    frontendApi,
+    pkgVersion,
+    publishableKey,
+    proxyUrl,
+    isSatellite = false,
+    domain,
+  } = options as LoadInterstitialOptions;
   return `
     <head>
         <meta charset="UTF-8" />
@@ -28,6 +38,7 @@ export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions,
             window.__clerk_frontend_api = '${frontendApi}';
             window.__clerk_debug = ${JSON.stringify(debugData || {})};
             ${proxyUrl ? `window.__clerk_proxy_url = ${proxyUrl}` : ''}
+            ${domain ? `window.__clerk_domain = '${domain}'` : ''}
             window.startClerk = async () => {
                 function formRedirect(){
                     const form = '<form method="get" action="" name="redirect"></form>';
@@ -50,13 +61,17 @@ export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions,
 
                 const Clerk = window.Clerk;
                 try {
-                    await Clerk.load();
-                    if(window.location.href.indexOf("#") === -1){
-                      window.location.href = window.location.href;
-                    } else if (window.navigator.userAgent.toLowerCase().includes("firefox/")){
-                        formRedirect();
-                    } else {
-                        window.location.reload();
+                    await Clerk.load({
+                        isSatellite: ${isSatellite},
+                    });
+                    if(Clerk.loaded){
+                      if(window.location.href.indexOf("#") === -1){
+                        window.location.href = window.location.href;
+                      } else if (window.navigator.userAgent.toLowerCase().includes("firefox/")){
+                          formRedirect();
+                      } else {
+                          window.location.reload();
+                      }
                     }
                 } catch (err) {
                     console.error('Clerk: ', err);
@@ -69,9 +84,11 @@ export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions,
                     ? `script.setAttribute('data-clerk-publishable-key', '${publishableKey}');`
                     : `script.setAttribute('data-clerk-frontend-api', '${frontendApi}');`
                 }
+
+                ${domain ? `script.setAttribute('data-clerk-domain', '${domain}');` : ''}
                 ${proxyUrl ? `script.setAttribute('data-clerk-proxy-url', '${proxyUrl}')` : ''};
                 script.async = true;
-                script.src = '${getScriptUrl(frontendApi, pkgVersion)}';
+                script.src = '${getScriptUrl(domain || frontendApi, pkgVersion)}';
                 script.crossOrigin = 'anonymous';
                 script.addEventListener('load', startClerk);
                 document.body.appendChild(script);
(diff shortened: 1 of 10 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -948,7 +952,36 @@ export default class Clerk implements ClerkInterface {
     this.#componentControls?.updateProps(props);
   };
 
-  #loadInStandardBrowser = async (): Promise<void> => {
+  #redirectToSyncWithPrimary = () => {
+    const q = new URLSearchParams({
+      redirect_url: `${window.location.href}?synced=true`,
```
SokratisVidros, line 957: ❓ Is this the function that initiates the syncing process? If yes, it's a bit confusing that the query string is `synced=true` at this point. Am I missing something?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -948,7 +952,36 @@ export default class Clerk implements ClerkInterface {
...
+    });
+    window.location.replace(new URL(`/v1/client/sync?${q.toString()}`, `https://${this.domain}`).toString());
```
SokratisVidros, line 959: 🔧 You can also use `this.navigate`.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -948,7 +952,36 @@ export default class Clerk implements ClerkInterface {
...
+      this.#redirectToSyncWithPrimary();
+      return true;
+    }
+    this.#stripSyncedQueryParam(q);
+    return false;
+  };
+
+  #loadInStandardBrowser = async (): Promise<boolean> => {
+    if (this.#options.isSatellite) {
+      if (this.#shouldSyncWithMain()) {
```
SokratisVidros, line 980: 🔧 The name of this function makes the reader think that it will only do a check and return a boolean. In reality, it also does the redirection and strips the query string parameter. 

I'd suggest to move this functionality inside the #loadInStandardBrowser method, as it will be easier to comprehend the flow.

packages/nextjs/src/client/index.tsx
```diff
@@ -39,7 +51,9 @@ export function ClerkProvider({ children, ...rest }: NextClerkProviderProps): JS
       publishableKey={publishableKey || process.env.NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY || ''}
       clerkJSUrl={clerkJSUrl || process.env.NEXT_PUBLIC_CLERK_JS}
       // @ts-expect-error
-      proxyUrl={proxyUrl || process.env.NEXT_PUBLIC_CLERK_PROXY_URL}
+      proxyUrl={proxyUrl || process.env.NEXT_PUBLIC_CLERK_PROXY_URL || ''}
+      domain={domain || ''}
```
SokratisVidros, line 55: 🔧 We should introduce NEXT_PUBLIC_CLERK_DOMAIN and NEXT_PUBLIC_CLERK_IS_SATELLITE.

SokratisVidros (changes requested):

packages/clerk-js/src/core/clerk.ts
panteliselef: Better naming for the syncing process was introduced [here](https://github.com/clerkinc/javascript/pull/727).

But to give you more context: There we are constructing the return URL which will point back to our app. Once you came back if `synced=true` the syncing process will not fire again (avoiding an infinite loop)

Until now this is handled inside clerk-js, I believe this could also be handled by the `/link` endpoint, but I don't believe this is necessary

packages/clerk-js/src/core/clerk.ts
panteliselef: Fixed [here](https://github.com/clerkinc/javascript/pull/727).

packages/nextjs/src/client/index.tsx
panteliselef: I see your point about a `*CLERK_DOMAIN` env var, but `CLERK_IS_SATELLITE` is supposed to be a boolean. I am not a big fan of converting strings to booleans. I think the `isSatellite` is sufficient.

packages/clerk-js/src/core/clerk.ts
panteliselef: 👍

packages/backend/src/tokens/interstitial.ts
```diff
@@ -26,6 +35,7 @@ export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions,
         <script>
             window.__clerk_frontend_api = '${frontendApi}';
             window.__clerk_debug = ${JSON.stringify(debugData || {})};
+            ${domain ? `window.__clerk_domain = '${domain}'` : ''}
```
dimkl, line 38: ❓ why don't we use `window.__clerk_domain = '${domain}'` ?

packages/backend/src/tokens/request.ts
```diff
@@ -59,10 +59,24 @@ export type AuthenticateRequestOptions = RequiredVerifyTokenOptions &
...
+    /**
+     * @experimental
+     */
+    isSatellite?: boolean;
   };
 
-export async function authenticateRequest(options: AuthenticateRequestOptions): Promise<RequestState> {
-  options.frontendApi = parsePublishableKey(options.publishableKey)?.frontendApi || options.frontendApi || '';
+export async function authenticateRequest(
+  options: Omit<AuthenticateRequestOptions, 'isSatellite' | 'domain'>,
```
dimkl, line 73: ❓ why do we omit those 2 since and use`options as AuthenticateRequestOptions` below since we have added them to the exported `AuthenticateRequestOptions` as **experimental** ?

dimkl (approved):

packages/backend/src/tokens/interstitial.ts
raptisj: @dimkl I think there is no need to include it if it doesn't exist, right?

packages/backend/src/tokens/request.ts
raptisj: @panteliselef @dimkl we should remove them from being experimental and visible to end users.

SokratisVidros (approved):

packages/backend/src/tokens/request.ts
panteliselef: @raptisj  @dimkl  I do this, because I want to hide the types from the function signature, but I want to get Intellisense and no type errors inside the function.

That being said I revised this, so it may worth taking another look

PR #763, opened 2023-02-06
FAPI determines if we're synced

gkats:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

FAPI now appends the `__clerk_synced` query string parameter as a result of syncing a satellite domain with its primary. 

There's no need to include it in the `redirect_url` parameter, but we still have to check its value to determine if a sync has completed.

Refactored the logic a bit so that each function does only one thing.

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -955,6 +955,10 @@ export default class Clerk implements ClerkInterface {
     this.#componentControls?.updateProps(props);
   };
 
+  #hasSynced = () => getClerkQueryParam(clerkSyncedParam) === 'true';
+
+  #clearSynced = () => removeClerkQueryParam(clerkSyncedParam);
+
   #syncWithPrimary = async () => {
     const q = new URLSearchParams({
       redirect_url: window.location.href,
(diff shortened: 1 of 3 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -1165,3 +1163,5 @@ export default class Clerk implements ClerkInterface {
     }
   }
 }
+
+const clerkSyncedParam = '__clerk_synced';
```
dimkl, line 1167: ❓ isn't there a `constants` or something like that module to this?

dimkl (approved):

packages/clerk-js/src/core/clerk.ts
gkats: Good question, I'm not sure. Open to any suggestions of course :smile:.

anagstef: Do you mean [this one](https://github.com/clerkinc/javascript/blob/00b44167c0cc789ebf0bf25af33cfbbe4a310d52/packages/clerk-js/src/core/constants.ts)?

anagstef (approved):

packages/clerk-js/src/core/clerk.ts
SokratisVidros: ☝️

SokratisVidros (approved):

packages/clerk-js/src/core/clerk.ts
panteliselef: I mean [this one](https://github.com/clerkinc/javascript/blob/824727a7bf52e6c30569e6e59476e7a34fed630f/packages/clerk-js/src/utils/getClerkQueryParam.ts#L7)

raptisj: @gkats @panteliselef I think it should go where @anagstef suggested

raptisj (approved):

packages/clerk-js/src/core/clerk.ts
panteliselef: I see, I was not aware of this pattern. I pushed a commit that should fix this :)

panteliselef: @gkats you can squash it into yours

PR #765, opened 2023-02-06
Allow function to be passed in isSatellite prop

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [x] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

After this PR, you should be able to pass function that accepts a parameter of type `URL` and returns a boolean

```typescript
isSatellite={(url) => url.hostname !== 'exaple.com'}
```

We are also introducing the `CLERK_IS_SATELLITE` environment variable with expected values of `'true'` `'false'`

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -147,6 +147,13 @@ export default class Clerk implements ClerkInterface {
     return this.#isReady;
   }
 
+  get isSatellite(): boolean {
+    if (typeof this.#options.isSatellite === 'function') {
+      return this.#options.isSatellite(new URL(window.location.href));
+    }
+    return this.#options.isSatellite || false;
+  }
+
   public constructor(key: string, options?: unknown) {
     key = (key || '').trim();
 
@@ -971,7 +978,7 @@ export default class Clerk implements ClerkInterface {
   };
 
   #loadInStandardBrowser = async (): Promise<boolean> => {
-    if (this.#options.isSatellite) {
+    if (this.isSatellite) {
       if (!this.#hasSynced()) {
         await this.#syncWithPrimary();
         return false;
--- a/packages/remix/src/ssr/authenticateRequest.ts
+++ b/packages/remix/src/ssr/authenticateRequest.ts
@@ -4,15 +4,14 @@ import { Clerk } from '@clerk/backend';
 import { noSecretKeyOrApiKeyError } from '../errors';
 import { getEnvVariable } from '../utils';
 import type { LoaderFunctionArgs, RootAuthLoaderOptions, RootAuthLoaderOptionsWithExperimental } from './types';
-import { parseCookies } from './utils';
+import { handleIsSatelliteBooleanOrFn, parseCookies } from './utils';
 
 /**
  * @internal
  */
 export function authenticateRequest(args: LoaderFunctionArgs, opts: RootAuthLoaderOptions = {}): Promise<RequestState> {
   const { request, context } = args;
-  const { loadSession, loadUser, loadOrganization, authorizedParties, isSatellite } =
-    opts as RootAuthLoaderOptionsWithExperimental;
+  const { loadSession, loadUser, loadOrganization, authorizedParties } = opts as RootAuthLoaderOptionsWithExperimental;
 
   // Fetch environment variables across Remix runtimes.
   // 1. First try from process.env if exists (Node).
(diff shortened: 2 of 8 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -975,7 +975,12 @@ export default class Clerk implements ClerkInterface {
     return false;
   };
 
-  #shouldSyncWithPrimary = () => this.#options.isSatellite && this.#handleSyncedQueryParam();
+  #shouldSyncWithPrimary = () => {
+    if (typeof this.#options.isSatellite === 'function') {
```
SokratisVidros, line 979: 🙃  It's more verbose but easier to follow the flow as the if clause handles the isSatellite calculation and then based on that we invoke another action.

```suggestion
let isSatellite = false;
if (typeof this.#options.isSatellite === 'function') {
  isSatellite = this.#options.isSatellite(new URL(window.location.href));
} else {
  isSatellite = !!this.#options.isSatellite;
}

if (isSatellite) {
  this.#handleSyncedQueryParam();
}
```

SokratisVidros (approved):

packages/clerk-js/src/core/clerk.ts
panteliselef: @SokratisVidros  How about moving the above lines 979-984 inside a public getter ?

SokratisVidros: Fine by me

packages/remix/src/ssr/authenticateRequest.ts
```diff
@@ -6,13 +6,19 @@ import { getEnvVariable } from '../utils';
 import type { LoaderFunctionArgs, RootAuthLoaderOptions, RootAuthLoaderOptionsWithExperimental } from './types';
 import { parseCookies } from './utils';
 
+const handleIsSatelliteBooleanOrFn = (opts: RootAuthLoaderOptionsWithExperimental, url: URL) => {
```
raptisj, line 9: @panteliselef this is a good candidate to move into the `utils` file in the same directory

packages/remix/src/ssr/authenticateRequest.ts
```diff
@@ -6,13 +6,19 @@ import { getEnvVariable } from '../utils';
...
+  if (typeof opts.isSatellite === 'function') {
+    return opts.isSatellite(url);
```
raptisj, line 11: @panteliselef why do we need to pass all options? Judging from the name just passing `isSatellite` will do.

raptisj (approved):

packages/remix/src/ssr/authenticateRequest.ts
panteliselef: @raptisj Yes you are right. Although this is on my list for making it a utility available across packages as we handle similar logic in clerk-js, remix, nextjs.

packages/remix/src/ssr/authenticateRequest.ts
```diff
@@ -41,7 +40,17 @@ export function authenticateRequest(args: LoaderFunctionArgs, opts: RootAuthLoad
     (opts as RootAuthLoaderOptionsWithExperimental).domain ||
     '';
 
-  const proxyUrl = getEnvVariable('CLERK_PROXY_URL') || (context?.CLERK_PROXY_URL as string) || opts.proxyUrl || '';
+  const isSatellite =
+    getEnvVariable('CLERK_IS_SATELLITE') === 'true' ||
+    (context?.CLERK_IS_SATELLITE as string) === 'true' ||
+    handleIsSatelliteBooleanOrFn((opts as RootAuthLoaderOptionsWithExperimental).isSatellite, new URL(request.url)) ||
```
gkats, line 46: :upside_down_face: Isn't it clearer if we simply do something like 
```
    getEnvVariable('CLERK_IS_SATELLITE') === 'true' ||
    (context?.CLERK_IS_SATELLITE as string) === 'true' ||
    opts.isSatellite === true || 
    opts.isSatellite(url);
```
If `isSatellite` is a function it will never be strictly equal to `true`.

gkats (approved):

packages/remix/src/ssr/authenticateRequest.ts
panteliselef: Unfortunately this will not work and we will still need to do explicit type casting or check if `isSatellite ` is type of function because at this point it could either be `false` or `undefined`. And now this gets messy.

That was the reason why I moved this logic to a function

gkats: You're right. The code sample I wrote was not meant to be used as is, just to illustrate my point. :smile: 

Feel free to leave things as they are!

nikosdouvlis: This looks fine to me. 
Please ignore this comment as this is extremely nitpicky, I'm just adding it for the sake of discussion. The only change I'd make would be to change this to 
```
    isSatellite(req, opts)
```
and move the implementation details inside the function because I totally agree with what you said :) :

> That was the reason why I moved this logic to a function

nikosdouvlis (approved):

packages/remix/src/ssr/authenticateRequest.ts
panteliselef: >  isSatellite(req, opts)
That is very similar to what i had before in terms of a parameters. Naming was deferent.

Would it make sense to encapsulate the whole logic of 
``` typescript
getEnvVariable('CLERK_IS_SATELLITE') === 'true' ||
    (context?.CLERK_IS_SATELLITE as string) === 'true' ||
    handleIsSatelliteBooleanOrFn((opts as RootAuthLoaderOptionsWithExperimental).isSatellite, new URL(request.url)) ||
    false;
```    
inside a `isSatellite(args: LoaderFunctionArgs, opts)` helper function ?

PR #779, opened 2023-02-08
fix(clerk-react): Add `redirectWithAuth` to isomorphicClerk

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Adds redirectWithAuth to IsomorphicClerk, so it is correctly returned by useClerk during runtime

### DIFF
--- a/packages/react/src/isomorphicClerk.ts
+++ b/packages/react/src/isomorphicClerk.ts
@@ -1,6 +1,7 @@
 import type {
   ActiveSessionResource,
   AuthenticateWithMetamaskParams,
+  Clerk,
   ClientResource,
   CreateOrganizationParams,
   CreateOrganizationProps,
@@ -44,6 +45,7 @@ type GenericFunction<TArgs = never> = (...args: TArgs[]) => unknown;
 type MethodName<T> = {
   [P in keyof T]: T[P] extends GenericFunction ? P : never;
 }[keyof T];
+
 type MethodCallback = () => Promise<unknown> | unknown;
 
 export default class IsomorphicClerk {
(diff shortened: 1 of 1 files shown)


SokratisVidros (approved):

dimkl (approved):

dimkl:
If we are going to keep the `IsomorphicClerk`, i think we should use an interface to make `IsomorphicClerk` and `Clerk` expose the same methods to avoid future inconsistencies.

nikosdouvlis:
> If we are going to keep the `IsomorphicClerk`, i think we should make `IsomorphicClerk` use the same interface with `Clerk` to enforce exposing the same methods. This will allow us to avoid future inconsistencies.

We definitely should. Alternatively, we can re-open a previous PR that replaced isomorphicClerk with a dynamic proxy-based object.

We're closer but still a good deal away from removing hotloading completely, so let me try to find the PR i mentioned above

PR #784, opened 2023-02-09
chore(backend): Improved types for interstitial rules

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Removed unnecessary async from functions
Created `InterstitialRuleResult`
Reduced occasions of `as any`

### DIFF
--- a/packages/backend/src/tokens/interstitialRule.ts
+++ b/packages/backend/src/tokens/interstitialRule.ts
@@ -2,10 +2,13 @@ import { isDevelopmentFromApiKey, isProductionFromApiKey } from '../util/instanc
 import { checkCrossOrigin } from '../util/request';
 import type { RequestState } from './authStatus';
 import { AuthErrorReason, interstitial, signedIn, signedOut } from './authStatus';
-import type { AuthenticateRequestOptionsWithExperimental } from './request';
+import type { AuthenticateRequestOptions, AuthenticateRequestOptionsWithExperimental } from './request';
 import { verifyToken } from './verify';
 
-export type InterstitialRule = <T>(opts: T) => Promise<RequestState | undefined>;
+type InterstitialRuleResult = RequestState | undefined;
+type InterstitialRule = <T extends AuthenticateRequestOptions>(
+  opts: T,
+) => Promise<InterstitialRuleResult> | InterstitialRuleResult;
 
 // In development or staging environments only, based on the request's
 // User Agent, detect non-browser requests (e.g. scripts). Since there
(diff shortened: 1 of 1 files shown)


dimkl:
☁️ The interstitial rules was changed:
- to be async for consistency since there is at least 1 rule that is async
- to use generic options to avoid cyclic dependency in types 
   request.ts **imports** interstitialRule.ts which **imports** AuthenticateRequestOptions from request.ts
  I would prefer that we don't apply those changes and discuss about handling with some other way. If it makes more sense to you and the team,i will approve it.

SokratisVidros (approved):

packages/backend/src/tokens/interstitialRule.ts
```diff
@@ -49,8 +53,8 @@ export const potentialFirstLoadInDevWhenUATMissing: InterstitialRule = async opt
   return undefined;
 };
 
-export const potentialRequestAfterSignInOrOurFromClerkHostedUiInDev: InterstitialRule = async options => {
-  const { apiKey, secretKey, referrer, host, forwardedHost, forwardedPort, forwardedProto } = options as any;
+export const potentialRequestAfterSignInOrOurFromClerkHostedUiInDev: InterstitialRule = options => {
```
raptisj, line 56: @panteliselef isn't this a typo in the function name? `...InOrOutFrom...` instead of `...InOrOurFrom...`

panteliselef: Yes it seems like it is. @dimkl can you verify ?

SokratisVidros (approved):

packages/backend/src/tokens/interstitialRule.ts
dimkl: yes you are correct, it's a typo.

PR #786, opened 2023-02-10
feat(nextjs): Improve types and add support for localization/appearance in app-beta

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Localization is easy. For appearance, the simplest solution seems to be to get rid of the function in our prebuilt themes, since there is no need for it right now (we just return an object without using the arguments). If a user wants to use the callback for elements he will get the error from next.js (as with any function prop while in a server component) and will have to add the use client directive in his file (same thing would apply if the developer wanted to pass the `navigate` prop).

Closes #637

### DIFF
--- a/packages/nextjs/src/app-beta/ClerkProvider.tsx
+++ b/packages/nextjs/src/app-beta/ClerkProvider.tsx
@@ -7,7 +7,6 @@ import { ClerkProvider as ClerkProviderClient } from './client/ClerkProvider';
 
 type NextAppClerkProviderProps = {
   children: React.ReactNode;
-  initialState?: InitialState;
 } & Omit<IsomorphicClerkOptions, keyof PublishableKeyOrFrontendApi> &
   Partial<PublishableKeyOrFrontendApi>;
 
--- a/packages/nextjs/src/app-beta/ClerkProvider.tsx
+++ b/packages/nextjs/src/app-beta/ClerkProvider.tsx
@@ -1,9 +1,26 @@
+import type { IsomorphicClerkOptions } from '@clerk/clerk-react/dist/types';
+import type { InitialState, PublishableKeyOrFrontendApi } from '@clerk/types';
 import React from 'react';
 
 import { initialState } from './auth';
 import { ClerkProvider as ClerkProviderClient } from './client/ClerkProvider';
 
-export function ClerkProvider({ children }: { children: React.ReactNode }) {
-  const state = initialState()?.__clerk_ssr_state || { sessionId: null, orgId: null, userId: null };
-  return <ClerkProviderClient initialState={state}>{children}</ClerkProviderClient>;
+type NextAppClerkProviderProps = {
+  children: React.ReactNode;
+  initialState?: InitialState;
+} & Omit<IsomorphicClerkOptions, keyof PublishableKeyOrFrontendApi> &
+  Partial<PublishableKeyOrFrontendApi>;
+
+export function ClerkProvider(props: NextAppClerkProviderProps) {
+  const state = (initialState()?.__clerk_ssr_state || { sessionId: null, orgId: null, userId: null }) as InitialState;
+  return (
+    <ClerkProviderClient
+      frontendApi={process.env.NEXT_PUBLIC_CLERK_FRONTEND_API || ''}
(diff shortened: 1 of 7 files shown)


packages/nextjs/src/app-beta/ClerkProvider.tsx
```diff
@@ -1,9 +1,26 @@
...
 
 import { initialState } from './auth';
 import { ClerkProvider as ClerkProviderClient } from './client/ClerkProvider';
 
-export function ClerkProvider({ children }: { children: React.ReactNode }) {
-  const state = initialState()?.__clerk_ssr_state || { sessionId: null, orgId: null, userId: null };
-  return <ClerkProviderClient initialState={state}>{children}</ClerkProviderClient>;
+type NextAppClerkProviderProps = {
+  children: React.ReactNode;
+  initialState?: InitialState;
```
panteliselef, line 10: ❓ Why do we include `initialState` here but not in `NextClerkProviderProps`

panteliselef:
@desiprisg  Does this PR also resolve this #637  ?

nikosdouvlis:
@desiprisg 
As far as I understand, this PR does not introduce support for the full `appearance` API but instead it provides a temp workaround.

Do you have any more insight on what needs to be done for us to support passing functions/non-serializable values to ClerkProvider? Eg a combination of client and server side providers etc.

desiprisg:
@nikosdouvlis I mean, we can calculate the appearance object after it is passed to our first provider (since we use a server provider that renders a client provider) and pass that to the client one. But that would only work for `ClerkProvider` and not the other components, as they would not be aware of the `variables` set higher up. Does that make sense?

packages/nextjs/src/app-beta/ClerkProvider.tsx
desiprisg: This might be unnecessary, I'll take a look and get back to you

desiprisg: Yeap, it shouldn't be there. Good catch!

desiprisg:
If we go ahead with this, I'd prefer we don't have 2 `ClerkProvider` components, but instead only one, which will be a client component. This will make it more apparent to the developer, that the error on passing a function is not a Clerk issue, but something that just isn't possible (server component - - -(function prop)- - > client component.

packages/nextjs/src/app-beta/client/ClerkProvider.tsx
```diff
@@ -1,5 +1,6 @@
 'use client';
 // !!! Note the import from react
+import type { ClerkProviderProps } from '@clerk/clerk-react';
 import { ClerkProvider as ReactClerkProvider } from '@clerk/clerk-react';
```
nikosdouvlis, line 4: @desiprisg Any ideas why this still comes from clerk/react and not clerk/nextjs?

desiprisg: We probably can reuse it. I'll try to clean it up.

desiprisg: Actually, we can't right now since the `useRouter` hook needs to be imported from `next/navigation` and not `next/router`, like it's done here.https://github.com/clerkinc/javascript/blob/46fcea92c69ef052eeebd54c5adf9ef82272e3fd/packages/nextjs/src/client/index.tsx#L4

nikosdouvlis (approved):

PR #787, opened 2023-02-10
Fastify touch ups 

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/packages/backend/src/tokens/interstitial.ts
+++ b/packages/backend/src/tokens/interstitial.ts
@@ -19,7 +19,10 @@ export type LoadInterstitialOptions = {
   domain?: string;
 };
 
-export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions, 'apiUrl' | 'isSatellite' | 'domain' | 'proxyUrl'>) {
+export function loadInterstitialFromLocal(
+  options: Omit<LoadInterstitialOptions, 'apiUrl' | 'isSatellite' | 'domain' | 'proxyUrl'>,
+) {
+  options.frontendApi = parsePublishableKey(options.publishableKey)?.frontendApi || options.frontendApi || '';
   const {
     debugData,
     frontendApi,
--- a/packages/fastify/src/types.ts
+++ b/packages/fastify/src/types.ts
@@ -1,0 +1,3 @@
+import type { ClerkOptions } from '@clerk/backend';
+
+export type ClerkFastifyOptions = Omit<ClerkOptions, 'apiKey'>;
--- a/packages/fastify/src/withClerkMiddleware.ts
+++ b/packages/fastify/src/withClerkMiddleware.ts
@@ -1,14 +1,16 @@
-import type { ClerkOptions } from '@clerk/types';
 import { parse } from 'cookie';
 import type { FastifyReply, FastifyRequest } from 'fastify';
 
 import { clerkClient } from './clerkClient';
 import * as constants from './constants';
+import type { ClerkFastifyOptions } from './types';
 import { getSingleValueFromArrayHeader } from './utils';
 
-export const withClerkMiddleware = (options: ClerkOptions) => {
+export const withClerkMiddleware = (options: ClerkFastifyOptions) => {
   return async (req: FastifyRequest, reply: FastifyReply) => {
     const cookies = parse(req.raw.headers.cookie || '');
+    const secretKey = options.secretKey || constants.SECRET_KEY;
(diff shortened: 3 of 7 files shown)


packages/fastify/src/types.ts
```diff
@@ -0,0 +1,3 @@
+import type { ClerkOptions } from '@clerk/backend';
+
+export type ClerkFastifyOptions = Omit<ClerkOptions, 'apiKey'>;
```
dimkl, line 3: 🙃 to avoid maintenance of this type on changes i would suggest 1 of the following: 
```
//1.
export type { ClerkOptions as ClerkFastifyOptions} from '@clerk/backend';
// 2. or don't even define a separate ClerkFastifyOptions and use ClerkOptions 
```

packages/backend/src/tokens/interstitial.ts
```diff
@@ -19,7 +19,10 @@ export type LoadInterstitialOptions = {
   domain?: string;
 };
 
-export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions, 'apiUrl' | 'isSatellite' | 'domain' | 'proxyUrl'>) {
+export function loadInterstitialFromLocal(
+  options: Omit<LoadInterstitialOptions, 'apiUrl' | 'isSatellite' | 'domain' | 'proxyUrl'>,
+) {
+  options.frontendApi = parsePublishableKey(options.publishableKey)?.frontendApi || options.frontendApi || '';
```
dimkl, line 25: 🎉 This fixes an issue i have with interstitial in remix app using publishableKey.

packages/fastify/src/withClerkMiddleware.ts
```diff
@@ -1,14 +1,16 @@
...
 import * as constants from './constants';
+import type { ClerkFastifyOptions } from './types';
 import { getSingleValueFromArrayHeader } from './utils';
 
-export const withClerkMiddleware = (options: ClerkOptions) => {
+export const withClerkMiddleware = (options: ClerkFastifyOptions) => {
   return async (req: FastifyRequest, reply: FastifyReply) => {
     const cookies = parse(req.raw.headers.cookie || '');
+    const secretKey = options.secretKey || constants.SECRET_KEY;
+    const publishableKey = options.publishableKey || constants.PUBLISHABLE_KEY;
```
dimkl, line 13: ❓ should we create a util method in backend to load
```
/*
 * prefix: should end with a `_` if provided
 */
const loadOptionsFromEnv = ( prefix?: string = '' ): Record<string, string | undefined> => {
  return {
   secretKey: process.env[prefix + 'CLERK_API_KEY],
   publishableKey: process.env[prefix + 'CLERK_PUBLISHABLE_KEY],
  }
}

// example calling it from remix
const envOptions = loadFromEnv();
// example calling it from next
const envOptions = loadFromEnv('NEXT_PUBLIC_');
```
and merge it with options like this?
```
authenticateRequest({
  ...envOptions,
  ...options,
  //...rest
})
```

nikosdouvlis: Definitely, 100% agree.
A layer to abstract this is necessary. I was thinking about a mechanism of some kind that would act as an adapter, eg: accept a method that reads the any config we need in a framework-specific way (prefixes might not be enough) and output a framework-agnostic config object we will use internally.

WDYT?

Cc @SokratisVidros

packages/fastify/src/types.ts
nikosdouvlis: Valid point Dimitri; we cannot remove `apiKey` from the common backend type right now - otherwise, we will break backward compatibility.

In general, though, I like separate high-level types on a per-framework basis because it's easy to make fine-grained adjustments later without introducing more types.

PR #810, opened 2023-02-14
fix(clerk-js): Don't append width url var for external images

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Don't append the `width` url variable for external images since they may return an error, and initials will be displayed instead of the image.


packages/clerk-js/src/ui/common/constants.ts
```diff
@@ -88,3 +88,5 @@ export const ERROR_CODES = {
   OAUTH_ACCESS_DENIED: 'oauth_access_denied',
   NOT_ALLOWED_ACCESS: 'not_allowed_access',
 };
+
+export const CLERK_IMAGE_BASE_URL = 'https://image.clerk.dev/';
```
SokratisVidros, line 92: 🔧 We need to cater for staging and local development enviroment.

SokratisVidros (changes requested):

packages/clerk-js/src/ui/common/constants.ts
desiprisg: Right! PTAL.

nikosdouvlis (approved):

packages/clerk-js/src/utils/url.ts
```diff
@@ -319,6 +324,10 @@ export function hasBannedProtocol(val: string | URL) {
   return BANNED_URI_PROTOCOLS.some(bp => bp === protocol);
 }
 
+export const isClerkImage = (src?: string): boolean => {
+  return !!CLERK_IMAGE_URL_BASES.some(base => src?.includes(base));
```
dimkl, line 328: 🙃 Why `src` param is optional? I think we should make it mandatory and avoid `?.` in the function body.

PR #814, opened 2023-02-15
Jit ignore playground

dimkl:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/.jit/config.yml
+++ b/.jit/config.yml
@@ -1,4 +1,33 @@
 folders:
-  - path: /
+  - path: /packages/backend
     exclude:
       - ./**/*.test.ts
+  - path: /packages/clerk-js
+    exclude:
+      - ./**/*.test.ts
+  - path: /packages/eslint-config-custom
+  - path: /packages/expo
+  - path: /packages/fastify
+    exclude:
+      - ./**/*.test.ts
+  - path: /packages/gatsby-plugin-clerk
+    exclude:
+      - ./**/*.test.ts
+  - path: /packages/localizations
+  - path: /packages/nextjs
+    exclude:
+      - ./**/*.test.ts
+  - path: /packages/react
+    exclude:
+      - ./**/*.test.ts
+  - path: /packages/remix
+    exclude:
+      - ./**/*.test.ts
+  - path: /packages/sdk-node
+    exclude:
+      - ./**/*.test.ts
+  - path: /packages/shared
+    exclude:
+      - ./**/*.test.ts
+  - path: /packages/themes
+  - path: /packages/types


dimkl:
This PR will update the **Jit** config and to be executed only for `packages/`.
Pros:
- issue with `"Invalid Version: file:packages/eslint-config-custom"` will be resolved
- no issues for playground apps and development dependencies of monorepo

Cons:
- security is focused on part of monorepo and secrets or issues in other parts may slip
- each time a new package or part of the code is added and we want to include it to JIT we should update the config.yml
- issues in dependencies of monorepo (eg lerna, ...) will not be alerted

I'm not so sure if we should proceed with this. The reason we checked this, is to avoid always having the JIT failing when a change is made to a dependency in any part of the monorepo.
I think we have notified JIT team to check for our issue. It might be better to communicate again and wait for a solution from their side.
cc: @SokratisVidros , @nikosdouvlis

nikosdouvlis:
> This PR will update the **Jit** config and to be executed only for `packages/`. Pros:
> 
> * issue with `"Invalid Version: file:packages/eslint-config-custom"` will be resolved
> * no issues for playground apps and development dependencies of monorepo
> 
> Cons:
> 
> * security is focused on part of monorepo and secrets or issues in other parts may slip
> * each time a new package or part of the code is added and we want to include it to JIT we should update the config.yml
> * issues in dependencies of monorepo (eg lerna, ...) will not be alerted
> 
> I'm not so sure if we should proceed with this. The reason we checked this, is to avoid always having the JIT failing when a change is made to a dependency in any part of the monorepo. I think we have notified JIT team to check for our issue. It might be better to communicate again and wait for a solution from their side. cc: @SokratisVidros , @nikosdouvlis

Can we simply exclude the `playground/*` directory? I think we'd want everything else to be includes, except for the playground dir that's not in the scope of the monorepo itself.

dimkl:
I tried to exclude only the playground but couldn't make it work and also that will not resolve the issue since the `/package-lock.json` will cause the "Invalid version" issue.

nikosdouvlis (approved):

PR #818, opened 2023-02-15
AUTH-134 fix(clerk-js,types): Resume an existing sign-up on SignUpStart only i…

yourtallness:
…f it's a ticket flow

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [X] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

In the case of a ticket flow with missing requirements, the sign-up created for the ticket will be resumed when a social connection is initiated (since it may help complete the setup).

However this was done unconditionally in `authenticateWithRedirect()` and introduced an interaction with back button navigation from a verification screen back to the initial sign-up screen.

This PR explicitly instructs `authenticateWithRedirect()` to resume an existing sign-up if it's a ticket flow scenario, otherwise it will default to creating a new sign-up.

Note: Needs further manual testing & unit tests.

### DIFF
--- a/packages/clerk-js/src/core/resources/SignUp.ts
+++ b/packages/clerk-js/src/core/resources/SignUp.ts
@@ -178,10 +178,14 @@ export class SignUp extends BaseResource implements SignUpResource {
     return this.authenticateWithWeb3({ identifier, generateSignature: generateSignatureWithMetamask });
   };
 
-  public authenticateWithRedirect = async (params: AuthenticateWithRedirectParams): Promise<void> => {
-    const { redirectUrl, redirectUrlComplete, strategy } = params || {};
+  public authenticateWithRedirect = async ({
+    redirectUrl,
+    redirectUrlComplete,
+    strategy,
+    continueSignUp = false,
+  }: AuthenticateWithRedirectParams): Promise<void> => {
     const authenticateFn = (args: SignUpCreateParams | SignUpUpdateParams) =>
-      this.id ? this.update(args) : this.create(args);
+      continueSignUp && this.id ? this.update(args) : this.create(args);
     const { verifications } = await authenticateFn({
       strategy,
       redirectUrl,
--- a/packages/clerk-js/src/ui/components/SignUp/SignUpSocialButtons.tsx
+++ b/packages/clerk-js/src/ui/components/SignUp/SignUpSocialButtons.tsx
@@ -10,7 +10,9 @@ import { SocialButtons } from '../../elements/SocialButtons';
 import { useNavigate } from '../../hooks';
 import { handleError } from '../../utils';
 
-export const SignUpSocialButtons = React.memo((props: SocialButtonsProps) => {
+export type SignUpSocialButtonsProps = SocialButtonsProps & { continueSignUp?: boolean };
+
+export const SignUpSocialButtons = React.memo((props: SignUpSocialButtonsProps) => {
   const clerk = useCoreClerk();
   const { navigate } = useNavigate();
   const card = useCardState();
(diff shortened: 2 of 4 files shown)


packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -203,7 +207,9 @@ export class SignUp extends BaseResource implements SignUpResource {
     });
   };
 
-  protected fromJSON(data: SignUpJSON | null): this {
+  protected;
+
+  fromJSON(data: SignUpJSON | null): this {
```
gkats, line 212: :wrench: Is this change intentional? I believe the `protected` modifier should come right before the method definition.

packages/types/src/oauth.ts
```diff
@@ -244,4 +244,8 @@ export type AuthenticateWithRedirectParams = {
    * Full URL or path to navigate after the OAuth flow completes.
    */
   redirectUrlComplete: string;
+  /**
+   * Whether to continue (i.e. PATCH) an existing SignUp (if present) or create a new SignUp
+   */
+  continueSignUp?: boolean;
```
gkats, line 250: :upside_down_face: Not sure about property naming and whether this property should be part of the public API. 

How about `shouldCreateNew`? Let's see what the others think.

gkats (commented):
Looks good to me overall!

packages/clerk-js/src/core/resources/SignUp.ts
yourtallness: Messed up while auto-formatting, thanks for spotting

packages/clerk-js/src/ui/components/SignUp/SignUpSocialButtons.tsx
```diff
@@ -10,7 +10,9 @@ import { SocialButtons } from '../../elements/SocialButtons';
 import { useNavigate } from '../../hooks';
 import { handleError } from '../../utils';
 
-export const SignUpSocialButtons = React.memo((props: SocialButtonsProps) => {
+export type SignUpSocialButtonsProps = SocialButtonsProps & { continueSignUp?: boolean };
```
dimkl, line 13: ❓ why didn't we add `continueSignUp?: boolean` to `SocialButtonsProps` type in `elements/SocialButtons` ?

dimkl (approved):

packages/clerk-js/src/ui/components/SignUp/SignUpSocialButtons.tsx
yourtallness: Because it doesn't apply to sign-in.

packages/types/src/oauth.ts
yourtallness: Since `shouldCreateNew` should be true by default, I'd prefer to go for a flag that should be false by default.

We have the concept of "continuing a sign-up" already, so it's not a new term.

Will proceed as is to unblock this fix, but open to revisiting this.

PR #822, opened 2023-02-16
Fix(clerk-js): Sync __session cookie on token:update event

dimkl:
## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

- [ ] add tests after PR approval

Introduce `core/eventBus` singleton module to support event bus architecture. To fix the issue of cookie not being in sync after `getToken()` is being called we have introduced an `token:update` custom event with the `TokenResource` passed in event data when token is being set and added an event listener to update the cookie using that event and it's token resource.

The `core/eventBus` module is internal so it's not being exposed.

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -40,6 +40,7 @@ import type {
   SignOutOptions,
   SignUpProps,
   SignUpResource,
+  TokenResource,
   UnsubscribeCallback,
   UserButtonProps,
   UserProfileProps,
@@ -82,6 +83,7 @@ import {
   clerkMissingProxyUrlAndDomain,
   clerkOAuthCallbackDidNotCompleteSignInSIgnUp,
 } from './errors';
+import { eventBus } from './eventBus';
 import type { FapiClient, FapiRequestCallback } from './fapiClient';
 import createFapiClient from './fapiClient';
 import {
@@ -1120,6 +1122,13 @@ export default class Clerk implements ClerkInterface {
       if (data.type === 'signout') {
         void this.handleUnauthenticated({ broadcast: false });
       }
+    });
+
+    // set cookie on token update
+    eventBus.addEventListener('token:update', event => {
+      const tokenEvent = event as CustomEvent;
+      const token = tokenEvent.detail.token as TokenResource;
+      this.#authService?.setAuthCookiesFromToken(token.getRawString());
     });
   };
 
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -40,7 +40,6 @@ import type {
   SignOutOptions,
   SignUpProps,
   SignUpResource,
-  TokenResource,
   UnsubscribeCallback,
   UserButtonProps,
   UserProfileProps,
@@ -83,7 +82,7 @@ import {
   clerkMissingProxyUrlAndDomain,
   clerkOAuthCallbackDidNotCompleteSignInSIgnUp,
 } from './errors';
-import { eventBus } from './eventBus';
+import { eventBus, events } from './events';
 import type { FapiClient, FapiRequestCallback } from './fapiClient';
 import createFapiClient from './fapiClient';
 import {
@@ -1125,10 +1124,8 @@ export default class Clerk implements ClerkInterface {
     });
 
     // set cookie on token update
-    eventBus.addEventListener('token:update', event => {
-      const tokenEvent = event as CustomEvent;
-      const token = tokenEvent.detail.token as TokenResource;
-      this.#authService?.setAuthCookiesFromToken(token.getRawString());
+    eventBus.on(events.TokenUpdate, ({ token }) => {
+      this.#authService?.setAuthCookiesFromToken(token?.getRawString());
     });
   };
 
--- a/packages/clerk-js/src/core/events.ts
+++ b/packages/clerk-js/src/core/events.ts
@@ -14,14 +14,38 @@ type EventPayload = {
 };
 
 class ClerkEventBus extends EventTarget {
+  #handlers = {} as Record<ClerkEvent, Array<any>>;
+
   on<E extends ClerkEvent, P extends EventPayload[E]>(eventName: E, callback: (evt: P) => void): void {
-    this.addEventListener(eventName, (e: Event) => {
+    const handler = (e: Event) => {
       callback((e as CustomEvent).detail);
-    });
+    };
+
+    if (!this.#handlers[eventName]) {
+      this.#handlers[eventName] = [];
+    }
+    this.#handlers[eventName].push({ callback, handler });
+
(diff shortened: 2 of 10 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -1098,6 +1100,13 @@ export default class Clerk implements ClerkInterface {
         void this.handleUnauthenticated({ broadcast: false });
       }
     });
+
+    // set cookie on token update
+    eventBus.addEventListener('token:update', event => {
```
SokratisVidros, line 1105: 🙃  We should consider centralizing event definitions

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1098,6 +1100,13 @@ export default class Clerk implements ClerkInterface {
...
+      const tokenEvent = event as CustomEvent;
+      const token = tokenEvent.detail.token as TokenResource;
```
SokratisVidros, line 1107:
```suggestion
      const token = tokenEvent?.detail?.token as TokenResource;
```

SokratisVidros: If token is not available we should noop

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -84,7 +85,10 @@ export class Session extends BaseResource implements SessionResource {
     const path = template ? `${this.path()}/tokens/${template}` : `${this.path()}/tokens`;
     const tokenResolver = Token.create(path);
     SessionTokenCache.set({ tokenId, tokenResolver });
-    return tokenResolver.then(res => res.getRawString());
+    return tokenResolver.then(token => {
+      eventBus.dispatchEvent(new CustomEvent('token:update', { detail: { token } }));
```
SokratisVidros, line 89: 🔧 I'd suggest abstracting the CustomEvent via a createEvent helper that belongs to the eventBus module. That way, we can change the internal implementation of the event bus without touching all the event emission points.

packages/clerk-js/src/core/services/authentication/AuthenticationService.ts
```diff
@@ -35,6 +35,11 @@ export class AuthenticationService {
     this.setClientUatCookieForDevelopmentInstances();
   }
 
+  public setAuthCookiesFromToken(token: string): void {
```
SokratisVidros, line 38:
```suggestion
  public onTokenUpdate(token: string): void {
```

SokratisVidros (changes requested):

packages/clerk-js/src/core/resources/Session.ts
dimkl: Since the `addEventListener` will depend on the `CustomEvent` to parse the data from `event` argument should i also introduce a `eventBus.parseEvent` ?

SokratisVidros: Yes. Or just introduce an `on` function.

packages/clerk-js/src/core/services/authentication/AuthenticationService.ts
dimkl: Why name this Service method based on where it's gonna be used and not based on what it does?

packages/clerk-js/src/core/clerk.ts
dimkl: `event` and `token` should be available but the content of the token could be empty. I have added types to enforce this.
see: 5d67113ea7e6ec42987234fac948739462da8ee4

packages/clerk-js/src/core/clerk.ts
dimkl: like this ? 5d67113ea7e6ec42987234fac948739462da8ee4

SokratisVidros (approved):

packages/clerk-js/src/core/events.ts
```diff
@@ -0,0 +1,25 @@
+import type { TokenResource } from '@clerk/types';
+
+export const eventBus = new EventTarget();
+
+export const createEvent = <T extends EventType>(eventName: string, data: T): CustomEvent => {
```
nikosdouvlis, line 5: This looks great!

I've got a few minor comments/ideas regarding the "ergonomics" of the internal event bus API but these can be either ignored or tackled in the future. 

1. Individual exports vs an `events` constant
Instead of exporting each event on its own, I'd probably group them under an `events` constant: 

```tsx
export const events = {
  TokenUpdate: 'token:update'
} as const;
```
(or even namespaced, eg `events.Token.Update`)

2. `createEvent` can constraint `eventName` based on the available event

3. We can create an association map between the event name and the event payload and use it in createEvent/parseEvent for extra type safety, a quick example: 
```
type TokenResource = {
    lastUpdatedAt: Date;
}

export const eventBus = new EventTarget();

export const events = {
    TokenUpdate: 'token:update',
    UserSignedIn: 'user:signedIn'
} as const

type Event = typeof events[keyof typeof events];
type EventType<T> = Record<string, unknown> & T;

type TokenUpdatePayload = { token: TokenResource | null };
type UserSignedInPayload = { user: { firstName: string } | null };

type EventPayload = {
    [events.TokenUpdate]: EventType<TokenUpdatePayload>,
    [events.UserSignedIn]: EventType<UserSignedInPayload>
}


export const createEvent = <E extends Event, P extends EventPayload[E]>(eventName: E, data: P): CustomEvent => {
    return new CustomEvent(eventName, { detail: data });
};

```
image

4. We can wrap eventBus and expose an `on` prop so we can have strict types on parseEvent as well and avoid the `addEventListener` etc boilerplate, eg: `eventBus.on(events.TokenUpdate, ({user}) => user....  )'. Similarly, we could have a `dispatch` method

5. (optional, nit-picky and debatable :) ) A final touch would be to make `.on` and `.dispatch` accept a function as the first argument pass the events as a prop so we don't need the explicit import, eg
```
eventBus.on(e => e.TokenUpdate, ({ user }) => {} )
```

nikosdouvlis (approved):
💯

packages/clerk-js/src/core/events.ts
nikosdouvlis: After discussing with @dimkl offline, we decided that point 5 makes no sense to let's ignore it :)

PR #829, opened 2023-02-17
COR-136 Use proxyUrl for as src for script

panteliselef:
## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

- Uses proxyUrl as src of clerk script
- Validates for absoluteUrl proxyUrl and throws error in Remix & nextjs

### DIFF
--- a/packages/backend/src/tokens/interstitial.ts
+++ b/packages/backend/src/tokens/interstitial.ts
@@ -93,7 +93,7 @@ export function loadInterstitialFromLocal(
                 ${domain ? `script.setAttribute('data-clerk-domain', '${domain}');` : ''}
                 ${proxyUrl ? `script.setAttribute('data-clerk-proxy-url', '${proxyUrl}')` : ''};
                 script.async = true;
-                script.src = '${getScriptUrl(frontendApi, pkgVersion)}';
+                script.src = '${getScriptUrl(proxyUrl || frontendApi, pkgVersion)}';
                 script.crossOrigin = 'anonymous';
                 script.addEventListener('load', startClerk);
                 document.body.appendChild(script);
@@ -174,6 +174,7 @@ const getClerkJsMajorVersionOrTag = (frontendApi: string, pkgVersion?: string) =
 };
 
 const getScriptUrl = (frontendApi: string, pkgVersion?: string) => {
+  const noSchemeFrontendApi = frontendApi.replace(/http(s)?:\/\//, '');
   const major = getClerkJsMajorVersionOrTag(frontendApi, pkgVersion);
-  return `https://${frontendApi}/npm/@clerk/clerk-js@${major}/dist/clerk.browser.js`;
+  return `https://${noSchemeFrontendApi}/npm/@clerk/clerk-js@${major}/dist/clerk.browser.js`;
 };
--- a/packages/backend/src/util/parsePublishableKey.ts
+++ b/packages/backend/src/util/parsePublishableKey.ts
@@ -13,7 +13,7 @@ export function buildPublishableKey(frontendApi: string): string {
   return `${keyPrefix}${btoa(`${frontendApi}$`)}`;
 }
 
-export function parsePublishableKey(key: string): PublishableKey | null {
+export function parsePublishableKey(key: string | undefined): PublishableKey | null {
   key = key || '';
 
   if (!isPublishableKey(key)) {
--- a/packages/nextjs/src/server/utils.ts
+++ b/packages/nextjs/src/server/utils.ts
@@ -144,3 +144,8 @@ export const handleIsSatelliteBooleanOrFn = (opts: WithAuthOptionsExperimental, 
   }
   return opts.isSatellite || false;
 };
+
+// TODO: use @clerk/shared once is tree-shakeable
+export function isHttpOrHttps(key: string | undefined) {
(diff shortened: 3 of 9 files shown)


packages/backend/src/tokens/interstitial.ts
```diff
@@ -93,7 +93,7 @@ export function loadInterstitialFromLocal(
                 ${domain ? `script.setAttribute('data-clerk-domain', '${domain}');` : ''}
                 ${proxyUrl ? `script.setAttribute('data-clerk-proxy-url', '${proxyUrl}')` : ''};
                 script.async = true;
-                script.src = '${getScriptUrl(frontendApi, pkgVersion)}';
+                script.src = '${getScriptUrl(proxyUrl?.replace(/http(s)?:\/\//, '') || frontendApi, pkgVersion)}';
```
SokratisVidros, line 96: ❓ Should we maybe move this inside getScriptUrl?

SokratisVidros (approved):

packages/backend/src/util/parsePublishableKey.ts
```diff
@@ -13,7 +13,7 @@ export function buildPublishableKey(frontendApi: string): string {
   return `${keyPrefix}${btoa(`${frontendApi}$`)}`;
 }
 
-export function parsePublishableKey(key: string): PublishableKey | null {
+export function parsePublishableKey(key: string | undefined): PublishableKey | null {
```
dimkl, line 16: ❓ why did we allowed the `undefined` here? Was there a bug reported for legacy (using frontendApi) customers ?

packages/nextjs/src/server/utils.ts
```diff
@@ -144,3 +144,8 @@ export const handleIsSatelliteBooleanOrFn = (opts: WithAuthOptionsExperimental,
   }
   return opts.isSatellite || false;
 };
+
+// TODO: use @clerk/shared once is tree-shakeable
```
dimkl, line 148: 👍

packages/nextjs/src/server/withClerkMiddleware.ts
```diff
@@ -44,6 +45,10 @@ export const withClerkMiddleware: WithClerkMiddleware = (...args: unknown[]) =>
 
   const proxyUrl = opts?.proxyUrl || PROXY_URL;
 
+  if (!!proxyUrl && !isHttpOrHttps(proxyUrl)) {
```
dimkl, line 48: 🙃 i think we `isHttpOrHttps(proxyUrl)` covers the case of falsy proxyUrl. Do we need `!!proxyUrl &&` check ?

packages/nextjs/src/server/withClerkMiddleware.ts
```diff
@@ -44,6 +45,10 @@ export const withClerkMiddleware: WithClerkMiddleware = (...args: unknown[]) =>
...
+    throw new Error(`Only a absolute URL that starts with https is allowed to be used in SSR`);
```
dimkl, line 49: @nikosdouvlis i think we should add a ticket to create `errors.ts` to nextjs, like our other packages (eg remix).

packages/shared/src/utils/proxy.ts
```diff
@@ -3,7 +3,11 @@ export function isValidProxyUrl(key: string | undefined) {
     return true;
   }
 
-  return key.startsWith('https://') || isProxyUrlRelative(key);
+  return isHttpOrHttps(key) || isProxyUrlRelative(key);
+}
+
+export function isHttpOrHttps(key: string | undefined) {
+  return key?.startsWith('https://') || key?.startsWith('http://') || false;
```
dimkl, line 10: 🙃 i would go for something like this.
```javascript
export function isHttpOrHttps([key?: string](key: string | undefined)) {
  return /^http(s)?:\/\//.test(key)
end
```

dimkl (commented):
🙃 Could we at least add a test in interstitial html page :
when proxyUrl is passed
then `script.src={expectedProxyUrl}` exists in result ?

packages/backend/src/util/parsePublishableKey.ts
panteliselef: @dimkl  I made this change because I wanted to maintain the consistency between this function and the same one written inside `packages/shared/src/utils/keys.ts`.

packages/nextjs/src/server/withClerkMiddleware.ts
panteliselef: @dimkl  isHttpOrHttps will return false if proxyUrl is undefined, but in this case I want  to throw the error if the proxyUrl was set but it was set with the wrong format.

Reminder:  proxyUrl is something optional

PR #836, opened 2023-02-20
COR-145: Hide children on sync/link

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

 This ensures that nothing will be displayed until clerk is loaded
 If the app requires for the sync/link to be performed, clerkLoaded will be false
 and none the content of the page will not flicker and no other request will be made

 At this moment checking clerkLoaded to achieve this makes sense because
 sync/link fires always by clerk-js
 
 While for nextjs this works as expected, to make remix work I had to refactored my app as seen below, otherwise nothing was rendered in the browser and DOM was completely empty
 ```typescript
 const MyClerkApp = ClerkApp(() => {
  return (
    <>
      <Outlet />
    </>
  );
});

function App() {
  return (
    <html lang='en'>
      <head>
        <Meta />
        <Links />
      </head>
      <body>
        <MyClerkApp />
        <ScrollRestoration />
        <Scripts />
        <LiveReload />
      </body>
    </html>
  );
}

export default App;

 ```

### DIFF
--- a/packages/react/src/contexts/ClerkContextProvider.tsx
+++ b/packages/react/src/contexts/ClerkContextProvider.tsx
@@ -76,7 +76,15 @@ export function ClerkContextProvider(props: ClerkContextProvider): JSX.Element |
         <SessionContext.Provider value={sessionCtx}>
           <OrganizationContext.Provider value={organizationCtx}>
             <AuthContext.Provider value={authCtx}>
-              <UserContext.Provider value={userCtx}>{children}</UserContext.Provider>
+              {/**
+               * This ensures that nothing will be displayed until clerk is loaded
+               * If the app requires for the sync/link to be performed, clerkLoaded will be false
+               * and none the content of the page will not flicker and no other request will be made
+               *
+               * At this moment checking clerkLoaded to achieve this makes sense because
+               * sync/link fires always by clerk-js
+               */}
+              <UserContext.Provider value={userCtx}>{clerkLoaded && children}</UserContext.Provider>
             </AuthContext.Provider>
           </OrganizationContext.Provider>
         </SessionContext.Provider>


packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -76,7 +76,15 @@ export function ClerkContextProvider(props: ClerkContextProvider): JSX.Element |
...
-              <UserContext.Provider value={userCtx}>{children}</UserContext.Provider>
+              {/**
+               * This ensures that nothing will be displayed until clerk is loaded
+               * If the app requires for the sync/link to be performed, clerkLoaded will be false
+               * and none the content of the page will not flicker and no other request will be made
+               *
+               * At this moment checking clerkLoaded to achieve this makes sense because
+               * sync/link fires always by clerk-js
+               */}
+              <UserContext.Provider value={userCtx}>{clerkLoaded && children}</UserContext.Provider>
```
nikosdouvlis, line 87: ❓ @panteliselef please ignore this comment if I'm mistaken, but I think this means that we won't be rendering any of the user's app at all before Clerk loads. Is that correct?

If so, this is definitely something that we should not do, as ClerkProvider in almost all scenarios wraps the whole app, blocking rendering until the `/client` and `/environment` requests resolve. 

Please confirm the above and let's discuss alternatives if this is something we should be doing  :)

panteliselef: @nikosdouvlis I know that theses changes raise some red flags.
Currently sync/link fires before `/client` and `/enviroment` request are fired

SokratisVidros:
@panteliselef Let's close this for now as we discussed offline.

PR #837, opened 2023-02-20
fix(clerk-js): Add session cookie before sync/link

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

The reasoning behind this, is to avoid throwing an extra interstitial because the session cookies was missing

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -980,8 +980,11 @@ export default class Clerk implements ClerkInterface {
   };
 
   #loadInStandardBrowser = async (): Promise<boolean> => {
+    this.#authService = new AuthenticationService(this);
     if (this.isSatellite && this.#options.shouldSyncLink) {
       if (!this.#hasSynced()) {
+        // set session cookie to avoid throwing an extra interstitial
+        this.#authService?.setAuthCookiesFromSession(this.session);
         await this.#syncWithPrimary();
         return false;
       }
@@ -990,7 +993,6 @@ export default class Clerk implements ClerkInterface {
       this.#clearSynced();
     }
 
-    this.#authService = new AuthenticationService(this);
     this.#pageLifecycle = createPageLifecycle();
 
     this.#devBrowserHandler = createDevBrowserHandler({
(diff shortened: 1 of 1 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -980,8 +980,11 @@ export default class Clerk implements ClerkInterface {
   };
 
   #loadInStandardBrowser = async (): Promise<boolean> => {
+    this.#authService = new AuthenticationService(this);
     if (this.isSatellite && this.#options.shouldSyncLink) {
       if (!this.#hasSynced()) {
+        // set session cookie to avoid throwing an extra interstitial
+        this.#authService?.setAuthCookiesFromSession(this.session);
```
nikosdouvlis, line 987: ❓ I'm certainly missing context, but do we have a `this.session` at this point? The `/client` request happens after that if I'm not mistaken.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1025,6 +1019,15 @@ export default class Clerk implements ClerkInterface {
         this.#environment = environment;
         this.updateClient(client);
 
+        if (this.isSatellite && this.#options.shouldSyncLink) {
+          if (!this.#hasSynced()) {
```
dimkl, line 1023: ☁️ If make move the code below the next line (updated 983 line)
```
if (this.isSatellite && this.#hasSynced()) {
  this.#clearSynced();
}
```
the `this.#hasSynced()` which is used will always be false even though these lines of code should run as in sync flow.
I don't think we should move it here. Let's discuss it and see where it should be.

panteliselef:
This was attempt to fix an issue with clerk-js originated from interstitial and how interstitial works currently. It ended up to be a failed experiment and that's the reason I will close this PR

PR #838, opened 2023-02-20
fix(clerk-js,backend): Only interstitial will dictate sync/link

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This is an attempt to be less aggressive when syncing with primary. We should let interstitial decide if the app should sync/link and not run this process on every page refresh

### Photos when the app has valid `__session` &`__client_uat` cookies and no interstitial is thrown
Before this PR 
Screenshot 2023-02-20 at 5 39 33 PM

After this PR 
Screenshot 2023-02-20 at 5 33 34 PM

### Photo when the app throws interstitial
Before this PR
image

After this PR
image

### DIFF
--- a/packages/backend/src/tokens/interstitial.ts
+++ b/packages/backend/src/tokens/interstitial.ts
@@ -6,6 +6,7 @@ import { callWithRetry } from '../util/callWithRetry';
 import { isStaging } from '../util/instance';
 import { parsePublishableKey } from '../util/parsePublishableKey';
 import { joinPaths } from '../util/path';
+import { AuthErrorReason } from './authStatus';
 import { TokenVerificationError, TokenVerificationErrorAction, TokenVerificationErrorReason } from './errors';
 import type { DebugRequestSate } from './request';
 
@@ -67,7 +68,7 @@ export function loadInterstitialFromLocal(
                 try {
                     await Clerk.load({
                         isSatellite: ${isSatellite},
-                        shouldSyncLink: ${isSatellite && debugData?.reason !== 'satellite-cookie-missing'},
+                        shouldSyncLink: ${debugData?.reason === AuthErrorReason.SatelliteCookieNeedsSync},
                     });
                     if(Clerk.loaded){
                       if(window.location.href.indexOf("#") === -1){
@@ -140,7 +141,7 @@ export function buildPublicInterstitialUrl(
     url.searchParams.append('proxy_url', proxyUrl);
   }
 
-  const shouldSync = isSatellite && debugData?.reason !== 'satellite-cookie-missing';
+  const shouldSync = debugData?.reason === AuthErrorReason.SatelliteCookieNeedsSync;
   if (shouldSync) {
     url.searchParams.append('should_sync_link', 'true');
   }
--- a/packages/backend/src/tokens/interstitial.ts
+++ b/packages/backend/src/tokens/interstitial.ts
@@ -6,7 +6,6 @@ import { callWithRetry } from '../util/callWithRetry';
 import { isStaging } from '../util/instance';
 import { parsePublishableKey } from '../util/parsePublishableKey';
 import { joinPaths } from '../util/path';
-import { AuthErrorReason } from './authStatus';
 import { TokenVerificationError, TokenVerificationErrorAction, TokenVerificationErrorReason } from './errors';
 import type { DebugRequestSate } from './request';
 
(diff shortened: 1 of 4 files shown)


packages/backend/src/tokens/interstitial.ts
```diff
@@ -67,7 +68,7 @@ export function loadInterstitialFromLocal(
                 try {
                     await Clerk.load({
                         isSatellite: ${isSatellite},
-                        shouldSyncLink: ${isSatellite && debugData?.reason !== 'satellite-cookie-missing'},
+                        shouldSyncLink: ${debugData?.reason === AuthErrorReason.SatelliteCookieNeedsSync},
```
nikosdouvlis, line 71: 🔧 

The change makes sense to me! I'd just like to raise that we probably shouldn't be using `debugData` here, as this is just a bucket of data we assign to the `window.__clerk_debug` var to help us debug some edge cases.

Given that this is a "bucket of data" I expect that we'll end up breaking this check. I think it'd be better if we were more explicit about this.

panteliselef: You are correct about not using debugData in that manner. `reason` is also accessible as return value of `authenticateRequest` so I will pass it in a different way here.

@nikosdouvlis Please check my new commit in this PR where I am tackling this.

packages/backend/src/tokens/interstitial.ts
```diff
@@ -67,7 +69,7 @@ export function loadInterstitialFromLocal(
                 try {
                     await Clerk.load({
                         isSatellite: ${isSatellite},
-                        shouldSyncLink: ${isSatellite && debugData?.reason !== 'satellite-cookie-missing'},
+                        shouldSyncLink: ${shouldSyncLink},
```
SokratisVidros, line 72: 🙃 Maybe rename this to 
```suggestion
                        shouldSyncWithPrimary: ${shouldSyncWithPrimary},
```

syncLink sounds a bit weird.

SokratisVidros:
@panteliselef Reminder these changes need to be applied to the Go backend interstitial.

panteliselef:
> @panteliselef Reminder these changes need to be applied to the Go backend interstitial.

@SokratisVidros  Already there :) , unless I change the naming no more changes will be required to the Go backend interstitial

SokratisVidros (approved):

packages/nextjs/src/server/withClerkMiddleware.ts
```diff
@@ -1,5 +1,6 @@
 import type { AuthStatus, RequestState } from '@clerk/backend';
 import { constants, debugRequestState } from '@clerk/backend';
+import { AuthErrorReason } from '@clerk/backend/src/tokens/authStatus';
```
dimkl, line 3: 🔧 i think we should not use non package-level exports since this couple more those 2 packages. If we want to use this i would suggest to either export it from the package or add a property to `requestState` like `isInterstitial` to be used instead of this check (eg `requestState.shouldSyncWithPrimary`).

dimkl (commented):
I would suggest that @clerk/backend should contain the logic of `shouldSyncWithPrimary` instead of each backend (eg remix/next/fastify/...) because in cases where the customer is going to use `@clerk/backend` package with another backend, we need to inform them that in the case of satellite apps it should add the condition of should `shouldSyncWithPrimary` even though it's something internal (related to how we implement the satellite apps).

panteliselef:
> I would suggest that @clerk/backend should contain the logic of `shouldSyncWithPrimary` instead of each backend (eg remix/next/fastify/...) because in cases where the customer is going to use `@clerk/backend` package with another backend, we need to inform them that in the case of satellite apps it should add the condition of should `shouldSyncWithPrimary` even though it's something internal (related to how we implement the satellite apps).

@dimkl I see your point there. This approach could eliminate duplication across the rest of our packages. But i propose this could be solved in another PR.

packages/nextjs/src/server/withClerkMiddleware.ts
panteliselef: I will use the equivalent string representation for now

dimkl (approved):
Lets proceed with the current implementation and tackle some things in a new PR.

PR #866, opened 2023-02-24
Expect actual satellite domain and not FAPI

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

- [x]  `npm run build` runs as expected.

This PR supports an actual domain to be passed for a satellite and and not FAPI. 

Before: `CLERK_DOMAIN=clerk.satellite.dev`
After: `CLERK_DOMAIN=satellite.dev`

This changes applies for the environment variable, props, and options for wrapper functions.

We continue to pass around the domain value without the prefix and we add it when it is needed(clerk-js & clerk-react)
It will be used when constructing the src url of for clerk-js script and inside Clerk class by clerk-js

We also send the domain to BAPI without the prefix and the endpoint will handle the rest.

While this solution may not seem ideal, because we need to construct the correct url each time we want to use it, the other approach of constructing the url at top level creates more issues than it solves. Happy to talk more about it in the comments if anyone wants to know more

### DIFF
--- a/packages/backend/src/tokens/interstitial.ts
+++ b/packages/backend/src/tokens/interstitial.ts
@@ -21,11 +21,19 @@ export type LoadInterstitialOptions = {
 } & DomainOrProxyUrl;
 
 // TODO: use the same function from @clerk/shared once treeshakable
-function addClerkPrefix(str: string | undefined) {
-  if (typeof str === 'undefined') {
-    return undefined;
+export function addClerkPrefix(str: string | undefined) {
+  if (!str) {
+    return '';
   }
-  return `clerk.${str}`;
+  let regex;
+  if (str?.match(/(clerk\.)?dev$/)) {
+    regex = /(clerk\.)*(?=clerk\.)/;
+  } else {
+    regex = /clerk\./gi;
+  }
+
+  const stripped = str.replace(regex, '');
+  return `clerk.${stripped}`;
 }
 
 export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions, 'apiUrl'>) {
--- a/packages/backend/src/tokens/interstitial.ts
+++ b/packages/backend/src/tokens/interstitial.ts
@@ -82,7 +82,7 @@ export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions,
                 ${domain ? `script.setAttribute('data-clerk-domain', '${domain}');` : ''}
                 ${proxyUrl ? `script.setAttribute('data-clerk-proxy-url', '${proxyUrl}')` : ''};
                 script.async = true;
-                script.src = '${getScriptUrl(proxyUrl || frontendApi, pkgVersion)}';
+                script.src = '${getScriptUrl(proxyUrl || `clerk.${domain}` || frontendApi, pkgVersion)}';
                 script.crossOrigin = 'anonymous';
                 script.addEventListener('load', startClerk);
                 document.body.appendChild(script);
--- a/packages/backend/src/tokens/interstitial.ts
+++ b/packages/backend/src/tokens/interstitial.ts
@@ -20,6 +20,14 @@ export type LoadInterstitialOptions = {
   isSatellite?: boolean;
 } & DomainOrProxyUrl;
 
+// TODO: use the same function from @clerk/shared once treeshakable
+function addClerkPrefix(str: string | undefined) {
+  if (typeof str === 'undefined') {
+    return undefined;
+  }
+  return `clerk.${str}`;
+}
+
 export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions, 'apiUrl'>) {
   options.frontendApi = parsePublishableKey(options.publishableKey)?.frontendApi || options.frontendApi || '';
   const { debugData, frontendApi, pkgVersion, publishableKey, proxyUrl, isSatellite = false, domain } = options;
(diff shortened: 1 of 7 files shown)


raptisj (approved):
@panteliselef I'm curious a bit about the issues though

packages/backend/src/tokens/interstitial.ts
```diff
@@ -20,6 +20,14 @@ export type LoadInterstitialOptions = {
...
 };
 
+// TODO: use the same function from @clerk/shared once treeshakable
+function addClerkPrefix(str: string | undefined) {
+  if (typeof str === 'undefined') {
+    return undefined;
+  }
+  return `clerk.${str}`;
+}
+
```
dimkl, line 30: 🙃 
```typescript 
function addClerkPrefix(str: string | undefined) {
  return str && `clerk.${str}`;
}
```

(if you find this more readable, apply it also to the @clerk/shared)

panteliselef:
> @panteliselef I'm curious a bit about the issues though

@raptisj Could you clarify more ? I tested all the flows locally with the changes from this PR, I did not see anything breaking. But another pair of 👀 are always welcome. These changes should also be tested once we have a staging version.

packages/backend/src/tokens/request.ts
```diff
@@ -82,11 +82,7 @@ export type AuthenticateRequestOptions = RequiredVerifyTokenOptions &
   };
 
 export async function authenticateRequest(options: AuthenticateRequestOptions): Promise<RequestState> {
-  options.frontendApi =
-    (options as AuthenticateRequestOptionsWithExperimental).domain ||
-    parsePublishableKey(options.publishableKey)?.frontendApi ||
-    options.frontendApi ||
-    '';
+  options.frontendApi = parsePublishableKey(options.publishableKey)?.frontendApi || options.frontendApi || '';
```
dimkl, line 85: ❓ the `domain` is removed? Are we sure about that?

packages/react/src/isomorphicClerk.ts
```diff
@@ -94,7 +95,7 @@ export default class IsomorphicClerk {
     this.frontendApi = frontendApi;
     this.publishableKey = publishableKey;
     this.proxyUrl = (options as ClerkConstructorOptions | undefined)?.proxyUrl;
-    this.domain = (options as ClerkConstructorOptions | undefined)?.domain;
+    this.domain = addClerkPrefix((options as ClerkConstructorOptions).domain);
```
dimkl, line 98: ☁️ I think this change will cause to add `clerk.` twice. `this.domain` is passed as domain param to Clerk at `/packages/react/src/isomorphicClerk.ts#L133` which also prefixes the domain with `.clerk` from the change in this PR at `/packages/clerk-js/src/core/clerk.ts`.

packages/backend/src/tokens/interstitial.ts
```diff
@@ -20,6 +20,14 @@ export type LoadInterstitialOptions = {
   domain?: string;
 };
 
+// TODO: use the same function from @clerk/shared once treeshakable
+function addClerkPrefix(str: string | undefined) {
```
SokratisVidros, line 24: 🔧 This function can only return an empty string if str is not provided. That would create a simpler type.

SokratisVidros: I'd also suggest making it bullet prof and ensuring that `clerk.` is added only once via regex replacement ;). That is

```
addClerkProfix("") -> ""
addClerkProfix("foo.com") -> "clerk.foo.com"
addClerkProfix("clerk.foo.com") -> "clerk.foo.com"
```

SokratisVidros (changes requested):
@panteliselef since the domain is only used on the FAPI client, does it make sense to add the `clerk.` prefix only there and not on the top level when the parameter is passed in initializing functions?

dimkl (approved):
☁️ I would use the `addClerkPrefix` to `@clerk/backend` and `@clerk/clerk-js` and not the other packages depending on it (eg @clerk/react).
❓ Does the `domain` param refer to the **primary domain** ?

packages/backend/src/tokens/interstitial.ts
dimkl: i agree but wouldn't our use case would make the regex complicated?
eg
```
addClerkProfix("clerk.clerk.dev") -> "clerk.clerk.dev"
addClerkProfix("clerk.dev") -> "clerk.clerk.dev"
```

raptisj:
> the other approach of constructing the url at top level creates more issues than it solves. Happy to talk more about it in the comments if anyone wants to know more

I was referring to this :-)

packages/backend/src/tokens/interstitial.ts
panteliselef: @SokratisVidros  should we handle the case that @dimkl  mentioned  ?

packages/backend/src/tokens/request.ts
panteliselef: @dimkl This was intentional as we handle domain always separately and we should not expect the domain when reading frontendApi.

packages/react/src/isomorphicClerk.ts
panteliselef: Yes you are correct, preparing a fix

packages/backend/src/tokens/interstitial.ts
SokratisVidros: Yes.

panteliselef: @SokratisVidros you can see the new cases in our [tests](https://github.com/clerkinc/javascript/pull/866/commits/14cd4008ed6ba91bccfd236c450ff9b247983212#diff-1a43d1aabd7f12b0efe74b95add43652387ef9b16e70fa0ba015a63b72c78771L34)

I think we are handling everything now

SokratisVidros (approved):

gkats (approved):

PR #877, opened 2023-02-28
Include proxyUrl, isSatellite, domain in developer facing apis

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description


packages/backend/src/index.ts
```diff
@@ -9,8 +9,14 @@ export * from './tokens/jwt';
 export * from './tokens/verify';
 export { constants } from './constants';
 
+// TODO: COR-164 improve type-safety
```
SokratisVidros, line 12: ❓ Can you please elaborate on this?

packages/backend/src/tokens/factory.ts
```diff
@@ -2,14 +2,24 @@ import type { ApiClient } from '../api';
...
+      | 'secretKey'
+      | 'apiUrl'
+      | 'apiVersion'
+      | 'frontendApi'
+      | 'publishableKey'
+      | 'jwtKey'
+      | 'proxyUrl'
+      | 'domain'
+      | 'isSatellite'
+      | 'hasJustSynced'
```
SokratisVidros, line 22: ⛔ This function exposes the internal logic of the `authenticateRequest` function that is hard to explain to the end user. Instead of passing it as a boolean, we should replace it with a query parameter of type String or URLSearchParams that the `authenticateRequest` function will receive, do the necessary computations without exposing any of that to its public API.

SokratisVidros (changes requested):

packages/backend/src/index.ts
panteliselef: @SokratisVidros I am thinking of using the type `DomainOrProxyUrl` wherever possible

```typescript
export type DomainOrProxyUrl = {
    proxyUrl?: never;
    domain: string;
} | {
    proxyUrl: string;
    domain?: never;
};
```

SokratisVidros (approved):

PR #879, opened 2023-02-28
COR-110 Reduce ClerkJs bundle size 

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### React.lazy() to the rescue

This PR is the first attempt to decrease the bundle size of our clerk-js bundle. 
By using React.lazy and React.Suspense we are now  lazy load our UI components when they are mounted on a users application. 

⚠️ Further optimisation can be introduced to each lazy loaded UI component to incrementally load only the parts the user needs or visits.

We also lazy load some of the providers that wrap our UI components. This was a necessary step the allowed us to decrease the bundle size even more. These providers are only used by the UI components so it makes sense that if a UI component is not loaded, we do not need to load the providers as well. For now we lazy load some of them eagerly, but this is an optimisation that could be done in a later step.

This PR decreases the bundle size of `clerk.browser.js` from ~187KB down to ~55KB. This is a remarkable **70% decrease** 🎉

In this PR we also addressed
- Hydration issues related to lazy loading
- Generating unique hashes for the webpack chunk

### Hydration issues related to lazy loading
We stumbled upon an interesting hydration issue when our UI modal components where about to be mounted. We solved that by adding an additional <Suspense> and wrapping those components directly without other lazy load component in between.

### Generating unique hashes for the webpack chunk

We come up with the following naming conventions
- For common chunks: shared-[id].[hash].[version].js
- For top level chunks : [name].[hash].[version].js

`name` name of the module or the given name of webpackChunkName name
`hash` will consist of 6 characters unique for each build & will be the same of all chunks generated in a particular build
`version`  is the readable (human friendly) representation of the clerk-js package

image


packages/clerk-js/src/ui/lazyModules/LazyImpersonationFAB.tsx
```diff
@@ -0,0 +1,30 @@
+import type { Appearance } from '@clerk/types';
+import { Suspense } from 'react';
+
+import { ImpersonationFab } from './../components/ImpersonationFab';
+import { AppearanceProvider } from './commonDeps';
+
+console.log('top level fab');
```
panteliselef, line 7: 🙃 remove logs

nikosdouvlis (approved):
:D

desiprisg:
:O

PR #881, opened 2023-02-28
feat(backend): Throw error if the secret key is missing

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This PR validates the presence of the `secretKey` in the backend package.

### DIFF
--- a/packages/backend/src/api/request.ts
+++ b/packages/backend/src/api/request.ts
@@ -72,7 +72,6 @@ export function buildRequest(options: CreateBackendApiOptions) {
       httpOptions = {},
     } = options;
     const { path, method, queryParams, headerParams, bodyParams } = requestOptions;
-
     const key = secretKey || apiKey;
 
     assertValidSecretKey(key);
--- a/packages/backend/src/api/request.ts
+++ b/packages/backend/src/api/request.ts
@@ -6,8 +6,8 @@ import { API_URL, API_VERSION, USER_AGENT } from '../constants';
 // DO NOT CHANGE: Runtime needs to be imported as a default export so that we can stub its dependencies with Sinon.js
 // For more information refer to https://sinonjs.org/how-to/stub-dependency/
 import runtime from '../runtime';
+import { assertValidSecretKey } from '../util/assertValidSecretKey';
 import { joinPaths } from '../util/path';
-import { validateSecretKey } from '../util/validateSecretKey';
 import type { CreateBackendApiOptions } from './factory';
 import { deserialize } from './resources/Deserializer';
 
@@ -73,9 +73,9 @@ export function buildRequest(options: CreateBackendApiOptions) {
     } = options;
     const { path, method, queryParams, headerParams, bodyParams } = requestOptions;
 
-    const key = secretKey || apiKey || '';
+    const key = secretKey || apiKey;
 
-    validateSecretKey(key);
+    assertValidSecretKey(key);
 
     const url = joinPaths(apiUrl, apiVersion, path);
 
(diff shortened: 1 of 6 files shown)


packages/backend/src/api/request.ts
```diff
@@ -72,13 +73,9 @@ export function buildRequest(options: CreateBackendApiOptions) {
     } = options;
     const { path, method, queryParams, headerParams, bodyParams } = requestOptions;
 
-    const key = secretKey || apiKey;
+    const key = secretKey || apiKey || '';
```
nikosdouvlis, line 76: Why do we need to fallback to ''?

packages/backend/src/util/validateSecretKey.ts
```diff
@@ -0,0 +1,9 @@
+export const validateSecretKey = (key: string) => {
```
nikosdouvlis, line 1: Let's make this an [assertion function ](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions)

```suggestion
export function assertValidSecretKey(val: unknown): asserts val is string {
    if(!val || typeof val !== "string") {
        throw ...
    }
}
```

packages/backend/src/api/request.ts
desiprisg: We can omit it if the validation function can receive undefined.

packages/backend/src/util/validateSecretKey.ts
desiprisg: Done!

dimkl (approved):

packages/backend/src/index.ts
```diff
@@ -19,6 +20,7 @@ export type ClerkOptions = CreateBackendApiOptions &
 
 export function Clerk(options: ClerkOptions) {
   const opts = { ...options };
+  assertValidSecretKey(opts.secretKey || opts.apiKey);
```
dimkl, line 23: ⛔ this check should not exist since the `secretKey` could be passed as runtime parameter in authenticateRequest.  If we add this call here, it will fail in `Cleck` instance inititialization.

dimkl: Based on the above comment (if we remove this line) the functionality of the code will not change. Let's add a warning in the `assertValidSecretKey` to have a kind of indication in the logs of our application (nextJS, remix) and see how we can tackle it after a discussion.

nikosdouvlis (approved):

dimkl:
Some extra context: Currently the users are getting a `signed-out` state in case the secretKey & apiKey are missing and there is log about the missing secretKey. After this PR the user will get a `Missing Clerk Secret Key or ...` error upon each request!
The assertion will be executed per request instead of module initialization due to the use case of passing keys in runtime.
cc: @nikosdouvlis , @SokratisVidros

dimkl (approved):
🔧 please revert the changes to `playground/nextjs/package-lock.json`.

PR #882, opened 2023-03-01
fix(clerk-js): Remove sleep when clicking manage account in navigation mode

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Remove the 1 second sleep when clicking the manage account button in the user button popover. This seems to cause the popover to remain open when the navigation has completed.

Can't seem to figure out why this was added there. Did the popover close too fast when not in a SPA? Let me know you see an issue with this fix so we can tailor it better if possible.


packages/clerk-js/src/ui/components/UserButton/useMultisessionActions.tsx
```diff
@@ -35,8 +35,7 @@ export const useMultisessionActions = (opts: UseMultisessionActionsParams) => {
 
   const handleManageAccountClicked: React.MouseEventHandler<HTMLButtonElement> = () => {
     if (opts.userProfileMode === 'navigation') {
-      return navigate(opts.userProfileUrl).finally(async () => {
-        await sleep(1000);
+      return navigate(opts.userProfileUrl).finally(() => {
```
nikosdouvlis, line 38: We're waiting for 1sec for any *external* navigation to happen before closing the popup, as it's rather hard to consistently wait until a browser-navigation takes place, especially across different browsers. 

So we have 2 scenarios we need to handle here: 
- there's an external navigation and the browser will redirect from the host app to accounts (full page reload): we need to show the spinner while the browser is loading and until the page unmounts
- there's an client-side only navigation (eg:  from /home to /user): this we can await so there's no need for the sleep

desiprisg: Any ideas on how to check which case we're on?

packages/clerk-js/src/ui/components/UserButton/useMultisessionActions.tsx
```diff
@@ -33,10 +33,10 @@ export const useMultisessionActions = (opts: UseMultisessionActionsParams) => {
     );
   };
 
-  const handleManageAccountClicked: React.MouseEventHandler<HTMLButtonElement> = () => {
+  const handleManageAccountClicked = () => {
     if (opts.userProfileMode === 'navigation') {
       return navigate(opts.userProfileUrl).finally(async () => {
-        await sleep(1000);
+        await sleep(300);
```
anagstef, line 39: ❓ So, the final change is to reduce this sleep timer? Why?

nikosdouvlis (approved):

packages/clerk-js/src/ui/components/UserButton/useMultisessionActions.tsx
nikosdouvlis: Checking whether this is a external or internal navigation is too complicated for this simple scenario. We might revise in the future, but this should be enough for now

PR #890, opened 2023-03-03
fix(clerk-js): Correctly truncate the hash after redirectWithAuth

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

I've opted for a different approach with this after some discussion. We should not use the URL api to avoid uncertainty and always preserve the exact same url that was passed to us. I've added 2 separators that we will use to detect the jwt. One is used when there was not hash present in the passed url, and the other in the opposite case. That way, we can know whether or not we need to remove the hash upon load.


desiprisg:
Are we ok with removing the fragment here? If so, let me know so I can go ahead and update the tests!

nikosdouvlis:
@desiprisg Can you share more details about this one please? Not sure I get the solution!

dimkl (approved):

SokratisVidros (changes requested):
The `dev_browser_jwt` was designed to pass from the hash instead of query string on purpose as it doesn't travel on the server logs. Let me find the relevant design doc to provide more context.

dimkl (approved):

desiprisg:
@SokratisVidros I made the changes to keep the hash. Please take a look.

packages/clerk-js/src/utils/url.ts
```diff
@@ -280,18 +280,20 @@ export function setSearchParameterInHash({
...
   const h = hash.startsWith('#') ? hash.substring(1) : hash;
   const dummyUrlForHash = new URL(h, DUMMY_URL_BASE);
   dummyUrlForHash.searchParams.delete(paramName);
 
-  // The following line will prepend the hash with a `/`.
+  // The following line will prepend the hash with a `/` if removeTrailingSlash is false(default).
   // This is required for ClerkJS Components Hash router to work as expected
   // as it treats the hash as sub-path with its nested querystring parameters.
-  return dummyUrlForHash.href.replace(DUMMY_URL_BASE, '');
+  return dummyUrlForHash.href.replace(`${DUMMY_URL_BASE}${removeTrailingSlash ? '/' : ''}`, '');
```
anagstef, line 296: ❓ Isn't this a little confusing? The variable name is `removeTrailingSlash`, but when it is truthy we add a trailing slash.

desiprisg: I thought I removed this. My bad. This is not a requirement any more.
For the record though, we add the trailing slash to the string to be replaced by `''`, so it's essentially removed.

desiprisg: This function will continue as it was before.

packages/clerk-js/src/core/devBrowserHandler.ts
```diff
@@ -107,17 +106,19 @@ export default function createDevBrowserHandler({
...
+    }
+    const devBrowserToken = arr.length > 1 ? arr[1] : undefined;
 
     if (devBrowserToken) {
-      window.location.hash = removeSearchParameterFromHash({
-        hash: window.location.hash,
-        paramName: DEV_BROWSER_SSO_JWT_PARAMETER,
-      });
-
+      window.location.hash = window.location.hash.split(DEV_BROWSER_SSO_JWT_PARAMETER_SEPARATOR)[0];
```
anagstef, line 118: ❓ Is there any cleaner way to write this? Accessing array positions directly feels a little confusing.

SokratisVidros:
@desiprisg regarding this fix, wouldn't it make sense just to remove [this line](https://github.com/clerkinc/javascript/blob/3d4fe82452e196207cc5c4285ade2ca37c71174c/packages/clerk-js/src/utils/url.ts#L274)?

desiprisg:
@SokratisVidros This function did not change, so why would we remove the comment there?

packages/clerk-js/src/core/devBrowserHandler.ts
desiprisg: We're just getting the first part of the hash without the JWT in this case. I think it's a little overkill to make a util for this though.

SokratisVidros (approved):

PR #911, opened 2023-03-09
COR-152 fix(clerk-js): Read country code from FAPI response

raptisj:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Read the country code from FAPI response and make it the default value. This is a change that will only affect newcomers.

### DIFF
--- a/packages/clerk-js/src/core/fapiClient.ts
+++ b/packages/clerk-js/src/core/fapiClient.ts
@@ -197,8 +197,15 @@ export default function createFapiClient(clerkInstance: Omit<Clerk, 'proxyUrl'>)
     } catch (e) {
       clerkNetworkError(urlStr, e);
     }
+
+    const responseHeaders = {
+      country: response.headers.get('x-country'),
+    };
+
     const json: FapiResponseJSON<T> = await response.json();
-    const fapiResponse: FapiResponse<T> = Object.assign(response, { payload: json });
+    const fapiResponse: FapiResponse<T> = Object.assign(response, {
+      payload: { ...json, response_headers: responseHeaders },
+    });
     await runAfterResponseCallbacks(requestInit, fapiResponse);
     return fapiResponse;
   }
--- a/packages/clerk-js/src/core/fapiClient.ts
+++ b/packages/clerk-js/src/core/fapiClient.ts
@@ -2,6 +2,7 @@ import { camelToSnake } from '@clerk/shared';
 import type { Clerk, ClerkAPIErrorJSON, ClientJSON } from '@clerk/types';
 import qs from 'qs';
 
+import type { CountryIso } from '../ui/elements/PhoneInput/countryCodeData';
 import { buildEmailAddress as buildEmailAddressUtil, buildURL as buildUrlUtil } from '../utils';
 import { clerkNetworkError } from './errors';
 
@@ -43,6 +44,7 @@ export interface FapiResponseJSON<T> {
   meta?: {
     client?: ClientJSON;
     session_id?: string;
+    responseHeaders?: { country: CountryIso } | unknown;
   };
 }
 
@@ -204,7 +206,7 @@ export default function createFapiClient(clerkInstance: Omit<Clerk, 'proxyUrl'>)
 
     const json: FapiResponseJSON<T> = await response.json();
     const fapiResponse: FapiResponse<T> = Object.assign(response, {
-      payload: { ...json, response_headers: responseHeaders },
+      payload: { ...json, meta: { ...json?.meta, responseHeaders } },
     });
     await runAfterResponseCallbacks(requestInit, fapiResponse);
     return fapiResponse;
--- a/packages/clerk-js/src/core/resources/Environment.ts
+++ b/packages/clerk-js/src/core/resources/Environment.ts
@@ -7,6 +7,7 @@ import type {
   UserSettingsResource,
 } from '@clerk/types';
 
+import type { CountryIso } from '../../ui/elements/PhoneInput/countryCodeData';
 import { AuthConfig, BaseResource, DisplayConfig, UserSettings } from './internal';
 import { OrganizationSettings } from './OrganizationSettings';
 
@@ -18,7 +19,7 @@ export class Environment extends BaseResource implements EnvironmentResource {
   displayConfig!: DisplayConfigResource;
   userSettings!: UserSettingsResource;
   organizationSettings!: OrganizationSettingsResource;
-  responseHeaders: { country: string };
+  country!: CountryIso;
 
   public static getInstance(): Environment {
     if (!Environment.instance) {
@@ -59,7 +60,7 @@ export class Environment extends BaseResource implements EnvironmentResource {
   protected fromJSON(data: EnvironmentJSON | null): this {
     if (data) {
       this.authConfig = new AuthConfig(data.auth_config);
-      this.responseHeaders = data.response_headers;
+      this.country = data.meta.responseHeaders.country;
       this.displayConfig = new DisplayConfig(data.display_config);
       this.userSettings = new UserSettings(data.user_settings);
       this.organizationSettings = new OrganizationSettings(data.organization_settings);
--- a/packages/clerk-js/src/core/resources/Environment.ts
+++ b/packages/clerk-js/src/core/resources/Environment.ts
@@ -18,6 +18,7 @@ export class Environment extends BaseResource implements EnvironmentResource {
   displayConfig!: DisplayConfigResource;
   userSettings!: UserSettingsResource;
   organizationSettings!: OrganizationSettingsResource;
+  responseHeaders: { country: string };
 
   public static getInstance(): Environment {
     if (!Environment.instance) {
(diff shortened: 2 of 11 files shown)


dimkl (approved):

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
```diff
@@ -20,14 +20,15 @@ const createSelectOption = (country: CountryEntry) => {
 
 const countryOptions = [...IsoToCountryMap.values()].map(createSelectOption);
 
-type PhoneInputProps = PropsOfComponent<typeof Input>;
+type PhoneInputProps = PropsOfComponent<typeof Input> & { defaultSelectedIso?: CountryIso };
```
panteliselef, line 23: 🙃 This is extreme nitpicking, you can ignore it
```suggestion
type PhoneInputProps = PropsOfComponent<typeof Input> & { defaultSelectedISO?: CountryIso };
```

packages/clerk-js/src/ui/elements/PhoneInput/useFormattedPhoneNumber.ts
```diff
@@ -16,7 +16,10 @@ const format = (str: string, iso: CountryIso) => {
 };
 
 export const useFormattedPhoneNumber = (props: UseFormattedPhoneNumberProps = {}) => {
-  const [selectedIso, setSelectedIso] = useLocalStorage<CountryIso>('selectedCountryIso', 'us');
+  const [selectedIso, setSelectedIso] = useLocalStorage<CountryIso>(
+    'selectedCountryIso',
```
panteliselef, line 20: ❓ Should we store these keys in constants ?

panteliselef (approved):

desiprisg (approved):

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
nikosdouvlis: Even more nitpicky (sorry for this hehe) I think that at some point we landed on using camelCase for abbreviations as the newer w3c standard/airbnb guide suggests, eg:

✅  `redirectUrl` 
❌  `redirectURL`

packages/clerk-js/src/core/fapiClient.ts
```diff
@@ -197,8 +197,15 @@ export default function createFapiClient(clerkInstance: Omit<Clerk, 'proxyUrl'>)
...
     }
+
+    const responseHeaders = {
+      country: response.headers.get('x-country'),
+    };
+
     const json: FapiResponseJSON<T> = await response.json();
-    const fapiResponse: FapiResponse<T> = Object.assign(response, { payload: json });
+    const fapiResponse: FapiResponse<T> = Object.assign(response, {
+      payload: { ...json, response_headers: responseHeaders },
```
nikosdouvlis, line 207: It looks like we're adding the `response_headers` for all requests.
If so, we need to also update the `FapiResponse` and `FapiResponse` to reflect this.

nikosdouvlis: On a second thought, I see that `FapiResponseJSON` includes an optional `meta` key: https://github.com/clerkinc/javascript/blob/ec2262459d34f1a443d272dfa405d117d97251ae/packages/clerk-js/src/core/fapiClient.ts#L43

Can we maybe reuse that instead of adding the `response_headers` object? Eg `payload.meta.country`

packages/clerk-js/src/ui/elements/PhoneInput/useFormattedPhoneNumber.ts
```diff
@@ -5,7 +5,7 @@ import { extractDigits, formatPhoneNumber } from '../../utils';
 import type { CountryIso } from './countryCodeData';
 import { IsoToCountryMap } from './countryCodeData';
 
-type UseFormattedPhoneNumberProps = { defaultPhone?: string };
+type UseFormattedPhoneNumberProps = { defaultPhone?: string; defaultSelectedISO?: CountryIso };
```
nikosdouvlis, line 8: We use `ISO` instead of  `Iso` as we do everywhere else?  Lets still with camelCase for abbreviations :)

packages/clerk-js/src/core/fapiClient.ts
```diff
@@ -197,8 +197,15 @@ export default function createFapiClient(clerkInstance: Omit<Clerk, 'proxyUrl'>)
...
     }
+
+    const responseHeaders = {
+      country: response.headers.get('x-country'),
+    };
+
     const json: FapiResponseJSON<T> = await response.json();
-    const fapiResponse: FapiResponse<T> = Object.assign(response, { payload: json });
+    const fapiResponse: FapiResponse<T> = Object.assign(response, {
+      payload: { ...json, meta: { ...json?.meta, responseHeaders } },
```
nikosdouvlis, line 207: Why do we need `...json?.meta` instead of just  `...json.meta`?

packages/clerk-js/src/core/resources/Environment.ts
```diff
@@ -58,6 +60,7 @@ export class Environment extends BaseResource implements EnvironmentResource {
   protected fromJSON(data: EnvironmentJSON | null): this {
     if (data) {
       this.authConfig = new AuthConfig(data.auth_config);
+      this.meta = data.meta;
```
nikosdouvlis, line 63: Lets keep this as it was before - env does not need to know about everything in meta, just the country
```suggestion
      this.country = data.meta.headers.country
```

packages/types/src/json.ts
```diff
@@ -52,11 +52,18 @@ export interface ImageJSON {
...
+export interface ResponseHeaderOptionsJSON {
+  responseHeaders: { country: any };
+}
+
 export interface EnvironmentJSON extends ClerkResourceJSON {
   auth_config: AuthConfigJSON;
   display_config: DisplayConfigJSON;
   user_settings: UserSettingsJSON;
   organization_settings: OrganizationSettingsJSON;
+  meta: ResponseHeaderOptionsJSON;
```
nikosdouvlis, line 66: Lets revert it back to a country: string as we had before

packages/clerk-js/src/core/fapiClient.ts
nikosdouvlis: Also, the FapiResponseJSON needs to be updated with the responseHeaders key:

```
export interface FapiResponseJSON<T> {
  response: T;
  client?: ClientJSON;
  errors?: ClerkAPIErrorJSON[];
  meta?: {
    client?: ClientJSON;
    session_id?: string;
    responseHeaders?: { country: string }
  };
}
```

packages/clerk-js/src/ui/elements/PhoneInput/useFormattedPhoneNumber.ts
raptisj: @nikosdouvlis ok I'm switching from `defaultSelectedISO` to `defaultSelectedIso `

packages/clerk-js/src/core/fapiClient.ts
raptisj: @nikosdouvlis this is purely for safety. Do you think it is not needed?

packages/types/src/json.ts
```diff
@@ -52,11 +52,18 @@ export interface ImageJSON {
   public_url: string;
 }
 
+// move CountryIso types from clerkjs to this repo. Hence the any type in country
+// packages/clerk-js/src/ui/elements/PhoneInput/countryCodeData.ts
+export interface ResponseHeaderOptionsJSON {
```
nikosdouvlis, line 57: I dont think we still need this

packages/types/src/json.ts
nikosdouvlis, line 66: This needs to be country, right?

raptisj: @nikosdouvlis we need this for because of this `data: EnvironmentJSON `

nikosdouvlis: I see..
not a fan of this as EnvironmentJSON does not need to know of `meta` but lets revise when w need to read headers for a second place

nikosdouvlis (approved):

packages/types/src/json.ts
SokratisVidros: An alternative solution is to use a `Clerk.setCountry` function similar to `Clerk.updateClient` that is specific to this functionality until we have another use case and need to read from FAPI headers. Having said that, I'd expect the Env JSON to be responsible for this only if the country was returned by FAPI as part of the environment payload.

@raptisj let's add a story for this week to revise this please.

PR #912, opened 2023-03-09
[JS-209] Report all FAPI network errors only for valid online browsers

anagstef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

In this PR we try to reduce the errors that bubble up to user's apps by retrying more times on every `GET` request when it fails due to network issues and also show the error as a console warn if the browser is offline at the time of throwing the error.


packages/clerk-js/src/core/clerk.ts
```diff
@@ -1063,7 +1063,7 @@ export default class Clerk implements ClerkInterface {
           await this.#devBrowserHandler.setup();
         } else if (!isValidBrowserOnline()) {
           console.warn(err);
-          break;
+          return false;
```
anagstef, line 1066: ℹ️ This is to return `false` when `loadInStandardBrowser` fails, so it reflects that the clerk-js has not loaded yet.

packages/clerk-js/src/core/fapiClient.ts
```diff
@@ -184,16 +184,26 @@ export default function createFapiClient(clerkInstance: Omit<Clerk, 'proxyUrl'>)
...
-        : // Mock an empty json response
-          new Response('{}', requestInit);
+      if (beforeRequestCallbacksResult) {
+        response =
+          // retry only on GET requests for safety
+          overwrittenRequestMethod === 'GET'
+            ? await runWithExponentialBackOff(() => fetch(urlStr, fetchOpts), {
+                maxRetries: 6,
+                firstDelay: 500,
+                maxDelay: 5000,
```
anagstef, line 201: ℹ️  This config results in the following timeline:

**0sec -- 0,5sec -- 1,5sec -- 3,5sec -- 7,5sec -- 15,5sec -- THROW.**

packages/clerk-js/src/core/resources/BackupCode.ts
```diff
@@ -16,11 +16,13 @@ export class BackupCode extends BaseResource implements BackupCodeResource {
...
-    this.codes = data.codes;
-    this.updatedAt = unixEpochToDate(data.updated_at);
-    this.createdAt = unixEpochToDate(data.created_at);
+  protected fromJSON(data: BackupCodeJSON | null): this {
+    if (data) {
+      this.id = data.id;
+      this.codes = data.codes;
+      this.updatedAt = unixEpochToDate(data.updated_at);
+      this.createdAt = unixEpochToDate(data.created_at);
+    }
```
anagstef, line 25: ℹ️ Refactored all the resources to handle `null` instead of a data object in their `fromJSON`

packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -34,15 +34,17 @@ export abstract class BaseResource {
...
-        maxRetries: 4,
-        firstDelay: 500,
-      });
-    } else {
+    try {
       fapiResponse = await BaseResource.fapiClient.request<J>(requestInit);
+    } catch (e) {
+      if (!isValidBrowserOnline()) {
+        console.warn(e);
+        return null;
```
anagstef, line 42: ℹ️ Mute all network errors if the browser is not online

packages/clerk-js/src/core/resources/DisplayConfig.ts
```diff
@@ -35,33 +35,35 @@ export class DisplayConfig extends BaseResource implements DisplayConfigResource
...
-    this.branded = data.branded;
-    this.supportEmail = data.support_email || '';
-    this.clerkJSVersion = data.clerk_js_version;
-    this.experimental__forceOauthFirst = data.experimental_force_oauth_first || false;
-    this.organizationProfileUrl = data.organization_profile_url;
-    this.createOrganizationUrl = data.create_organization_url;
-    this.afterLeaveOrganizationUrl = data.after_leave_organization_url;
-    this.afterCreateOrganizationUrl = data.after_create_organization_url;
+  protected fromJSON(data: DisplayConfigJSON | null): this {
+    if (data) {
```
panteliselef, line 39: 🙃 to avoid the indentation
```suggestion
    if (!data) {return this}
```

packages/clerk-js/src/core/fapiClient.ts
```diff
@@ -184,16 +184,26 @@ export default function createFapiClient(clerkInstance: Omit<Clerk, 'proxyUrl'>)
...
+              })
+            : await fetch(urlStr, fetchOpts);
+      } else {
+        response = new Response('{}', requestInit); // Mock an empty json response
```
dimkl, line 205: ❓ is `Response` supported in react native? We should check our `expo` package.

packages/clerk-js/src/core/resources/BackupCode.ts
dimkl: ❓ why? Do we call the `fromJSON()` when there is no data?

dimkl (approved):

panteliselef:
> ❓ why? Do we call the fromJSON() when there is no data?
@dimkl  in order not to refactor [this] 🤔(https://github.com/clerkinc/javascript/blob/4a1ce9a2bf64e9e724098e69b108e15763057839/packages/backend/src/api/resources/Deserializer.ts#L46) ?

dimkl:
> > ❓ why? Do we call the fromJSON() when there is no data?
> > @dimkl  in order not to refactor [this] 🤔(https://github.com/clerkinc/javascript/blob/4a1ce9a2bf64e9e724098e69b108e15763057839/packages/backend/src/api/resources/Deserializer.ts#L46
> > ) ?

With the current implementation there is no need. Why does the current PR changes it?
Also the ref you provided is in `backend` where the current PR refers to changes in FAPI network calls which are only related with FE code `clerk-js`.

panteliselef:
you are correct, nvm

packages/clerk-js/src/core/resources/BackupCode.ts
anagstef: @dimkl This is a valid question. The default mechanism when calling requests in `BaseResource` (and in extend to all the other resources) is to call `fromJSON()` after getting the response to update the resource data. 

Currently, if the `fetch` throws an error, the execution thread stops, and the error reaches the browser.

After this PR, if `fetch` fails, in some cases we mute the error and return `null`. Maybe we want to catch it earlier. Open to suggestions. :)

packages/types/src/deletedObject.ts
```diff
@@ -1,6 +1,4 @@
-import type { ClerkResource } from './resource';
-
-export interface DeletedObjectResource extends ClerkResource {
+export interface DeletedObjectResource {
```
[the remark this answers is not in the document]
anagstef: @dimkl This was added by me in a commit some days ago, and TypeScript is not happy about it because of how this implements some of the properties. I thought `DeletedObjectResource` might be a special resource, so I reverted to what it was before.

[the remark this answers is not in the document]
dimkl: ok. i think we need note and investigate about what happens in resources since we have to enforce that all of them somehow extend a base resource interface (comment unrelated to this PR).

dimkl (commented):
There are some requests where the retry mechanism will not be triggered.
eg https://github.com/clerkinc/javascript/blob/dab58e970a1fed3cf486b33e25bc9f3c3dc531a4/packages/clerk-js/src/core/resources/SessionWithActivities.ts#L42

anagstef:
> There are some requests where the retry mechanism will not be triggered. eg
> 
> https://github.com/clerkinc/javascript/blob/dab58e970a1fed3cf486b33e25bc9f3c3dc531a4/packages/clerk-js/src/core/resources/SessionWithActivities.ts#L42

@dimkl The retry mechanism is added inside the `request` method of the FAPI client, so it's triggered in the above example.

packages/clerk-js/src/core/resources/Organization.ts
```diff
@@ -155,16 +155,18 @@ export class Organization extends BaseResource implements OrganizationResource {
...
-    this.name = data.name;
-    this.slug = data.slug;
-    this.logoUrl = data.logo_url;
-    this.publicMetadata = data.public_metadata;
-    this.membersCount = data.members_count;
-    this.pendingInvitationsCount = data.pending_invitations_count;
-    this.createdAt = unixEpochToDate(data.created_at);
-    this.updatedAt = unixEpochToDate(data.updated_at);
+  protected fromJSON(data: OrganizationJSON | null): this {
+    if (data) {
```
nikosdouvlis, line 159: **This is an extremely nitpicky comment and there's absolutely no need to address it**
I usually like to reverse if statements like that and use them as guard expressions, eg:

```tsx
if (!data) {
 return this;
} 
...
```

as it makes the edge case stand out more

packages/clerk-js/src/core/resources/OrganizationSettings.ts
```diff
@@ -11,10 +11,9 @@ export class OrganizationSettings extends BaseResource implements OrganizationSe
     this.fromJSON(data);
   }
 
-  protected fromJSON(data: OrganizationSettingsJSON): this {
-    data = data || {};
-    this.enabled = data.enabled || false;
-    this.maxAllowedMemberships = data.max_allowed_memberships || 0;
+  protected fromJSON(data: OrganizationSettingsJSON | null): this {
```
nikosdouvlis, line 14: This looks to be handled differently than the rest. Since it looks like we need some default values, we should probably make it more explicit:
```tsx
data = data || { enabled: false,  maxAllowedMemberships: 0 } 
this.enabled = data.enabled;
this.maxAllowedMemberships = data.max_allowed_memberships;
```

packages/clerk-js/src/core/resources/BackupCode.ts
nikosdouvlis: So `null` means that no change will be made to the state of the app?

PR #929, opened 2023-03-13
Accept string or function for `domain` prop

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [x] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### Before (remix)
```typescript
export const loader: LoaderFunction = (args) => {
  return rootAuthLoader(
    args,
    {
      loadUser: true,
      isSatellite: true,
      domain: 'satellite.dev',
    } 
  );
};
```
### After (remix)
```typescript
export const loader: LoaderFunction = (args) => {
  return rootAuthLoader(
    args,
    {
      loadUser: true,
      isSatellite: true,
      domain: (url) => url.host,
    } as const
  );
};
```

### Before (nextjs middleware)
```typescript
export default withClerkMiddleware(... ,{
    isSatellite: true,
    domain: 'satellite.dev'
});
```
### After (nextjs middleware)
```typescript
export default withClerkMiddleware(... ,{
    isSatellite: true,
    domain:(url) => url.host
});
```

### Before (nextjs ClerkProvider)
```jsx
<ClerkProvider
  isSatellite={false}
  domain={'satellite.dev'}
>
{...}
</ClerkProvider>
```

### After (nextjs ClerkProvider)
```jsx
<ClerkProvider
  isSatellite={false}
  domain={url => url.host}
>
{...}
</ClerkProvider>
```


packages/nextjs/src/server/utils.ts
```diff
@@ -143,14 +143,22 @@ export const injectSSRStateIntoObject = <O, T>(obj: O, authObject: T) => {
...
 };
 
+// TODO: Use the same function defined in @clerk/shared once the package is tree shakeable
+export const handleDomainStringOrFn = (opts: WithAuthOptions, url: URL): string => {
+  if (typeof opts.domain === 'function') {
+    return opts.domain(url);
+  }
+  return opts.domain || '';
+};
+
```
dimkl, line 161: ❓ what do you think about using a generic function for both domain and isSatellite ?
```
type Fn<T> = (v: any) => T;
export function handleValueOrFn<T>(value: T | Fn<T>, url: URL, defaultValue: T): T {
    if (typeof value === 'function') {
        return value(url);
    }
    return value || defaultValue;
}

var request = { url: '' };
var opts = { isSatellite: true };
handleValueOrFn<boolean>(opts.isSatellite, new URL(request.url), false);

var request = { url: '' };
handleValueOrFn<boolean>((url: URL) => !!url, new URL(request.url), false);


var request = { url: '' };
var opts2 = { domain: 'true' };
handleValueOrFn<string>(opts2.domain, new URL(request.url), '');

var request = { url: '' };
handleValueOrFn<string>((url: URL) => url.host, new URL(request.url), '');
```

packages/react/src/isomorphicClerk.ts
```diff
@@ -88,12 +89,19 @@ export default class IsomorphicClerk {
     return this.#instance;
   }
 
+  get domain(): string {
+    if (typeof this.#domain === 'function') {
+      return this.#domain(new URL(window.location.href));
+    }
+    return this.#domain || '';
+  }
+
```
dimkl, line 98: ❓ shouldn't there be an `isSatellite()` method like in `core/clerk.ts` ?

packages/types/src/clerk.ts
```diff
@@ -73,7 +73,7 @@ export interface Clerk {
    * @experimental
    * Clerk Satellite Frontend API string.
    */
-  domain?: string;
+  domain?: string | ((url: URL) => string);
```
dimkl, line 76: ❓ shouldn't there be an `isSatellite?` here ?

packages/types/src/multiDomain.ts
```diff
@@ -7,19 +7,46 @@ import type { ClerkOptions } from './clerk';
...
+      proxyUrl?: never | string;
+      domain?: never;
+    }
+  | {
+      isSatellite: Exclude<ClerkOptions['isSatellite'], undefined>;
+      proxyUrl?: never;
+      domain: Exclude<ClerkInterface['domain'], undefined>;
+    }
+  | {
+      isSatellite: Exclude<ClerkOptions['isSatellite'], undefined>;
```
dimkl, line 22: ❓ Wouldn't it be more readable to move the `isSatellite` from `ClerkOptions` to here  and avoid `Exclude`  ? Also semantically if `isSatellite` only used as part of multi-domain feature i would expect to find it as part of the `MultiDomainAndOrProxy` type.
Also i would rename the type from `MultiDomainAndOrProxy` to `MultiDomainAndProxy`. WDYT ?

dimkl: Could you explain what are the validations of this type? What are the combination 😓 ?

dimkl (approved):

packages/types/src/clerk.ts
panteliselef: @dimkl I think the question should be, do we still need domain there 🤔?

packages/types/src/multiDomain.ts
panteliselef: Yes, I've listed all valid and invalid combos in a [previous PR](https://github.com/clerkinc/javascript/pull/901)

panteliselef: Regarding your 1st comment

isSatellite lives inside ClerkOptions because it is used by `new Clerk().load()`

The patterns that existed before ( and I followed ) is: Pass to the Clerk Class construct only the information necessary to create the instance (e.g createFapiClient)

And pass everything else through `load` as `ClerkOptions`

We can for sure refactor this. I'm just laying out my reasoning behind this

packages/nextjs/src/server/utils.ts
panteliselef: For sure we can 😀, this was my thought as well but I decided to delay it. Glad that you brought this up.

PR #941, opened 2023-03-16
fix(clerk-js): Removed unsupported props when inviting org members

alex-ntousias:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This commit removes removes the unused `redirectUrl` property from the organization member invitation.

This property is supported when creating organization invitations from BAPI, but not from FAPI.
Unfortunately, they were accidentally added in our FE types and should be removed because they cause confusion, given that the FAPI will return an error if someone tries to use them.

### DIFF
--- a/packages/clerk-js/src/core/resources/OrganizationInvitation.ts
+++ b/packages/clerk-js/src/core/resources/OrganizationInvitation.ts
@@ -22,13 +22,13 @@ export class OrganizationInvitation extends BaseResource implements Organization
 
   static async create(
     organizationId: string,
-    { emailAddress, role, redirectUrl }: CreateOrganizationInvitationParams,
+    { emailAddress, role }: CreateOrganizationInvitationParams,
   ): Promise<OrganizationInvitationResource> {
     const json = (
       await BaseResource._fetch<OrganizationInvitationJSON>({
         path: `/organizations/${organizationId}/invitations`,
         method: 'POST',
-        body: { email_address: emailAddress, role, redirect_url: redirectUrl } as any,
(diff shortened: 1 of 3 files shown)


nikosdouvlis (approved):
Looks good!
Just double checking.. in theory this could be an annoying breaking change if we simply ignored the prop till now. If passing it resulted in an error, we're good to go 🙂

dimkl:
❓ Do we have any metrics if we anyone is using this parameter? 
Removing a parameter is a breaking change but if it's always cause the request to fail i believe we should make this change without deprecation & major version bump process.

panteliselef (approved):

alex-ntousias:
@nikosdouvlis @dimkl We've been monitoring the logs for several weeks now and noone is using it. Also, if anyone uses it, they would get an error anyway. So it's safe to be removed.

PR #946, opened 2023-03-17
chore(nextjs): Use tsup to bundle and minify package

dimkl:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

## Before
Screenshot 2023-03-17 at 14 44 04

## After
Screenshot 2023-03-17 at 14 44 33

### CJS
Screenshot 2023-03-17 at 14 50 21

### ESM
Screenshot 2023-03-17 at 14 49 35

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -30,11 +30,12 @@
   ],
   "scripts": {
     "prepublishOnly": "npm run build",
-    "build": "tsc -p tsconfig.build.json",
+    "dev": "tsup --watch",
+    "build": "tsup --env.NODE_ENV production",
+    "build:analyze": "tsup --metafile --env.NODE_ENV production",
     "clean": "rimraf ./dist",
     "lint": "eslint .",
-    "test": "jest",
-    "dev": "tsc -p tsconfig.build.json --watch"
+    "test": "jest"
   },
   "dependencies": {
     "@clerk/backend": "^0.16.0",
@@ -50,6 +51,7 @@
     "jest": "*",
     "node-fetch-native": "^0.1.8",
     "ts-jest": "*",
+    "tsup": "*",
     "typescript": "*"
   },
   "peerDependencies": {
--- a/packages/nextjs/tsup.config.ts
+++ b/packages/nextjs/tsup.config.ts
@@ -1,0 +1,30 @@
+import { defineConfig } from 'tsup';
+
+import { name, version } from './package.json';
+
+export default defineConfig(overrideOptions => {
+  const isProd = overrideOptions.env?.NODE_ENV === 'production';
+
+  return {
+    entry: [
+      'src/index.ts',
+      'src/ssr/index.ts',
+      'src/server/index.ts',
+      'src/api/index.ts',
+      'src/middleware/index.ts',
+      'src/app-beta/index.ts',
+      'src/client/index.ts',
+      'src/constants.ts',
+    ],
+    onSuccess: 'tsc --emitDeclarationOnly --declaration',
+    minify: isProd,
+    sourcemap: true,
+    format: ['cjs', 'esm'],
+    define: {
+      PACKAGE_NAME: `"${name}"`,
+      PACKAGE_VERSION: `"${version}"`,
+      __DEV__: `${!isProd}`,
+    },
+    legacyOutput: true,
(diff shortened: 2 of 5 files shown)


packages/nextjs/package.json
```diff
@@ -30,11 +30,11 @@
   ],
   "scripts": {
     "prepublishOnly": "npm run build",
-    "build": "tsc -p tsconfig.build.json",
+    "dev": "tsup --minify --watch",
```
panteliselef, line 33: ❓Why do we minify dev ?

dimkl: no reason. i will remove it.

panteliselef (approved):

SokratisVidros (approved):

packages/nextjs/package.json
```diff
@@ -30,11 +30,11 @@
   ],
   "scripts": {
     "prepublishOnly": "npm run build",
-    "build": "tsc -p tsconfig.build.json",
+    "dev": "tsup --watch",
+    "build": "tsup --minify --env.NODE_ENV production",
```
nikosdouvlis, line 34:
```suggestion
    "build": "tsup --env.NODE_ENV production",
```
I think that minify is enabled by default for prod builds :)

packages/nextjs/package.json
```diff
@@ -50,6 +50,7 @@
     "jest": "*",
     "node-fetch-native": "^0.1.8",
     "ts-jest": "*",
+    "tsup": "^6.6.3",
```
nikosdouvlis, line 53: Shall we use the tsup version from the root package.json?
```suggestion
    "tsup": "*",
```

packages/nextjs/tsup.config.ts
```diff
@@ -0,0 +1,22 @@
+import { defineConfig } from 'tsup';
+
+import { name, version } from './package.json';
+
+export default defineConfig(overrideOptions => {
+  const isProd = overrideOptions.env?.NODE_ENV === 'production';
+
+  return {
+    entry: ['src/index.ts', 'src/ssr/index.ts', 'src/server/index.ts', 'src/api/index.ts', 'src/client/index.ts'],
```
nikosdouvlis, line 9: Does this change alter the structure of the `/dist` directory? 

We implement subpath imports (eg `@clerk/nextjs/server` by including the top-level .js and d.ts files (see https://github.com/clerkinc/javascript/blob/00bc852a38c1bc5fc46e9f433c7279c9884108ba/packages/nextjs/package.json#L19-L18)

We need to make sure that these point to the correct /dist files, otherwise the package will break even if publishing succeeds

packages/nextjs/tsup.config.ts
```diff
@@ -0,0 +1,22 @@
...
+    minify: isProd,
+    sourcemap: true,
+    format: ['cjs', 'esm'],
+    // Uncomment this to get metafiles for esbuild bundle analysis
+    // metafile: true,
```
nikosdouvlis, line 14: Can we add a npm script to build and analyze?
EG: `npm run build:analyze` or similar (up to you) 

We can make this standard across our packages.

SokratisVidros (approved):

packages/nextjs/tsup.config.ts
dimkl: there are some differences in the `dist/`  folder compared to the old one. I will update the tsup config include some subpaths.
@nikosdouvlis  let'ts discuss offline about a way to test if all currently supported exports are covered int he new bundle.

nikosdouvlis (approved):

PR #948, opened 2023-03-17
Cor 213

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This PR is the first of a many that will add support for the multi-domain feature for development instances
### adds support for the following `signInUrl` prop
```jsx
<ClerkProvider isSatellite signInUrl="https://primary.com/sign-in" >
```

### Redirects a satellite app to primary
By using the value set in `signInUrl` we perform a redirect with a query string appended `?__clerk_syncing=true&redirect_url={window.location.href}`

### Primary app redirects back to satellite app
- Use the `redirect_url` from the query string that was set from the step above
- Redirect back to satellite app :
  - Append the query string param ?__clerk_synced=true to the value of `redirect_url`.  This will prevent the satellite app from request syncing again from primary (causing an infinite loop of redirects)
  - Use `redirectWithAuth()` to go back to satellite. That will allow for the `dev_session` hash to be appended and later retrieved form the satellite app

### Back to satelllite
At this point we have completed the syncing process and we either have an app with a dev_session if one existed in primary or we don't because there was none to retrieve.
The existing `__clerk_synced=true` query param in the url will prevent the satellite from syncing again.

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -587,7 +587,7 @@ export default class Clerk implements ClerkInterface {
     return await customNavigate(stripOrigin(toURL));
   };
 
-  public buildUrlWithAuth(to: string): string {
+  public buildUrlWithAuth(to: string, strict = true): string {
     if (this.#instanceType === 'production' || !this.#devBrowserHandler?.usesUrlBasedSessionSync()) {
       return to;
     }
@@ -596,15 +596,17 @@ export default class Clerk implements ClerkInterface {
 
     if (toURL.origin !== window.location.origin) {
       const devBrowserJwt = this.#devBrowserHandler?.getDevBrowserJWT();
-      if (!devBrowserJwt) {
+      if (!devBrowserJwt && strict) {
         return clerkMissingDevBrowserJwt();
       }
 
-      toURL.hash = setSearchParameterInHash({
-        hash: toURL.hash,
-        paramName: DEV_BROWSER_SSO_JWT_PARAMETER,
-        paramValue: devBrowserJwt,
-      });
+      if (devBrowserJwt) {
+        toURL.hash = setSearchParameterInHash({
+          hash: toURL.hash,
+          paramName: DEV_BROWSER_SSO_JWT_PARAMETER,
+          paramValue: devBrowserJwt,
+        });
+      }
     }
 
     return toURL.href;
@@ -661,6 +663,16 @@ export default class Clerk implements ClerkInterface {
     }
     return this.buildUrlWithAuth(this.#environment.displayConfig.organizationProfileUrl);
   }
+
+  public redirectToSatellite = async (): Promise<unknown> => {
+    const redirectUrl = new URL(window.location.href).searchParams.get('redirect_url') as string;
+    const primarySyncUrl = new URL(`/`, redirectUrl);
+    primarySyncUrl.searchParams.append(CLERK_SYNCED, 'true');
+
+    if (inBrowser()) {
+      return this.navigate(this.buildUrlWithAuth(primarySyncUrl.toString(), false));
+    }
+  };
 
   public redirectWithAuth = async (to: string): Promise<unknown> => {
     if (inBrowser()) {
@@ -1066,18 +1078,6 @@ export default class Clerk implements ClerkInterface {
     return this.#isRequestingSyncing();
   };
 
-  #redirectToSatelliteDev = async () => {
-    await this.redirectWithAuth(this.#buildReturnSatelliteUrlDev());
-  };
-
-  #buildReturnSatelliteUrlDev = (): string => {
-    const redirectUrl = new URL(window.location.href).searchParams.get('redirect_url') as string;
-    const primarySyncUrl = new URL(`/`, redirectUrl);
-    primarySyncUrl?.searchParams.append(CLERK_SYNCED, 'true');
-
-    return primarySyncUrl?.toString() || '';
-  };
-
   #loadInStandardBrowser = async (): Promise<boolean> => {
     this.#devBrowserHandler = createDevBrowserHandler({
       frontendApi: this.frontendApi,
@@ -1090,7 +1090,7 @@ export default class Clerk implements ClerkInterface {
       await this.#syncWithPrimaryDev();
       return false;
     } else if (this.#shouldRedirectToSatelliteDev()) {
-      await this.#redirectToSatelliteDev();
+      await this.redirectToSatellite();
       return false;
     } else if (this.#shouldSyncWithPrimary()) {
       await this.#syncWithPrimary();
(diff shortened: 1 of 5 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -585,7 +587,7 @@ export default class Clerk implements ClerkInterface {
     return await customNavigate(stripOrigin(toURL));
   };
 
-  public buildUrlWithAuth(to: string): string {
+  public buildUrlWithAuth(to: string, strict = true): string {
```
SokratisVidros, line 590: @panteliselef Can you please elaborate why do we need the `strict` parameter in this function?

packages/clerk-js/src/core/constants.ts
```diff
@@ -6,6 +6,7 @@ export const DEV_BROWSER_SSO_JWT_HTTP_HEADER = 'Clerk-Cookie';
 
 export const CLERK_MODAL_STATE = '__clerk_modal_state';
 export const CLERK_SYNCED = '__clerk_synced';
+export const CLERK_SYNCING = '__clerk_syncing';
```
SokratisVidros, line 9: ❓ Is there a way to avoid the extra parameter?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1039,8 +1049,50 @@ export default class Clerk implements ClerkInterface {
...
+  };
+
   #loadInStandardBrowser = async (): Promise<boolean> => {
-    if (this.#shouldSyncWithPrimary()) {
+    this.#devBrowserHandler = createDevBrowserHandler({
+      frontendApi: this.frontendApi,
+      fapiClient: this.#fapiClient,
+    });
+
+    if (this.#hasJustSynced()) {
```
SokratisVidros, line 1087: ⛔ @panteliselef as the logic evolves, I'd suggest moving all the multidomain-related code in `syncWithPrimary` method for both dev and production instances.

SokratisVidros (changes requested):

packages/clerk-js/src/core/clerk.ts
panteliselef: @SokratisVidros  This was an attempt to prevent the error from throwing for the following scenario.

- User lands on the satellite website (first time visit)
- Redirecting to primary website (first time visit)
- Primary does not have a _dev_session in localstorage
- redirectWithAuth for satellite will throw an error because _dev_session was not found.

packages/clerk-js/src/core/clerk.ts
panteliselef: I will refactor this :)

packages/clerk-js/src/core/constants.ts
panteliselef: Just for clarification 
- `__clerk_synced` is used by a satellite app in order to prevent syncing from firing more that once & to mark that syncing has been attempted.
- `__clerk_syncing` is expected from a primary app in order to perform the redirection back to the satellite. It will also be used to throw the interstitial so the redirection happens without flickering

When we are constructing the URL that contains `__clerk_syncing=true` we always pair it with a `redirect_url`. 
Do you think it will be sufficient enough to only use the `redirect_url` query param ,that points to the satellite app, for this flow to work ?

dimkl (approved):

packages/clerk-js/src/core/clerk.ts
```diff
@@ -660,6 +664,16 @@ export default class Clerk implements ClerkInterface {
     return this.buildUrlWithAuth(this.#environment.displayConfig.organizationProfileUrl);
   }
 
+  public redirectToSatellite = async (): Promise<unknown> => {
+    const redirectUrl = new URL(window.location.href).searchParams.get('redirect_url') as string;
```
SokratisVidros, line 668: ❓ Do we need to parse the redirect_url parameter as a URL? How will this behave if it's a path and not a full URL? My suggestion is to decode the search param and use it as is.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1001,10 +1015,19 @@ export default class Clerk implements ClerkInterface {
...
-  #buildPrimarySyncUrl = (): string => {
+  #buildSyncUrlWithSignInUrl = (): string => {
+    const primarySyncUrl = new URL(`/`, this.#options.signInUrl);
+    primarySyncUrl?.searchParams.append(CLERK_SYNCING, 'true');
+    primarySyncUrl?.searchParams.append('redirect_url', window.location.href);
+
+    return primarySyncUrl?.toString() || '';
+  };
+
+  #buildSyncUrlWithFAPI = (): string => {
```
SokratisVidros, line 1030:
```suggestion
  #buildSyncUrlForProductionInstances = (): string => {
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1022,24 +1045,54 @@ export default class Clerk implements ClerkInterface {
...
+      this.#clearJustSynced();
       return false;
     }
+    return this.#shouldSyncWithPrimaryDev() || this.#shouldSyncWithPrimaryProd() || this.#shouldRedirectToSatellite();
+  };
 
-    if (this.#hasJustSynced()) {
-      this.#clearJustSynced();
+  #shouldSyncWithPrimaryDev = (): boolean => {
+    if (this.isSatellite && !this.#options.signInUrl && this.#instanceType === 'development') {
```
SokratisVidros, line 1056:
```suggestion
    if (this.#instanceType === 'development' && this.isSatellite && !this.#options.signInUrl) {
```

packages/clerk-js/src/core/errors.ts
```diff
@@ -97,3 +97,7 @@ export function clerkMissingProxyUrlAndDomain(): never {
     `${errorPrefix} Missing domain and proxyUrl. A satellite application needs to specify a domain or a proxyUrl`,
   );
 }
+
+export function clerkMissingSignInUrlAsSatellite(): never {
+  throw new Error(`${errorPrefix} Missing signInUrl. Pass a signInUrl for dev instances if an app is satellite`);
```
SokratisVidros, line 102:
```suggestion
throw new Error(`${errorPrefix} Missing signInUrl. A satellite application needs to specify the signInUrl for development instnaces.`);
```

SokratisVidros (changes requested):

packages/clerk-js/src/core/constants.ts
panteliselef: Yes probably `redirect_url` is not sufficient enough on its own. Let's combine both of them into the new `__clerk_satellite_url` as proposed.

SokratisVidros (approved):

PR #962, opened 2023-03-21
fix(remix): Fix support for plain objects in handler result

dimkl:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

- [] `npm test` runs as expected.
- [] `npm run build` runs as expected.

issue: https://github.com/clerkinc/javascript/issues/960
introduced in: https://github.com/clerkinc/javascript/pull/750


nikosdouvlis (approved):

packages/remix/src/ssr/rootAuthLoader.ts
```diff
@@ -68,5 +68,5 @@ export const rootAuthLoader: RootAuthLoader = async (
   }
 
   // if the user returned a plain object, simply inject requestState into an empty response
```
nikosdouvlis, line 70:
```suggestion
  // if the user returned a plain object, create an empty response and inject requestState and handlerResult
```

nikosdouvlis (approved):

dimkl:
@nikosdouvlis PTAL

packages/remix/src/ssr/rootAuthLoader.ts
```diff
@@ -25,8 +25,8 @@ interface RootAuthLoader {
 
 export const rootAuthLoader: RootAuthLoader = async (
   args: LoaderFunctionArgs,
-  handlerOrOptions: any,
-  options?: any,
+  handlerOrOptions?: RootAuthLoaderCallback<RootAuthLoaderOptions> | RootAuthLoaderOptions,
+  options?: RootAuthLoaderOptions,
```
dimkl, line 29: ❓ since `RootAuthLoader` type is defined above to have `callback: RootAuthLoaderCallback<RootAuthLoaderOptions> ` and `options?: RootAuthLoaderOptions`, do you think this change is breaking change?

playground/remix-node/app/root.tsx
```diff
@@ -42,6 +42,9 @@ export const meta: MetaFunction = () => ({
 export const CatchBoundary = ClerkCatchBoundary();
 
 function App() {
+  const loaderData = useLoaderData();
+  console.log('root: ', { loaderData });
```
dimkl, line 46: i've validated the fix using the log of this line. Before the fix this was logging only the clerk state and after the fix this was logging also an `user` .

packages/remix/src/ssr/rootAuthLoader.ts
panteliselef: I think this only affects the types inside the function. (Due to more strict types set to `rootAuthLoader: RootAuthLoader`) But I am not 100% sure.
A safer approach would be to type cast with `as` those variables inside the function, and the signature will not be affected for sure

playground/remix-node/app/root.tsx
nikosdouvlis: Great

packages/remix/src/ssr/rootAuthLoader.ts
nikosdouvlis: Would you think this is a cleaner approach? https://github.com/clerkinc/javascript/blob/c7e61d9eb067694b790b52983cd57f968938ed69/packages/nextjs/src/server/withClerkMiddleware.ts#L44

cc @dimkl , @panteliselef

panteliselef: yes

dimkl: me too. but i will revert this change for the moment to merge this PR and we will refactor it in the future

nikosdouvlis (approved):

PR #966, opened 2023-03-22
Append `__domain` query param to FAPI requests for satellite apps with dev instances

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [x] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This PR tackles:

- changes to fapiClient when domain is used in dev instances
- constructing clerk.js script url when domain is used in dev instances (affects the react package and interstitial)
- Clerk.js throws an error if isSatellite is set without proxyUrl or domain. (for both prod and dev instances)
- withClerkMiddleware and rootAuthLoader  throw the same error mentioned in the previous line

In order to complete the above items some refactor was needed:
- `isDevOrStagingUrl` utility function  was moved from `clerk-js` to `shared`
- fixed types for Clerk.js Singleton (domain, isSatellite)

### DIFF
--- a/packages/backend/src/tokens/interstitial.ts
+++ b/packages/backend/src/tokens/interstitial.ts
@@ -6,6 +6,7 @@ import { API_VERSION } from '../constants';
 import runtime from '../runtime';
 import { callWithRetry } from '../util/callWithRetry';
 import { isStaging } from '../util/instance';
+import { isDevOrStagingUrl } from '../util/isDevOrStagingUrl';
 import { parsePublishableKey } from '../util/parsePublishableKey';
 import { joinPaths } from '../util/path';
 import { TokenVerificationError, TokenVerificationErrorAction, TokenVerificationErrorReason } from './errors';
@@ -40,6 +41,7 @@ export function addClerkPrefix(str: string | undefined) {
 
 export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions, 'apiUrl'>) {
   options.frontendApi = parsePublishableKey(options.publishableKey)?.frontendApi || options.frontendApi || '';
+  const domainOnlyInProd = !isDevOrStagingUrl(options.frontendApi) ? addClerkPrefix(options.domain) : '';
   const { debugData, frontendApi, pkgVersion, publishableKey, proxyUrl, isSatellite = false, domain } = options;
   return `
     <head>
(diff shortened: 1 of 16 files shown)


packages/backend/src/tokens/interstitial.ts
```diff
@@ -40,6 +41,7 @@ export function addClerkPrefix(str: string | undefined) {
 
 export function loadInterstitialFromLocal(options: Omit<LoadInterstitialOptions, 'apiUrl'>) {
   options.frontendApi = parsePublishableKey(options.publishableKey)?.frontendApi || options.frontendApi || '';
+  const domainOnlyInProd = !isDevOrStagingUrl(options.frontendApi) ? addClerkPrefix(options.domain) : '';
```
SokratisVidros, line 44:
```suggestion
  const isDomainForProduction = !isDevOrStagingUrl(options.frontendApi) ? addClerkPrefix(options.domain) : '';
```

packages/backend/src/util/isDevOrStagingUrl.ts
```diff
@@ -0,0 +1,34 @@
+// TODO: use the same function from @clerk/shared once treeshakable
```
SokratisVidros, line 1: @nikosdouvlis Did we resolve the tree-shaking issues with `@clerk/shared` yet?

SokratisVidros (approved):

packages/backend/src/util/isDevOrStagingUrl.ts
panteliselef: #946 should handle the issue with the treeshakable packages
cc @dimkl

packages/backend/src/tokens/interstitial.ts
panteliselef: @SokratisVidros  this does return a boolean but a string.

We expect that `domainOnlyInProd` will have a valid value if instance is prod otherwise we assign an empty string. This is on purpose so that the following line works as expected
```typescript
script.src = '${getScriptUrl(proxyUrl || domainOnlyInProd || frontendApi, pkgVersion)}';
```

packages/backend/src/util/isDevOrStagingUrl.ts
dimkl: The mentioned PR is related to the `nextjs` package and not the entire project. This change is inside the `backend` package. Why is that related to the treeshaking of the other PR ?

PR #969, opened 2023-03-22
COR-199 fix(clerk-js): Refactor country code method and remove local storage …

raptisj:
…for country

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected
Refactor country code method and remove local storage for country
- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -127,6 +127,7 @@ export default class Clerk implements ClerkInterface {
   public session?: ActiveSessionResource | null;
   public organization?: OrganizationResource | null;
   public user?: UserResource | null;
+  public __internal_country?: string | null;
   public readonly frontendApi: string;
   public readonly publishableKey?: string;
   public readonly proxyUrl?: ClerkInterface['proxyUrl'];
@@ -965,6 +966,12 @@ export default class Clerk implements ClerkInterface {
     this.#environment = environment;
     this.#authService?.setEnvironment(environment);
   }
+
+  __internal_setCountry = (country: string | null) => {
+    if (!this.__internal_country) {
+      this.__internal_country = country;
+    }
+  };
 
   updateClient = (newClient: ClientResource): void => {
     if (!this.client) {
--- a/packages/clerk-js/src/core/resources/Base.ts
+++ b/packages/clerk-js/src/core/resources/Base.ts
@@ -47,6 +47,11 @@ export abstract class BaseResource {
 
     const { payload, status, statusText } = fapiResponse;
 
+    if (payload?.meta) {
+      const country = payload.meta.responseHeaders?.country;
+      this.clerk.__internal_setCountry(country ? country.toLowerCase() : null);
+    }
+
     // TODO: Link to Client payload piggybacking design document
     if (requestInit.method !== 'GET' || opts.forceUpdateClient) {
       this._updateClient<J>(payload);
--- a/packages/clerk-js/src/core/resources/Base.ts
+++ b/packages/clerk-js/src/core/resources/Base.ts
@@ -45,10 +45,10 @@ export abstract class BaseResource {
       }
     }
 
-    const { payload, status, statusText } = fapiResponse;
+    const { payload, status, statusText, headers } = fapiResponse;
 
     if (payload?.meta) {
-      const country = payload.meta.responseHeaders?.country;
+      const country = headers.get('country');
       this.clerk.__internal_setCountry(country ? country.toLowerCase() : null);
     }
 
@@ -119,8 +119,7 @@ export abstract class BaseResource {
       opts,
     );
 
-    const data = json?.response ? { ...json.response, ...json.meta } : json;
-    return this.fromJSON(data as J);
+    return this.fromJSON((json?.response || json) as J);
   }
 
   protected async _baseMutate<J extends ClerkResourceJSON | null>({
--- a/packages/clerk-js/src/core/resources/Environment.ts
+++ b/packages/clerk-js/src/core/resources/Environment.ts
@@ -7,7 +7,6 @@ import type {
   UserSettingsResource,
 } from '@clerk/types';
 
-import type { CountryIso } from '../../ui/elements/PhoneInput/countryCodeData';
 import { AuthConfig, BaseResource, DisplayConfig, UserSettings } from './internal';
 import { OrganizationSettings } from './OrganizationSettings';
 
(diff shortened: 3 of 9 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -945,6 +947,12 @@ export default class Clerk implements ClerkInterface {
     this.#authService?.setEnvironment(environment);
   }
 
+  setCountry = (country: CountryIso | null) => {
```
SokratisVidros, line 950: ⛔ This method needs to be private to Clerk so as not to be exposed as a public API, for example `private __internal_setCountry = (...)`.

SokratisVidros (changes requested):

packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -1,6 +1,7 @@
 import { isValidBrowserOnline } from '@clerk/shared';
 import type { ClerkAPIErrorJSON, ClerkResourceJSON, ClerkResourceReloadParams, DeletedObjectJSON } from '@clerk/types';
 
+import type { CountryIso } from '../../ui/elements/PhoneInput/countryCodeData';
```
panteliselef, line 4: 🔧 Can we refactor this to avoid using any `/ui/`paths inside core ?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -51,6 +51,7 @@ import type {
 } from '@clerk/types';
 
 import type { MountComponentRenderer } from '../ui/Components';
+import type { CountryIso } from '../ui/elements/PhoneInput/countryCodeData';
```
panteliselef, line 54: 🔧 Can we refactor this to avoid using any /ui/paths inside core ?

packages/types/src/countryIso.ts
```diff
@@ -0,0 +1,234 @@
+export type CountryIso =
```
nikosdouvlis, line 1: I'm not sure about the duplicated type here. Can we use a simple `string` type?
The correct approach would be to extract countryData.ts into clerk/shared and reuse that, but this looks unnecessary right now!

packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -47,6 +53,11 @@ export abstract class BaseResource {
 
     const { payload, status, statusText } = fapiResponse;
 
+    if (payload?.meta) {
+      const country = payload?.meta.responseHeaders?.country || 'us';
```
nikosdouvlis, line 57: Why do we fallback to `us` and check for `: null` right below? 
We probably shouldn't fallback to anything. If we didn't get a country, the value should be `null` and the components should handle `null` accordingly

nikosdouvlis (changes requested):

packages/clerk-js/src/core/resources/Base.ts
panteliselef: Yes let's allow for nullable country to be set. And let the component handle he fallback. Although we may endup using `us` there in order to display a default selected option for the country extension

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useLayoutEffect, useState } from 'react';
 
-import { useEnvironment } from '../../contexts';
+import { getLongestValidCountryCode } from '../../../ui/utils/phoneUtils';
```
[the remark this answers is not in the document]
raptisj: @nikosdouvlis I removed this. Probably a leftover from a rebase

packages/types/src/countryIso.ts
raptisj: @nikosdouvlis I removed the redundant type and used a type of string instead

packages/types/src/clerk.ts
```diff
@@ -89,6 +89,9 @@ export interface Clerk {
   /** Current User. */
   user?: UserResource | null;
 
+  /** User country code */
+  country?: string | null;
```
nikosdouvlis, line 93: Similarly to my previous comment, this should be hidden/internal

nikosdouvlis (approved):

packages/clerk-js/src/core/resources/Environment.ts
```diff
@@ -59,9 +58,10 @@ export class Environment extends BaseResource implements EnvironmentResource {
 
   protected fromJSON(data: EnvironmentJSON | null): this {
     if (data) {
+      const country = data.meta?.responseHeaders?.country;
```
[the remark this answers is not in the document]
nikosdouvlis: John enchanced the `FapiResponseJSON` interface by adding the `meta.reponseHeaders` prop which is populated within `fapiClient`, before the response object reaches the Base resource.

```tsx
export interface FapiResponseJSON<T> {
  response: T;
  client?: ClientJSON;
  errors?: ClerkAPIErrorJSON[];
  meta?: {
    client?: ClientJSON;
    session_id?: string;
    responseHeaders?: { country: string | null };
  };
}
```

I think we're set, but do you require any other adjustments?

packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -47,6 +47,11 @@ export abstract class BaseResource {
 
     const { payload, status, statusText } = fapiResponse;
 
+    if (payload?.meta) {
+      const country = payload?.meta.responseHeaders?.country;
```
panteliselef, line 51:
```suggestion
      const country = payload.meta.responseHeaders?.country;
```
🙃 payload.meta exists inside `if`

panteliselef (approved):

SokratisVidros (approved):

packages/clerk-js/src/core/resources/Environment.ts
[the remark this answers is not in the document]
nikosdouvlis: Update: we moved it to `FapiResponseJSON.headers`

PR #977, opened 2023-03-23
fix(clerk-react): Fallback to Clerk.isReady if Clerk.loaded is not available

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This change makes clerk/clerk-react compatible with both clerk-js@3 and clerk-js@4 in regards to loading in standards browser. This allows the accounts project to load any of these 2 versions on a per-request basis, even if the latest clerk/clerk-react package is installed

### DIFF
--- a/packages/react/src/isomorphicClerk.ts
+++ b/packages/react/src/isomorphicClerk.ts
@@ -3,6 +3,7 @@ import type {
   ActiveSessionResource,
   AuthenticateWithMetamaskParams,
   Clerk,
+  Clerk as ClerkInterface,
   ClientResource,
   CreateOrganizationParams,
   CreateOrganizationProps,
@@ -22,7 +23,6 @@ import type {
   UserProfileProps,
   UserResource,
 } from '@clerk/types';
-import type { Clerk as ClerkInterface } from '@clerk/types';
 import type { OrganizationProfileProps, OrganizationSwitcherProps } from '@clerk/types/src';
 
 import type {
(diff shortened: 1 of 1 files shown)


packages/react/src/isomorphicClerk.ts
```diff
@@ -166,7 +166,10 @@ export default class IsomorphicClerk {
         await global.Clerk.load(this.options);
       }
 
-      return global.Clerk?.loaded ? this.hydrateClerkJS(global.Clerk) : undefined;
+      if (global.Clerk?.loaded || global.Clerk?.isReady()) {
```
panteliselef, line 169: ❓Aren't both returning the same value ? (this.#isReady)

packages/react/src/isomorphicClerk.ts
```diff
@@ -22,7 +23,6 @@ import type {
   UserProfileProps,
   UserResource,
 } from '@clerk/types';
-import type { Clerk as ClerkInterface } from '@clerk/types';
```
panteliselef, line 25: 💯

packages/react/src/isomorphicClerk.ts
nikosdouvlis: Yup, but Clerk.loaded does not exist on clerk-js@3 and we'd like clerk-react to be able to load either both Clerk v3 and v4

panteliselef (approved):

PR #989, opened 2023-03-27
Enable switching through phone and emailOrUsername in SignInStart

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

As of this PR, the sign in flow will be much more fluid, showing either an input for the email/username, either the phone input for the phone number. The user will be able to switch between the two via an action right on top of the input (as seen in the pictures). 4 localization keys have been added. One for each possible case (`Use email`, `Use phone`, `Use username`, `Use email or username`)

Before: 
Screenshot 2023-03-28 at 13 46 41

After:
Screenshot 2023-03-28 at 13 44 43
Screenshot 2023-03-28 at 13 45 05


packages/clerk-js/src/ui/components/SignIn/SignInFactorOnePasswordCard.tsx
```diff
@@ -66,6 +66,14 @@ export const SignInFactorOnePasswordCard = (props: SignInFactorOnePasswordProps)
           gap={8}
         >
           <Form.Root onSubmit={handlePasswordSubmit}>
+            {/* For password managers */}
+            <input
+              readOnly
```
panteliselef, line 71: ❓Why did we decided not to go with type='hidden' ? Do we have extra benefits with this approach ?

packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
```diff
@@ -33,26 +38,59 @@ export function _SignInStart(): JSX.Element {
...
   });
 
   const identifierField = useFormControl('identifier', '', {
-    ...identifierInputDisplayValues,
+    ...currentIdentifier,
     isRequired: true,
   });
 
+  const switchToNextIdentifier = () => {
+    setIdentifierAttribute(
```
SokratisVidros, line 69: 💯

packages/clerk-js/src/ui/utils/mergeRefs.ts
```diff
@@ -0,0 +1,10 @@
+export const mergeRefs = <T>(...refs: React.RefObject<T>[]) => {
```
SokratisVidros, line 1: ❓ Can you please elaborate on this?

packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
```diff
@@ -235,16 +275,39 @@ export function _SignInStart(): JSX.Element {
 }
 
 const InstantPasswordRow = ({ field }: { field?: FormControlState<'password'> }) => {
+  const [autofilled, setAutofilled] = useState(false);
+  const ref = useRef<HTMLInputElement>(null);
+
+  // show password if it's autofilled by the browser
+  React.useLayoutEffect(() => {
+    const intervalId = setInterval(() => {
```
SokratisVidros, line 283: 💡  Instead of a poller, how about using a mutation observer and possibly avoid the getComputedStyle invocation?

SokratisVidros (approved):

packages/clerk-js/src/ui/utils/mergeRefs.ts
desiprisg: This is used in case we want to forward a ref in the component, and also `useRef` inside the component itself for some other reason.

packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
desiprisg: trying this out

desiprisg: It doesn't seem to fire in that case.

packages/types/src/localization.ts
```diff
@@ -121,6 +139,10 @@ type _LocalizationResource = {
       subtitle: LocalizationValue;
       actionText: LocalizationValue;
       actionLink: LocalizationValue;
+      useEmailActionLink: LocalizationValue;
```
nikosdouvlis, line 142: Shouldn't these have the same base name and an extra id after `__` like
```
  actionLink__use_email_username: LocalizationValue;
```

nikosdouvlis (approved):

packages/types/src/localization.ts
desiprisg: Right

packages/types/src/localization.ts
```diff
@@ -36,9 +45,18 @@ type _LocalizationResource = {
...
+   * @deprecated
+   */
   formFieldInputPlaceholder__emailAddress_phoneNumber: LocalizationValue;
   formFieldInputPlaceholder__emailAddress_username: LocalizationValue;
+  /**
+   * @deprecated
+   */
   formFieldInputPlaceholder__phoneNumber_username: LocalizationValue;
+  /**
+   * @deprecated
```
dimkl, line 58: ❓ should we not add here what should be used instead ?

PR #993, opened 2023-03-28
Throw error when domain is set as function for non-…

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Accessing this.domain inside Isomorphic clerk will result to runtime error when the window is undefined. 

For non-browser environments, prohibit the usage of a function for now.

### DIFF
--- a/packages/react/src/errors.ts
+++ b/packages/react/src/errors.ts
@@ -31,3 +31,6 @@ export function isMagicLinkError(err: any): err is MagicLinkError {
 
 export const invalidStateError =
   'Invalid state. Feel free to submit a bug or reach out to support here: https://clerk.com/support';
+
+export const domainAsFunctionError =
+  'The usage of a function as domain is not supported in non-browser environments. Please contact support@clerk.com';
--- a/packages/react/src/errors.ts
+++ b/packages/react/src/errors.ts
@@ -32,5 +32,5 @@ export function isMagicLinkError(err: any): err is MagicLinkError {
 export const invalidStateError =
   'Invalid state. Feel free to submit a bug or reach out to support here: https://clerk.com/support';
 
-export const domainAsFunctionError =
-  'The usage of a function as domain is not supported in non-browser environments. Please contact support@clerk.com';
+export const unsupportedNonBrowserDomainFunction =
+  'Unsupported usage of domain. The usage of domain as function is not supported in non-browser environments.';
--- a/packages/react/src/isomorphicClerk.ts
+++ b/packages/react/src/isomorphicClerk.ts
@@ -3,7 +3,6 @@ import type {
   ActiveSessionResource,
   AuthenticateWithMetamaskParams,
   Clerk,
-  Clerk as ClerkInterface,
   ClientResource,
   CreateOrganizationParams,
   CreateOrganizationProps,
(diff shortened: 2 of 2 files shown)


packages/react/src/errors.ts
```diff
@@ -31,3 +31,6 @@ export function isMagicLinkError(err: any): err is MagicLinkError {
 
 export const invalidStateError =
   'Invalid state. Feel free to submit a bug or reach out to support here: https://clerk.com/support';
+
+export const domainAsFunctionError =
+  'The usage of a function as domain is not supported in non-browser environments. Please contact support@clerk.com';
```
gkats, line 36: :upside_down_face: Perhaps the "contact" part of the message is redundant?

packages/react/src/isomorphicClerk.ts
```diff
@@ -92,6 +92,16 @@ export default class IsomorphicClerk {
   }
 
   get domain(): string {
+    /**
+     * This getter can run in environments where window is not available.
+     * In those cases we should expect and use domain a string
```
gkats, line 97: :upside_down_face: 
```suggestion
     * In those cases we should expect and use domain as a string
```

gkats (approved):
:bug:  

Let's also add a commit message with more information on why this change is needed and if possible reproduce it in a test.

packages/react/src/isomorphicClerk.ts
panteliselef: Nice catch 😀

packages/react/src/errors.ts
panteliselef: TBH this was requested, but I may have taken it a bit seriously 
cc @nikosdouvlis

nikosdouvlis: I suggested this error message as apart from the components, this is one of the few features we don't support on non-browser environment for now. This  error does not indicate missuse of our apis or an actual error - just a missing feature :)

Feel free to adjust accordingly or remove completely - no real issues here!

nikosdouvlis (approved):

packages/react/src/errors.ts
panteliselef: I am thinking of changing this one to 
```typescript
export const unsupporterDomainFunctionInNonBrowser =
  'Unsupported usage of domain. The usage of domain as function is not supported in non-browser environments.';
```

nikosdouvlis: This is better, thanks Panteli

PR #999, opened 2023-03-30
fix(clerk-react,shared): Update inClientSide() util to check for document

dimkl:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This change resolve issues with react Native where `window` is defined but the API is not the same with the Browser resulting in failures when trying to use window.location or any other property.

## Changes

- Drop `inClientSide()` from `react/utils` in favour of `shared` package implementation
- Fix `inClientSide()` in `shared` to check for `document` instead of `window`

issue: https://github.com/clerkinc/javascript/issues/998

### DIFF
--- a/packages/shared/src/utils/ssr.ts
+++ b/packages/shared/src/utils/ssr.ts
@@ -1,3 +1,3 @@
 export const inClientSide = (): boolean => {
-  return typeof window !== 'undefined';
+  return typeof window !== 'undefined' && typeof document !== 'undefined';
 };
--- a/packages/react/src/isomorphicClerk.ts
+++ b/packages/react/src/isomorphicClerk.ts
@@ -1,4 +1,4 @@
-import { handleValueOrFn } from '@clerk/shared';
+import { handleValueOrFn, inClientSide } from '@clerk/shared';
 import type {
   ActiveSessionResource,
   AuthenticateWithMetamaskParams,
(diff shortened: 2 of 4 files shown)


packages/shared/src/utils/ssr.ts
```diff
@@ -1,3 +1,3 @@
 export const inClientSide = (): boolean => {
-  return typeof window !== 'undefined';
```
nikosdouvlis, line 2: Shall we check for both just in case?

nikosdouvlis (approved):

packages/shared/src/utils/ssr.ts
```diff
@@ -1,3 +1,3 @@
 export const inClientSide = (): boolean => {
```
panteliselef, line 1: 🙃 `isClientSide` or `isCSR` ?

Feel free to keep it as is

nikosdouvlis: I prefer the longest one tbh

packages/shared/src/utils/ssr.ts
dimkl: updated. PTAL

PR #1007, opened 2023-03-31
chore(backend): Use helper functions to construct interstitial script

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

The first tests for checking the contents of the interstitial html template are introduced by this commit. This allows us to test specific parts for the interstitial script that we are creating. We want to ensure that we are passing the correct parameters with the correct format to clerk-js.

### DIFF
--- a/packages/backend/src/tokens/interstitial.test.ts
+++ b/packages/backend/src/tokens/interstitial.test.ts
@@ -1,0 +1,96 @@
+import { addScriptAttributes, getClerkLoadArgs } from './interstitial';
+
+export default (QUnit: QUnit) => {
+  const { module, test } = QUnit;
+
+  module('interstitial.getClerkLoadArgs', () => {
+    test('returns empty object', assert => {
+      const payload = getClerkLoadArgs({});
+      assert.propEqual(payload, {});
+    });
+
+    test('returns empty object', assert => {
+      const payload = getClerkLoadArgs({
+        isSatellite: undefined,
+        signInUrl: undefined,
+      });
+      assert.propEqual(payload, {});
+    });
+
+    test('returns {isSatellite: true}', assert => {
+      const payload = getClerkLoadArgs({
+        isSatellite: true,
+      });
+      assert.propEqual(payload, {
+        isSatellite: true,
+      });
+    });
+    test('returns empty object', assert => {
+      const payload = getClerkLoadArgs({
+        isSatellite: false,
+      });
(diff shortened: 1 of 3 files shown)


dimkl (commented):
☁️ Is the reasoning behind this change is to make this part of code more readable and easy to maintain ?
If so, i don't know if `getClerkLoadArgs` and `addScriptAttributes` are more readable and maintainable for me :/

panteliselef:
> ☁️ Is the reasoning behind this change is to make this part of code more readable and easy to maintain ? If so, i don't know if `getClerkLoadArgs` and `addScriptAttributes` are more readable and maintainable for me :/

My intention is to ensure that we are passing the correct parameters with the correct format to clerk-js. 
While the implementation of these function is not as readable as just having to read a plain string from inside the html template, the way we call these functions is very straight forward.

My goal with this PR is to provide tests for the "moving" parts of the interstitial html template, in order to avoid mistyping a character that may brake the script and it would be hard to identify

SokratisVidros:
I agree with @dimkl. My suggestion was to approach testing the interstitial script with string matching.

alex-ntousias:
@SokratisVidros @dimkl @panteliselef What should we do with this PR? It has been open for a few weeks now.
Does the change improve the interstitial handling and debugging? Or did we identify something that's missing, but this particular solution is not the right one?
In other words, shall we merge this or close it?

dimkl:
@alex-ntousias  I believe that the introduced interstitial tests should be merged but my personal suggestion would be to use string matching to assert the cases of missing values and populated values.
If the added code makes sense and is more readable to everyone, i would agree to merge it as is.

panteliselef:
I am closing this, in order to address this issue with string matching at a later point.

PR #1033, opened 2023-04-05
feat(clerk-expo): Introduce useOAuth.ts

SokratisVidros:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [x] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

The useOAuth hook handles sign-ins or sign-ups OAuth flows automatically for Expo v48 applications.

Recommended usage:

```ts
export function OauthButtons() {
  // Warm up the android browser to improve UX
  // https://docs.expo.dev/guides/authentication/#improving-user-experience
  useWamUpBrowser();

  const { startOAuthFlow } = useOAuth({ strategy: "oauth_google" });

  const onPress = React.useCallback(async () => {
    const { createdSessionId, signIn, signUp, setSession } =
      await startOAuthFlow();

    if (createdSessionId) {
      setSession(createdSessionId);
    } else {
      // Use signIn or signUp for next steps such as MFA
    }
  }, []);

  return (
    <TouchableOpacity
      style={{ ...styles.secondaryButton, marginBottom: 20 }}
      onPress={onPress}
    >
      <Text style={styles.secondaryButtonText}>Continue with Google</Text>
    </TouchableOpacity>
  );
}
```


packages/expo/src/useOAuth.ts
```diff
@@ -0,0 +1,84 @@
...
+    await signIn.reload({ rotatingTokenNonce });
+
+    const { status, firstFactorVerification } = signIn;
+
+    let createdSessionId = '';
+
+    if (status === 'complete') {
+      createdSessionId = signIn.createdSessionId!;
+    } else if (firstFactorVerification.status === 'transferable') {
+      await signUp.create({
```
perkinsjr, line 72: If we are creating the function we need to handle what happens if the user has signed up but requires additional fields, right now we aren't handling this in this flow. 

Are we expecting that to handle this by calling the function? Right now the code wouldn't complete.

I'd expect at bit more grace like

```
const redirectUrl = AuthSession.makeRedirectUri({
        path: "/oauth-native-callback",
      });

      await signIn.create({
        strategy: parsedStrat,
        redirectUrl,
      });

      const {
        firstFactorVerification: { externalVerificationRedirectURL },
      } = signIn;

      if (!externalVerificationRedirectURL)
        throw "Something went wrong during the OAuth flow. Try again.";

      const authResult = await AuthSession.startAsync({
        authUrl: externalVerificationRedirectURL.toString(),
        returnUrl: redirectUrl,
      });

      if (authResult.type !== "success") {
        throw "Something went wrong during the OAuth flow. Try again.";
      }

      // Get the rotatingTokenNonce from the redirect URL parameters
      const { rotating_token_nonce: rotatingTokenNonce } = authResult.params;

      await signIn.reload({ rotatingTokenNonce });

      const { createdSessionId } = signIn;

      if (createdSessionId) {
        // If we have a createdSessionId, then auth was successful
        await setSession(createdSessionId);
      } else {
        // If we have no createdSessionId, then this is a first time sign-in, so
        // we should process this as a signUp instead
        // Throw if we're not in the right state for creating a new user
        if (
          !signUp ||
          signIn.firstFactorVerification.status !== "transferable"
        ) {
          throw "Something went wrong during the Sign up OAuth flow. Please ensure that all sign up requirements are met.";
        }

        console.log(
          "Didn't have an account transferring, following through with new account sign up",
        );

        // Create user
        await signUp.create({ transfer: true });
        await signUp.reload({
          rotatingTokenNonce: authResult.params.rotating_token_nonce,
        });
        await setSession(signUp.createdSessionId);
      }
```

SokratisVidros: Good point. I was thinking about all the cases and depending on the flow we have multiple outcomes for PSU sign ups, MFA in sign-ins, new security checks in the future during for both sign-in or sign-up, etc... 

The different outcomes are already modelled in signIn and signUp objects. Adding a conditional for all the combination will introduce maintenance overhead for the hook. 

So, the hook returns `{ createdSessionId, signIn, signUp, setSession}`. If createdSessionId exists then we are done.    Note that setSession is not called by the hook to give control to the developer to call it when they need to.

If createdSessionId doesn't exist,then the developer needs to use signIn or signUp, read their status and do the necessary handling.

packages/expo/package.json
```diff
@@ -44,7 +44,9 @@
     "typescript": "*"
   },
   "peerDependencies": {
-    "react": ">=16"
+    "react": ">=16",
+    "expo-auth-session": ">=4",
```
nikosdouvlis, line 48: Did we always implicitly require these versions? 
Just checking as otherwise `npm i` might throw errors about conflicting dependency versions.

packages/expo/src/useOAuth.ts
```diff
@@ -0,0 +1,84 @@
+import { useSignIn, useSignUp } from '@clerk/clerk-react';
+import type { OAuthStrategy } from '@clerk/types';
+import * as AuthSession from 'expo-auth-session';
+import * as WebBrowser from 'expo-web-browser';
+
+export type UseOAuthFlowParams =
```
nikosdouvlis, line 6: ❓ Is there a specific reason we're changing the signature of the function instead of skipping redirectUrl?

```tsx
useOAuth({ strategy: 'oauth_google' })
```
vs
```tsx
useOAuth('oauth_google')
```

I think we usually go with the 1rst one as its more extensible.

packages/expo/src/useOAuth.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+  | OAuthStrategy;
+
+export function useOAuth(params: UseOAuthFlowParams) {
+  let strategy: OAuthStrategy;
+  let redirectUrl: string;
+
+  if (typeof params === 'string') {
+    strategy = params;
+  } else if (typeof params === 'object') {
```
nikosdouvlis, line 19: Just to be extra safe we can also check for:
```suggestion
  } else if ('strategy' in params)  {
```

Also, what happens if `params` is missing? Are we throwing further down the flow?

packages/expo/src/useOAuth.ts
```diff
@@ -0,0 +1,84 @@
...
+    strategy = params.strategy;
+    redirectUrl = params.redirectUrl || '';
+  }
+
+  const { signIn, setSession, isLoaded: isSignInLoaded } = useSignIn();
+  const { signUp, isLoaded: isSignUpLoaded } = useSignUp();
+
+  async function startOAuthFlow() {
+    if (!isSignInLoaded || !isSignUpLoaded) {
```
nikosdouvlis, line 28: This is going to throw if someone tries to call `startOAuthFlow` without checking `isLoaded` in advance with a Typerror (probably `undefined is not a function`).

Since `startOAuthFlow` returns a promise already, we can wait for clerk to load within the function like we do for `useAuth.getToken` and `useAuth.signOut`: https://github.com/clerkinc/javascript/blob/a285a4e4c38bc1ed1cca6a910d2ec6d76b911822/packages/react/src/hooks/utils.ts#L18

packages/expo/src/useOAuth.ts
```diff
@@ -0,0 +1,84 @@
...
+    // @ts-expect-error
+    const { type, url } = result || {};
+
+    // TODO: Check all the possible AuthSession results
+    // https://docs.expo.dev/versions/latest/sdk/auth-session/#returns-7
+    if (type !== 'success') {
+      throw 'Something went wrong during the OAuth flow. Try again.';
+    }
+
+    const params = new URL(url).searchParams;
```
nikosdouvlis, line 60: I'm asking because I'm not sure so please ignore if I'm completely off :) Is `URL` supported by all expo versions we support?

nikosdouvlis (approved):

packages/expo/src/useOAuth.ts
nikosdouvlis: We probably need to create a helper to handle validation + type inference for overloaded functions and params. Something like a mini-zod implementation tailored to our use cases.

packages/expo/package.json
SokratisVidros: The expo-* dependencies are new.

packages/expo/src/useOAuth.ts
SokratisVidros: Fixed. I've also just kept the object although I do like the flexibiity of passing just a strategy.

packages/expo/src/useOAuth.ts
SokratisVidros: Yes

nikosdouvlis (approved):

PR #1035, opened 2023-04-05
Type testing infra

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] other: Testing

## Packages affected

- [ ] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

We can test our types with the `expect-type` package. This package gives us an api with which we can test a value against a type. By setting `declarations: true` in our jest config, we will get errors from typescript as well.

The idea is to test our top level types in this PR, and set a precedent as to how these tests will be implemented.

In case we want to test a function directly (and not its exposed type), a cool trick would be to create a callback that executes that function with the desired arguments. This way, we can verify the validity of the types without actually running the function. Props @nikosdouvlis for the idea. I will create a utility for this purpose and I'll add a couple of tests using it.


packages/react/src/contexts/__tests__/ClerkProvider.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+      expectTypeOf({
+        ...defaultProps,
+        appearance: { layout: { nonExistentKey: '' } },
+      }).not.toMatchTypeOf<ClerkProviderProps>();
+
+      // expectTypeOf({
+      //   ...defaultProps,
+      //   appearance: { elements: { nonExistentKey: '' } },
+      // }).not.toMatchTypeOf<ClerkProviderProps>();
```
panteliselef, line 107: ❓Should this be removed ?

nikosdouvlis (approved):
Excellent @desiprisg

PR #1036, opened 2023-04-05
Password Complexity

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description


packages/clerk-js/src/ui/hooks/usePasswordComplexity.ts
```diff
@@ -0,0 +1,99 @@
+import type { PasswordSettingsData } from '@clerk/types';
+import { useCallback, useMemo, useState } from 'react';
+
+import { localizationKeys, useLocalizations } from '../localization';
+
+const testComplexityCases = (password: string) => {
+  return {
+    min_length: password.length >= 8,
```
dimkl, line 8: 🔧 there are some settings missing (eg `max_length`) based on the defined `PasswordSettingsData ` type.

packages/clerk-js/src/ui/components/UserProfile/PasswordPage.tsx
```diff
@@ -119,6 +117,7 @@ export const PasswordPage = withCardStateProvider(() => {
             <Form.Control
               {...passwordField.props}
               minLength={6}
+              complexity={true}
```
raptisj, line 120: @panteliselef  🙃(super-nit) couldn’t this be just `complexity` with no value true? Just like required

packages/clerk-js/src/ui/elements/PasswordInput.tsx
```diff
@@ -1,13 +1,48 @@
...
-  const { id, ...rest } = props;
+  const { id, onChange, complexity = false, ...rest } = props;
+  const formControlProps = useFormControl();
+  const {
+    userSettings: { passwordSettings },
+  } = useEnvironment();
+
+  const { setPassword } = usePasswordComplexity(
+    //TODO: Remove this once default values are populated in DB
+    { ...passwordSettings, max_length: passwordSettings.max_length === 0 ? 999 : 0 },
```
raptisj, line 26: @panteliselef 🙃  `max_length` is set to 100 in the BE. Let's replace the 999 with 100

packages/types/src/localization.ts
```diff
@@ -91,6 +91,15 @@ type _LocalizationResource = {
   membershipRole__admin: LocalizationValue;
   membershipRole__basicMember: LocalizationValue;
   membershipRole__guestMember: LocalizationValue;
+  passwordComplexity: {
```
raptisj, line 94: @panteliselef I don't this is the proper place to put these. We have `unstable__errors` when it comes to non-FAPI errors. `passwordComplexity.minimumLength` doesn’t indicate that this is about an error while reading it

packages/localizations/src/es-ES.ts
```diff
@@ -71,6 +71,15 @@ export const esES: LocalizationResource = {
   membershipRole__admin: 'Administrador',
   membershipRole__basicMember: 'Miembro',
   membershipRole__guestMember: 'Invitado',
+  passwordComplexity: {
+    sentencePrefix: '',
```
raptisj, line 75: @panteliselef I think you can omit these and they will default to `us` no?

cc: @nikosdouvlis

packages/types/src/localization.ts
nikosdouvlis: I agree with @raptisj :) All error-related messages should go under the `unstable_errors` key, as we need to keep them marked as `unstable` until we revamp our error-handling API.

One minor correction: the `unstable_errors` key also localizes FAPI errors by matching the error status code

packages/clerk-js/src/ui/components/UserProfile/PasswordPage.tsx
panteliselef: @raptisj I don't think that this is correct. The default value is false as this "feature" is an opt-in mechanism (we don't want all password fields to have this behaviour).

`required` is a valid html input attribute, whereas complexity is a custom prop

packages/clerk-js/src/ui/elements/PasswordInput.tsx
panteliselef: I will update this :)

packages/localizations/src/es-ES.ts
nikosdouvlis: Yup, they will default to the `us` resource - however Id keep these so its apparent that they are not yet localised.

packages/types/src/localization.ts
panteliselef: These might not be displayed as errors after all. Probably as indications in gray text. Do you think we should still move this somewhere else ?

packages/localizations/src/es-ES.ts
panteliselef: Hmm, I've seen using empty strings in other places as well, so I wanted to follow the pattern. If this prevent the US (default) translation to be displayed, I can remove them.

packages/clerk-js/src/ui/utils/useFormControl.ts
```diff
@@ -41,6 +45,7 @@ export const useFormControl = <Id extends string>(
   const [value, setValueInternal] = React.useState<string>(initialState);
   const [checked, setCheckedInternal] = React.useState<boolean>(opts?.checked || false);
   const [errorText, setErrorText] = React.useState<string | undefined>(undefined);
+  const [_isSuccess, _setSuccess] = React.useState(false);
```
nikosdouvlis, line 48: 🙃 
```suggestion
  const [isSuccessful, setIsSuccessful] = React.useState(false);
```

packages/clerk-js/src/ui/elements/PasswordInput.tsx
```diff
@@ -1,13 +1,48 @@
...
+    {
+      onValidationFailed: (_, errorMessage) => {
+        formControlProps.setError?.(errorMessage);
+      },
+      onValidationSuccess: () => formControlProps.setSuccessful?.(true),
+    },
+  );
+
+  const __internalOnChange = (e: ChangeEvent<HTMLInputElement>) => {
```
nikosdouvlis, line 35: Nit:

```suggestion
  const onChange = (e: ChangeEvent<HTMLInputElement>) => {
     ...
     onChangeProp()
  }
```

packages/clerk-js/src/ui/hooks/usePasswordComplexity.ts
```diff
@@ -0,0 +1,126 @@
...
+    max_length: password.length < maxLength,
+    min_length: password.length >= minLength,
+    require_numbers: /\d/.test(password),
+    require_lowercase: /[a-z]/.test(password),
+    require_uppercase: /[A-Z]/.test(password),
+    require_special_char: /[!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~]/.test(password),
+  };
+};
+
+type ComplexityErrorMessages = {
```
nikosdouvlis, line 26: Is this equivalent? 
```suggestion
type ComplexityErrorMessages = Partial<Record<Omit<PasswordSettingsData, 'disable_hibp' | 'min_zxcvbn_strength' | 'show_zxcvbn'>>, string>
```

packages/clerk-js/src/ui/hooks/usePasswordComplexity.ts
```diff
@@ -0,0 +1,126 @@
...
+  [key in keyof Partial<Omit<PasswordSettingsData, 'disable_hibp' | 'min_zxcvbn_strength' | 'show_zxcvbn'>>]: string;
+};
+
+export const usePasswordComplexity = (
```
nikosdouvlis, line 30: Very good job extracting this as a hook!

packages/clerk-js/src/ui/hooks/usePasswordComplexity.ts
```diff
@@ -0,0 +1,126 @@
...
+  config: Omit<PasswordSettingsData, 'disable_hibp' | 'min_zxcvbn_strength' | 'show_zxcvbn'>,
+  callbacks?: {
+    onValidationFailed?: (validationErrorMessages: ComplexityErrorMessages, errorMessage: string) => void;
+    onValidationSuccess?: () => void;
+  },
+) => {
+  const { min_length, max_length, require_lowercase, require_numbers, require_uppercase, require_special_char } =
```
nikosdouvlis, line 37: How come these are snake_cased? Aren't we parsing them in UserSettings?

packages/clerk-js/src/ui/hooks/usePasswordComplexity.ts
```diff
@@ -0,0 +1,126 @@
...
+    require_special_char: /[!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~]/.test(password),
+  };
+};
+
+type ComplexityErrorMessages = {
+  [key in keyof Partial<Omit<PasswordSettingsData, 'disable_hibp' | 'min_zxcvbn_strength' | 'show_zxcvbn'>>]: string;
+};
+
+export const usePasswordComplexity = (
+  config: Omit<PasswordSettingsData, 'disable_hibp' | 'min_zxcvbn_strength' | 'show_zxcvbn'>,
```
nikosdouvlis, line 31: Nit: In general I feel that keeping the types and the function inputs separated is more readable.
Eg:
```ts
type UsePasswordComplexityConfig = ...
type UsePasswordComplexityCbs = ...
type UsePasswordComplexity = (config: UsePasswordComplexityConfig, cbs: UsePasswordComplexityCbs) => void;

export const usePasswordComplexity: UsePasswordComplexity = (config, callbacks) => {
  const { onValidationFailed = noop, onValidationSuccess = noop } = callbacks || {};
  ...
}
```

packages/clerk-js/src/ui/hooks/usePasswordComplexity.ts
```diff
@@ -0,0 +1,126 @@
...
+  callbacks?: {
+    onValidationFailed?: (validationErrorMessages: ComplexityErrorMessages, errorMessage: string) => void;
+    onValidationSuccess?: () => void;
+  },
+) => {
+  const { min_length, max_length, require_lowercase, require_numbers, require_uppercase, require_special_char } =
+    config;
+
+  const [_password, _setPassword] = useState('');
```
nikosdouvlis, line 40:
```suggestion
  const [password, _setPassword] = useState('');
```

packages/clerk-js/src/ui/hooks/usePasswordComplexity.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  const passwordComplexity = useMemo(() => {
+    return testComplexityCases(_password, {
+      maxLength: max_length,
+      minLength: min_length,
+    });
+  }, [_password, min_length, max_length]);
+
+  const hasPassedComplexity = useMemo(
+    () => Object.keys(failedValidations || {}).length === 0 && !!_password,
```
nikosdouvlis, line 65: Extremely nitpicky: `!!_password && ...`

nikosdouvlis (approved):
Good job @panteliselef !!
Approving this in order to unblock you but please don't forget to move the localization keys as discussed :)

desiprisg:
Couple of small things.
* I think it would be better if we started showing the errors after the first blur. It seems too agressive to show all these errors when someone starts typing for the first time.
* It seems that the 8 character limit is not enforced if the field in DAPi is not updated at least once. That may be a db problem as the value may not be populated.

packages/clerk-js/src/ui/hooks/usePasswordComplexity.ts
panteliselef: Seems like it but then I have to handle the following TS error

```
Type 'Omit ' does not satisfy the constraint 'string | number | symbol'.
```

packages/clerk-js/src/ui/hooks/usePasswordComplexity.ts
panteliselef: It seems that we do. We only to do that for the first level, we do not follow the pattern for all nested objects.

For example
`userSettings.signIn.second_factor` and `userSettings.signUp.allowlist_only`

panteliselef: 👉https://github.com/clerkinc/javascript/blob/8f190fcb72a9c8c21df66842db9bf26543f98bcb/packages/clerk-js/src/core/resources/UserSettings.ts#L56

PR #1042, opened 2023-04-06
fix(clerk-js,types): Fix typescript errors

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -80,6 +80,7 @@ import { CLERK_SATELLITE_URL, CLERK_SYNCED, ERROR_CODES } from './constants';
 import type { DevBrowserHandler } from './devBrowserHandler';
 import createDevBrowserHandler from './devBrowserHandler';
 import {
+  clerkErrorInitFailed,
   clerkMissingDevBrowserJwt,
   clerkMissingProxyUrlAndDomain,
   clerkMissingSignInUrlAsSatellite,
@@ -502,7 +503,7 @@ export default class Clerk implements ClerkInterface {
       session = (this.client.sessions.find(x => x.id === session) as ActiveSessionResource) || null;
     }
 
-    let newSession = session === undefined ? this.session : session;
+    let newSession = session ? session : this.session || null;
 
     // At this point, the `session` variable should contain either an `ActiveSessionResource`
     // or `null`.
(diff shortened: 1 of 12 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -502,7 +503,7 @@ export default class Clerk implements ClerkInterface {
       session = (this.client.sessions.find(x => x.id === session) as ActiveSessionResource) || null;
     }
 
-    let newSession = session === undefined ? this.session : session;
+    let newSession = session ? session : this.session || null;
```
SokratisVidros, line 506: This change might look innocent but I am sure it's not. In ClerkJS we have three states for sessions

A. Session is set -> Signed in
B. Null -> Signed out
C. Undefined -> Transitioning from signed out to signed in 

That is this change might have negative side effects.

SokratisVidros (changes requested):

packages/clerk-js/src/core/clerk.ts
desiprisg: I was influenced by the commend under this. We used to have an `as ActiveSessionResource | null` there but we removed it by accident (we didn't see it caused a typescript error.

desiprisg: https://github.com/clerkinc/javascript/pull/1042/files/ce824451e6cb28619825e3fcefbc154e90695ce5#diff-1952d6b8ac4b5495e1dba57c32bff9ce72d43f83595c1bc8532a76cf31feb753R508

desiprisg: This is causing the error to fail anyway, so we'll take a closer look!

dimkl: I wouldn't change this line of code. 😅

nikosdouvlis: As everyone said, these 2 lines are not equivalent unfortunately! 
We only want to fallback to `this.session` when `session` is undefined, as `undefined` and `null` have a significant meaning for this flow. This is not your fault of course, we need to revisit the `undefined / null / session` approach

nikosdouvlis (commented):
Let's also remove the `( ... || true)` from the `npm run build` script before this gets merged :)

packages/clerk-js/src/core/clerk.ts
panteliselef: Should we add a comment highlighting A,B,C as Sokratis mentioned ?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -685,49 +686,56 @@ export default class Clerk implements ClerkInterface {
       { base: getClerkQueryParam(CLERK_SATELLITE_URL) as string, searchParams },
       { stringify: true },
     );
-    return this.navigate(this.buildUrlWithAuth(backToSatelliteUrl, false));
+    return this.navigate(this.buildUrlWithAuth(backToSatelliteUrl));
```
panteliselef, line 689: 💯

nikosdouvlis (approved):
Excellent work @desiprisg

PR #1047, opened 2023-04-07
deDe should be deDE instead

harrisrobin:
Seems like deDe is wrong, it should be deDE to stay consistent with all the other keys.

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [X] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/packages/localizations/src/de-DE.ts
+++ b/packages/localizations/src/de-DE.ts
@@ -1,6 +1,6 @@
 import type { LocalizationResource } from '@clerk/types';
 
-export const deDe: LocalizationResource = {
+export const deDE: LocalizationResource = {
   socialButtonsBlockButton: 'Weiter mit {{provider|titleize}}',
   dividerText: 'oder',
   formFieldLabel__emailAddress: 'E-Mail-Adresse',
@@ -556,6 +556,5 @@ export const deDe: LocalizationResource = {
 } as const;
 
 
-export const deDE = deDe;
 
 
--- a/packages/localizations/src/de-DE.ts
+++ b/packages/localizations/src/de-DE.ts
@@ -554,3 +554,8 @@ export const deDe: LocalizationResource = {
     numeric: "{{ date | numeric('de-DE') }}",
   },
 } as const;
+
+
+export const deDE = deDe;
+
+
--- a/packages/localizations/src/de-DE.ts
+++ b/packages/localizations/src/de-DE.ts
@@ -554,7 +554,3 @@ export const deDE: LocalizationResource = {
     numeric: "{{ date | numeric('de-DE') }}",
   },
 } as const;
-
-
-
-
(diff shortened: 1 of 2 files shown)


packages/localizations/src/index.ts
```diff
@@ -1,6 +1,6 @@
 export { enUS } from './en-US';
 export { frFR } from './fr-FR';
-export { deDe } from './de-DE';
+export { deDE } from './de-DE';
```
dimkl, line 3: 🔧  This change is consider a breaking change since you change the exports of the package. To avoid breaking those using `import { deDe } from '@clerk/localizations` i would suggest to keep the current localization, add a deprecation warning and export the one with the correct key. Eg
```
/*
 * @deprecated use `deDE` instead
 */
export { deDe } from './de-DE';
export { deDE } from './de-DE';
```

dimkl (commented):
@harrisrobin thank you for the contribution!! Nice work noticing the typo :D

packages/localizations/src/index.ts
harrisrobin: Makes sense! Done.

packages/localizations/src/de-DE.ts
```diff
@@ -554,3 +554,8 @@ export const deDe: LocalizationResource = {
     numeric: "{{ date | numeric('de-DE') }}",
   },
 } as const;
+
+
+export const deDE = deDe;
```
harrisrobin, line 559: @dimkl we also need to re-export `deDe` as `deDE`. Once we deprecate `deDe` we can fix the name properly.

dimkl: We could update the current export since it's internal and use the following in `packages/localizations/src/index.ts`:
```
/*
 * @deprecated use `deDE` instead
 */
export { deDE as deDe } from './de-DE';
export { deDE } from './de-DE';
```
With that we can limit the cleanup changes to be the deprecation warning and the `...deDE as deDe...` line.

harrisrobin: makes sense & good call. done

dimkl (approved):

packages/localizations/src/index.ts
nikosdouvlis: Can't unsee now. 
Thanks for spotting that @harrisrobin :)

PR #1050, opened 2023-04-07
COR-235 Password Stength

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [x] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### zxcvbn
This PR introduces [zxcvbn](https://zxcvbn-ts.github.io/zxcvbn/) integration.

### Bundle size
For this feature we have introduced the `@zxcvbn-ts` package. We decided to use our utility `loadScript` in order to lazy load the library.

We chose `loadScript` over the native dynamic `import()` due to issues with support in mobile browsers.

### Lazy load
As we mentioned, we are lazy loading the library but we do so only when it is absolutely necessary as almost 200KB will have to be downloaded on the user's device

### Localized zxcvbn suggestions and warnings
We can leverage the localization keys of [Zxcvbn-ts](https://zxcvbn-ts.github.io/zxcvbn/) in order to deliver our own translations. This allows us not to depend to localizations from an external library and their language support.
- (Note: ) Currently I am using the exact keys the lib is using as well.

## Demo video
https://user-images.githubusercontent.com/19269911/230662452-f0f525de-cad2-434b-b96a-d9864ced02e8.mov


packages/clerk-js/src/ui/hooks/usePasswordStrength.ts
```diff
@@ -0,0 +1,47 @@
...
+    },
+    [callbacks?.onValidationFailed, callbacks?.onValidationSuccess, min_zxcvbn_strength],
+  );
+  return {
+    getScore,
+    zxcvbnResult,
+  };
+};
+
+export const scoreDescriptions = ['Very weak', 'Weak', 'Average', 'Strong', 'Very strong'];
```
SokratisVidros, line 47: ❓ Do these values appear in the UI? If yes, they should be localizable.

packages/clerk-js/webpack.config.js
```diff
@@ -48,6 +48,16 @@ const common = ({ mode }) => {
     optimization: {
       splitChunks: {
         cacheGroups: {
+          zxcvbnTSCoreVendor: {
```
SokratisVidros, line 51: ⛔ Since we lazy load, these settings should be removed.

SokratisVidros (changes requested):

packages/clerk-js/webpack.config.js
panteliselef: If we do not create a separate cacheGroup, zxcvbn will be included in defaultVendors and we will end up loading the library eagerly.

Am i missing something ?

packages/clerk-js/package.json
```diff
@@ -47,6 +47,9 @@
     "@emotion/cache": "11.10.5",
     "@emotion/react": "11.10.5",
     "@floating-ui/react": "0.19.0",
+    "@zxcvbn-ts/core": "^2.2.1",
+    "@zxcvbn-ts/language-common": "^2.0.1",
```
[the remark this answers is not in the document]
panteliselef: The implementation is using those, why do they need to be removed ?

packages/clerk-js/package.json
```diff
@@ -47,6 +47,9 @@
...
+    "@zxcvbn-ts/language-en": "^2.1.0",
```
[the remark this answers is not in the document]
panteliselef: Yes, "@zxcvbn-ts/language-en" can be removed

packages/clerk-js/webpack.config.js
SokratisVidros: This case is similar to https://github.com/clerkinc/javascript/blob/493ddc19f1fbbbaaa34bdbf5beaa15be9eeb75bc/packages/clerk-js/src/utils/url.ts#L67. Do we want webpack to coordinate the loading or do we want to do it ourselves?

cc: @nikosdouvlis

packages/clerk-js/package.json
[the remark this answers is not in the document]
SokratisVidros: Don't we lazy load them?

packages/clerk-js/webpack.config.js
panteliselef: I see, thank you Sokratis.
They do offer [cdn links](https://zxcvbn-ts.github.io/zxcvbn/guide/getting-started/#as-script-tag) that we can leverage.

packages/clerk-js/package.json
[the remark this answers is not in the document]
panteliselef: Yes we do, but based on this implementation, we need to bundle the dependencies ourselves, so this is needed right ?

If we end up loading them with `loadScript` we can remove these from package.json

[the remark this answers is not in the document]
SokratisVidros: My comments came from the fact that I was thinking that either we will use loadScript or `import(/*webpackIgnore: true*/ 'cdn.jsdelivr.net/path/to/module.js');` so that the package doesn't passes from our services or webpack.

packages/clerk-js/webpack.config.js
SokratisVidros: Great. PTAL https://github.com/clerkinc/javascript/pull/1050#discussion_r1161679217

nikosdouvlis (approved):

packages/clerk-js/src/ui/elements/PasswordInput.tsx
```diff
@@ -1,34 +1,102 @@
...
+      setComplexityError(errorMessage);
     },
-    onValidationSuccess: () => formControlProps.setSuccessful?.(true),
+    onValidationSuccess: () => setComplexitySuccess(),
   });
 
   const onChange = (e: ChangeEvent<HTMLInputElement>) => {
+    // Lazy load `zxcvbn` on interaction
+    if (strengthMeter && show_zxcvbn) {
+      void Promise.all([
```
SokratisVidros, line 82: 💯

packages/localizations/src/en-US.ts
```diff
@@ -73,6 +73,43 @@ export const enUS: LocalizationResource = {
   membershipRole__admin: 'Admin',
   membershipRole__basicMember: 'Member',
   membershipRole__guestMember: 'Guest',
+  zxcvbn: {
```
panteliselef, line 76: Let's move this under unstable errors for now

packages/clerk-js/src/ui/hooks/usePasswordStrength.ts
```diff
@@ -0,0 +1,48 @@
...
+        const fErrors = errors.map(er => t(localizationKeys(`zxcvbn.suggestions.${er}` as any)));
+        if (result.score < min_zxcvbn_strength) {
+          fErrors.unshift(t(localizationKeys('zxcvbn.notEnough')));
+        }
+        onValidationFailed(fErrors, fErrors.join(' '));
+      } else if (result.score >= min_zxcvbn_strength) {
+        onValidationSuccess?.();
+      }
+    },
+    [callbacks?.onValidationFailed, callbacks?.onValidationSuccess, min_zxcvbn_strength],
```
panteliselef, line 42:
```suggestion
    [onValidationFailed, onValidationSuccess, min_zxcvbn_strength],
```

packages/clerk-js/src/ui/elements/PasswordInput.tsx
panteliselef, line 82: extract this to a function outside of the component. `loadZxcvbn`

packages/clerk-js/src/ui/elements/FormControl.tsx
```diff
@@ -37,6 +37,7 @@ type FormControlProps = Omit<PropsOfComponent<typeof Input>, 'label' | 'placehol
   isSuccessful: boolean;
   hasLostFocus: boolean;
   complexity?: boolean;
+  strengthMeter?: boolean;
```
SokratisVidros, line 40: 🔧 This must be a left over.

packages/clerk-js/src/ui/hooks/usePasswordStrength.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const { t } = useLocalizations();
+  const [zxcvbnResult, setZxcvbnResult] = useState<ZxcvbnResult | undefined>(undefined);
+
+  const getScore = useCallback(
+    (zxcvbn: zxcvbnFN) => (password: string) => {
+      const result = zxcvbn(password);
+      setZxcvbnResult(result);
+
+      if (result.feedback.suggestions?.length > 0) {
```
SokratisVidros, line 31:
```suggestion
      if (result?.feedback?.suggestions?.length > 0) {
```

packages/clerk-js/src/ui/elements/PasswordInput.tsx
```diff
@@ -1,34 +1,110 @@
...
+  };
+};
+
+const loadZxcvbn = () => {
+  return Promise.all([
+    // @ts-ignore
+    import(/*webpackIgnore: true*/ 'https://cdn.jsdelivr.net/npm/@zxcvbn-ts/core@2.2.1/+esm'),
+    // @ts-ignore
+    import(/*webpackIgnore: true*/ 'https://cdn.jsdelivr.net/npm/@zxcvbn-ts/language-common@2.0.1/+esm'),
+  ]).then(([core, zxcvbnCommonPackage]) => {
```
nikosdouvlis, line 64: What do we do if loading fails for some reason? Shall we retry or let the FAPI handle validation?

packages/clerk-js/src/ui/elements/PasswordInput.tsx
```diff
@@ -1,34 +1,110 @@
...
-  const { id, onChange: onChangeProp, complexity = false, ...rest } = props;
-  const formControlProps = useFormControl();
+  const { id, onChange: onChangeProp, strengthMeter = false, complexity = false, ...rest } = props;
   const {
     userSettings: { passwordSettings },
   } = useEnvironment();
 
+  const { show_zxcvbn } = passwordSettings;
+
+  const { setStrengthError, setComplexityError, setStrengthSuccess, setComplexitySuccess } = useComplexityOverStrength({
```
nikosdouvlis, line 84: ❓ How come we need 3 separate hooks here? I wonder if the orchestration you need here would become simpler we combined everything into a single hook, avoiding the callbacks. 

Eg:
```tsx
const { password, setPassword, score } = usePassword(settings);
```  

`setPassword` would take care of the orchestration internally

panteliselef: @nikosdouvlis That was one way to go, which crossed my mind. But I was told that we may end up shipping this functionality as part of the public API, and I wanted to keep things separate. 

Do you propose merging usePasswordComplexity and usePasswordStrength

or create a new hook (`usePassword`) that uses both under the hood and does the coordination internally ?

packages/clerk-js/src/ui/elements/PasswordInput.tsx
panteliselef: Hmm, FAPI will not accept any password that does not meet the criteria (once submitted), having said that I don't think we want to make requests to FAPI (on keystroke).

packages/clerk-js/src/ui/elements/PasswordInput.tsx
nikosdouvlis: Would the public hooks need this kind of orchestration?
To be honest, I'd go with the simpler approach for now (usePassword) as I'd like us to start with a design document for any additions to the public api :)
Also  note that this helper is within clerk-js - it cannot be exposed as a toplevel api from clerk/clerk-react directly so we'll probably have to reimplement it at some point.

packages/clerk-js/src/ui/elements/PasswordInput.tsx
```diff
@@ -1,34 +1,133 @@
...
+      if (!strengthMeter) {
+        formControlProps.setSuccessful?.(true);
+      }
+    }, []),
+  };
+};
+
+const loadZxcvbn = () => {
+  return Promise.all([
+    loadScript('https://cdn.jsdelivr.net/npm/@zxcvbn-ts/core@2.2.1/dist/zxcvbn-ts.min.js', {
```
nikosdouvlis, line 79: Nice!

packages/clerk-js/src/ui/elements/PasswordInput.tsx
panteliselef: Screenshot 2023-04-12 at 11 02 59 AM

In any case the FE allows the request to go through with a weak password, FAPI will catch it. But I see your point.

if(library is downloaded) {dontAllowForWeakPasswordsOnSubmit}
if(library has failed download) {allowForAnyPassword > 8 length to be submitted}

packages/clerk-js/src/ui/elements/PasswordInput.tsx
panteliselef: TBH I don't have an idea of what will be the intended use of the public hook(s).

I'll have another look, on how `usePassword` could improve what we have now

msakbar:
https://user-images.githubusercontent.com/6579893/231518090-5bbdfedf-69c0-4737-afcd-e2b4fab61f81.mov

I took some time to quickly do a test and enabled all the settings, some feedback:

- Complexity messages should be comprehensively provided. You can see how after I add a lowercase, it is asking me to provide an uppercase. We need to provide all the error cases simultaneously so they can attempt completion AND then be smart enough to update the error as they type to show the completion. 

- For Strength, "Your password is not strong enough" should be accompanied by a suggestion on how to make it stronger. I believe this comes from zcxvbn directly and we can pipe it through, especially if we're not displaying an actual meter it's really confusing.

- Also, just to be consistent with the change password dialogue, the input field focus ring didn't turn green when the password WAS strong enough & complied, we should make it the same.

msakbar:
Hokay, did some testing and it works really nicely! (complexity testing isn't in this video because I checked and it was smooth!) 

https://user-images.githubusercontent.com/6579893/231624031-40fbdb90-b25b-4528-b8a2-8ccf4cf0ec84.mov

There are just two issues:

1. The enforcement doesn't seem to line up properly. I used https://lowe.github.io/tryzxcvbn/ to generate passwords of various strengths. Even if i set weak/average, it will only allow me to submit strong passwords. 
2. The actual message around these changes on submit (see video), if it doesn't comply with minimum strength, it should always say: 
    - Your password isn't strong enough. {zxcvbn message}.
  
For this specific scenario, we don't need to provide more messages IF the password strength complies with the setting, we can set the password input box to green at that point and let it go through.

That piece will come in the next iteration with the animations.

msakbar:
FTR - here is the complexity testing as well: 

https://user-images.githubusercontent.com/6579893/231624120-b72b2f00-bc24-4d8c-904e-77ad432ece04.mov

desiprisg:
Small comment: 
The `Confirm password` field should remove its error when the passwords match, and not on blur if possible.

panteliselef:
> Hokay, did some testing and it works really nicely! (complexity testing isn't in this video because I checked and it was smooth!)
> 
>  Screen.Recording.2023-04-12.at.9.30.09.PM.mov 
> There are just two issues:
> 
> 1. The enforcement doesn't seem to line up properly. I used https://lowe.github.io/tryzxcvbn/ to generate passwords of various strengths. Even if i set weak/average, it will only allow me to submit strong passwords.
> 2. The actual message around these changes on submit (see video), if it doesn't comply with minimum strength, it should always say:
>    
>    * Your password isn't strong enough. {zxcvbn message}.
> 
> For this specific scenario, we don't need to provide more messages IF the password strength complies with the setting, we can set the password input box to green at that point and let it go through.
> 
> That piece will come in the next iteration with the animations.

@msakbar  Both of these will be resolved once we build a universal way for using the zxcvbn implementation. This can be done in a following PR

SokratisVidros (approved):
💯

PR #1064, opened 2023-04-11
feat(backend): Add types for webhook events

### LINKED ISSUE #982: Provide webhook types
## Package

- [x] `@clerk/types`

## Description

As far as I can tell, Clerk does not provide TypeScript types for Webhook contents. There's a generic/naive implementation [in the Next.js example](https://github.com/clerkinc/clerk-nextjs-examples/blob/main/examples/widget/pages/api/webhooks/user.ts#L49). It would be nice to have something more concrete, and more clear documentation on this point - a type that included all of the possible variations and their contents. The [documentation for webhooks](https://clerk.com/docs/integration/webhooks#payload-structure) is fine but not complete or explicit.

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Add types for webhook events. Usage example :
```
import type { WebhookEvent } from "@clerk/clerk-sdk-node" 

const handler = req => {
  const evt = req.body.evt as WebhookEvent; 
  switch (evt.type) {
    case 'user.created': // this is typed
      evt.data.firstName // this is also typed
  }
}
```

Fixes #982


dimkl (approved):

packages/backend/src/api/resources/JSON.ts
```diff
@@ -309,3 +309,37 @@ export interface DeletedObjectJSON {
   slug?: string;
   deleted: boolean;
 }
+
```
nikosdouvlis, line 312: Can we place these in a webhooks.ts file? 
Also, I realiase that the JSONs exist in the backend pkg, but do we place this in backend or in types?

cc @dimkl

nikosdouvlis (approved):

packages/backend/src/api/resources/JSON.ts
dimkl: Should we use `webhooks.ts` or `webhooks.d.ts` ? Let's extract the types. 
I think it's `backend` specific and depends on the `JSON` types that are only accessible from this package. To move the `Webhook` type to `types` we should also move the `JSON` types, which i think we shouldn't.

nikosdouvlis: > Should we use webhooks.ts or webhooks.d.ts ?
Should be webhooks.ts, TS will generate the d.ts file during the build step. 

> To move the Webhook type to types we should also move the JSON types, which i think we shouldn't.
Let's stick with @clerk/backend though

@desiprisg Let's move these to a webhooks.ts file  and then we're good to go. Thanks both :)

nikosdouvlis: @desiprisg Once this is merged and release, we should also add a "typescript support" section to the webhooks docs here: https://clerk.com/docs/integration/webhooks

PR #1078, opened 2023-04-18
fix(clerk-js): Handle SSO flows in dev instances with URL based session syncing

mzhong9723:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

URL based session synced dev instances don't set a cookie on FAPI. For OAuth2 IDP and future SSO flows, we get around this by appending the dev browser JWT as a query parameter instead of hash. This way, FAPI can retrieve the dev browser jwt from the url and complete the sso flow.

### DIFF
--- a/packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
+++ b/packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
@@ -1,3 +1,4 @@
+import type { BuildUrlWithAuthParams } from '@clerk/types';
 import React from 'react';
 
 import { buildURL } from '../../utils/url';
@@ -16,6 +17,7 @@ import type {
   UserButtonCtx,
   UserProfileCtx,
 } from '../types';
+import { isRedirectForFAPIInitiatedFlow } from './utils';
 
 export const ComponentContext = React.createContext<AvailableComponentCtx | null>(null);
 
@@ -113,13 +115,22 @@ export const useSignInContext = (): SignInContextType => {
     }),
   );
 
-  const afterSignInUrl = clerk.buildUrlWithAuth(
-    extractAuthProp('afterSignInUrl', {
-      ctx,
-      queryParams,
-      displayConfig,
-    }),
-  );
+  const extractedAfterSignInUrl = extractAuthProp('afterSignInUrl', {
+    ctx,
+    queryParams,
+    displayConfig,
+  });
+
+  const buildUrlWithAuthParams: BuildUrlWithAuthParams = {};
+  if (clerk.instanceType !== 'production' && isRedirectForFAPIInitiatedFlow(clerk, extractedAfterSignInUrl)) {
+    buildUrlWithAuthParams.useQueryParam = true;
+
+    if (clerk.session) {
+      navigate(clerk.buildUrlWithAuth(extractedAfterSignInUrl, buildUrlWithAuthParams));
+    }
+  }
+
+  const afterSignInUrl = clerk.buildUrlWithAuth(extractedAfterSignInUrl, buildUrlWithAuthParams);
 
   const navigateAfterSignIn = () => navigate(afterSignInUrl);
 
--- a/packages/clerk-js/src/ui/contexts/utils.ts
+++ b/packages/clerk-js/src/ui/contexts/utils.ts
@@ -13,3 +13,16 @@ export function assertContextExists(contextVal: unknown, providerName: string): 
     clerkCoreErrorContextProviderNotFound(providerName);
   }
 }
+
+export function isRedirectForFAPIInitiatedFlow(clerk: Clerk, redirectUrl: string): boolean {
+  const url = new URL(redirectUrl);
+  const path = url.pathname;
+
+  return clerk.frontendApi === url.host && frontendApiRedirectPaths.includes(path);
+}
+
+const frontendApiRedirectPaths: string[] = [
+  '/oauth/authorize', // OAuth2 identify provider flow
+  '/v1/verify', // magic links
+  '/v1/tickets/accept', // ticket flow
+];
--- a/packages/types/src/clerk.ts
+++ b/packages/types/src/clerk.ts
@@ -293,8 +293,9 @@ export interface Clerk {
    * Decorates the provided url with the auth token for development instances.
    *
    * @param {string} to
+   * @param opts A {@link BuildUrlWithAuthParams} object
    */
-  buildUrlWithAuth(to: string): string;
+  buildUrlWithAuth(to: string, opts?: BuildUrlWithAuthParams): string;
 
   /**
    * Returns the configured url where <SignIn/> is mounted or a custom sign-in page is rendered.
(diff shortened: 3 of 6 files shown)


packages/types/src/clerk.ts
```diff
@@ -475,6 +476,13 @@ export interface ClerkOptions {
   signInUrl?: string;
 }
 
+export type AuthTokenOptions = {
```
SokratisVidros, line 479:
```suggestion
export type BuildUrlWithAuthParams = {
```

SokratisVidros (changes requested):

packages/clerk-js/src/ui/contexts/utils.ts
```diff
@@ -13,3 +13,20 @@ export function assertContextExists(contextVal: unknown, providerName: string):
...
+  const path = url.pathname;
+
+  if (url.host === fapiUrl && ssoRedirectPaths.includes(path)) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+const ssoRedirectPaths: string[] = [
```
SokratisVidros, line 30: This mechanism of "stealing" the dev browser JWT from ClerkJS and redirecting back to FAPI will be generalized for SSO and other redirection flows.

@alex-ntousias Do you feel the approach of hardcoding FAPI paths in ClerkJS can scale or should we consider introducing a new query string parameter that ClerkJS will identify, and will always redirect to it adding the dev browser JWT?

mzhong9723: Yes agreed, we should implement this keeping in mind other SSO and redirection flows. I hard coded the `/oauth/authorize` path to have something simple that worked, but this is also my biggest question for this PR. Happy to change this to a new query string param (or some other solution that is more general), depending on what we think is best here.

packages/types/src/clerk.ts
```diff
@@ -451,6 +452,13 @@ export type HandleOAuthCallbackParams = {
   continueSignUpUrl?: string | null;
 };
 
+export type BuildUrlWithAuthParams = {
+  /**
+   * Controls if dev browser JWT is added as a query param
+   */
+  queryParam?: boolean | null;
```
SokratisVidros, line 459:
```suggestion
  useQueryParam?: boolean | null;
```

SokratisVidros (commented):
Another nit picky naming comment :)

packages/clerk-js/src/ui/contexts/utils.ts
alex-ntousias: Yes, hardcoding the path will work. We'll probably need to add the following:
* `/v1/verify`: this is what magic links are using currently
* `/v1/tickets/accept`: this is the generic one for the ticket flow...ideally all these redirection flows should go via this endpoint, but this is something we can discuss in the future

mzhong9723: Added those paths!

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
```diff
@@ -113,13 +115,22 @@ export const useSignInContext = (): SignInContextType => {
...
-      displayConfig,
-    }),
-  );
+  const extractedAfterSignInUrl = extractAuthProp('afterSignInUrl', {
+    ctx,
+    queryParams,
+    displayConfig,
+  });
+
+  const buildUrlWithAuthOpts: BuildUrlWithAuthParams = {};
```
SokratisVidros, line 124:
```suggestion
  const buildUrlWithAuthParams: BuildUrlWithAuthParams = {};
```

packages/clerk-js/src/ui/contexts/utils.ts
```diff
@@ -13,3 +13,22 @@ export function assertContextExists(contextVal: unknown, providerName: string):
     clerkCoreErrorContextProviderNotFound(providerName);
   }
 }
+
+export function isRedirectForSSOFlow(clerk: Clerk, redirectUrl: string): boolean {
+  const fapiUrl = clerk.publishableKey || clerk.frontendApi;
```
SokratisVidros, line 18: ⛔ The publishableKey is the base64(`pk_(test|live)_${frontendApi}`) value. If not mistaken, `clerk.frontendApi` should be always available. It should be used to be compared with the `url.host`.

packages/clerk-js/src/ui/contexts/utils.ts
```diff
@@ -13,3 +13,22 @@ export function assertContextExists(contextVal: unknown, providerName: string):
...
+
+  const url = new URL(redirectUrl);
+  const path = url.pathname;
+
+  if (url.host === fapiUrl && ssoRedirectPaths.includes(path)) {
```
SokratisVidros, line 23: 🔧 
```suggestion
return clerk.frontendApi === url.host && frontendApiRedirectPaths.includes(path);
```

SokratisVidros (changes requested):

packages/clerk-js/src/ui/contexts/utils.ts
mzhong9723: Hmm the reason I had this logic in the first place is that `frontendApi` is marked deprecated here: https://github.com/clerkinc/javascript/blob/main/packages/types/src/clerk.ts#L62 - is that not the case? For now, I can add parsing for the publishable key, but we can remove it if `frontendApi` will always be available going forward.

SokratisVidros (approved):

packages/clerk-js/src/ui/contexts/utils.ts
mzhong9723: Gotcha, thanks for the clarification!

packages/clerk-js/src/ui/contexts/utils.ts
```diff
@@ -13,3 +13,16 @@ export function assertContextExists(contextVal: unknown, providerName: string):
     clerkCoreErrorContextProviderNotFound(providerName);
   }
 }
+
+export function isRedirectForFAPIInitiatedFlow(clerk: Clerk, redirectUrl: string): boolean {
+  const url = new URL(redirectUrl);
```
desiprisg, line 18: This will throw an error if `redirectUrl` is a relative path.

packages/clerk-js/src/ui/contexts/utils.ts
```diff
@@ -13,3 +13,16 @@ export function assertContextExists(contextVal: unknown, providerName: string):
...
+  const path = url.pathname;
+
+  return clerk.frontendApi === url.host && frontendApiRedirectPaths.includes(path);
+}
+
+const frontendApiRedirectPaths: string[] = [
+  '/oauth/authorize', // OAuth2 identify provider flow
+  '/v1/verify', // magic links
+  '/v1/tickets/accept', // ticket flow
+];
```
dimkl, line 28: 🙃 for object constants prefer using `const variable = [...] as const` to avoid mistakenly mutating them.

PR #1091, opened 2023-04-20
Cor 273 Animate form error and success messages

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description


packages/clerk-js/src/ui/primitives/FormErrorText.tsx
```diff
@@ -10,27 +12,40 @@ const { applyVariants } = createVariants(theme => ({
     willChange: 'transform, opacity, height',
```
desiprisg, line 12: Is this needed?

packages/clerk-js/src/ui/styledSystem/animations.ts
```diff
@@ -32,6 +32,37 @@ const fadeIn = keyframes`
...
+const inAnimation = keyframes`
+  0% {
+    opacity: 0;
+    transform: translateY(-5px);
+    max-height: 0;
+  }
+  100% {
+    opacity: 1;
+    transform: translateY(0px);
+    max-height: 6rem;
```
desiprisg, line 44: Can we have `100%` here instead of `6rem`?

packages/clerk-js/src/ui/styledSystem/animations.ts
```diff
@@ -32,6 +32,37 @@ const fadeIn = keyframes`
...
+  }
+`;
+
+const outAnimation = keyframes`
+  20% {
+    opacity: 1;
+    transform: translateY(0px);
+    max-height: 6rem;
```
desiprisg, line 52: `100%` also?!

packages/clerk-js/src/ui/styledSystem/animations.ts
desiprisg, line 44: Since this is an animation we might want to use for taller elements, should we do something like:
`max-height: var(--clerk-collapse-max-height, 6rem)` ?

packages/clerk-js/src/ui/elements/FormControl.tsx
```diff
@@ -53,6 +64,55 @@ const getInputElementForType = (type: FormControlProps['type']) => {
...
+
+  const getFormTextAnimation = useCallback(
+    (enterAnimation: boolean): ThemableCssProp => {
+      if (prefersReducedMotion) {
+        return {
+          animation: 'none',
+        };
+      }
+      return t => ({
+        animation: `${enterAnimation ? animations.inAnimation : animations.outAnimation} 300ms ${
```
desiprisg, line 78: Can't we use `t.transitionDuration.slower` here?

packages/clerk-js/src/ui/elements/FormControl.tsx
```diff
@@ -53,6 +64,55 @@ const getInputElementForType = (type: FormControlProps['type']) => {
...
+
+  return {
+    getFormTextAnimation,
+  };
+}
+
+const useCalculateErrorTextHeight = () => {
+  const [height, setHeight] = useState(24);
+
+  const calculateHeight = useCallback((element: HTMLElement | null, messageToDisplay: string | undefined) => {
```
desiprisg, line 94: Can we simplify this and use the `offsetHeight` property. If we want to calculate something before it's in the dom, we can always add it with `position:absolute` and `opacity:0` and calculate the values, although I don't think we need that in this case since this receives an element.

panteliselef: I gave this a try:
- When the callback runs you get access to the element, offsetHeight is 0. Interestingly if wait for 100ms inside a setTimeout the offsetHeight is now a proper value i can use. Sadly this delays the animation.

I am not quite sure why this is happening as the element.innerText seems to be the appropriate text value.

packages/clerk-js/src/ui/elements/FormControl.tsx
panteliselef: I hard coded for the following reason:

Inside FormControl, I use `useDelayUnmount` and I pass a specific value of 200ms, I found out through trial and error that the sweet point for the animation duration needs to be 100ms for than the input value in `useDelayUnmount`

Using this `t.transitionDuration.slower` means that a developer can override these values, causing the animation to break

packages/clerk-js/src/ui/styledSystem/animations.ts
panteliselef: This will probably work, but imagine this scenario:

- We define `--clerk-collapse-max-height` to be 10rem
-  There are 2 places in the screen where this animation will run.
-  One item has a shorter height
- 2nd item has 10rem height.
We managed to run the animation as intended for the 2nd item, but now the animation for the 1st item will run faster

packages/clerk-js/src/ui/hooks/useDelayUnmount.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect, useState } from 'react';
+
+export function useDelayUnmount<T>(isMounted: T, delayTime: number) {
```
SokratisVidros, line 3: ❓ Can you please elaborate on this hook?

packages/clerk-js/src/ui/hooks/usePrefersReducedMotion.ts
```diff
@@ -0,0 +1,19 @@
...
+export function usePrefersReducedMotion() {
+  const [prefersReducedMotion, setPrefersReducedMotion] = useState(true);
+  useEffect(() => {
+    const mediaQueryList = window.matchMedia(mediaQueryNoPreference);
+    setPrefersReducedMotion(!window.matchMedia(mediaQueryNoPreference).matches);
+    const listener = (event: MediaQueryListEvent) => {
+      setPrefersReducedMotion(!event.matches);
+    };
+    mediaQueryList.addEventListener('change', listener);
+    return () => {
```
SokratisVidros, line 14: 🙃 Please add a new line for better readability.

packages/clerk-js/src/ui/hooks/usePrefersReducedMotion.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from 'react';
+
+const mediaQueryNoPreference = '(prefers-reduced-motion: no-preference)';
+
+export function usePrefersReducedMotion() {
```
SokratisVidros, line 5: 💯

packages/clerk-js/src/ui/icons/index.ts
```diff
@@ -8,6 +8,7 @@ export { default as ArrowLeftIcon } from './arrow-left.svg';
 export { default as ArrowRightIcon } from './arrow-right.svg';
 export { default as AuthApp } from './auth-app.svg';
 export { default as Clipboard } from './clipboard.svg';
+export { default as CheckCircle } from './check-circle.svg';
```
SokratisVidros, line 11: ❓ Did we pass these SVGs from an SVG optimizer or does webpack do this for us?

packages/clerk-js/src/ui/primitives/FormErrorText.tsx
```diff
@@ -10,27 +12,40 @@ const { applyVariants } = createVariants(theme => ({
...
+      ref={ref}
       variant='smallRegular'
       colorScheme='danger'
       aria-live='polite'
       id={errorMessageId}
-      {...props}
+      {...rest}
       css={applyVariants(props)}
-    />
+    >
```
SokratisVidros, line 43: ❓ These icons affect all error messages across ClerkJS screens. Is this intentional? If yes, we need to make sure that errors are rendered correctly.

packages/clerk-js/src/ui/primitives/FormText.tsx
```diff
@@ -0,0 +1,26 @@
+import { forwardRef } from 'react';
```
SokratisVidros, line 1: ❓ What's the component about? The  <FormText/> naming is a bit generic. Should we maybe rename it to FormInfoText to complete the success, warning triplet?

packages/clerk-js/src/ui/utils/useFormControl.ts
```diff
@@ -60,13 +66,17 @@ export const useFormControl = <Id extends string>(
...
     options: [],
     enableErrorAfterBlur: false,
+    direction: '',
   };
+
   const { translateError } = useLocalizations();
   const [value, setValueInternal] = React.useState<string>(initialState);
   const [checked, setCheckedInternal] = React.useState<boolean>(opts?.checked || false);
   const [errorText, setErrorText] = React.useState<string | undefined>(undefined);
+  const [warningText, setWarningText] = React.useState<string | undefined>(undefined);
```
SokratisVidros, line 76: 🔧 It would be simpler to set the default value of the error and warning texts to `""` and avoid the <string | undefined> type. That would also simplify the following `typeof` checks as you can just do `if (!warningText)`.

packages/clerk-js/src/ui/utils/useFormControl.ts
```diff
@@ -85,13 +100,23 @@ export const useFormControl = <Id extends string>(
...
   };
   const setSuccessful: FormControlState['setSuccessful'] = isSuccess => {
     setErrorText(undefined);
+    setWarningText(undefined);
     setIsSuccessful(isSuccess);
   };
 
+  const setWarning: FormControlState['setWarning'] = warning => {
+    setWarningText(warning || undefined);
+    if (typeof warning !== 'undefined') {
```
SokratisVidros, line 114: Ditto about removing the `undefined` type.

packages/clerk-js/src/ui/utils/useFormControl.ts
```diff
@@ -125,3 +154,74 @@ export const buildRequest = (fieldStates: Array<FormControlStateLike>): Record<s
...
+type DebouncingOption = {
+  hasLostFocus: boolean;
+  warningText: string | undefined;
+  errorText: string | undefined;
+  enableErrorAfterBlur: boolean | undefined;
+  isSuccessful: boolean;
+  isFocused: boolean;
+  direction: string | undefined;
+};
+export const useFormControlFeedback = (
```
SokratisVidros, line 177: ❓ Please elaborate on this hook. How is skipBlur used? Is it for keyboard navigation?

SokratisVidros: On a separate note, since you already have a DebouncingOption object, consider adding the `skipblur` and `delayTime` keys to the opts object.

packages/clerk-js/src/ui/elements/FormControl.tsx
```diff
@@ -53,6 +64,56 @@ const getInputElementForType = (type: FormControlProps['type']) => {
...
+    [prefersReducedMotion],
+  );
+
+  return {
+    getFormTextAnimation,
+  };
+}
+
+const useCalculateErrorTextHeight = () => {
+  const [height, setHeight] = useState(24);
```
SokratisVidros, line 93: ❓ What's this unit? Is it pixels?

packages/clerk-js/src/ui/elements/FormControl.tsx
SokratisVidros: Folks, if I get it right we want to animate height but the final height is still being determined due to the variable text. For such cases, there are three other CSS techniques that can be used:

1. Work with max-height and height: auto
2. Use scale-y
3. Use flexbox

Have we tried these? The max-height one might work for us we don't have very big variations between the hardcoded max-height value we will pick and the actual height.

For more information please refer to https://css-tricks.com/using-css-transitions-auto-dimensions/.

SokratisVidros (changes requested):

packages/clerk-js/src/ui/hooks/useDelayUnmount.ts
panteliselef: So this allow me to delay the unMount of a component in order to allow the "exit" animation to run before the actual unmount happens and the node is removed from the dom

packages/clerk-js/src/ui/icons/index.ts
panteliselef: All I need was copy the generated SVG code from the Figma icons. I will check for further actions

packages/clerk-js/src/ui/elements/FormControl.tsx
panteliselef: Yes

packages/clerk-js/src/ui/elements/FormControl.tsx
panteliselef: I will give these a try, but the animation are already very efficient, the only inefficient part is that we change the height dynamically, this only happens in very rare cases and based on the most common flows a user this may not even get triggered.

packages/clerk-js/src/ui/primitives/FormErrorText.tsx
panteliselef: Yes we want to display this icon in all error messages. In the videos above some different errors can be seen as well.

packages/clerk-js/src/ui/utils/useFormControl.ts
panteliselef: So i needed a way to group the logic for when an error or the so called "direction" should actually become visible to the end user. I didn't want to add this logic in the useFormControl, and I didn't want to add this inside the FormControl component either.

packages/clerk-js/src/ui/primitives/FormText.tsx
panteliselef: This displays a grey label without an icon. As you proposed `FormInfoText` seems to be a more appropriate name

packages/clerk-js/src/ui/utils/useFormControl.ts
panteliselef: `skipBlur` is used for the keyboard navigation as you said.
I can ditch the option as param and use the `const { submittedWithEnter } = useFormState();` directly inside the hook. This way the callers of `useFormControlFeedback` will not have to worry about passing it.

packages/clerk-js/src/ui/elements/FormControl.tsx
panteliselef: I don't see how the above methods will prevent the height only increasing and not decreasing in order to avoid the elements below bouncing when we are changing between neutral/error/success states.

packages/clerk-js/src/ui/primitives/FormText.tsx
panteliselef: It's used for this
image

packages/clerk-js/src/ui/elements/FormControl.tsx
nikosdouvlis: We can make this more apparent by making a type alias, `type Px = number` and typing the hook as such: 
```
  const [height, setHeight] = useState<Px>(24);
```

packages/clerk-js/src/ui/elements/FormControl.tsx
panteliselef: - I could not find a way to override them from the appearance prop.
- t.transitionDuration.slower is currently set to 280ms, the animation still looks pretty ok with this.

I am switching to `t.transitionDuration.slower`

packages/clerk-js/src/ui/primitives/FormErrorText.tsx
panteliselef: This is now optimised

packages/clerk-js/src/ui/icons/index.ts
panteliselef: This is now optimized

packages/clerk-js/src/ui/elements/FormControl.tsx
```diff
@@ -53,6 +64,57 @@ const getInputElementForType = (type: FormControlProps['type']) => {
...
+      if (prefersReducedMotion) {
+        return {
+          animation: 'none',
+        };
+      }
+      return t => ({
+        animation: `${enterAnimation ? animations.inAnimation : animations.outAnimation} ${
+          t.transitionDuration.$slower
+        } ${t.transitionTiming.$common}`,
+        transition: `height ${t.transitionDuration.$slow}  ease`, // This is expensive but required for a smooth layout shift
```
SokratisVidros, line 81:
```suggestion
        transition: `height ${t.transitionDuration.$slow} ease`, // This is expensive but required for a smooth layout shift
```

SokratisVidros: ❓ Should easing be part of the `t.transitionDuration` interface so that it's customizable by user?

packages/clerk-js/src/ui/hooks/useDelayUnmount.ts
SokratisVidros: Thanks for the extra information. Consider naming the hook to `useDelayedUnmount`

packages/clerk-js/src/ui/elements/FormControl.tsx
```diff
@@ -53,6 +64,57 @@ const getInputElementForType = (type: FormControlProps['type']) => {
...
+  };
+}
+
+type Px = number;
+const useCalculateErrorTextHeight = () => {
+  const [height, setHeight] = useState<Px>(24);
+
+  const calculateHeight = useCallback((element: HTMLElement | null, messageToDisplay: string | undefined) => {
+    if (element) {
+      const fontSize = parseInt(getComputedStyle(element).fontSize.replace('px', ''));
```
SokratisVidros, line 98: ⛔ Please invoke `getComputedStyle` once for the same element, as it [causes layout thrashing](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#calling-getcomputedstyle).

packages/clerk-js/src/ui/elements/FormControl.tsx
```diff
@@ -53,6 +64,57 @@ const getInputElementForType = (type: FormControlProps['type']) => {
...
+      const width = parseInt(getComputedStyle(element).width.replace('px', ''));
+      const lineHeight = parseInt(getComputedStyle(element).lineHeight.replace('px', '')) / 16;
+      const characters = messageToDisplay?.length || 0;
+
+      setHeight(prevHeight => {
+        const newHeight = 10 + fontSize * lineHeight * Math.ceil(characters / (width / (fontSize * 0.6))); //0.6 is an average character width
```
SokratisVidros, line 104: ❓ What does `10` represent?

packages/clerk-js/src/ui/elements/FormControl.tsx
SokratisVidros: I'd also expect us using requestAnimationFrame.

packages/clerk-js/src/ui/hooks/useDebounce.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from 'react';
+
+export function useDebounce<T>(value: T, delay?: number): T {
```
SokratisVidros, line 3: 🔧 Consider renaming this hook to useSetTimeout as it's not strictly about debouncing user input.

SokratisVidros (changes requested):

packages/clerk-js/src/ui/elements/FormControl.tsx
```diff
@@ -73,129 +135,237 @@ export const FormControl = forwardRef<HTMLInputElement, FormControlProps>((props
...
+
+  const errorMessage = useDelayUnmount(debouncedState.errorText, 200);
+  const _successMessage = debouncedState.isSuccessful
+    ? t(localizationKeys('unstable__errors.zxcvbn.goodPassword'))
+    : '';
+  const successMessage = useDelayUnmount(_successMessage, 200);
+  const directionMessage = useDelayUnmount(debouncedState.direction, 200);
+  const warningMessage = useDelayUnmount(debouncedState.warningText, 200);
+
+  const messageToDisplay = directionMessage || successMessage || errorMessage || warningMessage;
```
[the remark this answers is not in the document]
panteliselef: But we need each message to be managed separately so that we know which message to animate out and which to animate in.

packages/clerk-js/src/ui/elements/FormControl.tsx
panteliselef: That refers to the 10px margin-top a form field text has by default

packages/clerk-js/src/ui/elements/FormControl.tsx
panteliselef: We will get an new reference when an element mounts, unmounts and mounts again. 
So I'd propose to simply use a useRef and store the font size there for all elements. This will ensure the getComputedStyle is called once. 
Obviously this works under the assumption that all elements will share the same styles which in our case it does.

panteliselef:
> Wouldn't you avoid changing the height dynamically if we've set a max-height of about 2.5 * line-height for all the error messages and animate based on that? (Technique number 1 in my previous message).

Currently I'm not animating the height when a message is animated in or out. I calculate that height and I update it when a short message is already displayed and a longer one is about to enter and it needs more space. Also technique 1 does not work if the items inside the box that you try to animate, have position absolute.

dimkl (approved):

packages/clerk-js/src/ui/hooks/useSetTimeout.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from 'react';
+
+export function useSetTimeout<T>(value: T, delay?: number): T {
+  const [debouncedValue, setDebouncedValue] = useState<T>(value);
```
SokratisVidros, line 4: 🙃 

```suggestion
  const [timer, setTimer] = useState<T>(value);
```

packages/clerk-js/src/ui/utils/useFormControl.ts
```diff
@@ -85,13 +100,23 @@ export const useFormControl = <Id extends string>(
     setErrorText(translateError(error || undefined));
     if (typeof error !== 'undefined') {
       setIsSuccessful(false);
+      setWarningText('');
     }
   };
   const setSuccessful: FormControlState['setSuccessful'] = isSuccess => {
     setErrorText(undefined);
+    setWarningText('');
```
SokratisVidros, line 108: ❓ Why is warning text set to `""` but error text to `undefined`? I'd expect them to be aligned.

packages/clerk-js/src/ui/utils/useFormControl.ts
```diff
@@ -125,3 +154,85 @@ export const buildRequest = (fieldStates: Array<FormControlStateLike>): Record<s
...
+    hasLostFocus = false,
+    errorText = '',
+    warningText = '',
+    enableErrorAfterBlur = false,
+    isSuccessful = false,
+    isFocused = false,
+    informationText = '',
+  } = opts;
+
+  const canDisplayFeedback = useMemo(() => {
```
SokratisVidros, line 192: ❓ Is it necessary to use a `useMemo`? If not, let's just inline them.

packages/clerk-js/src/ui/utils/useFormControl.ts
SokratisVidros: 🔧 @panteliselef Please merge skipBluer and delay into the `DebouncingOption` interface.

packages/clerk-js/src/ui/hooks/useDelayedUnmount.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect, useState } from 'react';
+
+export function useDelayedUnmount<T>(isMounted: T, delayTime: number) {
```
SokratisVidros, line 3:
```suggestion
export function useDelayedUnmount<T>(isMounted: T, delayInMs: number) {
```

SokratisVidros (changes requested):

SokratisVidros (changes requested):
Last round :)

packages/clerk-js/src/ui/utils/useFormControl.ts
panteliselef: @SokratisVidros This is something that was there initially, which is yeah I also followed that pattern when I started adding more state hooks.

I will align these so they only accept strings

SokratisVidros (approved):

PR #1098, opened 2023-04-21
Add Environment-level and ClerkProvider-level path configuration

anagstef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This PR adds the functionality to add the following path configuration props on the ClerkProvider: `signInUrl`, `signUpUrl`, `afterSignInUrl`, and `afterSignUpUrl`. These props allow users to handle the whole path configuration in one place in their codebase.
Example:

```
<ClerkProvider
	signInUrl="/sign-in"
	signUpUrl="/sign-up"
	afterSignInUrl="/"
	afterSignUpUrl="https://example.com/enroll"
>
   <MyApp />
</ClerkProvider>
```

In Next.js, the following environment variables can be used instead to provide better DX:
- `NEXT_PUBLIC_CLERK_SIGN_IN_URL`
- `NEXT_PUBLIC_CLERK_SIGN_UP_URL`
- `NEXT_PUBLIC_CLERK_AFTER_SIGN_IN_URL`
- `NEXT_PUBLIC_CLERK_AFTER_SIGN_UP_URL`

Example in `.env.local` file:
```
NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=pk_test_...
CLERK_SECRET_KEY=sk_test_...
NEXT_PUBLIC_CLERK_SIGN_IN_URL=/sign-in
NEXT_PUBLIC_CLERK_SIGN_UP_URL=/sign-up
NEXT_PUBLIC_CLERK_AFTER_SIGN_IN_URL=/
NEXT_PUBLIC_CLERK_AFTER_SIGN_UP_URL=/
```


packages/clerk-js/src/utils/authPropHelpers.ts
```diff
@@ -1,48 +1,48 @@
...
-    (typeof queryParams.redirect_url === 'string' ? queryParams.redirect_url : null) ||
-    ctx[key] ||
-    ctx.redirectUrl ||
-    displayConfig[key];
+    (accessRedirectUrl && typeof queryParams?.redirect_url === 'string' ? queryParams?.redirect_url : null) ||
+    ctx?.[key] ||
+    (accessRedirectUrl ? ctx?.redirectUrl : null) ||
+    options?.[key] ||
+    displayConfig?.[key] ||
+    undefined;
```
dimkl, line 45: 🙃 why do we need `undefined` as default ? Also what do you think about replacing the ternary operator conditions with `&&` ? eg:
```
(
   (accessRedirectUrl && queryParams?.redirect_url) || 
   ctx?.[key] ||
   (accessRedirectUrl && ctx?.redirectUrl) ||
   options?.[key] ||
   displayConfig?.[key] ||
   undefined
 )

**Review by dimkl — APPROVED** [review]

File: packages/clerk-js/src/core/clerk.ts [hunk 2]
```diff
@@ -1324,6 +1312,24 @@ export default class Clerk implements ClerkInterface {
     });
   };
 
+  #buildSignInOrUpUrl = (key: 'signInUrl' | 'signUpUrl', options?: RedirectOptions): string => {
+    const opts: RedirectOptions = {
+      afterSignInUrl: extractAuthPathByPriority('afterSignInUrl', { ctx: options, options: this.#options }, false),
+      afterSignUpUrl: extractAuthPathByPriority('afterSignUpUrl', { ctx: options, options: this.#options }, false),
```
dimkl, line 1318: ❓ shall we rename `options?` param to `context?` ? I think it's a little confusing that `{ ctx: options, options: this.#options }` we have 2 options here.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1324,6 +1312,24 @@ export default class Clerk implements ClerkInterface {
     });
   };
 
+  #buildSignInOrUpUrl = (key: 'signInUrl' | 'signUpUrl', options?: RedirectOptions): string => {
```
SokratisVidros, line 1315:
```suggestion
  #buildUrl = (key: 'signInUrl' | 'signUpUrl', options?: RedirectOptions): string => {
```

packages/clerk-js/src/utils/authPropHelpers.ts
```diff
@@ -1,48 +1,48 @@
...
  *
  * The priority is:
- * Querystring parameters > ClerkJS component props > Display configuration payload
+ * 1. Querystring parameters
+ * 2. ClerkJS component props
+ * 3. ClerkProvider props
+ * 4. Display configuration payload
  */
-export const extractAuthProp = (
+export const extractAuthPathByPriority = (
```
SokratisVidros, line 30:
```suggestion
export const pickRedirectionProp = (
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1324,6 +1312,24 @@ export default class Clerk implements ClerkInterface {
...
+    const opts: RedirectOptions = {
+      afterSignInUrl: extractAuthPathByPriority('afterSignInUrl', { ctx: options, options: this.#options }, false),
```
SokratisVidros, line 1317: 🔧 We should consider moving these to a redirection service that computes these values once and then has getters that are used to get the computed values.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1324,6 +1312,24 @@ export default class Clerk implements ClerkInterface {
...
+      afterSignUpUrl: extractAuthPathByPriority('afterSignUpUrl', { ctx: options, options: this.#options }, false),
+      redirectUrl: options?.redirectUrl || window.location.href,
+    };
+    if (!this.#environment || !this.#environment.displayConfig) {
```
SokratisVidros, line 1321: ❓ Should we replace this condition with an early guard (first check in the function) checking if Clerk.isLoaded?

SokratisVidros (changes requested):

SokratisVidros:
Yes, that's right. It's the safest approach, though. Let me know if you discussed something else with the team already.

packages/clerk-js/src/core/clerk.ts
anagstef: This is a fair point, although the available values are different depending on where this redirection service is called. For example, inside the `useSignInContext` hook, we can access query params, SignIn component props, ClerkProvider-level props and displayConfig. On the other hand, in `clerk.ts` we only have access to ClerkProvider props, displayConfig and maybe `RedirectOptions` from some components. So the computation maybe cannot be done once for all.

I can document where we use each of these, so that we can come up with a better solution that centralizes the whole logic for redirections.

packages/clerk-js/src/utils/authPropHelpers.ts
anagstef: @dimkl Do you think replacing with `&&` will be easier to read?

dimkl: (personal opinion) yes but leave it as is?

nikosdouvlis (approved):

PR #1101, opened 2023-04-24
feat: add signup/signin redirect helper

dimkl:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description


packages/backend/src/redirect.ts
```diff
@@ -0,0 +1,24 @@
+type RedirectAdapter = (url: string, response?: any) => void;
+
+type SingUpParams = { signUpUrl: string; url: string; response?: any };
+type SingInParams = { signInUrl: string; url: string; response?: any };
+
+export class RedirectHelper {
```
SokratisVidros, line 6: 🙃 I'd suggest using JS POJOs instead of TS classes, as they add unnecessary bloat.

cc: @nikosdouvlis

packages/backend/src/redirect.ts
```diff
@@ -0,0 +1,24 @@
...
+  toSignup({ signUpUrl, url, response }: SingUpParams): void {
+    this.#redirectAdapter(this.#buildPath(signUpUrl, url), response);
+  }
+
+  toSignIn({ signInUrl, url, response }: SingInParams): void {
+    this.#redirectAdapter(this.#buildPath(signInUrl, url), response);
+  }
+
+  #buildPath(destinationUrl: string, currentUrl: string) {
+    return `${destinationUrl}#/?redirect_url=${currentUrl}`;
```
SokratisVidros, line 22: ❓ Since we operate on the server context, I expect this to be path-based, not hash-based. Am I missing something?

SokratisVidros (changes requested):

packages/backend/src/redirect.ts
dimkl: The only reasoning behind class instead of POJOs is to provide an API that you can define once how to the redirect works (parameter in constructor) and use it either for signup or singin redirections.
By changing it to JS POJOs i would have to add another parameter to each function (`toSignUp`/ `toSignIn`) for the redirectAdapter. Are we okay with that?

packages/backend/src/redirect.ts
dimkl: I followed the description of the ticket. Let's take it offline and came back to it with the changes.

packages/backend/src/redirect.ts
SokratisVidros: An alternative option is to leverage closures

```
function (options...) {
  // closure vars
  return {
    redirectToSignIn,
    redirectToSignUp
  }
}
```

dimkl: Refactored as suggested. Could you take another look?

packages/backend/src/redirect.ts
```diff
@@ -0,0 +1,20 @@
+type RedirectAdapter = (url: string, response?: any) => void;
+
+type SingUpParams = { signUpUrl: string; url: string; response?: any };
+type SingInParams = { signInUrl: string; url: string; response?: any };
+
+const buildPath = (destinationUrl: string, currentUrl: string) => {
+  return `${destinationUrl}#/?redirect_url=${currentUrl}`;
```
anagstef, line 7: ❓ Maybe this url should use URL utilities, because of encoding, etc. ?

dimkl:
@SokratisVidros , @anagstef please proceed with reviewing the PR.

PR #1102, opened 2023-04-24
feat(clerk-js): Use path-based routing as default when repo-level paths exist

anagstef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

If `signInUrl` or `signUpUrl` are set in the repo-level, we can default to path-based routing for the SignIn or SignUp components respectively.

This PR depends on #1098

### DIFF
--- a/packages/nextjs/src/client/index.tsx
+++ b/packages/nextjs/src/client/index.tsx
@@ -1,10 +1,16 @@
-import { __internal__setErrorThrowerOptions, ClerkProvider as ReactClerkProvider } from '@clerk/clerk-react';
+import {
+  __internal__setErrorThrowerOptions,
+  ClerkProvider as ReactClerkProvider,
+  SignIn as BaseSignIn,
+  SignUp as BaseSignUp,
+} from '@clerk/clerk-react';
 import type { IsomorphicClerkOptions } from '@clerk/clerk-react/dist/types';
-import type { MultiDomainAndOrProxy, PublishableKeyOrFrontendApi } from '@clerk/types';
+import type { MultiDomainAndOrProxy, PublishableKeyOrFrontendApi, SignInProps, SignUpProps } from '@clerk/types';
 import { useRouter } from 'next/router';
 import React from 'react';
 
 import { invalidateNextRouterCache } from './invalidateNextRouterCache';
+import { ClerkNextProvider, useClerkNextContext } from './NextProviderContext';
 
 // TODO: Import from shared once [JS-118] is done
 const useSafeLayoutEffect = typeof window !== 'undefined' ? React.useLayoutEffect : React.useEffect;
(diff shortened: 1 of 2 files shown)


anagstef:
I have only one note about this. The `signInUrl` exists at repo-level (`ClerkProvider`, or environment variable) for some time now (a month or so), accompanying the Multi-domain feature. This may cause breaking changes to users using this feature and the default hash-based routing.

(cc @nikosdouvlis @SokratisVidros )

SokratisVidros (approved):

SokratisVidros:
@anagstef This is a valid concern. Multi-domain has yet to be widely used, so we have some room for adjustments. As you said, this PR will force path-based routing for the multi-domain sign-in component. Two approaches:

1. Keep it as we would like to promote path-based routing in the future.
2. Enforce path-based routing only when both SIGN_IN_URL and SIGN_UP_URL are defined in ClerkProvider.
3. Do not enforce path routing if proxyUrl, or domain props are set. These props are used in the Multi-domain feature.

For now my suggestion is to go with 1 until an issue is raised.

packages/nextjs/src/client/index.tsx
```diff
@@ -67,26 +101,29 @@ export function ClerkProvider({ children, ...rest }: NextClerkProviderProps): JS
     };
   }, []);
 
+  const nextProps = {
+    frontendApi: frontendApi || process.env.NEXT_PUBLIC_CLERK_FRONTEND_API || '',
```
SokratisVidros, line 105: 🔧 We can remove the frontendApi since it's consider deprecated

SokratisVidros (approved):

packages/nextjs/src/client/index.tsx
anagstef: @SokratisVidros  I think this will be a breaking change for people still using `frontendApi` and having a caret  (`^`)  in their `package.json` for `@clerk/nextjs`. After releasing this, their next build will break.

SokratisVidros: Can we mark it as deprecated then?

PR #1103, opened 2023-04-24
feat(clerk-js,types): Introduce basic types & resources for SAML support

yourtallness:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [X] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This introduces basic types & resources that we'll be using for the upcoming SAML support.

New types & properties are marked as experimental until the feature is ready for GA.


packages/types/src/signUp.ts
```diff
@@ -114,6 +115,12 @@ export type PrepareVerificationParams =
       strategy: OAuthStrategy;
       redirectUrl?: string;
       actionCompleteRedirectUrl?: string;
+    }
+  | {
+      strategy: SamlStrategy;
+      samlIdentifier: string;
+      redirectUrl?: string;
+      actionCompleteRedirectUrl?: string;
     };
```
dimkl, line 124: ❓ isn't this like `SamlConfig` type extended with a `samlIdentifier`?

dimkl (approved):

packages/types/src/redirects.ts
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Whether to continue (i.e. PATCH) an existing SignUp (if present) or create a new SignUp.
+   */
+  continueSignUp?: boolean;
+
+  /**
+   * One of the supported OAuth providers you can use to authenticate with, eg 'oauth_google'.
+   * Or alternatively `saml`, to authenticate with SAML.
+   */
+  strategy: OAuthStrategy | SamlStrategy;
```
SokratisVidros, line 25: 🔧 Let's add more type safety here with conditional types. When strategy equals `SamlStrategy` the `identifier` key will be required. All the other keys, are available for both strategies.

packages/types/src/redirects.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Identifier that can be used to select the target SAML connection for this request,
+   * in the case of the `saml` strategy.
+   *
+   * @experimental
+   */
+  samlIdentifier?: string;
```
SokratisVidros, line 33: ✂️ As discussed offline, this will be removed.

packages/types/src/signUp.ts
```diff
@@ -154,7 +161,13 @@ export type SignUpCreateParams = Partial<
     externalAccountStrategy: string;
     externalAccountRedirectUrl: string;
     externalAccountActionCompleteRedirectUrl: string;
-    strategy: OAuthStrategy | TicketStrategy;
+    strategy: OAuthStrategy | SamlStrategy | TicketStrategy;
+
+    /**
+     * @experimental
+     */
+    samlIdentifier: string;
```
SokratisVidros, line 169: ✂️

SokratisVidros (changes requested):

packages/types/src/signUp.ts
yourtallness: Yes, but since we're not DRYing this up for the oauth case either, I preferred to be consistent.

packages/types/src/redirects.ts
yourtallness: Discussed this with @nikosdouvlis.

While discriminated union types are nice for strictness, they make consumption e.g. in `authenticateWithRedirect` very tricky, because props can no longer easily be destructured.

packages/types/src/signUp.ts
yourtallness: Removed

packages/types/src/redirects.ts
nikosdouvlis: This makes sense to me, its a discussion of type strictness versus ease of use.
@yourtallness can you verify that FAPI returns an appropriate/readable error if `identnfier` is not passed in?
Also, if you have any examples from our discussion that demonstrate the issues caused by the extra strictness, could you please share them for future reference?

nikosdouvlis (approved):

packages/types/src/redirects.ts
yourtallness: @nikosdouvlis Checked that `identifier` & `email_address` respectively are not submitted when not provided, so we should be OK.

yourtallness: Will get back to you with examples with regard to discriminated union types.

PR #1108, opened 2023-04-26
Feat(backend, nextjs): introduce redirectToSignIn / redirectToSignup helper

dimkl:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Introduced `redirectToSignIn` / `redirectToSignup` helper in `@clerk/backend` package and a wrapper of those in `@clerk/nextjs` to support a better DX for redirecting to sing-in and sign-up pages.

The new NextJS helpers are gonna work as :
- redirect to sign-in/sign-up provided by NEXT_PUBLIC_CLERK_SIGN_{UP|IN}_URL  env variables
- redirect to hosted pages sign-in/sign-up as fallback

Some key notes:
- `redirectBackUrl` is optional

## Notes

Review it per commit


dimkl:
Ignore the failing tests. They are related to the addition of `NEXT_PUBLIC_CLERK_SIGN_UP_URL` env which will be also added by another PR from Stefanos before this PR is merged.

anagstef (approved):

packages/backend/src/redirect.ts
```diff
@@ -0,0 +1,67 @@
+import { parsePublishableKey } from './util/parsePublishableKey';
+
+type RedirectAdapter = (url: string) => any;
+
+type SingUpParams = { url?: string; returnBackUrl?: string };
+type SingInParams = { url?: string; returnBackUrl?: string };
+
+const buildPath = (destinationUrl: string, currentUrl?: string) => {
```
SokratisVidros, line 8: ⛔ This function build the full sign-(in|up)-url with the redirect_url parameter. The `buildPath` name is a bit misleading. Can we revise it?

Same for destinationUrl and currentUrl. The currentUrl is the redirectUrl.

packages/backend/src/redirect.ts
```diff
@@ -0,0 +1,67 @@
...
+type RedirectParams = {
+  redirectAdapter: RedirectAdapter;
+  signInUrl?: string;
+  signUpUrl?: string;
+  publishableKey?: string;
+  frontendApi?: string;
+};
+
+export function redirect({ redirectAdapter, signUpUrl, signInUrl, frontendApi, publishableKey }: RedirectParams) {
+  const accountsBaseUrl = getAccountsBaseUrl(frontendApi, publishableKey);
```
SokratisVidros, line 37: 🙃 It's more reasonable to do the following:

1. Check if frontendApi is passed
2. If yes, pass it to buildAccountsBaseUrl(frontendApi)
3. If not passed, check if publishableKey is passed
4. If yes, parse it and pass it to buildAccountsBaseUrl(frontendApi)
5. Otherwise the accountsBaseUrl should be "" so that you can easily throw an error later if signIn or signUp URLs are missing.

packages/backend/src/redirect.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const redirectToSignUp = ({ returnBackUrl }: SingUpParams = {}) => {
+    if (!signUpUrl && !accountsBaseUrl) {
+      throw new Error('signUpUrl or frontendApi/publishableKey should be provided!');
```
SokratisVidros, line 41: ❓ Do we have a format for errors thrown from @clerk/backend?

packages/backend/src/redirect.ts
```diff
@@ -0,0 +1,67 @@
+import { parsePublishableKey } from './util/parsePublishableKey';
```
SokratisVidros, line 1: 🙃 Nit picky one. Let's rename the module to `redirections.ts`

SokratisVidros (changes requested):

packages/backend/src/redirect.ts
dimkl: suggested name? is `buildUrl` a better one?

packages/backend/src/redirect.ts
dimkl: as far as know, we don't:
- `packages/backend/src/util/assertValidSecretKey.ts#L4`
- `packages/backend/src/tokens/request.ts#L89`

packages/backend/src/redirect.ts
dimkl: I will refactor it. Are you also suggesting to move the check that `signUpUrl` / `sigInUrl` or `frontentApi`/`publishableKey` before the execution of the `redirectToSignIn` / `redirectToSignUp` and enforce that either both `signUpUrl` and `signInUrl` or `frontentApi`/`publishableKey` should be defined ?

packages/backend/src/redirect.ts
SokratisVidros: Yes

packages/backend/src/redirect.ts
SokratisVidros: The keys should exist at this point but we can rethrow the missing key error from the errorthrower to be on the safe side

SokratisVidros (approved):

packages/backend/src/redirect.ts
dimkl: i'm gonna keep the error being thrown upon helper execution and change the error message to be the same as `errorThrower.throwMissingPublishableKeyError()` from `@clerk/shared`. We currently do not have `@clerk/shared` as dependency due to some bundling issues so i will leave a TODO to replace that errorMessage with the errorThrower when `@clerk/shared` is added.

PR #1122, opened 2023-04-28
Add ISO locale in @clerk/localizations + Intl.ListFormat

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Example of the new structure for our Localization resources
```typescript
const enUS = {
 locale: 'en-US',
 resources: {
      ... // All localized key value pairs inside here
 }
}
```

### Intl.ListFormat
Used Intl.ListFormat web api to localize conjuctions in a sentence. At the moment this is needed for the Password complexity error messages.

`Intl.ListFormat().format` accepts a locale. We should *always* provide this locale based on the language resource `@clerk/localization`. The following example indicates which situation we try to avoid

If undefined is passed to `Intl.ListFormat().format` the browser's default locale will be used which may cause inconsistencies.  Let's say a Developer using Clerk components sets the lanugage to always be english. A  greek user is visiting the website and has the default  browser  language settings to greek as well. He/shed will end up seeing a label like
"Music, food και wine" instead of "Music, food, and wine" 

### Changes in structure & types
This PR introduces breaking changes as the it is changing the types and the structure for each language object inside `@clerk/localizations`

For clerk-js, we ensure backwards compatibility with the following snippet. Old versions of @clerk/localization will still work for latest changes of clerk-js.
```typescript
const migrateResources = (lresource: LocalizationResource | undefined): LocalizationResources => {
  return 'resources' in ((lresource as any) || {})
    ? lresource?.resources || {}
    : (lresource as LocalizationResources) || {};
};
```
This is mandatory as we hot load the latest clerk-js versions in browsers and the passed translations may have not beed updated by the developers yet.

**Updated types**
Now we support the old and the new structure
```typescript
  ...
  localization?: LocalizationResource | LocalizationResources;
  ...
}
```


packages/clerk-js/src/ui/hooks/usePasswordComplexity.ts
```diff
@@ -75,7 +76,13 @@ export const usePasswordComplexity = (config: UsePasswordComplexityConfig, callb
...
     (failedValidations: ComplexityErrorMessages) => {
-      const messageWithPrefix = Object.values(failedValidations).join(', ');
+      let messageWithPrefix: string;
+      if (canUseListFormat(locale)) {
+        const formatter = new Intl.ListFormat(locale, { style: 'long', type: 'conjunction' });
+        messageWithPrefix = formatter.format(Object.values(failedValidations));
+      } else {
+        messageWithPrefix = Object.values(failedValidations).join(', ');
+      }
       return `${t(localizationKeys('unstable__errors.passwordComplexity.sentencePrefix'))} ${messageWithPrefix}`;
```
anagstef, line 86: 💡  I just noticed this one. Isn't this somewhat inconsistent because it produces the final string out of concatenation between a `localizationKey` and the `failedValidations` values? But I guess this is the only way, due to the dynamic list of failed validations, we cannot pass the values to the localization files, right? 🤔

panteliselef:
Closing in favour of #1150

PR #1152, opened 2023-05-04
fix(clerk-js): Redirect for FAPI initiated flows if user is signed in

mzhong9723:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

For flows like OAuth2 IDP in dev instances that use URL based session syncing, we need to add the dev session jwt as a query parameter so FAPI can recognize it on redirect (i.e. redirect to `/oauth/authorize`). If the user is signed in, we perform the redirect with this added query param.

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -67,6 +67,7 @@ import {
   isAccountsHostedPages,
   isDevOrStagingUrl,
   isError,
+  isRedirectForFAPIInitiatedFlow,
   noOrganizationExists,
   noUserExists,
   pickRedirectionProp,
@@ -630,10 +631,7 @@ export default class Clerk implements ClerkInterface {
       return clerkMissingDevBrowserJwt();
     }
 
-    let asQueryParam = false;
-    if (options && options.useQueryParam) {
-      asQueryParam = options.useQueryParam;
-    }
+    const asQueryParam = !!options?.useQueryParam;
 
     return setDevBrowserJWTInURL(toURL.href, devBrowserJwt, asQueryParam);
   }
@@ -1178,6 +1176,10 @@ export default class Clerk implements ClerkInterface {
         // set in updateClient
         this.updateEnvironment(environment);
 
+        if (await this.#redirectFAPIInitiatedFlow()) {
+          return false;
+        }
+
         if (Clerk.mountComponentRenderer) {
           this.#componentControls = Clerk.mountComponentRenderer(this, this.#environment as Environment, this.#options);
         }
@@ -1350,4 +1352,27 @@ export default class Clerk implements ClerkInterface {
       throw new Error('ClerkJS components are not ready yet.');
     }
   }
+
+  #redirectFAPIInitiatedFlow = async (): Promise<boolean> => {
+    const redirectUrl = new URLSearchParams(window.location.search).get('redirect_url');
+    const isProdInstance = this.instanceType === 'production';
+    const shouldRedirect = redirectUrl !== null && isRedirectForFAPIInitiatedFlow(this.frontendApi, redirectUrl);
+
+    if (isProdInstance || !shouldRedirect) {
+      return false;
+    }
+
+    const userSignedIn = this.session;
+    const signInUrl = this.#environment?.displayConfig.signInUrl;
+    const referrerIsSignInUrl = signInUrl && window.location.href.startsWith(signInUrl);
+
+    // don't redirect if user is not signed in and referrer is sign in url
+    if (!userSignedIn && referrerIsSignInUrl) {
+      return false;
+    }
+
+    const buildUrlWithAuthParams: BuildUrlWithAuthParams = { useQueryParam: true };
+    await this.navigate(this.buildUrlWithAuth(redirectUrl, buildUrlWithAuthParams));
+    return true;
+  };
 }


packages/clerk-js/src/core/clerk.ts
```diff
@@ -1173,6 +1171,8 @@ export default class Clerk implements ClerkInterface {
         // set in updateClient
         this.updateEnvironment(environment);
 
+        await this.redirectFAPIInitiatedFlowIfSignedIn();
```
dimkl, line 1174: 🔧 i would expect that if a navigation is trigger to stop the execution of this function. Eg
```
if(await this.redirectFAPIInitiatedFlowIfSignedIn()) return;
```
For the above code to work we should return `true` from the `redirectFAPIInitiatedFlowIfSignedIn` when we trigger navigate.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1345,4 +1345,18 @@ export default class Clerk implements ClerkInterface {
...
   }
+
+  redirectFAPIInitiatedFlowIfSignedIn = async () => {
+    const redirectUrl = new URLSearchParams(window.location.search).get('redirect_url');
+    if (
+      this.instanceType !== 'production' &&
+      redirectUrl !== null &&
+      isRedirectForFAPIInitiatedFlow(this.frontendApi, redirectUrl)
+    ) {
+      if (this.session) {
```
dimkl, line 1356: ❓ why do we also check the `this.session` ? Where do we need it?

SokratisVidros (approved):

packages/clerk-js/src/core/clerk.ts
mzhong9723: If a session is present, does it mean the user is logged in?

For the OAuth 2 IDP flow in particular, if we redirect without an active session, FAPI oauth server will redirect the user back to the sign in URL to sign in, but clerkJS will redirect back to FAPI  `/oauth/authorize` and so on. We end up with an infinite loop of redirects (sign in URL -> `/oauth/authorize` -> sign in URL -> ...). 

Do other flows (magic link, ticket) that require this redirect with dvb query param need presence of a session? If not, we may need to differentiate the logic between oauth idp and the others.

dimkl: > If a session is present, does it mean the user is logged in?
yes
I agree with @mzhong9723 that we should redirect only the user is signedIn to avoid an infinite redirect loop.

> Do other flows (magic link, ticket) that require this redirect with dvb query param need presence of a session? If not, we may need to differentiate the logic between oauth idp and the others.
cc: @SokratisVidros 

Finally we should check if session is populated when this `redirectFAPIInitiatedFlowIfSignedIn` is called ?

SokratisVidros: In ticket and magic links flows, there is no session yet. So redirection can't be conditional. We can stop the redirection flow in FAPI in the two following ways:

1. If the referrer is `sign-in-url`.
2. If we introduce another `__clerk_synced` parameter, ClerkJS will also add to the replayed FAPI URL when it passes from redirectWithAuth.

mzhong9723: @SokratisVidros Could you expand a bit on 2? How would it work for OAuth2 idp vs. ticket and magic link flows?

SokratisVidros: Basically, my suggestion is to add `__clerk_fapi_replayed=true` parameter [here](https://github.com/clerkinc/javascript/blob/d53432e4a57df64cd0556d53ff20eb0a23f4d3f5/packages/clerk-js/src/utils/devBrowser.ts#L14). Then on FAPI if no session exist, but the `__clerk_fapi_replayed` is present we continue the flow without redirecting again to SignInURL for Clerk.js.

Let me know if this helps.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1345,4 +1347,28 @@ export default class Clerk implements ClerkInterface {
       throw new Error('ClerkJS components are not ready yet.');
     }
   }
+
+  shouldRedirectFAPIInitiatedFlow = async (): Promise<boolean> => {
```
dimkl, line 1351: ❓ @nikosdouvlis should we make this function private or do we consider a method that can be used in custom flows? I think it should be private.
```
#redirectFAPIInitiatedFlow  = async (): Promise<boolean> => {
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1345,4 +1347,28 @@ export default class Clerk implements ClerkInterface {
...
+      ) {
+        return false;
+      }
+
+      const buildUrlWithAuthParams: BuildUrlWithAuthParams = { useQueryParam: true };
+      await this.navigate(this.buildUrlWithAuth(redirectUrl, buildUrlWithAuthParams));
+      return true;
+    }
+
+    return false;
```
dimkl, line 1372: 🙃 i would suggest to use guards instead of big control flow blocks. eg
```typescript
    const prodInstance =  this.instanceType === 'production'
    const shouldRedirect =  redirectUrl !== null && isRedirectForFAPIInitiatedFlow(this.frontendApi, redirectUrl);
    
    if (!prodInstance && shouldRedirect) {
      return false
    }
   
     const userSignedIn = this.session;
     const signInUrl = this.#environment?.displayConfig?.signInUrl
     const comesFromSignInUrl = signInUrl && window.location.href.startsWith(signInUrl);
    // don't redirect if user is not signed in and referrer is sign in url
    if (!userSignedIn && comesFromSignInUrl) {
      return false;
    }
    // .... redirect code 
    return true;
```

dimkl (approved):
Nice work and sorry for all those comments 😅 !!
Some final touches and i think we are good to merge.
Since this part of the code is not tested, i would like to point out that we should perform thorough manual testing in staging since this change will be automatically passed to all our customers (due to hot-loading) when its released in production.

packages/clerk-js/src/core/clerk.ts
mzhong9723: I agree that it should be private too. I will update the function declaration!

mzhong9723:
@dimkl I appreciate the thorough review! I have tested my changes for the oauth2 IDP flow locally, and I will do so in staging after merge as well. Is there any regression test cases I should pay special attention to to verify that these changes don't break any existing functionality?

packages/clerk-js/src/core/clerk.ts
mzhong9723: Updated to use the first solution. We will check if the referrer is the sign in url. If it is and there is no session, we will not redirect and proceed so the user can sign in.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1350,4 +1352,27 @@ export default class Clerk implements ClerkInterface {
       throw new Error('ClerkJS components are not ready yet.');
     }
   }
+
+  #shouldRedirectFAPIInitiatedFlow = async (): Promise<boolean> => {
```
desiprisg, line 1356: Should we rename this since it might trigger a navigation?

mzhong9723: Renamed back to `redirectFAPIInitiatedFlow` which I think makes more sense, lmk what you think!

dimkl:
You could check the following flows :
- Sign-up / Sign-in with creds
- Sign-up / Sign-in with oauth
- Interstitial flow
We will also make various manual tests, so i think we will check if it affects something else.
I think you can proceed with merging after renaming the method 😁.

dimkl:
@panteliselef Could i ping you when this is released on staging to perform to validate for multi domain applications?

PR #1153, opened 2023-05-05
fix(nextjs): Enable authMiddleware testing for sign-in/sign-up urls from env

dimkl:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Add skipped failing test for supporting sign-in and sign-up urls from env and update the code to retrieve the envs each time the middleware tun to support support the testing.

### DIFF
--- a/packages/nextjs/src/server/authMiddleware.ts
+++ b/packages/nextjs/src/server/authMiddleware.ts
@@ -7,7 +7,6 @@ import { NextResponse } from 'next/server';
 import { isRedirect, mergeResponses, paths, setHeader, stringifyHeaders } from '../utils';
 import { withLogger } from '../utils/debugLogger';
 import { authenticateRequest, handleInterstitialState, handleUnknownState } from './authenticateRequest';
-import { SIGN_IN_URL, SIGN_UP_URL } from './clerkClient';
 import { receivedRequestForIgnoredRoute } from './errors';
 import { redirectToSignIn } from './redirect';
 import type { NextMiddlewareResult, WithAuthOptions } from './types';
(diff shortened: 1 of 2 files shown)


packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -176,12 +175,19 @@ const withDefaultPublicRoutes = (publicRoutes: RouteMatcherParam | undefined) =>
   if (typeof publicRoutes === 'function') {
     return publicRoutes;
   }
+
   const routes = [publicRoutes || ''].flat().filter(Boolean);
-  if (SIGN_IN_URL) {
-    routes.push(matchRoutesStartingWith(SIGN_IN_URL));
+  // TODO: refactor it to use common config file eg SIGN_IN_URL from ./clerkClient
```
SokratisVidros, line 180: ❓ Don't we do the same thing in clerkClient.ts?

dimkl: what do you mean?
the comment is a reminder that should replace the usage of `process.env.ENV_KEY` with a more centralized solution which will also allow mocking the environment variable in tests without having to reset the modules.
Currently by using the following part we cannot mock the environment variable (since it's referenced by value in the imported file) unless we reset all modules and import the testing modules in each test.
```typescript
// file: clerkClient.ts
export const SIGN_IN_URL = 'something'

// file: authMiddleware.ts

import { SIGN_IN_URL } from "./clerkClient"
```

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -176,12 +175,19 @@ const withDefaultPublicRoutes = (publicRoutes: RouteMatcherParam | undefined) =>
...
+  // we use process.env for now to support testing
+  const signInUrl = process.env.NEXT_PUBLIC_CLERK_SIGN_IN_URL || '';
```
nikosdouvlis, line 182: I absolutely agree that we need to create a utility to read the env/config values, however I think it's best if we mock the module for now, even if it requires changing the authMiddleware.test.ts, instead of pulling this logic out of the ./clerkClient file

nikosdouvlis (approved):
After reviewing offline, we decided to proceed with the proposed change. 
@dimkl please add a ticket so we can introduce the config helper in an upcoming cycle :)

nikosdouvlis (approved):
After reviewing offline, we decided to proceed with the proposed change. 
@dimkl please add a ticket so we can introduce the config helper in an upcoming cycle :)

PR #1161, opened 2023-05-08
App router improvements

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

- Introduce withLogger and more debug logs for authMiddleware
- passing debug: true to authMiddleware also enables debug for auth() and getAuth()
- split auth and getAuth to improve error messages
- better TRPC support by changing the default matchers to allow for /api and /trpc routes
- improve the default ignoredRoute warning message

### DIFF
--- a/packages/nextjs/src/server/authMiddleware.ts
+++ b/packages/nextjs/src/server/authMiddleware.ts
@@ -28,15 +28,16 @@ type RouteMatcherWithNextTypedRoutes =
   | (string & {});
 
 /**
- * The default ideal matcher that excludes the _next directory (internals) and all static files.
+ * The default ideal matcher that excludes the _next directory (internals) and all static files,
+ * but it will match the root route (/) and any routes that start with /api or /trpc.
  */
-export const DEFAULT_CONFIG_MATCHER = ['/((?!.*\\..*|_next).*)', '/'];
+export const DEFAULT_CONFIG_MATCHER = ['/((?!_next|.+\\..+))', '/', '/(api|trpc)(.*)'];
 
 /**
  * Any routes matching this path will be ignored by the middleware.
  * This is the inverted version of DEFAULT_CONFIG_MATCHER.
  */
-export const DEFAULT_IGNORED_ROUTES = ['/(.*\\..*|_next.*)'];
+export const DEFAULT_IGNORED_ROUTES = ['/((?!api|trpc))(_next|.+\\..+)(.*)'];
 
 type RouteMatcherParam =
   | Array<RegExp | RouteMatcherWithNextTypedRoutes>
--- a/packages/nextjs/src/server/authMiddleware.ts
+++ b/packages/nextjs/src/server/authMiddleware.ts
@@ -4,13 +4,14 @@ import type Link from 'next/link';
 import type { NextFetchEvent, NextMiddleware, NextRequest } from 'next/server';
 import { NextResponse } from 'next/server';
 
-import { isRedirect, mergeResponses, paths, setHeader } from '../utils';
+import { isRedirect, mergeResponses, paths, setHeader, stringifyHeaders } from '../utils';
+import { withLogger } from '../utils/debugLogger';
 import { authenticateRequest, handleInterstitialState, handleUnknownState } from './authenticateRequest';
 import { SIGN_IN_URL, SIGN_UP_URL } from './clerkClient';
 import { receivedRequestForIgnoredRoute } from './errors';
 import { redirectToSignIn } from './redirect';
 import type { NextMiddlewareResult, WithAuthOptions } from './types';
-import { decorateRequest } from './utils';
+import { decorateRequest, setRequestHeadersOnNextResponse } from './utils';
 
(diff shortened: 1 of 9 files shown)


packages/nextjs/src/server/getAuth.ts
```diff
@@ -10,42 +10,65 @@ import {
...
+      const authMessage = getAuthKeyFromRequest(req, 'AuthMessage');
+      const authReason = getAuthKeyFromRequest(req, 'AuthReason');
+      logger.log('Headers debug', { authStatus, authMessage, authReason });
 
-  if (authStatus !== AuthStatus.SignedIn) {
-    return signedOutAuthObject(options);
-  }
+      if (!authStatus) {
+        throw new Error(noAuthStatusMessage);
+      }
```
nikosdouvlis, line 45: For the reviewers: This has not changed, this is just terrible git diff formatting

nikosdouvlis:
> ☁️ if we are going to use a logger i would suggest to use an existing logger library or use an a logger with the same interface as an existing library to avoid the mental overhead and allow use replacing our existing logger with a library in the future without replacing all the references. WDYT?

This is the intention that's why the loggerFactory and the types are injectable/pluggable. Replacing this temp logger with something along the lines of pino/winston etc will happen as part of an upcoming ticket as ideally we'd like to create a isomorphic solution we can use across edge/ node/ browser envs.

dimkl (approved):

packages/nextjs/src/utils/debugLogger.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const withLogger: WithLogger = (loggerFactoryOrName, handlerCtor) => {
+  return ((...args: any) => {
+    const factory =
+      typeof loggerFactoryOrName === 'string' ? createDebugLogger(loggerFactoryOrName) : loggerFactoryOrName;
+    const logger = factory();
+    const handler = handlerCtor(logger as any);
+    try {
+      const res = handler(...args);
+      if (typeof res === 'object' && 'then' in res && typeof res.then === 'function') {
```
SokratisVidros, line 54: ❓ Why don't we just always `await handler` to keep it simpler? Try/catch is already in place for async errors.

packages/nextjs/src/utils/response.ts
```diff
@@ -37,3 +37,15 @@ export const setHeader = (res: Response, name: string, val: string) => {
   res.headers.set(name, val);
   return res;
 };
+
+export const stringifyHeaders = (headers: { forEach: Headers['forEach'] }) => {
+  if (!headers) {
+    return JSON.stringify({});
+  }
+
+  const obj: Record<string, string> = {};
```
SokratisVidros, line 46: 🙃 Try this one-liner to see if it's working `JSON.stringify(Object.fromEntries(headers) || {})`.

SokratisVidros (approved):

SokratisVidros (approved):

packages/nextjs/src/utils/debugLogger.ts
nikosdouvlis: I don't want to make `withLogger` async as it can wrap synchronous methods as well

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -100,41 +101,57 @@ const authMiddleware: AuthMiddleware = (...args: unknown[]) => {
...
     if (requestState.isUnknown) {
+      logger.debug('authenticateRequest state is unknown', requestState);
       return handleUnknownState(requestState);
     } else if (requestState.isInterstitial) {
+      logger.debug('authenticateRequest state is interstitial', requestState);
       return handleInterstitialState(requestState, options);
     }
 
     const auth = Object.assign(requestState.toAuth(), { isPublicRoute: isPublicRoute(req) });
+    logger.debug(() => ({ auth: JSON.stringify(auth) }));
```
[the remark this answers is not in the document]
nikosdouvlis: Pretty much yes, but feel free to suggest different logs here! 
In general, we have a couple of logs describing the flow and some JSON-dumps to help us identify why a flow run.

Logs that could be "expensive", eg stringifying a potentially large object, are only evaluated if the logger is enabled, that's why they are wrapped with a function.

packages/nextjs/src/utils/debugLogger.ts
SokratisVidros: Valid point.

PR #1162, opened 2023-05-08
feat(clerk-sdk-node): Multi-domain support

gkats:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Added support for the multi-domain feature in our Node SDK. The authenticateRequest method now accepts four additional options; isSatellite, proxyUrl, domain and signInUrl.


packages/sdk-node/src/authenticateRequest.ts
```diff
@@ -56,3 +81,26 @@ export const decorateResponseWithObservabilityHeaders = (res: ServerResponse, re
...
+};
+
+const getRequestProto = (req: IncomingMessage): string => {
+  const proto = req.headers[constants.Headers.ForwardedProto] as string;
+  if (!proto) {
+    return req.headers.host && req.headers.host.match('localhost') ? 'http' : 'https';
+  }
+  const index = proto.indexOf(',');
+  return index > -1 ? proto.substring(0, index).trim() : proto.trim();
+};
```
gkats, line 99: I had a lot of trouble getting the request protocol from the [IncomingMessage](https://nodejs.org/api/http.html#class-httpincomingmessage) request.

I ended up looking for the "x-forwarded-proto" header and default to a hacky way to determine the protocol. All requests are considered https unless coming from localhost. 

Any suggestion for a better way to deal with this?

panteliselef: What if someone simply added a entry to their `/etc/hosts`  and they host now is `custom.local`, wouldn't that basically break this assumption ?

packages/sdk-node/src/authenticateRequest.ts
```diff
@@ -56,3 +81,26 @@ export const decorateResponseWithObservabilityHeaders = (res: ServerResponse, re
...
+  return new URL(req.url as string, `${getRequestProto(req)}://${req.headers.host}`);
+};
+
+const getRequestProto = (req: IncomingMessage): string => {
+  const proto = req.headers[constants.Headers.ForwardedProto] as string;
+  if (!proto) {
+    return req.headers.host && req.headers.host.match('localhost') ? 'http' : 'https';
+  }
+  const index = proto.indexOf(',');
+  return index > -1 ? proto.substring(0, index).trim() : proto.trim();
```
panteliselef, line 98: 🙃 ```suggestion
  return proto?.split(',')[0]?.trim();
```

**Review by panteliselef — APPROVED** [review]

File: packages/sdk-node/src/authenticateRequest.ts (hunk 1 above)
Comment (panteliselef) [reply packages/sdk-node/src/authenticateRequest.ts:99]: I yesterday gave it a try to find a better way to handle the fallback. I could not find something. Most servers correctly set the X-forwarded-proto header, so we should be good.

File: packages/sdk-node/src/authenticateRequest.ts [hunk 3]
```diff
@@ -56,3 +81,26 @@ export const decorateResponseWithObservabilityHeaders = (res: ServerResponse, re
...
 };
+
+const isDevelopmentFromApiKey = (apiKey: string): boolean =>
+  apiKey.startsWith('test_') || apiKey.startsWith('sk_test_');
+
+const getRequestUrl = (req: IncomingMessage): URL => {
+  return new URL(req.url as string, `${getRequestProto(req)}://${req.headers.host}`);
+};
+
+const getRequestProto = (req: IncomingMessage): string => {
```
SokratisVidros, line 92: ❓ What about [request.protocol](https://nodejs.org/docs/latest-v18.x/api/http.html#requestprotocol)?

SokratisVidros (commented):
Done with the first pass.

packages/sdk-node/src/authenticateRequest.ts
```diff
@@ -56,3 +81,26 @@ export const decorateResponseWithObservabilityHeaders = (res: ServerResponse, re
   requestState.reason && res.setHeader(constants.Headers.AuthReason, encodeURIComponent(requestState.reason));
   requestState.status && res.setHeader(constants.Headers.AuthStatus, encodeURIComponent(requestState.status));
 };
+
+const isDevelopmentFromApiKey = (apiKey: string): boolean =>
+  apiKey.startsWith('test_') || apiKey.startsWith('sk_test_');
+
+const getRequestUrl = (req: IncomingMessage): URL => {
+  return new URL(req.url as string, `${getRequestProto(req)}://${req.headers.host}`);
```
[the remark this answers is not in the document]
panteliselef: req.url seems to alway be a relative path

packages/sdk-node/src/authenticateRequest.ts
panteliselef: `request.protocol` is populated but the because `authenticateRequest` accepts request as `IncomingMessage` protocol is not part of the type.

Should we update the type to `http.ClientRequest` ?. 

@dimkl I saw that you updated that type from `ExpressRequest` to  `IncomingMessage`. Any particular reason why we did not go with `http.ClientRequest` back then ?

dimkl: based on the docs the `ExpressRequest` extends the `http.IncomingMessage`, not the `http.ClientRequest` interface.

packages/sdk-node/src/authenticateRequest.ts
```diff
@@ -56,3 +81,26 @@ export const decorateResponseWithObservabilityHeaders = (res: ServerResponse, re
...
+};
+
+const getRequestProto = (req: IncomingMessage): string => {
+  const proto = req.headers[constants.Headers.ForwardedProto] as string;
+  if (!proto) {
+    return req.headers.host && req.headers.host.match('localhost') ? 'http' : 'https';
```
dimkl, line 95: based on the express codebase we probably could do this: 
```typescript
proto = req.connection.encrypted ? 'https' : 'http';
```
ref: https://github.com/expressjs/express/blob/0debedf4f31bb20203da0534719b9b10d6ac9a29/lib/request.js#LL307C1-L309C14

dimkl (approved):

packages/sdk-node/src/authenticateRequest.ts
gkats: Thanks for the suggestion @dimkl but unfortunately this check does not apply in our case.

The `req` object's type is `IncomingMessage` and the [connection property](https://nodejs.org/api/http.html#messageconnection) is different than ClientRequest's `connection` property. This is an alias for the socket.

SokratisVidros (approved):

packages/sdk-node/src/authenticateRequest.ts
gkats: Discussed offline with @dimkl and @SokratisVidros and we agreed on the following.

- We won't (and can't) change the `req` object's type as some frameworks extend `IncomingMessage` for their custom request objects. A good example is express and its own [Request](https://github.com/expressjs/express/blob/353348a83e6874db7ded27102e91e7b6cad7e89c/lib/request.js#L31).
- Since we're stuck with `IncomingMessage` and there's no easy way to determine the protocol, we'll use the following algorithm:
  1. Check the "X-Forwarded-Proto" header. This should take precedence.
  2. Check if `req.connection.encrypted` is set. Perhaps we're lucky and the `req` object comes in as a `ClientRequest`.
  3. If you can't find the request protocol, throw an error. Fail fast with a descriptive message.
  
Updated the code with the above changes.

packages/sdk-node/src/authenticateRequest.ts
```diff
@@ -1,5 +1,6 @@
 import type { Clerk, RequestState } from '@clerk/backend';
 import { constants } from '@clerk/backend';
+import { handleValueOrFn, isHttpOrHttps, isProxyUrlRelative, isValidProxyUrl } from '@clerk/shared';
```
dimkl, line 3: @nikosdouvlis is it okay to use the import of `@clerk/shared` in `@clerk/sdk-node` or will we have issues with bundling?

dimkl (approved):

panteliselef (approved):

packages/sdk-node/src/authenticateRequest.ts
nikosdouvlis: I think this is safe to do, but we do need to explicitly list the dependency in the package.json of clerk-sdk-node

panteliselef:
@dimkl  @SokratisVidros  I request again for a review as a few changes were added.

We now support both localInterstitial and remotePrivate as before, this was to avoid updating the remote private interstitial endpoint with the "multi-domain" logic and have one more place with duplicated logic

packages/sdk-node/src/authenticateRequest.ts
SokratisVidros: 👍

packages/sdk-node/src/authenticateRequest.ts
```diff
@@ -56,3 +81,40 @@ export const decorateResponseWithObservabilityHeaders = (res: ServerResponse, re
...
+    return relativeOrAbsoluteUrl;
+  }
+  return new URL(relativeOrAbsoluteUrl, baseUrl).toString();
+};
+
+const satelliteAndMissingProxyUrlAndDomain =
+  'Missing domain and proxyUrl. A satellite application needs to specify a domain or a proxyUrl';
+const satelliteAndMissingSignInUrl = `
+Invalid signInUrl. A satellite application requires a signInUrl for development instances.
+Check if signInUrl is missing from your configuration or it is not a absolute URL.`;
```
SokratisVidros, line 118:
```suggestion
Check if signInUrl is missing from your configuration or if it is not an absolute URL.`;
```

packages/sdk-node/src/authenticateRequest.ts
```diff
@@ -56,3 +81,40 @@ export const decorateResponseWithObservabilityHeaders = (res: ServerResponse, re
...
+const missingProto =
+  "Cannot determine the request protocol. Please make sure you've set the X-Forwarded-Proto header with the request protocol (http or https).";
```
SokratisVidros, line 120:
```suggestion
  "Cannot determine the request protocol. Please ensure you've set the X-Forwarded-Proto header with the request protocol (http or https).";
```

packages/sdk-node/src/clerkExpressRequireAuth.ts
```diff
@@ -43,7 +43,25 @@ export const createClerkExpressRequireAuth = (createOpts: CreateClerkExpressMidd
         return handleUnknownCase(res, requestState);
       }
       if (requestState.isInterstitial) {
-        const interstitial = await clerkClient.remotePrivateInterstitial();
+        let interstitial;
+
+        /**
+         * This is a step for deprecating the usage of `remotePrivateInterstitial`
```
SokratisVidros, line 49: 🙃 If the PK is unavailable in Express, we can't get a local interstitial. So, for the future reader, the comment should state that this is an improvement if we have PK or FrontendAPI key to avoid an extra network call. Mentioning deprecation is a bit confusing.

packages/sdk-node/src/clerkExpressWithAuth.ts
```diff
@@ -25,7 +25,25 @@ export const createClerkExpressWithAuth = (createOpts: CreateClerkExpressMiddlew
...
       }
       if (requestState.isInterstitial) {
-        const interstitial = await clerkClient.remotePrivateInterstitial();
+        let interstitial;
+
+        /**
+         * This is a step for deprecating the usage of `remotePrivateInterstitial`
+         * For the multi-domain feature and when frontendApi is set prefer the localInterstitial
+         */
+        if (requestState.publishableKey || requestState.frontendApi) {
```
SokratisVidros, line 34: 🔧 Let's introduce an interstitial helper function to make the code more dry.

SokratisVidros (approved):

SokratisVidros (approved):

PR #1187, opened 2023-05-12
COR-394: fix infinite redirect for Multi domain development instances

panteliselef:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### Background
So far for development instances, the sync/link flow boils down to:
- Visit the satellite app.
- Satellite redirects to `signInUrl`
- Primary (with clerk loaded) is accessed, redirects back to satellite
- Satellite receives a request with `__clerk_synced=true` and sets the `client_uat` cookie
- Satellite reloads in order to read the `client_uat` cookie

###  The problem
All the above work perfectly fine when both apps are on localhost. But, let's say that we add an nginx configuration and we now have [primary.dev](http://primary.dev/) and [satellite.dev](http://satellite.dev/) as the domains with self-signed certificates. 

Let's repeat the above steps.
- Visit `https://satellite.dev/`
- Satellite redirects to `https://primary.dev/whatever`
- Primary is accessed, redirects back to `https://satellite.dev?__clerk_synced=true`
- Satellite sets the client_uat cookie.

In this final step, the `client_uat` cookie cannot be set. The `client_uat` cookie's SameSite policy is "Strict". The request is considered cross-origin by the browser, because the referrer is `primary.dev`, so the "Strict" policy will block the cookie.
As a result, when satellite reloads in step (5), the client_uat cookie isn't there and the satellite tries to sync again. Visits https://primary.dev/whatever and we get stuck in an infinite loop of redirects.

### The solution of this PR
1. Create a new interstitial rule when isDev=true &  isSatellite=true  and we are returning back from a primary. Previously we were not throwing interstitial in that occasion. Now we do and a new query param was added to handle that case (__clerk_referrer_primary).
2. In development instances we look at __clerk_referrer_primary  and if it exists we know that interstitial has run, and we replace it with clerk_synced and then reload. This eliminates the infinite loop, as now the client_uat is set properly and we persist clerk_synced on reload.

### DIFF
--- a/packages/backend/src/tokens/request.test.ts
+++ b/packages/backend/src/tokens/request.test.ts
@@ -49,6 +49,7 @@ function assertInterstitial(
     reason: AuthReason;
     isSatellite?: boolean;
     domain?: string;
+    signInUrl?: string;
   },
 ) {
   assert.propContains(requestState, {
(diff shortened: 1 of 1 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -1045,6 +1046,13 @@ export default class Clerk implements ClerkInterface {
 
   #clearJustSynced = () => removeClerkQueryParam(CLERK_SYNCED);
 
+  #isReturningFromPrimary = () => getClerkQueryParam(CLERK_REFERRER_PRIMARY) === 'true';
+
+  #clearPrimaryReferrer = () => {
```
raptisj, line 1051: @panteliselef 🙃 the function name should indicate that it sets another query param.

raptisj (approved):

PR #1194, opened 2023-05-16
COR-370 fix(backend): Add delete domain method

raptisj:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

 Add delete domain method


packages/backend/src/api/resources/Domains.ts
```diff
@@ -0,0 +1,9 @@
+import type { DomainJSON } from './JSON';
+
+export class Domains {
+  constructor(readonly id: string, readonly name: string) {}
+
```
panteliselef, line 5: ❓ I'd expect a delete method here, am I missing something ?

packages/backend/src/util/mockFetch.ts
```diff
@@ -1,8 +1,8 @@
-export function jsonOk(body: unknown) {
+export function jsonOk(body: unknown, status = 200) {
```
panteliselef, line 1: 💯

packages/backend/src/api/resources/Domains.ts
```diff
@@ -0,0 +1,9 @@
+import type { DomainJSON } from './JSON';
+
+export class Domains {
```
panteliselef, line 3: ❓Shouldn't this extend BaseResource ?

packages/backend/src/api/endpoints/DomainApi.ts
```diff
@@ -0,0 +1,14 @@
+import { joinPaths } from '../../util/path';
+import type { Domains } from '../resources/Domains';
+import { AbstractAPI } from './AbstractApi';
+
+const basePath = '/domains';
+
+export class DomainAPI extends AbstractAPI {
+  public async deleteDomain(id: string) {
+    return this.request<Domains>({
```
gkats, line 9: :question: Is `Domains` the return type here? If yes, it should be a [DeletedObject](https://github.com/clerkinc/javascript/blob/3be2c91a0e9119ade32cf195ba1cae2e97e1e74e/packages/backend/src/api/resources/DeletedObject.ts).

packages/backend/src/api/resources/Domains.ts
gkats: This whole class is probably not needed, since the `deleteDomain()` function will return a `DeletedObject`.

playground/express/README.md
```diff
@@ -13,7 +13,7 @@ Execute in current folder:
 touch .env # set PUBLISHABLE_KEY and SECRET_KEY from Clerk Dashboard API keys
 npm i
 rm -rf node_modules/@clerk
-yalc add @clerk/clerk-sdk-node --pure
+yalc add @clerk/clerk-sdk-node # also add the packages you made changes to, e.g. @clerk/bakcned @clerk/types.
```
gkats, line 16: :upside_down_face: 

```suggestion
yalc add @clerk/clerk-sdk-node # also add the packages you made changes to, e.g. @clerk/backend @clerk/types.
```

gkats (changes requested):
Nice one @raptisj. 

Left one comment about the return type. Other than that, looks good to me!

packages/backend/src/api/resources/Domains.ts
raptisj: @panteliselef the convention here is to have the methods in the `/endpoints` folder

packages/backend/src/api/resources/Domains.ts
raptisj: @panteliselef BaseResource isn't present here like it is in clerkjs. The `factory` file handles the request

packages/backend/src/api/endpoints/DomainApi.ts
raptisj: @gkats I'm a bit confused about this. What determines if we are going to use the `DeleteObject`? 

We seem to have it both ways. We seem to use it in [deletePhoneNumber](https://github.com/clerkinc/javascript/blob/3be2c91a0e9119ade32cf195ba1cae2e97e1e74e/packages/backend/src/api/endpoints/PhoneNumberApi.ts#L50) and [deleteEmailAddress](https://github.com/clerkinc/javascript/blob/3be2c91a0e9119ade32cf195ba1cae2e97e1e74e/packages/backend/src/api/endpoints/EmailAddressApi.ts#L50) but not in [deleteOrganization](https://github.com/clerkinc/javascript/blob/3be2c91a0e9119ade32cf195ba1cae2e97e1e74e/packages/backend/src/api/endpoints/OrganizationApi.ts#LL129C16-L129C34), [deleteUser](https://github.com/clerkinc/javascript/blob/3be2c91a0e9119ade32cf195ba1cae2e97e1e74e/packages/backend/src/api/endpoints/UserApi.ts#LL139C16-L139C26) and [disableUserMFA](https://github.com/clerkinc/javascript/blob/3be2c91a0e9119ade32cf195ba1cae2e97e1e74e/packages/backend/src/api/endpoints/UserApi.ts#LL163C16-L163C30) for example.

cc @dimkl @nikosdouvlis 
Can you clarify the thought behind this a bit more?

packages/backend/src/api/resources/Domains.ts
panteliselef: My bad, didn't see the path of the file

packages/backend/src/api/endpoints/DomainApi.ts
gkats: Hey @raptisj, I can see why you might be confused by this. It's true that we're not consistent in our return types and we probably use some wrong types in certain functions. Sorry for that.

> What determines if we are going to use the DeleteObject

We basically have to check the API response. At the moment there's no documentation for the DELETE /v1/domains so you have every right to be confused. Here's an example for the [delete organization operation](https://clerk.com/docs/reference/backend-api/tag/Organizations#operation/DeleteOrganization). You'll notice the successful response body is a "DeletedObject".

So, again, apologies for not making that clear to you in the first place, but the DELETE /v1/domains endpoint will return a "DeletedObject" on successful invocation.

I'll update our docs shortly.

packages/backend/src/api/factory.test.ts
```diff
@@ -173,5 +174,23 @@ export default (QUnit: QUnit) => {
         }),
       );
     });
+
+    test('executes a successful backend API request to delete a domain', async assert => {
+      fakeFetch = sinon.stub(runtime, 'fetch');
+      fakeFetch.onCall(0).returns(jsonOk(deletedObjectJson, 204));
```
gkats, line 180:
```suggestion
      const domainId = "dmn_123";
      const fakeResponse = {
        object: "domain",
        id: domainId,
        deleted: true,
      };
      fakeFetch.onCall(0).returns(jsonOk(fakeResponse, 204));
```

packages/backend/src/fixtures/responses/deletedObject.json
```diff
@@ -0,0 +1,6 @@
+{
+  "object": "domains",
+  "id": "123123123",
+  "slug": "",
+  "deleted": true
+}
```
gkats, line 6: :wrench: This is not a generic "DeletedObject". It's the JSON payload for a "domain" deleted object. If we want to make this re-usable we should leave any mention of "domain" out of the payload.

How about we remove this file and instead move the fake response inside the test? See the other comment above.

playground/express/README.md
```diff
@@ -13,7 +13,7 @@ Execute in current folder:
 touch .env # set PUBLISHABLE_KEY and SECRET_KEY from Clerk Dashboard API keys
 npm i
 rm -rf node_modules/@clerk
-yalc add @clerk/clerk-sdk-node --pure
+yalc add @clerk/clerk-sdk-node # also add the packages you made changes to, e.g. @clerk/backned @clerk/types.
```
gkats, line 16: :wrench: 
```suggestion
yalc add @clerk/clerk-sdk-node # also add the packages you made changes to, e.g. @clerk/backend @clerk/types.
```

packages/backend/src/api/resources/JSON.ts
```diff
@@ -302,3 +302,8 @@ export interface DeletedObjectJSON {
   slug?: string;
   deleted: boolean;
 }
+
+export interface DomainJSON {
+  id: string;
+  name: string;
+}
```
gkats, line 309: :wrench: Is this used anywhere? If not, let's remove.

packages/backend/src/api/factory.test.ts
```diff
@@ -173,5 +174,23 @@ export default (QUnit: QUnit) => {
...
+
+      await apiClient.domains.deleteDomain('123123123');
```
gkats, line 182:
```suggestion
      await apiClient.domains.deleteDomain(domainId);
```

packages/backend/src/api/factory.test.ts
```diff
@@ -173,5 +174,23 @@ export default (QUnit: QUnit) => {
...
+
+      assert.ok(
+        fakeFetch.calledOnceWith('https://api.clerk.test/v1/domains/123123123', {
```
gkats, line 185:
```suggestion
        fakeFetch.calledOnceWith(`https://api.clerk.test/v1/domains/${domainId}`, {
```

gkats (changes requested):
Nice! Left some additional comments. Let's fix those and then we can merge. :rocket:

PR #1198, opened 2023-05-17
feat(clerk-js): Retry get requests less frequently if online

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description


packages/shared/src/utils/browser.ts
```diff
@@ -61,3 +61,18 @@ export function isValidBrowserOnline(): boolean {
 
   return !isUserAgentRobot && !isWebDriver && isExperimentalConnectionOnline && isNavigatorOnline;
 }
+
+export function isBrowserOnline(): boolean {
```
nikosdouvlis, line 65: We already have an `isValidBrowserOnline` function. Can we refactor/reuse that?

desiprisg: We want the opposite logic here. If it's not a valid browser, but it's online, we don't want to retry forever.

desiprisg: We could split the two checks into two smaller utilities and use those.

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -66,8 +66,7 @@ export class Session extends BaseResource implements SessionResource {
 
   getToken: GetToken = async (options?: GetTokenOptions): Promise<string | null> => {
     return runWithExponentialBackOff(() => this._getToken(options), {
-      shouldRetry: e => !isUnauthorizedError(e),
-      maxRetries: 8,
+      shouldRetry: (e: unknown, i: number) => !isUnauthorizedError(e) || i < 8,
```
SokratisVidros, line 69: 💯

packages/shared/src/utils/browser.ts
SokratisVidros: ☝️ Let's do that and wrap this up.

SokratisVidros (commented):
@desiprisg How does the new retry policy feel in the browser? Is it too slow for the first iteration or not?

packages/clerk-js/src/core/fapiClient.ts
```diff
@@ -199,13 +199,16 @@ export default function createFapiClient(clerkInstance: Clerk): FapiClient {
 
     try {
       if (beforeRequestCallbacksResult) {
+        const maxTries = isBrowserOnline() ? 4 : 11;
```
nikosdouvlis, line 202: This check here determines the maxTries the moment the request begins.
If we move it inside shouldRetry, it will also catch cases where the browser goes offline in the middle of the request. Which one do you prefer @desiprisg ?

nikosdouvlis (approved):

packages/clerk-js/src/core/fapiClient.ts
desiprisg: I don't think that's better. In that case, if the browser's offline and then becomes online and we're at the 6th try, the check will fail.

PR #1200, opened 2023-05-17
fix(clerk-js): Serve static provider SVGs via new img-service

nikosdouvlis:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

We need to serve the static files, eg oauth provider icons, through the new img-service worker.

We already have a proxy route in place, however this route requires its payload to be a base64 encoded json. There's no need to do this for static assets clerk-js needs, so we will introduce a /static endpoint:

img.clerk.com/static/google.svg

that will use the proxy route internally. Clerkjs will only be aware of the above route, requiring no changes in logic

### DIFF
--- a/packages/clerk-js/src/ui/common/constants.ts
+++ b/packages/clerk-js/src/ui/common/constants.ts
@@ -96,6 +96,19 @@ export function getWeb3ProviderData(name: Web3Provider): Web3ProviderData | unde
   return WEB3_PROVIDERS[name];
 }
 
+/**
+ * Returns the URL for a static SVG image
+ * using the old images.clerk.com service
+ * @deprecated In favor of iconImageUrl
+ */
 export function svgUrl(id: string): string {
   return `https://images.clerk.com/static/${id}.svg`;
+}
+
+/**
+ * Returns the URL for a static SVG image
+ * using the new img.clerk.com service
+ */
+export function iconImageUrl(id: string): string {
+  return `https://img.clerk.com/static/${id}.svg`;
 }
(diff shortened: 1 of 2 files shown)


packages/clerk-js/src/ui/common/constants.ts
```diff
@@ -96,6 +96,19 @@ export function getWeb3ProviderData(name: Web3Provider): Web3ProviderData | unde
...
 export function svgUrl(id: string): string {
   return `https://images.clerk.com/static/${id}.svg`;
 }
+
+/**
+ * Returns the URL for a static SVG image
+ * using the new img.clerk.com service
+ */
+export function iconImageUrl(id: string): string {
+  return `https://img.clerk.com/static/${id}.svg`;
```
desiprisg, line 113: I know it was like that before as well, but shouldn't this be different for staging?

nikosdouvlis: This can be done when we start splitting the bundles based on the build target, right now I'm not sure if it's worth propagating this info to this specific function as the logic and the URL won't really change.

anagstef: @nikosdouvlis That's because static assets/files will eventually be the same on both staging and production environments, right?

nikosdouvlis: Yes. Actually, we only upload them to the prod bucket right now (it's just the oauth provider logos)

panteliselef:
Is this a step towards deliver our internal svg assets of clerk js via img-service or simply to support svg avatars ?

dimkl (approved):

nikosdouvlis:
> Is this a step towards deliver our internal svg assets of clerk js via img-service or simply to support svg avatars ?

Only changing the img-service URL :)

PR #1210, opened 2023-05-19
fix(clerk.js): Don't duplicate the protocol separator while producing the request URL, for issue #1209

### LINKED ISSUE #1209: Error caused by wrong url assembled using URL::protocol
The utility method `getRequestUrl` tries to acquire the protocol from the header "X-Forwarded-Host" and when it is not available, it falls back to `url.protocol`.

The problem is "X-Forwarded-Host" contains just the protocol name (ex: "http") while NodeJS URL class's 
`protocol` property returns the protocol and the separator colon, so the assembled URL has 2 colons.

The error:
```
Error [TypeError]: Invalid base URL: http:://localhost:3003
```

This happens here:
https://github.com/clerkinc/javascript/blob/5ec4c3e79ed6b4ba58a24cfe9a5279fad36b2f13/packages/nextjs/src/server/utils.ts#L185

brunoschmidt:
## Type of change

- [X] 🐛 Bug fix

## Packages affected

- [X] `@clerk/nextjs`

## Description

Small fix to not duplicate the comma separating the protocol, when there is no "X-Forwarded-Host" header.

Fixes #1209


packages/nextjs/src/server/utils.ts
```diff
@@ -182,7 +182,7 @@ export function getRequestUrl({ request, relativePath }: { request: Request; rel
 
   // X-Forwarded-Proto could be 'https, http'
   const protocol =
-    (headers.get('X-Forwarded-Proto') ?? (headers as any)['X-Forwarded-Proto'])?.split(',')[0] ?? url.protocol;
+    (headers.get('X-Forwarded-Proto') ?? (headers as any)['X-Forwarded-Proto'])?.split(',')[0] ?? (url.protocol && url.protocol.slice(0, -1));
```
panteliselef, line 185: This feels more robust, thanks for raising this 😀
```suggestion
    (headers.get('X-Forwarded-Proto') ?? (headers as any)['X-Forwarded-Proto'])?.split(',')[0] ?? (url.protocol && url.protocol.replace(/\:$/, '');
```

panteliselef (changes requested):

packages/nextjs/src/server/utils.ts
```diff
@@ -182,7 +182,7 @@ export function getRequestUrl({ request, relativePath }: { request: Request; rel
 
   // X-Forwarded-Proto could be 'https, http'
   const protocol =
-    (headers.get('X-Forwarded-Proto') ?? (headers as any)['X-Forwarded-Proto'])?.split(',')[0] ?? url.protocol;
+    (headers.get('X-Forwarded-Proto') ?? (headers as any)['X-Forwarded-Proto'])?.split(',')[0] ?? (url.protocol && url.protocol.replace(/\:$/, '');
```
panteliselef, line 185: 🔧  Please add a parenthesis at the end of the line 
```suggestion
    (headers.get('X-Forwarded-Proto') ?? (headers as any)['X-Forwarded-Proto'])?.split(',')[0] ?? (url.protocol && url.protocol.replace(/\:$/, ''));
```

panteliselef:
@brunoschmidt  could you please squash and rebase 🙃

nikosdouvlis:
No need to rebase or anything, I will go ahead and merge this so we can get it out with our next release.
Many thanks for the contribution @brunoschmidt :)

PR #1229, opened 2023-05-23
fix(nextjs): Resolve awaitable navigate if provided with the current url

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

The awaitable navigate promise never resolves if we pass the current url as an argument, because a rerender is never triggered. This causes our `await beforeEmit()` function call to block the execution in `clerk.ts` and thus we are stuck in the transitive state and `Clerk` properties are undefined. 

You can reproduce this bug by signing in with a modal while being in the root page in the app-router playground app.


desiprisg:
Can we just instantly resolve here instead of the whole rerender thing?

packages/nextjs/src/app-router/client/useAwaitableNavigate.ts
```diff
@@ -26,6 +27,13 @@ export const useAwaitableNavigate = () => {
         } else {
           window.__clerk_nav_resolves_ref = [res];
         }
+
+        if (to === window.location.href) {
+          setTimeout(() => {
+            forceUpdate();
```
nikosdouvlis, line 33: Why are we doing a forceUpdate rather than resolving all pending promises here?

desiprisg: Was mostly a POC as I was figuring out how this function worked. I've updated the code. What's also weird is that it seemed to not be working because I was not stripping the origin, but I've fixed that. Although I remember it being fixed. That was weird :/

PR #1230, opened 2023-05-23
fix(clerk-sdk-node): load config on first use

dimkl:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### DIFF
--- a/.changeset/twelve-avocados-kneel.md
+++ b/.changeset/twelve-avocados-kneel.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-sdk-node': patch
+---
+
+Load env variables upon first usage of middlewares or clerkClient
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,6 +14,13 @@
         "request": "launch",
         "command": "npm run dev",
         "cwd": "${workspaceFolder}/playground/nextjs12"
+      },
+      {
+        "name": "playground/express: debug server",
+        "type": "node-terminal",
+        "request": "launch",
+        "command": "npm start",
+        "cwd": "${workspaceFolder}/playground/express"
       }
     ]
   }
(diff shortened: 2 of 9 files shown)


SokratisVidros:
@dimkl Can you please elaborate on the issue that this change addresses?

nikosdouvlis:
> @dimkl Can you please elaborate on the issue that this change addresses?

Currently in Clerk Node, the following code doesn't run because Clerk doesn't read any of the keys even though they are clearly imported, this leads to a bad DX and reports that Clerk doesn't work correctly.

```ts
import { ClerkExpressWithAuth } from "@clerk/clerk-sdk-node";
import express from "express";

require('dotenv').config()

const port = process.env.PORT || 3000;

const app = express();
// Use the lax middleware that returns an empty auth object when unauthenticated
app.get(
  "/protected-endpoint",
  ClerkExpressWithAuth({
    // ...options
  }),
  (req, res) => {
    res.json(req.auth);
  }
);

app.listen(port, () => {
  console.log(`Example app listening at http://localhost:${port}`);
});
```

Notice that since there is no underlying framework to load the env *before* the Clerk imports are evaluated, the order of the imports and the `dotenv.config` invocation is significant here, it's very likely for Clerk to be initialized before `env.process` is.

We need to read the *current* env value when needed, not just once during initialization.

clerk-cookie:

PR #1235, opened 2023-05-23
fix(clerk-js): Remove invalid props on DOM element

tmilewski:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

`npm test` was outputting `Warning: React does not recognize the 'X' prop on a DOM element.`

This PR ensures that those props don't make their way to the DOM element.

### DIFF
--- a/packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
+++ b/packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
@@ -53,7 +53,20 @@ export const InviteMembersForm = (props: InviteMembersFormProps) => {
   });
 
   const {
-    props: { errorText, hasLostFocus, setError, successfulText, setSuccessful, ...restEmailAddressProps },
+    props: {
+      /* eslint-disable @typescript-eslint/no-unused-vars */
+      enableErrorAfterBlur,
+      errorText,
+      hasLostFocus,
+      isFocused,
+      setError,
+      setWarning,
+      setSuccessful,
+      successfulText,
+      warningText,
+      /* eslint-enable @typescript-eslint/no-unused-vars */
+      ...restEmailAddressProps
+    },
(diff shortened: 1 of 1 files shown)


panteliselef (approved):

nikosdouvlis (approved):
This looks good :) Feel free to merge anytime

cc @desiprisg, @panteliselef 
For future reference, I'd like to point out that we can probably handle filtering all the invalid DOM props for all components in the `primitives` or `customizable` layer.

For example: 
https://github.com/clerkinc/javascript/blob/85e83e391e6a4d9b7ac664bd6f6f677accc52d5e/packages/clerk-js/src/ui/primitives/Input.tsx#L47

This wouldn't be the right place for it, but it's a good starting point for a discussion.  In general, all components go through the `makeCustomizable` HOF and every component uses the `createStyles` helper, so both would be good candidates for a future refactor.

Let me know what you think :)

PS: Just in case I wasn't clear, the above wouldn't be a part of this PR - the fix proposed here is fine :)

tmilewski:
@nikosdouvlis Thanks for the context,  and yes, that all makes sense!

I have a few thoughts here and there on potential refactors, but I want to get a better lay of the land first. 👍🏻

PR #1241, opened 2023-05-24
Append Dev Browser to redirect URL when using Next.js middleware and Hosted Pages

anagstef:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

# The problem

Next.js middleware has no access to browser’s local storage (because it runs on the server) and redirects the user before loading the Clerk-js script. That is, the **Dev Browser JWT** is not getting propagated to **Clerk Hosted Pages**, when redirecting there for signing in. If Clerk Hosted Pages have a different auth state on its Dev Browser JWT, then the user ends up in one of the following scenarios:

### 1. Infinite loop between `localhost` and Clerk Hosted Pages.

To recreate this scenario:

1. Create a new Next.Js app-router app with Clerk
2. Do not use repo-level environment variables for sign-in, sign-up, etc. (and keep the default Clerk Hosted Pages paths)
3. Use Clerk’s `authMiddleware` without any arguments (this will result in all pages to be protected)
4. Visit `localhost` and sign in (through the Clerk Hosted Pages)
5. Go to `localhost` and clear cookies and local storage
6. We end up in an infinite loop between `localhost` and Clerk Hosted Pages

### 2. Accessing a `localhost` protected page will always redirect you to `xyz.accounts.dev/default-redirect`

To recreate this scenario:

1. Create a new Next.Js app-router app with Clerk
2. Do not use repo-level environment variables for sign-in, sign-up, etc. (and keep the default Clerk Hosted Pages paths)
3. Use Clerk’s `authMiddleware` without any arguments (this will result in all pages to be protected)
4. Visit Clerk Hosted Pages first and sign in
5. Try to visit `localhost` and it will always redirect to `xyz.accounts.dev/default-redirect`

Both scenarios suffer from the same bug. The main issue is that `localhost` and Clerk Hosted Pages have different Dev Browser JWT, the first one is unauthenticated (or totally missing), and the Hosted Pages have a signed-in one.

When trying to access an unauthenticated page on `localhost`, it redirects to Clerk Hosted Pages. 

In scenario 1, Clerk Hosted Pages have their `Development Origin` updated to localhost, so it redirects back to localhost, and the infinite loop happens.

In scenario 2, we signed in directly to Clerk Hosted Pages, so the `HomeUrl` redirects to `/default-redirect`. That’s why we cannot access the `localhost` protected pages.

# Proposed solution

To solve this, we need to access the `localhost` dev browser from the Next.js middleware (server-side), where Clerk-js has not yet loaded. So, we store the Dev Browser JWT in a `__clerk_db_jwt` cookie, which can be accessed by Clerk’s `authMiddleware` and then appended to the URL when redirecting from localhost to Clerk Hosted Pages, so that the Dev Browser JWT from localhost overrides the one in Clerk Hosted Pages. 

**The localhost should be the source of truth for the auth state.**

-------------------

>**Side note:** 
>An initial approach involved an internal setting on Clerk-js to choose between Cookies and Local Storage for storing and accessing the Dev Browser JWT. This approach hits a wall when loading the Clerk-js script through interstitial, where we cannot append any options. So, we end up saving the Dev Browser cookie on both cookies and local storage.


dimkl (approved):

packages/clerk-js/src/core/devBrowserHandler.ts
```diff
@@ -38,11 +38,17 @@ export default function createDevBrowserHandler({
 
   function setDevBrowserJWT(jwt: string) {
     localStorage.setItem(key, jwt);
+    // Append dev browser JWT to cookies, because Next.js middleware
```
nikosdouvlis, line 41: nitpick: this is not specific to nextjs, let's update the comment so it doesn't confuses us in the future.

packages/clerk-js/src/utils/cookies/devBrowser.ts
```diff
@@ -0,0 +1,5 @@
+import { createCookieHandler } from '@clerk/shared';
+
+const DEV_BROWSER_COOKIE_NAME = '__clerk_db_jwt';
```
nikosdouvlis, line 3: Can we try not redeclaring the constant here? 
Here's another place in the code that defines the same value: https://github.com/clerkinc/javascript/blob/d53432e4a57df64cd0556d53ff20eb0a23f4d3f5/packages/clerk-js/src/utils/devBrowser.ts#L3

nikosdouvlis (approved):

packages/clerk-js/src/core/devBrowserHandler.ts
```diff
@@ -38,11 +38,14 @@ export default function createDevBrowserHandler({
 
   function setDevBrowserJWT(jwt: string) {
     localStorage.setItem(key, jwt);
+    // Append dev browser JWT to cookies, because server-side redirects (e.g. middleware) has no access to local storage
+    cookieHandler.setDevBrowserCookie(jwt);
```
SokratisVidros, line 42: @anagstef Can we do this conditionally only for Next.js for now? We want to avoid sending the __client cookie in other frameworks as it confused customers in the past, especially for those who implement their own JWT verification mechanism.

SokratisVidros: Discussed offline with Nikos.

anagstef: @SokratisVidros Unfortunately, we cannot distinguish which framework is used when executing the local interstitial. This is where this actually runs, when called by interstitial.

PR #1246, opened 2023-05-26
fix(clerk-js): Fix uneven profile card left/right spacing

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

An invisible scrollbar was adding more space to the right side of components like `<CreateOrganization/>` and `<UserProfile/>`. I've modified the `maxHeightScroller` to use `overflowY: auto` and `scrollbar-gutter: stable both-edges` to create the appropriate space in the opposite side.


packages/clerk-js/src/ui/styledSystem/common.ts
```diff
@@ -194,12 +194,13 @@ const unstyledScrollbar = (t: InternalTheme) => ({
   '::-webkit-scrollbar-track': {
     background: 'transparent',
   },
+  scrollbarGutter: 'stable both-edges',
```
panteliselef, line 197: I believe this is not available for Safari, will that be a problem ?

nikosdouvlis:
Why is there a scroller in the first place? Can you share more context please?

packages/clerk-js/src/ui/styledSystem/common.ts
desiprisg: The alternative is `overflow: overlay` which is a legacy value.

desiprisg:
@nikosdouvlis With `overflow: auto`, there will no longer be a scrollbar when the parent height is enough. I.e. it will appear in our modals and if the developer has a container that is limited in its height.
https://github.com/clerkinc/javascript/blob/abb076a171480722d3ee6c0ff116d3bda500ed3e/packages/clerk-js/src/ui/elements/ProfileCardContent.tsx#L41 This is the code that caused the scrollbar

anagstef (approved):

PR #1262, opened 2023-05-29
fix(backend): Issue with forwarded ports with load balancers(eg AWS load balancing)

dimkl:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Use port from the origin or `x-forwarded-port` header to construct URL for host and use protocol to determine ports for both origin and final URLs if the ports are missing.


packages/backend/src/util/request.ts
```diff
@@ -37,10 +37,23 @@ export function checkCrossOrigin({
...
+export function convertHostHeaderValueToURL(host: string, protocol = 'https:'): URL {
   /**
    * The protocol is added for the URL constructor to work properly.
    * We do not check for the protocol at any point later on.
    */
-  return new URL(`https://${host}`);
+  return new URL(`${protocol}//${host}`);
+}
+
+const PROTOCOL_TO_PORT_MAPPING: Record<string, string> = {
```
nikosdouvlis, line 48: 🙃 No need to explicitly type as this is using `as const`
```suggestion
const PROTOCOL_TO_PORT_MAPPING = {
```

nikosdouvlis (approved):

colinclerk:
Which vendors/proxies have we tested this with?

As part of definition of done here, can we write a docs page about the how reverse proxies must be configured to work properly in production?  What should the `X-Forwarded-*` headers be set to?

And, can we produce an error that's more helpful for debugging/support?  If the message said Referer=X while Host=Y it would be easier to catch at a glance.

desiprisg (approved):

packages/backend/src/util/request.ts
dimkl: Removing the type causes a TS error.
Screenshot 2023-05-31 at 11 49 35
Screenshot 2023-05-31 at 11 49 29

clerk-cookie:

PR #1266, opened 2023-05-31
fix(remix): Use location to verify navigation completion

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

When using a catchall route (`$.tsx`) in remix, the transition state `useTransition().state` remains idle during the navigation, so we are never able to resolve the promise we create in our `awaitableNavigate`.

I've replaced that with the `useLocation` hook, which is exported straight from the `react-router` package. This will rerender the component even if we navigate to the same url (in that case `location.key` will be different).

### DIFF
--- a/packages/remix/src/client/useAwaitableNavigate.tsx
+++ b/packages/remix/src/client/useAwaitableNavigate.tsx
@@ -1,24 +1,22 @@
-import { useNavigate, useTransition } from '@remix-run/react';
+import { useLocation, useNavigate } from '@remix-run/react';
 import React from 'react';
 
 type Resolve = (value?: unknown) => void;
 
 export const useAwaitableNavigate = () => {
   const navigate = useNavigate();
-  const transition = useTransition();
+  const location = useLocation();
   const resolveFunctionsRef = React.useRef<Resolve[]>([]);
-
   const resolveAll = () => {
-    if (transition.state !== 'idle') {
-      return;
-    }
     resolveFunctionsRef.current.forEach(resolve => resolve());
     resolveFunctionsRef.current.splice(0, resolveFunctionsRef.current.length);
   };
 
+  // location.key will change even when navigating to the same url,
+  // so we will successfully resolve in that case as well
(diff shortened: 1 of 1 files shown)


packages/remix/src/client/useAwaitableNavigate.tsx
```diff
@@ -1,24 +1,22 @@
...
     resolveFunctionsRef.current.forEach(resolve => resolve());
     resolveFunctionsRef.current.splice(0, resolveFunctionsRef.current.length);
   };
 
+  // location.key will change even when navigating to the same url,
+  // so we will successfully resolve in that case as well
   React.useEffect(() => {
     resolveAll();
-  }, [transition.state]);
+  }, [location]);
```
panteliselef, line 19: ❓ Should this be `location.key` ?

desiprisg: It could but it's used like that in react router docs so I think it's better this way. The key changing is just an observation.

panteliselef (approved):
I tested this with an app, works as expected 💯

nikosdouvlis (commented):
#til 

Good job, do we know when this changed?

nikosdouvlis (approved):

clerk-cookie:

PR #1272, opened 2023-06-01
fix(clerk-js): Only prefix url with the origin in buildUrlWithAuth

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This was resulting in unwanted behaviour in the case that the `to` string is a relative path. In that case we would prefix it with the full current path. 
Example:
`to = 'foo'` and we're on `http://localhost:4011/sign-in`
`buildUrlWithAuth` would return `http://localhost:4011/sign-in/foo` instead of `http://localhost:4011/foo`. This is an actual scenario when using magic links in dev instances.


SokratisVidros (approved):

nikosdouvlis:
@desiprisg let's add some tests before merging :)

clerk-cookie:

PR #1276, opened 2023-06-02
feat(nextjs): Return 401 Response on protected API routes

anagstef:
## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This commit introduces a new `apiRoutes` param on `authMiddleware`.

If `apiRoutes` is omitted, then the following heuristic is used: 
- If the route path is `['/api/(.*)', '/trpc/(.*)']` 
- or Request has `Content-Type: application/json` 
- or Request method is not-`GET,OPTIONS,HEAD`,

then this is considered an API route.

After this commit, a 401 JSON Response is returned on the following 3 cases:
- When the `authenticateRequest` function returns the `Unknown` state.
- When the `authenticateRequest` function returns the `Interstitial` state and the route is an API route.
- When the user is signed out, the route is protected (not public), and the route is an API route. _(this one happens inside our default implementation of `afterAuth`. If it is overwritten, then the user needs to handle this case themselves)_

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -86,7 +86,7 @@
     "@types/react-dom": "*",
     "expect-type": "^0.15.0",
     "jest": "*",
-    "node-fetch-native": "^0.1.8",
+    "node-fetch-native": "1.1.1",
     "ts-jest": "*",
     "typescript": "*"
   },
--- a/packages/nextjs/src/server/authMiddleware.ts
+++ b/packages/nextjs/src/server/authMiddleware.ts
@@ -11,7 +11,7 @@ import { DEV_BROWSER_JWT_MARKER, setDevBrowserJWTInURL } from './devBrowser';
 import { receivedRequestForIgnoredRoute } from './errors';
 import { isDevOrStagingUrl, redirectToSignIn } from './redirect';
 import type { NextMiddlewareResult, WithAuthOptions } from './types';
-import { decorateRequest, setRequestHeadersOnNextResponse } from './utils';
+import { apiEndpointUnauthorizedNextResponse, decorateRequest, setRequestHeadersOnNextResponse } from './utils';
 
 type WithPathPatternWildcard<T> = `${T & string}(.*)`;
 type NextTypedRoute<T = Parameters<typeof Link>['0']['href']> = T extends string ? T : never;
@@ -38,6 +38,10 @@ export const DEFAULT_CONFIG_MATCHER = ['/((?!.*\\..*|_next).*)', '/', '/(api|trp
  * This is the inverted version of DEFAULT_CONFIG_MATCHER.
  */
 export const DEFAULT_IGNORED_ROUTES = ['/((?!api|trpc))(_next|.+\\..+)(.*)'];
+/**
+ * Any routes matching this path will be treated as API endpoints by the middleware.
+ */
+export const DEFAULT_API_ROUTES = ['/api/(.*)', '/trpc/(.*)'];
 
 type RouteMatcherParam =
   | Array<RegExp | RouteMatcherWithNextTypedRoutes>
@@ -46,6 +50,7 @@ type RouteMatcherParam =
   | ((req: NextRequest) => boolean);
 
 type IgnoredRoutesParam = Array<RegExp | string> | RegExp | string | ((req: NextRequest) => boolean);
+type ApiRoutesParam = Array<RegExp | string> | RegExp | string | ((req: NextRequest) => boolean);
 
 type BeforeAuthHandler = (
   req: NextRequest,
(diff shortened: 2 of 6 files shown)


packages/nextjs/package.json
```diff
@@ -86,7 +86,7 @@
     "@types/react-dom": "*",
     "expect-type": "^0.15.0",
     "jest": "*",
-    "node-fetch-native": "^0.1.8",
+    "node-fetch-native": "1.1.1",
```
anagstef, line 89: This is updated because, the previous version does not implement the `Response.json` function. The dependency is only used for testing.

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -223,3 +243,31 @@ const appendDevBrowserOnDevOrStaging = (req: NextRequest, res: Response) => {
...
+};
+
+const isRequestContentTypeJson = (req: NextRequest): boolean => {
+  const requestContentType = req.headers.get(constants.Headers.ContentType);
+  return requestContentType === constants.ContentTypes.Json;
+};
+
+const isRequestMethodIndicatingApiRoute = (req: NextRequest): boolean => {
+  const requestMethod = req.method.toLowerCase();
+  return requestMethod !== 'get' && requestMethod !== 'head' && requestMethod !== 'options';
```
panteliselef, line 272: 🙃 Makes it more readable, less performant, merely a suggestion
```suggestion
  return !['get','head','options'].includes(requestMethod)
```

packages/nextjs/src/server/utils.ts
```diff
@@ -182,7 +182,8 @@ export function getRequestUrl({ request, relativePath }: { request: Request; rel
 
   // X-Forwarded-Proto could be 'https, http'
   const protocol =
-    (headers.get('X-Forwarded-Proto') ?? (headers as any)['X-Forwarded-Proto'])?.split(',')[0] ?? (url.protocol && url.protocol.replace(/\:$/, ''));
+    (headers.get('X-Forwarded-Proto') ?? (headers as any)['X-Forwarded-Proto'])?.split(',')[0] ??
+    (url.protocol && url.protocol.replace(/\:$/, ''));
```
panteliselef, line 186: ❓ Was this prettier, or is there a change I could not see here ?

packages/nextjs/src/server/authMiddleware.ts
dimkl, line 272: 🔧 i don't think we should include `get` hear. There may be an API route that we want to retrieve data with `get`.

dimkl (approved):
❓ should we also add the `isApiRoute` to the `auth` object ?

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -223,3 +243,31 @@ const appendDevBrowserOnDevOrStaging = (req: NextRequest, res: Response) => {
...
+//    or Request has `Content-Type: application/json`
+//    or Request method is not-GET,OPTIONS,HEAD,
+//    then this is considered an API route.
+//
+// - If the user has provided a specific `apiRoutes` prop in `authMiddleware` then all the above are discarded,
+//   and only routes that match the user’s provided paths are considered API routes.
+const createApiRoutes = (apiRoutes: RouteMatcherParam | undefined): ((req: NextRequest) => boolean) => {
+  if (apiRoutes) {
+    return createRouteMatcher(apiRoutes);
+  } else {
```
panteliselef, line 258: 🔧 `else` is redundant here

panteliselef (approved):

packages/nextjs/src/server/authMiddleware.ts
anagstef: @dimkl This is just an indication for our heuristic algorithm. If it's not a `GET` we are sure it's an API route, if it's a `GET` we don't know if this is a page or API endpoint, so it should return `false`.

nikosdouvlis:
!snapshot

nikosdouvlis (approved):

clerk-cookie:

PR #1304, opened 2023-06-06
fix(clerk-js): Add push/replace props & replace state when merging fragment into url

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

When using path routing and clicking on a sign-in/up link, we get rerouted to a hash based url which automatically gets converted to a path based url (that's happening inside PathRouter.tsx). After that, when the back button is pressed, we arrive at the hash-based url that got converted and not at the actual previous page we were on.

Solution: Add a `routerReplace` prop to `<ClerkProvider/>` and use that for merging the fragment into the url.

Edit: This PR also includes the new `routerPush` prop that replaces the `navigate` prop for v5.

Before: 

https://github.com/clerkinc/javascript/assets/73396808/436b9029-22b0-47ad-9fb1-d93135eaa8dd

After:

https://github.com/clerkinc/javascript/assets/73396808/818ed562-01f0-4013-80a5-832d9e3dabb2

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -31,6 +31,7 @@ import type {
   HandleOAuthCallbackParams,
   InstanceType,
   ListenerCallback,
+  NavigateOptions,
   OrganizationListProps,
   OrganizationProfileProps,
   OrganizationResource,
@@ -665,13 +666,14 @@ export class Clerk implements ClerkInterface {
     return unsubscribe;
   };
 
-  public navigate = async (to: string | undefined): Promise<unknown> => {
+  public navigate = async (to: string | undefined, options?: NavigateOptions): Promise<unknown> => {
     if (!to || !inBrowser()) {
       return;
     }
 
     const toURL = new URL(to, window.location.href);
-    const customNavigate = this.#options.navigate;
+    const customNavigate =
+      options?.replace && this.#options.routerReplace ? this.#options.routerReplace : this.#options.routerPush;
 
     if (toURL.origin !== window.location.origin || !customNavigate) {
       windowNavigate(toURL);
--- a/packages/gatsby-plugin-clerk/src/GatsbyClerkProvider.tsx
+++ b/packages/gatsby-plugin-clerk/src/GatsbyClerkProvider.tsx
@@ -21,7 +21,8 @@ export function ClerkProvider({ children, ...rest }: GatsbyClerkProviderProps) {
 
   return (
     <ReactClerkProvider
-      navigate={to => navigate(to)}
+      routerPush={to => navigate(to)}
+      routerReplace={to => navigate(to, { replace: true })}
       initialState={__clerk_ssr_state || {}}
       sdkMetadata={SDK_METADATA}
       telemetry={
--- a/packages/remix/src/client/RemixClerkProvider.tsx
+++ b/packages/remix/src/client/RemixClerkProvider.tsx
@@ -89,7 +89,8 @@ export function ClerkProvider({ children, ...rest }: RemixClerkProviderProps): J
   return (
     <ClerkRemixOptionsProvider options={mergedProps}>
       <ReactClerkProvider
-        navigate={(to: string) => awaitableNavigateRef.current?.(to)}
+        routerPush={(to: string) => awaitableNavigateRef.current?.(to)}
+        routerReplace={(to: string) => awaitableNavigateRef.current?.(to, { replace: true })}
         initialState={__clerk_ssr_state}
         sdkMetadata={SDK_METADATA}
         {...mergedProps}
--- a/packages/types/src/clerk.ts
+++ b/packages/types/src/clerk.ts
@@ -498,14 +498,15 @@ export type BuildUrlWithAuthParams = {
 
 // TODO: Make sure Isomorphic Clerk navigate can work with the correct type:
 // (to: string) => Promise<unknown>
-export type CustomNavigation = (to: string) => Promise<unknown> | void;
+export type CustomNavigation = (to: string, options?: NavigateOptions) => Promise<unknown> | void;
 
 export type ClerkThemeOptions = DeepSnakeToCamel<DeepPartial<DisplayThemeJSON>>;
 
 export interface ClerkOptions {
   appearance?: Appearance;
   localization?: LocalizationResource;
-  navigate?: (to: string) => Promise<unknown> | unknown;
+  routerPush?: (to: string) => Promise<unknown> | unknown;
+  routerReplace?: (to: string) => Promise<unknown> | unknown;
   polling?: boolean;
   selectInitialSession?: (client: ClientResource) => ActiveSessionResource | null;
   /** Controls if ClerkJS will load with the standard browser setup using Clerk cookies */
@@ -537,6 +538,10 @@ export interface ClerkOptions {
       };
 
   sdkMetadata?: SDKMetadata;
+}
+
+export interface NavigateOptions {
+  replace?: boolean;
 }
 
 export interface Resources {
--- a/playground/vite-react-ts/src/App.tsx
+++ b/playground/vite-react-ts/src/App.tsx
@@ -1,7 +1,3 @@
-import reactLogo from './assets/react.svg';
-import viteLogo from '/vite.svg';
-import clerkLogo from '/clerk.svg';
-import './App.css';
 import {
   ClerkProvider,
   RedirectToSignIn,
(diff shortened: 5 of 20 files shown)


panteliselef (approved):

colinclerk:
This is a good change, but there's a reasonable chance it will lead to errors. The issue is we can't predict if we're setting an appropriate state variable for the framework's router.

We probably won't see an issue unless we:
* Call `window.history.replaceState`
* Complete the first step of sign-up
* Then click the back button, to return to the start step of the sign-up

The back button will fire the Window [popstate event](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event), which the framework's router will be listening to so it can update it's own internal state. At that point, it's possible the framework can get confused by our manual entry via `window.history.replaceState` and throw an error.

Passing `window.history.state` is a good countermeasure and I suspect it will work for navigations to the root of the component, but I'd still worry about navigations to the non-root (for example, `/sign-in#/choose` for afterSignOutOne).

This issue is why we don't call `window.history.pushState` directly, but instead use the framework's wrapper of pushState (e.g. `router.push` in Next.js).

The safer approach for this fix is to use the framework's wrapper around `window.history.replaceState`, too (e.g. `router.replace` in Next.js)

I've been thinking we should accept both `push` and `replace` variants to `<ClerkProvider>` in v5, but if there's urgency in v4 we can find a way.

desiprisg:
@colinclerk We'll do some more tests to make sure this works, if we're going to go ahead with it. I didn't think accepting a `replace` variant was an option right now as well.

packages/clerk-js/src/ui/router/__tests__/PathRouter.test.tsx
```diff
@@ -54,6 +54,7 @@ describe('PathRouter', () => {
   beforeAll(() => {
     // @ts-ignore
     delete window.location;
+    history.replaceState = jest.fn();
```
dimkl, line 57: 🙃 could you please add an assertion that it's being called with the expected values?

dimkl (approved):

packages/types/src/clerk.ts
```diff
@@ -525,14 +525,15 @@ export type BuildUrlWithAuthParams = {
...
+export type CustomNavigation = (to: string, options?: NavigateOptions) => Promise<unknown> | void;
 
 export type ClerkThemeOptions = DeepSnakeToCamel<DeepPartial<DisplayThemeJSON>>;
 
 export interface ClerkOptions {
   appearance?: Appearance;
   localization?: LocalizationResource;
-  navigate?: (to: string) => Promise<unknown> | unknown;
+  push?: (to: string) => Promise<unknown> | unknown;
+  replace?: (to: string) => Promise<unknown> | unknown;
```
panteliselef, line 536: ❓Shouldn't we also add a deprecation warning in v4 ?

packages/remix/src/client/RemixClerkProvider.tsx
```diff
@@ -84,7 +84,7 @@ export function ClerkProvider({ children, ...rest }: RemixClerkProviderProps): J
   return (
     <ClerkRemixOptionsProvider options={mergedProps}>
       <ReactClerkProvider
-        navigate={(to: string) => awaitableNavigateRef.current?.(to)}
+        push={(to: string) => awaitableNavigateRef.current?.(to)}
```
panteliselef, line 87: ❓Don't we need to pass replace ?
```suggestion
      push={(to: string) => awaitableNavigateRef.current?.(to)}
      replace={(to: string) => awaitableNavigateRef.current?.(to, { replace:true })}
```

packages/nextjs/src/app-beta/client/ClerkProvider.tsx
```diff
@@ -59,9 +59,12 @@ export const useAwaitableNavigate = () => {
  */
 export function ClerkProvider(props: ClerkProviderProps) {
   const navigate = useAwaitableNavigate();
+  const router = useRouter();
+
   return (
     <ReactClerkProvider
-      navigate={navigate}
+      push={navigate}
+      replace={(to: string) => router.replace(to)}
```
panteliselef, line 67: If these changes only go to v5, i think you can skip this, becase we will be dropping the `/app-beta` subpath.

packages/gatsby-plugin-clerk/src/GatsbyClerkProvider.tsx
```diff
@@ -25,7 +25,7 @@ export function ClerkProvider({ children, ...rest }: GatsbyClerkProviderProps) {
 
   return (
     <ReactClerkProvider
-      navigate={to => navigate(to)}
+      push={to => navigate(to)}
```
panteliselef, line 28: ❓
```suggestion
      push={to => navigate(to)}
      replace={to => navigate(to, {replace: true})}
```

playground/vite-react-ts/src/App.tsx
```diff
@@ -67,7 +67,7 @@ function ClerkProviderWithRoutes() {
   return (
     <ClerkProvider
       publishableKey={clerkPubKey}
-      navigate={to => navigate(to)}
+      push={to => navigate(to)}
```
panteliselef, line 70: ❓ Shall we pass replace ?
```suggestion
      push={to => navigate(to)}
      repalce={to => navigate(to, {replace: true} )}
```

packages/types/src/clerk.ts
desiprisg: will do

BRKalow (approved):
Implementation looks good, seems like CI is failing though.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -665,13 +666,13 @@ export class Clerk implements ClerkInterface {
...
 
-  public navigate = async (to: string | undefined): Promise<unknown> => {
+  public navigate = async (to: string | undefined, options?: NavigateOptions): Promise<unknown> => {
     if (!to || !inBrowser()) {
       return;
     }
 
     const toURL = new URL(to, window.location.href);
-    const customNavigate = this.#options.navigate;
+    const customNavigate = options?.replace && this.#options.replace ? this.#options.replace : this.#options.push;
```
dimkl, line 675: This got me a bit confused! We have `options.replace: boolean` and `this.#options.replace: () => void`. I am the only one? I would suggest we rename one of those 2 to something else.

package-lock.json
dimkl, line 1: 🔧  What caused those changes? Are we sure we want to merge this? I think we should drop the changes since there is no change in the any `package.json` file.

dimkl (commented):
Please also add the changes to the `Upgrade v4 -> v5 Guide [DRAFT]`.

packages/clerk-js/src/core/clerk.ts
BRKalow: @dimkl I agree this is a bit confusing, but I think the usage makes more sense:

```ts
const clerk = new Clerk({ push: router.push, replace: router.replace })

// calls this.#options.push
clerk.navigate('/')

// calls this.#options.replace
clerk.navigate('/', { replace: true })
```

I'm not sure what would we rename either to considering this. 🤔

dimkl: Should we consider naming the `#options` as `historyReplace` / `historyPush`?
This suggestion comes from the comment on the `SDK-683` ticket. 
@desiprisg  Also doesn't this PR resolve that ticket?

package-lock.json
nikosdouvlis: + 1

nikosdouvlis:
@desiprisg let's please add at least one e2e test to cover this.

package-lock.json
desiprisg: My bad

packages/clerk-js/src/core/clerk.ts
desiprisg: Will coordinate with Colin and rename.

PR #1307, opened 2023-06-06
Add ability for user to delete themselves if permission present

jescalan:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [x] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

- When permissions are present, users will see a section in their settings allowing them to delete their account. User deletion permissions are controlled via the Clerk dashboard.
- A `createOrganizationEnabled` property was also added to users in anticipation of a feature allowing permission-gated org creation that should be landing in the next couple days.
- When a user's account is deleted, they are signed out and nav to the home page.

### DIFF
--- a/packages/clerk-js/package.json
+++ b/packages/clerk-js/package.json
@@ -37,7 +37,8 @@
     "start": "echo \"Noop\"",
     "test": "jest",
     "test:ci": "jest --maxWorkers=50%",
-    "test:coverage": "jest --collectCoverage && open coverage/lcov-report/index.html"
+    "test:coverage": "jest --collectCoverage && open coverage/lcov-report/index.html",
+    "watch": "webpack --config webpack.config.js --env production --watch"
   },
   "dependencies": {
     "@clerk/localizations": "^1.18.1",
--- a/packages/clerk-js/playground.html
+++ b/packages/clerk-js/playground.html
@@ -1,0 +1,170 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script>
+      // 👇 PUT YOUR API KEY HERE 👇
+      window.__clerk_frontend_api = '';
+      // 👆 PUT YOUR API KEY HERE 👆
+    </script>
+    <script src="./dist/clerk.browser.js"></script>
+    <style>
+      body {
+        font-family: sans-serif;
+        margin: 0;
+        padding: 0;
+        -webkit-font-smoothing: antialiased;
+      }
+
+      .text {
+        line-height: 1.5em;
+      }
+
+      #errors p {
+        color: red;
+        padding: 20px 25px;
+        background: #faecec;
+        border-radius: 4px;
+        margin: 0 auto;
+        max-width: 960px;
+      }
+
+      header {
+        background: #3e3e3e;
+        color: white;
+      }
+
+      header .wrapper {
+        display: flex;
+        justify-content: space-between;
+        align-items: center;
+        margin: 0 auto;
+        max-width: 960px;
+        min-height: 32px;
+      }
+
+      .user-info {
+        display: flex;
+        align-items: center;
+      }
+
+      #username {
+        margin-right: 20px;
+      }
+
+      .buttons {
+        display: flex;
+      }
+
+      #orgswitcher {
+        margin-right: 20px;
+      }
+
+      main {
+        margin: 50px auto;
+        max-width: 960px;
+      }
+
+      button {
+        appearance: none;
+        padding: 10px 20px;
+        border-radius: 3px;
+        border: none;
+        background: #4a7ad8;
+        color: white;
+        font-weight: bold;
+        cursor: pointer;
+      }
+
+      button:hover {
+        background: #6391ec;
+      }
+    </style>
+  </head>
+  <body>
+    <header>
+      <div class="wrapper">
+        <h3>Clerk JS Playground</h3>
+        <div class="buttons">
+          <div id="orgswitcher"></div>
+          <div class="user-info">
+            <div id="username"></div>
+            <div id="userbutton"></div>
+          </div>
+        </div>
+      </div>
+    </header>
+    <main>
+      <p class="text">
+        A simple playground for tinkering with vanilla js clerk and testing changes to components. To get started,
+        create an application on Clerk and set your API key to the <code>window.__clerk_frontend_api</code> variable at
+        the top of the source code.
+      </p>
+      <p>Use the buttons below to trigger basic functionality. Feel free to add more buttons as well!</p>
+
+      <button onClick="openSignUp()">Sign Up</button>
+      <button onClick="openSignIn()">Sign In</button>
+      <button onClick="showUserProfile()">User Profile</button>
+      <button onClick="openCreateOrganization()">Create Org</button>
+      <button onClick="showOrganizationProfile()">Org Profile</button>
+      <button onClick="signOut()">Sign Out</button>
+    </main>
+
+    <div id="errors"></div>
+    <div id="stage"></div>
+
+    <script>
+      const stage = document.getElementById('stage');
+      // clerkjs functions return null when there's an error and use console.info for reporting
+      // i recommend changing this, but for now we patch console.info to be able to display the err
+      const consoleInfo = console.info.bind(console);
+      console.info = message => {
+        const errs = document.getElementById('errors');
+        errs.innerHTML = '';
+        errs.append(el('p', message));
+        consoleInfo(message);
+      };
+
+      function openSignIn() {
+        Clerk.openSignIn();
+      }
+
+      function openSignUp() {
+        Clerk.openSignUp();
+      }
+
+      function signOut() {
+        Clerk.signOut();
+      }
+
+      function openCreateOrganization() {
+        Clerk.openCreateOrganization();
+      }
+
+      function showUserProfile() {
+        Clerk.mountUserProfile(stage);
+      }
+
+      function showOrganizationProfile() {
+        Clerk.mountOrganizationProfile(stage);
+      }
+
+      // this does not appear to be working
+      Clerk.load().then(() => {
+        if (Clerk.user) {
+          document.getElementById('username').innerText = Clerk.user.primaryEmailAddress.emailAddress;
+          Clerk.mountUserButton(document.getElementById('userbutton'));
+        }
+        if (Clerk.organization) {
+          Clerk.mountOrganizationSwitcher(document.getElementById('orgswitcher'));
+        }
+      });
+
+      // utility fn
+      function el(tagName, content) {
+        const tag = document.createElement(tagName);
+        tag.innerText = content;
+        return tag;
+      }
+    </script>
+  </body>
+</html>
--- a/packages/clerk-js/src/core/resources/Organization.ts
+++ b/packages/clerk-js/src/core/resources/Organization.ts
@@ -26,6 +26,7 @@ export class Organization extends BaseResource implements OrganizationResource {
   slug!: string;
   logoUrl!: string;
   imageUrl!: string;
+  adminDeleteEnabled!: boolean;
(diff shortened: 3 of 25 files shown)


jescalan:
I don't think we need a changeset for this alone, but I can continue to actually implement the effects of these properties on this PR if desired, and that would definitely merit one.

packages/types/src/organization.ts
```diff
@@ -32,6 +32,7 @@ export interface OrganizationResource extends ClerkResource {
   membersCount: number;
   pendingInvitationsCount: number;
   publicMetadata: OrganizationPublicMetadata;
+  adminDeleteEnabled: boolean;
```
SokratisVidros, line 35: ❓ Shouldn't we parse this setting in /core/resources/Organization.ts?

jescalan: 🤦‍♂️ i for some reason did not push this change - ty!

SokratisVidros (approved):

packages/clerk-js/package.json
```diff
@@ -36,7 +36,8 @@
     "lint": "eslint .",
     "start": "echo \"Noop\"",
     "test": "jest",
-    "test:coverage": "jest --collectCoverage && open coverage/lcov-report/index.html"
+    "test:coverage": "jest --collectCoverage && open coverage/lcov-report/index.html",
+    "watch": "webpack --config webpack.config.js --env production --watch"
```
jescalan, line 40: Added this command to pair with the playground file I added here while developing the feature. When running this in the terminal, it will rebuild changed files more quickly than running a full build after every save.

packages/clerk-js/playground.html
```diff
@@ -0,0 +1,170 @@
+<!DOCTYPE html>
```
jescalan, line 1: This playground file is something I built alongside the feature to help me to develop it. This requires little to no setup - you just add your API key, open the file, and its off to the races. I also feel like having a pure js testbed for things is helpful in general - I turned up a number of bugs that I bet are not present in higher abstractions like react & nextjs libs.

packages/clerk-js/src/core/resources/Organization.ts
```diff
@@ -169,6 +170,7 @@ export class Organization extends BaseResource implements OrganizationResource {
     this.publicMetadata = data.public_metadata;
     this.membersCount = data.members_count;
     this.pendingInvitationsCount = data.pending_invitations_count;
+    this.adminDeleteEnabled = data.admin_delete_enabled;
```
jescalan, line 173: I kind of snuck this addition in here - the original PR was just too add these new properties as they were finished in the backend, but it expanded to covering the user delete feature. We decided to move the other two paired features, allowing admins to delete organizations, and gating organization creation permissions for users, to separate PRs, which I will be submitting soon.

If you want, I can move these out to the other PRs to keep the history squeaky clean, just let me know.

packages/clerk-js/src/core/resources/User.ts
```diff
@@ -80,6 +80,8 @@ export class User extends BaseResource implements UserResource {
   publicMetadata: UserPublicMetadata = {};
   unsafeMetadata: UserUnsafeMetadata = {};
   lastSignInAt: Date | null = null;
+  createOrganizationEnabled: boolean = false;
```
jescalan, line 83: Same here with the leaked properties note above

packages/clerk-js/src/core/resources/User.ts
```diff
@@ -315,6 +321,9 @@ export class User extends BaseResource implements UserResource {
     this.backupCodeEnabled = data.backup_code_enabled;
     this.twoFactorEnabled = data.two_factor_enabled;
 
+    this.createOrganizationEnabled = data.create_organization_enabled;
```
jescalan, line 324: ...and here

packages/clerk-js/src/core/resources/UserSettings.ts
```diff
@@ -65,6 +67,7 @@ export class UserSettings extends BaseResource implements UserSettingsResource {
     this.attributes = Object.fromEntries(
       Object.entries(data.attributes).map(a => [a[0], { ...a[1], name: a[0] }]),
     ) as Attributes;
+    this.actions = data.actions;
```
jescalan, line 70: This small modification allows the environment context to have access to this new "actions" property which contains the values for permissions associated with user actions, which are needed for the user delete feature. Big thanks for @SokratisVidros for helping me to find this and patch this value into the right place.

packages/clerk-js/src/ui/components/UserProfile/DeletePage.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const session = useCoreSession();
+  const user = useCoreUser();
+
+  const deleteUser = async () => {
+    try {
+      await user.delete();
+      await clerk.signOut({ sessionId: session.id });
+      router.navigate(environment.displayConfig.homeUrl);
+      // TODO: for routerless mode, explicitly open the "choose" route
+      clerk.openSignIn();
```
jescalan, line 22: Want to add a little extra context here - after deleting a user, you get signed out, and the router will navigate to the home route, which should contain the sign in UI, in theory. If you have a multi-session app, the sign in UI should allow the user to select the next available session data to log into.

However, if the implementation is just using pure JS and is not associated with a router, the navigation will do nothing other than changing the url hash, so we explicitly open the sign in window. This is a slight bug IMO for router-less apps, since we want to show the "choose" route within the sign in UI to allow to user to choose other sessions if there are any. If you refresh the page, it will automatically log into any other available sessions for a multi-session app, but a refresh shouldn't be necessary.

packages/localizations/src/en-US.ts
```diff
@@ -354,6 +354,12 @@ export const enUS: LocalizationResource = {
         destructiveActionSubtitle: 'Remove this web3 wallet from your account',
         destructiveAction: 'Remove wallet',
       },
+      dangerSection: {
+        title: 'Danger',
+        deleteAccountButton: 'Delete Account',
+        deleteAccountTitle: 'Delete Account',
+        deleteAccountDescription: 'Delete your account and all its associated data',
+      },
```
jescalan, line 362: Just as a note, I only ran the english translation here. I can probably cover a couple more but I'm not sure how we normally tackle translation.

jescalan (commented):
Self-review as this is my first PR to Clerk!

packages/clerk-js/playground.html
panteliselef: This is great, I wonder if we should move this inside the top-level `playground` dir instead

packages/localizations/src/en-US.ts
panteliselef: 😀 We officially support only the english translation. It's a personal decision if you want to do the rest as well.

packages/clerk-js/src/core/resources/Organization.ts
SokratisVidros: We can leave it as is. No need for extra work.

packages/clerk-js/src/ui/components/UserProfile/DeletePage.tsx
SokratisVidros: Is this fix targeted for multi-session or single-session apps? The sign-in choose screen is only available in multi-session apps.

This behavior is opinionated when it comes to UX. ClerkJS supports [afterSignOut props](https://github.com/clerkinc/javascript/blob/24915e26581a39651f389ded9d1ae0c6533e167f/packages/clerk-js/src/ui/components/UserButton/useMultisessionActions.tsx#L27) that should also be used in this case as they are the main control for Clerk users to handle post sign-out behavior.

Most importantly, I'd also that FAPI signs out the user during `await user.delete`. So, the `clerk.signOut` shouldn't be necessary as the payload of `user.delete` should purge the current session, and <UserProfile/> should unmount. Let me double-check that.

packages/clerk-js/src/ui/components/UserProfile/DeletePage.tsx
```diff
@@ -0,0 +1,44 @@
...
+    } catch (e) {
+      handleError(e, [], card.setError);
+    }
+  };
+
+  return (
+    <>
```
SokratisVidros, line 29: ❓ Is this fragment required?

SokratisVidros (changes requested):

packages/clerk-js/package.json
nikosdouvlis: Makes sense to me. Just curious, does HMR work correctly when using `playground.html`? I *think* HMR won't work unless the file is served under `https` as HMR relies on webhooks.

packages/clerk-js/src/core/resources/User.ts
nikosdouvlis: ❓ Are these named based on the FAPI payload? Another naming that'd make sense to me would be `canCreateOrganization` but I'm fine with `createOrganizationEnabled` as well.

nikosdouvlis (approved):

jescalan:
@SokratisVidros removed the unneeded fragment, tested whether `user.delete()` will kill the session and it does, so removed clerk log out line!

packages/clerk-js/src/core/resources/User.ts
SokratisVidros: Yes, it is based on FAPI. We iterated a lot on naming these parameters. I can't recall why we didn't pick the `canX` proposal that was discussed at some point.

@mzhong9723 @jescalan Do you maybe remember?

@nikosdouvlis FYI [this is the design doc](https://www.notion.so/clerkdev/Allow-our-users-to-delete-themselves-and-their-orgs-94180a172b0e428c9352835274fa8bfe). You can also check the comments section for naming decisions.

packages/clerk-js/src/ui/components/UserProfile/DeletePage.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const DeletePage = withCardStateProvider(() => {
+  const card = useCardState();
+  const environment = useEnvironment();
+  const router = useRouter();
+  const user = useCoreUser();
+
+  const deleteUser = async () => {
+    try {
+      await user.delete();
+      router.navigate(environment.displayConfig.homeUrl);
```
SokratisVidros, line 17:
```suggestion
      await router.navigate(environment.displayConfig.afterSignOutUrl);
```

SokratisVidros (approved):
Nicely done @jescalan.

One last comment about the router.navigate and it's good to go!

packages/clerk-js/src/core/resources/User.ts
jescalan: as far as `canX`, I'm not sure to be honest, I just mirrored the property name from bapi for consistency

packages/clerk-js/src/core/resources/User.ts
```diff
@@ -224,6 +224,10 @@ export class User extends BaseResource implements UserResource {
     });
   };
 
+  delete = (): Promise<void> => {
+    return this._baseDelete({ path: '/me' });
```
jescalan, line 228: Just want to flag this, as it's a somewhat non-standard path. Once we fully roll this out in the API there's no going back without a major version boost.

clerk-cookie:

PR #1321, opened 2023-06-08
fix(backend): Fix infinite redirect loop in localhost apps

dimkl:
This issue was caused by the fix introduced for Railway App with NextJS in which there are more forwarded protos than forwarded ports causing an issue in constructing the final url origin.

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

The fix for `Railway` (checking of forwarded-port is the same as the forwarded-proto)  was causing an infinite redirect loop when running a NextJS application in `localhost:{port}` since the `{port}` is passed as `forwarded-port` which is different from the `forwarded-proto` (equals to http).
The issue is fixed by changing the logic of checking for cross-origin to use the forwarded proto only when there are more forwarded proto than forwarded ports (which is the case of Railway).


nikosdouvlis (approved):

clerk-cookie:

PR #1329, opened 2023-06-09
feat(backend,clerk-sdk-node): Simplify the authenticateRequest signature

anagstef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `@clerk/chrome-extension`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

- [x] `npm test` runs as expected.
- [x] `npm run build` runs as expected.

This PR simplifies the signature of `authenticateRequest` for better DX on both `@clerk/backend` and `@clerk/clerk-sdk-node` packages:
- One pair of legacy or new instance keys are required now and not all 4 of them
- `@clerk/backend` now can handle the `"Bearer "` prefix in Authorization header for better DX
- `host` parameter is now optional in `@clerk/backend`

## Example of using `authenticateRequest`

Before:
```ts
import clerkClient from '@clerk/clerk-sdk-node';

// `Bearer asdgfjdkhgmsfngl`
const authHeader = request.headers.get('authorization') || '';
const authToken = authHeader.split(` `)[1];

const res = await clerkClient.authenticateRequest({
  headerToken: authToken,
  publishableKey: process.env.CLERK_PUB_KEY,
  secretKey: process.env.CLERK_SECRET_KEY,
  apiKey: ``,
  frontendApi: ``,
  host: ``,
});

...
```

After this PR:
```ts
import clerkClient from '@clerk/clerk-sdk-node';

const res = await clerkClient.authenticateRequest({
  headerToken: request.headers.get('authorization'),
  publishableKey: process.env.CLERK_PUB_KEY,
  secretKey: process.env.CLERK_SECRET_KEY
});

...
```

PR #1348, opened 2023-06-13
Introduce `RequestAdapter` interface for `authenticateRequest`

anagstef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [x] `@clerk/fastify`
- [x] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### This PR is built upon changes of #1329 

This PR introduces a new `RequestAdapter` interface that may be implemented by any framework that uses the  `@clerk/backend` package. The custom `RequestAdapter` instance will be passed to `authenticateRequest` instead of each header/cookie/searchParams as a prop.

Example usage:
```ts
// CustomRequestAdapter.ts
// How we access the headers/cookies/searchParams depends on the specific framework's request representation
export class CustomRequestAdapter implements RequestAdapter {
  constructor(readonly req: CustomRequest) {}

  headers(key: string) {
    return this.req?.headers?.[key];
  }
  cookies(key: string) {
    return this.req?.cookies?.[key];
  }
  searchParams() {
    return new URL(this.req.url).searchParams;
  }
}
```
```ts
// FrameworkClerkMiddleware.ts
...
export const withClerkMiddleware = (options: ClerkCustomOptions) => {
  return async (req: CustomFrameworkRequest, ...) => {
    const requestState = await clerkClient.authenticateRequest({
      ...options,
      secretKey,
      publishableKey,
      requestAdapter: new CustomRequestAdapter(req),
    });
   ...
};
```

### DIFF
--- a/packages/backend/src/tokens/request.ts
+++ b/packages/backend/src/tokens/request.ts
@@ -37,9 +37,59 @@ export type OptionalVerifyTokenOptions = Partial<
   Pick<VerifyTokenOptions, 'authorizedParties' | 'clockSkewInSeconds' | 'jwksCacheTtlInMs' | 'skipJwksCache' | 'jwtKey'>
 >;
 
-export type AuthenticateRequestOptions = RequiredVerifyTokenOptions &
+type PublicKeys =
+  | {
+      publishableKey: string;
+      /**
+       * @deprecated Use `publishableKey` instead.
+       */
+      frontendApi: never;
+    }
+  | {
+      publishableKey: never;
+      /**
+       * @deprecated Use `publishableKey` instead.
+       */
+      frontendApi: string;
+    }
+  | {
+      publishableKey: string;
+      /**
+       * @deprecated Use `publishableKey` instead.
+       */
+      frontendApi: string;
+    };
+
+type SecretKeys =
+  | {
+      secretKey: string;
+      /**
+       * @deprecated Use `secretKey` instead.
+       */
+      apiKey: never;
+    }
+  | {
+      secretKey: never;
+      /**
+       * @deprecated Use `secretKey` instead.
+       */
+      apiKey: string;
+    }
+  | {
+      secretKey: string;
+      /**
+       * @deprecated Use `secretKey` instead.
+       */
+      apiKey: string;
+    };
+
+export type InstanceKeys = PublicKeys & SecretKeys;
+
+export type AuthenticateRequestOptions = InstanceKeys &
   OptionalVerifyTokenOptions &
   LoadResourcesOptions & {
+    apiVersion?: string;
+    apiUrl?: string;
     /* Client token cookie value */
     cookieToken?: string;
     /* Client uat cookie value */
@@ -48,12 +98,8 @@ export type AuthenticateRequestOptions = RequiredVerifyTokenOptions &
     headerToken?: string;
     /* Request origin header value */
     origin?: string;
-    /* Clerk frontend Api value */
-    frontendApi: string;
-    /* Clerk Publishable Key value */
-    publishableKey: string;
     /* Request host header value */
-    host: string;
+    host?: string;
     /* Request forwarded host value */
     forwardedHost?: string;
     /* Request forwarded port value */
@@ -102,6 +148,7 @@ export async function authenticateRequest(options: AuthenticateRequestOptions): 
   options.frontendApi = parsePublishableKey(options.publishableKey)?.frontendApi || options.frontendApi || '';
   options.apiUrl = options.apiUrl || API_URL;
   options.apiVersion = options.apiVersion || API_VERSION;
+  options.headerToken = options.headerToken?.replace('Bearer ', '');
 
   assertValidSecretKey(options.secretKey || options.apiKey);
 
--- a/packages/backend/src/tokens/request.ts
+++ b/packages/backend/src/tokens/request.ts
@@ -1,4 +1,4 @@
-import { API_URL, API_VERSION } from '../constants';
+import { API_URL, API_VERSION, constants } from '../constants';
 import { assertValidSecretKey } from '../util/assertValidSecretKey';
 import { isDevelopmentFromApiKey } from '../util/instance';
 import { parsePublishableKey } from '../util/parsePublishableKey';
@@ -22,6 +22,31 @@ import {
   satelliteInDevReturningFromPrimary,
 } from './interstitialRule';
 import type { VerifyTokenOptions } from './verify';
+
+export interface RequestAdapter {
+  /**
+   * The headers method returns the value of the specified header. Keys given to this method are
+   * normalized to lowercase, as they exist in `packages/backend/src/constants.ts`.
+   * Implementing this method requires the user to manipulate the header keys accordingly.
+   *
+   * @param key The header name to retrieve
+   */
+  headers(key: string): string | undefined;
+
+  /**
+   * The cookies method returns the value of the specified cookie.
+   * Keys given to this method exist in `packages/backend/src/constants.ts`.
+   *
+   * @param key The cookie name to retrieve
+   */
+  cookies(key: string): string | undefined;
+
+  /**
+   * The searchParams method returns a URLSearchParams object that contains the query parameters
+   * of the request.
+   */
+  searchParams(): URLSearchParams | undefined;
+}
 
 export type LoadResourcesOptions = {
   loadSession?: boolean;
@@ -130,6 +155,7 @@ export type AuthenticateRequestOptions = InstanceKeys &
      * @experimental
      */
     signInUrl?: string;
+    requestAdapter?: RequestAdapter;
   };
 
 function assertSignInUrlExists(signInUrl: string | undefined, key: string): asserts signInUrl is string {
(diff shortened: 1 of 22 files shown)


packages/backend/src/tokens/request.ts
```diff
@@ -145,10 +152,29 @@ function assertProxyUrlOrDomain(proxyUrlOrDomain: string | undefined) {
...
+    forwardedHost:
+      options.forwardedHost || options.requestAdapter?.headers(constants.Headers.ForwardedHost) || undefined,
+    forwardedPort:
+      options.forwardedPort || options.requestAdapter?.headers(constants.Headers.ForwardedPort) || undefined,
+    forwardedProto:
+      options.forwardedProto || options.requestAdapter?.headers(constants.Headers.ForwardedProto) || undefined,
+    referrer: options.referrer || options.requestAdapter?.headers(constants.Headers.Referrer) || undefined,
+    userAgent: options.userAgent || options.requestAdapter?.headers(constants.Headers.UserAgent) || undefined,
+    searchParams: options.searchParams || options.requestAdapter?.searchParams() || undefined,
+  };
```
dimkl, line 177: ☁️ why not something like this:
```typescript
if (options.headerToken){
  options.headerToken = stripAuthorizationHeader(options.headerToken)
}

options.frontendApi = parsePublishableKey(options.publishableKey)?.frontendApi || options.frontendApi

const requestOptions = {
  headerToken: options.requestAdapter?.headers(constants.Headers.Authorization),
  cookieToken: options.requestAdapter?.cookies(constants.Cookies.Session),
  clientUat: options.requestAdapter?.cookies(constants.Cookies.ClientUat),
  origin: options.requestAdapter?.headers(constants.Headers.Origin),
  host: options.requestAdapter?.headers(constants.Headers.Host),
  forwardedHost: options.requestAdapter?.headers(constants.Headers.ForwardedHost),
  forwardedPort: options.requestAdapter?.headers(constants.Headers.ForwardedPort),
  forwardedProto: options.requestAdapter?.headers(constants.Headers.ForwardedProto),
  referrer: options.requestAdapter?.headers(constants.Headers.Referrer),
  userAgent: options.requestAdapter?.headers(constants.Headers.UserAgent),
  searchParams: options.requestAdapter?.searchParams(),
}

const defaultOptions = {
  apiUrl: API_URL,
  apiVersion: API_VERSION,
}

mergeWithPriority(options, requestOptions, defaultOptions)



const mergeWithPriority = (...records) => {
  const keys = new Set(records.map(r => Object.keys(r)).flat());
  const mergedRecords = {};

  keys.map(key => {
    record = records.find(r => r[key]);
    mergedRecords[key] = record && record[key];
  });

  return mergedRecords;
}
```

dimkl (approved):

packages/sdk-node/src/authenticateRequest.ts
```diff
@@ -102,7 +89,7 @@ export const decorateResponseWithObservabilityHeaders = (res: ServerResponse, re
 const isDevelopmentFromApiKey = (apiKey: string): boolean =>
   apiKey.startsWith('test_') || apiKey.startsWith('sk_test_');
 
-const getRequestUrl = (req: IncomingMessage): URL => {
+export const getRequestUrl = (req: IncomingMessage): URL => {
```
dimkl, line 92: 🙃 please check if this also gets exported from the `sdk-node` . If it's get exported I would advise to move this to utils and import it here.

anagstef: @dimkl Thanks for mentioning it, but I think we don't have any issues with exports.

packages/backend/src/tokens/request.ts
```diff
@@ -23,6 +23,31 @@ import {
...
 
+export interface RequestAdapter {
+  /**
+   * The headers method returns the value of the specified header. Keys given to this method are
+   * normalized to lowercase, as they exist in `packages/backend/src/constants.ts`.
+   * Implementing this method requires the user to manipulate the header keys accordingly.
+   *
+   * @param key The header name to retrieve
+   */
+  headers(key: string): string | undefined;
```
nikosdouvlis, line 34: I'm somewhat concerned with the proposed API.
1. Why did we use our own interface for `headers` here? Is there a reason we cannot use [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers)?
2. The naming confuses me. I'd expect `headers` to be an object/instance, not a function that returns a string. I'd expect something along the lines of `headers.get` (see point 1)
3. How can we write a header? This is necessary for the loop detection mechanism

nikosdouvlis: I thought we'd be doing the following:
- clerk/backend polyfills fetch and Headers for all runtimes
- clerk/backend exposes a util to convert an object into a Headers instance 
- clerk/backend exposes an interface requiring at least a `headers` instance to be present.

packages/backend/src/tokens/request.ts
```diff
@@ -23,6 +23,31 @@ import {
 } from './interstitialRule';
 import type { VerifyTokenOptions } from './verify';
 
+export interface RequestAdapter {
+  /**
+   * The headers method returns the value of the specified header. Keys given to this method are
+   * normalized to lowercase, as they exist in `packages/backend/src/constants.ts`.
+   * Implementing this method requires the user to manipulate the header keys accordingly.
```
nikosdouvlis, line 30: Having the caller re-implement the headers spec is risky. Small discrepancies can cause serious issues that are very hard to debug when deployed. 

Also, we've been bitten by runtimes that don't respect the spec (Vercel runtime for edge API routes, it was fixed about 8 months ago by Vercel).

packages/backend/src/tokens/request.ts
```diff
@@ -23,6 +23,31 @@ import {
...
+   *
+   * @param key The cookie name to retrieve
+   */
+  cookies(key: string): string | undefined;
+
+  /**
+   * The searchParams method returns a URLSearchParams object that contains the query parameters
+   * of the request.
+   */
+  searchParams(): URLSearchParams | undefined;
```
nikosdouvlis, line 48: Can we accept a `URL` instead?

packages/fastify/src/utils.ts
```diff
@@ -1,3 +1,28 @@
+import type { RequestAdapter } from '@clerk/backend';
+import { constants } from '@clerk/backend';
+import { parse } from 'cookie';
+import type { FastifyRequest } from 'fastify';
+
 export const getSingleValueFromArrayHeader = (value?: string[] | string): string | undefined => {
   return (Array.isArray(value) ? value[0] : value) || undefined;
 };
+
+export class FastifyRequestAdapter implements RequestAdapter {
```
nikosdouvlis, line 10: Do we require using a class here? 
If not, let's use a simple function instead. It should return a `RequestAdapter` eg:
```ts
export const fastifyRequestAdapter = (req): RequestAdapter => {
   ...
} 
```

nikosdouvlis (changes requested):

packages/backend/src/tokens/request.ts
```diff
@@ -23,6 +23,31 @@ import {
...
+   */
+  headers(key: string): string | undefined;
+
+  /**
+   * The cookies method returns the value of the specified cookie.
+   * Keys given to this method exist in `packages/backend/src/constants.ts`.
+   *
+   * @param key The cookie name to retrieve
+   */
+  cookies(key: string): string | undefined;
```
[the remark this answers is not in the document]
anagstef: @nikosdouvlis Each framework does a separate thing for accessing the cookies, even using a lib for that. So, we kept the functionality for cookies as-is, giving each framework the flexibility to implement a way to access cookies.

packages/backend/src/tokens/request.ts
anagstef: @nikosdouvlis 
> Why did we use our own interface for headers here? Is there a reason we cannot use [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers)?

We want to give full control to each framework. `@clerk/backend` package only passes a header key and the implementation is totally up to each framework.

>The naming confuses me. I'd expect headers to be an object/instance, not a function that returns a string. I'd expect something along the lines of headers.get (see point 1)

I agree with the naming, we could rename to something like `getHeader`.

>How can we write a header? This is necessary for the loop detection mechanism

I think it's not. This is an adapter on how the `@clerk/backend` will have access to headers to call `authenticateRequest`. Writing on the headers is the framework's job. Am I missing a use case?

packages/backend/src/tokens/request.ts
anagstef: @nikosdouvlis We did not change the implementation for our supported frameworks. We just moved the logic inside the RequestAdapter.

packages/fastify/src/utils.ts
anagstef: @nikosdouvlis No we don't require a class. It could be changed with a simple function.

nikosdouvlis: Let's use with simple functions then!

packages/backend/src/tokens/request.ts
nikosdouvlis: Sorry, I'm still a bit confused. I'm referring to this comment:
> Implementing this method requires the user to manipulate the header keys accordingly

We can use clerk/backend to convert a `Record<string,string>` to a Headers instance without doing that manually. Is this not the preferred approach? Otherwise the method caller needs to reimplement Headers themselves

packages/backend/src/tokens/request.ts
nikosdouvlis: `RequestAdapter` can expect a `Headers` instance instead of this proprietary `headers` method. Each framework will have full control on how to transform data from its request to a `Headers` instance.

nikosdouvlis: Writing headers is needed in order to get the loop detection mechanism work for every framework without reimplementing the logic.

nikosdouvlis:
Closing in favor of #1393

clerk-cookie:

PR #1380, opened 2023-06-19
fix(clerk-js): Also include the redirect_url query for create/leave org redirects

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Include the redirect_url query param in all redirect actions. Also refactored the pickRedirectionProp util to be more versatile. It now accepts a keys array, a sources array, a validator and a formatter for the keys. It will check for valid urls and will fallback to the next key if a check fails.


packages/clerk-js/src/utils/authPropHelpers.ts
```diff
@@ -1,69 +1,36 @@
...
-    (accessRedirectUrl ? ctx?.redirectUrl : undefined) ||
-    options?.[key] ||
-    displayConfig?.[key];
-
-  if (!isValidUrl(url, { includeRelativeUrls: true }) || hasBannedProtocol(url)) {
-    return '';
-  }
-
-  return url;
+export const pickUrl = (key: string | string[], source: Record<string, any>, opts?: PickUrlOptions): string => {
```
anagstef, line 11: 🔧  Options for `pickUrl` util could be an object like:

```ts
type PickUrlParams = {
  key: string | string[];
  source: Record<string, any>;
  opts?: PickUrlOptions
}
```

anagstef, line 11: 💡 Should we also add a new comment explaining in short how this works (?)

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
```diff
@@ -38,26 +39,22 @@ export const useSignUpContext = (): SignUpContextType => {
...
-    pickRedirectionProp('afterSignUpUrl', {
-      queryParams,
-      ctx,
-      options,
-      displayConfig,
-    }),
+    pickUrl(['afterSignUpUrl', 'redirectUrl'], queryParams, {
+      validator: url => isAllowedRedirectOrigin(url, options.allowedRedirectOrigins),
+      formatter: camelToSnake,
+    }) || pickUrl('afterSignUpUrl', [ctx, options, displayConfig]),
```
anagstef, line 45: ❓ The priority for this before refactoring was:

1. queryParams.afterSignUpUrl
2. queryParams.redirectUrl
3. ctx.afterSignUpUrl
4. **ctx.redirectUrl**
5. options.afterSignUpUrl
6. displayConfig.afterSignUpUrl

Is the priority still respected? Especially `ctx.redirectUrl`.

desiprisg: Yes it should. For every source, we check all the values, in that order.

dimkl (commented):
☁️ We have already tried 1 time to refactor and since I am not sure that the current it the previous refactoring attempts make it more readable, I would like to propose creating a document with suggesting and discussing it before an actual refactor.

PS: I would expect something more explicit about the rules and the fallback etc to make it easy to read instead of some magic :/

packages/clerk-js/src/utils/authPropHelpers.ts
desiprisg: I don't think the two first params can change in the future, so I just think this will make it so we have to type `key` and `source` every time.

desiprisg:
@dimkl I've moved the logic a bit higher and used a simple search util to make the code more readable. I'd argue that the previous approach was more magical :P. This refactor was needed to provide the functionality described in the title. We can always make it better in the future. Happy to hear any other approaches but that's all I could come up with.

panteliselef (approved):

clerk-cookie:
This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 10 days.

clerk-cookie:
This PR was closed because it has been stalled for 10 days with no activity.

desiprisg:
Closing this since a lot of changes have been made to org redirects. Will make a PR for the refactos since they make adding more logic much easier.

PR #1391, opened 2023-06-20
fix(clerk-js,types): Use buildUrlWithAuth for all links/navigations

desiprisg:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Changes: 
1. `buildUrlWithAuth` now returns the same value as passed if it is a relative url;
2. All links and navigations will be filtered by `buildUrlWithAuth` in order take care of url session sync needs globally.


desiprisg:
Closing this as we can handle it in #1391

packages/clerk-js/src/core/clerk.test.ts
```diff
@@ -1458,13 +1458,13 @@ describe('Clerk singleton', () => {
   });
 
   describe('buildUrlWithAuth', () => {
-    it('builds an absolute url from a relative url in development for url based session syncing', async () => {
+    it('builds a relative url from a relative url in development for url based session syncing', async () => {
```
panteliselef, line 1461: ❓ Should we add a new test for the `absolute url` case ?

panteliselef (approved):

colinclerk:
Awesome - can we double check oauth flows?  We don't want to append the JWT there.  I can't think of other navigations we'd want to opt-out.

What do you think of explicitly using `<a>` and `window.location.href` for the opt-out cases, instead of `useBuildUrlWithAuth: false`?

In general, I think this is a smart change.  For navigations we trigger, we almost always want it, so it's a better default. It's a lot easier to remember to opt-out of buildUrlWithAuth than to opt-in.

desiprisg:
@colinclerk You mean when navigating to the oauth providers page, right? We may have to add an `options` optional param to our `navigate` and conditionally `buildUrlWithAuth` then. 

I added the prop for the `<Link/>` primitive in order to keep using that with it's styles. Maybe I got wrong what you're suggesting here.

Thanks for the feedback!

packages/clerk-js/src/ui/primitives/Link.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import { useCoreClerk } from '../contexts';
```
nikosdouvlis, line 3: Primitives should not be aware of clerk utils - let's use a HOC for that.

packages/shared/src/utils/url.ts
```diff
@@ -25,3 +27,18 @@ export function addClerkPrefix(str: string | undefined) {
   const stripped = str.replace(regex, '');
   return `clerk.${stripped}`;
 }
+
+export function isRelativeUrl(str: string) {
```
nikosdouvlis, line 31: If we use it, lets add a couple of tests if not, lets remove it please :)

packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -129,6 +130,14 @@ export abstract class BaseResource {
     method = 'POST',
     path,
   }: BaseMutateParams): Promise<this> {
+    if (typeof body === 'object') {
```
nikosdouvlis, line 133: This shouldn't be a concern of the fapi client - lets discuss alternatives

packages/clerk-js/src/core/clerk.test.ts
desiprisg: It's not supposed to build an absolute url in that case anymore.

packages/clerk-js/src/ui/primitives/Link.tsx
desiprisg: Good suggestion!

packages/clerk-js/src/core/resources/Base.ts
desiprisg: Do you have anything in mind. This is the only place to intercept I think and it might not be a concert of the fapi client but it removes complexity from a lot of places. We can always refactor if we can't find an alternative now.

nikosdouvlis: I prefer some duplication instead of making fapiClient aware of URL-based session syncing. These are two completely different concerns and should be handled accordingly. 

`actionCompleteRedirectUrl` can only be found in SignIn and SignUp resources. If not all, most `redirectUrl` props pass through SignIn/ SignUp as well - can we check that this is the case indeed and refactor accordingly? 

Another util we could explore is `buildSSOCallbackURL`.

desiprisg: I'll check if there's a way to abstract it without the current implementation, but the point was to make this something we don't need to worry about anymore. I'll get back to you.

desiprisg: I made some changes. PTAL.

desiprisg:
I've made some changes that I want to test some more so let's not close this PR just yet.

packages/clerk-js/src/ui/elements/withAuthHref.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { forwardRef } from 'react';
+
+import { useCoreClerk } from '../contexts';
+
+type WithAuthHrefProps<T> = T & { href?: string; useBuildUrlWithAuth?: boolean };
```
nikosdouvlis, line 5: Why do we need the `useBuildUrlWithAuth` option here?

nikosdouvlis (approved):

packages/clerk-js/src/ui/elements/withAuthHref.tsx
desiprisg: We need it on links such as the `signUpUrl` on the sign in page. In general, every link should have that except for the clerk marketing page url.

clerk-cookie:

PR #1393, opened 2023-06-20
feat(backend): Introduce `createIsomorphicRequest`

anagstef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [x] `@clerk/fastify`
- [ ] `@clerk/chrome-extension`
- [x] `gatsby-plugin-clerk`
- [x] `build/tooling/chore`

## Description

- [x] `npm test` runs as expected.
- [x] `npm run build` runs as expected.

Introducing the new utility `createIsomorphicRequest` for the `@clerk/backend` package, so that the `authenticateRequest` signature will be more simplified, and it will be easier to integrate with more frameworks.

PR #1456, opened 2023-07-05
feat(backend): Add methods for uploading images to User and Organization

anagstef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

This PR adds: 

- `updateOrganizationLogo` method in `OrganizationApi` class
- `updateUserProfileImage` method in `UserApi` class

### DIFF
--- a/packages/backend/src/api/endpoints/OrganizationApi.ts
+++ b/packages/backend/src/api/endpoints/OrganizationApi.ts
@@ -33,6 +33,11 @@ type UpdateParams = {
   maxAllowedMemberships?: number;
 } & MetadataParams;
 
+type UpdateLogoParams = {
+  file: Blob | File | null;
+  uploaderUserId: string;
+};
+
 type UpdateMetadataParams = MetadataParams;
 
 type GetOrganizationMembershipListParams = {
@@ -113,6 +118,34 @@ export class OrganizationAPI extends AbstractAPI {
       method: 'PATCH',
       path: joinPaths(basePath, organizationId),
       bodyParams: params,
+    });
+  }
+
+  public async updateOrganizationLogo(organizationId: string, params: UpdateLogoParams) {
+    this.requireId(organizationId);
+
+    if (params?.file === null) {
+      return this.request<Organization>({
+        method: 'DELETE',
+        path: joinPaths(basePath, organizationId, 'logo'),
+      });
+    }
+
+    if (!params?.file) {
+      throw new Error('A valid Blob or File is required.');
+    }
+    if (!params?.uploaderUserId) {
+      throw new Error('A valid uploaderUserId is required.');
+    }
+
+    const formData = new FormData();
+    formData.append('file', params.file);
+    formData.append('uploader_user_id', params.uploaderUserId);
+
+    return this.request<Organization>({
+      method: 'PUT',
+      path: joinPaths(basePath, organizationId, 'logo'),
+      formData,
     });
   }
 
--- a/packages/backend/src/api/endpoints/OrganizationApi.ts
+++ b/packages/backend/src/api/endpoints/OrganizationApi.ts
@@ -34,7 +34,7 @@ type UpdateParams = {
 } & MetadataParams;
 
 type UpdateLogoParams = {
-  file: Blob | File | null;
+  file: Blob | File;
   uploaderUserId: string;
 };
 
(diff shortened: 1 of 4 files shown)


packages/backend/src/api/endpoints/OrganizationApi.ts
```diff
@@ -116,6 +121,34 @@ export class OrganizationAPI extends AbstractAPI {
     });
   }
 
+  public async updateOrganizationLogo(organizationId: string, params: UpdateLogoParams) {
+    this.requireId(organizationId);
+
+    if (params?.file === null) {
```
gkats, line 127: :upside_down_face: Hmmm.. isn't it a cleaner API if we have two separate methods, one for each endpoint? One uploads a logo and another deletes a logo.

```ts
type UpdateLogoParams = {
  file: Blob | File;
  uploaderUserId: string;
};

updateOrganizationLogo(organizationId: string, params: UpdateLogoParams)
deleteOrganizationLogo(organizationId: string)
```

dimkl (approved):

packages/backend/src/api/endpoints/OrganizationApi.ts
```diff
@@ -116,6 +121,34 @@ export class OrganizationAPI extends AbstractAPI {
     });
   }
 
+  public async updateOrganizationLogo(organizationId: string, params: UpdateLogoParams) {
```
gkats, line 124: :upside_down_face: Do we need to mention "organization" in the method name? We should avoid stuttering if we can.

What's the usage?

```ts
organizations.updateLogo()
```

packages/backend/src/api/endpoints/OrganizationApi.ts
```diff
@@ -116,6 +121,34 @@ export class OrganizationAPI extends AbstractAPI {
...
+        path: joinPaths(basePath, organizationId, 'logo'),
+      });
+    }
+
+    if (!params?.file) {
+      throw new Error('A valid Blob or File is required.');
+    }
+    if (!params?.uploaderUserId) {
+      throw new Error('A valid uploaderUserId is required.');
+    }
```
gkats, line 139: :upside_down_face: Feel free to let the Backend API do the validations for us. Not sure if we want fast feedback at least for non empty values.

packages/backend/src/api/endpoints/UserApi.ts
```diff
@@ -126,6 +126,23 @@ export class UserAPI extends AbstractAPI {
     });
   }
 
+  public async updateUserProfileImage(userId: string, params: { file: Blob | File }) {
+    this.requireId(userId);
+
+    if (!params?.file) {
+      throw new Error('A valid Blob or File parameter is required.');
+    }
```
gkats, line 134: :upside_down_face: Same comment about validations. Feel free to resolve.

packages/backend/src/api/request.ts
```diff
@@ -16,6 +16,7 @@ export type ClerkBackendApiRequestOptions = {
   queryParams?: Record<string, unknown>;
   headerParams?: Record<string, string>;
   bodyParams?: object;
+  formData?: FormData;
```
gkats, line 19: :question: Forgive me for the newbie question, but is `FormData` available in all target runtimes we want to support?

gkats (commented):
Nice one! Left a few comments, feel free to ignore them.

packages/backend/src/api/endpoints/OrganizationApi.ts
anagstef: This is indeed a great suggestion. I just followed the Clerkjs implementation, but it makes more sense to separate them.

packages/backend/src/api/endpoints/OrganizationApi.ts
anagstef: I agree with you, although all the methods in this file include the word `Organization`. It would be consistent if this one omits it.

packages/backend/src/api/endpoints/UserApi.ts
```diff
@@ -126,6 +126,23 @@ export class UserAPI extends AbstractAPI {
     });
   }
 
+  public async updateUserProfileImage(userId: string, params: { file: Blob | File }) {
```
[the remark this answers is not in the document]
anagstef: Same reason as with Organizations.

packages/backend/src/api/request.ts
anagstef: This is actually a spot-on question. Indeed, `FormData` is available in node 18.x and above, and all the runtimes that support the Fetch API.

I'll be adding a complementary change when https://github.com/clerkinc/javascript/pull/1393 gets merged, which adds `FormData` support for node versions prior to 18.

anagstef: @nikosdouvlis  We will use it when #1393 gets merged, which re-exports `FormData` from the pollyfill.

packages/backend/src/api/endpoints/OrganizationApi.ts
gkats: You're right, I need :eyeglasses: :smile:

panteliselef:
@gkats @anagstef I believe we did some work in order to support images to be uploaded as `DataURLs`. This was done on FAPI. Is this something we could support here as well ?

anagstef:
> @gkats @anagstef I believe we did some work in order to support images to be uploaded as `DataURLs`. This was done on FAPI. Is this something we could support here as well ?

@panteliselef It would be a nice-to-have thing, and also to be consistent between FAPI and BAPI. Although, I don't think we actually need it. The Base64 support was implemented mainly to support Expo users. You can create a Blob/FIle from a Base64 string easily.

packages/backend/src/api/request.ts
anagstef: Just added support for Node prior to 18.x versions.

clerk-cookie:

PR #1482, opened 2023-07-13
USR-135 fix(clerk-js): Address spacing issues when password feedback message changes

raptisj:
fix(clerk-js): Add changeset

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

- Address spacing issues when password feedback message changes
- Add a full stop in form feedback(errors and warnings) when needed

### Before:
https://github.com/clerkinc/javascript/assets/23478420/6c211c51-5be4-487a-840b-2e5ab1d7a34c

### After: 
https://github.com/clerkinc/javascript/assets/23478420/700e0bfb-fa79-4072-9309-101e9b2f7a02

### DIFF
--- a/.changeset/tidy-worms-tan.md
+++ b/.changeset/tidy-worms-tan.md
@@ -2,5 +2,5 @@
 '@clerk/clerk-js': patch
 ---
 
-- Merge the two password field onBlur behaviours into one and address spacing issues
+- Address spacing issues when password feedback message changes
 - Add a full stop in form feedback(errors and warnings) when needed
--- a/.changeset/tidy-worms-tan.md
+++ b/.changeset/tidy-worms-tan.md
@@ -2,4 +2,5 @@
 '@clerk/clerk-js': patch
 ---
 
-Merge the two password onBlur behaviours into one and address spacing issues
+- Merge the two password field onBlur behaviours into one and address spacing issues
+- Add a full stop in form feedback(errors and warnings) when needed
--- a/.changeset/tidy-worms-tan.md
+++ b/.changeset/tidy-worms-tan.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Merge the two password onBlur behaviours into one and address spacing issues
--- a/packages/clerk-js/src/ui/elements/FormControl.tsx
+++ b/packages/clerk-js/src/ui/elements/FormControl.tsx
@@ -95,7 +95,7 @@ function useFormTextAnimation() {
 }
 
 type CalculateConfigProps = {
-  recalculate?: string | boolean | undefined;
+  recalculate?: string | undefined;
 };
 type Px = number;
 const useCalculateErrorTextHeight = (config: CalculateConfigProps = {}) => {
@@ -147,7 +147,9 @@ export const FormFeedback = (props: FormFeedbackProps) => {
   const messageToDisplay = informationMessage || successMessage || errorMessage || warningMessage;
   const isSomeMessageVisible = !!messageToDisplay;
 
-  const { calculateHeight, height } = useCalculateErrorTextHeight({ recalculate: warningMessage || errorMessage });
+  const { calculateHeight, height } = useCalculateErrorTextHeight({
+    recalculate: warningMessage || errorMessage || informationMessage,
+  });
   const { getFormTextAnimation } = useFormTextAnimation();
   const defaultElementDescriptors = {
     error: descriptors.formFieldErrorText,
--- a/packages/clerk-js/src/ui/elements/FormControl.tsx
+++ b/packages/clerk-js/src/ui/elements/FormControl.tsx
@@ -95,10 +95,10 @@ function useFormTextAnimation() {
 }
 
 type CalculateConfigProps = {
-  recalculate: string | boolean | undefined;
+  recalculate?: string | boolean | undefined;
 };
 type Px = number;
-const useCalculateErrorTextHeight = (config: CalculateConfigProps) => {
+const useCalculateErrorTextHeight = (config: CalculateConfigProps = {}) => {
   const [height, setHeight] = useState<Px>(24);
 
   const calculateHeight = useCallback(
@@ -111,12 +111,22 @@ const useCalculateErrorTextHeight = (config: CalculateConfigProps) => {
         setHeight(newHeight);
       }
     },
-    [config.recalculate],
+    [config?.recalculate],
   );
   return {
     height,
     calculateHeight,
   };
+};
+
+const addFullStop = (string: string | undefined) => {
+  if (!string) {
+    return '';
+  }
+
+  const wordSet = string?.split(' ');
+  const lastWordHasFullStop = wordSet.slice(-1)[0].includes('.') || '';
+  return lastWordHasFullStop || wordSet.length <= 1 ? string : string + '.';
 };
 
 type FormFeedbackDescriptorsKeys = 'error' | 'warning' | 'info' | 'success';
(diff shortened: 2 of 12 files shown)


packages/clerk-js/src/ui/utils/useFormControl.ts
```diff
@@ -195,19 +195,9 @@ export const useFormControlFeedback = (opts: DebouncingOption): DebouncedFeedbac
     hasPassedComplexity = false,
   } = opts;
 
-  const canDisplayFeedback = useMemo(() => {
```
panteliselef, line 198: This cannot be deleted. As it affects the UI and when errors are displayed.

Here is the result, and the bug that got introduced by this change

https://github.com/clerkinc/javascript/assets/19269911/0bfbeb6d-336b-476c-953a-a192c6d21e9e

Also this is the code that creates the confirm password field. Note the `enableErrorAfterBlur` that indicates the error should only be visible after the user has left the field at least once,

```ts
  const confirmField = useFormControl('confirmPassword', '', {
    type: 'password',
    label: localizationKeys('formFieldLabel__confirmPassword'),
    isRequired: true,
    enableErrorAfterBlur: true,
  });
```

packages/clerk-js/src/ui/elements/FormControl.tsx
```diff
@@ -175,7 +184,7 @@ export const FormFeedback = (props: FormFeedbackProps) => {
           ref={calculateHeight}
           sx={getFormTextAnimation(!!props.informationText && !props?.successfulText && !props.warningText)}
         >
-          {informationMessage}
+          {addFullStop(informationMessage)}
```
panteliselef, line 187: ❓ Why don't we update the localized messages ?

packages/clerk-js/src/ui/elements/FormControl.tsx
```diff
@@ -94,30 +94,39 @@ function useFormTextAnimation() {
...
+    (element: HTMLElement | null) => {
+      if (element) {
+        const computedStyles = getComputedStyle(element);
+        const marginTop = parseInt(computedStyles.marginTop.replace('px', ''));
+
+        const newHeight = 1.1 * marginTop + element.scrollHeight;
+        setHeight(newHeight);
+      }
+    },
+    [config?.recalculate],
```
panteliselef, line 114: My understanding here is that this is used simply to force a re-render. And its value is not actually used.

Would the following work ? 

```jsx
const calculateHeight = useCallback((element: HTMLElement | null, string) => {},[])

<FormInfoText ref={ref =>calculateHeight(ref,informationMessage)} />
<FormErrorText ref={ref => calculateHeight(ref,errorMessage)} />
<FormWarningText ref={ref => calculateHeight(ref,warningMessage)} />

```

packages/clerk-js/src/ui/elements/FormControl.tsx
```diff
@@ -136,7 +145,7 @@ export const FormFeedback = (props: FormFeedbackProps) => {
   const messageToDisplay = informationMessage || successMessage || errorMessage || warningMessage;
   const isSomeMessageVisible = !!messageToDisplay;
 
-  const { calculateHeight, height } = useCalculateErrorTextHeight();
+  const { calculateHeight, height } = useCalculateErrorTextHeight({ recalculate: warningMessage || errorMessage });
```
panteliselef, line 148: I believe `informationMessage` is missing

panteliselef (dismissed):

packages/clerk-js/src/ui/elements/FormControl.tsx
raptisj: @panteliselef we can't do this, or at least it would be very error-prone if we do. 
For example, the warning/error will be a combination of these keys which. happens dynamically.
```   passwordComplexity: {
      sentencePrefix: 'Your password must contain',
      minimumLength: '{{length}} or more characters',
      maximumLength: 'less than {{length}} characters',
      requireNumbers: 'a number',
      requireLowercase: 'a lowercase letter',
      requireUppercase: 'an uppercase letter',
      requireSpecialCharacter: 'a special character',
    }, 

I think it is better to do this in the end. Here I wrapped all messages with the helper purely for convention and possible future refactors.

File: packages/clerk-js/src/ui/utils/useFormControl.ts (hunk 1 above)
Comment (raptisj) [reply packages/clerk-js/src/ui/utils/useFormControl.ts]: @panteliselef nice catch. I reverted this change and took another approach.

File: packages/clerk-js/src/ui/elements/FormControl.tsx (hunk 2 above)
Comment (panteliselef) [reply packages/clerk-js/src/ui/elements/FormControl.tsx:189]: You are correct, I missed this one.
In that case, because we always want to append the full-stop in the end, consider updating the `generateErrorTextUtil` function that generates this string.
This will enable you to remove the custom function for adding a full-stop.

File: packages/clerk-js/src/ui/elements/FormControl.tsx (hunk 3 above)
Comment (panteliselef) [reply packages/clerk-js/src/ui/elements/FormControl.tsx:114]: this makes sense

File: packages/clerk-js/src/ui/elements/FormControl.tsx [hunk 5]
```diff
@@ -94,30 +94,41 @@ function useFormTextAnimation() {
   };
 }
 
+type CalculateConfigProps = {
+  recalculate?: string | boolean | undefined;
```
panteliselef, line 98: How come we need boolean here ?

packages/clerk-js/src/ui/elements/FormControl.tsx
panteliselef: @raptisj ☝️

packages/clerk-js/src/ui/elements/FormControl.tsx
raptisj: @panteliselef we don't. I had the thought initially to make it a boolean value but later revised

packages/clerk-js/src/ui/elements/FormControl.tsx
raptisj: @panteliselef If we move this to `generateErrorTextUtil` we will lose the ability to apply this throughout the app, something I think we want in general. WDYT?

panteliselef: I don't think it's safe to assume that we can add full stops like this. This should be something included in a localized value. 
For the case of Password Complexity errors we can add a full-stop "by hand" the moment that we construct the string. Same for password strength errors. These two are the only strings generated by concatinating other localized values.

.changeset/tidy-worms-tan.md
```diff
@@ -0,0 +1,6 @@
+---
+'@clerk/clerk-js': patch
+---
+
+- Address spacing issues
```
desiprisg, line 5: Should we clarify here?

packages/clerk-js/src/ui/hooks/usePasswordComplexity.ts
```diff
@@ -51,6 +51,16 @@ const errorMessages = {
   require_special_char: 'unstable__errors.passwordComplexity.requireSpecialCharacter',
 };
 
+const addFullStop = (string: string | undefined) => {
+  if (!string) {
+    return '';
+  }
+
+  const wordSet = string?.split(' ');
+  const lastWordHasFullStop = wordSet.slice(-1)[0].includes('.') || '';
```
desiprisg, line 60: Can we just do something like this here just to simplify?
```
if (text.charAt(a.length-1) === "."){
  //add the dot
}
```

desiprisg:
Did we lose the animations here. I can see they're missing in the before as well? @panteliselef Am I missing something?

panteliselef (approved):
Looks good to me 💯

clerk-cookie:

PR #1493, opened 2023-07-17
USR-102 fix(clerk-react): Add authStateKey field to useAuth

raptisj:
fix(clerk-react): Add changeset

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Add authStateKey field to `useAuth` in order to have the auth state in a hashed format.

https://github.com/clerkinc/javascript/assets/23478420/7b24e49e-7d8a-40cb-bbe6-98b358bfbb84

### DIFF
--- a/.changeset/funny-toys-hide.md
+++ b/.changeset/funny-toys-hide.md
@@ -1,0 +1,32 @@
+---
+'@clerk/clerk-react': minor
+---
+
+Add an `authStateKey` field to the `useAuth` hook so it returns a value that is unique per auth state.
+
+The `authStateKey` is a string value that was derived by hashing the `sessionId`, `userId` and the `orgId`. If any of these change in value, it will result in a completely different hash value. This can be of value if there is a need for caching.
+
+
+Example with SWR
+
+```jsx
+
+const fetcher = {
+    fetch('/user/123')
+}
+
+const singleUser = '/user'
+
+useSWR(singleUser, fetcher); // this won't fetch '/user/456'
+```
+
+With `authStateKey` this can become
+```jsx
+const { authStateKey }  = useAuth()
+
+const singleUser = '/user'
+
+useSWR(`${singleUser}?hash=${authStateKey}`, fetcher);
+
+```
+This will invalidate the cache if `sessionId`, `userId` or `orgId` changed.
--- a/packages/clerk-js/src/utils/memoizeStateListenerCallback.ts
+++ b/packages/clerk-js/src/utils/memoizeStateListenerCallback.ts
@@ -46,7 +46,7 @@ function userMembershipsChanged(prev: UserResource, next: UserResource): boolean
 }
 
 // TODO: Decide if this belongs in the resources
-function resourceChanged<T extends AcceptedResource | undefined | null>(prev: T, next: T): boolean {
+function resourceChanged<T extends AcceptedResource | undefined | null | string>(prev: T, next: T): boolean {
   // support the setSession undefined intermediate state
   if ((!prev && !!next) || (!!prev && !next)) {
     return true;
@@ -90,7 +90,8 @@ function getSameOrUpdatedResource<
     | OrganizationInvitationResource
     | OrganizationMembershipResource
     | undefined
-    | null,
+    | null
+    | string,
 >(prev: T, next: T): T {
   return resourceChanged(prev, next) ? next : prev;
 }
@@ -106,6 +107,7 @@ function stateWithMemoizedResources(cur: Resources, next: Resources): Resources 
       next.lastOrganizationInvitation,
     ),
     lastOrganizationMember: getSameOrUpdatedResource(cur.lastOrganizationMember, next.lastOrganizationMember),
+    authStateKey: getSameOrUpdatedResource(cur.authStateKey, next.authStateKey),
   };
 }
 
--- a/packages/clerk-js/src/utils/sha1.ts
+++ b/packages/clerk-js/src/utils/sha1.ts
@@ -1,0 +1,145 @@
+/////////
+///////
+// The SHA (Secure Hash Algorithm) can be used to encrypt data for secure transfer between applications.
+// The SHA1() function returns a string with the SHA1 encrypted hash as a string with 40 characters.
+// It is fully compatible with UTF-8 encoding.
+///
+//
+export const SHA1 = (msg: string) => {
+  function rotate_left(n: number, s: number) {
+    const t4 = (n << s) | (n >>> (32 - s));
+    return t4;
+  }
+
+  function cvt_hex(val: number) {
+    let str = '';
+    let i;
+    let v;
+    for (i = 7; i >= 0; i--) {
+      v = (val >>> (i * 4)) & 0x0f;
+      str += v.toString(16);
+    }
+    return str;
+  }
+  function Utf8Encode(string: string) {
+    string = string.replace(/\r\n/g, '\n');
+    let utftext = '';
+    for (let n = 0; n < string.length; n++) {
+      const c = string.charCodeAt(n);
+      if (c < 128) {
+        utftext += String.fromCharCode(c);
+      } else if (c > 127 && c < 2048) {
+        utftext += String.fromCharCode((c >> 6) | 192);
+        utftext += String.fromCharCode((c & 63) | 128);
+      } else {
+        utftext += String.fromCharCode((c >> 12) | 224);
+        utftext += String.fromCharCode(((c >> 6) & 63) | 128);
+        utftext += String.fromCharCode((c & 63) | 128);
+      }
+    }
+    return utftext;
+  }
+  let blockstart;
+  let i, j;
+  const W = new Array(80);
+  let H0 = 0x67452301;
+  let H1 = 0xefcdab89;
+  let H2 = 0x98badcfe;
+  let H3 = 0x10325476;
+  let H4 = 0xc3d2e1f0;
+  let A, B, C, D, E;
+  let temp;
+  msg = Utf8Encode(msg);
+  const msg_len = msg.length;
+  const word_array: number[] = [];
+  for (i = 0; i < msg_len - 3; i += 4) {
+    j =
+      (msg.charCodeAt(i) << 24) | (msg.charCodeAt(i + 1) << 16) | (msg.charCodeAt(i + 2) << 8) | msg.charCodeAt(i + 3);
+    word_array.push(j);
+  }
+  switch (msg_len % 4) {
+    case 0:
+      i = 0x080000000;
+      break;
+    case 1:
+      i = (msg.charCodeAt(msg_len - 1) << 24) | 0x0800000;
+      break;
+    case 2:
+      i = (msg.charCodeAt(msg_len - 2) << 24) | (msg.charCodeAt(msg_len - 1) << 16) | 0x08000;
+      break;
+    case 3:
+      i =
+        (msg.charCodeAt(msg_len - 3) << 24) |
+        (msg.charCodeAt(msg_len - 2) << 16) |
+        (msg.charCodeAt(msg_len - 1) << 8) |
+        0x80;
+      break;
+  }
+  word_array.push(i);
+
+  while (word_array.length % 16 != 14) {
+    word_array.push(0);
+  }
+
+  word_array.push(msg_len >>> 29);
+  word_array.push((msg_len << 3) & 0x0ffffffff);
+  for (blockstart = 0; blockstart < word_array.length; blockstart += 16) {
+    for (i = 0; i < 16; i++) {
+      W[i] = word_array[blockstart + i];
+    }
+
+    for (i = 16; i <= 79; i++) {
+      W[i] = rotate_left(W[i - 3] ^ W[i - 8] ^ W[i - 14] ^ W[i - 16], 1);
+    }
+
+    A = H0;
+    B = H1;
+    C = H2;
+    D = H3;
+    E = H4;
+    for (i = 0; i <= 19; i++) {
+      // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+      temp = (rotate_left(A, 5) + ((B & C) | (~B & D)) + E + W[i] + 0x5a827999) & 0x0ffffffff;
+      E = D;
+      D = C;
+      C = rotate_left(B, 30);
+      B = A;
+      A = temp;
+    }
+    for (i = 20; i <= 39; i++) {
+      // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+      temp = (rotate_left(A, 5) + (B ^ C ^ D) + E + W[i] + 0x6ed9eba1) & 0x0ffffffff;
+      E = D;
+      D = C;
+      C = rotate_left(B, 30);
+      B = A;
+      A = temp;
+    }
+    for (i = 40; i <= 59; i++) {
+      // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+      temp = (rotate_left(A, 5) + ((B & C) | (B & D) | (C & D)) + E + W[i] + 0x8f1bbcdc) & 0x0ffffffff;
+      E = D;
+      D = C;
+      C = rotate_left(B, 30);
+      B = A;
+      A = temp;
+    }
+    for (i = 60; i <= 79; i++) {
+      // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+      temp = (rotate_left(A, 5) + (B ^ C ^ D) + E + W[i] + 0xca62c1d6) & 0x0ffffffff;
+      E = D;
+      D = C;
+      C = rotate_left(B, 30);
+      B = A;
+      A = temp;
+    }
+    H0 = (H0 + A) & 0x0ffffffff;
+    H1 = (H1 + B) & 0x0ffffffff;
+    H2 = (H2 + C) & 0x0ffffffff;
+    H3 = (H3 + D) & 0x0ffffffff;
+    H4 = (H4 + E) & 0x0ffffffff;
+  }
+  const result = cvt_hex(H0) + cvt_hex(H1) + cvt_hex(H2) + cvt_hex(H3) + cvt_hex(H4);
+
+  return result.toLowerCase();
+};
--- a/packages/clerk-js/src/utils/syncDigest.ts
+++ b/packages/clerk-js/src/utils/syncDigest.ts
@@ -1,0 +1,14 @@
+import { SHA1 } from './sha1';
+
+export const syncDigest = (values: Record<string, string | undefined>) => {
+  const sortedValues = Object.keys(values)
+    .sort()
+    .filter(s => values[s])
+    .map(s => values[s]);
+
+  if (!sortedValues.length) {
+    return '';
+  }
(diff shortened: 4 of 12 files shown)


.changeset/funny-toys-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/clerk-react': minor
+---
+
+Add authStateKey field to useAuth in order to have the auth state in a hashed format
```
desiprisg, line 5:
```suggestion
Add an `authStateKey` field to the `useAuth` hook's return value that is unique per auth state.
```

packages/react/src/hooks/useAuth.ts
```diff
@@ -117,6 +124,7 @@ export const useAuth: UseAuth = () => {
       orgSlug: undefined,
       signOut,
       getToken,
+      authStateKey: undefined,
```
desiprisg, line 127: Shouldn't we use a hash here as well? It is still an "authState".

packages/react/src/hooks/useAuth.ts
```diff
@@ -132,6 +140,7 @@ export const useAuth: UseAuth = () => {
       orgSlug: null,
       signOut,
       getToken,
+      authStateKey: null,
```
desiprisg, line 143: Shouldn't we use a hash here as well? It is still an "authState".

packages/react/src/hooks/useAuth.ts
nikosdouvlis: 🔧 
It'd expect `authStateKey` to  be a `string` always so it needs to take a value here so it can be used to bust caches when going from loaded -> signed in or loaded -> signed out states

packages/react/src/hooks/useAuth.ts
nikosdouvlis: 🔧  Same.

packages/react/src/utils/sha1.ts
```diff
@@ -0,0 +1,170 @@
...
+  const sortedValues = Object.keys(values)
+    .sort()
+    .reduce((accumulator, key) => {
+      if (values[key as keyof HashValueProps]) {
+        // @ts-ignore
+        accumulator.push(values[key]);
+      }
+
+      return accumulator;
+    }, []);
```
nikosdouvlis, line 17: I think we can replace this with a `filter` function?
Eg
```
Object.entries(values).sort((a,b) => a[1] - b[1]).filter(e => e[1]) 
```

packages/react/src/hooks/useAuth.ts
raptisj: @desiprisg Maybe we could just hash an empty string and have that as its value

packages/react/src/hooks/useAuth.ts
raptisj: @desiprisg https://github.com/clerkinc/javascript/pull/1493#discussion_r1266511299

packages/react/src/utils/sha1.ts
```diff
@@ -0,0 +1,170 @@
...
+};
+
+/////////
+///////
+// The SHA (Secure Hash Algorithm) can be used to encrypt data for secure transfer between applications.
+// The SHA1() function returns a string with the SHA1 encrypted hash as a string with 40 characters.
+// It is fully compatible with UTF-8 encoding.
+///
+//
+export const SHA1 = (msg: string) => {
```
nikosdouvlis, line 33: Good thinking extracting sha1 like this. I'm debating whether we need to go with SHA1 after all. 
Take a look at this: https://clerk.com/blog/generating-sortable-stripe-like-ids-with-segment-ksuids#clerk-s-id-generator  - Clerk is using [KSUID](https://github.com/segmentio/ksuid) to generate IDs

Given that our ID generation logic will likely always stay the same, would it be sufficient to use the existing ids to create a hash? I think we can extract part of each id:
 
Screenshot 2023-07-18 at 12 36 07

and concatenate the parts together. If an id is undefined, we consider it to be  "" instead.

We're going to end up with signatures of around ~15chars long if orgs are enabled which is longer than the 8chars of the original design, but I think the tradeoff is worth it.

For the loading state where you have no ids available, we can probably default to `Math.random().toString(36).substr(2, 8);`

What do you think? @gkats @desiprisg

nikosdouvlis (changes requested):
Finally, I'd like to ask if we will expose this as a clerk-js API as well. This is implemented in `clerk/clerk-react` so `clerk-js` will not benefit from it.

Does it make sense to create the a public `authStateKey` getter on the Clerk singleton and make useAuth return that?

packages/react/src/utils/sha1.ts
raptisj: @nikosdouvlis This makes sense to me to be honest

packages/react/src/hooks/useAuth.ts
desiprisg: Why can't we do exactly as we do in the other cases?

packages/react/src/utils/sha1.ts
nikosdouvlis: It depends - I think implementing what I wrote above would be much simpler than embedding the SHA1 algorithm in our packages. 

A different approach would be to use https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest - did you explore that option? A small downside is that `digest` returns a Promise and we'd probably like `authStateKey` to be sync.

raptisj: @gkats Valid point. What if we used the [MD5](https://coursesweb.net/javascript/md5-hash-string_cs) hashing algorithm that is faster, instead of SHA1? 

Regarding cons(point 2) if we add something like organization role, for example, we can append that as a string since we don't care so much about security(e.4jdfjd8klj453kjh4admin)

nikosdouvlis: I think its safe to say that SubtleCrypto will be available in every platform where Clerk runs as far as web or cross-platform native apps are concerned. The digest method can produce a hash using the SHA1 algorithm so that wouldn't change across platforms anyway.
The biggest drawback of using the web crypto API directly is the fact that it is async and not sync as the original proposal. 

Regarding the performance of MD5 and SHA1, I think it doesn't matter in this case as the key will rarely change, both algorithms are fast enough.

What @gkats said about the size is true though, its going to go from 6-8chars to ~15 for our use case. We could pick less chars per id though as entropy should be enough regardless, I wouldn't worry too much about collisions or duplicate keys (false-negatives) on the frontend.

nikosdouvlis: Bundling sha1 definitely feels weird to me but it makes sense if you consider the above discussion so the decision is up to you!

If we decide to proceed with sha1 we should probably do the following refactors in this PR:
- separate the sha1 algo from the rest of the utils. Right now, the `sha1.ts` file has utils specific to `authStateKey`
- stop exporting sha1, export a `syncDigest` method instead

raptisj: @nikosdouvlis When went with sha1 and moved the functionality in Clerk-js.

packages/clerk-js/src/utils/syncDigest.ts
```diff
@@ -0,0 +1,26 @@
+import { SHA1 } from './sha1';
+
+type HashValueProps = {
+  sessionId: string | undefined | null;
+  userId: string | undefined | null;
+  orgId: string | undefined | null;
+};
+
+export const syncDigest = (values: HashValueProps) => {
```
nikosdouvlis, line 9: Can we make this more generic by having it accept a `string[]` or a `Record<string,string>`?
There's no reason to type the keys here I think

packages/clerk-js/src/utils/syncDigest.ts
```diff
@@ -0,0 +1,26 @@
...
+  const sortedValues = Object.keys(values)
+    .sort()
+    .reduce((accumulator, key) => {
+      if (values[key as keyof HashValueProps]) {
+        // @ts-ignore
+        accumulator.push(values[key]);
+      }
+
+      return accumulator;
+    }, []);
```
nikosdouvlis, line 19: Can we simplify this using a filter function as discussed?

packages/types/src/clerk.ts
```diff
@@ -89,6 +89,9 @@ export interface Clerk {
   /** Current User. */
   user?: UserResource | null;
 
+  /** Current state hash derived from userId, sessionId and orgId */
+  authStateKey?: string;
```
nikosdouvlis, line 93: As we've discussed, we'll be moving our FE SDK docs inline using JSDoc. In preparation for that, let's treat this JSDoc comment as the documentation for this prop and provide a better/fuller description and an example (eg This is especially useful if you want to invalidate caches (eg: for react-quiery, useSwr etc) . . .`

Also, let's stick with the 
```
/**
*
*/
```
even for single-line comments from now on

packages/clerk-js/src/utils/memoizeStateListenerCallback.ts
```diff
@@ -106,6 +107,7 @@ function stateWithMemoizedResources(cur: Resources, next: Resources): Resources
       next.lastOrganizationInvitation,
     ),
     lastOrganizationMember: getSameOrUpdatedResource(cur.lastOrganizationMember, next.lastOrganizationMember),
+    authStateKey: getSameOrUpdatedResource(cur.authStateKey, next.authStateKey),
```
nikosdouvlis, line 110: Can you explain why this is needed?

.changeset/funny-toys-hide.md
```diff
@@ -0,0 +1,32 @@
+---
+'@clerk/clerk-react': minor
+---
```
nikosdouvlis, line 3: very nice!

packages/clerk-js/src/utils/memoizeStateListenerCallback.ts
raptisj: @nikosdouvlis From what I understand this is needed to notify the listener that the resource has been changed so it can get the new value. Without this, the authStateKey just has the initial value(`' '`) and never changes after that.

.changeset/funny-toys-hide.md
```diff
@@ -0,0 +1,32 @@
...
+const fetcher = {
+    fetch('/user/123')
+}
+
+const singleUser = '/user'
+
+useSWR(singleUser, fetcher); // this won't fetch '/user/456'
+```
+
+With `authStateKey` this can be become
```
desiprisg, line 23:
```suggestion
With `authStateKey` this will become
```

packages/clerk-js/src/utils/syncDigest.ts
raptisj: @nikosdouvlis doing with filter returns an array of arrays that we have to map as well. Is this better than reduce?
e.g. 
```  
something.sort()
    .filter(s => values[s])
    .map(s => values[s]);
```

raptisj:
@nikosdouvlis all comments were addressed

packages/clerk-js/src/utils/sha1.ts
```diff
@@ -0,0 +1,145 @@
+/////////
+///////
+// The SHA (Secure Hash Algorithm) can be used to encrypt data for secure transfer between applications.
+// The SHA1() function returns a string with the SHA1 encrypted hash as a string with 40 characters.
+// It is fully compatible with UTF-8 encoding.
+///
+//
+export const SHA1 = (msg: string) => {
```
SokratisVidros, line 8: Team, do you know if this polyfill is necessary? We could use webcrypto. It was great browser support.

```async function sha1(str) {
  const enc = new TextEncoder();
  const hash = await crypto.subtle.digest('SHA-1', enc.encode(str));
  return Array.from(new Uint8Array(hash))
    .map(v => v.toString(16).padStart(2, '0'))
    .join('');
}```

File: packages/react/src/hooks/useAuth.ts [hunk 13]
```diff
@@ -31,6 +32,7 @@ type UseAuthReturn =
       orgSlug: null;
       signOut: SignOut;
       getToken: GetToken;
+      authStateKey: string;
```
SokratisVidros, line 35: @raptisj Have we settled on the name in the design doc?

SokratisVidros (changes requested):

packages/react/src/hooks/useAuth.ts
raptisj: @SokratisVidros We settled on this `authStateKey`. I saw your comments on the doc and it kinda makes sense too. Do you find `authStateKey` to be problematic?

packages/clerk-js/src/utils/sha1.ts
raptisj: @SokratisVidros unfortunately the implementation you provided is async. Going this route means we will have to make `authStateKey` a function(similar to getToken) and let the user resolve it. I don't think we want this. This was the main reason we didn't choose this kind of handling and went with a more straight forward sync way.

cc @nikosdouvlis

nikosdouvlis: This was discussed in this PR again, the initial design described a sync `authStateKey` prop. However, given that this has come up twice so far, let's explore what making this async would look like.

```
import useSWR from 'swr';
import { useAuth } from '@clerk/nextjs';

export default function useClerkSWR(url) {
  const { getToken, authStateKey } = useAuth();

  const fetcher = async (...args) => {
    return fetch(...args, {
     // this is valid as this Promise is correctly handled within SWR 
      headers: { Authorization: `Bearer ${await getToken()}` }
    }).then(res => res.json());
  };

  // await cannot be used here currently
  return useSWR([url, authStateKey], fetcher);
}
```
(References: https://swr.vercel.app/docs/arguments)

Meaning that if we were to use the built-in crypto methods, authStateKey would need to:
1. either be async (bad DX as it requires manually unwrapping the value from the promise inside hooks)
2. or, be a sync variable holding `undefined` or `""` while the Promise is not resolved, and a `string` once it resolves. If this is implemented with `useState`, `useAuth` will render multiple times.

A different approach would be to use `useRef` within `useAuth` (so no extra renders because of `useState) to hold a reference to the latest resolved key and expose a sync method to either retrieve the current value of the promise if the promise is resolved, or fallback to a different value (eg `""`) if the promise is not resolved

packages/clerk-js/src/utils/syncDigest.ts
nikosdouvlis: THis is very minor so don't worry about it, personally I'd go with `Object.entries` as you don't need to reference `values` outside of the initial entries call :)

For the nested array, you can take a look at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

packages/clerk-js/src/utils/sha1.ts
desiprisg: I agree with @SokratisVidros. Let's use webcrypto. It should work since localhost is considered a secure context.

nikosdouvlis: @desiprisg Please take a look at my answer above as this possible but not as straightforward as we'd like

raptisj:
!snapshot

clerk-cookie:
Hey @raptisj - the snapshot version command generated the following package versions:
| Package | Version |
| --- | --- |
| @clerk/backend | 0.26.1-snapshot.e77e2ff |
| @clerk/chrome-extension | 0.3.26-snapshot.e77e2ff |
| @clerk/clerk-js | 4.54.3-snapshot.e77e2ff |
| eslint-config-custom | 0.3.0 |
| @clerk/clerk-expo | 0.18.17-snapshot.e77e2ff |
| @clerk/fastify | 0.6.3-snapshot.e77e2ff |
| gatsby-plugin-clerk | 4.4.4-snapshot.e77e2ff |
| @clerk/localizations | 1.24.1-snapshot.e77e2ff |
| @clerk/nextjs | 4.23.2-snapshot.e77e2ff |
| @clerk/clerk-react | 4.24.0-snapshot.e77e2ff |
| @clerk/remix | 2.9.1-snapshot.e77e2ff |
| @clerk/clerk-sdk-node | 4.12.2-snapshot.e77e2ff |
| @clerk/shared | 0.20.0 |
| @clerk/themes | 1.7.5 |
| @clerk/types | 3.48.2-snapshot.e77e2ff |

Tip: use the snippet copy button below to quickly install the required packages.
```sh
# @clerk/backend
npm i @clerk/backend@0.26.1-snapshot.e77e2ff
```
```sh
# @clerk/chrome-extension
npm i @clerk/chrome-extension@0.3.26-snapshot.e77e2ff
```
```sh
# @clerk/clerk-js
npm i @clerk/clerk-js@4.54.3-snapshot.e77e2ff
```
```sh
# eslint-config-custom
npm i eslint-config-custom@0.3.0
```
```sh
# @clerk/clerk-expo
npm i @clerk/clerk-expo@0.18.17-snapshot.e77e2ff
```
```sh
# @clerk/fastify
npm i @clerk/fastify@0.6.3-snapshot.e77e2ff
```
```sh
# gatsby-plugin-clerk
npm i gatsby-plugin-clerk@4.4.4-snapshot.e77e2ff
```
```sh
# @clerk/localizations
npm i @clerk/localizations@1.24.1-snapshot.e77e2ff
```
```sh
# @clerk/nextjs
npm i @clerk/nextjs@4.23.2-snapshot.e77e2ff
```
```sh
# @clerk/clerk-react
npm i @clerk/clerk-react@4.24.0-snapshot.e77e2ff
```
```sh
# @clerk/remix
npm i @clerk/remix@2.9.1-snapshot.e77e2ff
```
```sh
# @clerk/clerk-sdk-node
npm i @clerk/clerk-sdk-node@4.12.2-snapshot.e77e2ff
```
```sh
# @clerk/shared
npm i @clerk/shared@0.20.0
```
```sh
# @clerk/themes
npm i @clerk/themes@1.7.5
```
```sh
# @clerk/types
npm i @clerk/types@3.48.2-snapshot.e77e2ff
```

packages/react/src/utils/sha1.ts
SokratisVidros: Folks have we considered returning the key from FAPI on the session object?

packages/clerk-js/src/utils/sha1.ts
SokratisVidros: Great call on it being async.

nikosdouvlis:
Folks I'm going to close this one for now :( 
There are multiple projects running at the same time so I don't think we have enough capacity for this one right now, so let's please re-open if/when we revisit this one

clerk-cookie:

PR #1505, opened 2023-07-21
USR-193 fix(clerk-js): Introduce ability to read multiple errors

raptisj:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Currently, client-side password validation constructs dynamically all error messages based on the password complexity settings. In this PR we want to support reading multiple errors in order to construct a password error with information 
that comes from FAPI. This way we can achieve parity between password client-side validations and errors 
returned from FAPI, resulting in the same exact message.

Currently, we support handling only one error object that comes from FAPI so if we have many errors regarding passwords(e.g. too short password, uppercase required, lowercase required) it will only show the first. We want the error handler to read all incoming errors(only password relating errors will be multiple) so a more concise password can be shown.

### Before
https://github.com/clerkinc/javascript/assets/23478420/338cbe9e-ff5f-4149-9a5d-b7bc1494368d

### After
https://github.com/clerkinc/javascript/assets/23478420/12b8b50c-68cc-4878-8541-4e50dab80a00


packages/clerk-js/src/ui/utils/errorHandler.ts
```diff
@@ -1,24 +1,44 @@
...
+) {
   if (!errors || errors.length < 1) {
     return;
   }
 
   fieldStates.forEach(field => {
-    const error = errors.find(err => {
-      return err.meta!.paramName === field.id || snakeToCamel(err.meta!.paramName) === field.id;
+    const passwordError = errors.filter(err => {
+      return err.meta!.paramName === PASSWORD || snakeToCamel(err.meta!.paramName) === PASSWORD;
```
panteliselef, line 26: I really don't like hard coding this inside of `setFieldErrors`. This will not work for other paramNames that would benefit from the same functionality. (See `newPassword` in PasswordPage

packages/clerk-js/src/ui/utils/errorHandler.ts
```diff
@@ -79,10 +104,16 @@ export function getFieldError(err: Error): ClerkAPIError | undefined {
   if (!isClerkAPIResponseError(err)) {
     return;
   }
+
   const { fieldErrors } = parseErrors(err.errors);
+  if (fieldErrors.length) {
```
panteliselef, line 109: This line makes `return fieldErrors[0]` to never be called

packages/clerk-js/src/ui/utils/passwordUtils.ts
```diff
@@ -0,0 +1,52 @@
+import type { PasswordSettingsData } from '@clerk/types';
+
+import { localizationKeys } from '../customizables';
+import { addFullStop, canUseListFormat } from '../utils';
+
+// match FAPI error codes with localization keys
+export const getCodes = (passwordSettings: Pick<PasswordSettingsData, 'max_length' | 'min_length'>) => {
```
panteliselef, line 7:
```suggestion
export const mapComplexityErrors = (passwordSettings: Pick<PasswordSettingsData, 'max_length' | 'min_length'>) => {
```

packages/clerk-js/src/ui/utils/passwordUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    if (Array.isArray(localizedKey)) {
+      const [lk, attr, val] = localizedKey;
+      return t(localizationKeys(lk, { [attr]: val }));
+    }
+    return t(localizationKeys(localizedKey));
+  });
+
+  let messageWithPrefix: string;
+  if (canUseListFormat(locale)) {
```
panteliselef, line 40: This code reminds me of `generateErrorTextUtil`, can we do something to avoid duplication ?

panteliselef (changes requested):
The UX feels like what we want to accomplish, let's focus on getting the code in the right state

packages/clerk-js/src/ui/utils/errorHandler.ts
raptisj: @panteliselef I agree. I added only `password` to do the proof of concept

packages/clerk-js/src/ui/utils/passwordUtils.ts
raptisj: @panteliselef we can do what I did for `addFullStop` helper. Move them to a `localicationUtils` file(the name is questionable) and import them where is needed

packages/clerk-js/src/ui/utils/errorHandler.ts
```diff
@@ -1,24 +1,49 @@
...
+  localizationConfig?: any,
+) {
   if (!errors || errors.length < 1) {
     return;
   }
 
   fieldStates.forEach(field => {
-    const error = errors.find(err => {
-      return err.meta!.paramName === field.id || snakeToCamel(err.meta!.paramName) === field.id;
+    const passwordError = errors.filter(err => {
```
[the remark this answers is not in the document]
raptisj: @nikosdouvlis We want to take all the password-related errors and "merge" them in a way. We don't do this manipulation in all other errors. All other errors will likely return a single error object which we will, later on, choose to localize or show the error.message coming from FAPI as is.

packages/clerk-js/src/ui/components/SignIn/ResetPassword.tsx
```diff
@@ -18,13 +18,16 @@ export const _ResetPassword = () => {
...
+  const { t, locale } = useLocalizations();
+
   const passwordField = useFormControl('password', '', {
     type: 'password',
     label: localizationKeys('formFieldLabel__newPassword'),
     isRequired: true,
     enableErrorAfterBlur: true,
     validatePassword: true,
     informationText: failedValidationsText,
+    callback: errors => createPasswordError(errors, { t, locale, passwordSettings }),
```
anagstef, line 30: ❓ This function accepts an array of one or more errors and returns a string with the final error message. Should we change the naming from `callback` to something less generic?

packages/clerk-js/src/ui/utils/errorHandler.ts
```diff
@@ -15,10 +15,17 @@ function setFieldErrors(fieldStates: Array<FormControlState<string>>, errors: Cl
...
 
   fieldStates.forEach(field => {
-    const error = errors.find(err => {
+    let errorCallback = field?.callback;
+
+    if (!errorCallback) {
+      errorCallback = (err: ClerkAPIError[]) => err[0];
+    }
+
+    const error = errors.filter(err => {
```
anagstef, line 24: 🔧 The `error` variable here is an array of one or more errors. We should rename it.

packages/clerk-js/src/ui/utils/errorHandler.ts
```diff
@@ -15,10 +15,17 @@ function setFieldErrors(fieldStates: Array<FormControlState<string>>, errors: Cl
   }
 
   fieldStates.forEach(field => {
-    const error = errors.find(err => {
+    let errorCallback = field?.callback;
```
anagstef, line 18: 🔧 We should also rename this variable. `errorCallback` hints a function that runs when erroring. This is a function that manipulates the errors and returns the final error string.

packages/clerk-js/src/ui/utils/errorHandler.ts
```diff
@@ -15,10 +15,17 @@ function setFieldErrors(fieldStates: Array<FormControlState<string>>, errors: Cl
...
+
+    if (!errorCallback) {
+      errorCallback = (err: ClerkAPIError[]) => err[0];
```
anagstef, line 21: ❓ Should we return a string in this function here? I think it's a bit confusing to return either a `ClerkAPIError` object or a string. Is this handled appropriately?

raptisj: @anagstef No need to do this here. This is handled later on from `translateError` in useLocalization([here](https://github.com/clerkinc/javascript/blob/d0816ee1385ae942b9853d617ce5d2e4ad6f5cd1/packages/clerk-js/src/ui/localization/makeLocalizable.tsx#L73))

SokratisVidros:
!snapshot

packages/clerk-js/src/ui/hooks/usePassword.ts
```diff
@@ -153,12 +153,14 @@ export const useConfirmPassword = ({
...
-    (confirmPassword: string) =>
-      passwordField.value.trim().length >= MIN_PASSWORD_LENGTH && passwordField.value === confirmPassword,
+    (confirmPassword: string) => passwordField.value === confirmPassword,
     [passwordField.value],
   );
 
-  const isPasswordMatch = useMemo(() => checkPasswordMatch(confirmPasswordField.value), [confirmPasswordField.value]);
+  const isPasswordMatch = useMemo(
+    () => checkPasswordMatch(confirmPasswordField.value),
+    [passwordField, confirmPasswordField.value],
```
panteliselef, line 162: `passwordField` is not directly affecting `isPasswordMatch` 
```suggestion
    [checkPasswordMatch, confirmPasswordField.value],
```

packages/clerk-js/src/ui/utils/errorHandler.ts
```diff
@@ -79,10 +86,13 @@ export function getFieldError(err: Error): ClerkAPIError | undefined {
   if (!isClerkAPIResponseError(err)) {
     return;
   }
+
```
panteliselef, line 89: ❓ Is this prettier ?

packages/clerk-js/src/ui/utils/passwordUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+  passwordSettings: Pick<PasswordSettingsData, 'max_length' | 'min_length'>;
+};
+export const createPasswordError = (errors: ClerkAPIError[], localizationConfig: LocalizationConfigProps) => {
+  if (!localizationConfig) {
+    return errors[0].longMessage;
+  }
+
+  const { t, locale, passwordSettings } = localizationConfig;
+
+  const message = errors.map((s: any) => {
```
panteliselef, line 39: ❓ There is a similar mapping logic inside `generateErrorTextUtil`. Is this something we can merge here as well or it is not possible ?

packages/clerk-js/src/ui/utils/useFormControl.ts
```diff
@@ -19,11 +19,13 @@ type Options = {
 } & (
   | {
       validatePassword?: never;
+      buildErrorMessage?: never;
       type?: Exclude<HTMLInputTypeAttribute, 'password'>;
     }
   | {
       type: Extract<HTMLInputTypeAttribute, 'password'>;
       validatePassword: boolean;
+      buildErrorMessage?: (err: ClerkAPIError[]) => ClerkAPIError | string | undefined;
```
panteliselef, line 28: ❤️ Loving this, as it depends on the type of the input and not the name

packages/clerk-js/src/ui/utils/errorHandler.ts
```diff
@@ -15,10 +15,17 @@ function setFieldErrors(fieldStates: Array<FormControlState<string>>, errors: Cl
   }
 
   fieldStates.forEach(field => {
-    const error = errors.find(err => {
+    let buildErrorMessage = field?.buildErrorMessage;
+
+    if (!buildErrorMessage) {
+      buildErrorMessage = (err: ClerkAPIError[]) => err[0];
```
panteliselef, line 21: Just to confirm, is this how we support backwards compatibility with single error fields ?

panteliselef (approved):
Nicely done 🎉

packages/clerk-js/src/ui/utils/errorHandler.ts
nikosdouvlis: Let's create a function called `getFirstError` or similar to make this more explicit :)

nikosdouvlis (approved):

anagstef (approved):
🎉

packages/clerk-js/src/ui/utils/passwordUtils.ts
raptisj: @panteliselef I looked into this and it is slightly different. Merging would make it a bit verbose for no real benefit so I chose to keep them separate.

clerk-cookie:

PR #1520, opened 2023-07-25
ORG-83 userInvitations in useOrganizationList

panteliselef:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

### This PR
- Creates `UserOrganizationInvitation` and `UserOrganizationInvitationResource`
- Updates useOrganization to return `userInvitations`
- Adds the ability to aggregate the `userInvitations` when fetched from `useOrganization` (Very useful from infinite scrolling)

You might be familiar with `OrganizationInvitation` that already exists in our codebase.

Both `UserOrganizationInvitation` and `OrganizationInvitation` describe the same entity. The former is from the perspective of a user where the later from the perspective of an organization.

### Exposed API

```ts
  const {
    isLoaded,
    organizationList, // Same as before
    createOrganization, // Same as before
    setActive, // Same as before
    userInvitations: {
      data,
      count,
      isFetching,
      isLoading,
      isError,
      page,
      pageCount,
      fetchPage,
      fetchNext,
      fetchPrevious,
      hasNextPage,
      hasPreviousPage,
    },
  } = useOrganizationList({
    userInvitations: {
      page: 1,
      pageSize: 10,
      infinite: true, // Aggregate the data or not. Ideal for infinite lists
      keepPreviousData: true,
    },
  });
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42132,7 +42132,7 @@
       "dependencies": {
         "glob-to-regexp": "0.4.1",
         "js-cookie": "3.0.1",
-        "swr": "1.3.0"
+        "swr": "2.2.0"
       },
       "devDependencies": {
         "@clerk/types": "^3.48.1",
@@ -42143,6 +42143,17 @@
       },
       "peerDependencies": {
         "react": ">=16"
+      }
+    },
+    "packages/shared/node_modules/swr": {
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/swr/-/swr-2.2.0.tgz",
+      "integrity": "sha512-AjqHOv2lAhkuUdIiBu9xbuettzAzWXmCEcLONNKJRba87WAefz8Ca9d6ds/SzrPc235n1IxWYdhJ2zF3MNUaoQ==",
+      "dependencies": {
+        "use-sync-external-store": "^1.2.0"
+      },
+      "peerDependencies": {
+        "react": "^16.11.0 || ^17.0.0 || ^18.0.0"
       }
     },
     "packages/themes": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -37884,9 +37884,12 @@
       }
     },
     "node_modules/swr": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/swr/-/swr-1.3.0.tgz",
-      "integrity": "sha512-dkghQrOl2ORX9HYrMDtPa7LTVHJjCTeZoB1dqTbnnEDlSvN8JEKpYIYurDfvbQFUUS8Cg8PceFVZNkW0KNNYPw==",
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/swr/-/swr-2.2.0.tgz",
+      "integrity": "sha512-AjqHOv2lAhkuUdIiBu9xbuettzAzWXmCEcLONNKJRba87WAefz8Ca9d6ds/SzrPc235n1IxWYdhJ2zF3MNUaoQ==",
+      "dependencies": {
+        "use-sync-external-store": "^1.2.0"
+      },
       "peerDependencies": {
         "react": "^16.11.0 || ^17.0.0 || ^18.0.0"
       }
@@ -39901,7 +39904,6 @@
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/use-sync-external-store/-/use-sync-external-store-1.2.0.tgz",
       "integrity": "sha512-eEgnFxGQ1Ife9bzYs6VLi8/4X6CObHMw9Qr9tPY43iKwsPw8xE8+EFsf/2cFZ5S3esXgpWgtSCtLNS41F+sKPA==",
-      "peer": true,
       "peerDependencies": {
         "react": "^16.8.0 || ^17.0.0 || ^18.0.0"
       }
@@ -41402,10 +41404,10 @@
     },
     "packages/backend": {
       "name": "@clerk/backend",
-      "version": "0.26.0",
+      "version": "0.27.0",
       "license": "MIT",
       "dependencies": {
-        "@clerk/types": "^3.48.1",
+        "@clerk/types": "^3.49.0",
         "@peculiar/webcrypto": "1.4.1",
         "@types/node": "16.18.6",
         "cookie": "0.5.0",
@@ -41448,11 +41450,11 @@
     },
     "packages/chrome-extension": {
       "name": "@clerk/chrome-extension",
-      "version": "0.3.25",
+      "version": "0.3.26",
       "license": "MIT",
       "dependencies": {
-        "@clerk/clerk-js": "^4.54.2",
-        "@clerk/clerk-react": "^4.23.1"
+        "@clerk/clerk-js": "^4.55.0",
+        "@clerk/clerk-react": "^4.23.2"
       },
       "devDependencies": {
         "@types/chrome": "*",
@@ -41468,12 +41470,12 @@
     },
     "packages/clerk-js": {
       "name": "@clerk/clerk-js",
-      "version": "4.54.2",
+      "version": "4.55.0",
       "license": "MIT",
       "dependencies": {
-        "@clerk/localizations": "^1.24.0",
-        "@clerk/shared": "^0.20.0",
-        "@clerk/types": "^3.48.1",
+        "@clerk/localizations": "^1.24.1",
+        "@clerk/shared": "^0.21.0",
+        "@clerk/types": "^3.49.0",
         "@emotion/cache": "11.10.5",
         "@emotion/react": "11.10.5",
         "@floating-ui/react": "0.19.0",
@@ -41890,16 +41892,16 @@
     },
     "packages/expo": {
       "name": "@clerk/clerk-expo",
-      "version": "0.18.16",
+      "version": "0.18.17",
       "license": "MIT",
       "dependencies": {
-        "@clerk/clerk-js": "^4.54.2",
-        "@clerk/clerk-react": "^4.23.1",
+        "@clerk/clerk-js": "^4.55.0",
+        "@clerk/clerk-react": "^4.23.2",
         "base-64": "1.0.0",
         "react-native-url-polyfill": "1.3.0"
       },
       "devDependencies": {
-        "@clerk/types": "^3.48.1",
+        "@clerk/types": "^3.49.0",
         "@types/base-64": "^1.0.0",
         "@types/node": "^16.11.55",
         "@types/react": "*",
@@ -41918,11 +41920,11 @@
     },
     "packages/fastify": {
       "name": "@clerk/fastify",
-      "version": "0.6.2",
+      "version": "0.6.3",
       "license": "MIT",
       "dependencies": {
-        "@clerk/backend": "^0.26.0",
-        "@clerk/types": "^3.48.1",
+        "@clerk/backend": "^0.27.0",
+        "@clerk/types": "^3.49.0",
         "cookies": "0.8.0"
       },
       "devDependencies": {
@@ -41938,13 +41940,13 @@
       }
     },
     "packages/gatsby-plugin-clerk": {
-      "version": "4.4.3",
+      "version": "4.4.4",
       "license": "MIT",
       "dependencies": {
-        "@clerk/backend": "^0.26.0",
-        "@clerk/clerk-react": "^4.23.1",
-        "@clerk/clerk-sdk-node": "^4.12.1",
-        "@clerk/types": "^3.48.1",
+        "@clerk/backend": "^0.27.0",
+        "@clerk/clerk-react": "^4.23.2",
+        "@clerk/clerk-sdk-node": "^4.12.2",
+        "@clerk/types": "^3.49.0",
         "cookie": "0.5.0",
         "tslib": "2.4.1"
       },
@@ -41967,10 +41969,10 @@
     },
     "packages/localizations": {
       "name": "@clerk/localizations",
-      "version": "1.24.0",
+      "version": "1.24.1",
       "license": "MIT",
       "dependencies": {
-        "@clerk/types": "^3.48.1"
+        "@clerk/types": "^3.49.0"
       },
       "devDependencies": {
         "tsup": "*",
@@ -41985,13 +41987,13 @@
     },
     "packages/nextjs": {
       "name": "@clerk/nextjs",
-      "version": "4.23.1",
+      "version": "4.23.2",
       "license": "MIT",
       "dependencies": {
-        "@clerk/backend": "^0.26.0",
-        "@clerk/clerk-react": "^4.23.1",
-        "@clerk/clerk-sdk-node": "^4.12.1",
-        "@clerk/types": "^3.48.1",
+        "@clerk/backend": "^0.27.0",
+        "@clerk/clerk-react": "^4.23.2",
+        "@clerk/clerk-sdk-node": "^4.12.2",
+        "@clerk/types": "^3.49.0",
         "path-to-regexp": "6.2.1",
         "tslib": "2.4.1"
       },
@@ -42024,11 +42026,11 @@
     },
     "packages/react": {
       "name": "@clerk/clerk-react",
-      "version": "4.23.1",
+      "version": "4.23.2",
       "license": "MIT",
       "dependencies": {
-        "@clerk/shared": "^0.20.0",
-        "@clerk/types": "^3.48.1",
+        "@clerk/shared": "^0.21.0",
+        "@clerk/types": "^3.49.0",
         "tslib": "2.4.1"
       },
       "devDependencies": {
@@ -42051,13 +42053,13 @@
     },
     "packages/remix": {
       "name": "@clerk/remix",
-      "version": "2.9.0",
+      "version": "2.9.1",
       "license": "MIT",
       "dependencies": {
-        "@clerk/backend": "^0.26.0",
-        "@clerk/clerk-react": "^4.23.1",
-        "@clerk/shared": "^0.20.0",
-        "@clerk/types": "^3.48.1",
+        "@clerk/backend": "^0.27.0",
+        "@clerk/clerk-react": "^4.23.2",
+        "@clerk/shared": "^0.21.0",
+        "@clerk/types": "^3.49.0",
         "cookie": "0.5.0",
         "tslib": "2.4.1"
       },
@@ -42085,11 +42087,11 @@
     },
     "packages/sdk-node": {
       "name": "@clerk/clerk-sdk-node",
-      "version": "4.12.1",
+      "version": "4.12.2",
       "license": "MIT",
       "dependencies": {
-        "@clerk/backend": "^0.26.0",
-        "@clerk/types": "^3.48.1",
+        "@clerk/backend": "^0.27.0",
+        "@clerk/types": "^3.49.0",
         "@types/cookies": "0.7.7",
         "@types/express": "4.17.14",
         "@types/node-fetch": "2.6.2",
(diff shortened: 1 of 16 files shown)


packages/types/src/api.ts
```diff
@@ -19,3 +19,11 @@ export interface ClerkPaginationParams {
   limit?: number;
   offset?: number;
 }
+
+/**
+ * Pagination params
+ */
+export interface ClerkPaginationResponse<T> {
```
nikosdouvlis, line 26: 🙃 
```suggestion
export interface ClerkPaginatedResponse<T> {
```

package-lock.json
nikosdouvlis, line 1: Is this needed? If not, let's remove this from the PR - we'll update the lock file with a separate commit.

packages/clerk-js/src/core/resources/UserOrganizationInvitation.ts
```diff
@@ -0,0 +1,64 @@
...
+      .then(res => {
+        const { data: invites, total_count } =
+          res?.response as unknown as ClerkPaginationResponse<UserOrganizationInvitationJSON>;
+
+        return {
+          total_count,
+          data: invites.map(invitation => new UserOrganizationInvitation(invitation)),
+        };
+      })
+      .catch(() => ({
```
nikosdouvlis, line 41: Just checking - is this how we handle similar cases?

packages/clerk-js/src/ui/contexts/CoreClerkContextWrapper.tsx
```diff
@@ -3,14 +3,15 @@ import React from 'react';
...
-import { CoreOrganizationContext } from './CoreOrganizationContext';
+import { CoreOrganizationProvider } from './CoreOrganizationContext';
 import { CoreSessionContext } from './CoreSessionContext';
 import { CoreUserContext } from './CoreUserContext';
 import { assertClerkSingletonExists } from './utils';
 
 type CoreClerkContextWrapperProps = {
   clerk: Clerk;
   children: React.ReactNode;
+  swrConfig?: any;
```
nikosdouvlis, line 14: Could you please explain why we need this?

packages/react/package.json
```diff
@@ -40,7 +40,7 @@
   "dependencies": {
     "@clerk/shared": "^0.20.0",
     "@clerk/types": "^3.48.1",
-    "swr": "1.3.0",
+    "swr": "2.2.0",
```
nikosdouvlis, line 43: Let's rebase from main as this has been removed already :)

packages/shared/package.json
```diff
@@ -52,6 +52,6 @@
   "dependencies": {
     "glob-to-regexp": "0.4.1",
     "js-cookie": "3.0.1",
-    "swr": "1.3.0"
+    "swr": "2.2.0"
```
nikosdouvlis, line 55: Out of curiosity - did we face any issues with v1.3.0?

packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -74,11 +74,11 @@ export function ClerkContextProvider(props: ClerkContextProvider): JSX.Element |
     <IsomorphicClerkContext.Provider value={clerkCtx}>
       <ClientContext.Provider value={clientCtx}>
         <SessionContext.Provider value={sessionCtx}>
-          <OrganizationContext.Provider value={organizationCtx}>
+          <OrganizationProvider {...organizationCtx.value}>
```
nikosdouvlis, line 77: I guess we're treating this differently because of `swrConfig`... Not sure how I feel about this small discrepancy to be honest - if this is a testing-only issue, we might be able to find a different solution that works for the tests. What was the original issue?

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -1,39 +1,191 @@
...
+        count: number;
+        isLoading: boolean;
+        isValidating: boolean;
+        isError: boolean;
+        size: number | undefined;
+        setSize:
+          | ((
+              size: number | ((_size: number) => number),
+            ) => Promise<(ClerkPaginationResponse<UserOrganizationInvitationResource> | undefined)[] | undefined>)
+          | undefined;
```
nikosdouvlis, line 59: Can we probably use the `Dispatch<SetStateAction<...>>` interface from React?

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -1,39 +1,191 @@
...
     };
 
-type UseOrganizationList = () => UseOrganizationListReturn;
+type UseOrganizationList = (params?: UseOrganizationListParams) => UseOrganizationListReturn;
 
-export const useOrganizationList: UseOrganizationList = () => {
+export const useOrganizationList: UseOrganizationList = params => {
+  const { userInvitations } = params || {};
+
+  const triggerInfinite = userInvitations !== true ? userInvitations?.aggregate ?? false : false;
```
nikosdouvlis, line 68: Due to the type of `userInvitations`, this check is a little confusing. Let's make it more explicit:

```suggestion
  const triggerInfinite = typeof userInvitations === 'boolean' && userInvitations ? false : 
!!userInvitations?.aggregate;
```

(please double-check the above for correctness)

packages/shared/package.json
nikosdouvlis: Replying to myself: we're using `swr/infinite`

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -43,3 +195,12 @@ function createOrganizationList(organizationMemberships: OrganizationMembershipR
     organization: organizationMembership.organization,
   }));
 }
+
+function cacheKey(type: 'userInvitations', user: UserResource, pagination: GetUserOrganizationInvitations) {
```
nikosdouvlis, line 199: Do we care about typing these objects here? Do you think this would be enough? If it's generic enough in can be used in other similar places as well - other hooks using `swr` for example.
```suggestion
function cacheKey(...keys: Array<string | number | undefined>) {
```

packages/types/src/api.ts
panteliselef: I was thinking the same but then I decided to follow the pattern of `ClerkPaginationParams`. I could go back to `ClerkPaginatedResponse`

packages/react/package.json
panteliselef: Cool, thanks :)

packages/shared/package.json
panteliselef: Correct, v2 introduces swr/infinite + some nice properties return from useSWR like `isLoading`

packages/react/src/contexts/ClerkContextProvider.tsx
panteliselef: We need a way to clear cache between tests. The recommended way from the SWR docs [is this](https://swr.vercel.app/docs/advanced/cache#reset-cache-between-test-cases)

Wrapping our `CoreClerkContextWrapper` with <SWRConfig /> will not work. I verified this by logging the results of useSWRConfig inside our `useOrganization` while tests where running.

My solution was to expose a OrganizationProvider instead of the "low-level" context and include the `<SWRConfig/> provider` inside of the `OrganizationProvider`. 
☝️ **_This is now working as expecting._**

packages/shared/src/hooks/useOrganizationList.tsx
panteliselef: This is the type from SWR, `Dispatch<SetStateAction<...>>` returns void, right ?

Probably this will change, as I am working on a more uniform exposed API

packages/shared/src/hooks/useOrganizationList.tsx
panteliselef: I think you are correct. Although swr@2 seems to automatically parse objects as keys, which will make this function obsolete

packages/shared/src/hooks/useOrganizationList.tsx
nikosdouvlis: A type abstraction would be nice, but yes you're right :)

packages/shared/src/hooks/useOrganizationList.tsx
nikosdouvlis: We need to make sure to check whether `swr` uses referential equality checks or if it somehow builds a scalar value by paring the object's values - otherwise we might risk invalidating the cache on every render.

packages/react/src/contexts/ClerkContextProvider.tsx
nikosdouvlis: I see that the cache provider  follows the following interface: 
```
interface Cache<Data> {
  get(key: string): Data | undefined
  set(key: string, value: Data): void
  delete(key: string): void
  keys(): IterableIterator<string>
}
```

Did you try wrapping the test contexts with a `SWRConfig`  that uses a "noop" cache provider, eg: 
```
{
get: () => undefined,
set: noop,
delete: noop,
keys: () => []
}
```

so that we don't need to reset it between the tests?

panteliselef: I don't see why this would help. It seemed like the useSWRConfig within useOrganization would return the default values, as it couldn't reach the correct provider of SWRConfig. 

For example is I was passing <SWRConfig value={{ dedupingInterval: 0 }}/> the useSWRConfig would return `{dedupingInterval: 2000}` which is the default value

packages/shared/src/hooks/useOrganizationList.tsx
panteliselef: Does [this](https://swr.vercel.app/docs/arguments#passing-objects) answer your question ?

packages/clerk-js/src/ui/contexts/CoreClerkContextWrapper.tsx
panteliselef: This is passed from `MockClerkProvider` down to  `CoreOrganizationProvider` in order to clear swr cache between tests

packages/react/src/contexts/ClerkContextProvider.tsx
nikosdouvlis: Does this mean that a SWRConfig provider is found deeper in the tree so you cannot override the defaults in the tests?

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -1,39 +1,248 @@
...
+      userInvitations: PaginatedDataAPI<UserOrganizationInvitationResource>;
     };
 
-type UseOrganizationList = () => UseOrganizationListReturn;
+type UseOrganizationList = (params?: UseOrganizationListParams) => UseOrganizationListReturn;
+
+export const useOrganizationList: UseOrganizationList = params => {
+  const { userInvitations } = params || {};
+
+  const shouldUseDefaults = typeof userInvitations === 'boolean' && userInvitations;
```
SokratisVidros, line 70: 🔧 
```suggestion
  const shouldUseDefaultOptions = userInvitations === true
```

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -1,39 +1,248 @@
...
+  const [paginatedPage, setPaginatedPage] = useState(shouldUseDefaults ? 1 : userInvitations?.initialPage ?? 1);
+
+  // Cache initialPage and initialPageSize until unmount
+  const initialPageRef = useRef(shouldUseDefaults ? 1 : userInvitations?.initialPage ?? 1);
```
SokratisVidros, line 74: ❓ Do we need the last default fallback? Can't we ensure that the `initialPage` attribute always has a value?

SokratisVidros: Follow-up question? Do we need a ref for this?

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -1,39 +1,248 @@
...
+          initialPageSize: initialPageSizeRef.current,
+          status: internalStatus,
+        };
+
+  const canFetch = !!(clerk.loaded && user);
+
+  // Some gymnastics to adhere to the rules of hooks
+  // We need to make sure useSWR is called on every render
+  const fetchInvitations = !clerk.loaded
+    ? () => ({ data: [], total_count: 0 } as ClerkPaginatedResponse<UserOrganizationInvitationResource>)
```
SokratisVidros, line 98: ❓ Have we discussed the hook's return value before Clerk is ready? Is it possible to return `null` at this point as it's a bit of a code smell that we hardcode the format of the FAPI response in the hook? If not mistaken `null` is our go-to thing in other hooks for now, right?

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -1,39 +1,248 @@
...
+  const isomorphicIsError = !!(triggerInfinite ? userInvitationsInfiniteError : userInvitationsError);
+  /**
+   * Helpers
+   */
+  const fetchNext = useCallback(() => {
+    isomorphicSetPage(n => n + 1);
+  }, [isomorphicSetPage]);
+
+  const fetchPrevious = useCallback(() => {
+    isomorphicSetPage(n => n - 1);
```
SokratisVidros, line 189: 🔧 
```suggestion
    isomorphicSetPage(n => Math.max(0, n - 1));
```

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -1,39 +1,248 @@
...
+  }, [triggerInfinite, userInvitationsDataInfinite, userInvitationsData]);
+
+  const isomorphicIsLoading = triggerInfinite ? userInvitationsLoadingInfinite : userInvitationsLoading;
+  const isomorphicIsFetching = triggerInfinite ? userInvitationsInfiniteValidating : userInvitationsValidating;
+  const isomorphicIsError = !!(triggerInfinite ? userInvitationsInfiniteError : userInvitationsError);
+  /**
+   * Helpers
+   */
+  const fetchNext = useCallback(() => {
+    isomorphicSetPage(n => n + 1);
```
SokratisVidros, line 185: 🔧 
```suggestion
    isomorphicSetPage(n => Math.max(0, n + 1));
```

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -1,39 +1,248 @@
...
+      isFetching: isomorphicIsFetching,
+      isError: isomorphicIsError,
+      page: isomorphicPage,
+      pageCount,
+      fetchPage: isomorphicSetPage,
+      fetchNext,
+      fetchPrevious,
+      hasNextPage,
+      hasPreviousPage,
+      unstable__mutate,
```
SokratisVidros, line 244: ❓ Is it necessary to export mutate at this point since invitations are handled from ClerkJS/FAPI calls at this point?

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -1,39 +1,248 @@
...
+  }, [triggerInfinite, userInvitationsDataInfinite, userInvitationsData]);
+
+  const isomorphicCount = useMemo(() => {
+    if (triggerInfinite) {
+      return userInvitationsDataInfinite?.[userInvitationsDataInfinite?.length - 1]?.total_count || 0;
+    }
+    return userInvitationsData?.total_count ?? 0;
+  }, [triggerInfinite, userInvitationsDataInfinite, userInvitationsData]);
+
+  const isomorphicIsLoading = triggerInfinite ? userInvitationsLoadingInfinite : userInvitationsLoading;
```
SokratisVidros, line 178: ❓ What does isomorphic mean in this context? It feels that it's a bit confusing.

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -1,39 +1,248 @@
...
+  const {
+    data: userInvitationsDataInfinite,
+    isLoading: userInvitationsLoadingInfinite,
+    isValidating: userInvitationsInfiniteValidating,
+    error: userInvitationsInfiniteError,
+    size,
+    setSize,
+    mutate: userInvitationsInfiniteMutate,
+  } = useSWRInfinite(getInfiniteKey, ({ initialPage, initialPageSize, status }) => {
+    return !clerk.loaded || !user
```
SokratisVidros, line 137: ❓ I believe this hook should trigger the fetch only if triggerInfinite is set to true

SokratisVidros (changes requested):

packages/shared/src/hooks/useOrganizationList.tsx
```diff
@@ -1,39 +1,248 @@
...
+    | (GetUserOrganizationInvitationsParams & {
+        infinite?: boolean;
+        keepPreviousData?: boolean;
+      });
+};
+
 type OrganizationList = ReturnType<typeof createOrganizationList>;
 
+type CustomSetAction<T = unknown> = (size: T | ((_size: T) => T)) => void;
+type PaginatedDataAPI<T = unknown> = {
```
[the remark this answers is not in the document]
panteliselef: Done :)

packages/shared/src/hooks/useOrganizationList.tsx
panteliselef: initialPage does not always have a value as someone can simply pass as params the following 
```ts
useOrganizationList({infinite:true})
```
Here we want to not use the defaults in general but use the default value of initialPage

packages/shared/src/hooks/useOrganizationList.tsx
panteliselef: I'm not sure what you are referring to. Mutate will be used for revalidation, in order to refresh the data on demand

packages/shared/src/hooks/useOrganizationList.tsx
panteliselef: I needed a way to introduce parity between the returned values of useSWR and useSWRInfinite.

based on the params received in the hook return the appropriate values

packages/shared/src/hooks/useOrganizationList.tsx
panteliselef: It does check `getInfiniteKey`. The triggering is decided based on the key not the fetcher function. I updated that part and removed the conditional from this place to avoid confusions

packages/clerk-js/src/core/resources/UserOrganizationInvitation.ts
```diff
@@ -29,10 +30,22 @@ export class UserOrganizationInvitation extends BaseResource implements UserOrga
   static async retrieve(
     params?: GetUserOrganizationInvitationsParams,
   ): Promise<ClerkPaginatedResponse<UserOrganizationInvitation>> {
+    function convertPageToOffset(pageParams: GetUserOrganizationInvitationsParams | undefined): ClerkPaginationParams {
```
SokratisVidros, line 33: 🔧 This function can be a separate module as I assume it will be required in more resources soon.

packages/shared/src/hooks/useOrganizationList.tsx
SokratisVidros: The naming still feels a bit off. I'd just remove the isomorphic prefix.

SokratisVidros (approved):

packages/shared/src/hooks/useOrganizationList.tsx
panteliselef: Sure, i'll update this shortly in a following PR

clerk-cookie:

PR #1563, opened 2023-08-08
feat(clerk-js): Improve redirects on OAuth callback

kostaspt:
## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Currently, if you try to sign up with a provider that allows unverified accounts, the flow appears to be broken because it asks for the email after the OAuth callback. With this change, it will navigate to the appropriate change when needed.

#### Before

https://github.com/clerkinc/javascript/assets/684680/94298206-2a1b-4997-9adb-760908eec7fd

#### After

https://github.com/clerkinc/javascript/assets/684680/2b205fbb-6e19-43a8-9f2b-c6b6eedadbdf


packages/clerk-js/src/core/clerk.ts
```diff
@@ -873,6 +876,23 @@ export default class Clerk implements ClerkInterface {
         buildURL({ base: displayConfig.signUpUrl, hashPath: '/continue' }, { stringify: true }),
     );
 
+    const navigateToNextStepSignUp = () => {
+      if (su.unverifiedFields.length === 0 || su.missingFields.length !== 0) {
```
alex-ntousias, line 880: 🔧 We shouldn't care about unverified fields here. If we have missing fields, we need to prompt the user to complete them, i.e. navigate to continue sign up.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -873,6 +876,23 @@ export default class Clerk implements ClerkInterface {
...
+        signUp,
+        verifyEmailPath:
+          params.verifyEmailAddressUrl ||
+          buildURL({ base: displayConfig.signUpUrl, hashPath: '/verify-email-address' }, { stringify: true }),
+        verifyPhonePath:
+          params.verifyPhoneNumberUrl ||
+          buildURL({ base: displayConfig.signUpUrl, hashPath: '/verify-phone-number' }, { stringify: true }),
+        navigate,
+      });
+    };
```
alex-ntousias, line 894: 🔧 As discussed offline, let's reuse the `completeSignUpFlow` here and avoid introducing more utility functions.
Given that this is a change that will affect one of our core flows, let's keep the changeset as straight forward as possible, to make sure that FE folks can review it easier and spot any issues.
If a refactoring is required, we can always do it in a subsequent PR.

alex-ntousias (changes requested):
Nice work @kostaspt .
Added a comment to reduce the changes on this PR, so that FE folks can review it easier and spot any important issues.

packages/clerk-js/src/core/clerk.ts
kostaspt: This `handleMissingRequirements` is being extracted from `completeSignUpFlow` and used in this. Do you mean revert that and use it as it was?

packages/clerk-js/src/core/clerk.ts
gkats:
```suggestion
      if (su.missingFields.length) {
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -873,6 +876,23 @@ export default class Clerk implements ClerkInterface {
         buildURL({ base: displayConfig.signUpUrl, hashPath: '/continue' }, { stringify: true }),
     );
 
+    const navigateToNextStepSignUp = () => {
+      if (su.unverifiedFields.length === 0 || su.missingFields.length !== 0) {
+        return navigateToContinueSignUp();
```
gkats, line 881: :question: Is calling `navigateToContinueSignUp()` with no arguments the correct way to proceed here? I would expect that `navigateToNextStepSignUp` is built like the other navigation functions.

Why don't we put this logic inside `SignUpContinue`?

packages/types/src/clerk.ts
```diff
@@ -451,6 +451,16 @@ export type HandleOAuthCallbackParams = {
...
+
+  /**
+   * Full URL or path to navigate after requesting email verification.
+   */
+  verifyEmailAddressUrl?: string | null;
+
+  /**
+   * Full URL or path to navigate after requesting phone verification.
+   */
+  verifyPhoneNumberUrl?: string | null;
```
gkats, line 463: :question: Are these required for this fix? Seems like this is a broader API change for our components.

packages/clerk-js/src/core/clerk.ts
kostaspt: > ❓ Is calling `navigateToContinueSignUp()` with no arguments the correct way to proceed here? I would expect that `navigateToNextStepSignUp` is built like the other navigation functions.

Hmm, I'm not sure I get this. This function combines the functionality needed to determine the next step, making it reusable below. The other navigation functions don't have any logic.

> Why don't we put this logic inside `SignUpContinue`?

Do you mean the actual component? The `handleRedirectCallback` seems like a more appropriate place as it determines when to redirect next after the SSO callback.

packages/types/src/clerk.ts
kostaspt: It shouldn't need these, but for sign-in, using a relative URL won't work.

I want to redirect in these two cases:
- `sign-up/sso-callback` -> `sign-up/verify-email-address`
- `sign-in/sso-callback` -> `sign-up/verify-email-address`

For the first one to work, a relative URL like `../verify-email-address` is working as expected, but for the second one, the hash router will be necessary to navigate here.

Or I might be possibly missing something.

kostaspt:
Closing in favor of USR-279.

packages/clerk-js/src/core/clerk.ts
alex-ntousias: Yes, just reuse the existing `completeSignUpFlow`. I understand the motivation behind this change, but given how  critical this part of the code is, I would leave the refactoring as a separate piece of work.

packages/clerk-js/src/core/clerk.ts
gkats: > Why don't we put this logic inside SignUpContinue?

Please ignore this comment, it's a leftover. I had written it initially, but then read the PR more carefully and understood better what we're doing here. Sorry for the confusion.

My question about the `navigateToNextStepSignUp` is about making sure we're not missing any arguments that we might need to pass to navigateToContinueSignUp, such as the signUpUrl. Upon closer inspection, I see that this is already included by the `makeNavigate` call above. So, again, probably a void comment.

packages/types/src/clerk.ts
gkats: I see, thanks for explaining. I guess the problem is the the transfer flow from sign in?

We are on the `sign-in/` "base" path and we're mounting components in there. Wonder if there's an alternative where we always go through `sign-up/` first or we can somehow reuse the `signUpUrl`. I'll leave this comment open to get feedback from an expert in our javascript repo. :pray:

alex-ntousias (approved):
Nice work @kostaspt , thank you for addressing my comments. 🚀

anagstef (approved):
This is actually a good example of when we should add an E2E test. (cc @nikosdouvlis)

packages/types/src/clerk.ts
kostaspt: Hey @nikosdouvlis, there is no design document for this, but there is more information about the goal of this task in the ticket: https://linear.app/clerk/issue/USR-248

Do you have any concerns about these changes?

nikosdouvlis (approved):

clerk-cookie:

PR #1567, opened 2023-08-09
fix(clerk-js): Pass dev_browser to AP via query param, fix AP origin …

yourtallness:
…detection util

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

To make the dev_browser JWT available to the SSR context on Account Portal, we need to pass the JWT via the query, not the hash.

The current PR detects whether the URL we are navigating to is an AP URL & passes the db JWT as a query param.

Also fixed the AP origin detection util, which did not cover kima instances (ported logic from our backend).

Adding @chanioxaris to help verify if the AP origin detection logic is accurate.

### DIFF
--- a/packages/clerk-js/src/utils/devBrowser.ts
+++ b/packages/clerk-js/src/utils/devBrowser.ts
@@ -3,41 +3,62 @@ import { DEV_BROWSER_SSO_JWT_PARAMETER } from '../core/constants';
 export const DEV_BROWSER_JWT_MARKER = '__clerk_db_jwt';
 const DEV_BROWSER_JWT_MARKER_REGEXP = /__clerk_db_jwt\[(.*)\]/;
 
-function extractDevBrowserJWT(url: string): string {
-  const matches = url.match(DEV_BROWSER_JWT_MARKER_REGEXP);
-  return matches ? matches[1] : '';
+// Sets the dev_browser JWT in the hash or the search
+export function setDevBrowserJWTInURL(url: URL, jwt: string, asQueryParam: boolean): URL {
+  const resultURL = new URL(url);
+
+  // extract & strip existing jwt from hash
+  const jwtFromHash = extractDevBrowserJWTFromHash(resultURL.hash);
+  resultURL.hash = resultURL.hash.replace(DEV_BROWSER_JWT_MARKER_REGEXP, '');
+  if (resultURL.href.endsWith('#')) {
+    resultURL.hash = '';
+  }
+
+  // extract & strip existing jwt from search
+  const jwtFromSearch = resultURL.searchParams.get(DEV_BROWSER_SSO_JWT_PARAMETER);
+  resultURL.searchParams.delete(DEV_BROWSER_SSO_JWT_PARAMETER);
+
+  // Existing jwt takes precedence
+  const jwtToSet = jwtFromHash || jwtFromSearch || jwt;
+
+  if (jwtToSet) {
+    if (asQueryParam) {
+      resultURL.searchParams.append(DEV_BROWSER_SSO_JWT_PARAMETER, jwtToSet);
+    } else {
+      resultURL.hash = resultURL.hash + `${DEV_BROWSER_JWT_MARKER}[${jwtToSet}]`;
+    }
+  }
+
+  return resultURL;
 }
 
-export function setDevBrowserJWTInURL(url: string, jwt: string, asQueryParam: boolean): string {
-  if (asQueryParam) {
-    const hasQueryParam = (url || '').includes('?');
-    return `${url}${hasQueryParam ? '&' : '?'}${DEV_BROWSER_SSO_JWT_PARAMETER}=${(jwt || '').trim()}`;
+// Gets the dev_browser JWT from either the hash or the search
+// Side effect:
+// Removes dev_browser JWT from the URL as a side effect and updates the browser history
+export function getDevBrowserJWTFromURL(url: URL): string {
+  const resultURL = new URL(url);
+
+  // extract & strip existing jwt from hash
+  const jwtFromHash = extractDevBrowserJWTFromHash(resultURL.hash);
+  resultURL.hash = resultURL.hash.replace(DEV_BROWSER_JWT_MARKER_REGEXP, '');
+  if (resultURL.href.endsWith('#')) {
+    resultURL.hash = '';
   }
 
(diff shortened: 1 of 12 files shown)


panteliselef (approved):
Looks good to me 💯 but maybe wait for Haris as requested

packages/clerk-js/src/utils/url.ts
```diff
@@ -27,26 +27,53 @@ export const DEV_OR_STAGING_SUFFIXES = [
...
-    res = DEV_OR_STAGING_SUFFIXES.some(s => /^(https?:\/\/)?accounts\./.test(hostname) && hostname.endsWith(s));
-    accountsCache.set(hostname, res);
+    res = isLegacyDevAccountPortalOrigin(hostname) || isKimaDevAccountPortalOrigin(hostname);
+    accountPortalCache.set(hostname, res);
   }
+
   return res;
 }
 
+// Returns true for hosts such as:
```
chanioxaris, line 55: Maybe we should be a bit more strict with our checks here. For example we will return true for hostname `accounts.foo.13.bar-1.lcl.dev` which is wrong

packages/clerk-js/src/utils/url.ts
```diff
@@ -27,26 +27,53 @@ export const DEV_OR_STAGING_SUFFIXES = [
...
+// * accounts.foo.bar-13.lcl.dev
+// * accounts.foo.bar-13.lclstage.dev
+// * accounts.foo.bar-13.dev.lclclerk.com
+function isLegacyDevAccountPortalOrigin(host: string): boolean {
+  return LEGACY_DEV_SUFFIXES.some(legacyDevSuffix => {
+    return host.startsWith('accounts.') && host.endsWith(legacyDevSuffix);
+  });
+}
+
+// Returns true for hosts such as:
```
chanioxaris, line 65: Same here we can be more strict with our checks. For example we will return true for hostname `foo.bar.13.whatever.accounts.dev` which is wrong. Maybe use a regex instead?

packages/clerk-js/src/utils/url.ts
```diff
@@ -27,26 +27,53 @@ export const DEV_OR_STAGING_SUFFIXES = [
   'accounts.dev',
 ];
 
+export const LEGACY_DEV_SUFFIXES = ['.lcl.dev', '.lclstage.dev', '.lclclerk.com'];
+
+export const KIMA_DEV_SUFFIXES = ['.accounts.dev', '.accountstage.dev', '.accounts.lclclerk.com'];
```
SokratisVidros, line 32:
```suggestion
export const DEV_SUFFIXES = ['.accounts.dev', '.accountstage.dev', '.accounts.lclclerk.com'];
```

Kima was an internal project naming that shouldn't leak in the code.

packages/clerk-js/src/utils/url.ts
```diff
@@ -27,26 +27,53 @@ export const DEV_OR_STAGING_SUFFIXES = [
...
+  });
+}
+
+// Returns true for hosts such as:
+// * foo-bar-13.accounts.dev
+// * foo-bar-13.accountstage.dev
+// * foo-bar-13.accounts.lclclerk.com
+// But false for:
+// * foo-bar-13.clerk.accounts.lclclerk.com
+function isKimaDevAccountPortalOrigin(host: string): boolean {
```
SokratisVidros, line 71: Ditto about Kima

SokratisVidros (changes requested):

packages/clerk-js/src/utils/devBrowser.ts
```diff
@@ -3,41 +3,62 @@ import { DEV_BROWSER_SSO_JWT_PARAMETER } from '../core/constants';
...
+// Sets the dev_browser JWT in the hash or the search
+export function setDevBrowserJWTInURL(url: URL, jwt: string, asQueryParam: boolean): URL {
+  const resultURL = new URL(url);
 
-export function setDevBrowserJWTInURL(url: string, jwt: string, asQueryParam: boolean): string {
-  if (asQueryParam) {
-    const hasQueryParam = (url || '').includes('?');
-    return `${url}${hasQueryParam ? '&' : '?'}${DEV_BROWSER_SSO_JWT_PARAMETER}=${(jwt || '').trim()}`;
+  // extract & strip existing jwt from hash
+  const jwtFromHash = extractDevBrowserJWTFromHash(resultURL.hash);
```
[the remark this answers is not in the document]
yourtallness: Yes, but it might be in the hash & we need to move it to search or vice versa.

packages/clerk-js/src/utils/url.ts
yourtallness: I was thinking of a regex, but preferred to use the same logic as the BE has for this.

Note that this regex would also require that `.clerk` is not contained be contained and I'm a bit worried about using a negative lookahead (`?!`).

packages/clerk-js/src/utils/url.ts
yourtallness: Renamed

SokratisVidros (approved):

packages/clerk-js/src/utils/url.ts
```diff
@@ -27,26 +27,52 @@ export const DEV_OR_STAGING_SUFFIXES = [
   'accounts.dev',
 ];
 
+export const LEGACY_DEV_SUFFIXES = ['.lcl.dev', '.lclstage.dev', '.lclclerk.com'];
```
nikosdouvlis, line 30: Adding a comment for posterity/ future reference mostly, this change needs to be made to all similar helpers until we revamp clerk/shared.

clerk-cookie:

PR #1598, opened 2023-08-15
Limit debug logs to 4096 bytes on vercel to avoid errors

jescalan:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

When deployed to Vercel, logs cannot exceed 4096 bytes, or no logs are displayed at all and the user gets an error instead. This PR truncates debug logs to 4096 bytes to ensure that they display correctly on Vercel.

Fixes JS-608

### DIFF
--- a/packages/nextjs/src/utils/debugLogger.ts
+++ b/packages/nextjs/src/utils/debugLogger.ts
@@ -1,7 +1,5 @@
 // TODO: Replace with a more sophisticated logging solution
 
-import truncate from 'truncate-utf8-bytes';
-
 import { logFormatter } from './logFormatter';
 
 export type Log = string | Record<string, unknown>;
@@ -78,3 +76,15 @@ export const withLogger: WithLogger = (loggerFactoryOrName, handlerCtor) => {
     }
   }) as ReturnType<typeof handlerCtor>;
 };
+
+// ref: https://stackoverflow.com/questions/57769465/javascript-truncate-text-by-bytes-length
+function truncate(str: string, maxLength: number) {
+  const encoder = new TextEncoder();
+  const decoder = new TextDecoder('utf-8');
+
+  const encodedString = encoder.encode(str);
+  const truncatedString = encodedString.slice(0, maxLength);
+
+  // return the truncated string, removing any replacement characters that result from partially truncated characters
+  return decoder.decode(truncatedString).replace(/\uFFFD/g, '');
+}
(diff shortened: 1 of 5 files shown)


packages/nextjs/src/utils/debugLogger.ts
```diff
@@ -1,5 +1,7 @@
 // TODO: Replace with a more sophisticated logging solution
 
+import truncate from 'truncate-utf8-bytes';
```
BRKalow, line 3: I believe we intentionally try to mitigate usage of 3p dependencies in our packages. We can inline this pretty easily using `TextEncoder` and `TextDecoder`:

```js
function truncate(str: string, maxLength: number) {
	const encoder = new TextEncoder();
	const decoder = new TextDecoder('utf-8');

	const encodedString = encoder.encode(str);
	const truncatedString = encodedString.slice(0, maxLength + 1);

	// return the truncated string, removing any replacement characters that result from partially truncated characters
	return decoder.decode(truncatedString).replace(/\uFFFD/g, '');
}
```

([ref](https://stackoverflow.com/questions/57769465/javascript-truncate-text-by-bytes-length))

jescalan: Happy to shift over to this, but I was reluctant to do so initially since it doesn't handle multi-byte characters, where the library I included does, and [is hardly any more code](https://github.com/parshap/truncate-utf8-bytes/blob/master/lib/truncate.js) overall. We could copy-paste the 3p library code too, but that seems a little silly

jescalan: 🤷‍♂️ done!

BRKalow (approved):
LGTM! I think we need a `patch` changeset here and then it's good to go. Nice work

clerk-cookie:

PR #1658, opened 2023-08-30
Consume CAPTCHA errors during sign up and apply retries in case of error

chanioxaris:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description


SokratisVidros (approved):

packages/clerk-js/src/utils/captcha.ts
```diff
@@ -1,55 +1,132 @@
...
+          'error-callback': function (errorCode) {
+            /**
+             * By setting retry to 'never' the responsibility for implementing retrying is ours
+             * https://developers.cloudflare.com/turnstile/reference/client-side-errors/#retrying
+             */
+            if (retries < 2) {
+              setTimeout(() => {
+                captcha.reset(id);
+                retries++;
+              }, 500);
```
SokratisVidros, line 99: This might feel a bit too slow.

packages/clerk-js/src/utils/captcha.ts
```diff
@@ -1,55 +1,132 @@
...
+          sitekey,
+          retry: 'never',
+          // 'retry-interval': 500,
+          'refresh-expired': 'auto',
+          callback: function (token: string) {
+            resolve([token, id]);
+          },
+          'error-callback': function (errorCode) {
+            /**
+             * By setting retry to 'never' the responsibility for implementing retrying is ours
```
SokratisVidros, line 92: Why don't we switch to the default retry policy by CF?

SokratisVidros (changes requested):

packages/clerk-js/src/utils/captcha.ts
```diff
@@ -1,55 +1,132 @@
...
+      try {
+        const id = captcha.render(`.${WIDGET_CLASSNAME}`, {
+          sitekey,
+          retry: 'never',
+          // 'retry-interval': 500,
+          'refresh-expired': 'auto',
+          callback: function (token: string) {
+            resolve([token, id]);
+          },
+          'error-callback': function (errorCode) {
```
colinclerk, line 90: Cloudflare marks some error codes as ones that should not be retried.  Can we be selective about our retries?

Can we record all errors that arise during retries and pass them with comma separation to captcha_error, instead of just the last one?

packages/clerk-js/src/utils/captcha.ts
panteliselef: with retry:auto
once a challenge fails the  'error-callback' will be called.
and cf will decide if it want to retry.  For some error codes it will retry but for others it will not. (There is no way of knowing for which unless we transfer the logic that the mention in [their docs ](https://developers.cloudflare.com/turnstile/reference/client-side-errors/#error-codes)into our code which is not optimal).

It's important to note that if cf decides to retry, there is no stoping mechanism (e.g maxRetries) unless you explicitly remove the challenge

In our case their retry mechanism is not what we want as we simply need to attempt a few retries and then reject. With their implementation we cannot know when to reject because we don't know if the error-code  will trigger retries or not.
the default retry-interval is set to 8000ms (8s). I've set ours to 500ms which we can further decrease

:point_up: this is why i went with retry:never alongside manually calling .reset()
btw this is not a hack, [it is mentioned here](https://developers.cloudflare.com/turnstile/reference/client-side-errors/#retrying) in the "Retrying" seciont (edited)

packages/clerk-js/src/utils/captcha.ts
panteliselef: Tbh, i'd expect upon retries to get back the same error code. So create a list of it, may not bring to much value.

But yes we can more selective about the retries. We need to manually do it for now, sadly turnstile does not give the option to.

colinclerk: > Tbh, i'd expect upon retries to get back the same error code. So create a list of it, may not bring to much value

I agree, but it's easy to add the list, and it would be interesting if we see otherwise.  This is a bit of a black box to us, so let's be aggressive about logging the observability they do offer.

panteliselef: Sounds good, I already pushed one commit that opts out of retries based on error codes.

I will add one for the concatinated list of errors.


Btw what's ur take on going with our own retry mechanism ? I guess after these changes we can go with turnstile's way by passing `retry-interval`

packages/clerk-js/src/utils/captcha.ts
```diff
@@ -1,55 +1,139 @@
...
+    turnstile: Turnstile;
   }
 }
 
 const WIDGET_CLASSNAME = 'clerk-captcha';
 
+export const shouldRetryTurnstileErrorCode = (errorCode: string) => {
+  const codesWithRetries = ['crashed', 'undefined_error', '102', '103', '104', '106', '110600', '300100', '600'];
+
+  return !!(codesWithRetries.includes(errorCode) || codesWithRetries.find(w => errorCode.startsWith(w)));
```
nikosdouvlis, line 58: could probably simplify this to 
```suggestion
  return codesWithRetries.find(w => errorCode.startsWith(w));
```

SokratisVidros (approved):

packages/clerk-js/src/utils/__tests__/captcha.test.ts
```diff
@@ -0,0 +1,23 @@
+import { shouldRetryTurnstileErrorCode } from '../captcha';
+
+describe('wow', () => {
```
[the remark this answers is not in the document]
panteliselef: me and my dummy var names 😂. Changing it before merging

clerk-cookie:

PR #1667, opened 2023-08-31
chore(repo): Remove non-snapshot versions from snapshot comment

BRKalow:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Updates our snapshot workflow to filter out non-snapshot versions from the comment that gets posted. This should reduce noise and avoid confusion when people see non-snapshot versions in the table.

fixes JS-574


.github/workflows/release-snapshot.yml
```diff
@@ -53,18 +53,28 @@ jobs:
...
-            });
-
+            const fs = require("fs");
+            const files = await (await glob.create("./packages/*/package.json")).glob();
+            const descriptors = files
+              .map((file) => {
+                const { name, version } = JSON.parse(fs.readFileSync(file, "utf8"));
+                return { name, version };
+              })
+              .filter(({ version }) => version.includes("-snapshot"));
```
nikosdouvlis, line 63: Calling `!snapshot prefix` will trigger a snapshot release using `prefix` instead of `snapshot` for the version name - note that the package will still be published under the `snapshot` tag. Ref: https://github.com/clerkinc/javascript/blob/main/docs/PUBLISH.md#publishing-snapshot-package-versions-snapshot

Checking for `-snapshot` here won't catch versions with custom prefixes. I think we can simply interpolate the output of the `extract-snapshot-name` step here? Eg:

```suggestion
              .filter(({ version }) => version.includes("-${{ steps.extract-snapshot-name.outputs.name }}"));
```

nikosdouvlis (approved):
Good improvement! Honestly, this should be like that from the beginning, sorry for scaring you a couple of times :) 

Left a minor comment, everything else looks good.

.github/workflows/release-snapshot.yml
BRKalow: Good catch, forgot about that feature!

clerk-cookie:

PR #1695, opened 2023-09-07
fix(nextjs): Only ignore paths that have a '.' in the final path segment

BRKalow:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Updates our middleware default matcher and `IGNORED_ROUTES` pattern to allow matching of paths that contain a `.` that is not in the last path segment, but does not match paths that contain a `.` in the last segment (presumably this would be a static asset).

Fixes JS-694


packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -46,13 +46,13 @@ const INFINITE_REDIRECTION_LOOP_COOKIE = '__clerk_redirection_loop';
...
  */
-export const DEFAULT_CONFIG_MATCHER = ['/((?!.*\\..*|_next).*)', '/', '/(api|trpc)(.*)'];
+export const DEFAULT_CONFIG_MATCHER = ['/((?!.+[\\w]+\\.[\\w]+$|_next).*)', '/', '/(api|trpc)(.*)'];
 
 /**
  * Any routes matching this path will be ignored by the middleware.
  * This is the inverted version of DEFAULT_CONFIG_MATCHER.
  */
-export const DEFAULT_IGNORED_ROUTES = ['/((?!api|trpc))(_next|.+\\..+)(.*)'];
+export const DEFAULT_IGNORED_ROUTES = ['/((?!api|trpc))(_next.*|.+[\\w]+\\.[\\w]+$)'];
```
dimkl, line 55: ❓ why did we change it from `.next` to `_next.*`?

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -46,13 +46,13 @@ const INFINITE_REDIRECTION_LOOP_COOKIE = '__clerk_redirection_loop';
  * The default ideal matcher that excludes the _next directory (internals) and all static files,
  * but it will match the root route (/) and any routes that start with /api or /trpc.
  */
-export const DEFAULT_CONFIG_MATCHER = ['/((?!.*\\..*|_next).*)', '/', '/(api|trpc)(.*)'];
+export const DEFAULT_CONFIG_MATCHER = ['/((?!.+[\\w]+\\.[\\w]+$|_next).*)', '/', '/(api|trpc)(.*)'];
```
dimkl, line 49: ☁️ i think we should use the `.+\.[\w]+$` instead of `+[\w]+\\.[\w]+$` because filenames may contain symbols outside of `[a-zA-Z_0-9]` (eg `-`).

The next part is outside of the scope of this PR but I would like to open a discussion on that:
Could we split this regex into parts to make it more maintainable? Eg
```typescript
const allowAPIs = '/(api|trpc)(.*)';
const allowRootPaths = '/';
const ignoreNextUrls = '/(?!_next.*)';
const ignoreFileExtensions = '/(?!.+\.[\w]+$)';
export const DEFAULT_CONFIG_MATCHER = [allowAPIs, allowRootPaths, ignoreNextUrls, ignoreFileExtensions];
```

packages/nextjs/src/server/authMiddleware.ts
BRKalow: The pattern has changed slightly, we now want to check for:
* Any paths that start with `/_next` (denoted by `_next.*`)
* Any paths that end in `*.*`

Previously we didn't care where the `*.*` was, so we had a `(.*)` after the capture group.

packages/nextjs/src/server/authMiddleware.ts
BRKalow: Good suggestion. The negative rules need to remain together it seems like, otherwise one of them will match the other and they effectively cancel each other out, but I do like this!

SokratisVidros (approved):

packages/nextjs/src/server/authMiddleware.ts
dimkl: i would insist a bit on allowing any character before the file extension and not only the specified ones:
```
const VALID_FILENAME_CHARS = '\\w-\\$\\(\\)';
```
If the rest of the team agrees to keep only the limited set of chars, then you may proceed with merging it.

cc: @nikosdouvlis , @anagstef

dimkl (approved):

packages/nextjs/src/server/authMiddleware.ts
anagstef: I agree with @dimkl! I also think we should allow any character (including `~`) in the file name.

anagstef (approved):

packages/nextjs/src/server/authMiddleware.ts
BRKalow: Sounds good 👍

clerk-cookie:

PR #1704, opened 2023-09-11
fix(remix): Adjust RootAuthLoaderCallbackReturn to allow returning null

### LINKED ISSUE #1615: Remix rootAuthLoader callback's typescript types don't allow returning null.
- [X] Review the documentation: https://clerk.com/docs
- [X] Search for existing issues: https://github.com/clerkinc/javascript/issues
- [X] Go through package changelog files.
- [ ] Provide the Frontend API key from your application dashboard.

## Package + Version

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [X] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`
- [ ] other:

### Dependencies + versions

Provide a json with the dependencies used in your project (copy paste from yarn.lock / package-lock.json) or a github project / template that reproduces the issue.

**Include the @clerk/* packages and their versions!**

Example:

```json
{  
  "dependencies": {
    "react": "18.2.0",
    "@remix-run/cloudflare": "1.19.3",
    "@remix-run/cloudflare-pages": "1.19.3",
    "@remix-run/css-bundle": "1.19.3",
    "@remix-run/react": "1.19.3",
    "@clerk/remix": "2.9.1",
    "@clerk/clerk-react": "4.13.0"
  }
}

```

## Description

This one is fairly simple. In Remix, you're allowed to return "null" from a loader when it should return nothing. That's useful when you have conditional redirects, for example.

Eyeballing the code of @clerk/remix, it seems to be handling that correctly. [The types](https://github.com/clerkinc/javascript/blob/435d2cff5dfc86c58690d3f0d843f567ac4f3c04/packages/remix/src/ssr/types.ts) however, are declared separately to handle the overloads, and it loosely ends up with "Return a response or some data. If returning data, it has to extend `{}`", which does not allow null.

So something like this will file the type check:

```ts
export async function loader(args: LoaderArgs) {
  return rootAuthLoader(args, ({ request }) => {
    const url = new URL(request.url);
    if (url.pathname === "/") {
      return redirect("/mypage");
    }

    return null;
  });
}
```

As a workaround, you can replace `return null;` with `return {};`, so it isn't a big deal, but it would be nice for consistency to have the types 1:1 with the Remix types, which do allow returning null.

BRKalow:
## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

Adjust types in our Remix SDK to allow returning `null` from the `rootAuthLoader` callback. This aligns with how Remix's loaders work. They are allowed to return `null`, and so from a DX perspective it makes sense for us to support the same.

We still cannot support primitive values or arrays, as we aren't able to decorate this values with authentication state. I have updated the check and the underlying types to reflect this.

Here is Remix's loader error when an invalid value is returned, for reference:
> Error: You defined a loader for route "root" but didn't return anything from your loader function. Please return a value or null.

fixes JS-677
fixes #1615


panteliselef (approved):
Nice 🧹

BRKalow:
Looks like we also have a runtime check here, I'll need to update the logic there as well.

packages/remix/src/ssr/types.ts
```diff
@@ -35,7 +35,9 @@ export type RootAuthLoaderCallbackReturn =
   | Promise<Response>
   | Response
   | Promise<Record<string, unknown>>
-  | Record<string, unknown>;
+  | Record<string, unknown>
+  | Promise<null>
+  | null;
```
nikosdouvlis, line 40: 👋🏻 
Going through the Remix rootAuthLoader code, I see that we currently do not allow `null` to be returned from the rootAuthLoader callback. Take a look at [this assertion](https://github.com/clerkinc/javascript/blob/6c095fba9d37b1e657f5a87f3c1c2e56e0ac45cd/packages/remix/src/ssr/rootAuthLoader.ts#L54) and [this error](https://github.com/clerkinc/javascript/blob/42e05494cb692c2056be020d720dfca66679ecea/packages/remix/src/errors.ts#L51)

As I recall, the reason for this is that we rely on receiving an object-like response in which we can inject the auth state of the app in order to render the interstitial page, if needed, from the client-side components.

Do we have any specific use-cases where return `null` would be required?

BRKalow: Yep, I noticed the runtime checks! See the associated issue: #1615, though I'm not sure it makes sense for us to support it given that it's not just a type issue and we rely on an object being returned.

Alternatively, we could handle this case internally and make sure we return an object with the injected auth state even if nothing is returned.

BRKalow: @nikosdouvlis To improve DX, I've opted to allow returning `null` from the loader callback and have updated the logic to handle creating an empty response accordingly. Aligning with what Remix allows seems like a good choice here. Remix's error:

> Error: You defined a loader for route "root" but didn't return anything from your `loader` function. Please return a value or `null`.

packages/remix/src/ssr/types.ts
```diff
@@ -31,11 +31,14 @@ export type RootAuthLoaderCallback<Options extends RootAuthLoaderOptions> = (
   args: LoaderFunctionArgsWithAuth<Options>,
 ) => RootAuthLoaderCallbackReturn;
 
-export type RootAuthLoaderCallbackReturn =
-  | Promise<Response>
-  | Response
-  | Promise<Record<string, unknown>>
-  | Record<string, unknown>;
+type ObjectLike = Record<string, unknown> | null;
```
BRKalow, line 34: I think this provides some clarity around the intent of the type, along with the comment below.

packages/remix/src/ssr/rootAuthLoader.ts
```diff
@@ -67,6 +78,8 @@ export const rootAuthLoader: RootAuthLoader = async (
     }
   }
 
-  // if the user returned a plain object, create an empty response and inject requestState and handlerResult
-  return injectRequestStateIntoResponse(new Response(JSON.stringify(handlerResult)), requestState);
+  // if the return value of the user's handler is null or a plain object, create an empty response to inject Clerk's state into
+  const responseBody = JSON.stringify(handlerResult ?? {});
```
BRKalow, line 82: by this point `handlerResult` should be an object or null, so the use of `??` feels meaningful here.

panteliselef (approved):

packages/remix/src/ssr/types.ts
nikosdouvlis: Thank you Bryce, this makes sense.

This sounds like an edge scenario but I'd still like to make sure we are all aligned. In the following example, what would be the return value for the `[useLoaderData](https://remix.run/docs/en/main/hooks/use-loader-data)` hook? 

```
export const loader: LoaderFunction = args => {
  return rootAuthLoader(args, ({ request }) => {
    const { userId } = request.auth;
    return userId ? moviesRepo.getMoviesFor(userId) : null;
  });
};

function App() {
  const movies = useLoaderData();
  return (
    <html lang='en'>
     ...
}
```

Of course, the ideal way to right this would be to either use `[]` as a fallback value or wrap the return in a `{ movies: ... }` object.

dimkl (approved):

packages/remix/src/ssr/types.ts
nikosdouvlis: Bryce and I discussed about this offline, we're going to proceed with the changes shown in the PR :)

nikosdouvlis (approved):

clerk-cookie:

PR #1715, opened 2023-09-13
feat(shared): Introduce deprecate wrapper to output warnings for deprecated props/methods

dimkl:
## Description

Add wrapper function to be used in all the packages of this monorepo to warn customers in non-production environments about using a deprecated class method/prop or function.

A follow-up PR will be created to use this wrapper in the `@deprecated` props/methods.
I think since this is an internal tool there is no need to add any docs.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`


packages/shared/src/utils/deprecate.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+export const deprecate: DeprecateType = (fnOrPropName, warning, fnOrMethodOrCls, isStatic = false) => {
+  let hasWarned = false;
+
+  const warn = () => {
+    const isProduction = process.env.NODE_ENV === 'production';
+
+    if (isProduction) {
+      return;
+    }
```
anagstef, line 56: ❓ Should we go the other way around for this? `Do not show warning if it's not in development mode.`

packages/shared/src/utils/deprecate.ts
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * deprecate('something', 'Use `somethingElse` instead.', Example, true);
+ *
+ * 4. Deprecate function
+ * const getSomething = deprecate('getSomething', 'Use `getSomethingElse` instead.', () => {
+ *   return 'getSomethingValue';
+ * });
+ */
+export const deprecate: DeprecateType = (fnOrPropName, warning, fnOrMethodOrCls, isStatic = false) => {
+  let hasWarned = false;
```
SokratisVidros, line 47:
```suggestion
  let wasWarned = false;
```

or

```suggestion
  let warned = false;
```

packages/shared/src/utils/deprecate.ts
```diff
@@ -0,0 +1,91 @@
...
+ * }
+ *
+ * deprecate('something', 'Use `somethingElse` instead.', Example, true);
+ *
+ * 4. Deprecate function
+ * const getSomething = deprecate('getSomething', 'Use `getSomethingElse` instead.', () => {
+ *   return 'getSomethingValue';
+ * });
+ */
+export const deprecate: DeprecateType = (fnOrPropName, warning, fnOrMethodOrCls, isStatic = false) => {
```
SokratisVidros, line 46: ❓ Can we combine `fnOrPropName`, `fnOrMethodOrCls`, and `isStatic` into a single argument? Classes are functions and they have prototypes. I'd suggest trying to do it using the typeof operator.

SokratisVidros (changes requested):

packages/shared/src/utils/deprecate.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const warn = () => {
+    const isDevelopment = process.env.NODE_ENV === 'development';
+
+    if (!isDevelopment) {
+      return;
+    }
+
+    if (!hasWarned) {
+      console.warn(`${fnOrPropName} will be deprecated in the next major release.\n${warning}`);
```
BRKalow, line 57: ❓ Is this more accurate?
```suggestion
      console.warn(`${fnOrPropName} is deprecated and will be removed in the next major release.\n${warning}`);
```

dimkl: oops 😅

packages/shared/src/utils/deprecate.ts
dimkl: I am open to suggestions.
The reasons behind this cumbersome method signature:
- `fnOrPropName` and `fnOrMethodOrCls` are required for class properties since you cannot wrap them somehow and the only way I found to decorate them was, by rewriting the property on the class and output a console log when the property accessor is executed.
- `isStatic` is required to be set when a property is static because there is no way to check its existence if there is no default value set.
- if `fnOrPropName` is not passed when decorating the method, the warning should contain the deprecated method name since using the arrow function as `fnOrMethodOrCls` would result in showing a warning with missing information about what should be removed. We could enforce the use of named functions instead of arrow functions but this would cause an issue with decorated functions using `this` inside.

I could try to implement something like this, if it makes more sense (or split the it into 2 wrappers):
```typescript
interface DeprecateType {
    (warning: string, fnOrMethodOrCls: AnyClass, fnOrPropName: string): void;
    (warning: string, fnOrMethodOrCls: AnyFn, isStatic?: boolean): AnyFn;
  }

// method
class Example {
   getSomething = deprecate('Use `getSomethingElse` instead.', (arg1, arg2) => {
       return `getSomethingValue:${arg1 || '-'}:${arg2 || '-'}`;
    });
}

// function
const getSomething = deprecate('Use `getSomethingElse` instead.', (arg1, arg2) => {
    return `getSomethingValue:${arg1 || '-'}:${arg2 || '-'}`;
});

// property
class Example {
    something: string;
}

deprecate('Use `somethingElse` instead.', Example, 'something');


// static property
class Example {
    static something: string;
}

deprecate('Use `somethingElse` instead.', Example, 'something', true);
```
Does anyone have an example of how we could decorate (without using TS decorators) a class property without using fnOrPropName` and `fnOrMethodOrCls`?

dimkl:
@LekoArts we could create a utility function that will allow us to output a consistent message.

> And add a @deprecated JSDoc comment to the property/function in addition.
I didn't understand that. We already have marked all the deprecated class methods/properties, functions and constants with JSDoc `@deprecated` in our codebase.

LekoArts:
> I didn't understand that. We already have marked all the deprecated class methods/properties

Sorry if that wasn't clear! I meant that we should add those JSdoc comments when we'd use that utility. I didn't check if we already had that. But if that's already the case, even better!

dimkl:
I will open a new PR that will provide a `deprecated` wrapper for class properties and a `deprecationWarning` helper method to be used everywhere else.
It will combine this PR and the https://github.com/clerkinc/javascript/pull/1687

clerk-cookie:

PR #1743, opened 2023-09-18
feat(shared): Introduce deprecated/deprecatedProperty for deprecation warnings

dimkl:
## Description

This PR introduces `deprecated` and `deprecatedProperty` helpers that will be used to warn users when using a marked (already marked using JSDoc) class property/method or function by showing a console WARNING.
The warnings will be shown EACH the property is being accessed on the method is being called to urge the users to remove the deprecated usages.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

### DIFF
--- a/.changeset/seven-rivers-greet.md
+++ b/.changeset/seven-rivers-greet.md
@@ -1,0 +1,5 @@
+---
+'@clerk/shared': minor
+---
+
+Introduce `deprecated()` and `deprecatedProperty()` helper methods to report deprecation warnings once on console
--- a/packages/shared/src/utils/deprecated.test.ts
+++ b/packages/shared/src/utils/deprecated.test.ts
@@ -1,0 +1,259 @@
+jest.mock('./runtimeEnvironment', () => {
+  return { isDevelopmentEnvironment: jest.fn(() => true) };
+});
+
+import { deprecated, deprecatedProperty } from './deprecated';
+import { isDevelopmentEnvironment } from './runtimeEnvironment';
+
+describe('deprecated(fnName, warning)', () => {
+  let consoleWarnSpy;
+
+  beforeEach(() => {
+    consoleWarnSpy = jest.spyOn(global.console, 'warn').mockImplementation();
+  });
+  afterEach(() => {
+    consoleWarnSpy.mockRestore();
+  });
+
+  test('deprecate class method shows warning', () => {
+    class Example {
+      getSomeMethod = (arg1?, arg2?) => {
+        deprecated('getSomeMethod', 'Use `getSomeMethodElse` instead.');
+        return `getSomeMethodValue:${arg1 || '-'}:${arg2 || '-'}`;
+      };
+    }
+
+    const example = new Example();
+
+    expect(consoleWarnSpy).not.toBeCalled();
+    expect(example.getSomeMethod('a', 'b')).toEqual('getSomeMethodValue:a:b');
+    expect(consoleWarnSpy).toBeCalledTimes(1);
+    expect(consoleWarnSpy).toBeCalledWith(
+      'DEPRECATION WARNING: "getSomeMethod" is deprecated and will be removed in the next major release.\nUse `getSomeMethodElse` instead.',
+    );
+
+    expect(example.getSomeMethod()).toEqual('getSomeMethodValue:-:-');
+    expect(consoleWarnSpy).toBeCalledTimes(1);
+  });
+
+  test('deprecate static method shows warning', () => {
+    class Example {
+      static getSomeStaticMethod = () => {
+        deprecated('getSomeStaticMethod', 'Use `getSomeStaticMethodElse` instead.');
+        return 'getSomeStaticMethodValue';
+      };
+    }
+
+    expect(consoleWarnSpy).not.toBeCalled();
+    expect(Example.getSomeStaticMethod()).toEqual('getSomeStaticMethodValue');
+    expect(consoleWarnSpy).toBeCalledTimes(1);
+    expect(consoleWarnSpy).toBeCalledWith(
+      'DEPRECATION WARNING: "getSomeStaticMethod" is deprecated and will be removed in the next major release.\nUse `getSomeStaticMethodElse` instead.',
+    );
+
+    expect(Example.getSomeStaticMethod()).toEqual('getSomeStaticMethodValue');
+    expect(consoleWarnSpy).toBeCalledTimes(1);
+  });
+
+  test('deprecate function shows warning', () => {
+    const getSomeFunction = () => {
+      deprecated('getSomeFunction', 'Use `getSomeFunctionElse` instead.');
+      return 'getSomeFunctionValue';
+    };
+
+    expect(consoleWarnSpy).not.toBeCalled();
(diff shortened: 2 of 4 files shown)


packages/shared/src/utils/deprecated.ts
```diff
@@ -0,0 +1,61 @@
...
+ * }
+ *
+ * 2. Deprecate function
+ * const getSomething = () => {
+ *   deprecated('getSomething', 'Use `getSomethingElse` instead.');
+ *   return 'getSomethingValue';
+ * };
+ */
+export const deprecated = (fnName: string, warning: string): void => {
+  console.warn(`${fnName} is deprecated and will be removed in the next major release.\n${warning}`);
```
BRKalow, line 23: Can we restrict these warnings to development only? They aren't actionable in a production environment and will likely cause a lot of log pollution and user frustration.

Also, I understand wanting to be noisy here, but I think warning on every function call might be too much. That's potentially a lot of noise if we're calling this in a commonly used method. Users might not be able to upgrade immediately for some reason, and with no way to turn this off it could be frustrating.

Adding a module-level check seems like a reasonable approach, that way it would only warn once per session if in the browser, or once per process otherwise.

dimkl: >Can we restrict these warnings to development only? They aren't actionable in a production environment and will likely cause a lot of log pollution and user frustration.

Yes, we restrict those warnings to be visible in development mode using https://github.com/clerkinc/javascript/pull/1752

> Also, I understand wanting to be noisy here, but I think warning on every function call might be too much. That's potentially a lot of noise if we're calling this in a commonly used method. Users might not be able to upgrade immediately for some reason, and with no way to turn this off it could be frustrating.

Are you suggesting we make it called once?

> Adding a module-level check seems like a reasonable approach, that way it would only warn once per session if in the browser, or once per process otherwise.

Could you elaborate more on this? Could you provide a usage example?

LekoArts: > Are you suggesting we make it called once?

Yes, I'd suggest that, too.

For example, if `functionX` is used by the user in multiple places, it only makes sense to throw this warning the first time `functionX` is used. Then people know what they need to switch. All other occurrences wouldn't give any additional benefit to the user (and clutter the console).

> Could you elaborate more on this? Could you provide a usage example?

Instead of checking inside the `deprecate` function if it was already called, you move the scope to the **module scope**.

An example would be:

```ts
const displayedWarnings = new Set<string>()

export const reportOnce = (
  message: string,
  key?: string
): void => {
  const messageId = key ?? message
  if (!displayedWarnings.has(messageId) && !isDevelopmentMode) {
    console.warn(message)
  }
}
```

dimkl: @LekoArts I will proceed with your suggestion but I would like to add here that due to HMR in Remix and Nextjs the modules are reloaded in every request disabling the caching of information using module-level scope.

Relevant: https://www.prisma.io/docs/guides/other/troubleshooting-orm/help-articles/nextjs-prisma-client-dev-practices

PS: You can check the JS-164 internal ticket.

.changeset/seven-rivers-greet.md
```diff
@@ -0,0 +1,2 @@
+---
```
LekoArts, line 1: Changeset for a new minor of `@clerk/shared` to highlight that new helper 👍

packages/shared/src/utils/deprecated.test.ts
```diff
@@ -0,0 +1,201 @@
...
+    consoleWarnSpy = jest.spyOn(global.console, 'warn').mockImplementation();
+  });
+  afterEach(() => {
+    consoleWarnSpy.mockRestore();
+  });
+
+  test('deprecate class method shows warning', () => {
+    class Example {
+      getSomething = (arg1?, arg2?) => {
+        deprecated('getSomething', 'Use `getSomethingElse` instead.', 'key:getSomething:1');
```
LekoArts, line 21: Commenting here but that applies to all `deprecated` instances:

You're testing with the `key` provided here. It would be good to also add test cases where `key` is not given (since it's optional)

LekoArts (commented):
Only two smaller comments 👍

packages/shared/src/utils/deprecated.test.ts
dimkl: The issue is that having a module-scope variable (as cache) causes issues in testing since the deprecations are being silenced because they are cached as displayed if we are using the same key and message. To avoid exposing the cache (to use .clear() after each test) we should either use the key to generate a different key for each deprecation or use a different message.
The only options I could think of are:
- expose `displayedWarnings` Set and `.clear()` in after each test
- use different example (eg use different method names) per each test case
- provide keys in each test case so that each deprecation is unique

What are your thoughts on that? Is there any suggestion or preference to the above?
cc: @LekoArts

LekoArts: Since we want to test the codepath without a `key` the options that would rely on a `key` are not something we should do.

Use different `fnName` to keep things unique :)

dimkl: tests updated!

.changeset/seven-rivers-greet.md
dimkl: changeset updated!

packages/shared/src/utils/deprecated.test.ts
```diff
@@ -0,0 +1,201 @@
...
+    expect(consoleWarnSpy).not.toBeCalled();
+    expect(getSomeFunction()).toEqual('getSomeFunctionValue');
+    expect(consoleWarnSpy).toBeCalledTimes(1);
+    expect(consoleWarnSpy).toBeCalledWith(
+      'DEPRECATION WARNING: "getSomeFunction" is deprecated and will be removed in the next major release.\nUse `getSomeFunctionElse` instead.',
+    );
+
+    expect(getSomeFunction()).toEqual('getSomeFunctionValue');
+    expect(consoleWarnSpy).toBeCalledTimes(1);
+  });
```
LekoArts, line 73: Last nit:

Copy/paste that test and use a `key` for the deprecated function. Then no key + key usage is tested

dimkl: I've introduced the following 2 tests:
- deprecate function with key shows warning
- deprecate function with the same key does not show warning

LekoArts (approved):
Awesome 🎉

clerk-cookie:

PR #1760, opened 2023-09-21
chore(nextjs): Apply deprecation warnings

dimkl:
## Description

** REVIEW IT PER COMMIT **

Warn about deprecations of the following in Clerk Nextjs:

- `withClerkMiddleware`
- `withServerSideAuth`
- `CLERK_JS_VERSION` environment variable
- `CLERK_API_KEY` environment variable
- `NEXT_PUBLIC_FRONTEND_API` environment variable

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

### DIFF
--- a/.changeset/spicy-toys-change.md
+++ b/.changeset/spicy-toys-change.md
@@ -1,0 +1,5 @@
+---
+'@clerk/nextjs': patch
+---
+
+Warn about deprecations that will be dropped in next major version
--- a/packages/nextjs/src/server/clerkClient.ts
+++ b/packages/nextjs/src/server/clerkClient.ts
@@ -1,23 +1,7 @@
 /* eslint-disable turbo/no-undeclared-env-vars */
 import { Clerk } from '@clerk/backend';
 
-/**
- * @deprecated
- */
-export const JS_VERSION = process.env.CLERK_JS_VERSION || '';
-export const CLERK_JS_VERSION = process.env.NEXT_PUBLIC_CLERK_JS_VERSION || '';
-export const CLERK_JS_URL = process.env.NEXT_PUBLIC_CLERK_JS || '';
-export const API_URL = process.env.CLERK_API_URL || 'https://api.clerk.dev';
-export const API_VERSION = process.env.CLERK_API_VERSION || 'v1';
-export const API_KEY = process.env.CLERK_API_KEY || '';
-export const SECRET_KEY = process.env.CLERK_SECRET_KEY || '';
-export const FRONTEND_API = process.env.NEXT_PUBLIC_CLERK_FRONTEND_API || '';
-export const PUBLISHABLE_KEY = process.env.NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY || '';
-export const DOMAIN = process.env.NEXT_PUBLIC_CLERK_DOMAIN || '';
-export const PROXY_URL = process.env.NEXT_PUBLIC_CLERK_PROXY_URL || '';
-export const IS_SATELLITE = process.env.NEXT_PUBLIC_CLERK_IS_SATELLITE === 'true' || false;
-export const SIGN_IN_URL = process.env.NEXT_PUBLIC_CLERK_SIGN_IN_URL || '';
-export const SIGN_UP_URL = process.env.NEXT_PUBLIC_CLERK_SIGN_UP_URL || '';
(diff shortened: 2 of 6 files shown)


packages/nextjs/src/shared/deprecated.ts
```diff
@@ -0,0 +1,63 @@
+// copy of packages/shared/src/utils/deprecated.ts
```
dimkl, line 1: `@clerk/shared` package cannot be used in `@clerk/nextjs` server context due to the following error:


Screenshot 2023-09-21 at 21 32 30

LekoArts, line 1: We need to spend time on fixing this error and the copy/paste should only be the very last resort.

What happens if you put `"use client"` into the `usePagesOrInfinite` file?

packages/nextjs/src/shared/url.ts
```diff
@@ -0,0 +1,25 @@
+// TODO: Use the same function defined in @clerk/shared once the package is tree shakeable
```
SokratisVidros, line 1: ❓ What's the latest update on this? Is @clerk/shared still not tree-shakeable?

SokratisVidros (approved):

packages/nextjs/src/shared/deprecated.ts
dimkl: I wanted to avoid introducing changes in `@clerk/shared` that would require extensive testing in this PR so I followed the naive approach of copy / paste to proceed with this PR objective and introduce another PR to support importing `@clerk/shared` in server-side frameworks.
Based on your comment, if we consider blocking this PR until the @clerk/shared is fixed and used I have opened the following PR in which we should also check if the @clerk/shared is tree-shakeable or not before we merge it: https://github.com/clerkinc/javascript/pull/1767
cc: @SokratisVidros

packages/nextjs/src/shared/url.ts
dimkl: I will investigate it as part of https://github.com/clerkinc/javascript/pull/1767

dimkl: It seems like `@clerk/shared` is not entirely tree-shakeable due to `window.__SWR_DEVTOOLS_USE__ = undefined;`
I used the `@next/bundle-analyzer` and ran our A`NALYZE=true npm run build` to generate the bundle analysis below (the screenshot is part of the client.html analysis).
Screenshot 2023-09-27 at 17 49 19

```
// next.config.js
const withBundleAnalyzer = require('@next/bundle-analyzer')({
  enabled: process.env.ANALYZE === 'true',
})
module.exports = withBundleAnalyzer(nextConfig) 
```

I've created a discussion in the team channel to resolve this.

ref: https://github.com/clerkinc/javascript/blob/0a17322842393670344369b0217ed3e989f912bb/packages/shared/src/hooks/clerk-swr.ts#L4C1-L4C1

packages/nextjs/src/shared/deprecated.ts
dimkl: I have tested adding a deprecation in `authMiddleware` and `getAuth` and it seems that the tree-shaking works as expected and the above error is fixed!
I have also tried the `deprecated()` helper in `getServerSideProps` helper of a page and there is also no error. it seems that we can use it safely in server context.

packages/nextjs/src/shared/url.ts
dimkl: It seems that tree-shaking is fixed. 
The above dependency still exists in the client analysis because it is being used in some `@clerk/clerk-react` hooks that are being used by the clerk UI components.

clerk-cookie:

PR #1772, opened 2023-09-25
feat(backend,clerk-js,types): Support the new LinkedIn OAuth Provider

fragoulis:
## Description

LinkedIn has changed the OAuth flow by replacing it with OpenID Connect. We have already added support for this in the backend.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/types/src/oauth.ts
+++ b/packages/types/src/oauth.ts
@@ -19,6 +19,7 @@ export type DiscordOauthProvider = 'discord';
 export type TwitterOauthProvider = 'twitter';
 export type TwitchOauthProvider = 'twitch';
 export type LinkedinOauthProvider = 'linkedin';
+export type LinkedinOIDCOauthProvider = 'linkedin_oidc';
 export type DropboxOauthProvider = 'dropbox';
 export type AtlassianOauthProvider = 'atlassian';
 export type BitbucketOauthProvider = 'bitbucket';
@@ -45,6 +46,7 @@ export type OAuthProvider =
   | TwitterOauthProvider
   | TwitchOauthProvider
   | LinkedinOauthProvider
+  | LinkedinOIDCOauthProvider
   | DropboxOauthProvider
   | AtlassianOauthProvider
   | BitbucketOauthProvider
@@ -108,6 +110,12 @@ export const OAUTH_PROVIDERS: OAuthProviderData[] = [
     strategy: 'oauth_linkedin',
     name: 'LinkedIn',
     docsUrl: 'https://clerk.com/docs/authentication/social-connection-with-linkedin',
+  },
+  {
+    provider: 'linkedin_oidc',
+    strategy: 'oauth_linkedin_oidc',
+    name: 'LinkedIn',
+    docsUrl: 'https://clerk.com/docs/authentication/social-connections/linkedin-oidc',
   },
   {
     provider: 'github',
--- a/.changeset/serious-zebras-sip.md
+++ b/.changeset/serious-zebras-sip.md
@@ -1,0 +1,7 @@
+---
+'@clerk/clerk-js': minor
+'@clerk/backend': minor
+'@clerk/types': minor
+---
+
+Add support for LinkedIn OIDC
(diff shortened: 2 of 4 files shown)


anagstef (approved):

packages/types/src/oauth.ts
```diff
@@ -106,9 +108,15 @@ export const OAUTH_PROVIDERS: OAuthProviderData[] = [
   {
     provider: 'linkedin',
     strategy: 'oauth_linkedin',
-    name: 'LinkedIn',
+    name: 'LinkedIn (Deprecated)',
```
nikosdouvlis, line 111: ❓ @fragoulis Unless I'm mistaken, this is used by clerk-js as well. Is this visile to the end users as well?

fragoulis: Yes, this is what is shown in the Dashboard. See [here](https://github.com/clerkinc/dashboard/pull/1636). Do you think there would be a problem somewhere else?

SokratisVidros (approved):

packages/types/src/oauth.ts
nikosdouvlis: @fragoulis 
Unfortunately, it seems that this string is being used by the UI. Do we need the deprecated comment here? 
image

To share a little bit of context, It looks like this type and the utils that exist here are shared between the dashboard and the Clerk sdks. The `(deprecated)` string needs to be moved in the dashboard. We also should discuss how we want to approach this in the future with the dashboard team, as it seems that the types package includes actual runtime code like `getOAuthProviderData` and `sortedOAuthProviders` for the sole purpose of code reuse between the sdks and the dashboard. I believe, however, that the sdks should not be aware of the dashboard at all.

Happy to talk with you and Mark about this when you have some free time next week :)

nikosdouvlis (changes requested):

packages/types/src/oauth.ts
fragoulis: This feels completely out of my comfort zone.

fragoulis: image

image

These are the other places it is shown.

gkats: I think @nikosdouvlis is right. We don't want to display the "(Deprecated)" part in our components. We should probably create a new provider with the same name as the old one.

Something like the following
```js
{
  provider: "linkedin",
  strategy: "oauth_linkedin",
  name: "LinkedIn",
},
{  
  provider: "linkedin_oidc",
  strategy: "oauth_linkedin_oidc",
  name: "LinkedIn",
}
```

But we should make sure that only one of the two providers can be enabled.

fragoulis:
This is ready. The new provider's name will be set in dashboard.

nikosdouvlis (approved):

packages/types/src/oauth.ts
```diff
@@ -109,6 +111,12 @@ export const OAUTH_PROVIDERS: OAuthProviderData[] = [
     name: 'LinkedIn',
     docsUrl: 'https://clerk.com/docs/authentication/social-connection-with-linkedin',
   },
+  {
+    provider: 'linkedin_oidc',
+    strategy: 'oauth_linkedin_oidc',
+    name: 'LinkedIn',
+    docsUrl: 'https://clerk.com/docs/authentication/social-connections/linkedin-oidc',
```
desiprisg, line 118: This does not exist yet, right? Are we adding it?

kostaspt: @desiprisg Yes: https://github.com/clerkinc/clerk-docs/pull/342

clerk-cookie:

PR #1786, opened 2023-09-27
PLT 155 - Introduce jwt signing js backend

Nikpolik:
## Description

Added a new `signJwt(payload, secret, options)` function to  allow us to sign tokens.

Updated the `hasValidSignature(jwt, secret)` method to be able to verify tokens using PEM formatted keys.

The process for signing JWT was heavily inspired from another library [cloudflare-worker-jwt](https://github.com/tsndr/cloudflare-worker-jwt) and its [sign](https://github.com/tsndr/cloudflare-worker-jwt/blob/main/src/index.ts#L179-L221) method.

**Quick overview of process for signing JWT**

1. JSON.stringify header and payload
2. URL encode the two strings
3. Create the first part of the token `headerEncoded.payloadEncoded`
4. Import key
5. Signing using the algorithm provided in the options and encode signature
6. Return the final JWT `headerEncoded.payloadEncoded.signature`

**Changes implemented**

1. Extract isomorphicAtob to separate file to reuse for key importing.
2. Extract jwk to crypto algorithm mapping to separate file. This is will be shared between verify and sign.
3. Create shared importKey function that handles multiple key formats.
4. Use new importKey function inside hasValidSignature and verifyJWT also added tests for hasValidSignature.
6. Implement signing procedure described above.
7. Export new function.
8. Added tests and updated docs.

**Notes**

1. Since we only use *RSASSA-PKCS1-v1_5* currently only added support for importing keys formatted `jwk`,`pkcs8`,`spki` and not `raw`.
2. ~~signJwt was prefixed with __unstable__ since its goal is to be used internally. ~~ marked with JSDocs that will be used internally
3. ~~Isomorphic atob could be imported from `@clerk/shared` (https://github.com/clerkinc/javascript/pull/1769)~~ **Done**
4. ~~In the cloudflare-worker-jwt implementation of the sign method, the IAT (Issued at) field in the payload is updated. I am not sure if this is something that should be handled in the signing process or when we create the payload 🤔~~ **Changed to update IAT.**
5. ~~[verifyJWT](https://github.com/clerkinc/javascript/blob/975412ed5307ac81128c87289178bd1e6c2fb1af/packages/backend/src/tokens/jwt/verifyJwt.ts) method was not updated since it has its own implementation of `hasValidSignature` maybe it should use the updated one? This will allow us to use it PEM pem keys also.~~ **Fixed and removed duplicate (unused `hasValidSignature`)**

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`


packages/backend/src/shared/isomorphicAtob.ts
```diff
@@ -0,0 +1,10 @@
+const isomorphicAtob = (data: string) => {
+  if (typeof atob !== 'undefined' && typeof atob === 'function') {
+    return atob(data);
+  } else if (typeof globalThis !== 'undefined' && globalThis.Buffer) {
+    return globalThis.Buffer.from(data, 'base64').toString();
+  }
+  return data;
+};
+
+export default isomorphicAtob;
```
SokratisVidros, line 10: 🙃 Prefer named exports to default exports to avoid potential bundling issues.

SokratisVidros (commented):
@Nikpolik can you please elaborate on the comment about uat?

Nikpolik:
> @Nikpolik excellent PR description!
> 
> Can you please elaborate on the comment about iat? Is this about updating the iat of the token when we compute the signature?

Basically overwrite the iat `payload.iat = Math.floor(Date.now() / 1000)` to make sure that the signed token always has the correct timestamp of when it was signed. After thinking about it more clearly I need to add it here.

SokratisVidros (approved):

packages/backend/src/tokens/jwt/signJwt.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  const { module, test } = QUnit;
+
+  module('signJwt(payload, options)', () => {
+    test('signs a JWT with a JWK formatted secret', async assert => {
+      const jwt = await __unstable__signJwt(mockJwtPayload, signingJwks, {
+        algorithm: mockJwtHeader.alg,
+        header: mockJwtHeader,
+      });
+
+      assert.true(await hasValidSignature(decodeJwt(jwt), publicJwks));
```
SokratisVidros, line 24: 🙃 Isn't it more straightforward if you use `verifyJwt`?

packages/backend/src/tokens/jwt/signJwt.test.ts
```diff
@@ -0,0 +1,36 @@
...
+} from '../fixtures';
+import { __unstable__signJwt } from './signJwt';
+import { decodeJwt, hasValidSignature } from './verifyJwt';
+
+export default (QUnit: QUnit) => {
+  const { module, test } = QUnit;
+
+  module('signJwt(payload, options)', () => {
+    test('signs a JWT with a JWK formatted secret', async assert => {
+      const jwt = await __unstable__signJwt(mockJwtPayload, signingJwks, {
```
georgepsarakis, line 19: ❓ Is the generated signature static? Perhaps we can add an assertion for signature (string) equality if so. It makes sense to assert symmetric behavior with `verifyJwt` or `hasValidSignature` but it may also cause a edge case of "two wrongs making one right".

georgepsarakis (commented):
👏  Nice work! Overall seems fine to me, just not approving since I'm not so well-versed in JS/TS and there might be some subtle issue with crypto I'm missing.

packages/backend/src/tokens/jwt/signJwt.test.ts
Nikpolik: Yes! Initially I didn't update verifyJWT so I used its parts. Will change.

packages/backend/src/tokens/jwt/signJwt.test.ts
Nikpolik: No unfortunately the jwt is not static. 

Each time we sign the JWT it gets an updated IAT timestamp that changes the payload and the signature. 

The `hasValidSignature` test has its own check though with a pregenerated JWT for this reason.

nikosdouvlis (approved):

packages/backend/src/tokens/jwt/signJwt.ts
```diff
@@ -0,0 +1,45 @@
...
+function encodeJwtData(value: unknown): string {
+  const stringified = JSON.stringify(value);
+  const encoder = new TextEncoder();
+  const encoded = encoder.encode(stringified);
+  return base64url.stringify(encoded, { pad: false });
+}
+
+export async function __unstable__signJwt(
+  payload: Record<string, unknown>,
+  secret: string | JsonWebKey,
```
dimkl, line 19: 🙃 i would suggest we keep the same naming as the parameter used in `verifyJwt` 
```
key:  JsonWebKey | string,
```

packages/backend/src/tokens/jwt/signJwt.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function encodeJwtData(value: unknown): string {
+  const stringified = JSON.stringify(value);
+  const encoder = new TextEncoder();
+  const encoded = encoder.encode(stringified);
+  return base64url.stringify(encoded, { pad: false });
+}
+
+export async function __unstable__signJwt(
```
dimkl, line 17: ☁️ I would suggest we expose it as `signJwt` and add a JSDoc to indicate that this function will be used internally.
Since this method does not contain Clerk-specific logic, why not expose it without the `__unstable` prefix?
cc: @nikosdouvlis , @SokratisVidros

dimkl (approved):

packages/backend/src/tokens/jwt/assertions.test.ts
```diff
@@ -153,26 +153,35 @@ export default (QUnit: QUnit) => {
       assert.equal(undefined, assertHeaderAlgorithm('RS256'));
       assert.equal(undefined, assertHeaderAlgorithm('RS384'));
       assert.equal(undefined, assertHeaderAlgorithm('RS512'));
-      assert.equal(undefined, assertHeaderAlgorithm('ES256'));
-      assert.equal(undefined, assertHeaderAlgorithm('ES384'));
-      assert.equal(undefined, assertHeaderAlgorithm('ES512'));
```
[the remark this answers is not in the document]
Nikpolik: I am not sure this actually worked before 😅 

If you read the instructions for importing [ECDSA](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign#ecdsa) or [ECDH](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/deriveKey#ecdh) [keys](https://developer.mozilla.org/en-US/docs/Web/API/EcKeyImportParams) it expectes something like `{ name: 'ECDSA', namedCurve: 'P-256' }` and not `{ name: 'ECDSA', hash: 'SHA-256' }` this means that when using these types of algorithms and keys you would get an error.

Screenshot 2023-10-11 at 10 45 55 AM

Additionally if you look at the previous implementation of `hasValidSignature` it has hardcoded the algorithm to `RSASSA-PKCS1-v1_5`

`return runtime.crypto.subtle.verify('RSASSA-PKCS1-v1_5', cryptoKey, signature, data);`  this means that it even if the key importing worked, the `hasValidSignature` could only actually accept RSA keys.

So I decided to remove them for now and we can decide later if we want to support all kinds of algorithms (there is also HMAC to consider) when clerk only uses RSA.

This is the same reason I did not add support for `raw` format of keys.

packages/backend/src/tokens/jwt/signJwt.test.ts
Nikpolik: Changed https://github.com/clerkinc/javascript/pull/1786/commits/c21e08a5c4363f2be544af4945a040a07f114f08

packages/backend/src/tokens/jwt/signJwt.ts
georgepsarakis: @dimkl the naming convention helps with also preventing confusion on the user's side on whether they should use this function in their flow or not as well. We could add a JSDoc annotation regardless though.

packages/backend/src/tokens/jwt/assertions.test.ts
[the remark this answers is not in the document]
dimkl: Nice!! :)

packages/backend/src/tokens/jwt/signJwt.ts
dimkl: I think we should not use the approach of exposing internal methods as `__unstable`, since it does not keep this as internal and it pollutes the public API of our package with "weird" prefixed exports. I know that `__unstable` prefixed exports are used in some parts of our packages, but I think that we should only use them as a dirty way to support a case that we don't have another to way to support at the moment (eg `__unstable_options`).

If it's something users should not use, why add it in `@clerk/backend`?
How possible is it that the `signJwt` signature will change with a breaking change?

I am okay with `__unstable_signJwt` but my personal suggestion is:
- export `signJwt`
- add JSDoc to indicate that this is internal
- (optional) add docs for `signJwt` to explain that this should not be used to generate clerk JWTs, that is being used internally and that it could be used as a helper to generate a signature for other cases expect creating a Clerk JWT (eg sign a payload in the customer application server).

.changeset/cold-bags-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
+Added \_\_unstable\_\_signJwt
```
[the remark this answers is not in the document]
Nikpolik: Updated the changelogs https://github.com/clerkinc/javascript/pull/1786/commits/6470ea28b0644efebd8d418fcb950c4f7b22a594

packages/backend/src/tokens/jwt/algorithms.ts
```diff
@@ -0,0 +1,66 @@
...
+    throw new Error(`Unsupported algorithm ${algorithmName}, expected one of ${algs.join(',')}.`);
+  }
+
+  return {
+    hash: { name: algToHash[algorithmName] },
+    name: jwksAlgToCryptoAlg[algorithmName],
+  };
+}
+
+export function importKey(
```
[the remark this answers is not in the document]
Nikpolik: Moved in in an external file and also added some tests https://github.com/clerkinc/javascript/pull/1786/commits/53f655916d39fb700ab7ab484fef8b0efe5046e6

packages/backend/src/tokens/jwt/signJwt.ts
Nikpolik: Went with this approach 😄  Also added js docs about the signature of the function!

clerk-cookie:

PR #1803, opened 2023-09-29
feat(clerk-js,types): Add appearance keys for pagination and table head

octoper:
## Description

Add missing appearance keys for Pagination text and buttons and Table head.

### Before

Before

### After

After

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [X] `@clerk/types`
- [x] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/clerk-js/src/ui/elements/Pagination.tsx
+++ b/packages/clerk-js/src/ui/elements/Pagination.tsx
@@ -1,6 +1,6 @@
 import { useState } from 'react';
 
-import { Button, Flex, localizationKeys, Text } from '../customizables';
+import { Button, descriptors, Flex, localizationKeys, Text } from '../customizables';
 import type { PropsOfComponent } from '../styledSystem';
 import { mqu } from '../styledSystem';
 import { range } from '../utils';
@@ -31,6 +31,7 @@ const PageButton = (props: PropsOfComponent<typeof Button> & { isActive?: boolea
         },
         sx,
       ]}
+      elementDescriptor={descriptors.paginationPageButton}
       {...rest}
     />
   );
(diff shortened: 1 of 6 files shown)


packages/clerk-js/src/ui/elements/Pagination.tsx
```diff
@@ -52,32 +53,27 @@ const RowInformation = (props: RowInfoProps) => {
...
         as='span'
-        sx={t => ({
-          color: t.colors.$blackAlpha700,
-        })}
+        colorScheme='inherit'
         localizationKey={localizationKeys('paginationRowText__displaying')}
       />{' '}
       <Text
         as='span'
+        colorScheme='inherit'
```
nikosdouvlis, line 64: @octoper I see we made some minor UI changes here, could you please add a screenshot in the PR description if needed? Even if the changes are minor, this always helps and we can use it for future reference as well.

desiprisg: Yes, let's add a before and after showcasing why we need to use `inherit` here.

octoper: The main reason I used `colorScheme='inherit'` here is just to have the colour to the parent element so I can use the element descriptor only theme.

nikosdouvlis (approved):
This looks good :) Approving to unblock, let's discuss about the type issue offline

packages/clerk-js/src/ui/elements/Pagination.tsx
panteliselef: @octoper I'm very skeptic about the result of the before/active. This is something i recently touched as well, but endup not making the change.

I don't think using the primary colour here makes sense, and the result is not that elegant.

octoper: @panteliselef Yeah it's not that elegant but I thought that this should match the other parts of our components that have text, because if this is not changed every other part will have the primary colour and this will be black.

panteliselef (approved):

dimkl (approved):

clerk-cookie:

PR #1813, opened 2023-10-02
feat(clerk-js,shared,types): Add ClerkRuntimeError class

panteliselef:
## Description

With the introduction of ClerkRuntimeError class, we can localize error messages and display them in ClerkJS components. This functionality existed for FAPI errors, and we are now adding support for runtime errors.

### Example
Displays an error message caused when lazy loading a 3rd party library. In this case, the sign up process cannot continue and we need to provided the end user with some feedback.
Screenshot 2023-10-02 at 2 12 53 PM

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/shared/src/errors/Error.ts
+++ b/packages/shared/src/errors/Error.ts
@@ -14,11 +14,35 @@ export interface MetamaskError extends Error {
 }
 
 export function isKnownError(error: any) {
-  return isClerkAPIResponseError(error) || isMetamaskError(error);
+  return isClerkAPIResponseError(error) || isMetamaskError(error) || isClerkRuntimeError(error);
 }
 
 export function isClerkAPIResponseError(err: any): err is ClerkAPIResponseError {
+  if (err instanceof ClerkAPIResponseError) {
+    return true;
+  }
+
   return 'clerkError' in err;
+}
+
+/**
+ * Checks if the provided error object is an instance of ClerkRuntimeError.
+ *
+ * @param {any} err - The error object to check.
+ * @returns {boolean} True if the error is a ClerkRuntimeError, false otherwise.
+ *
+ * @example
+ * const error = new ClerkRuntimeError('An error occurred');
+ * if (isClerkRuntimeError(error)) {
+ *   // Handle ClerkRuntimeError
+ *   console.error('ClerkRuntimeError:', error.message);
+ * } else {
+ *   // Handle other errors
+ *   console.error('Other error:', error.message);
+ * }
+ */
+export function isClerkRuntimeError(err: any): err is ClerkRuntimeError {
+  return err instanceof ClerkRuntimeError;
 }
 
 export function isMetamaskError(err: any): err is MetamaskError {
@@ -44,8 +68,6 @@ export function parseError(error: ClerkAPIErrorJSON): ClerkAPIError {
 }
 
 export class ClerkAPIResponseError extends Error {
-  clerkError: true;
-
   status: number;
   message: string;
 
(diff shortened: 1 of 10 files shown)


packages/shared/src/errors/Error.ts
```diff
@@ -69,6 +73,25 @@ export class ClerkAPIResponseError extends Error {
   };
 }
 
+export class ClerkRuntimeError extends Error {
```
BRKalow, line 76: Given that we are adding new exports here, can we add JSDoc comments above `isClerkRuntimeError` and `ClerkRuntimeError` briefly explaining intended usage?

BRKalow (approved):

packages/shared/src/errors/Error.ts
```diff
@@ -69,6 +90,58 @@ export class ClerkAPIResponseError extends Error {
...
+  }
+
+  /**
+   * Returns a string representation of the error.
+   *
+   * @returns {string} A formatted string with the error name and message.
+   * @memberof ClerkRuntimeError
+   */
+  public toString = () => {
+    return `[${this.name}]\nMessage:${this.message}`;
```
dimkl, line 141: ❓ If the `toString()` is used to show the localized error message to the end-user I think we should probably need to remove the `this.name` prefix and show the localized message. Otherwise, shouldn't we also add the `code` ?

packages/shared/src/errors/Error.ts
```diff
@@ -69,6 +90,58 @@ export class ClerkAPIResponseError extends Error {
...
+ *   throw new ClerkRuntimeError('An error occurred', { code: 'password_invalid' });
+ */
+export class ClerkRuntimeError extends Error {
+  /**
+   * Indicates that this is a Clerk runtime error.
+   *
+   * @type {boolean}
+   * @memberof ClerkRuntimeError
+   */
+  clerkRuntimeError: true;
```
dimkl, line 107: ❓ should we use the `clerkError` instead? Also since `isClerkRuntimeError` helper exists, do we need this?

dimkl (approved):
❓ Instead of introducing a new error class, couldn't we add the `code` to our existing error classes (if exists) and use a helper `isTranslatableError(error)` that will check if the `code` exists in the supported localization keys to translate the error?
I left some questions in the PR but i will approve this since I may be missing some context.

packages/shared/src/errors/Error.ts
panteliselef: `toString` is not used to display a localized message.

This class cannot have access to our localization mechanism that exist in `clerk-js` as this would introduce a cyclic dependency.

`translateError` inside our UI layer is in charge of localizing.

We are only interested in localizating the errors displayed in the UI, not the ones we are logging to console.

packages/shared/src/errors/Error.ts
panteliselef: I removed this one

dimkl: i know. i'm saying that instead of removing it, using it here.

clerk-cookie:

PR #1835, opened 2023-10-06
fix(clerk-react,remix,shared): Make `exports.types` the first key in `exports`

nikosdouvlis:
## Description
The [TypeScript docs](https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing) recommends so, but it's also because the "exports" field is order-based.

For example, a scenario where both the "types" and "import" condition could be active, "types" should be first so that it matches and returns a .d.ts file, rather than a .js file from the "import" condition.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [x] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/remix/package.json
+++ b/packages/remix/package.json
@@ -18,9 +18,18 @@
   "types": "dist/index.d.ts",
   "sideEffects": false,
   "exports": {
-    ".": "./dist/index.js",
-    "./ssr.server": "./dist/ssr/index.js",
-    "./api.server": "./dist/api/index.js"
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./ssr.server": {
+      "types": "./dist/ssr/index.d.ts",
+      "default": "./dist/ssr/index.js"
+    },
+    "./api.server": {
+      "types": "./dist/api/index.d.ts",
+      "default": "./dist/api/index.js"
+    }
   },
   "typesVersions": {
     "*": {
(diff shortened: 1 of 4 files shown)


packages/remix/package.json
```diff
@@ -18,9 +18,18 @@
   "types": "dist/index.d.ts",
   "sideEffects": false,
   "exports": {
-    ".": "./dist/index.js",
-    "./ssr.server": "./dist/ssr/index.js",
-    "./api.server": "./dist/api/index.js"
+    ".": {
```
nikosdouvlis, line 21: @anagstef @LekoArts 
Could you please verify that this is correct?

LekoArts: LGTM. Looking at https://webpack.js.org/guides/package-exports/#conditional-syntax

> The last condition in the object might be the special "default" condition, which is always matched.

I'm not 100% sure if we could use `require` here, but `default` seems like a safe bet for now

nikosdouvlis: Exactly - used default as it should mimic the `exports` behavior we had before. Thank you!

desiprisg (approved):

anagstef (approved):

dimkl (approved):

clerk-cookie:

PR #1868, opened 2023-10-12
fix(shared): Remove testUtils & smaller updates

LekoArts:
## Description

While working on JS-509 I quickly realized that putting that into one big PR would be a mess. So this is a slice of work that can be shipped separately and removes some noise from the original purpose of that ticket.

This PR does a couple of things:

- Remove `testUtils` from `@clerk/shared`. There's no need to put this into such a package, it can live inside the repo outside of a npm package. The contents of the `testUtils` were re-exports from `@testing-library/react` and an unused wrapper. So I didn't have to move it somewhere else, I just had to update the imports
- Fix the Jest config `@clerk/shared`. Through an incorrect `testRegex` one test was never run. Jest's defaults are fine for our use case, so we don't even need to set it
- Remove the `package.json` we put into `dist/cjs` and `dist/esm`. They have no benefit since the root `package.json` already has `sideEffects` defined.
- Add [`peerDependenciesMeta`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta) to make `react` peerDep optional
- Use correct `LICENSE`
- Add `author`

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`


packages/clerk-js/src/utils/__tests__/ignoreEventValue.test.ts
```diff
@@ -1,7 +1,8 @@
-import { noop } from '@clerk/shared/testUtils';
-
 import { ignoreEventValue } from '../ignoreEventValue';
 
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noop = (..._args: any[]): void => {};
```
LekoArts, line 4: IMO fine to inline since this was only used once

packages/shared/src/testUtils/testUtils.ts
LekoArts, line 1: These were just all re-exports that did nothing

packages/clerk-js/setupJestAfterEnv.ts
```diff
@@ -1,11 +1 @@
 import '@testing-library/jest-dom';
```
LekoArts, line 1: We had `@testing-library/jest-dom/extend-expect` in our testUtils file. This line already covers this

.changeset/serious-meals-wink.md
```diff
@@ -0,0 +1,9 @@
+---
+"@clerk/shared": patch
```
dimkl, line 2: 🔧 I think that changing the public API (eg removing an export in our case) of a package requires a major version bump.

packages/clerk-js/src/utils/__tests__/ignoreEventValue.test.ts
dimkl: 🔧 There are more references of noop in the codebase:
Screenshot 2023-10-12 at 13 08 39

dimkl (approved):

packages/clerk-js/src/utils/__tests__/ignoreEventValue.test.ts
LekoArts: `import { noop } from '@clerk/shared/testUtils` was only used once. The other imports won't be touched by the testUtils change

.changeset/serious-meals-wink.md
LekoArts: It's an internal package without a README only used by our Jest tests that we also don't tell people to install or use. It'd consider it purely internal.

So while technically you're correct, I wouldn't care about that :)

BRKalow (approved):

nikosdouvlis (approved):

nikosdouvlis (commented):
Awesome job @LekoArts 🔥 

> Remove the package.json we put into dist/cjs and dist/esm. They have no benefit since the root package.json already has sideEffects defined.

For future reference, old Vite React apps weren't picking up the root package.json's `sideEffects` prop. It seems that this has been fixed since then so all changes in this PR as spot on.

clerk-cookie:

PR #1877, opened 2023-10-13
chore(repo,clerk-js): Update TS to v5 & Update emotion minor

LekoArts:
## Description

So to be able to use the `Bundler` option in https://www.typescriptlang.org/tsconfig#moduleResolution we need TypeScript v5. 5.1 introduced something new for JSX: https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/#decoupled-type-checking-between-jsx-elements-and-jsx-tag-types

To make TS v5 and emotion work, we need to update emotion a little bit.
Both updates are harmless:
- https://github.com/emotion-js/emotion/releases/tag/%40emotion%2Freact%4011.11.0
- https://github.com/emotion-js/emotion/releases/tag/%40emotion%2Freact%4011.11.1

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [x] `build/tooling/chore`


packages/clerk-js/src/ui/elements/ArrowBlockButton.tsx
```diff
@@ -85,7 +85,7 @@ export const ArrowBlockButton = (props: ArrowBlockButtonProps) => {
             <Icon
               elementDescriptor={leftIconElementDescriptor}
               elementId={leftIconElementId}
-              icon={leftIcon}
+              icon={leftIcon as React.ComponentType}
```
LekoArts, line 88: So this is a bit complicated 😆 
Due to https://github.com/clerkinc/javascript/blob/caf1f2fd1a8f7b187e152e0e095d01095b41ca98/packages/clerk-js/src/ui/elements/ArrowBlockButton.tsx#L12 this type here is wrong. But you just can't set it to the correct `React.ComponentType` because the type is referenced in other places (as a type alias).

So yeah, instead of fixing stuff in 10 places I just type-casted 🙈

packages/clerk-js/tsconfig.declarations.json
```diff
@@ -8,6 +8,7 @@
     "declarationMap": false,
     "sourceMap": false,
     "outDir": "./dist",
-    "declarationDir": "./dist/types"
+    "declarationDir": "./dist/types",
+    "noImplicitReturns": false
```
LekoArts, line 12: We only warn against this in our code (and have some instances where we need to fix this). During compilation then it fails unless turned off. Since we don't enforce it in our normal code I opted for this for now.

packages/clerk-js/src/ui/elements/ArrowBlockButton.tsx
nikosdouvlis: What do we need to fix the underlying type issue?
Can we either try to do it now or create a ticket with more details?

LekoArts: I don't know yet, since this would require some bigger investigation/changes rather than just keeping things as-is for now.

It comes down to use reusing the props here:

https://github.com/clerkinc/javascript/blob/8434782c53147e4e3333746fd6096212dfcaa51d/packages/clerk-js/src/ui/common/BlockButtons.tsx#L6

Then in other places we use it like this: https://github.com/clerkinc/javascript/blob/8434782c53147e4e3333746fd6096212dfcaa51d/packages/clerk-js/src/ui/components/SignIn/SignInAccountSwitcher.tsx#L51-L57

But the code I changed here only wants `ComponentType`:

https://github.com/clerkinc/javascript/blob/8434782c53147e4e3333746fd6096212dfcaa51d/packages/clerk-js/src/ui/primitives/Icon.tsx#L30

So it's a big problem and fixing this would block me from shipping the more important work

BRKalow: The type isn't actually wrong here, see the `isIconElement` boolean on L47 and handling the case where `leftIcon` is an element on L99. It seems that with the TS upgrade though the type narrowing isn't working as it was before 😞 :

Screenshot 2023-10-16 at 12 18 13 PM

I think @LekoArts's fix is good for now, we could also use @ts-expect-error and leave a comment stating we already have a type guard.

BRKalow (approved):
Approving! I think we can opt to use `@ts-expect-error` instead of casting for the type issue, but either approach works.

clerk-cookie:

PR #1881, opened 2023-10-14
feat(shared): Support passing env source into runtime environment helpers [SDK-773]

dimkl:
## Description

Changes:
- Support per-request env of Cloudflare workers 
- Add JSDoc with examples

Currently we cannot support per-request environment variables for helpers using the runtimeEnvironment helpers since we should pass the `env` parameter from `fetch()` handler down to all functions using it. An alternative way to support this is to  manually copy these values into `process.env` (see [Cloudflare ref](https://developers.cloudflare.com/workers/runtime-apis/nodejs/process/#relationship-to-per-request-env-argument-in-fetch-handlers)).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/shared/src/utils/runtimeEnvironment.ts
+++ b/packages/shared/src/utils/runtimeEnvironment.ts
@@ -1,30 +1,83 @@
-export const isDevelopmentEnvironment = (): boolean => {
+import { parsePublishableKey } from './keys';
+
+/**
+ * Check for development runtime environment using the environment variables of the Clerk publishableKey.
+ *
+ * Examples:
+ *
+ * 1. using process.env: `isDevelopmentEnvironment()`
+ * 2. passing custom process.env source: `isDevelopmentEnvironment(import.meta.env)`
+ * 3. using publishableKey: `isDevelopmentEnvironment('pk_********')`
+ *
+ * @param envOrPublishableKey
+ * @returns boolean
+ */
+export const isDevelopmentEnvironment = (envOrPublishableKey?: EnvironmentOrPublishableKey): boolean => {
+  return isEnvironment('development', envOrPublishableKey);
+};
+
+/**
+ * Check for testing runtime environment using the environment variables of the Clerk publishableKey.
+ *
+ * Examples:
+ *
+ * 1. using process.env: `isTestEnvironment()`
+ * 2. passing custom process.env source: `isTestEnvironment(import.meta.env)`
+ * 3. using publishableKey: `isTestEnvironment('pk_test_********')`
+ *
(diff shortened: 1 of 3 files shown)


packages/shared/src/utils/runtimeEnvironment.ts
```diff
@@ -1,30 +1,83 @@
...
+ * @returns boolean
+ */
+export const isProductionEnvironment = (envOrPublishableKey?: EnvironmentOrPublishableKey): boolean => {
+  return isEnvironment('production', envOrPublishableKey);
+};
 
-  // TODO: add support for import.meta.env.DEV that is being used by vite
-  return false;
+const getRuntimeFromProcess = (env?: typeof process.env) => {
+  return env?.NODE_ENV;
```
anagstef, line 55: ❓ @dimkl this one applies only on node environments, right?
Or, if you pass an `env` object, it should include a `NODE_ENV` property for this to work, right?

dimkl: No, it applies for all runtimes (Node, Cloudflare Workers, ...).
If you pass an `env` object with the `NODE_ENV` property, it should work.
I probably need to test if the `process.env` type works for Vite and Cloudflare workers.

anagstef: @dimkl That's what I meant. You need to have `NODE_ENV` in your passed object for this to work. `NODE_ENV` usually refers to node environments.

LekoArts: I agree that this name is a bit misleading. At the moment the function assumes that `NODE_ENV` exists and that it's Node.js

You can do a check for the runtime like this: https://github.com/unjs/std-env/blob/main/src/runtimes.ts

Probably the whole "runtime" name (also for the file name) is a bit misleading as we only check the `NODE_ENV` environment 🤔

clerk-cookie:

PR #1894, opened 2023-10-16
Include clerkTraceId for backend api errors

Nikpolik:
## Description

This PR adds `clerkTraceId` in `ClerkBackendApiResponse`. This can help debug issues with API responses especially when errors happen server side or transiently and they can't be viewed in the console. 

## Changes

1. Use `ClerkAPIResponseError` from `@clerk/shared`. 
2. Updated `ClerkBackendApiResponse` interface to include `clerkTraceId` when available.
3. Extract `clerkTraceId` either from backend error response or headers in `buildRequest`.
4. Pass the new `clerkTraceId` value to `withLegacyReturn` and then to the constructor of `ClerkAPIResponseError`.
5. Update `toString` for `ClerkAPIResponseError` so it can also log the clerkTraceId.

Now callining `.toString()` on an error should show the trace id

Screenshot 2023-10-17 at 12 45 50 PM

The same is true for simply calling console.log on the error

Screenshot 2023-10-16 at 4 49 20 PM

## Notes

DX is a bit weird here since fields that are different between success and error cannot be accessed by default in typescript and they do not appear in code completion(intelisense/language server). They can be accessed however by
checking if they exist in the object.

```ts
if ("clerkTraceId" in clerkBackendApiResponse) {
  console.log(`Error Trace: ${response.clerkTraceId}`);
}
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

### DIFF
--- a/.changeset/sour-comics-stare.md
+++ b/.changeset/sour-comics-stare.md
@@ -1,0 +1,7 @@
+---
+'@clerk/backend': patch
+'@clerk/shared': patch
+---
+
+Add clerkTraceId to ClerkBackendApiResponse and ClerkAPIResponseError to allow for better tracing and debugging API error responses. 
+Uses `clerk_trace_id` when available in a response and defaults to [`cf-ray` identifier](https://developers.cloudflare.com/fundamentals/reference/cloudflare-ray-id/) if missing.  
--- a/packages/backend/src/api/request.ts
+++ b/packages/backend/src/api/request.ts
@@ -1,3 +1,4 @@
+import { ClerkAPIResponseError } from '@clerk/shared/error';
 import type { ClerkAPIError, ClerkAPIErrorJSON } from '@clerk/types';
 import deepmerge from 'deepmerge';
 import snakecaseKeys from 'snakecase-keys';
@@ -37,6 +38,7 @@ export type ClerkBackendApiResponse<T> =
   | {
       data: null;
       errors: ClerkAPIError[];
+      clerkTraceId?: string;
     };
 
 export type RequestFunction = ReturnType<typeof buildRequest>;
@@ -52,13 +54,14 @@ const withLegacyReturn =
   (cb: any): LegacyRequestFunction =>
   async (...args) => {
     // @ts-ignore
-    const { data, errors, status, statusText } = await cb<T>(...args);
+    const { data, errors, status, statusText, clerkTraceId } = await cb<T>(...args);
     if (errors === null) {
       return data;
     } else {
       throw new ClerkAPIResponseError(statusText || '', {
         data: errors,
         status: status || '',
+        clerkTraceId,
       });
     }
   };
@@ -161,6 +164,7 @@ export function buildRequest(options: CreateBackendApiOptions) {
               message: err.message || 'Unexpected error',
             },
           ],
+          clerkTraceId: getTraceId(err, res?.headers),
         };
       }
 
@@ -171,11 +175,23 @@ export function buildRequest(options: CreateBackendApiOptions) {
         // @ts-expect-error
         status: res?.status,
         statusText: res?.statusText,
+        clerkTraceId: getTraceId(err, res?.headers),
       };
     }
   };
 
   return withLegacyReturn(request);
+}
+
+// Returns either clerk_trace_id if present in response json, otherwise defaults to CF-Ray header
+// If the request failed before receiving a response, returns undefined
+function getTraceId(data: unknown, headers?: Headers): string {
+  if (data && typeof data === 'object' && 'clerk_trace_id' in data && typeof data.clerk_trace_id === 'string') {
+    return data.clerk_trace_id;
+  }
+
+  const cfRay = headers?.get('cf-ray');
+  return cfRay || '';
 }
 
 function parseErrors(data: unknown): ClerkAPIError[] {
@@ -196,24 +212,4 @@ function parseError(error: ClerkAPIErrorJSON): ClerkAPIError {
       sessionId: error?.meta?.session_id,
     },
   };
-}
-
-class ClerkAPIResponseError extends Error {
-  clerkError: true;
-
-  status: number;
-  message: string;
-
-  errors: ClerkAPIError[];
-
-  constructor(message: string, { data, status }: { data: ClerkAPIError[]; status: number }) {
-    super(message);
-
-    Object.setPrototypeOf(this, ClerkAPIResponseError.prototype);
-
-    this.clerkError = true;
-    this.message = message;
-    this.status = status;
-    this.errors = data;
-  }
 }
--- a/packages/backend/src/api/factory.test.ts
+++ b/packages/backend/src/api/factory.test.ts
@@ -4,7 +4,7 @@ import sinon from 'sinon';
 import emailJson from '../fixtures/responses/email.json';
 import userJson from '../fixtures/responses/user.json';
 import runtime from '../runtime';
-import { jsonNotOk, jsonOk } from '../util/mockFetch';
+import { jsonError, jsonNotOk, jsonOk } from '../util/mockFetch';
 import { createBackendApiClient } from './factory';
 
 export default (QUnit: QUnit) => {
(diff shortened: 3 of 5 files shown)


packages/backend/src/api/request.ts
```diff
@@ -171,13 +175,28 @@ export function buildRequest(options: CreateBackendApiOptions) {
...
     }
   };
 
   return withLegacyReturn(request);
 }
 
+// Returns either clerk_trace_id if present in response json, otherwise defaults to CF-Ray header
+function getTraceId(data: unknown, headers?: Headers): string | undefined {
+  let traceId: unknown = headers?.get('cf-ray');
+  if (!!data && typeof data === 'object' && 'clerk_trace_id' in data) {
```
SokratisVidros, line 189:
```suggestion
  if (data && data.clerk_trace_id) {
    return data.clerk_trace_id;
  }
  return '';
```

packages/backend/src/api/request.ts
```diff
@@ -1,3 +1,4 @@
+import { ClerkAPIResponseError } from '@clerk/shared';
```
SokratisVidros, line 1: @nikosdouvlis Do we foresee any building issues with this import from shared? Should we wait for the @clerk/shared architecture as part of #1898 ?

SokratisVidros (changes requested):

packages/backend/src/api/request.ts
dimkl: @Nikpolik could you wait until the https://github.com/clerkinc/javascript/pull/1898 is merged and use the subpath import? Not using the subpath import would cause the backend package to fail in node runtimes that do not have react as peer dependency and bundling tools like vite.

packages/backend/src/api/request.ts
georgepsarakis: ❓ If we want to make it clearer that `clerk_trace_id` from error (non-200) responses takes precedence, you can consider moving line 188 to become the fallback (`cf-ray` is always populated in a response).

.changeset/sour-comics-stare.md
```diff
@@ -0,0 +1,7 @@
+---
+'@clerk/backend': patch
+'@clerk/shared': patch
+---
+
+Add clerkTraceId to ClerkBackendApiResponse and ClerkAPIResponseError to allow for better tracing and debugging api errors. 
+Uses clerk trace id when available and defaults to cf ray id if missing.
```
georgepsarakis, line 7:
```suggestion
Uses `clerk_trace_id` when available in a response and defaults to [`cf-ray` identifier](https://developers.cloudflare.com/fundamentals/reference/cloudflare-ray-id/) if missing. 
```

.changeset/sour-comics-stare.md
```diff
@@ -0,0 +1,7 @@
+---
+'@clerk/backend': patch
+'@clerk/shared': patch
+---
+
+Add clerkTraceId to ClerkBackendApiResponse and ClerkAPIResponseError to allow for better tracing and debugging api errors.
```
georgepsarakis, line 6:
```suggestion
Add clerkTraceId to ClerkBackendApiResponse and ClerkAPIResponseError to allow for better tracing and debugging API error responses. 
```

packages/backend/src/api/request.ts
```diff
@@ -197,23 +216,3 @@ function parseError(error: ClerkAPIErrorJSON): ClerkAPIError {
     },
   };
 }
-
-class ClerkAPIResponseError extends Error {
```
georgepsarakis, line 201: ❓ This was unused I guess?

packages/shared/src/errors/Error.ts
```diff
@@ -70,24 +71,32 @@ export class ClerkAPIResponseError extends Error {
...
 
   public toString = () => {
-    return `[${this.name}]\nMessage:${this.message}\nStatus:${this.status}\nSerialized errors: ${this.errors.map(e =>
-      JSON.stringify(e),
+    let message = `[${this.name}]\nMessage:${this.message}\nStatus:${this.status}\nSerialized errors: ${this.errors.map(
+      e => JSON.stringify(e),
     )}`;
+
+    if (this.clerkTraceId) {
+      message += `\nClerk trace id: ${this.clerkTraceId}`;
```
georgepsarakis, line 96:
```suggestion
      message += `\nClerk Trace ID: ${this.clerkTraceId}`;
```

packages/backend/src/api/request.ts
```diff
@@ -171,13 +175,28 @@ export function buildRequest(options: CreateBackendApiOptions) {
...
+        clerkTraceId: getTraceId(err, res?.headers),
       };
     }
   };
 
   return withLegacyReturn(request);
 }
 
+// Returns either clerk_trace_id if present in response json, otherwise defaults to CF-Ray header
+function getTraceId(data: unknown, headers?: Headers): string | undefined {
```
georgepsarakis, line 187: ❓ Is it worth adding some unit tests? The implementation is simple, but tests can also serve as examples!

georgepsarakis (commented):
Nice work @Nikpolik 👏 !

packages/backend/src/api/request.ts
Nikpolik: @SokratisVidros I think we must do the check using the `in` keyword. Otherwise it will not pick it up. 

That is because it is being narrowed from `unknown` -> `object` -> `object with clerk_trace_id in keys`

Screenshot 2023-10-18 at 1 43 29 PM

packages/backend/src/api/request.ts
Nikpolik: Yes it is being imported from `@clerk/shared` now!

packages/backend/src/api/request.ts
Nikpolik: 👍

packages/backend/src/api/request.ts
Nikpolik: @georgepsarakis there is a case where we may not even have `cf-ray`. 

If fetch throws an error before making a request, the response will be undefined. This can happen for example if fetch fails due to network issues.

packages/backend/src/api/request.ts
dimkl: @Nikpolik you can proceed with this PR

.changeset/sour-comics-stare.md
Nikpolik: Updated the changelog!

packages/backend/src/api/request.ts
```diff
@@ -171,13 +175,29 @@ export function buildRequest(options: CreateBackendApiOptions) {
...
+  if (data && typeof data === 'object' && 'clerk_trace_id' in data && typeof data.clerk_trace_id === 'string') {
+    return data.clerk_trace_id;
+  }
+
+  const cfRay = headers?.get('cf-ray');
+  if (typeof cfRay === 'string') {
+    return cfRay;
+  }
+
+  return undefined;
```
SokratisVidros, line 198:
```suggestion
  return "";
```

packages/backend/src/api/request.ts
SokratisVidros:
```suggestion
function getTraceId(data: unknown, headers?: Headers): string {
```

SokratisVidros (approved):

dimkl:
@Nikpolik Could you also cherry-pick this change into `main`?

PR #1896, opened 2023-10-16
feat(clerk-js): Use `org:sys_domains:read` for improved fine grain control of the UI

panteliselef:
## Description
This PR makes usage of `org:sys_domains:read` a permission that allows users to see a list of domains. Previously we were wrapping the whole functionality under the `manage` permission.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`


packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
```diff
@@ -21,6 +21,55 @@ type DomainListProps = GetDomainsParams & {
   fallback?: React.ReactNode;
 };
 
+const useDomainList = () => {
+  const { isAuthorizedUser: canDeleteDomain } = useGate({ permission: 'org:sys_domains:delete' });
```
SokratisVidros, line 25: ❓ What else does the `useGate` return? I'd expect to the DX to be like `const canDeleteDomain = useGate({ permission: 'org:sys_domains:delete' });` for the internal and external gate hook.

packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
```diff
@@ -21,6 +21,55 @@ type DomainListProps = GetDomainsParams & {
...
+}: Pick<DomainListProps, 'redirectSubPath'> & {
+  domainId: OrganizationDomainResource['id'];
+}) => {
+  const { navigate } = useRouter();
+  const { canDeleteDomain, canVerifyDomain } = useDomainList();
+
+  return (
+    <ThreeDotsMenu
+      actions={[
+        ...(canVerifyDomain
```
SokratisVidros, line 47: 🙃 Consider building the actions array outside JSX for improved readability using a `buildActions` function that handles the business logic.

packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
```diff
@@ -21,6 +21,55 @@ type DomainListProps = GetDomainsParams & {
...
+          ? [
+              {
+                label: localizationKeys(
+                  'organizationProfile.profilePage.domainSection.unverifiedDomain_menuAction__verify',
+                ),
+                onClick: () => navigate(`${redirectSubPath}${domainId}/verify`),
```
SokratisVidros, line 53: 🔧 Prefer the `buildUrl` or `buildPath` utilities that ensure the path is formed correctly without accidental double slashes. 

String concatenations in URLs should be avoided.

packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
```diff
@@ -82,23 +132,12 @@ export const DomainList = withGate(
...
-                      {
-                        label: localizationKeys(
-                          'organizationProfile.profilePage.domainSection.unverifiedDomain_menuAction__remove',
-                        ),
-                        isDestructive: true,
-                        onClick: () => navigate(`${redirectSubPath}${d.id}/remove`),
-                      },
-                    ]}
-                  />
+                  showDotMenu ? (
```
SokratisVidros, line 135: 🙃 `showDotMeny && <JSX/>`

packages/clerk-js/src/ui/hooks/useFetch.ts
```diff
@@ -35,7 +35,7 @@ export const useFetch = <T>(
         requestStatus.setError();
         setData(null);
       });
-  }, []);
+  }, [JSON.stringify(params)]);
```
SokratisVidros, line 38: ❓ Can you elaborate on this change? Is it necessary to be part of this PR?

packages/clerk-js/src/ui/utils/test/mockHelpers.ts
```diff
@@ -30,6 +30,46 @@ export const mockClerkMethods = (clerk: LoadedClerk): DeepJestMocked<LoadedClerk
   mockMethodsOf(clerk.client.signUp);
   clerk.client.sessions.forEach(session => {
     mockMethodsOf(session);
+    session.isAuthorized = jest.fn(args => {
```
SokratisVidros, line 33: ❓ Why do we need to mock this helper? Should we use the actual method and test it?

SokratisVidros (changes requested):

packages/clerk-js/src/ui/utils/test/mockHelpers.ts
panteliselef: The line above assigns jest.fn() to all methods of the object, i basically restore it. 

Tests already exists in Session.test.tsx, but re-implementing the function here was necessary for me to test the usage inside the components, as they are depending on it now.

panteliselef: Do u think we should change `mockMethodsOf` to accept a second argument that will skip assigning jest.fn to specified properties ?

packages/clerk-js/src/ui/hooks/useFetch.ts
panteliselef: Previously useFetch would only fire once, "on mount", there was a bug where the Gate component would not hide the content while not having the necessary permission because the "fetch" wouldn't fire when the parameters changed.

For context we are using `useFetch` to handle the async logic of isAuthorized.

packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
panteliselef: Already tried that, seems like `trailingComponent` does not accept null, I didn't want to mess with the signature of the component.

packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
panteliselef: At this stage of implementation this request is reasonable. We decides on this signature because in the future useGate may actually perform a request, which will require more properties to be returned. For example if in the future we depend on CF workers for authorization the following signature makes more sense
```tsx
const { isAuthorizedUser, isLoading, isError } = useGate({ permission: 'org:sys_domains:delete' });
```

packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
panteliselef: I created a hook that returns the actions, the conditional logic depended too much on the state + router.

packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
panteliselef: I went with a custom solution, seems like buildUrl and the hashPath did not work with path routing for internal navigations

packages/clerk-js/src/utils/url.ts
```diff
@@ -210,6 +210,18 @@ export const trimTrailingSlash = (path: string): string => {
...
 
+/**
+ * trimLeadingSlash(path: string): string
+ *
+ * Strips the leading slashes from a string
+ *
+ * @returns {string} Returns the string without leading slashes
+ * @param path
+ */
+export const trimLeadingSlash = (path: string): string => {
```
desiprisg, line 221: Can we add a couple of tests for this util?

desiprisg (approved):

packages/clerk-js/src/utils/url.ts
panteliselef: Yeap, thanks for the reminder

SokratisVidros (changes requested):

packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
```diff
@@ -17,10 +18,63 @@ type DomainListProps = GetDomainsParams & {
...
+
+const buildDomainListRelativeURL = (parentPath: string, domainId: string, mode?: 'verify' | 'remove') =>
+  trimLeadingSlash(stripOrigin(toURL(`${parentPath}/${domainId}/${mode || ''}`)));
+
+const useMenuActions = (parentPath: string, domainId: string) => {
+  const { canDeleteDomain, canVerifyDomain } = useDomainList();
+  const { navigate } = useRouter();
+
+  return [
+    ...(canVerifyDomain
```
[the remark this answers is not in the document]
panteliselef: I mostly did that for type inference to work out of the box, but i will use explicit types instead.

packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
panteliselef: The following has type of ` false | EmotionJSX.Element | null`

```tsx
showDotMenu && (
                    <DomainListDotMenu
                      redirectSubPath={redirectSubPath}
                      domainId={d.id}
                    />
                  )


```

So i would have to do 
```tsx

(showDotMenu  || null ) && (
                    <DomainListDotMenu
                      redirectSubPath={redirectSubPath}
                      domainId={d.id}
                    />
                  )
```

Which i'm not convinced it is better

packages/clerk-js/src/ui/hooks/useFetch.ts
panteliselef: This would only invoke a call if params change, not on every render.

packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
SokratisVidros: I am not sure I understand what you are talking about. Is this [the](https://github.com/clerkinc/javascript/blob/b2296d6304e1ca31a35450e0c67a12555c0142f9/packages/clerk-js/src/ui/elements/BlockWithTrailingComponent.tsx#L10) type?

SokratisVidros (approved):

packages/clerk-js/src/ui/components/OrganizationProfile/DomainList.tsx
panteliselef: This is the TS error 
image

PR #1906, opened 2023-10-17
chore(repo): Extend linting

tmilewski:
## Description

Our current ESLint configuration was global which left a lot to be desired as each of our packages have unique requirements.

This PR extends leans on a generic set of ESLint configurations while also allowing for a lot of customizability at both the ESLint "feature"-level as well as the package-level.

**Please focus on this commit as these are the changes:** https://github.com/clerkinc/javascript/pull/1906/commits/f2c3b647559d8ec165c78470505a2067b340177d 
_The other changes are to appease Prettier._

#### Please Note: 
In an effort to keep this PR as minimal as possible, I've disabled a number of rules. The plan is to re-enable them via follow-on PRs.)

#### Other Notes:
- The `/integration` directory has been factored into NPM workspaces and, ultimately, Turborepo. This doesn't have any tangible change other than running linting along with the packages.
- ESLint (when utilizing Typescript) will target `tsconfig.lint.json`. If that does not exists, it moves on to `tsconfig.json`. The `lint` TSConfig was required in a few cases as the default one wasn't permissive enough with it's `includes`/`excludes`.
- I had to skip the Husky tasks so that it didn't attempt to update and commit all of the "fixable" issues.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [x] `build/tooling/chore`


LekoArts:
How much of an effort would it be to to _not_ use `tsconfig.lint.json` across the board? Just as an estimated guess

tmilewski:
> How much of an effort would it be to to _not_ use `tsconfig.lint.json` across the board and instead rely on `tsconfig.json` (and fixing those)? Just as an estimated guess

@LekoArts Not much of an effort, but it would make the `tsconfig.json` quite all-encompassing. I just didn't want to cause any unnecessary side-effects.

packages/types/src/json.ts
```diff
@@ -380,15 +380,6 @@ export interface OrganizationSuggestionJSON extends ClerkResourceJSON {
   updated_at: number;
 }
 
-export interface OrganizationSuggestionJSON extends ClerkResourceJSON {
```
tmilewski, line 383: Note: This is an exact copy of the interface above it which ESLint didn't catch until now.

packages/clerk-js/src/utils/devBrowser.ts
```diff
@@ -51,7 +51,7 @@ export function getDevBrowserJWTFromURL(url: URL): string {
 
   const jwt = jwtFromHash || jwtFromSearch;
 
-  if (jwt && typeof globalThis.history !== undefined) {
+  if (jwt && typeof globalThis.history !== 'undefined') {
```
desiprisg, line 54: Shouldn't we check for `typeof globalThis` first?

.husky/pre-commit
```diff
@@ -1,4 +1,8 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
+# ESLint via Turborepo
+# npm run lint -- -- --fix
```
LekoArts, line 5: I think we want to enable this before the merge?

tmilewski: @LekoArts We'll likely want to do this in a follow-on PR as it's going to want to update a number of files.

LekoArts:
Is the addition of `integration` into the workspaces necessary? It'll mean that you have to install more dependencies for the monorepo even though you might not be interested in running the tests + you could more easily run into problems of "multiple different versions of packages".

In CI you could still cache the `node_modules` on an individual level for each test suite.

tmilewski:
> Is the addition of `integration` into the workspaces necessary? It'll mean that you have to install more dependencies for the monorepo even though you might not be interested in running the tests + you could more easily run into problems of "multiple different versions of packages".
> 
> In CI you could still cache the `node_modules` on an individual level for each test suite.

There's actually no noticeable difference. I should note that `integrations` is included, not `integations/*` so it's not attempting to install each integration test's deps.

LekoArts:
The integration folder doesn't have a `package.json` file (I see it in another commit but not in the total view of this PR). What you're saying makes sense then, but isn't that package.json then missing?

tmilewski:
> The integration folder doesn't have a `package.json` file (I see it in another commit but not in the total view of this PR). What you're saying makes sense then, but isn't that package.json then missing?

Ah, you are correct, it's a remnant of a past commit. Updated, thanks!

LekoArts (approved):
:shipit:

PR #1911, opened 2023-10-18
fix(clerk-js): Use the issued at timestamp from JWT for the MemoryTokenCache

octoper:
## Description
This PR changes the `createdAt` value when storing a token to the `MemoryTokenCache` from the current user time to the issued at time of the token to check how much time has elapsed from when the token has been issued.

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/clerk-js/src/core/tokenCache.test.ts
+++ b/packages/clerk-js/src/core/tokenCache.test.ts
@@ -12,8 +12,25 @@ jest.mock('./resources/Base', () => {
   };
 });
 
-const jwt =
-  'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2NzU4NzY3OTAsImRhdGEiOiJmb29iYXIiLCJpYXQiOjE2NzU4NzY3MzB9.Z1BC47lImYvaAtluJlY-kBo0qOoAk42Xb-gNrB2SxJg';
+const mockJwtToken = (now: number): string => {
+  const nowInSeconds = Math.floor(now / 1000);
+  const payload = {
+    iat: nowInSeconds,
+    exp: nowInSeconds + 60,
+    data: 'foobar',
+  };
+
+  const header = {
+    alg: 'HS256',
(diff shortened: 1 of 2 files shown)


dimkl (changes requested):
❓ What issue this change will resolve?
Also i think that using the `issuedAt` instead of the end-user clock will not work for end-user with a clockSkew. The current implementation keeps the creation DateTime of the token based on the end-user clock and the expiration duration from the token using `expiresAt - issuedAt` to calculate the exact time the token will expire based on the end-user clock.
Example:
End-user A with clockSkew 10 minutes (compared to the FAPI server), retrieves token with issuedAt at `00:10:00` that will expire in 60 seconds `00:11:00` and the end-user clock is `00:00:00`. After 5 seconds Frontend code executes `getToken()` and `tokenCache.get()` is executed. The code in packages/clerk-js/src/core/tokenCache.ts will be executed in the end-user with :
```typescript
    const nowSeconds = Math.floor(Date.now() / 1000);  // "00:00:05"
    const elapsedSeconds = nowSeconds - value.createdAt!; // "00:00:05" - "00:10:00"
     const expiresSoon = value.expiresIn! - elapsedSeconds < (leeway || 1) + SYNC_LEEWAY; // 60 - 595 < 10 - 5   
```

clerk-cookie:

PR #1957, opened 2023-10-30
feat(clerk-js,types,nextjs,remix): Terse path parameters

octoper:
## Description

This PR introduces some breaking changes

- By default, all components that use routing will have their `routing` prop assigned with `'path'` if the `path` prop is filled. Additionally, the types will be stricter, and if the `path` prop is filled and the `routing` prop is filled with a value other than `'path'`, TypeScript will throw an error.
- The `<UserButton />` component will have the `userProfileMode` prop assigned with `'navigation'` by default if the `userProfileUrl` prop is filled. Similarly, the types will be stricter, and if the `userProfileUrl` prop is filled and the `userProfileMode` prop is filled with a value other than `'navigation'`, TypeScript will throw an error.
- The `<OrganizationSwitcher />` component will have the `organizationProfileMode` and `createOrganizationMode` props assigned with `'navigation'` by default if the `organizationProfileUrl` and `createOrganizationUrl` props are filled accordingly. Additionally, the types will be stricter, and if the `organizationProfileUrl` or `createOrganizationUrl` prop is filled and the `organizationProfileMode` or `createOrganizationMode` prop is filled with a value other than `'navigation'`, TypeScript will throw an error.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [X] 🔨 Breaking change
- [ ] other:

## Packages affected

- [X] `@clerk/clerk-js`
- [x] `@clerk/clerk-react`
- [X] `@clerk/nextjs`
- [X] `@clerk/remix`
- [X] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`


packages/clerk-js/src/ui/components/CreateOrganization/CreateOrganization.tsx
```diff
@@ -38,7 +38,7 @@ export const CreateOrganization = withOrganizationsEnabledGuard(
   { mode: 'redirect' },
 );
 
-export const CreateOrganizationModal = (props: CreateOrganizationProps): JSX.Element => {
+export const CreateOrganizationModal = (props: Omit<CreateOrganizationProps, 'path'>): JSX.Element => {
```
desiprisg, line 41: Shouldn't we also omit the `routing` prop here? (and for the other modals as well)

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
```diff
@@ -241,6 +241,9 @@ export const useUserButtonContext = () => {
   const afterSwitchSessionUrl = ctx.afterSwitchSessionUrl || displayConfig.afterSwitchSessionUrl;
   const navigateAfterSwitchSession = () => navigate(afterSwitchSessionUrl);
 
+  const userProfileMode =
+    !!ctx.userProfileUrl && ctx.userProfileMode === undefined ? 'navigation' : ctx.userProfileMode;
```
desiprisg, line 245: Can we just do `!ctx.userProfileMode` here or do we care for `null` ?

packages/clerk-js/src/ui/portal/index.tsx
```diff
@@ -11,28 +13,29 @@ type PortalProps<CtxType extends AvailableComponentCtx, PropsType = Omit<CtxType
   node: HTMLDivElement;
   component: React.FunctionComponent<PropsType> | React.ComponentClass<PropsType, any>;
   // Aligning this with props attributes of ComponentControls
-  props?: PropsType & { path?: string; routing?: string };
+  props: PropsType & RoutingOptions;
```
desiprisg, line 16: Why did we decide to add this to the `portal` level?

packages/types/src/utils.ts
```diff
@@ -83,3 +83,10 @@ type IsSerializable<T> = T extends Function ? false : true;
 export type Serializable<T> = {
   [K in keyof T as IsSerializable<T[K]> extends true ? K : never]: T[K];
 };
+
+/**
+ * Omit without union flattening
+ * */
+export type Without<T, W> = {
```
desiprisg, line 90: I remember adding this in my old PR for terse params. Was it actually necessary here?

desiprisg (changes requested):
Good work! We should add some type tests to make sure this doesn't break in the future since these are some complex types. Check this file for referennce: https://github.com/clerkinc/javascript/blob/3ceb2a734a43f134956164377399fec46e01e0a1/packages/react/src/contexts/__tests__/ClerkProvider.test.tsx

packages/clerk-js/src/ui/components/CreateOrganization/CreateOrganization.tsx
octoper: Yeah you are tight we can also omit routing here as it's always `virtual`.

packages/types/src/utils.ts
octoper: Yeah it's an alternative to the built-in `Omit` that will omit the properties passed without doing union flattening, because we are having unions and they break if we use `Omit`.

packages/clerk-js/src/ui/portal/index.tsx
octoper: It's the same behaviour but I made the types stricter everywhere.

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
octoper: Yeah you are right we can change that to be `!ctx.userProfileMode`, I also don't think that this can be `null`

packages/clerk-js/src/ui/common/__tests__/authPropHelpers.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    it('returns routing: null if path was provided and routing was null (avoid breaking integrations)', () => {
+      // @ts-expect-error Need to test this case
+      expect(normalizeRoutingOptions({ path: 'test', routing: null })).toEqual({ routing: null, path: 'test' });
+    });
+
+    it('returns the same options if routing was provided (any routing) and path was provided (avoid breaking integrations)', () => {
+      expect(normalizeRoutingOptions({ routing: 'path', path: 'test' })).toEqual({ routing: 'path', path: 'test' });
+      expect(normalizeRoutingOptions({ routing: 'hash', path: 'test' })).toEqual({ routing: 'hash', path: 'test' });
+      expect(normalizeRoutingOptions({ routing: 'virtual' })).toEqual({ routing: 'virtual' });
```
LekoArts, line 17: What happens in this case?

```ts
normalizeRoutingOptions({ routing: 'virtual', path: 'test' })
```

packages/clerk-js/src/utils/authPropHelpers.ts
```diff
@@ -112,3 +112,13 @@ export const buildAuthQueryString = (data: BuildAuthQueryStringArgs): string | n
...
 };
+
+export const normalizeRoutingOptions = ({
+  routing,
+  path,
+}: {
+  routing?: RoutingStrategy;
+  path?: string;
+}): RoutingOptions => {
+  return { routing: !!path && !routing ? 'path' : routing, path } as RoutingOptions;
```
LekoArts, line 123: The `as` assertion here is a code-smell, it shouldn't be necessary. What happens when you remove it?

packages/clerk-js/src/ui/common/__tests__/authPropHelpers.test.ts
octoper: it will return `{ routing: 'virtual', path: 'test' }` just as in this case `normalizeRoutingOptions({ routing: 'hash', path: 'test' })`, I will add a test case for it.

packages/clerk-js/src/utils/authPropHelpers.ts
octoper: It's necessary as the the type is a union and in the case `virtual` or `hash` the objects does not have the `path` property.

LekoArts: Is the TS compiler happy if you don't use ternaries but a bit more verbose syntax?

```ts
if (!!path && !routing) {
return { routing: 'path', path }
} else {
  return { routing }
}
```

nikosdouvlis (changes requested):
@octoper You will need to apply the changes to the ui-retheme dir.  I will share more details within the day

packages/clerk-js/src/utils/authPropHelpers.ts
octoper: Yeah I've tried that but still it's not happy

packages/remix/src/client/uiComponents.tsx
```diff
@@ -20,12 +22,14 @@ export const SignIn = (props: SignInProps) => {
...
   const { signUpUrl } = useClerkRemixOptions();
+  const { routing: _routing, path: _path, ...restProps } = props;
+
   if (signUpUrl) {
     return (
       <BaseSignUp
         routing='path'
         path={signUpUrl}
-        {...props}
+        {...restProps}
```
dimkl, line 32: ❓ Why the current implementation overriding the `routing = 'path'` if `signUpUrl` was used and `routing` prop was passed to `SignUp` component?
Could we add a test for that we haven't already?

packages/remix/src/client/uiComponents.tsx
```diff
@@ -6,12 +6,14 @@ import { useClerkRemixOptions } from './RemixOptionsContext';
...
   const { signInUrl } = useClerkRemixOptions();
+  const { routing: _routing, path: _path, ...restProps } = props;
+
   if (signInUrl) {
     return (
       <BaseSignIn
         routing='path'
         path={signInUrl}
-        {...props}
+        {...restProps}
```
dimkl, line 16: same but for SignIn : https://github.com/clerk/javascript/pull/1957/files#r1389065093

packages/nextjs/src/client-boundary/uiComponents.tsx
```diff
@@ -35,12 +37,14 @@ export const SignIn = (props: SignInProps) => {
 
 export const SignUp = (props: SignUpProps) => {
   const { signUpUrl: repoLevelSignUpUrl } = useClerkNextOptions();
+  const { routing, path, ...restProps } = props;
```
dimkl, line 40: 🙃 `const { routing: _routing, path: _path, ...restProps } = props;`  (probably there is an unused variable eslint warning here that will be resolved with this).

packages/clerk-js/src/ui/components/CreateOrganization/CreateOrganization.tsx
```diff
@@ -38,7 +38,7 @@ export const CreateOrganization = withOrganizationsEnabledGuard(
   { mode: 'redirect' },
 );
 
-export const CreateOrganizationModal = (props: CreateOrganizationProps): JSX.Element => {
+export const CreateOrganizationModal = (props: Omit<CreateOrganizationProps, 'path' | 'routing'>): JSX.Element => {
```
dimkl, line 41: ❓ Should we create a `WithoutRouting<T> =  Omit<T, 'path' | 'routing'>` helper type to re-use it in all those cases?

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
```diff
@@ -241,6 +241,8 @@ export const useUserButtonContext = () => {
   const afterSwitchSessionUrl = ctx.afterSwitchSessionUrl || displayConfig.afterSwitchSessionUrl;
   const navigateAfterSwitchSession = () => navigate(afterSwitchSessionUrl);
 
+  const userProfileMode = !!ctx.userProfileUrl && !ctx.userProfileMode ? 'navigation' : ctx.userProfileMode;
```
dimkl, line 244: ❓ why did we change this and the next ones?

dimkl (approved):
Nice! 🚀

packages/remix/src/client/uiComponents.tsx
octoper: Yeah you are right the current implementation is overriding `routing` or `path` if the props are passed to the `SignUp` component, I've deployed a fix for that

packages/remix/src/client/uiComponents.tsx
octoper: Fixed, you can check my comment here https://github.com/clerk/javascript/pull/1957#discussion_r1389688018

packages/clerk-js/src/ui/components/CreateOrganization/CreateOrganization.tsx
octoper: Added!

packages/nextjs/src/client-boundary/uiComponents.tsx
octoper: Variables starting with `_` will be ignored from the `no-unused-vars` rule in ESLint

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
octoper: This is a necessary change to determine the `userProfileMode`, if the `ctx.userProfileUrl` is filled and no `ctx.userProfileMode` was provided the `userProfileMode` will be `navigation`

packages/clerk-js/src/ui/common/__tests__/authPropHelpers.test.ts
```diff
@@ -0,0 +1,15 @@
+import { normalizeRoutingOptions } from '../../../utils/authPropHelpers';
+
+describe('auth prop helpers', () => {
+  describe('normalizeRoutingOptions', () => {
+    it("returns routing: 'path' if path was provided and no routing was provided", () => {
+      expect(normalizeRoutingOptions({ path: 'test' })).toEqual({ routing: 'path', path: 'test' });
+    });
+
+    it('returns the same options if routing was provided (any routing) and path was provided (avoid breaking integrations)', () => {
```
dimkl, line 9: ❓ i thought we wanted to introduce this breaking change and force everyone to use the correct options.
cc: @nikosdouvlis

dimkl (approved):

packages/types/src/utils.ts
nikosdouvlis: @dimkl This might also solve what we were discussing yesterday

packages/clerk-js/src/ui.retheme/components/CreateOrganization/CreateOrganization.tsx
```diff
@@ -38,7 +38,7 @@ export const CreateOrganization = withOrganizationsEnabledGuard(
   { mode: 'redirect' },
 );
 
-export const CreateOrganizationModal = (props: CreateOrganizationProps): JSX.Element => {
+export const CreateOrganizationModal = (props: WithoutRouting<CreateOrganizationProps>): JSX.Element => {
```
nikosdouvlis, line 41: Good catch.
🙃 
Do you think that creating a type for `CreateOrganizationModal` explicitly and hide `WithoutRouting` in its definition would make this slightly more readable? I don't think there's a good enough reason to have types for the base components but not their modal versions but I'd like to hear your thoughts

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
```diff
@@ -310,11 +313,17 @@ export const useOrganizationSwitcherContext = () => {
 
   const navigateAfterSelectPersonal = (user: UserResource) => navigateAfterSelectOrganizationOrPersonal({ user });
 
+  const organizationProfileMode =
+    !!ctx.organizationProfileUrl && !ctx.organizationProfileMode ? 'navigation' : ctx.organizationProfileMode;
+
+  const createOrganizationMode =
+    !!ctx.createOrganizationUrl && !ctx.createOrganizationMode ? 'navigation' : ctx.createOrganizationMode;
+
```
nikosdouvlis, line 321: Quick question, do we have similar helpers for the other components? If no, what is the reason there are needed here?

packages/clerk-js/src/ui/portal/index.tsx
nikosdouvlis: Pretty sure it's not a public type.

packages/clerk-js/src/utils/authPropHelpers.ts
nikosdouvlis, line 123: Let's make this more verbose. 
In general, one-liners are fine, but in this case, expanding the ternary makes sense so we can have one block for each possible scenario, even if it's not completely necessary.

packages/nextjs/src/client-boundary/uiComponents.tsx
```diff
@@ -7,42 +7,47 @@ import React from 'react';
...
+  SignOutButton,
+  SignUpButton,
+  UserButton,
+  UserProfile,
 } from '@clerk/clerk-react';
 
 export const SignIn = (props: SignInProps) => {
   const { signInUrl: repoLevelSignInUrl } = useClerkNextOptions();
-  if (repoLevelSignInUrl) {
+  const { routing: _routing, path, ...restProps } = props;
```
nikosdouvlis, line 24:
```suggestion
  const path = props.path || repoLevelSignInUrl;
  
  if (path) {
    <BaseSignIn
       {...props}
       routing='path'
       path={path}
      />
        }
  }
```

packages/nextjs/src/client-boundary/uiComponents.tsx
```diff
@@ -7,42 +7,47 @@ import React from 'react';
...
-  if (repoLevelSignUpUrl) {
+  const { routing: _routing, path, ...restProps } = props;
+
+  if (repoLevelSignUpUrl || path) {
     return (
       <BaseSignUp
         routing='path'
-        path={repoLevelSignUpUrl}
-        {...props}
+        path={path || repoLevelSignUpUrl}
```
nikosdouvlis, line 47: Ditto

nikosdouvlis (approved):

packages/clerk-js/src/ui.retheme/components/CreateOrganization/CreateOrganization.tsx
octoper: Yeah indeed that will be more readable!

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
octoper: Yeah it's similar to [normalizeRoutingOptions](https://github.com/clerk/javascript/pull/1957/files/c4a994eb78c67d5fbae58badc3d03372cad3b42e#diff-f7fa0cdb70cc4f092d8ae5107fa3aacbc3bf2917ad56442ab5dd98169705d20eR116), this could be extracted as a helper too.

packages/clerk-js/src/utils/authPropHelpers.ts
octoper: Done!

packages/nextjs/src/client-boundary/uiComponents.tsx
octoper: Done!

packages/types/src/utils.ts
dimkl: it might. i will try that out :)

integration/tests/navigation.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  test('user profile with path router', async ({ page, context }) => {
+    const u = createTestUtils({ app, page, context });
+    await u.po.signIn.goTo();
+    await u.po.signIn.waitForMounted();
+    await u.po.signIn.signInWithEmailAndInstantPassword({ email: fakeUser.email, password: fakeUser.password });
+    await u.po.expect.toBeSignedIn();
+
+    await u.page.goToRelative('/user');
+    await u.page.waitForClerkComponentMounted();
```
nikosdouvlis, line 46:
```suggestion
    await u.page.userProfile.goTo();
    await u.page.userProfile.waitForMounted();
```

You'd probably need to create these helpers - look at the `signIn.goTo` implementation for examples

integration/tests/navigation.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    await u.page.getByRole('button', { name: /Set username/i }).click();
+
+    await u.page.waitForURL('**/user/username');
```
nikosdouvlis, line 50: Ditto - let's create a nice PO for userProfile

integration/tests/navigation.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    fakeUser = u.services.users.createFakeUser();
+    await u.services.users.createBapiUser(fakeUser);
+  });
+
+  test.afterAll(async () => {
+    await fakeUser.deleteIfExists();
+    await app.teardown();
+  });
+
+  test('user profile with hash router', async ({ page, context }) => {
```
nikosdouvlis, line 23: Shall we add a couple of tests for the path-based router as well?

nikosdouvlis (commented):
I noticed that you updated the next-app-router and the react-vite apps and added the `/hash` pages - is there any reason you did not update the rest?

In general, every app in the template dir must have the same interface so we can use the apps interchangeably throughout the tests. 

If you want to run a small navigation-focused suite without affecting any templates, you can use an existing `applicationConfig`  and the `addFile` method to programmatically create an app before the tests. The scope of the app will be just that specific test

integration/tests/navigation.test.ts
octoper: Added!

packages/clerk-js/src/utils/authPropHelpers.ts
```diff
@@ -112,3 +112,17 @@ export const buildAuthQueryString = (data: BuildAuthQueryStringArgs): string | n
   }
   return Object.keys(query).length === 0 ? null : qs.stringify(query);
 };
+
+export const normalizeRoutingOptions = ({
```
SokratisVidros, line 116: 🙃 This file is not just about `authPropHelpers` anymore. It is a collection of prop/options parsers and normalizers. So, at some point, we need to rename it.

SokratisVidros (approved):

packages/clerk-js/src/utils/authPropHelpers.ts
octoper: Noted!

integration/tests/navigation.test.ts
octoper: I have added some more cases!

integration/tests/navigation.test.ts
octoper: Done!

nikosdouvlis (approved):

PR #1970, opened 2023-10-30
chore(repo): Enter changesets Pre mode (for v5)

dimkl:
## Description

- Merge `Version packages` to produce next major version `@alpha-v5` release
- Rename references of `main-v4` to `release/v4` (are only used as reference for the future, those will actually be used to trigger flows when they are committed to `release/v4`)
- Merging to main will not publish any version

To verify `premajor` tags:
- `git checkout fix-version-packages`
- `npx changeset version`
- check Changelog and package.json changes

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [x] `build/tooling/chore`


.github/workflows/release-alpha-v5.yml
LekoArts, line 1: The file name won't be the same for the next major. To keep things consistent, let's name the file, job and action name "Premajor"

.github/workflows/release-alpha-v5.yml
```diff
@@ -25,12 +25,12 @@ jobs:
       - name: Setup
         uses: ./.github/actions/setup
 
-      - name: Version packages for canary
+      - name: Version packages for alpha-v5
         id: version-packages
-        run: npm run version:canary | tail -1 >> "$GITHUB_OUTPUT"
+        run: npm run version:alpha-v5 | tail -1 >> "$GITHUB_OUTPUT"
```
LekoArts, line 30: Let's name the package scripts also `premajor` (the `alpha-v5` is an implementation detail and will change for v6, thus requiring us to change scripts in multiple places)

scripts/alpha-v5.mjs
```diff
@@ -1,12 +1,12 @@
...
 
-import 'zx/globals';
+import { $, echo } from 'zx';
 
 import { constants } from './common.mjs';
 
 await $`npx json -I -f ${constants.ChangesetConfigFile} -e "this.changelog = false"`;
 
-const res = await $`npx changeset version --snapshot canary`;
+const res = await $`npx changeset version --snapshot alpha-v5`;
```
LekoArts, line 9: Snapshot is the wrong thing to do here as it won't publish at the next major, but the current version + the tag. See https://github.com/changesets/changesets/blob/main/docs/dictionary.md:

> snapshot prerelease - A snapshot pre-release is intended to make it easy to test all changes at a particular git commit. It is published at 0.0.0 and uses the git hash as the tag. This should be used as a less formal method than a release candidate to make it easy to install and test changes. An example, if we have a package at 1.0.0, and have a changeset for it to be major bumped, a snapshot release will be at 0.0.0-ABCDEFGH (the last github hash). If we add more commits, the next snapshot would be 0.0.0-HIJKLMNOP.

We'll want to use pre-release mode as explained in the same doc and by Andarist: https://github.com/changesets/action/issues/69#issuecomment-774909280

We can probably look into how stately did it: https://github.com/statelyai/xstate/pull/902

LekoArts (changes requested):

scripts/alpha-v5.mjs
LekoArts: Stately did it like this: https://github.com/statelyai/xstate/pull/878

dimkl: @LekoArts check the `To verify alpha-v5 tags` steps of the descriptions and check if the packages are published under the next major version.
This may not the best way to make the pre-release tag of the next major version but since we have this mechanism i would go with this one.

.changeset/pre.json
```diff
@@ -0,0 +1,22 @@
+{
+  "mode": "pre",
+  "tag": "premajor",
```
LekoArts, line 3: We'll want to tag it with this

```suggestion
  "tag": "alpha-v5",
```

.github/workflows/release-prod.yml
```diff
@@ -4,7 +4,7 @@ run-name: Production release PR
 on:
   push:
     branches:
-      - main-v4
+      - main
```
LekoArts, line 7: We could/should add `release/v4` here, too?

At least Stately did it: https://github.com/statelyai/xstate/pull/878/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34

.changeset/pre.json
nikosdouvlis, line 3: This should be `alpha` or `alpha-v5` I think
```suggestion
  "tag": "premajor",
```

nikosdouvlis (approved):
One minor comment, everything else makes sense!

.github/workflows/release-prod.yml
dimkl: I will do it but it does not really matter because the Github actions are executed per branch. If you check the `main-v4` you will see that this line is updated with the `main-v4`.
If we update this, we could avoid introducing a specific commit to each release branch.

nikosdouvlis: I think yes

dimkl: done!

.changeset/pre.json
dimkl: done!

LekoArts (approved):

.github/workflows/release-prod.yml
```diff
@@ -39,8 +40,6 @@ jobs:
         id: changesets
         uses: changesets/action@v1
         with:
-          title: Version Packages
-          commit: "chore(repo): Version packages"
```
dimkl, line 43: ❓ @LekoArts  why did we remove this?

LekoArts: With this the PRs/Commits will have the title "Version Packages (alpha-v5)" instead of "Version Packages" -- and we don't need to change it for every change of pre tag or new major

PR #1960, opened 2023-10-30
fix(nextjs): Deprecate `user`, `session`, `organization` from `auth()`

panteliselef:
## Description

Deprecation before removal of resources. See #1947 

## Checklist

- [ ] `npm test` runs as expected.
- [ ] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerkinc/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/chrome-extension`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

PR #1987, opened 2023-10-31
fix(nextjs): Deprecate user, session, organization from auth() (#1960)

LekoArts:
Backporting #1960 to the main branch

(cherry picked from commit 1a95065eba3809bfecd0a6584b6a1465d172b36c)


LekoArts:
We actually dropped this in https://github.com/clerkinc/javascript/pull/1947 so no need to cherry-pick

PR #2020, opened 2023-11-02
feat(clerk-js): Change the default behavior of afterSignOutUrl, afterSignIn and afterSignUp

octoper:
## Description

This PR changes the default behavior of the following

- Change the default behavior of after `afterSignOutUrl` prop in `<UserButton />` to be redirected to `/` instead of the Account Portal defined url.
- Change the default behavior of after `afterSignIn` and `afterSignUp` props to be redirected to `/` instead of the Account Portal defined url.
- Added a utility function that removed undefined values of objects in the `@clerk/shared` package.
- Use the default options for the `Clerk` singleton class when loading.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [X] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [X] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`


packages/clerk-js/src/core/clerk.ts
```diff
@@ -139,8 +140,8 @@ const defaultOptions: ClerkOptions = {
   isSatellite: false,
   signInUrl: undefined,
   signUpUrl: undefined,
-  afterSignInUrl: undefined,
-  afterSignUpUrl: undefined,
+  afterSignInUrl: '/',
+  afterSignUpUrl: '/',
```
anagstef, line 144: There are more places in our codebase where we use `displayConfig.afterSignInUrl` and `displayConfig.afterSignUpUrl`. Are those also considered?

octoper: After the changes I made the default values will be directly taken from the `[defaultOptions](https://github.com/clerk/javascript/pull/2020/files#diff-1952d6b8ac4b5495e1dba57c32bff9ce72d43f83595c1bc8532a76cf31feb753L135-L136)` so the `displayConfig.afterSignInUrl` or `displayConfig.afterSignUpUrl` are never used.

nikosdouvlis (changes requested):
@octoper You will need to apply the changes to the ui-retheme dir.  I will share more details within the day

packages/clerk-js/src/core/clerk.redirects.test.ts
```diff
@@ -128,18 +128,30 @@ describe('Clerk singleton - Redirects', () => {
 
       it('redirects to signInUrl', async () => {
         await clerkForProductionInstance.redirectToSignIn({ redirectUrl: 'https://www.example.com/' });
-        expect(mockNavigate).toHaveBeenNthCalledWith(1, '/sign-in#/?redirect_url=https%3A%2F%2Fwww.example.com%2F');
+        expect(mockNavigate).toHaveBeenNthCalledWith(
+          1,
+          '/sign-in#/?after_sign_in_url=%2F&after_sign_up_url=%2F&redirect_url=https%3A%2F%2Fwww.example.com%2F',
```
anagstef, line 133: I guess this is the correct result, cause it overrides the AP values with the new `/` default. My only concern is if this is considered too much noise on the address bar. But I don't think there is another way to do this without changing the AP values.

cc @dimkl @nikosdouvlis WDYT?

octoper: Yeah that was my concern to that maybe this will make to much noise, but I couldn't find a better way to do it without doing a major refactor on how do we build URLs.

anagstef (approved):

packages/clerk-js/src/core/clerk.ts
```diff
@@ -129,8 +130,8 @@ const defaultOptions: ClerkOptions = {
...
```
nikosdouvlis, line 134: I need to take another look, but my instinct tells me that this is not the direction we want to go as this will cause these to have actual values so, by default, the URLs will include the query params.

What we want to do instead, is to change the behavior of Clerk IF these are undefined

nikosdouvlis (changes requested):

packages/clerk-js/src/core/clerk.redirects.test.ts
dimkl: I think that we should use the `/` as a fallback in the code instead of passing them as parameters in the URL.

packages/clerk-js/src/utils/pagesToOffset.ts
```diff
@@ -1,17 +1,10 @@
+import { getNonUndefinedValues } from '@clerk/shared';
```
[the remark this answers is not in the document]
octoper: Fixed!

packages/clerk-js/src/core/clerk.ts
octoper: Made the changes to only fallback to `/` and the values will not be included to the query params.

packages/clerk-js/src/core/clerk.redirects.test.ts
octoper: @dimkl Done!

packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
```diff
@@ -210,7 +210,7 @@ function _SignUpStart(): JSX.Element {
     card.setError(undefined);
 
     const redirectUrl = buildSSOCallbackURL(ctx, displayConfig.signUpUrl);
-    const redirectUrlComplete = ctx.afterSignUpUrl || displayConfig.afterSignUpUrl;
+    const redirectUrlComplete = ctx.afterSignUpUrl || '/' || displayConfig.afterSignUpUrl;
```
anagstef, line 213: The final part of the clause here is never going to run.
```suggestion
    const redirectUrlComplete = ctx.afterSignUpUrl || '/';
```

packages/clerk-js/src/ui/components/SignUp/SignUpSocialButtons.tsx
```diff
@@ -20,7 +20,7 @@ export const SignUpSocialButtons = React.memo((props: SignUpSocialButtonsProps)
   const ctx = useSignUpContext();
   const signUp = useCoreSignUp();
   const redirectUrl = buildSSOCallbackURL(ctx, displayConfig.signUpUrl);
-  const redirectUrlComplete = ctx.afterSignUpUrl || displayConfig.afterSignUpUrl;
+  const redirectUrlComplete = ctx.afterSignUpUrl || '/' || displayConfig.afterSignUpUrl;
```
anagstef, line 23:
```suggestion
  const redirectUrlComplete = ctx.afterSignUpUrl || '/';
```

packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
octoper: Done!

packages/clerk-js/src/ui/components/SignUp/SignUpSocialButtons.tsx
octoper: Done!

SokratisVidros (approved):

packages/shared/src/underscore.ts
```diff
@@ -108,3 +108,12 @@ export function isTruthy(value: unknown): boolean {
   // Default to false
   return false;
 }
+
+export function getNonUndefinedValues<T extends object>(obj: T): Partial<T> {
```
panteliselef, line 112: 💯

LekoArts (dismissed):

packages/shared/src/__tests__/underscore.test.ts
```diff
@@ -1,4 +1,14 @@
-import { deepCamelToSnake, deepSnakeToCamel, isIPV4Address, isTruthy, titleize, toSentence } from '../underscore';
+import exp from 'constants';
```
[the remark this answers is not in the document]
octoper: Oh you are right, good catch, thank you!

dimkl:
@nikosdouvlis Could you take another look at this PR and request for other changes or approve this?

nikosdouvlis (approved):

PR #2038, opened 2023-11-03
feat(nextjs): Introduce `apiUnauthorizedResponseTransformer` function in `authMiddleware`

anagstef:
## Description

This pull request proposes a potential solution for users who have created a Nextjs + tRPC project and encounter errors on the tRPC client when Clerk's `authMiddleware` returns a NextResponse with a `null` body. This occurs when the client expects to receive a tRPC structured error object.

To address the issue, we are introducing the `apiUnauthorizedResponseTransformer` property on `authMiddleware`. This property is a function that users can pass to transform the `NextResponse` generated by Clerk when an API endpoint call results in the Unknown state (401 Unauthorized).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [x] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/nextjs/src/server/authMiddleware.ts
+++ b/packages/nextjs/src/server/authMiddleware.ts
@@ -132,6 +132,10 @@ type AuthMiddlewareParams = WithAuthOptions & {
    */
   apiRoutes?: ApiRoutesParam;
   /**
+   * A function that transforms the 401 Unauthorized response returned by the middleware when hitting an API endpoint.
+   */
+  apiUnauthorizedResponseTransformer?: (res: NextResponse) => NextResponse;
+  /**
    * Enables extra debug logging.
    */
   debug?: boolean;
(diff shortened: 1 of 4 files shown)


dimkl (approved):
❓ @anagstef Why not provide a more generic `errorHandler` option instead of an option for the specific case?

anagstef:
> ❓ @anagstef Why not provide a more generic `errorHandler` option instead of an option for the specific case?

This implementation catches all the errors thrown by Clerk authMiddleware, except the interstitial ones, that we don't want the users to manipulate them.

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -131,6 +131,10 @@ type AuthMiddlewareParams = WithAuthOptions & {
    * @default undefined
    */
   apiRoutes?: ApiRoutesParam;
+  /**
+   * A function that transforms the 401 Unauthorized response returned by the middleware when hitting an API endpoint.
+   */
```
panteliselef, line 136: 🙃 Maybe add a reminder here that this will be used only for routes marked as `apiRoutes`

panteliselef (approved):

anagstef:
We have decided not to introduce this new property, as the tRPC-related issue can be resolved with the use of a snippet similar to the following:
```
const middleware = async (request: NextRequest, event: NextFetchEvent) => {
  const res: NextResponse = (await authMiddleware({
    publicRoutes: ["/api/trpc/(.*)"],
    signInUrl: "/sign-in",
  })(request, event)) as NextResponse;

  if (res.status === 401 && request.nextUrl.pathname.startsWith("/api/trpc")) {
    // manipulate the error here
  }

  return res;
};

export default middleware;
```

PR #2040, opened 2023-11-03
chore(repo): Update remaining GHA workflows

tmilewski:
## Description

Updates existing GHA workflows to utilized the provided actions.

Fixes cache being constantly re-created.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [x] `build/tooling/chore`


.github/workflows/release-staging.yml
```diff
@@ -27,18 +27,25 @@ jobs:
...
+          playwright-enabled: true # Must be present to enable caching on branched workflows
 
       - name: Version packages for staging
         id: version-packages
         run: npm run version:staging | tail -1 >> "$GITHUB_OUTPUT"
 
       - name: Staging release
         if: steps.version-packages.outputs.success == '1'
-        run: npm run release:staging
+        run: npx turbo build $TURBO_ARGS && changeset publish --tag snapshot --no-git-tag
```
LekoArts, line 44:
```suggestion
        run: npx turbo build $TURBO_ARGS && changeset publish --tag staging --no-git-tag
```

dimkl:
@tmilewski i have also applied this fix, you probably need to apply this to your changes: https://github.com/clerk/javascript/pull/2048

LekoArts (changes requested):

.github/workflows/nightly-checks.yml
```diff
@@ -5,21 +5,56 @@ on:
...
+
+      - name: Verdaccio
+        uses: ./.github/actions/verdaccio
+        with:
+          publish-cmd: |
+            if [ "$(npm config get registry)" = "https://registry.npmjs.org/" ]; then echo 'Error: Using default registry' && exit 1; else npx turbo build $TURBO_ARGS && npx changeset publish --no-git-tag; fi
+
+      - name: Install @clerk/backend in /integration
+        working-directory: ./integration
+        run: npm init -y && npm install @clerk/backend
```
BRKalow, line 41: ❓ Why do we need to do this?

.github/workflows/nightly-checks.yml
```diff
@@ -5,21 +5,56 @@ on:
     - cron: '0 7 * * *'
 
 jobs:
-  build:
-    name: Build packages and upload artifacts
```
BRKalow, line 9: ❓ Is the build handled somewhere else now?

.github/workflows/release-canary.yml
```diff
@@ -27,18 +27,25 @@ jobs:
...
+          playwright-enabled: true # Must be present to enable caching on branched workflows
 
       - name: Version packages for canary
         id: version-packages
         run: npm run version:canary | tail -1 >> "$GITHUB_OUTPUT"
 
       - name: Canary release
         if: steps.version-packages.outputs.success == '1'
-        run: npm run release:canary
+        run: npx turbo build $TURBO_ARGS && changeset publish --tag canary --no-git-tag
```
BRKalow, line 44: ❓ Are we not using `release:canary` now so we can use `$TURBO_ARGS`?

tmilewski: That's correct.

.github/workflows/nightly-checks.yml
tmilewski: Yup, the build process is handled in line 37 within the `publish-cmd`.

.github/workflows/nightly-checks.yml
tmilewski: This is code which was simply moved from another location. I believe was present as the integration dir needs it for the test utilities vs only in the tested projects themselves. @nikosdouvlis would have more context here.

BRKalow (approved):

.github/workflows/release.yml
```diff
@@ -35,19 +37,26 @@ jobs:
...
+          turbo-token: ${{ secrets.TURBO_TOKEN }}
+          playwright-enabled: true # Must be present to enable caching on branched workflows
 
       - name: Create Release PR
         id: changesets
         uses: changesets/action@v1
         with:
           commit: "chore(repo): Version packages"
-          publish: npm run release
+          publish: npx turbo build $TURBO_ARGS && changeset publish && git push --follow-tags
```
dimkl, line 53: @tmilewski `npm run release` performs a clean build and then publishes the packages. I think that we explicitly wanted to build the packages with force to avoid any cache-related issue. With this change it seems that we may use the cache in the release action. Are we sure about this change?

cc: @nikosdouvlis

tmilewski: Yup, we can force elsewhere.

.github/workflows/nightly-checks.yml
nikosdouvlis: @BRKalow /integration uses `backend` as it needs access to a `clerkClient` instance (needs to hit BAPI)
We can move the dependency into a package.json file within /integration in a next iteration

.github/workflows/release-canary.yml
nikosdouvlis: I completely get why but I'm a bit skeptical about this change as we now have this script in 2 different places (across 2 different active branches, main and release/v4) and its very likely that these will diverge accidentally. Shall we turn them into node/shell scripts so we can pass args directly to `turbo build` without the need for duplication?

PR #2041, opened 2023-11-03
chore(shared): Browser global accessors [SDK-293]

tmilewski:
## Description

Create an accessor for browser globals such that they fail on non-browser environments.

```ts
import { browser } from '@clerk/shared'

console.log(browser.window) // throws on non-browser envs
console.log(browser.document) // throws on non-browser envs
```

I have follow-up work, converting call-sites to this new process, but I ultimately think we should update as-needed and in-line with other work to avoid potentially missed issues per cached modules in Jest.

SDK-293

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [x] `build/tooling/chore`


LekoArts (approved):

packages/shared/src/browser.ts
```diff
@@ -6,6 +11,27 @@ export function inBrowser(): boolean {
...
+
+/**
+ *
+ * Returns the browser globals if the current environment is a browser. Else, it will throw an error upon access.
+ * @returns {BrowserGlobals}
+ * @throws {Error} - If the current environment is not a browser.
+ */
+export const browser = inBrowser()
+  ? rawBrowserGlobals
+  : new Proxy(rawBrowserGlobals, {
```
nikosdouvlis, line 27: @tmilewski Could you please share some context here? Have we set up any specific test environments / eslint config to catch issues like these without testing with an actual expo app?

Unless I'm missing something, trying to access these in non-browser environments would result in errors even without the Proxy here.

packages/shared/src/browser.ts
```diff
@@ -52,8 +78,8 @@ export function userAgentIsRobot(userAgent: string): boolean {
  * @returns {boolean}
  */
 export function isValidBrowser(): boolean {
-  const navigator = window?.navigator;
-  if (!inBrowser() || !navigator) {
+  const navigator = browser.window.navigator;
```
nikosdouvlis, line 81: This changes the behavior - instead of returning `false` we're now throwing. Is this expected?

packages/shared/src/browser.ts
```diff
@@ -64,17 +90,17 @@ export function isValidBrowser(): boolean {
  * @returns {boolean}
  */
 export function isBrowserOnline(): boolean {
-  const navigator = window?.navigator;
-  if (!inBrowser() || !navigator) {
+  const navigator = browser.window.navigator;
+
+  if (!navigator) {
```
nikosdouvlis, line 95: Ditto. 
This refactor for these two cases might be fine, however I don't think we can simply replace the usage of the `inBrowser()` helper with the new proxy-based approach as the two are not equivalent.

nikosdouvlis (changes requested):

packages/shared/src/browser.ts
nikosdouvlis: Thinking about this a bit more, if this piece of code runs in an expo app, instead of returning false, its going to throw and stop app execution upon accessing `browser.window`

tmilewski:
Closing in favor of implementing linting rules.

PR #2073, opened 2023-11-07
fix(clerk-js): Emit session when permissions or role of the active memberships change

panteliselef:
## Description

This PR allows for clerk-js to detect when the role or permissions of the active organization memberships of the user has changed and emit them for subscribers (in this case clerk-react is what we are interested in).

### Before

https://github.com/clerk/javascript/assets/19269911/5a391888-d798-45f0-a1ed-6aa21e3ed9db

### After

https://github.com/clerk/javascript/assets/19269911/492f0e4b-27e0-46b0-b1d2-2803b02bab81

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/clerk-js/src/core/resources/Organization.ts
+++ b/packages/clerk-js/src/core/resources/Organization.ts
@@ -8,6 +8,7 @@ import type {
   GetInvitationsParams,
   GetMembershipRequestParams,
   GetMemberships,
+  GetMembershipsParams,
   GetPendingInvitationsParams,
   GetRolesParams,
   InviteMemberParams,
@@ -26,7 +27,6 @@ import type {
   UpdateMembershipParams,
   UpdateOrganizationParams,
 } from '@clerk/types';
-import type { GetMembershipsParams } from '@clerk/types';
 
 import { unixEpochToDate } from '../../utils/date';
 import { convertPageToOffset } from '../../utils/pagesToOffset';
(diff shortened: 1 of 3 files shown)


dimkl (approved):

packages/clerk-js/src/core/resources/Organization.ts
```diff
@@ -109,11 +109,16 @@ export class Organization extends BaseResource implements OrganizationResource {
   };
 
   getRoles = async (getRolesParams?: GetRolesParams) => {
-    return await BaseResource._fetch({
-      path: `/organizations/${this.id}/roles`,
-      method: 'GET',
-      search: convertPageToOffset(getRolesParams) as any,
-    }).then(res => {
+    return await BaseResource._fetch(
```
dimkl, line 112: 🙃 i think there is no need to `return await` in an async function. this PR didn't change the existing codebase but i think we should jest return the promise. This is a suggestion for the future.

dimkl (commented):
❓ Why do we need to pass `forceUpdateClient`? Isn't this change gonna force the re-rendering of all the Clerk components? Are we sure we want to use the forceUpdateClient to all those places?

panteliselef:
> ❓ @panteliselef Why do we need to pass `forceUpdateClient`? Isn't this change gonna force the re-rendering of all the Clerk components? Are we sure we want to use the forceUpdateClient to all those places?

My understanding is that these changes will only affect the rerender of components if the conditions inside `memoizeStateListenerCallback` are met, which is ok and what we should do.

Even if `forceUpdateClient` and the new client object has not changes this will not force the re-rendering of all components.

PR #2085, opened 2023-11-08
chore(repo): Cache extraneous node_module versions

tmilewski:
## Description

In our CI unit tests, we run against Node 18 and Node 20 and we cache the `node_modules` output.

Node 20, however, continuously creates a new cache and the reason for this is that we run CI on the `merge_group` event and not on `main`.

Branches and PRs only have access to caches created within the scope of their branch, their base branch, and the default branch, which does not include merge groups.

This adds a simple job which ensures that `node_modules` for specific node versions (in this case, just v20) are cached in a way that follow-on branches are able to access it.

The job itself will add a very minimal amount of time to `main` and `release/v4` while having significant benefits downstream.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [x] `build/tooling/chore`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,6 +35,9 @@ jobs:
 
       - name: Checkout repo
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          show-progress: false
 
       - name: Setup
         id: config
@@ -84,8 +87,6 @@ jobs:
               core.warning("Changeset in pre-mode should not prepare a ClerkJS production release")
             }
 
-
-
       - name: Generate notification payload
         id: notification
         if: steps.changesets.outputs.published == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -101,4 +102,29 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          show-progress: false
+
+      - name: Cache node_modules (Node v${{ matrix.version }})
+        uses: ./.github/actions/init
+        with:
+          node-version: ${{ matrix.version }}
+          turbo-team: ''
+          turbo-token: ''
```
BRKalow, line 130: It feels a bit odd to tack this caching thing onto the release workflow, do you think it makes sense to have a standalone workflow that runs on push to `main` that makes sure the cache is setup how we expect?

tmilewski: @BRKalow I was just trying to avoid having a sort of dangling workflow running in each push to `main` or `release/v4`.

BRKalow (approved):

PR #2111, opened 2023-11-10
fix(clerk-js): Remove internal useCore hooks

LekoArts:
## Description

This PR:
- Merges useCoreSession (clerk-js) and useSession(clerk-react) and moves it to `@clerk/shared/react` as `useSession`.
- Merges useCoreUser (clerk-js) and useUser(clerk-react) and moves it to `@clerk/shared/react` as `useUser`.
- Merges useCoreClerk (clerk-js) and useClerk(clerk-react) and moves it to `@clerk/shared/react` as `useClerk`.
- Renames useCoreOrganization(List) hooks to useOrganization(List) inside clerk-js.
- Improves parity between the hooks we are exposing and the hooks we use internally.

The changes for the clerk-js package are applied inside the `ui.retheme` and the `ui` directory

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`


BRKalow (approved):

packages/clerk-js/src/ui.retheme/components/OrganizationList/UserMembershipList.tsx
```diff
@@ -35,8 +36,12 @@ export const MembershipPreview = withCardStateProvider((props: { organization: O
...
   const card = useCardState();
   const { hidePersonal, navigateAfterSelectPersonal } = useOrganizationListContext();
-  const { isLoaded, setActive } = useCoreOrganizationList();
-  const user = useCoreUser();
+  const { isLoaded, setActive } = useOrganizationList();
+  const { user } = useUser();
+
+  if (!user) {
+    return null;
+  }
```
dimkl, line 44: ❓ is this a fix to avoid rendering `PersonalAccountPreview` when user is not available or the `useCoreUser()`  has different behaviour from the `useUser()` ?

packages/clerk-js/src/ui.retheme/components/OrganizationProfile/ActionConfirmationPage.tsx
```diff
@@ -17,10 +19,10 @@ import { OrganizationProfileBreadcrumbs } from './OrganizationProfileNavbar';
 export const LeaveOrganizationPage = () => {
   const card = useCardState();
   const { navigateAfterLeaveOrganization } = useOrganizationProfileContext();
-  const { organization } = useCoreOrganization();
-  const user = useCoreUser();
+  const { organization } = useOrganization();
+  const { user } = useUser();
 
-  if (!organization) {
+  if (!organization || !user) {
```
dimkl, line 25: same as above

packages/clerk-js/src/ui.retheme/components/OrganizationProfile/ActiveMembersList.tsx
```diff
@@ -76,9 +76,9 @@ const MemberRow = (props: {
   const { membership, onRemove, onRoleChange, options } = props;
   const { localizeCustomRole } = useLocalizeCustomRoles();
   const card = useCardState();
-  const user = useCoreUser();
+  const { user } = useUser();
```
dimkl, line 79: ❓ Don't we need  `if (!user) return null;` guard here?

packages/react/src/contexts/ClientContext.tsx
dimkl, line 1: 🔧 remove this empty file.

dimkl (approved):
❓ Instead of replacing `user.` with `user?.` in the components where `useUser()` hook is used, shouldn't we add a guard `if (!user) return null;`?
❓ Instead of adding the `if (!user) return null;` guard could we move that up the hierarchy to avoid repeating this guard to all the components of eg `UserProfile` / `OrganizationSwitcher`? (If so, let's verify that all components that need `user` have the guard and add a separate PR to make that improvement)

packages/clerk-js/src/ui.retheme/components/OrganizationProfile/ActiveMembersList.tsx
panteliselef: OrganizationProfile is wrapped in  `withCoreUserGuard` so we are safe.

packages/clerk-js/src/ui.retheme/components/OrganizationList/UserMembershipList.tsx
panteliselef: OrganizationList is wrapped in  `withCoreUserGuard` so we are safe. It doesn't really matter if we do `user?.something` or return null.

Here we did that because it was helping with the deconstruction.

PR #2114, opened 2023-11-11
feat(clerk-js,types): Pass unsafe metadata to signin component for oauth

itelo:
## Description

While I was using the SignIn component from `@clerk/nextjs`, I notice that the user is created if the user uses a social button, what I think is ok, but since I rely on the unsafe metadata this is a bad news for me because only the SignUp components supports it. This PR is to solve that issue, adding the unsafe metadata to the SignIn component too.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [x] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/clerk-js/src/core/resources/SignIn.ts
+++ b/packages/clerk-js/src/core/resources/SignIn.ts
@@ -170,8 +170,12 @@ export class SignIn extends BaseResource implements SignInResource {
     });
   };
 
-  public authenticateWithRedirect = async (params: AuthenticateWithRedirectParams): Promise<void> => {
-    const { strategy, redirectUrl, redirectUrlComplete, identifier } = params || {};
+  public authenticateWithRedirect = async (
+    params: AuthenticateWithRedirectParams & {
+      unsafeMetadata?: SignUpUnsafeMetadata;
+    },
+  ): Promise<void> => {
+    const { strategy, redirectUrl, redirectUrlComplete, identifier, unsafeMetadata } = params || {};
 
     const { firstFactorVerification } =
       strategy === 'saml' && this.id
@@ -184,6 +188,7 @@ export class SignIn extends BaseResource implements SignInResource {
             strategy,
             identifier,
             redirectUrl: SignIn.clerk.buildUrlWithAuth(redirectUrl),
+            unsafeMetadata,
             actionCompleteRedirectUrl: redirectUrlComplete,
           });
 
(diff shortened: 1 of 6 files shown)


packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -170,8 +170,12 @@ export class SignIn extends BaseResource implements SignInResource {
     });
   };
 
-  public authenticateWithRedirect = async (params: AuthenticateWithRedirectParams): Promise<void> => {
-    const { strategy, redirectUrl, redirectUrlComplete, identifier } = params || {};
+  public authenticateWithRedirect = async (
+    params: AuthenticateWithRedirectParams & {
+      unsafeMetadata?: SignUpUnsafeMetadata;
```
dimkl, line 175: ⛔  the `unsafeMetadata` is only used and allowed to be passed in`sign_ups` (which is already supported) not `sign_ins`. You could also set `unsafeMetadata` via `User#update()` but not in sign_ins.
Did you find any reference about it in our docs? How did you end up with suggesting this change? What is your use case?

dimkl (changes requested):
@itelo thank you for the contribution. I have added a comment and blocked this PR since adding the `unsafeMetadata` in `sign_ins` is not something we support or probably want to support.
Let's continue on the thread of my comment to see how we could cater your use case.

packages/clerk-js/src/core/resources/SignIn.ts
itelo: I noticed that if a user tries to use the social button to log in without having created an account before, a new account will be created automatically (which I think is acceptable). However, I am currently relying on the unsafe metadata to perform certain actions in my backend. This functionality is currently only available for the Signup component, but I believe it should also be available for the SignIn component to cover this use case.

clerk-cookie:
Hello again 👋

After 60 days of no activity, we'll close this PR. Keep in mind, I'm just a robot, so if I've closed this PR in error, please reply here and my human colleagues will reopen it.

PR #2126, opened 2023-11-13
chore(backend,nextjs,clerk-sdk-node): Drop legacy return response in BAPI responses

dimkl:
## Description

Currently the BAPI responses are in `{ data: unknown | null, errors: [] | null }` format but to avoid introducing a breaking change we convert those responses and return only the `data` part or `throw` an error if there are `errors`.
With this PR we will drop the transformation and return the actual response. This is a breaking change since from now on
- no error will be thrown from the BAPI requests client
- the caller will have to check for the `errors` prop in the return value to verify if the request succeeded
- the caller will have to retrieve the response data from the `data` prop in the return value

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [] `@clerk/fastify`
- [] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [x] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [] `@clerk/remix`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/.changeset/mighty-pugs-knock.md
+++ b/.changeset/mighty-pugs-knock.md
@@ -1,0 +1,17 @@
+---
+'@clerk/clerk-sdk-node': major
+'@clerk/backend': major
+'@clerk/nextjs': major
+---
+
+Change the response payload of Backend API requests to return `{ data, errors }` instead of return the data and throwing on error response.
+Code example to keep the same behavior:
+```typescript
+import { users } from '@clerk/backend';
+import { ClerkAPIResponseError } from '@clerk/shared/error';
+
+const { data, errors, clerkTraceId, status, statusText } = await users.getUser('user_deadbeef');
+if(errors){
+    throw new ClerkAPIResponseError(statusText, { data: errors, status, clerkTraceId });
+}
+```
--- a/packages/backend/src/api/factory.test.ts
+++ b/packages/backend/src/api/factory.test.ts
@@ -4,6 +4,7 @@ import sinon from 'sinon';
 import emailJson from '../fixtures/responses/email.json';
 import userJson from '../fixtures/responses/user.json';
 import runtime from '../runtime';
+import { assertErrorResponse, assertResponse } from '../util/assertResponse';
 import { jsonError, jsonNotOk, jsonOk } from '../util/mockFetch';
 import { createBackendApiClient } from './factory';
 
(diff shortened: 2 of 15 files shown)


packages/backend/src/api/factory.test.ts
```diff
@@ -26,20 +27,17 @@ export default (QUnit: QUnit) => {
       fakeFetch = sinon.stub(runtime, 'fetch');
       fakeFetch.onCall(0).returns(jsonOk(userJson));
 
-      const payload = await apiClient.users.getUser('user_deadbeef');
+      const response = await apiClient.users.getUser('user_deadbeef');
 
-      if (!payload) {
-        assert.false(true, 'This assertion should never fail. We need to check for payload to make TS happy.');
-        return;
-      }
```
dimkl, line 34: `assertResponse` / `assertErrorResponse` helpers were introduced to drop the `!payload` check.

packages/sdk-node/src/clerkExpressRequireAuth.ts
```diff
@@ -38,7 +38,12 @@ export const createClerkExpressRequireAuth = (createOpts: CreateClerkExpressMidd
           clerkClient,
           requestState,
         });
-        return handleInterstitialCase(res, requestState, interstitial);
+        if (interstitial.errors) {
+          // TODO(@dimkl): return interstitial errors ?
```
dimkl, line 42: @nikosdouvlis WDYT about passing the errors of the interstitial request to the end-user?
We would probably remove this handling if we drop the `remotePrivateInterstitial()`.

.changeset/mighty-pugs-knock.md
```diff
@@ -0,0 +1,7 @@
+---
+'@clerk/clerk-sdk-node': major
+'@clerk/backend': major
+'@clerk/nextjs': major
+---
+
+Change the response payload of Backend API requests to return `{ data, errors }` instead of return the data and throwing on error response.
```
LekoArts, line 7: We should give people a Before/After example to show how they can migrate

LekoArts (commented):
Rest LGTM

.changeset/mighty-pugs-knock.md
dimkl: I have provided an example of how to keep the same behavior.

BRKalow (approved):

nikosdouvlis (approved):

packages/sdk-node/src/clerkExpressRequireAuth.ts
dimkl: We haven't removed it yet, since we need to decide if we want to enforce the publishableKey in all our SDKs.
The local interstitial won't ever return an error but if the publishableKey is missing, fetching interstitial from BAPI will be triggered and that could result to the following errors (based on the Go backend):
- 401 due to missing secretKey or invalid Key
- 500

PR #2127, opened 2023-11-14
Fix CLERK_RETHEME build issue on clerk-js

nikosdouvlis:
## Description

Introduce the `~ui` path alias on the build and type level (webpack and tsconfig) so we can conditionally build with or without the new UI components. The new alias was needed because we `paths` in tsconfig will not work with relative imports, so we could not alias the top-level imports on `index.ts` and `index.browser.ts`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`


packages/clerk-js/tsconfig.json
```diff
@@ -20,8 +20,10 @@
     "useUnknownInCatchVariables": false,
     "declaration": false,
     "jsx": "react-jsx",
-    "jsxImportSource": "@emotion/react"
-    // "noUncheckedIndexedAccess": true // TODO: Enable and progressively fix related errors
```
panteliselef, line 24: @nikosdouvlis Shall we add a ticket to bring `noUncheckedIndexedAccess` back ?

packages/clerk-js/src/ui.retheme/components/OrganizationProfile/BillingWidget.tsx
```diff
@@ -1,6 +1,6 @@
-import { useRouter } from '../../../ui/router';
 import { runIfFunctionOrReturn } from '../../../utils';
 import { AlertIcon, Flex, Link, Text } from '../../customizables';
+import { useRouter } from '../../router';
```
panteliselef, line 3: Now that we have the alias isn't it better to do `from '~ui/router'` ?

dimkl (commented):
❓ Instead of creating the alias, could we move those to `utils`?

packages/clerk-js/tsconfig.json
nikosdouvlis: This wasn't changed as part of this PR but agreed, let's add a ticket in triage

desiprisg (approved):

nikosdouvlis:
> ❓ Instead of creating the alias, could we move those to `utils`?

No, as even if we moved them, we'd still need to conditionally import from `ui` or `ui.retheme` at some point, so it'd be the same.

PR #2128, opened 2023-11-14
feat(clerk-js): Add default Allowed redirect origins

octoper:
## Description

This PR introduces default values for the `allowedRedirectOrigins` option, the current implementation does not provide any defaults, with this change if there is no option provided the default will be similar to the example below.

Let's say the host of the application is `test.host`, the origins will be
- `https://test.host/`
- `https://yourawesomeapp.clerk.accounts.dev/`
- `https://*.yourawesomeapp.clerk.accounts.dev/`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [X] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [X] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/.changeset/fast-ads-mix.md
+++ b/.changeset/fast-ads-mix.md
@@ -6,6 +6,5 @@ Introducing default values for allowed redirect origins, this change will apply 
 
 Let's assume the host of the application is `test.host`, the origins will be
 - `https://test.host/`
-- `https://test.host/*`
+- `https://yourawesomeapp.clerk.accounts.dev/`
 - `https://*.yourawesomeapp.clerk.accounts.dev/`
-- `https://*.yourawesomeapp.clerk.accounts.dev/*`
--- a/.changeset/fast-ads-mix.md
+++ b/.changeset/fast-ads-mix.md
@@ -1,0 +1,11 @@
+---
+'@clerk/clerk-js': minor
+---
+
+Introducing default values for allowed redirect origins, this change will apply default values for the `allowedRedirectOrigins option`, if there is no value provided the defaults will be similar to the example below.
+
+Let's assume the host of the application is `test.host`, the origins will be
+- `https://test.host/`
+- `https://test.host/*`
+- `https://*.yourawesomeapp.clerk.accounts.dev/`
+- `https://*.yourawesomeapp.clerk.accounts.dev/*`
--- a/.changeset/fast-ads-mix.md
+++ b/.changeset/fast-ads-mix.md
@@ -2,7 +2,7 @@
 '@clerk/clerk-js': minor
 ---
 
-Introducing default values for allowed redirect origins, this change will apply default values for the `allowedRedirectOrigins option`, if there is no value provided the defaults will be similar to the example below.
+Introducing default values for `allowedRedirectOrigins`. If no value is provided, default values similar to the example below will apply.
 
 Let's assume the host of the application is `test.host`, the origins will be
 - `https://test.host/`
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -60,12 +60,12 @@ import {
   appendAsQueryParams,
   buildURL,
   completeSignUpFlow,
+  createAllowedRedirectOrigins,
   createBeforeUnloadTracker,
   createCookieHandler,
   createPageLifecycle,
   errorThrower,
   getClerkQueryParam,
-  getETLDPlusOneFromFrontendApi,
   hasExternalAccountSignUpError,
   ignoreEventValue,
   inActiveBrowserTab,
@@ -256,25 +256,6 @@ export class Clerk implements ClerkInterface {
 
   public isReady = (): boolean => this.#isReady;
 
-  get allowedRedirectOrigins(): (string | RegExp)[] | undefined {
-    if (!this.#options.allowedRedirectOrigins) {
-      const origins = [];
-      if (inBrowser()) {
-        origins.push(window.location.origin);
-        origins.push(window.location.origin + '/*');
-      }
-
-      const frontendApi = this.frontendApi;
-
-      origins.push(`https://*.${getETLDPlusOneFromFrontendApi(frontendApi)}`);
-      origins.push(`https://*.${getETLDPlusOneFromFrontendApi(frontendApi)}/*`);
-
-      return origins;
-    }
-
-    return this.#options.allowedRedirectOrigins;
-  }
-
   public load = async (options?: ClerkOptions): Promise<void> => {
     if (this.#isReady) {
       return;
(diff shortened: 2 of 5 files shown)


dimkl (commented):
🔧 Based on the definition of the origin from [MDN docs](https://developer.mozilla.org/en-US/docs/Glossary/Origin), I think that we should only allow passing origin values to the `allowedRedirectOrigins` option (eg `${protocol}://${domain}:${port}`). Doing that will also allow us to drop the `/*` path default origins. If we want to match paths in allowedRedirectOrigins i would advise we rename the options.
cc: @nikosdouvlis

octoper:
!preview

clerk-cookie:

| :-- | :-- | :-- |
| :cook: _Deploying..._ | [Inspect](https://javascript-preview-gbrvz58m4.clerkpreview.com) | Nov 15, 2023 10:48 AM |

packages/clerk-js/src/core/clerk.ts
```diff
@@ -255,6 +256,25 @@ export default class Clerk implements ClerkInterface {
...
+
+      origins.push(`https://*.${getETLDPlusOneFromFrontendApi(frontendApi)}`);
+      origins.push(`https://*.${getETLDPlusOneFromFrontendApi(frontendApi)}/*`);
+
+      return origins;
+    }
+
+    return this.#options.allowedRedirectOrigins;
+  }
+
```
nikosdouvlis, line 277: This affects the public API of `window.Clerk`. Let's make it a util instead for now, as we haven't discussed making it public; we can always add it on `window.Clerk` later if required

packages/clerk-js/src/core/clerk.ts
```diff
@@ -255,6 +256,25 @@ export default class Clerk implements ClerkInterface {
...
+      const origins = [];
+      if (inBrowser()) {
+        origins.push(window.location.origin);
+        origins.push(window.location.origin + '/*');
+      }
+
+      const frontendApi = this.frontendApi;
+
+      origins.push(`https://*.${getETLDPlusOneFromFrontendApi(frontendApi)}`);
+      origins.push(`https://*.${getETLDPlusOneFromFrontendApi(frontendApi)}/*`);
```
nikosdouvlis, line 270: Why do we require the `/*` part here? Can you show an example where this is needed? Same question for line 264.

nikosdouvlis (changes requested):
@octoper this looks good, we need a minor change but apart from that everything is in place. Let's try and merge it today :)

nikosdouvlis:
> 🔧 Based on the definition of the origin from [MDN docs](https://developer.mozilla.org/en-US/docs/Glossary/Origin), I think that we should only allow passing origin values to the `allowedRedirectOrigins` option (eg `${protocol}://${domain}:${port}`). Doing that will also allow us to drop the `/*` path default origins. If we want to match paths in allowedRedirectOrigins i would advise we rename the options. cc: @nikosdouvlis

The `/*` part should not be required for someone to use `allowedRedirectOrigins` successfully. These tests cover almost all valid use cases (https://github.com/clerk/javascript/blob/d37d44a68e83b8e895963415f000c1aaef66e87e/packages/clerk-js/src/utils/__tests__/url.test.ts#L421) without `/*` 

Are you suggesting we manually remove `/*` if the user passes it in? Would that offer any benefit DX-wise?

packages/clerk-js/src/core/clerk.ts
octoper: Oh my bad this is not required, removing it as path doesn't matter for the check of same-origin.

packages/clerk-js/src/utils/url.ts
```diff
@@ -350,3 +350,22 @@ export const isAllowedRedirectOrigin = (_url: string, allowedRedirectOrigins: Ar
...
+      origins.push(window.location.origin);
+    }
+
+    origins.push(`https://${getETLDPlusOneFromFrontendApi(frontendApi)}`);
+    origins.push(`https://*.${getETLDPlusOneFromFrontendApi(frontendApi)}`);
+
+    return origins;
+  }
+
+  return allowedRedirectOrigins;
```
nikosdouvlis, line 370: Nitpicky, feel free to ignore. I usually like to flip the conditional and write this the other way around: 

```
if (Array.isArray(allowedRedirectOrigins) && !!allowedRedirectOrigins.length) {
  return allowedRedirectOrigins;
} 
// more logic
```

I find it slightly more readable as you take care of the simple things right away and then handle the "difficult" part

nikosdouvlis (approved):

.changeset/fast-ads-mix.md
```diff
@@ -0,0 +1,10 @@
+---
+'@clerk/clerk-js': minor
+---
+
+Introducing default values for allowed redirect origins, this change will apply default values for the `allowedRedirectOrigins option`, if there is no value provided the defaults will be similar to the example below.
```
anagstef, line 5:
```suggestion
Introducing default values for allowed redirect origins, this change will apply default values for the `allowedRedirectOrigins` option, if there is no value provided the defaults will be similar to the example below.
```

anagstef (approved):

packages/clerk-js/src/utils/url.ts
```diff
@@ -350,3 +350,22 @@ export const isAllowedRedirectOrigin = (_url: string, allowedRedirectOrigins: Ar
...
   return isAllowed;
 };
+
+export function createAllowedRedirectOrigins(
+  allowedRedirectOrigins: Array<string | RegExp> | undefined,
+  frontendApi: string,
+): (string | RegExp)[] | undefined {
+  if (!allowedRedirectOrigins || allowedRedirectOrigins.length === 0) {
+    const origins = [];
+    if (typeof window !== 'undefined' && !!window.location) {
```
dimkl, line 360: ❓ could we use the `inBrowser()` helper instead? @anagstef wdyt?

dimkl (approved):

packages/clerk-js/src/core/clerk.ts
octoper: Done!

packages/clerk-js/src/utils/url.ts
octoper: Oh I had that but change it as another utility function in the same file had it like this so I stick with that for consistency, but I can change the `inBrowser` for every function here.

packages/clerk-js/src/utils/url.ts
octoper: Agree, Done!

dimkl (approved):
🚀

PR #2134, opened 2023-11-14
feat(clerk-js): Add TelemetryCollector instance to backend client

BRKalow:
## Description

Adds a `TelemetryCollector` instance to the clerk backend client to allow collection in server environments. I also updated the collector to accept an instance's `secretKey`. Ultimately we do only collect the first 16 characters of the secret key to allow for later analysis.

I'm not entirely happy with piggybacking off of the backend client like this, but for now I think it's an easy way to consistently handle server-side telemetry across our SDKs without too much extra boilerplate.

Fixes SDK-909

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/fastify`
- [x] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [x] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [x] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`


packages/shared/src/telemetry.ts
```diff
@@ -21,13 +21,13 @@ type TelemetryCollectorOptions = {
    */
   maxBufferSize?: number;
   /**
-   * Determines whether or not events will be logged to the console.
+   * The publishableKey to associate with the collected events.
    */
-  verbose?: boolean;
```
BRKalow, line 26: `verbose` is now unused with the `debug` option.

SokratisVidros (approved):

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -154,7 +146,7 @@ const authMiddleware: AuthMiddleware = (...args: unknown[]) => {
   const isApiRoute = createApiRoutes(apiRoutes);
   const defaultAfterAuth = createDefaultAfterAuth(isPublicRoute, isApiRoute, params);
 
-  telemetry.record('METHOD_CALLED', {
+  clerkClient.telemetry.record('METHOD_CALLED', {
```
dimkl, line 149: ☁️ I think we should not depend on the `clerkClient` as customers may have initialized and used a different clerk instance to authenticate their requests. We should record those telemetry metrics in that clerk instance and not on the default clerk instance.

BRKalow: The dependency is fairly loose here, it was a convenient spot to piggyback off of the PK and SK they are providing to the backend client. Alternatively we could have each SDK instantiate its own telemetry instance.

Ultimately the SK and PK that we are using here isn't necessarily critical...it's more important that the SDK attribution is correct I think.

dimkl: I thought the attribution was based on the PK / SK to determine the instance_id. If it's not, then ignore my previous comment. 
It's about correct attribution and not about making each SDK instantiate its own telemetry. In the current line there may not be a valid clerk client instantiate since we may instantiate the clerk client during the `authenticateRequest` step.

BRKalow: > I thought the attribution was based on the PK / SK to determine the instance_id.

@dimkl This is true, but what I'm saying is it's not absolutely critical. Telemetry is an internal concern, so my current thinking is that the most reliable approach we can take is to use the PK/SK that are provided through the recommended environment variables. We want a best-effort attribution to the PK/SK the application is using.

PR #2145, opened 2023-11-16
chore(types): Deprecate navigate

desiprisg:
## Description

Depracate the `navigate` prop as it will be replaced in v5 with `push`.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [x] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/types/src/clerk.ts
+++ b/packages/types/src/clerk.ts
@@ -532,6 +532,9 @@ export type ClerkThemeOptions = DeepSnakeToCamel<DeepPartial<DisplayThemeJSON>>;
 export interface ClerkOptions {
   appearance?: Appearance;
   localization?: LocalizationResource;
+  /**
+   * @deprecated This prop has been deprecated and will be removed in the next major release.
+   */
   navigate?: (to: string) => Promise<unknown> | unknown;
   polling?: boolean;
   selectInitialSession?: (client: ClientResource) => ActiveSessionResource | null;


packages/types/src/clerk.ts
```diff
@@ -532,6 +532,9 @@ export type ClerkThemeOptions = DeepSnakeToCamel<DeepPartial<DisplayThemeJSON>>;
 export interface ClerkOptions {
   appearance?: Appearance;
   localization?: LocalizationResource;
+  /**
+   * @deprecated This prop has been deprecated and will be removed in the next major release.
```
dimkl, line 536: ❓ is there a replacement for this prop?

panteliselef: I think it will be `push` and `replace` and they will be available in v5

nikosdouvlis:
@panteliselef @desiprisg 
I believe we cannot deprecate this method right now as we don't provide a replacement in v4. If we deprecate it right now, users will be right to look for a replacement or reach out to support for this. I thnk this is one of the few instances where a deprecation warning will cause more hard than good and I'm tempted to close the PR - what do you think?

nikosdouvlis:
This PR deprecates a prop that has no replacement yet. I'm fairly confident that this will cause more harm than good, as customers will try to find a replacement for it, but such a replacement does not exist yet - it will become available in v5.  I remember that during the release of the clerk/backend package, we accidentally pushed a single deprecation warning to NPM without stating what the replacement API is, and in less than an hour, we received support requests about it.
If we want to deprecate it, we should at least be as descriptive as possible (eg "this will be removed when v5 is released and will be replaced by the xxx APIs that will be available as parts of that release"). Unless we are 100% certain of the new APIs names, let's not do it yet.

PR #2154, opened 2023-11-17
feat(shared): SDK telemetry

BRKalow:
## Description

Introduces opt-out telemetry collection from our SDKs by way of a new `TelemetryCollector` class. Currently, this supports collection both on the client, by way of our core Clerk instance, and on the server, by way of the pre-initialized `@clerk/backend` client.

Client collection:
```js
Clerk.telemetry.record({ event: 'EVENT', payload: { ... } })
```

Server collection:
```js
clerkClient.telemetry.record({ event: 'EVENT', payload: { ... } })
```

The `clerkClient` factory from `@clerk/backend` now accepts `sdkMetadata`, which is passed from each SDK. This metadata is then passed to `TelemetryCollector` and included in telemetry events. Currently, collection is only done from development instances.

### Opting-out

To opt-out, we respect a `CLERK_TELEMETRY_DISABLE` environment variable. Frameworks with unique environment variable prefixes, such as Next's `NEXT_PUBLIC_` will also disable telemetry collection.

### Docs

Documentation is being prepared here: https://github.com/clerk/clerk-docs/pull/496

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/backend`
- [x] `@clerk/chrome-extension`
- [x] `@clerk/clerk-js`
- [x] `@clerk/clerk-expo`
- [x] `@clerk/fastify`
- [x] `gatsby-plugin-clerk`
- [x] `@clerk/localizations`
- [x] `@clerk/nextjs`
- [x] `@clerk/clerk-react`
- [x] `@clerk/remix`
- [x] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/themes`
- [x] `@clerk/types`
- [ ] `build/tooling/chore`


packages/clerk-js/src/ui.retheme/localization/defaultEnglishResource.ts
```diff
@@ -1,4 +1,4 @@
 import { enUS } from '@clerk/localizations';
 import type { DeepRequired } from '@clerk/types';
 
-export const defaultResource = enUS as DeepRequired<typeof enUS>;
```
BRKalow, line 4: Hmm, I think this might have been an automatic lint "fix" 🤔

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -145,6 +146,14 @@ const authMiddleware: AuthMiddleware = (...args: unknown[]) => {
   const isApiRoute = createApiRoutes(apiRoutes);
   const defaultAfterAuth = createDefaultAfterAuth(isPublicRoute, isApiRoute, params);
 
+  clerkClient.telemetry.record('METHOD_CALLED', {
+    method: 'authMiddleware',
+    publicRoutes: Boolean(publicRoutes),
+    ignoredRoutes: Boolean(ignoredRoutes),
+    beforeAuth: Boolean(beforeAuth),
+    afterAuth: Boolean(afterAuth),
+  });
```
BRKalow, line 155: This is how we'll collect events in a server environment.

packages/react/src/isomorphicClerk.ts
```diff
@@ -314,6 +320,15 @@ export default class IsomorphicClerk {
...
   }
 
+  get telemetry(): TelemetryCollector | undefined {
+    if (this.clerkjs) {
+      // @ts-expect-error -- We can't add the type here due to the TelemetryCollector type existing in shared, but the Clerk type existing in types
+      return this.clerkjs.telemetry;
+    } else {
+      return undefined;
+    }
+  }
```
BRKalow, line 330: This will allow us to record telemetry via the `useClerk` hook, which will let us instrument more easily in our components.

packages/shared/src/__tests__/telemetry.test.ts
```diff
@@ -0,0 +1,155 @@
+import 'cross-fetch/polyfill';
```
BRKalow, line 1: `fetch` isn't available in the `jsdom` jest environment 😢, so adding the polyfill here.

packages/shared/src/telemetry.ts
```diff
@@ -0,0 +1,292 @@
...
+      this.#pendingFlush = requestIdleCallback(() => {
+        this.#flush();
+      });
+    } else {
+      // This is not an ideal solution, but it at least waits until the next tick
+      this.#pendingFlush = setTimeout(() => {
+        this.#flush();
+      }, 0);
+    }
+  }
```
BRKalow, line 218: This was my attempt to implement buffering in a way that mitigates potential data loss. It uses `requestIdleCallback` to schedule sending events during idle time. When an event is recorded, it schedules a flush during idle, and if the buffer limit is reached it immediately flushes the events.

The scheduling of a flesh on every record might be a tad aggressive, but we can tweak later if we want.

packages/shared/src/telemetry.ts
```diff
@@ -0,0 +1,292 @@
...
+
+  #flush(): void {
+    fetch(new URL('/v1/event', this.#config.endpoint), {
+      method: 'POST',
+      // TODO: We send an array here with that idea that we can eventually send multiple events.
```
BRKalow, line 223:
```suggestion
```

packages/shared/src/telemetry.ts
```diff
@@ -0,0 +1,292 @@
...
+      }),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    })
+      .catch(() => void 0)
+      .then(() => {
+        this.#buffer = [];
+      })
+      .catch(() => void 0);
```
BRKalow, line 235: Catch into the void to avoid leaking any errors here. ESLint complained that the chain had to in a `.catch`, hence the final `.catch` as well.

packages/shared/src/__tests__/telemetry.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    });
+
+    collector.record('TEST_EVENT', {});
+    collector.record('TEST_EVENT', {});
+
+    expect(fetchSpy).toHaveBeenCalled();
+
+    fetchSpy.mockRestore();
+  });
+});
```
jescalan, line 155: Tests 😍

jescalan (approved):
Two small thoughts:

1) Should we add a post-install message notifying about telemetry and how to opt out as is done in nextjs etc here?
2) Can we get events sent for hook usage here too, or is that planned for a later PR?

BRKalow:
@jescalan

> Should we add a post-install message notifying about telemetry and how to opt out as is done in nextjs etc here?

Yes, good reminder! I'll add this.

> Can we get events sent for hook usage here too, or is that planned for a later PR?

Will open a follow-up PR with hooks usage events 👍

packages/shared/src/telemetry/collector.ts
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  get isEnabled(): boolean {
+    if (this.#metadata.instanceType !== 'development') {
+      return false;
+    }
+
+    // In browser or client environments, we most likely pass the disabled option to the collector, but in environments
+    // where environment variables are available we also check for `CLERK_TELEMETRY_DISABLED`.
+    if (this.#config.disabled || (typeof process !== 'undefined' && isTruthy(process.env.CLERK_TELEMETRY_DISABLED))) {
```
nikosdouvlis, line 157: I'm nitpicking here 🙃  but in general, I tend to avoid double negatives like `CLERK_TELEMETRY_DISABLED=τρθε` as I find them slightly confusing. I think that positive conditions like `CLERK_TELEMETRY_ENABLED=false` are a bit easier to work with. How do you feel about renaming this to `CLERK_TELEMETRY_ENABLED` and having the users disable telemetry by passing `0` or `false`?

packages/shared/src/telemetry/collector.ts
```diff
@@ -0,0 +1,300 @@
...
+  get isDebug(): boolean {
+    return this.#config.debug || (typeof process !== 'undefined' && isTruthy(process.env.CLERK_TELEMETRY_DEBUG));
+  }
+
+  record(event: Pick<TelemetryEvent, 'event' | 'payload'>): void;
+  record(event: TelemetryEvent['event'], payload: TelemetryEvent['payload']): void;
+  record(
+    event: TelemetryEvent['event'] | Pick<TelemetryEvent, 'event' | 'payload'>,
+    payload?: TelemetryEvent['payload'],
+  ): void {
```
nikosdouvlis, line 173: ❓ Why do we need multiple signatures here?

Another pattern I like that helps to eliminate the `|` unions for the final signature is simply typing the args for the final signature as `...args: unknown[]`. You only need to make the type wide enough to accept every possible type of the args, and `unknown[]` is always wide enough :)

packages/shared/src/telemetry/collector.ts
```diff
@@ -0,0 +1,300 @@
...
+      method: 'POST',
+      // TODO: We send an array here with that idea that we can eventually send multiple events.
+      body: JSON.stringify({
+        events: this.#buffer,
+      }),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    })
+      .catch(() => void 0)
```
nikosdouvlis, line 239: Efficiency!

packages/shared/src/telemetry/collector.ts
```diff
@@ -0,0 +1,300 @@
...
+      if (this.#pendingFlush) {
+        const cancel = typeof cancelIdleCallback !== 'undefined' ? cancelIdleCallback : clearTimeout;
+        cancel(this.#pendingFlush);
+      }
+      this.#flush();
+      return;
+    }
+
+    // If we have a pending flush, do nothing
+    if (this.#pendingFlush) return;
```
nikosdouvlis, line 214: How come eslint is not complaining about the missing braces here? 🤔

packages/shared/src/telemetry/collector.ts
```diff
@@ -0,0 +1,300 @@
...
+   */
+  sdk?: string;
+  /**
+   * SDK Version
+   */
+  sdkv?: string;
+  payload: Record<string, string | number | boolean>;
+};
+
+const DEFAULT_CONFIG: Partial<Required<TelemetryCollectorConfig>> = {
```
nikosdouvlis, line 102: Is there any specific reason for using Partial and Required here?

packages/shared/src/telemetry/events/method-called.ts
```diff
@@ -0,0 +1,21 @@
+const EVENT_METHOD_CALLED = 'METHOD_CALLED' as const;
+
+type EventMethodCalled = {
+  method: string;
+} & Record<string, string | number | boolean>;
+
+/**
+ * Fired when a helper method is called from a Clerk SDK.
+ */
+export function eventMethodCalled(
```
nikosdouvlis, line 10: Ditto

packages/shared/package.json
```diff
@@ -93,6 +95,7 @@
     "@types/glob-to-regexp": "0.4.1",
     "@types/js-cookie": "3.0.2",
     "@types/node": "^18.17.0",
+    "cross-fetch": "^4.0.0",
```
nikosdouvlis, line 98: Is this only needed for the tests?

nikosdouvlis: Now that we're using >= Node 18 locally and on CICD as well, can we use something like this https://github.com/jsdom/jsdom/issues/1724#issuecomment-1446858041 ?

packages/shared/src/telemetry/collector.ts
BRKalow: I think I prefer `CLERK_TELEMETRY_DISABLED` here, as the default is that it is enabled, so when we set it the value would always be false.

FWIW, I modeled it after how Next.js and Astro handle it:
* https://nextjs.org/telemetry#how-do-i-opt-out
* https://astro.build/telemetry/

packages/shared/src/telemetry/collector.ts
BRKalow: Not sure, do we have that rule enabled? 💭

packages/shared/src/telemetry/collector.ts
BRKalow: Probably not 🙃 good catch!

packages/shared/package.json
BRKalow: Yep, only for tests 👍

I can shift to that JSDOM workaround if you feel it's preferable. The use of `cross-fetch` seems simpler than the custom patched environment, but it's fine if we want to minimize dependencies.

packages/shared/src/telemetry/collector.ts
BRKalow: This let's us both pass a complete event, or use the more a more succinct form:
```ts
// full event
telemetry.record({ event: 'EVENT', payload: { ... } })

// short form
telemetry.record('EVENT', { ... })
```

In practice, the full event form is helpful when using the factory methods:
```ts
telemetry.record(eventComponentMounted(...))
```

> Another pattern I like that helps to eliminate the | unions for the final signature is simply typing the args for the final signature as ...args: unknown[]. You only need to make the type wide enough to accept every possible type of the args, and unknown[] is always wide enough :)

Is there a reason we might want to eliminate the union in the final signature? 🤔 I don't necessarily see a benefit to making that type wider in this scenario.

packages/shared/src/telemetry/events/component-mounted.ts
```diff
@@ -0,0 +1,22 @@
+const EVENT_COMPONENT_MOUNTED = 'COMPONENT_MOUNTED' as const;
+
+type EventComponentMounted = {
```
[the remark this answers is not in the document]
BRKalow: Thanks for the feedback! I initially refrained from creating a type like you've outlined to keep the implementation simple. When using the helper, implementors would get type errors from the call to `telemetry.record()` if the shape isn't correct.

The extra layer of helper types here creates some complexity and cognitive overhead that I was trying to avoid, but let me take another look!

packages/shared/src/telemetry/events/component-mounted.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function eventComponentMounted(
+  component: string,
+  props?: Record<string, unknown>,
+): { event: typeof EVENT_COMPONENT_MOUNTED; payload: EventComponentMounted } {
+  return {
+    event: EVENT_COMPONENT_MOUNTED,
+    payload: {
+      component,
+      appearanceProp: Boolean(props?.appearance),
```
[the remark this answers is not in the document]
BRKalow: Updated! It doesn't seem like we can access a theme name, but for the others I've added boolean values to help us understand usage a bit better.

packages/shared/src/telemetry/events/component-mounted.ts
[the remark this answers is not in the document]
BRKalow: Pushed an update. I've introduced a single type to be used as the return value from the factory methods. I think this strikes a good balance between too much typing boilerplate while still ensuring correctness when we are adding new events. Let me know what you think 👍

packages/shared/src/telemetry/events/method-called.ts
BRKalow: https://github.com/clerk/javascript/pull/2154/files#r1400989070

packages/shared/scripts/postinstall.mjs
```diff
@@ -0,0 +1,73 @@
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+
+// If we make significant changes to how telemetry is collected in the future, bump this version.
+const TELEMETRY_NOTICE_VERSION = '1';
+
+function telemetryNotice() {
+  console.log(`Attention: Clerk now collects telemetry data from its SDKs when connected to development instances.`);
```
jescalan, line 9:
```suggestion
  console.log(`Attention: Clerk collects anonymous telemetry data from its SDKs when connected to development instances.`);
```

jescalan (approved):
Minor change suggestion, feel free to reject

packages/shared/src/telemetry/collector.ts
nikosdouvlis: Ah nice, didn't know about the Next/Astro vars. That makes sense! Resolved.

packages/shared/src/telemetry/collector.ts
nikosdouvlis: Interesting, let me check when we and why we removed this one, probably by accident.

nikosdouvlis: Marking this as resolved as it is unrelated to this PR. I posted more on slack

packages/shared/src/telemetry/collector.ts
nikosdouvlis: Is there a reason we want to support the brief form? 
I prefer to limit the users of this API to the events created in the shared package for improved consistency, easier search replace, navigation, intellisense etc.

packages/shared/package.json
nikosdouvlis: No real preference here, both are good approaches

nikosdouvlis (approved):

packages/shared/src/telemetry/collector.ts
BRKalow: yeah that's reasonable! I'll simplify this. Thanks for the feedback 🙂

PR #1980, opened 2023-10-30
feat(backend): Also export token-related errors

flawnn:
## Checklist

- [x] `npm test` runs as expected.
- [x] `npm run build` runs as expected.

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [x] other: Developer Experience

## Packages affected

- [x] `@clerk/backend`
- [ ] `@clerk/chrome-extension`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

PR #2189, opened 2023-11-22
chore(backend): Limit exposed token verification related errors

tmilewski:
## Description

Removes both `TokenVerificationErrorAction` & `TokenVerificationErrorCode` from `@clerk/backend` exports as both are internal.

It also happens to support @flawnn's work/use-case in https://github.com/clerk/javascript/pull/1980, once we backport to `release/v4`.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [x] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [x] `@clerk/remix`
- [x] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/.changeset/smooth-baboons-shake.md
+++ b/.changeset/smooth-baboons-shake.md
@@ -1,0 +1,8 @@
+---
+'@clerk/clerk-sdk-node': major
+'@clerk/backend': major
+'@clerk/nextjs': major
+'@clerk/remix': major
+---
+
+Limit TokenVerificationError exports to TokenVerificationError and TokenVerificationErrorReason
--- a/packages/backend/src/exports.test.ts
+++ b/packages/backend/src/exports.test.ts
@@ -33,8 +33,6 @@ export default (QUnit: QUnit) => {
         'SignInToken',
         'Token',
         'TokenVerificationError',
-        'TokenVerificationErrorAction',
-        'TokenVerificationErrorCode',
         'TokenVerificationErrorReason',
         'User',
         'Verification',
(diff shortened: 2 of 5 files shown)


dimkl (commented):
❓ How is dropping those 2 exports going to resolve the use case of user of catching and handling those errors? I think i got confused. 😓

tmilewski:
@dimkl It doesn't. A [comment in the other PR](https://github.com/clerk/javascript/pull/1980#pullrequestreview-1705738246) suggested we wanted to keep some of these exports internal. This PR aims to narrow the scope of the exports on `main`. When backporting this to `release/v4`, it will enable access only to these narrowed exports which aren't currently exported at all.

dimkl:
I would suggest we close this PR because in the context of the Public APIs RFC we are going to change the exports of the `@clerk/backend` and probably export a `@clerk/backend/errors` subpath with the errors that we want to expose.
Also, I would suggest we keep those errors exported (from the `/errors` subpath) since the customer use case seems valid.

nikosdouvlis (approved):

PR #2275, opened 2023-12-06
chore(repo): Improve integration tests docs

LekoArts:
## Description

This improves our guide for integration testing by making it more structured, concise, and additional information.

[Read the preview](https://github.com/clerk/javascript/blob/lekoarts/sdk-1046-write-public-docs-for-e2e-tests/integration/README.md)

This PR also updates our contributing guide to point to this integration testing suite.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [x] `build/tooling/chore`

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -37,6 +37,7 @@ The current monorepo setup is based on:
 - [GitHub Actions](https://docs.github.com/en/actions), used for quality checks and automated release orchestration.
 - [Yalc](https://github.com/wclr/yalc), used for to publish packages locally and test them in other local projects.
 - [Jest](https://jestjs.io/), used for running unit tests.
+- [QUnit](https://qunitjs.com/), used for tests inside `@clerk/backend`.
 - [Playwright](https://playwright.dev/), used for running the [integration](../integration/) test suite.
 
 All packages of the monorepo are inside [packages](../packages). For package specific details on installation, architecture and usage, you can refer to the package's README file.
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -4,7 +4,7 @@ When contributing to this repository, please first discuss the change you wish t
 Please note we have a [code of conduct](https://github.com/clerk/javascript/blob/main/docs/CODE_OF_CONDUCT.md), please follow it in all your interactions with the project.
 
 <details open="open">
-<summary><strong>TABLE OF CONTENTS</strong></summary>
+<summary><strong>Table of contents</strong></summary>
 
 - [Contributing guide](#contributing-guide)
   - [Developing locally](#developing-locally)
@@ -12,6 +12,7 @@ Please note we have a [code of conduct](https://github.com/clerk/javascript/blob
     - [Prerequisites](#prerequisites)
     - [Setting up your local environment](#setting-up-your-local-environment)
     - [Documenting your changes](#documenting-your-changes)
+    - [Writing tests](#writing-tests)
   - [Opening a Pull Request](#opening-a-pull-request)
     - [Changesets](#changesets)
     - [Commit messages](#commit-messages)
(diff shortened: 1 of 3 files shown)


docs/CONTRIBUTING.md
```diff
@@ -35,6 +36,8 @@ The current monorepo setup is based on:
 - [Changesets](https://github.com/changesets/changesets), used for package versioning, publishing and changelog generation.
 - [GitHub Actions](https://docs.github.com/en/actions), used for quality checks and automated release orchestration.
 - [Yalc](https://github.com/wclr/yalc), used for to publish packages locally and test them in other local projects.
+- [Jest](https://jestjs.io/), used for running unit tests.
```
dimkl, line 39: We should probably add QUnit too (its being used in the @clerk/backend package)

docs/CONTRIBUTING.md
```diff
@@ -60,39 +63,53 @@ To set up your development environment, please follow these steps:
    git clone https://github.com/clerk/javascript
    ```
 
-2. Install the dependencies. We're using npm workspaces, so you **should always run `npm install` from the root of the monorepo**, as it will install dependencies for all the packages:
+1. Install the dependencies. We're using npm workspaces, so you **should always run `npm install` from the root of the monorepo**, as it will install dependencies for all the packages:
```
dimkl, line 66: ❓ shouldn't this be 2? The 1st item is the clone repo step above.

docs/CONTRIBUTING.md
```diff
@@ -60,39 +63,53 @@ To set up your development environment, please follow these steps:
...
 
    ```sh
    cd javascript
    npm install
    ```
 
-3. Build all the packages in the monorepo by running:
+1. Build all the packages in the monorepo by running:
```
dimkl, line 73: Ditto

integration/README.md
```diff
@@ -1,193 +1,460 @@
...
-### Env variables
+## Prerequisites
 
-The integration suite uses env variables to switch between Clerk instances or use 3rd party services to access emails sent to complete auth flows like signing in with email code or email links. Even though not all env variables are needed to run specific tests, we recommend setting them all at the beginning to avoid confusion later.
+Before you begin writing tests, you should already have:
 
-Note that these environment variables will be available to the tests via `process.env.INTEGRATION_INSTANCE_KEYS` on our CICD.
+- Followed the setup instructions in the [contributing guide](../docs/CONTRIBUTING.md)
+- Access to Clerk's 1Password
+- Access to the **Integration testing** organization on Clerk (it owns the Clerk instances used in the tests)
```
dimkl, line 19: 🤔 I think this is not mandatory. They could use their own instances, but they have to use the same settings. We could provide how many instances we are using and their settings as a way to avoid the need of having access to Clerk 1Password.

docs/CONTRIBUTING.md
LekoArts: You can use `1.` for everything, then you don't have to manually number things and change numbers if you add items in between: https://www.markdownguide.org/basic-syntax/#ordered-lists

integration/README.md
LekoArts: Right now the target audience is other Clerk employees since we rarely if ever have contributions from other people that would need an integration test. Thus they have these requirements for now - we can change that once we see people contributing more

PR #2299, opened 2023-12-08
chore(clerk-js,clerk-react,shared): Add hook to assert the presence of ClerkProvider [SDK-1043]

tmilewski:
## Description

Adds and implements `useAssertWrappedByClerkProvider` in order to assert that `ClerkProvider` is present. The goal of this being that we'd like to provide end-users a better development experience.

This applies to all public-facing components and hooks.

The intention behind introducing a new context was to limit the amount of refactoring needed to support what we were looking to achieve. 

#### Current error messaging: 
`AuthContext not found`
#### New error messaging: 
`@clerk/clerk-react: useAuth must be used within the <ClerkProvider> component. Please see: https://clerk. com/docs/components/clerk-provider`

SDK-1043

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [x] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/.changeset/slow-bugs-exist.md
+++ b/.changeset/slow-bugs-exist.md
@@ -6,6 +6,6 @@
 
 Add `useAssertWrappedByClerkProvider` to internal code. If you use hooks like `useAuth` outside of the `<ClerkProvider />` context an error will be thrown. For example:
 
-	```shell
-	@clerk/clerk-react: useAuth must be used within the <ClerkProvider> component
-	```
+```shell
+@clerk/clerk-react: useAuth must be used within the <ClerkProvider> component
+```
--- a/.changeset/slow-bugs-exist.md
+++ b/.changeset/slow-bugs-exist.md
@@ -4,4 +4,8 @@
 '@clerk/clerk-react': patch
 ---
 
-Adds useAssertWrappedByClerkProvider to assert the presence of ClerkProvider
+Add `useAssertWrappedByClerkProvider` to internal code. If you use hooks like `useAuth` outside of the `<ClerkProvider />` context an error will be thrown. For example:
+
+	```shell
+	@clerk/clerk-react: useAuth must be used within the <ClerkProvider> component
+	```
--- a/.changeset/slow-bugs-exist.md
+++ b/.changeset/slow-bugs-exist.md
@@ -1,0 +1,7 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/shared': patch
+'@clerk/clerk-react': patch
+---
+
+Adds useAssertWrappedByClerkProvider to assert the presence of ClerkProvider
--- a/.changeset/slow-bugs-exist.md
+++ b/.changeset/slow-bugs-exist.md
@@ -7,5 +7,5 @@
 Add `useAssertWrappedByClerkProvider` to internal code. If you use hooks like `useAuth` outside of the `<ClerkProvider />` context an error will be thrown. For example:
 
 ```shell
-@clerk/clerk-react: useAuth must be used within the <ClerkProvider> component
+@clerk/clerk-react: useAuth can only be used within the <ClerkProvider /> component
 ```
--- a/packages/shared/src/error.ts
+++ b/packages/shared/src/error.ts
@@ -190,6 +190,7 @@ const DefaultMessages = Object.freeze({
   InvalidProxyUrlErrorMessage: `The proxyUrl passed to Clerk is invalid. The expected value for proxyUrl is an absolute URL or a relative path with a leading '/'. (key={{url}})`,
   InvalidPublishableKeyErrorMessage: `The publishableKey passed to Clerk is invalid. You can get your Publishable key at https://dashboard.clerk.com/last-active?path=api-keys. (key={{key}})`,
   MissingPublishableKeyErrorMessage: `Missing publishableKey. You can get your key at https://dashboard.clerk.com/last-active?path=api-keys.`,
+  MissingClerkProvider: `{{source}} must be wrapped within the <ClerkProvider> component. Please see: https://clerk.com/docs/quickstarts/react#configure-clerk-provider`,
 });
 
 type MessageKeys = keyof typeof DefaultMessages;
@@ -213,6 +214,9 @@ export interface ErrorThrower {
   throwInvalidProxyUrl(params: { url?: string }): never;
 
   throwMissingPublishableKeyError(): never;
+
+  throwMissingClerkProviderError(params: { source?: string }): never;
+
   throw(message: string): never;
 }
 
@@ -263,6 +267,10 @@ export function buildErrorThrower({ packageName, customMessages }: ErrorThrowerO
 
     throwMissingPublishableKeyError(): never {
       throw new Error(buildMessage(messages.MissingPublishableKeyErrorMessage));
+    },
+
+    throwMissingClerkProviderError(params: { source?: string }): never {
+      throw new Error(buildMessage(messages.MissingClerkProvider, params));
     },
 
     throw(message: string): never {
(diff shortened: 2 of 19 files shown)


.changeset/slow-bugs-exist.md
```diff
@@ -0,0 +1,7 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/shared': patch
+'@clerk/clerk-react': patch
+---
+
+Adds useAssertWrappedByClerkProvider to assert the presence of ClerkProvider
```
LekoArts, line 7:
````suggestion
Add `useAssertWrappedByClerkProvider` to internal code. If you use hooks like `useAuth` outside of the `<ClerkProvider />` context an error will be thrown. For example:
```shell
@clerk/clerk-react: useAuth must be used within the <ClerkProvider> component
```
````

File: packages/shared/src/error.ts [hunk 2]
```diff
@@ -190,6 +190,7 @@ const DefaultMessages = Object.freeze({
   InvalidProxyUrlErrorMessage: `The proxyUrl passed to Clerk is invalid. The expected value for proxyUrl is an absolute URL or a relative path with a leading '/'. (key={{url}})`,
   InvalidPublishableKeyErrorMessage: `The publishableKey passed to Clerk is invalid. You can get your Publishable key at https://dashboard.clerk.com/last-active?path=api-keys. (key={{key}})`,
   MissingPublishableKeyErrorMessage: `Missing publishableKey. You can get your key at https://dashboard.clerk.com/last-active?path=api-keys.`,
+  MissingClerkProvider: `{{source}} must be wrapped within the <ClerkProvider> component. Please see: https://clerk.com/docs/quickstarts/react#configure-clerk-provider`,
```
LekoArts, line 193: The quickstart might change in the future so this is brittle to link to. We should link to the reference page. If that lacks details we need for this error to be useful, we should PR to the Clerk docs :)

Also removed the "see" to cater for people who have vision impairment / who are blind. (minor thing)

```suggestion
  MissingClerkProvider: `{{source}} must be wrapped within the <ClerkProvider /> component. Learn more: https://clerk.com/docs/components/clerk-provider`,
```

packages/shared/src/react/contexts.tsx
```diff
@@ -50,6 +50,25 @@ const OrganizationProvider = ({
...
+
+  if (!assertionCtx) {
+    if (typeof displayNameOrCustomHandler === 'function') {
+      displayNameOrCustomHandler();
+      return;
+    }
+
+    throw new Error(
+      `${displayNameOrCustomHandler || 'Clerk components'} must be wrapped within the <ClerkProvider> component.
+      Please see: https://clerk.com/docs/components/clerk-provider`,
```
LekoArts, line 67: Similar suggestions as above.

Also, I didn't test run this but I removed the indentation on the second line, from my experience this would otherwise be indented.

```suggestion
      `${displayNameOrCustomHandler || 'Clerk components'} must be wrapped within the <ClerkProvider /> component.
Learn more: https://clerk.com/docs/components/clerk-provider`,
```

LekoArts (changes requested):

tmilewski:
> ❓ Instead of adding another Context eg `ClerkProviderAssertionContext ` why not use `useIsomorphicClerkContext` or `useClientContext` and apply the same pattern as other hooks?
> 
> ```
>   const ctx = useClientContext();
>   assertContextExists(ctx, ClientContext);
> ```
> 
> The `assertContextExists` already exists in the shared package.

@dimkl So, the interesting thing about that example is that, if `ClientContext` doesn't exist, it'll never get to `assertContextExists` as it throws, by default, in `useClientContext`.

That should probably be fixed up, TBD. 👍 

Happy to use `ClerkInstanceContext`/`useIsomorphicClerkContext`, but we'll have to use the `useCtxWithoutGuarantee` version of the `createContextAndHook` so as to not throw by default within the `useCtx` the created hook, allowing us to throw an appropriate error. 

I'll fix up the existing usage mentioned above with this updated approach.

packages/shared/src/react/contexts.tsx
```diff
@@ -50,6 +50,23 @@ const OrganizationProvider = ({
   );
 };
 
+function useAssertWrappedByClerkProvider(displayNameOrCustomHandler?: string | (() => void)): void {
+  const ctx = React.useContext(ClerkInstanceContext);
+
+  if (!ctx) {
+    if (typeof displayNameOrCustomHandler === 'function') {
```
dimkl, line 57: 🔧 Since there is currently no usage of `displayNameOrCustomHandler` as a function i would omit this. Based on this PR there is no case of passing a function as a parameter.

packages/shared/src/react/contexts.tsx
```diff
@@ -50,6 +50,23 @@ const OrganizationProvider = ({
...
+      displayNameOrCustomHandler();
+      return;
+    }
+
+    throw new Error(
+      `${
+        displayNameOrCustomHandler || 'Clerk components'
+      } must be wrapped within the <ClerkProvider /> component. Learn more: https://clerk.com/docs/components/clerk-provider`,
+    );
+  }
```
dimkl, line 67: 🔧 i would expect to have a displayName. Could we make the `displayName` param to be required and remove the `'Clerk components'` fallback?

dimkl (approved):
As part of another PR, I think we also need to consolidate those:
- https://github.com/clerk/javascript/blob/3abb91f9f31acfe3ea1c4b06bfaa4d14df0e51a2/packages/clerk-js/src/ui/contexts/utils.ts#L11
- https://github.com/clerk/javascript/blob/3abb91f9f31acfe3ea1c4b06bfaa4d14df0e51a2/packages/clerk-js/src/ui.retheme/contexts/utils.ts#L11

packages/shared/src/react/contexts.tsx
tmilewski: @dimkl It's used here: https://github.com/clerk/javascript/pull/2299/files#diff-f023857f9a8061e1b588df1ac38d27994ca07359dbf77a99b53bc44fc3f82fda

packages/shared/src/error.ts
```diff
@@ -190,6 +190,7 @@ const DefaultMessages = Object.freeze({
   InvalidProxyUrlErrorMessage: `The proxyUrl passed to Clerk is invalid. The expected value for proxyUrl is an absolute URL or a relative path with a leading '/'. (key={{url}})`,
   InvalidPublishableKeyErrorMessage: `The publishableKey passed to Clerk is invalid. You can get your Publishable key at https://dashboard.clerk.com/last-active?path=api-keys. (key={{key}})`,
   MissingPublishableKeyErrorMessage: `Missing publishableKey. You can get your key at https://dashboard.clerk.com/last-active?path=api-keys.`,
+  MissingClerkProvider: `{{source}} must be wrapped within the <ClerkProvider /> component. Learn more: https://clerk.com/docs/components/clerk-provider`,
```
BRKalow, line 193:
```suggestion
  MissingClerkProvider: `{{source}} can only be used within the <ClerkProvider /> component. Learn more: https://clerk.com/docs/components/clerk-provider`,
```

.changeset/slow-bugs-exist.md
```diff
@@ -0,0 +1,11 @@
...
+'@clerk/clerk-js': patch
+'@clerk/shared': patch
+'@clerk/clerk-react': patch
+---
+
+Add `useAssertWrappedByClerkProvider` to internal code. If you use hooks like `useAuth` outside of the `<ClerkProvider />` context an error will be thrown. For example:
+
+	```shell
+	@clerk/clerk-react: useAuth must be used within the <ClerkProvider> component
+	```
```
LekoArts, line 11: Nit: Remove the indentation

````suggestion
```shell
@clerk/clerk-react: useAuth must be used within the <ClerkProvider> component
```
````

LekoArts (approved):
LGTM

PR #2290, opened 2023-12-08
feat(clerk-js): Retheme profile cards and scroll boxes

desiprisg:
## Checklist

- [ ] `npm test` runs as expected.
- [ ] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/chrome-extension`
- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

PR #2312, opened 2023-12-11
fix(clerk-react): Fix build process not catching type errors [SDK-1065]

tmilewski:
## Description

`tsup` doesn't reliably handle type-checking, thus we must continue to rely on `tsc`.

SDK-1065

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/types/src/localization.retheme.ts
+++ b/packages/types/src/localization.retheme.ts
@@ -271,6 +271,8 @@ type _LocalizationResource = {
     start: {
       headerTitle__account: LocalizationValue;
       headerTitle__security: LocalizationValue;
+      headerSubtitle__account: LocalizationValue;
+      headerSubtitle__security: LocalizationValue;
       profileSection: {
         title: LocalizationValue;
       };
@@ -554,6 +556,8 @@ type _LocalizationResource = {
     start: {
       headerTitle__members: LocalizationValue;
       headerTitle__settings: LocalizationValue;
+      headerSubtitle__members: LocalizationValue;
+      headerSubtitle__settings: LocalizationValue;
     };
     profilePage: {
       title: LocalizationValue;
--- a/packages/types/src/localization.ts
+++ b/packages/types/src/localization.ts
@@ -264,6 +264,10 @@ type _LocalizationResource = {
     formButtonPrimary__continue: LocalizationValue;
     formButtonPrimary__finish: LocalizationValue;
     formButtonReset: LocalizationValue;
+    navbar: {
+      title: LocalizationValue;
+      description: LocalizationValue;
+    };
     start: {
       headerTitle__account: LocalizationValue;
       headerTitle__security: LocalizationValue;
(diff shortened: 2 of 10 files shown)


panteliselef (approved):

dimkl:
@tmilewski We should probably do this for the rest of the packages:
- expo
- remix
- nextjs
- types
- shared
- backend
....
Could you take a look before proceeding with only this one? Also i would suggest we use `tsc` on Success and put `emitDeclarationsOnly: true` in tsconfig.json.

dimkl (approved):

packages/types/src/localization.retheme.ts
dimkl, line 1: 🔧 I think that those keys are dropped in Retheme project as part of https://github.com/clerk/javascript/pull/2290
Is there an error somewhere that indicated to make these changes?

packages/types/src/organizationMembership.ts
```diff
@@ -8,6 +8,11 @@ interface Base {
   role: string;
 }
 
+interface Placeholder {
+  permission: unknown;
+  role: unknown;
+}
+
```
dimkl, line 15: ❓ isn't this change added in https://github.com/clerk/javascript/pull/2314 ?

packages/types/src/localization.ts
dimkl, line 1: 🔧 Those keys are only added in Retheme project and they are supposed to exist in `packages/types/src/localization.retheme.ts` and not here. We should remove them from this file.

packages/types/src/organizationMembership.ts
tmilewski: It is, it was the only way I could get this to build and test, as it wasn't available at the time. My plan was to rebase everything and weed it out once it came time.

packages/types/src/localization.retheme.ts
tmilewski: Yes, it was giving me errors when running `tsc`. If they're supposed to be dropped then I can look into it more.

nikosdouvlis (commented):
Folks, are we sure this is the case? Are there any open issues I can read for more context?

I tested tsup locally with a few isolated scenarios and the build fails as expected.

In any case, let's wait before we consider merging this as we are making too many build-related changes across the codebase for v5.

tmilewski:
@nikosdouvlis I was able to get it to consistently succeed/fail in various scenarios. That said, I agree that we should hold, and I'm going to close for the time being.

PR #2328, opened 2023-12-12
chore(*): Improve @clerk/clerk-react DX

dimkl:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [x] `@clerk/chrome-extension`
- [ ] `@clerk/clerk-js`
- [x] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [x] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [x] `@clerk/nextjs`
- [x] `@clerk/clerk-react`
- [x] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`


.changeset/mighty-bulldogs-try.md
```diff
@@ -0,0 +1,33 @@
+---
+'gatsby-plugin-clerk': major
+'@clerk/chrome-extension': major
+'@clerk/nextjs': major
+'@clerk/clerk-react': major
+'@clerk/remix': major
+'@clerk/clerk-expo': major
+---
+
+- Introduce `@clerk/clerk-react/errors` and `@clerk/clerk-react/internal` subpath exports to expose some internal utilities. Eg
```
SokratisVidros, line 10: 💯

packages/react/src/components/withClerk.tsx
```diff
@@ -48,5 +45,5 @@ export const WithClerk: React.FC<{
     return null;
   }
 
-  return <LoadedGuarantee>{children(clerk as unknown as LoadedClerk)}</LoadedGuarantee>;
```
SokratisVidros, line 51: ❓ Can you please elaborate what replaces the loaded guarantee?

packages/react/src/errors/errors.ts
```diff
@@ -0,0 +1,7 @@
+export {
+  isClerkAPIResponseError,
+  isEmailLinkError,
+  isKnownError,
+  isMetamaskError,
+  EmailLinkErrorCode,
```
SokratisVidros, line 6: Do we need both isEmailLinkError and EmailLinkErrorCode?

SokratisVidros: How about exporting ClerkRuntimeError and ClerkAPIResponseError?

packages/react/src/internal.ts
```diff
@@ -0,0 +1,14 @@
+import type { ErrorThrowerOptions } from '@clerk/shared/error';
```
SokratisVidros, line 1: 💯

SokratisVidros (approved):

packages/react/src/errors/errors.ts
dimkl: >Do we need both isEmailLinkError and EmailLinkErrorCode?
I think we should keep only the `isEmailLinkError` 
> How about exporting ClerkRuntimeError and ClerkAPIResponseError?
I will :)

.changeset/mighty-bulldogs-try.md
```diff
@@ -0,0 +1,34 @@
...
+
+    // Before
+    import { isClerkAPIResponseError, isEmailLinkError, isKnownError, isMetamaskError } from '@clerk/clerk-react';
+    // After
+    import { isClerkAPIResponseError, isEmailLinkError, isKnownError, isMetamaskError } from '@clerk/clerk-react/error';
+
+    // Before
+    import { MultisessionAppSupport } from '@clerk/clerk-react';
+    // After
+    import { MultisessionAppSupport } from '@clerk/clerk-react/internal';
```
BRKalow, line 25: ❓ why are we moving this to `/internal`?

dimkl: > This is a component that handles the intermediate “undefined” state for multisession apps by unmounting and remounting the components during the session switch (setActive call) in order to solve theoretical edge-cases that can arise while switching sessions. It is not documented and as far as I know, no one has actually used it, however, we are currently using it in our dashboard. I suggest we either remove it or move it under /internal until we have more concrete use cases for this.

ref: `SDK Public API (v5)` RFC

BRKalow: Makes sense! I was going to question why we were re-exporting this from other SDKs, but with the changes in 99207b29de2fe868e8b0ad200dc65c276df609bf this seems fine.

tmilewski:
@dimkl Rebase `main` and CI should be good.

packages/react/src/components/withClerk.tsx
nikosdouvlis: Nothing used the LoadedGuarantee (or "structure context") since the v4 hook change. We kept it around in case we wanted to use it to throw better errors in cases where the Clerk hooks were used outside the Clerk provider but ended up doing that by simply checking wether the ClerkProvider context exists or not.

Extra context: The LoadedGuarantee (or "structure context") used to wrap all SignedIn and SignedOut components and the older hooks could only be used when nested under these components - this is no longer the case as all hooks return `isLoaded`

nikosdouvlis: cc @dimkl

jonahallibone:
```
useSafeLayoutEffect(() => {
    window.__unstable__onAfterSetActive = () => {
      // Re-run the middleware every time there auth state changes.
      // This enables complete control from a centralised place (NextJS middleware),
      // as we will invoke it every time the client-side auth state changes, eg: signing-out, switching orgs, etc.\
      if (__unstable_invokeMiddlewareOnAuthStateChange) {
        void push(window.location.href);
      }
    };
  }, []);
```

This line seems to cause a race condition with the example in the docs.

Following this example [here](https://clerk.com/docs/custom-flows/use-sign-up#use-sign-in)
```
if (result.status === "complete") {
  console.log(result);
  await setActive({ session: result.createdSessionId });
  router.push("/")
}
```
When I do this, I `router.push("/")` but immediately get sent back to the previous URL. The only way around this is to pop-out of nextjs and using `window.location.href = newUrl`.

While this works, it's definitely not ideal. Is there a new way to navigate after session changes or should I continue to use `window.location`?

PR #2347, opened 2023-12-14
feat(nextjs): Various DX improvements

nikosdouvlis:
## Description
Drop support for NextJS v12: v12 was released on 26 Oct 2021. Support for security updates stopped on 21 Nov 2022.

Drop support for NextJS <13.0.4: Various header-related bugs were introduced with the 12.1.0, 12.2.0, 13.0.1, 13.0.2, 13.0.3 NextJS releases which are now fixed since next@^13.0.4. We will be dropping support for these problematic versions in order to greatly reduce complexity in our codebase.

Drop support for NextJS < 14.0.3 because of the issues described here: https://github.com/clerk/javascript/issues/1436.

Deprecate `authMiddleware` in favor of `clerkMiddleware`. For more information, see https://clerk.com/docs/upgrade-guides/v5-introduction

Move the server-side APIs from `@clerk/nextjs` to the `@clerk/nextjs/server` module: `WebhookEventType`, `WebhookEvent`, `verifyToken`, `redirectToSignIn`, `auth`, `buildClerkProps`, `clerkClient`, `currentUser`, `getAuth`, `redirectToSignUp` and `authMiddleware`. For more information, see https://clerk.com/docs/upgrade-guides/v5-introduction

Drop unused AuthStatus search param constant

SDK-787

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [x] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`


SokratisVidros (approved):

packages/nextjs/src/server/index.ts
```diff
@@ -1,7 +1,15 @@
-export * from './clerkClient';
+/**
+ * Generic exports
+ */
+export { verifyToken, createClerkClient } from '@clerk/backend';
```
dimkl, line 4: After [this PR](https://github.com/clerk/javascript/pull/2365) is being merged this line will be the same as the public API of the ( eg `export * from @clerk/backend`)

packages/nextjs/src/server/index.ts
```diff
@@ -1,7 +1,15 @@
...
+export type { WebhookEvent, WebhookEventType } from '@clerk/backend';
```
dimkl, line 5: After [this PR](https://github.com/clerk/javascript/pull/2365) only the following exported types will be available from the `@clerk/backend` : Organization, Session, User, WebhookEvent, WebhookEventType
We should export all those.

packages/nextjs/tsconfig.json
```diff
@@ -19,7 +19,8 @@
     "skipLibCheck": true,
     "sourceMap": false,
     "strict": true,
-    "target": "ES2020"
+    "target": "ES2020",
+    "rootDir": "src"
```
dimkl, line 23: ❓ Should we also this to all the other packages of the monorepo? Is this change resolving some issue or something?

playground/nextjs/pages/session-examples/index.tsx
dimkl, line 1: 🤔 I think this change is already applied in main by https://github.com/clerk/javascript/pull/2357

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -132,6 +132,10 @@ export interface AuthMiddleware {
   (params?: AuthMiddlewareParams): NextMiddleware;
 }
 
+/**
+ * @deprecated Use `clerkMiddleware` instead.
+ * Migration guide: https://clerk.com/docs/upgrade-guides/v5-introduction
+ */
```
dimkl, line 138: 🔧  let's add the deprecation when `clerkMiddleware` is added and also add a deprecation warning to the `authMiddleware` using the `@clerk/shared/deprecated` in v5 version

packages/nextjs/src/server/__tests__/__snapshots__/exports.test.ts.snap
dimkl, line 1: 🤔 There may be some conflicts with backend DX improvements PRs since these cleanups are also applied.

.changeset/grumpy-drinks-turn.md
dimkl, line 1: Can we reserve the `https://clerk.com/docs/upgrade-guides/v5-introduction`  (without being searchable) and add a TODO or BETA text as a placeholder for those that may search the CHANGELOG and try to access the link?

dimkl (approved):
🚀 🚀 🚀 🚀 🚀 🚀 🚀

packages/nextjs/src/server/index.ts
nikosdouvlis: I feel its still easier for us to control the public API of the high level packages if we're explicit - we're not adding new APIs that often so it shouldn't be an issue. Happy to revise of course!

packages/nextjs/src/server/index.ts
nikosdouvlis: What about the API types?

nikosdouvlis: Adding a TODO, I will update the types in a 2nd PR right after

packages/nextjs/tsconfig.json
nikosdouvlis: No reason to :)

packages/nextjs/src/server/authMiddleware.ts
nikosdouvlis: I don't think we need to wait, `clerkMiddleware` will be released in v5 so both this warning and the new middleware will be released at the same time :)

In any case, clerkMiddleware will be introduced shortly

dimkl: I think it's more related to add the deprecation in the same PR or in a PR later than when that change is applied. That's why. But I'm fine with it.

nikosdouvlis: Ah, okay just now got what you mean. Yes that makes more sense :)

Issue #1981, opened 2023-10-30
<SignIn> Component Not Rendering After Signing Out When Component Has Appearance Props Added


brettabamonte:
### Preliminary Checks

- [X] I have reviewed the documentation: https://clerk.com/docs

- [X] I have searched for existing issues: https://github.com/clerkinc/javascript/issues

- [X] I have not already reached out to Clerk support via email or Discord (if you have, no need to open an issue here)

- [X] This issue is not a question, general help request, or anything other than a bug report directly related to Clerk. Please ask questions in our Discord community: https://clerk.com/discord.

### Reproduction / Replay Link

https://github.com/brettabamonte/Clerk-Minimal_Reproduction-SignIn_Component/tree/main

### Publishable key

pk_test_Y2hpZWYtc2N1bHBpbi0xMi5jbGVyay5hY2NvdW50cy5kZXYk

### Description

**Steps to reproduce:**

See steps and screenshots in linked git repository README.md (https://github.com/brettabamonte/Clerk-Minimal_Reproduction-SignIn_Component).

**Expected behavior:**
After clicking the <SignOutButton /> the user should be redirected to the sign in page with a <SignIn /> component displaying that has appearance props.

**Actual behavior:**
After clicking the <SignOutButton /> the user is redirected to the sign in page but the <SignIn /> component does not render when it has the appearance prop set. If the appearance prop is set the component displays properly.

### Environment

```Shell
System:
    OS: macOS 14.0
    CPU: (8) arm64 Apple M2
    Memory: 238.50 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.18.1 - /usr/local/bin/node
    npm: 9.8.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 118.0.5993.117
    Safari: 17.0
  npmPackages:
    @clerk/nextjs: ^4.26.1 => 4.26.1 
    @types/node: 20.1.4 => 20.1.4 
    @types/react: 18.2.6 => 18.2.6 
    @types/react-dom: 18.2.4 => 18.2.4 
    autoprefixer: 10.4.14 => 10.4.14 
    classnames: ^2.3.2 => 2.3.2 
    next: ^14.0.0 => 14.0.0 
    postcss: 8.4.23 => 8.4.23 
    react: 18.2.0 => 18.2.0 
    react-dom: 18.2.0 => 18.2.0 
    tailwindcss: 3.3.2 => 3.3.2 
[...]

PR #2348, opened 2023-12-14
feat(clerk-react): Replace `signOutCallback` with `redirectUrl` in SignOutButton

LekoArts:
## Description

Replace the `signOutCallback` prop on the `<SignOutButton />` with `redirectUrl`. This aligns the API surface with other UI components provided by `@clerk/clerk-react`.

If you previously used the `signOutCallback` prop to navigate to another page, you can migrate as shown below.

Before:

```jsx
import { SignOutButton } from "@clerk/clerk-react";

export const Signout = () => {
  return (
    <SignOutButton signOutCallback={() => { window.location.href = "/your-path" }}>
      <button>Sign Out</button>
    </SignOutButton>
  )
}
```
After:
```jsx
import { SignOutButton } from "@clerk/clerk-react";

export const Signout = () => {
  return (
    <SignOutButton redirectUrl="/your-path">
      <button>Sign Out</button>
    </SignOutButton>
  )
}
```

Related to https://github.com/clerk/javascript/issues/1981 which uncovered a bug. With the `redirectUrl` people could more easily (and aligned with our other APIs) decide where to redirect.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [x] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`


packages/react/src/components/SignOutButton.tsx
```diff
@@ -1,25 +1,29 @@
...
-  signOutCallback?: SignOutCallback;
+  redirectUrl?: string;
   signOutOptions?: SignOutOptions;
   children?: React.ReactNode;
 };
 
 export const SignOutButton = withClerk(
   ({ clerk, children, ...props }: React.PropsWithChildren<WithClerkProp<SignOutButtonProps>>) => {
-    const { signOutCallback, signOutOptions, ...rest } = props;
+    const { redirectUrl = '/', signOutOptions, ...rest } = props;
```
LekoArts, line 16: Open to suggestions to moving this default to somewhere else (if necessary)

SokratisVidros (approved):

packages/react/src/components/SignOutButton.tsx
dimkl: 🤔 I would suggest we open a ticket and move all the defaults (we have defaults for `afterSignInUrl`, `afterSingUpUrl`) to a separate module to have all the default settings in a centralized place. For it seems fine to leave it there.

packages/react/src/components/SignOutButton.tsx
```diff
@@ -1,25 +1,29 @@
...
 
     children = normalizeWithDefaultValue(children, 'Sign out');
     const child = assertSingleChild(children)('SignOutButton');
 
+    const navigateToRedirectUrl = () => {
+      return clerk.navigate(redirectUrl);
```
dimkl, line 22: 🤔  should we use this or `clerk.navigate(redirectUrl, { replace: true })`?

dimkl (approved):

packages/react/src/components/SignOutButton.tsx
LekoArts: Inside the `navigate` function:

```ts
const customNavigate =
  options?.replace && this.#options.routerReplace ? this.#options.routerReplace : this.#options.routerPush;
```

So I don't think this would do anything, right?

LekoArts: Ok, so this refers to push and replace in regards to the browser history. I'd want to push here, so code is fine as-is 👍

packages/react/src/components/SignOutButton.tsx
LekoArts: Done in SDK-1079

PR #2352, opened 2023-12-14
feat(backend): Support dev handshake when the redirect ends up tainted

BRKalow:
## Description

In cases where the handshake flow is triggered on fetch-initiated requests in development, we can end up in a scenario where the redirect becomes "tainted" and the origin is set to null. This treats all subsequent requests in the chain as cross origin, even if they aren't in practice.

To support this and allow the handshake to succeed, we detect the final redirect by looking for the presence of the dev browser cookie and a null origin. Note that this also requires credentials to be included with the request.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/backend/src/tokens/request.ts
+++ b/packages/backend/src/tokens/request.ts
@@ -175,7 +175,9 @@ ${err.getFullMessage()}`,
   async function authenticateRequestWithTokenInCookie() {
     const {
       derivedRequestUrl,
+      devBrowser,
       isSatellite,
+      origin,
       secFetchDest,
       signInUrl,
       clientUat: clientUatRaw,
@@ -185,11 +187,23 @@ ${err.getFullMessage()}`,
     const clientUat = parseInt(clientUatRaw || '', 10) || 0;
     const hasActiveClient = clientUat > 0;
     const hasSessionToken = !!sessionToken;
+    const isTaintedDevRedirect = !!devBrowser && origin === 'null';
 
     const isRequestEligibleForMultiDomainSync =
       isSatellite &&
       secFetchDest === 'document' &&
       !derivedRequestUrl.searchParams.has(constants.QueryParameters.ClerkSynced);
+
+    let headers = new Headers({});
+
+    /**
+     * In dev, the redirect chain for fetch requests eventually results in a "tainted redirect" in which the origin gets set to "null" and the request is treated as cross-origin, even if it's technically not.
+     * In order to allow the handshake to succeed in this edge case, we need to add the proper CORS headers to the response.
+     */
+    if (isTaintedDevRedirect) {
+      headers.set('Access-Control-Allow-Origin', 'null');
+      headers.set('Access-Control-Allow-Credentials', 'true');
+    }
 
     /**
      * If we have a handshakeToken, resolve the handshake and attempt to return a definitive signed in or signed out state.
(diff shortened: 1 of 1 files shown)


packages/backend/src/tokens/__tests__/request.test.ts
```diff
@@ -388,12 +388,13 @@ export default (QUnit: QUnit) => {
       assert.strictEqual(requestState.toAuth(), null);
     });
 
-    test('cookieToken: returns undefined when crossOriginReferrer in development and is satellite [6n]', async assert => {
+    test('cookieToken: returns signedIn when satellite but valid token and clientUat', async assert => {
```
BRKalow, line 391: This test name was inaccurate 🧐

packages/backend/src/tokens/__tests__/request.test.ts
```diff
@@ -299,6 +298,7 @@ export default (QUnit: QUnit) => {
         mockRequestWithCookies(
           {
             ...defaultHeaders,
+            'sec-fetch-dest': 'empty',
```
BRKalow, line 301: Now that we're defaulting to `sec-fetch-dest: document`, we need to explicitly set it to empty for this test case

packages/backend/src/tokens/request.ts
```diff
@@ -53,6 +53,25 @@ function assertSignInUrlFormatAndOrigin(_signInUrl: string, origin: string) {
...
 }
 
+/**
+ * Currently, a request is only eligible for a handshake if we can say it's *probably* a request for a document, not a fetch or some other exotic request.
+ * This heuristic should give us a reliable enough signal for browsers that support `Sec-Fetch-Dest` and for those that don't.
+ */
+function isRequestEligibleForHandshake(authenticateContext: { secFetchDest?: string; accept?: string }) {
+  const { accept, secFetchDest } = authenticateContext;
+
+  // NOTE: we could also check sec-fetch-mode === navigate here, but according to the spec, sec-fetch-dest: document should indicate that the request is the result of a user navigation.
```
BRKalow, line 63: Also, you can't overwrite `sec-fetch-mode`, so testing became difficult.

packages/backend/src/tokens/request.ts
```diff
@@ -159,6 +178,20 @@ ${err.getFullMessage()}`,
...
+    message: string,
+    headers?: Headers,
+  ) {
+    if (isRequestEligibleForHandshake(context)) {
+      // Right now the only usage of passing in different headers is for multi-domain sync, which redirects somewhere else.
+      // In the future if we want to decorate the handshake redirect with additional headers per call we need to tweak this logic.
+      return handshake(context, reason, message, headers ?? buildRedirectToHandshake());
+    }
+    return signedOut(context, reason, message, new Headers());
+  }
```
BRKalow, line 193: This felt like a reasonable abstraction, easy to tweak the handshake heuristic in one place going forward.

SokratisVidros (approved):

PR #2389, opened 2023-12-18
Refactor clerk/backend

nikosdouvlis:
## Description 

Review by PR

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`


packages/backend/src/tokens/authenticateContext.ts
dimkl, line 1: 🎉

packages/backend/src/tokens/types.ts
dimkl, line 1: 🔧 I would suggest we keep types close to their usage and use `types.ts` for generic types related to specific package. Eg move `RequestStateParams` to `tokens/authStatus.ts` or `tokens/request.ts`

packages/backend/src/tokens/clerkRequest.ts
```diff
@@ -0,0 +1,68 @@
...
+   * The `authMiddleware` uses the value of the available req.headers in order to construct
+   * and use the correct url internally. This url is then exposed as `experimental_clerkUrl`,
+   * intended to be used within `beforeAuth` and `afterAuth` if needed.
+   */
+  clerkUrl: URL;
+};
+
+class ClerkRequest extends Request {
+  clerkUrl: URL;
+  cookies: Map<string, string>;
```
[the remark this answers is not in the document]
nikosdouvlis: Done :)

packages/backend/src/util/__tests__/request.test.ts
```diff
@@ -1,135 +0,0 @@
...
-import { checkCrossOrigin } from '../request';
-
-export default (QUnit: QUnit) => {
-  const { module, test } = QUnit;
-
-  module('check cross-origin-referrer request utility', () => {
-    test('is not CO with IPv6', assert => {
-      const originURL = new URL('http://[::1]');
-      const host = new URL('http://[::1]').host;
-      assert.false(checkCrossOrigin({ originURL, host }));
```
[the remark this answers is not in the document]
nikosdouvlis: FYI, I've found no references in the codebase

packages/backend/src/tokens/types.ts
nikosdouvlis: I had to extract these to a new file to avoid circular dependencies between `authStatus` <> `request`, but let me take another look and see if I can avoid this

packages/backend/src/tokens/authStatus.ts
```diff
@@ -1,12 +1,10 @@
 import type { JwtPayload } from '@clerk/types';
 
 import type { TokenVerificationErrorReason } from '../errors';
-import type { SignedInAuthObject, SignedInAuthObjectOptions, SignedOutAuthObject } from './authObjects';
+import type { AuthenticateContext } from './authenticateContext';
+import type { SignedInAuthObject, SignedOutAuthObject } from './authObjects';
 import { signedInAuthObject, signedOutAuthObject } from './authObjects';
```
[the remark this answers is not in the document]
nikosdouvlis: Fixed

packages/backend/src/tokens/authenticateContext.ts
```diff
@@ -0,0 +1,88 @@
+import { constants } from '../constants';
```
nikosdouvlis, line 1: These classes will be moved to a different dir when I wrap up my work on clerkMiddleware

packages/backend/src/tokens/authenticateContext.ts
```diff
@@ -0,0 +1,88 @@
...
+
+interface AuthenticateContext extends AuthenticateContextInterface {}
+
+/**
+ * All data required to authenticate a request.
+ * This is the data we use to decide whether a request
+ * is in a signed in or signed out state or if we need
+ * to perform a handshake.
+ */
+class AuthenticateContext {
```
SokratisVidros, line 36: This naming feels weird as it looks like a React Context. Do we even need a class at this point? This is a very nit-picky comment, so you can ignore it for now, but I wanted to raise the point that the abstraction might not help as much as intended.

SokratisVidros: @nikosdouvlis Will we just pass a Request object afterall?

packages/backend/src/tokens/clerkRequest.ts
```diff
@@ -0,0 +1,68 @@
+import { parse as parseCookies } from 'cookie';
+
+import { constants } from '../constants';
+
+export type WithClerkUrl<T> = T & {
+  /**
```
SokratisVidros, line 6: ❓ Do we still need this in the handshake world?

packages/backend/src/tokens/authenticateContext.ts
SokratisVidros: BTW this is a great candidate for async local storage.

packages/backend/src/tokens/types.ts
SokratisVidros: ☝️

SokratisVidros (approved):

packages/backend/src/tokens/authenticateContext.ts
nikosdouvlis: > ... that the abstraction might not help as much as intended.
> BTW this is a great candidate for async local storage.

Let me reply to these two together, as my intention is for us to use `AuthenticateContext` or a similar struct with Async Local Storage, but only after we simplify type-handling in clerk/backend and we decide that the abstraction works for us

I will not refactor anything else to use `AuthenticateContext` as I've simplified the problematic types for now. Let's do a design session and talk more about it

packages/backend/src/tokens/clerkRequest.ts
nikosdouvlis: We still need it for `authenticateRequest` and `redirectToSignIn`/`protect` from within `clerkMiddleware` and `authMiddleware`

packages/backend/src/tokens/types.ts
nikosdouvlis: I need these for authenticateRequest and AuthenticateContext right now. Since we will continue refactoring the types and structs in clerk/backend after v5 as part of flex work, shall we fix this then?

packages/backend/src/tokens/clerkRequest.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  private decodeCookieValue(str: string) {
+    return str ? str.replace(/(%[0-9A-Z]{2})+/g, decodeURIComponent) : str;
+  }
+}
+
+export const createClerkRequest = (...args: ConstructorParameters<typeof ClerkRequest>): ClerkRequest => {
+  return new ClerkRequest(...args);
+};
```
BRKalow, line 66: ❓ What's the benefit of using a factory function here? If it's a straight pass-through to the constructor it feels unecessary.

packages/backend/src/tokens/authenticateContext.ts
```diff
@@ -0,0 +1,88 @@
+import { constants } from '../constants';
+import type { ClerkRequest } from './clerkRequest';
+import type { AuthenticateRequestOptions } from './types';
+
+interface AuthenticateContextInterface extends AuthenticateRequestOptions {
+  // header-based values
+  sessionTokenInHeader: string | undefined;
```
BRKalow, line 7: ❓ why do we use `string | undefined` instead of `?: string`

packages/backend/src/tokens/authenticateContext.ts
```diff
@@ -0,0 +1,88 @@
...
+  // handshake-related values
+  devBrowserToken: string | undefined;
+  handshakeToken: string | undefined;
+  // url derived from headers
+  clerkUrl: URL;
+  // cookie or header session token
+  sessionToken: string | undefined;
+}
+
+interface AuthenticateContext extends AuthenticateContextInterface {}
```
BRKalow, line 28: ❓ Why are we doing this?

packages/backend/src/tokens/__tests__/clerkRequest.test.ts
```diff
@@ -0,0 +1,116 @@
+import { createClerkRequest } from '../clerkRequest';
+
+export default (QUnit: QUnit) => {
+  const { module, test: it } = QUnit;
+
+  module('createClerkRequest', () => {
+    module.skip('cookies', () => {
```
BRKalow, line 7: ❓ why are these tests skipped?

packages/backend/src/tokens/types.ts
```diff
@@ -0,0 +1,14 @@
...
+  domain?: string;
+  isSatellite?: boolean;
+  proxyUrl?: string;
+  signInUrl?: string;
+  signUpUrl?: string;
+  afterSignInUrl?: string;
+  afterSignUpUrl?: string;
+};
+
+export type AuthenticateRequestOptions = RequestStateParams & VerifyTokenOptions;
```
BRKalow, line 14: Can we consolidate these two types, or are we using `RequestStateParams` independently somewhere?

BRKalow (commented):
Nice cleanup

packages/backend/src/tokens/clerkRequest.ts
nikosdouvlis: It's a pattern I like but don't have strong opinions about. It's a way to export a class as  type only while providing a factory so consumers can use `ClerkRequest` as a type but they are forced to initialize via `createClerkRequest`.

I find that this helps me better understand what is a type vs a constructor when going through code quickly, but as I said, no strong opinions here.

This is exported under /internals so we can iterate as much as we like as for now, it's a "private" API

packages/backend/src/tokens/authenticateContext.ts
nikosdouvlis: In general, the two variations are almost always interchangeable, with one semantic difference being that `string | undefined` would mean that the prop exists in the object but is undefined (aka `"something" in obj` would return `true`) while `?: string` means that it might now exist at all (`"something" in obj` might return false)

Here, we're initializing the values and even if some of them will be `undefined`, they do exist on `this`. Extremely nitpicky difference here so happy to change if you prefer the alternative syntax.

packages/backend/src/tokens/authenticateContext.ts
nikosdouvlis: We're merging the interface with the class so we don't have to redefine the interface's props in the class it self

packages/backend/src/tokens/__tests__/clerkRequest.test.ts
nikosdouvlis: Because I was working on them :) They are enabled again

packages/backend/src/tokens/types.ts
nikosdouvlis: Thanks for spotting that, done!

packages/backend/src/tokens/authenticateContext.ts
BRKalow: Makes sense :+1:. In that case, I believe we can get rid of the intermediary `AuthenticateContextInterface` type ([TS Playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPIAczAPYgM7IDeAUMsjNtgFzJ5hSgDmxAvscaJLIigMK6QAHmGQRhEEABMCGLLgIkyAIzhQadBiGZtiCADZw8BfuDEjFyFQC8AFAEoipMsigQwAVyghkYABbA8ADoVKCc2HQR5EQRkAF5kEAgAd2QTITB7dgRAoA)). 

This should work:

```typescript
interface AuthenticateContext extends AuthenticateRequestOptions {
// all the properties defined in AuthenticateContextInterface above...
}

class AuthenticateContext {
// ...
}
```

PR #2399, opened 2023-12-19
fix(clerk-js): Clear feedback when `undefined` is passed to `setError`

panteliselef:
## Description
We still have a lot of places in clerk-js where we call `setError(undefined)` instead of `clearFeedback`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/.changeset/thick-doors-smile.md
+++ b/.changeset/thick-doors-smile.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Bug fix: Clear feedback of input field if `setError` is called with undefined.
--- a/packages/clerk-js/src/ui/utils/useFormControl.ts
+++ b/packages/clerk-js/src/ui/utils/useFormControl.ts
@@ -110,6 +110,8 @@ export const useFormControl = <Id extends string>(
   const setError: FormControlState['setError'] = error => {
     if (error) {
       setFeedback({ message: translateError(error), type: 'error' });
+    } else {
+      clearFeedback();
     }
   };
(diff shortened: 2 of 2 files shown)


dimkl (commented):
❓ Searching for `setError()` in project i found that it's being called inside a promise `catch()` block in `packages/clerk-js/src/ui/hooks/useFetch.ts`. Is this intentional? Should we at least show a `console.error`?
I would suggest instead of calling `setError()` without args we could introduce a `clear()` method to unset the errors and the loading state or anything else related. By doing this, the intention of the action we want to do will be more clear.

panteliselef:
@dimkl I totally agree, this is my plan, but because this require a bit of refactoring and i don't want to mess too much with v4, shall we let this one be backported and make a new one with the proper refactoring ?

Btw you are referring to a different `setError`, this PR has to do with the one coming from `useFormControl`

dimkl:
@panteliselef let's proceed.

PR #2404, opened 2023-12-19
feat(nextjs): Introduce clerkMiddleware

nikosdouvlis:
### `clerkMiddleware` API:
```
export default clerkMiddleware(auth => {
  auth().protect();
  auth().protect({ permission: 'org:domains:delete' });
  auth().protect({ permission: 'org:domains:delete' }, { redirectUrl: 'https://clerk.com' });

  auth().redirectToSignIn();
  auth().redirectToSignIn({ returnBackUrl: 'https://clerk.com' });

  const { userId, redirectToSignIn } = auth();
  if (!userId) {
    return redirectToSignIn();
  }

  return NextResponse.next();
});
```

### `protect` API:
The `protect` API can be used to protect pages, route handlers, server actions and any other request that can be intercepted using the middleware. It can be used to handle both unauthenticated and unauthorized requests.

Calling `protect` without arguments will catch all unauthenticated requests, and depending on the context it's used in, it will either redirect to the sign-in URL (for page requests) or throw a `notFound` error (for route handlers and server actions). If a `redirectUrl` is provided, `protect` will always respect it and redirect to it.

Calling `protect` with the same arguments the `has` API accepts, will catch all unauthorized requests, and throw a `notFound` error (for route handlers and server actions). If a `redirectUrl` is provided, `protect` will always respect it and redirect to it.

SDK-647

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`


packages/backend/src/tokens/clerkUrl.ts
dimkl, line 1: 🙃 i would prefer if we move this functionality to `ClerkRequest` and move this in a separate file if things get a lot more complicated. Do you expect the `ClerkUrl` to grow a lot bigger in the future?

packages/nextjs/src/app-router/server/auth.ts
```diff
@@ -1,9 +1,10 @@
 import type { AuthObject } from '@clerk/backend/internal';
+import { notFound, redirect } from 'next/navigation';
```
dimkl, line 2: ❓ Are those exports supported in the minimum NextJS version defined in our package.json?

packages/nextjs/src/pages/__tests__/index.test.tsx
```diff
@@ -31,27 +31,6 @@ describe('ClerkProvider', () => {
...
-    });
-
-    it('proxyUrl + domain + isSatellite is not allowed', () => {
-      expectTypeOf({
-        ...defaultProps,
-        proxyUrl: 'test',
-        domain: 'test',
-        isSatellite: true,
-      }).not.toMatchTypeOf<ClerkProviderProps>();
-    });
```
dimkl, line 54: ❓ Did we loosen up the type of `MultiDomainProxy`?

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -0,0 +1,187 @@
...
+  (options?: ClerkMiddlewareOptions): NextMiddleware;
+  /**
+   * @example
+   * export default clerkMiddleware;
+   */
+  (request: NextMiddlewareRequestParam, event: NextMiddlewareEvtParam): NextMiddlewareReturn;
+}
+
+export const clerkMiddleware: ClerkMiddleware = (...args: unknown[]): any => {
+  const [request, event] = parseRequestAndEvent(args);
```
dimkl, line 55: ❓ Should we rename those 2 since there is a naming conflict with the nextMiddleware params defined below? Or this intentional? 
Reading the code inside the `clerkMiddleware` function may be confusing as to where do we intend to use  the request and event.

packages/types/src/multiDomain.ts
dimkl, line 1: 🤔 I agree with the simplification of the Multidomain types but this probably needs to be approved by the team that introduced those.

packages/nextjs/src/server/types.ts
dimkl, line 1: 🎉

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -0,0 +1,187 @@
...
+      clerkRequest,
+      createAuthenticateRequestOptions(clerkRequest, options),
+    );
+
+    const locationHeader = requestState.headers.get(constants.Headers.Location);
+    if (locationHeader) {
+      const res = new Response(null, { status: 307, headers: requestState.headers });
+      return decorateResponseWithObservabilityHeaders(res, requestState);
+    } else if (requestState.status === AuthStatus.Handshake) {
+      throw new Error('Clerk: handshake status without redirect');
```
dimkl, line 71: 🙃 let's use the `errorThrower` instead of `Error`.

dimkl (approved):

packages/types/src/multiDomain.ts
```diff
@@ -2,53 +2,19 @@ import type { ClerkOptions } from './clerk';
...
-  | {
-      isSatellite: Exclude<ClerkOptions['isSatellite'], undefined>;
-      proxyUrl: StringOrURLFnToString;
-      domain?: never;
-    };
+export type MultiDomainAndOrProxy = {
+  isSatellite?: ClerkOptions['isSatellite'];
+  proxyUrl?: StringOrURLFnToString;
+  domain?: StringOrURLFnToString;
+};
```
panteliselef, line 9: I feel that we are loosing some good dx here. Did you encounter any issues with these ? I believe these types were simplified enough especially after removing the discriminated unions for frontent api and publishable key.

packages/backend/src/tokens/clerkUrl.ts
nikosdouvlis: In my mind, `ClerkUrl` extends the `URL` class and adds more utilities (eg `isCrossOrigin` and all other url-related utils we have in shared right now). My intention is to use it in backend for now, and then move it to shared and add all url-related utils on it. 

In any case, `ClerkRequest` extends `Request` and `ClerkUrl` extends `URL`. Do you still feel that ClerkUrl should be merged with `ClerkRequest`? Happy to discuss more about this! :D

packages/nextjs/src/app-router/server/auth.ts
nikosdouvlis: Yes :)

packages/nextjs/src/pages/__tests__/index.test.tsx
nikosdouvlis: In my opinion, these types greatly increased our type system's complexity without providing the benefit I'd expect.

I'd like to discuss with @panteliselef more about this before merging as he has much more context than me :)

packages/types/src/multiDomain.ts
nikosdouvlis: Agreed, I'll be talking with @panteliselef about this today

packages/backend/src/tokens/clerkRequest.ts
```diff
@@ -62,7 +51,7 @@ class ClerkRequest extends Request {
 }
 
 export const createClerkRequest = (...args: ConstructorParameters<typeof ClerkRequest>): ClerkRequest => {
-  return new ClerkRequest(...args);
+  return args[0] instanceof ClerkRequest ? args[0] : new ClerkRequest(...args);
```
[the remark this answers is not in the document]
nikosdouvlis: Fight now, I'm using `ClerkRequest` directly in `clerk/nextjs` in order to calculate the correct URL for any redirects that need to take place (`redirectToSignIn` with `returnBackUrl` etc). Since I already have a `ClerkRequest` instance, I'm simply passing it in `authenticateRequest`, which normalizes `Request` to `ClerkRequest`.

I'd like us to move all redirect logic to clerk/backend but this needs to happen across all frameworks. For now, I'm simply avoiding initializing a second `ClerkRequest` with the `instanceof` check here.

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -0,0 +1,187 @@
...
+    const authObject = requestState.toAuth();
+
+    const authObjWithMethods: ClerkMiddlewareAuthObject = Object.assign(authObject, {
+      protect: createMiddlewareProtect(authObject),
+      redirectToSignIn: createMiddlewareRedirectToSignIn(clerkRequest, requestState),
+    });
+
+    let handlerResult: Response = NextResponse.next();
+    try {
+      handlerResult = (await handler?.(() => authObjWithMethods, request, event)) || handlerResult;
```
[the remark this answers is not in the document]
nikosdouvlis: Good point!

packages/nextjs/src/server/protect.ts
```diff
@@ -40,34 +59,34 @@ export const createProtect = (authObj: AuthObject): AuthProtect => {
     /**
      * User is not authenticated
      */
-    if (!authObj.userId) {
-      return handleUnauthorized();
+    if (!authObject.userId) {
+      return handleUnauthenticated ? handleUnauthenticated() : handleUnauthorized();
```
panteliselef, line 63: We should probably opt out of the redirect to sign-in if the request refers to an endpoint and not a apge

packages/nextjs/src/server/clerkMiddleware.ts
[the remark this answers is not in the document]
BRKalow: Ah yeah, I wasn't sure if that was the case 😅 thanks for trying!

packages/nextjs/src/server/protect.ts
nikosdouvlis: That is correct. Context for future reviewers: we're currently discussing this offline, will post more details soon.

packages/backend/src/tokens/clerkUrl.ts
BRKalow: Do we have many use cases in mind already for the `ClerkUrl` abstraction? If we're mostly using it in relation to `ClerkRequest` now, it might be better to wait to introduce this. The extra abstractions here add some overhead when trying to grok the code, so I would prefer to wait until they are needed.

nikosdouvlis: @BRKalow, @dimkl 

I thought about this a bit more and I realised that when discussing, we're always referring to "cross-origin requests"  and never "cross-origin URLs" (or similar)

I think the isCrossOrigin check makes sense for both, but `clerkRequest.isCrossOrigin(url)` feels more natural in our domain. WIll update the code shortly

packages/nextjs/src/server/clerkMiddleware.ts
nikosdouvlis: Moved `const [request, event] = parseRequestAndEvent(args);` below `nextMiddleware`, thanks!

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -50,16 +38,10 @@ export const DEFAULT_IGNORED_ROUTES = [`/((?!api|trpc))(_next.*|.+\\.[\\w]+$)`];
...
-  | Array<RegExp | RouteMatcherWithNextTypedRoutes>
-  | RegExp
-  | RouteMatcherWithNextTypedRoutes
-  | ((req: NextRequest) => boolean);
-
 type IgnoredRoutesParam = Array<RegExp | string> | RegExp | string | ((req: NextRequest) => boolean);
 type ApiRoutesParam = IgnoredRoutesParam;
 
-type WithClerkUrl<T> = T & {
+type WithExperimentalClerkUrl<T> = T & {
```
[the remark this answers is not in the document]
nikosdouvlis: (just to be clear - this is referring to the old middleware)
I aligned the names as the key we inject into `nextUrl` is `experimental_clerkUrl`

packages/nextjs/src/server/protect.ts
```diff
@@ -0,0 +1,130 @@
...
+    req.headers.get(constants.Headers.SecFetchDest) === 'document' ||
+    req.headers.get(constants.Headers.Accept)?.includes('text/html') ||
+    (!!req.headers.get(nextConstants.Headers.NextUrl) && !isServerActionRequest(req))
+  );
+};
+
+// In case we want to handle router handlers and server actions differently in the future
+// const isRouteHandler = (req: Request) => {
+//   return !isPageRequest(req) && !isServerAction(req);
+// };
```
nikosdouvlis, line 130: Many thanks to @panteliselef for taking the time to investigate, test, and help me simplify these checks 🥇

packages/nextjs/src/server/protect.ts
nikosdouvlis: https://github.com/clerk/javascript/pull/2404/commits/0939fdab64bac1a159073c8a38fc1b5c824cef08

packages/nextjs/src/server/clerkMiddleware.ts
SokratisVidros: @nikosdouvlis Reminder ☝️

packages/nextjs/src/server/protect.ts
SokratisVidros: Should we remove the commented out code?

packages/nextjs/src/utils/appendDevBrowserOnCrossOrigin.ts
```diff
@@ -0,0 +1,34 @@
...
+import { constants } from '@clerk/backend/internal';
+import { DEV_BROWSER_JWT_KEY, isDevelopmentFromSecretKey, setDevBrowserJWTInURL } from '@clerk/shared';
+import { NextResponse } from 'next/server';
+
+import { SECRET_KEY } from '../server/constants';
+
+/**
+ * Grabs the dev browser JWT from cookies and appends it to the redirect URL when redirecting to cross-origin.
+ */
+export const appendDevBrowserOnCrossOrigin = (
```
SokratisVidros, line 12: Why is this still required in the new handshake era where URL-based session syncing works for all instances?

packages/types/src/multiDomain.ts
SokratisVidros: I believe the simplification is in favor of DX. Having said that, `StringOrURLFnToString` can be renamed to `StringOrFunction` and `BooleanOrFunction`.

nikosdouvlis: I reverted these (and simplified other) types so we can discuss more about this. However I believe that the DX benefits here are not as great as we'd like them to be. The types are becoming too complex in some cases, and the errors generated by misusing the props are not readable anymore.

I think I'd rather have some extra runtime checks in dev mode throwing readable errors, rather than overcomplicated discriminated union types that are hard to understand

PR #2421, opened 2023-12-20
Fix backend error message details

Nikpolik:
## Description

This was caused by API errors being parsed twice. Once in the response  error handling code and again when initializing a new `ClerkAPIResponseError`. 

This is was fixed in v5 with the removal of `ClerkAPIResponseError`. 

`ClerkAPIResponseError` is also used by clerk-js the change was to the backend package.        

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/backend/src/api/request.ts
+++ b/packages/backend/src/api/request.ts
@@ -51,18 +51,22 @@ type LegacyRequestFunction = <T>(requestOptions: ClerkBackendApiRequestOptions) 
  * TODO: Simply remove this wrapper and the ClerkAPIResponseError before the v5 release.
  */
 const withLegacyReturn =
-  (cb: any): LegacyRequestFunction =>
+  (cb: (...args: any) => Promise<ClerkBackendApiResponse<any>>): LegacyRequestFunction =>
   async (...args) => {
-    // @ts-ignore
-    const { data, errors, status, statusText, clerkTraceId } = await cb<T>(...args);
-    if (errors === null) {
-      return data;
+    const response = await cb(...args);
+    if (response.errors === null) {
+      return response.data;
     } else {
-      throw new ClerkAPIResponseError(statusText || '', {
-        data: errors,
+      const { errors, clerkTraceId } = response;
+      // TODO: To be removed with withLegacyReturn
+      const { status, statusText } = response as any;
+      const error = new ClerkAPIResponseError(statusText || '', {
+        data: [],
         status: status || '',
         clerkTraceId,
       });
+      error.errors = errors;
+      throw error;
     }
   };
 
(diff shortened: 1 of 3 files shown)


packages/backend/src/api/request.ts
```diff
@@ -154,7 +154,7 @@ export function buildRequest(options: CreateBackendApiOptions) {
         data: deserialize(data),
         errors: null,
       };
-    } catch (err) {
+    } catch (err: any) {
```
SokratisVidros, line 157:
```suggestion
    } catch (err: unknown) {
```

packages/backend/src/api/request.ts
```diff
@@ -170,7 +170,7 @@ export function buildRequest(options: CreateBackendApiOptions) {
 
       return {
         data: null,
-        errors: parseErrors(err),
+        errors: err?.errors,
```
SokratisVidros, line 173: @Nikpolik, I am not sure I get this fix. These errors are reported on the server and don't propagate to the client where ClerkJS lives. Can you please elaborate?

Nikpolik: Initially parse errors is called here. This transforms it to camel case.

Then the result is passed to [`withLegacyReturn`](https://github.com/clerk/javascript/blob/nikpolik%2Fv4-fix-backend-error-message-details/packages/backend/src/api/request.ts#L61-L65) which initialises a new `ClerkAPIResponseError`
Finally parse errors is called again inside [`ClerkAPIResponseError`](https://github.com/clerk/javascript/blob/release%2Fv4/packages/shared/src/error.ts#L109-L109)

packages/backend/src/api/request.ts
Nikpolik: This was done on purpose. Since the result is passed to with `withLegacyReturn` all type safety is gone here. I can fix the types as well but wanted to get the fix out quick.

Nikpolik:
@SokratisVidros I updated the fix a bit to be a bit simpler

nikosdouvlis (approved):

PR #2430, opened 2023-12-21
fix(clerk-js): Fetching custom role in OrganizationSwitcher  with cache

panteliselef:
### Problem
In order to display a custom role we need to fetch them, so far we didn't have a robust way of fetching resources in clerk-js. A while back i had implemented `useFetch`, today I'm extending it in order to use a cache in order to fetch and share resources across components more efficiently.

### Before
When a component need to display the custom roles we had to fetch those each time the component was mounted.

### After
We fetch the resource once, other subscribers will not trigger a refetch unless the data is stale (in cache for 2 mins).

### Goal
The goal of this PR is to provide a good enough fix for the issue with an efficient solution. This is a bug in v4 as well, and a simple solution is better when backporting is necessary. If we think this solution is limiting, we should consider migrating to swr as its bundlesize is minimal and much more robust.

For tests:
- We need to clear cache before each test

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/clerk-js/src/ui/hooks/useFetch.ts
+++ b/packages/clerk-js/src/ui/hooks/useFetch.ts
@@ -15,10 +15,20 @@ export interface Cache<Data = any> {
   set(key: string, value: State<Data>): void;
 
   delete(key: string): void;
+
+  clear(): void;
 }
 
-const map = new WeakMap<Cache, State>();
+let requestCache = new WeakMap<Cache, State>();
 const subscribers = new Set<() => void>();
+
+/**
+ * This utility should only be used in tests to clear previously fetched data
+ */
+export const clearFetchCache = () => {
+  requestCache = new WeakMap<Cache, State>();
+};
+
 const useCache = (
   param: any,
 ): {
@@ -26,12 +36,10 @@ const useCache = (
   set: (state: State) => void;
   subscribe: (callback: () => void) => () => void;
 } => {
-  // const subscribers = useRef(new Set<() => void>());
-
-  const get = useCallback(() => map.get(param), [param]);
+  const get = useCallback(() => requestCache.get(param), [param]);
   const set = useCallback(
     (data: State) => {
-      map.set(param, data);
+      requestCache.set(param, data);
       subscribers.forEach(callback => callback());
     },
     [param],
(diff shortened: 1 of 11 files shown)


packages/clerk-js/src/ui/hooks/useFetch.ts
```diff
@@ -10,35 +68,56 @@ export const useFetch = <T>(
     onSuccess?: (data: T) => void;
   },
 ) => {
-  const [data, setData] = useSafeState<T | null>(null);
-  const requestStatus = useLoadingStatus({
-    status: 'loading',
-  });
+  const cache = useCache(params);
```
BRKalow, line 71: It appears that `params` is regularly an object here that is not memoized. If I'm understanding the `useCache` implementation correctly, we're using the raw `params` value as a cache key. This means that every re-render could potentially introduce a new entry in the cache. We likely need to stringify the object in a deterministic way (e.g. sorted) to get a consistent cache key (I see we're doing it below).

Additionally, the `useCallback` usage within `useCache` is not so useful for the same reason (`params` changing reference on each call).

BRKalow (changes requested):
We need to tweak the logic around `params` in `useCache` to ensure we're dealing with a stable cache key.

packages/clerk-js/src/ui/hooks/useFetch.ts
panteliselef: Thanks for caching this @BRKalow, shall i go the stringify approach ?

BRKalow (approved):
💯

PR #2548, opened 2024-01-10
Remove create email api function from backend package

Nikpolik:
## Description

Remove createEmail function from `@clerk/backend`.

The equivalent `/emails` Backend API endpoint will also be dropped in the future, since this feature will no longer be available for new instances.

For a brief period it will still be accessible for instances that have used it in the past 7
days.

New instances will get a 403 forbidden response if they try to access it.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Packages affected

- [x] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`


.changeset/curvy-suns-roll.md
```diff
@@ -0,0 +1,7 @@
+---
+'@clerk/backend': patch
+---
+
+Remove createEmail function from @clerk/backend.
+
+The corresponding endpoint is also being removed from the API.
```
nikosdouvlis, line 7: Discussed with @Nikpolik offline and we will not be providing an alternative method for the time being. 

Can we please add a note informing any customers currently using this to reach out to our support team if they wish to update to v5?

```suggestion
The `emails` endpoint helper and the corresponding `createEmail` method have been removed from the `@clerk/backend` SDK and `apiClint.emails.createEmail` will no longer be available. We will not be providing an alternative method for creating and sending emails directly from our JavaScript SDKs with this release. If you are currently using `createEmail` and you wish to update to the latest SDK version, please reach out to our support team (https://clerk.com/support) so we can assist you.
```

nikosdouvlis (approved):

packages/backend/src/api/__tests__/factory.test.ts
```diff
@@ -88,39 +86,6 @@ export default (QUnit: QUnit) => {
       );
     });
 
-    test('executes a successful backend API request for a resource that contains data (key related to pagination)', async assert => {
```
dimkl, line 91: 🔧  Instead of removing this test, replace the `apiClient.emails.createEmail(body);` with another API call. This test is not related to the `EmailAPI`, it was just used as an example.

dimkl (changes requested):
Please review the comment in the test file

packages/backend/src/api/__tests__/factory.test.ts
Nikpolik: Is it possible this test is no longer needed? 🤔 After looking around only `SMSMessage` and `Emails` contain the property `readonly data?: Record<string, any> | null` that is being checked here.

To be completely honest I am not 100% sure what is being tested here  😓  And whats the differences between this test and the first one that `executes a successful backend API request for a single resource and parses the response`

.changeset/curvy-suns-roll.md
Nikpolik: Much better message thanks!

packages/backend/src/api/__tests__/factory.test.ts
dimkl: You are correct. This case is no longer applicable. let's drop it.

dimkl (approved):

PR #2604, opened 2024-01-16
fix(clerk-js): Throw error for unsupported method `OrganizationMembership.reload`

panteliselef:
## Description

The only class that overrides the reload method from Base is OrganizationMembership. And it is broken, because the logic inside it no longer makes sense with the paginated endpoints. There is no endpoint to fetch by id so we need to notify developers with a descriptive error

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/.changeset/gold-shrimps-behave.md
+++ b/.changeset/gold-shrimps-behave.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Throw error for unsupported method `OrganizationMembership.reload`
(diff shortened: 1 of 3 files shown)


dimkl (approved):

nikosdouvlis (approved):

nikosdouvlis:
@panteliselef curious as to why we're not completely removing the method all together

dimkl:
@nikosdouvlis the `reload()` method is inherited from the `BaseResource` (base class). If we remove this, the base class implementation will be invoked which does not work (based on the description of this PR).

panteliselef:
I do have another ticket that will come up during when triaging and we can discuss there for alternative approaches

PR #2634, opened 2024-01-18
feat(nextjs): Infer the path the component is mounted at automatically in pages and app router

nikosdouvlis:
## Description

Detailed explanation is in the comments, please take a look at the code itself.
This PR allows using all components with path routing without passing the path manually

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/nextjs/src/client-boundary/uiComponents.tsx
+++ b/packages/nextjs/src/client-boundary/uiComponents.tsx
@@ -17,7 +17,7 @@ import type {
 } from '@clerk/types';
 import React from 'react';
 
-import { useClerkNextOptions } from './NextOptionsContext';
+import { usePathnameWithoutWithCatchAll } from './usePathnameWithoutCatchAll';
 
 export {
   OrganizationList,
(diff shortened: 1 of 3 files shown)


packages/nextjs/src/client-boundary/uiComponents.tsx
```diff
@@ -50,17 +52,18 @@ export const CreateOrganization = (props: CreateOrganizationProps) => {
...
   (props: OrganizationProfileProps) => {
-    return <BaseOrganizationProfile {...useRoutingProps('OrganizationProfile', props)} />;
+    const path = usePathnameWithoutWithCatchAll();
+    return <BaseOrganizationProfile {...useRoutingProps('OrganizationProfile', props, { path })} />;
   },
   { ...BaseOrganizationProfile },
 );
 
 export const SignIn = (props: SignInProps) => {
-  const { signInUrl } = useClerkNextOptions();
```
octoper, line 59: I don't think we should remove the `signInUrl` here, this is needed if we want to keep value coming from `<ClerkProvider/>` or env variable `NEXT_PUBLIC_CLERK_SIGN_IN_URL`.

packages/nextjs/src/client-boundary/uiComponents.tsx
```diff
@@ -50,17 +52,18 @@ export const CreateOrganization = (props: CreateOrganizationProps) => {
...
-  return <BaseSignIn {...useRoutingProps('SignIn', props, { path: signInUrl })} />;
+  const path = usePathnameWithoutWithCatchAll();
+  return <BaseSignIn {...useRoutingProps('SignIn', props, { path })} />;
 };
 
 export const SignUp = (props: SignUpProps) => {
-  const { signUpUrl } = useClerkNextOptions();
```
octoper, line 64: The same as `signInUrl` applies here

packages/nextjs/src/client-boundary/uiComponents.tsx
nikosdouvlis: I don't think we need to respect this - the env variable is supposed to be used so the SignIn component (and the middleware) knows where SignUp is mounted and vice versa.

You want to render the component at the path its mounted, and respecting the global value as the path means that the component will only be able to be mounted to a single route

PR #2750, opened 2024-02-07
fix(clerk-js): Fix neutral alpha shades generation

anagstef:
## Description

This PR fixes an issue in generating alpha shades from a user-provided color.
The value provided for the neutral alpha shades should not be included in the scale.

### Before:
Users could provide a single color, a part of the scale, or the whole scale to generate alpha colors. We generated an 50-950 alpha scale based on the 500 shade (or the single color) provided by the user, and then we would set the base color used for generation directly to the shade-500 position. This is problematic, because when user provides the color `black`, we end up with a scale that has the black color in the shade-500 position, where the shade-950 was lighter than black.

### After:
The user will be able to provide only a single color or the whole scale for the `colorNeutral` property.
- If user provides a single color: We generate the whole scale based on that single color, and we do not add the base color used into the scale.
- If user provides the whole scale: We just use the scale provided by the user.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`


clerk-cookie:

| :-- | :-- | :-- |
| :cook: _Deploying..._ | [Inspect](https://javascript-preview-m2ukyuviu.clerkstage.dev) | Feb 08, 2024 08:37 AM |

desiprisg (approved):
LGTM!

PR #2757, opened 2024-02-08
fix(clerk-react): Properly fire onLoad event when clerk-js is already loaded

panteliselef:
In case of mounting-unmounting-mounting the `IsomorphicClerk.addOnLoaded` would not fire correctly, hence `derivedState` would return the initialState instead of the client side state.

In addition to the above fix, we are now cleaning up the IsomorphicClerk instance when the provider is unmounted.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/react/src/contexts/ClerkContextProvider.tsx
+++ b/packages/react/src/contexts/ClerkContextProvider.tsx
@@ -83,5 +83,11 @@ const useLoadedIsomorphicClerk = (options: IsomorphicClerkOptions) => {
     isomorphicClerk.addOnLoaded(() => setLoaded(true));
   }, []);
 
+  React.useEffect(() => {
+    return () => {
+      IsomorphicClerk.clearInstance();
+    };
+  }, []);
+
   return { isomorphicClerk, loaded };
 };
--- a/.changeset/brown-lemons-film.md
+++ b/.changeset/brown-lemons-film.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-react': patch
+---
+
+Properly fire onLoad event when clerk-js is already loaded.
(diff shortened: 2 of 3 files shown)


packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -83,5 +83,11 @@ const useLoadedIsomorphicClerk = (options: IsomorphicClerkOptions) => {
     isomorphicClerk.addOnLoaded(() => setLoaded(true));
   }, []);
 
+  React.useEffect(() => {
+    return () => {
+      IsomorphicClerk.clearInstance();
+    };
+  }, []);
```
dimkl, line 90: ❓ shouldn't we merge this useEffect with the one above that initializes the ` isomorphicClerk.addOnLoaded` ?

dimkl (approved):

packages/react/src/contexts/ClerkContextProvider.tsx
panteliselef: The actual initialization happends in the useMemo
`  const isomorphicClerk = React.useMemo(() => IsomorphicClerk.getOrCreateInstance(options), []);`. So i'm not sure, the position shouldn't affect the result tho.

PR #2765, opened 2024-02-08
fix(nextjs,clerk-js,types): Stop Clerk component flickering when used in App Router

nikosdouvlis:
A few things going on in this PR.

### Navigation metadata
When using `router.push` from `next/navigation` (app router) the components in the current page will unmount before the navigation. In our case, if a component uses path-routing all internal navigations (for example `/user` > `/user/security` use `router.push` but that means that the component will unmount and immediately mount again in the new URL.

However, our components are client components and they can support client-side-only navigations. This type of navigation is usually called shallow navigation. App router apps can use the native `window.history` APIs in order to trigger shallow navigations.

We need to differentiate between internal navigations (navigations within the component) and external navigations (navigation outside the scope of the component) as we want to use shallow navigations for the former and normal `router.push` navigations for the latter.

To achieve that, the clerk-js router now sends a `metadata` object with each navigation that can be used by the consuming framework wrapper, enabling more fine-grained control over the type of navigation.

### Shallow navigations and nextjs
We want to be as little intrusive as possible, so when trigger shallow navigations, we need to make sure to preserve the default next router behavior. When calling the `window.history` APIs, we should respect how `windowHistorySupport` and let any changes to the URL update the next hooks as normal.
 
-  windowHistorySupport was enabled by default in `next@14.1.0`.
- For apps >= 14.1.0, we need to call push/replaceState with `null` as the first argument for WHS to work
- For older apps, we need to call the APIs with the `window.history.state`

Refs: 
- https://github.com/vercel/next.js/pull/58335
- https://github.com/vercel/next.js/discussions/48110
- https://github.com/vercel/next.js/discussions/18072
- https://github.com/47ng/nuqs/pull/482/files#diff-713238c17ffa37b63ad76fc5a5d9a2c6e923cc009e6593f86adf8fb32feef493R183

See inline comments for more details. 

### usePathnameWithoutCatchAll
This hook let us automatically determine the path a component is mounted on, while taking into account that components can be mounted in catch-all routes, so simply using `usePathname()` is not enough.
For example, imagine we are using path-based routing and we have a UserProfile component mounted at `/user/[[...rest]]/page.tsx`. The possible paths the component can use would be `/user` and `/user/security`, but the base path needs to be set as `/user` always for our internal router to work. 

We're now saving the initial value in a `ref` so it never gets updated after the initial (browser) navigation to the page. For details on what this hook does, please refer to the inline comments. 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Packages affected

- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/fastify`
- [ ] `gatsby-plugin-clerk`
- [ ] `@clerk/localizations`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/remix`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/themes`
- [ ] `@clerk/types`
- [ ] `build/tooling/chore`

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -726,8 +726,9 @@ export class Clerk implements ClerkInterface {
       return;
     }
 
+    const metadata = options?.metadata ? { __internal_metadata: options?.metadata } : undefined;
     // React router only wants the path, search or hash portion.
-    return await customNavigate(stripOrigin(toURL));
+    return await customNavigate(stripOrigin(toURL), metadata);
   };
 
   public buildUrlWithAuth(to: string): string {
--- a/packages/nextjs/src/app-router/client/useAwaitableNavigate.ts
+++ b/packages/nextjs/src/app-router/client/useAwaitableNavigate.ts
@@ -3,7 +3,7 @@
 import { useRouter } from 'next/navigation';
 import { useEffect, useRef, useTransition } from 'react';
 
-type NavigateFunction = ReturnType<typeof useRouter>['push'];
+import type { NextClerkProviderProps } from '../../types';
 
 /**
  * Creates an "awaitable" navigation function that will do its best effort to wait for Next.js to finish its route transition.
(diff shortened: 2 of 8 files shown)


packages/nextjs/src/app-router/client/useAwaitableNavigate.ts
```diff
@@ -12,31 +12,48 @@ type NavigateFunction = ReturnType<typeof useRouter>['push'];
...
+            // but only if passing `null` as the history state.
+            // Older versions need to maintain the history state for push to work,
+            // without affecting how the Next router works.
+            const state = ((window as any).next?.version ?? '') < '14.1.0' ? history.state : null;
+            window.history.pushState(state, '', to);
+          } else {
+            // If the navigation is external (usually when navigating away from the component but still within the app),
+            // we should use the Next.js router to navigate as it will handle updating the URL and also
+            // fetching the new page if necessary.
+            router.push(to);
```
panteliselef, line 40: Don't we need to pass options here ?

panteliselef (approved):

packages/nextjs/src/app-router/client/useAwaitableNavigate.ts
nikosdouvlis: We're not using these right now

packages/clerk-js/src/core/clerk.ts
```diff
@@ -722,12 +722,13 @@ export class Clerk implements ClerkInterface {
     }
 
     if (toURL.origin !== window.location.origin || !customNavigate) {
+      console.log('windownavigate', toURL);
```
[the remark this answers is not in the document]
nikosdouvlis: Woops, done

panteliselef:
💯 I enjoyed reading that PR description

PR #2782, opened 2024-02-12
feat(shared,clerk-react,clerk-js): Auto-revalidation of paginated resources based on events

panteliselef:
## Description

Three new events were added to `eventBus` in order to describe the action that was taken and broadcast it to subscribers.

This ensures that specific clerk-js methods can broadcast an event and our hooks can listen to that event and in this case revalidate the cache.

### Why do we need this
This is extremely useful of situations where our UI components are used in conjunction with custom implementations. For example, let's say that an application is using the Organization Profile and CreateOrganization components, but it has a custom implementation of OrganizationSwitcher. Until now there was no way to notify the custom switcher to refetch data when an organization was created via the component.

### Before
"test2" is not getting removed from custom switcher

https://github.com/clerk/javascript/assets/19269911/da74d1c3-6373-40f0-8ec4-55ab30b74eba

### After
"dwadaw" is getting correctly removed from custom switcher

https://github.com/clerk/javascript/assets/19269911/92df255d-dd68-44ea-8604-8062935a5257

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/events.ts
+++ b/packages/clerk-js/src/core/events.ts
@@ -2,7 +2,9 @@ import type { TokenResource } from '@clerk/types';
 
 export const events = {
   TokenUpdate: 'token:update',
-  OrganizationDestroy: 'organization:destroy',
+  OrganizationDeleted: 'organization:deleted',
+  OrganizationCreated: 'organization:created',
+  UserMembershipDeleted: 'user:membership_deleted',
 } as const;
 
 type ClerkEvent = (typeof events)[keyof typeof events];
@@ -12,7 +14,9 @@ type TokenUpdatePayload = { token: TokenResource | null };
 
 type EventPayload = {
   [events.TokenUpdate]: TokenUpdatePayload;
-  [events.OrganizationDestroy]: null;
+  [events.OrganizationDeleted]: null;
+  [events.OrganizationCreated]: null;
+  [events.UserMembershipDeleted]: null;
 };
 
 const createEventBus = () => {
@@ -40,6 +44,7 @@ const createEventBus = () => {
         event,
         handlers.filter(h => h !== handler),
       );
+      return;
     }
 
     eventToHandlersMap.set(event, []);
--- a/packages/clerk-js/src/core/events.ts
+++ b/packages/clerk-js/src/core/events.ts
@@ -2,6 +2,7 @@ import type { TokenResource } from '@clerk/types';
 
 export const events = {
   TokenUpdate: 'token:update',
+  OrganizationDestroy: 'organization:destroy',
 } as const;
 
 type ClerkEvent = (typeof events)[keyof typeof events];
@@ -11,6 +12,7 @@ type TokenUpdatePayload = { token: TokenResource | null };
 
 type EventPayload = {
   [events.TokenUpdate]: TokenUpdatePayload;
+  [events.OrganizationDestroy]: null;
 };
 
 const createEventBus = () => {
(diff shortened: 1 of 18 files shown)


packages/clerk-js/src/ui/components/CreateOrganization/CreateOrganizationForm.tsx
```diff
@@ -40,9 +39,7 @@ export const CreateOrganizationForm = withCardStateProvider((props: CreateOrgani
   const wizard = useWizard({ onNextStep: () => card.setError(undefined) });
 
   const lastCreatedOrganizationRef = React.useRef<OrganizationResource | null>(null);
-  const { createOrganization, isLoaded, setActive, userMemberships } = useOrganizationList({
-    userMemberships: organizationListParams.userMemberships,
-  });
+  const { createOrganization, isLoaded, setActive } = useOrganizationList();
```
panteliselef, line 42: We don't need to revalidate manually anymore

packages/clerk-js/src/ui/components/OrganizationProfile/ActionConfirmationPage.tsx
```diff
@@ -14,35 +14,26 @@ import {
...
-  });
 
   return () =>
     card
       .runAsync(async () => {
         await leavePromise?.();
       })
-      .then(() => {
-        void userMemberships.revalidate?.();
-        void userInvitations.revalidate?.();
```
panteliselef, line 36: We don't need to revalidate manually anymore

packages/shared/src/react/types.ts
```diff
@@ -44,6 +44,9 @@ export type PaginatedHookConfig<T> = T & {
    * Return the previous key's data until the new data has been loaded
    */
   keepPreviousData?: boolean;
+
+  __unstable__dependencyRevalidation?: boolean;
+  __unstable__defaultRevalidateOnEvents?: string[];
```
panteliselef, line 49: These will be exposed to public API. It is the only think that is exposed to public API

packages/clerk-js/src/core/events.ts
```diff
@@ -38,6 +44,7 @@ const createEventBus = () => {
         event,
         handlers.filter(h => h !== handler),
       );
+      return;
```
panteliselef, line 47: Previously all handlers were removed

packages/shared/src/react/hooks/usePagesOrInfinite.ts
```diff
@@ -85,6 +86,8 @@ export const usePagesOrInfinite: UsePagesOrInfinite = (params, fetcher, config,
   const enabled = config.enabled ?? true;
   const triggerInfinite = config.infinite ?? false;
   const keepPreviousData = config.keepPreviousData ?? false;
+  const __unstable__dependencyRevalidation = config.__unstable__dependencyRevalidation ?? true;
+  const __unstable__defaultRevalidateOnEvents = config.__unstable__defaultRevalidateOnEvents || [];
```
LekoArts, line 90: Should this default be different from the one you pass in at `packages/shared/src/react/hooks/useOrganizationList.tsx`?

panteliselef: I added this as "failsafe" values for that utility hook. I get I could change the types, but then I would need to pass the "empty" values from useOrganization hook. Not sure if we would be gaining something from that.

packages/clerk-js/src/core/events.ts
```diff
@@ -11,6 +14,9 @@ type TokenUpdatePayload = { token: TokenResource | null };
 
 type EventPayload = {
   [events.TokenUpdate]: TokenUpdatePayload;
+  [events.OrganizationDeleted]: null;
+  [events.OrganizationCreated]: null;
+  [events.UserMembershipDeleted]: null;
```
panteliselef, line 19: At the moment, I am not interested in passing payload as we are only need to know that an event has fired to trigger the revalidation later.

panteliselef:
We are closing this due to inactivity, we would like to experiment more with the solution before releasing it as stable.

packages/shared/src/react/types.ts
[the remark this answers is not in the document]
panteliselef: It really depends on the amount of configuration we want the developer to make.

I added `__unstable__dependencyRevalidation` as a way of manually turning off the "auto-revalidation"

And `__unstable__defaultRevalidateOnEvents` is used in order to manually select for which event the revalidation should run.

We can change this and have the `['organization:deleted', 'organization:created', 'user:membership_deleted']` be hard coded and not configurable by the caller, and also always perform "auto-revalidation"

PR #2794, opened 2024-02-13
feat(clerk-js,types): Ability to enable/disable animations

octoper:
## Description

This PR adds the ability to enable/disable animations through the `layout.animations` key in the `appearance` prop.
The animations are enabled by default but they always repsect the browser `prefers-reduced-motion` setting, so if the `prefers-reduced-motion` is set to `reduce` the animations will be disabled regardless the value of `layout.animations`.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


playground/nextjs/pages/_app.tsx
```diff
@@ -59,6 +59,7 @@ function MyApp({ Component, pageProps }: AppProps) {
           helpPageUrl: '/help',
           privacyPageUrl: '/privacy',
           termsPageUrl: '/terms',
+          animations: false,
```
anagstef, line 62: @octoper why do we turn them off in our playground?

anagstef (approved):

PR #2797, opened 2024-02-14
fix(elements): Misc stuff

LekoArts:
## Description

- Don't export Form from /common
- Make GlobalError a div
- Don't export SocialProviderIcon from /common

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/react/common/form/index.tsx
```diff
@@ -273,12 +273,12 @@ function GlobalError({ children, code, ...rest }: FormGlobalErrorProps) {
   const child = typeof children === 'function' ? children(error) : children;
 
   return (
-    <span
+    <div
```
LekoArts, line 276: This way the error will already have `display: block`

packages/elements/src/react/common/index.ts
```diff
@@ -1,2 +1 @@
-export { Field, FieldError, FieldState, Form, GlobalError, Input, Label, Submit } from '~/react/common/form';
```
LekoArts, line 1: Removes `Form` as it doesn't need to be a public API

packages/elements/src/react/common/index.ts
```diff
@@ -1,2 +1 @@
...
-export { SocialProviderIcon } from '~/react/common/third-party-providers/social-provider';
```
LekoArts, line 2: Removes `SocialProviderIcon` as it's useless outside of the `<SocialProvider>` context, so to remove confusion it's not exported here anymore

tmilewski (approved):

PR #2809, opened 2024-02-16
Nikos/bunch of v5 changes

nikosdouvlis:
## Description
- Rename beta-v5 to beta
- Port the cookie-setting strategy change from v4
- Stop adding "redirect" as authreason as it overrides AuthReason in some scenarios, making debugging harder 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -1,6 +1,6 @@
 {
   "mode": "pre",
-  "tag": "beta-v5",
+  "tag": "beta",
   "initialVersions": {
     "@clerk/backend": "0.31.3",
     "@clerk/chrome-extension": "0.4.10",
--- a/packages/clerk-js/src/core/resources/Session.ts
+++ b/packages/clerk-js/src/core/resources/Session.ts
@@ -160,7 +160,11 @@ export class Session extends BaseResource implements SessionResource {
     const cachedEntry = skipCache ? undefined : SessionTokenCache.get({ tokenId }, leewayInSeconds);
 
     if (cachedEntry) {
-      return cachedEntry.tokenResolver.then(res => res.getRawString());
+      const cachedToken = await cachedEntry.tokenResolver.then(res => res);
+      if (!template) {
+        eventBus.dispatch(events.TokenUpdate, { token: cachedToken });
+      }
+      return cachedToken.getRawString();
     }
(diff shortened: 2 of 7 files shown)


.changeset/pre.json
```diff
@@ -1,6 +1,6 @@
 {
   "mode": "pre",
-  "tag": "beta-v5",
+  "tag": "beta",
```
dimkl, line 3: 🤔 Currently all the package.json files in this monorepo refer to `@beta-v5.{something}`. Should we add a patch version bump in this changeset to make them convert to `beta`?

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -160,7 +160,11 @@ export class Session extends BaseResource implements SessionResource {
     const cachedEntry = skipCache ? undefined : SessionTokenCache.get({ tokenId }, leewayInSeconds);
 
     if (cachedEntry) {
-      return cachedEntry.tokenResolver.then(res => res.getRawString());
+      const cachedToken = await cachedEntry.tokenResolver.then(res => res);
+      if (!template) {
+        eventBus.dispatch(events.TokenUpdate, { token: cachedToken });
+      }
+      return cachedToken.getRawString();
```
dimkl, line 167: 🤔 This means we forgot to apply Bryce change to the beta version?
ref: https://github.com/clerk/javascript/pull/2585

dimkl (approved):

packages/clerk-js/src/core/resources/Session.ts
nikosdouvlis: We did not "forget" per se, as it was not necessarily needed but I decided to port it for two reasons:
- the vendor-level protection will be around for ~a year so there's a small chance that someone upgrading to v5 might break some of their users after deploying to a vendor that still has the protections active
- its going to be easier for us to debug any related issues that come up if both versions share the same implementation while people are upgrading

.changeset/pre.json
nikosdouvlis: I think this is going to happen in the next Vesion Packages PR even if we don't do anything :)

PR #2818, opened 2024-02-19
fix(backend): Align types based on FAPI/BAPI structs

nikosdouvlis:
## Description

Small type fixes based on the actual clerk_go types.
For the future reader, we're working internally to generate these automatically.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/api/resources/JSON.ts
+++ b/packages/backend/src/api/resources/JSON.ts
@@ -1,3 +1,5 @@
+import type { SamlAccountJSON } from '@clerk/types';
+
 import type {
   InvitationStatus,
   OrganizationInvitationStatus,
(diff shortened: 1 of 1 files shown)


packages/backend/src/api/resources/OrganizationMembership.ts
```diff
@@ -18,7 +18,7 @@ export class OrganizationMembership {
     return new OrganizationMembership(
       data.id,
       data.role,
-      data.public_metadata,
+      data.public_metadata || {},
```
SokratisVidros, line 21: Do we need this default value? My assumption is that the API always returns `{}`

SokratisVidros (approved):

packages/backend/src/api/resources/OrganizationMembership.ts
nikosdouvlis: There's an ongoing discussion on slack about this. I went through the backend code and I think that we always default to `{}` even if all struct use `json.RawMessage` which translates to `null | {}`

I will revert this one for now, however we need to keep in mind that we cannot simply add `{}` as a fallback in clerk_go, as in theory, people could be checking like `!metadata` which will return `false` for `{}`

packages/backend/src/api/resources/ExternalAccount.ts
```diff
@@ -28,7 +28,7 @@ export class ExternalAccount {
       data.email_address,
       data.first_name,
       data.last_name,
-      data.image_url,
+      data.image_url || '',
```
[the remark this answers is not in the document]
dimkl: I think it will always return a string (even of failures) it will return an empty string. We should update the OpenAPI spec and the BAPI/FAPI serializers (drop the `omitempty`).

PR #2845, opened 2024-02-22
fix(backend,shared): Emit telemetry notice in backend, not postinstall

jescalan:
## Description

Postinstall scripts no longer emit console logs, so we need to ensure the telemetry notice is shown in a different way for core 2.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/backend/src/telemetry-notification.ts
```diff
@@ -0,0 +1,80 @@
...
+    case 'darwin':
+      return macos();
+    case 'win32':
+      return win();
+    default:
+      return linux();
+  }
+}
+
+async function notifyAboutTelemetry() {
```
nikosdouvlis, line 50: @clerk/backend is an isomorphic package -  have we tested that this works across all runtimes, including CF workers, netlify etc?

nikosdouvlis: And a quick follow up: what's the plan for telemetry on the clerk-js side?

packages/backend/src/telemetry-notification.ts
```diff
@@ -0,0 +1,80 @@
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
```
nikosdouvlis, line 3: This is compatible with the node runtime only - it would break any apps not running on node. We need to make this logic isomorphic or think of a different way to show the telemetry notice.

For now, one other suggestion would be to simply show a shorter log (without the border) every time.

nikosdouvlis (changes requested):

packages/backend/src/telemetry-notification.ts
jescalan: I have not - just tested locally. Great catch! I'm not sure how we normally test into those runtimes, do you have any links to things I could read to learn about how to do so? I assume the main thing that could break down is the persistence of the config file here.

nikosdouvlis:
As discussed offline we're closing this PR for now for the reasons described in the comments above. We will move the warnings in the dashboard instead

PR #2857, opened 2024-02-26
fix(clerk-js,types): Always append db jwt as a searchParams

nikosdouvlis:
## Description
Append the devBrowser JWT to searchParams always in order to make v4 support both older v4 versions as well as v5 versions when a redirect flow is goes through AccountPortal

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/types/src/clerk.ts
```diff
@@ -411,7 +411,7 @@ export interface Clerk {
    * {@link Clerk.client.signIn.authenticateWithRedirect} or {@link Clerk.client.signUp.authenticateWithRedirect}
    */
   handleRedirectCallback: (
-    params: HandleOAuthCallbackParams | HandleSamlCallbackParams,
+    params: HandleOAuthCallbackParams,
```
dimkl, line 414: ❓ Why did we remove the `HandleSamlCallbackParams` ? I think it is intentional to keep both of them.

dimkl (approved):
☁️ With this change any site with pinned ClerkJS (using AP) that does not support cleaning the dev browser from both query params and hash will have all the dev browser stuck in the url forever.
Also we need to add some code to v5 to clean the dev browser from the hash. Currently the intention was to keep clean the v5 from any dev browser in hash logic.
If we are okay with this, we can proceed.

packages/types/src/clerk.ts
nikosdouvlis: This was automatically removed by eslint because the two types are the same - I will revert the change for now

PR #2858, opened 2024-02-26
fix(clerk-js,shared): Remove __clerk_db_jwt from hash on clerk-js init

nikosdouvlis:
Update the debBrowser handling logic to remove hash-based devBrowser JWTs from the URL. Even if v5 does not use the hash-based JWT at all, we still need to remove it from the URL in case clerk-js is initialised on a page after a redirect from an older clerk-js version, such as an AccountPortal using the v4 components

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/src/devBrowser.ts
+++ b/packages/shared/src/devBrowser.ts
@@ -19,20 +19,49 @@ export function setDevBrowserJWTInURL(url: URL, jwt: string): URL {
   return resultURL;
 }
 
-// Gets the dev_browser JWT from either the hash or the search
-// Side effect:
-// Removes dev_browser JWT from the URL as a side effect and updates the browser history
-export function getDevBrowserJWTFromURL(url: URL): string {
-  const resultURL = new URL(url);
-
-  // extract & strip existing jwt from search
-  const jwt = resultURL.searchParams.get(DEV_BROWSER_JWT_KEY) || '';
-  resultURL.searchParams.delete(DEV_BROWSER_JWT_KEY);
-
-  // eslint-disable-next-line valid-typeof
-  if (jwt && typeof globalThis.history !== undefined) {
-    globalThis.history.replaceState(null, '', resultURL.href);
+/**
+ * Gets the __clerk_db_jwt JWT from either the hash or the search
+ * Side effect:
+ * Removes __clerk_db_jwt JWT from the URL (hash and searchParams) and updates the browser history
+ */
+export function extractDevBrowserJWTFromURL(url: URL): string {
+  const jwt = readDevBrowserJwtFromSearchParams(url);
+  if (jwt && typeof globalThis.history !== 'undefined') {
+    globalThis.history.replaceState(null, '', removeDevBrowserJwt(url));
   }
-
   return jwt;
 }
+
+const readDevBrowserJwtFromSearchParams = (url: URL) => {
+  return url.searchParams.get(DEV_BROWSER_JWT_KEY) || '';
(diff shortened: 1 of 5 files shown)


packages/shared/src/devBrowser.ts
```diff
@@ -19,20 +19,48 @@ export function setDevBrowserJWTInURL(url: URL, jwt: string): URL {
...
+
+const removeDevBrowserJwt = (url: URL) => {
+  return removeDevBrowserJwtFromURLSearchParams(removeLegacyDevBrowserJwtFromURLHash(new URL(url)));
+};
+
+const removeDevBrowserJwtFromURLSearchParams = (_url: URL) => {
+  const url = new URL(_url);
+  url.searchParams.delete(DEV_BROWSER_JWT_KEY);
+  return url;
+};
```
dimkl, line 47: ☁️ I think we should follow the same pattern in our methods. I think we should adhere to the pattern where we construct a new instance and return the updated cloned instance instead of mutating the parameters passed.
This is a preferred pattern as it helps with improving visibility of the changes instead of hiding them which could cause unwanted side-effects and hard-to-find bugs.

`removeDevBrowserJwtFromURLSearchParams` returns a new url instance with the updated query params vs
`removeLegacyDevBrowserJwtFromURLHash` mutates the existing url instance.

packages/shared/src/devBrowser.ts
```diff
@@ -19,20 +19,48 @@ export function setDevBrowserJWTInURL(url: URL, jwt: string): URL {
   return resultURL;
 }
 
-// Gets the dev_browser JWT from either the hash or the search
-// Side effect:
-// Removes dev_browser JWT from the URL as a side effect and updates the browser history
-export function getDevBrowserJWTFromURL(url: URL): string {
```
dimkl, line 25: 🔧 This is a `@clerk/shared` public API method. Renaming it is a breaking change for the shared package. I would expect a major bump or at least a minor.

dimkl (approved):

packages/shared/src/devBrowser.ts
nikosdouvlis: 1. We're treating the `shared` package as an internal, low-level package
2. This change goes into `main` so this is the v5 branch anyway
3. The packages use fixed versions, meaning that you cannot have a Clerk package that depends on `@clerk/shared@^5` (notice the caret)

packages/shared/src/devBrowser.ts
nikosdouvlis: Good point, I updated the code 👍🏻

packages/shared/src/devBrowser.ts
```diff
@@ -19,20 +19,49 @@ export function setDevBrowserJWTInURL(url: URL, jwt: string): URL {
...
-
   return jwt;
 }
+
+const readDevBrowserJwtFromSearchParams = (url: URL) => {
+  return url.searchParams.get(DEV_BROWSER_JWT_KEY) || '';
+};
+
+const removeDevBrowserJwt = (url: URL) => {
+  return removeDevBrowserJwtFromURLSearchParams(removeLegacyDevBrowserJwtFromURLHash(new URL(url)));
```
dimkl, line 40: 🙃 Since both functions make a copy of the url passed there is no need to pass it as `new URL(url)`.

PR #2859, opened 2024-02-26
fix(clerk-js): Cleanup input related styles for polished appearance 

octoper:
## Description

This PR makes `polishedAppearance` respect the variants of inputs and added 2 new primitives.

**Primitives added:**
- `CheckboxInput`
- `RadioInput`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


octoper:
!preview

clerk-cookie:

| :-- | :-- | :-- |
| :cook: _Deploying..._ | [Inspect](https://javascript-preview-k5fvryex0.clerkstage.dev) | Feb 26, 2024 03:45 PM |

anagstef:
!preview

packages/clerk-js/src/ui/customizables/index.ts
```diff
@@ -32,6 +32,12 @@ export const Image = makeCustomizable(sanitizeDomProps(makeResponsive(Primitives
 export const Alert = makeCustomizable(sanitizeDomProps(Primitives.Alert));
 export const AlertIcon = makeCustomizable(sanitizeDomProps(Primitives.AlertIcon));
 export const Input = makeCustomizable(sanitizeDomProps(Primitives.Input), { defaultDescriptor: descriptors.input });
+export const CheckboxInput = makeCustomizable(sanitizeDomProps(Primitives.Input), {
+  defaultDescriptor: descriptors.checkboxInput,
+});
```
anagstef, line 37: I'd expect that this component includes properties that come with a checkbox, e.g: `type='checkbox'`

packages/clerk-js/src/ui/customizables/index.ts
```diff
@@ -32,6 +32,12 @@ export const Image = makeCustomizable(sanitizeDomProps(makeResponsive(Primitives
...
+export const RadioInput = makeCustomizable(sanitizeDomProps(Primitives.Input), {
+  defaultDescriptor: descriptors.radioInput,
+});
```
anagstef, line 40: Also, here: `type='radio'`

packages/clerk-js/src/ui/customizables/index.ts
octoper: We can't pass props like `type="checkbox"` here, the `makeCustomizable` accepts only some generic options `defaultDescriptor` and `defaultStyles`.

packages/clerk-js/src/ui/elements/RadioGroup.tsx
```diff
@@ -18,7 +18,6 @@ const RadioIndicator = forwardRef<HTMLInputElement, { value: string; id: string
       sx={t => ({
         width: 'fit-content',
         marginTop: t.space.$0x5,
-        boxShadow: 'none',
       })}
       type='radio'
```
desiprisg, line 22: I wouldn't expect this to be needed here.

packages/types/src/appearance.ts
```diff
@@ -135,6 +135,8 @@ export type ElementObjectKey<K extends string> = K extends `${infer Parent}-${in
 export type ElementsConfig = {
   button: WithOptions;
   input: WithOptions;
+  checkboxInput: WithOptions;
+  radioInput: WithOptions;
```
desiprisg, line 139: Can we get rid of the `Input` part?

packages/clerk-js/src/ui/elements/RadioGroup.tsx
octoper: Yeah I have removed that!

packages/types/src/appearance.ts
octoper: Yeah, good call!

anagstef (approved):

packages/clerk-js/src/ui/primitives/CheckboxInput.tsx
```diff
@@ -0,0 +1,16 @@
+import type { ComponentProps } from 'react';
+import React from 'react';
+
+import { Input } from './Input';
+
+export type CheckboxInputProps = ComponentProps<typeof Input>;
+
+export const CheckboxInput = React.forwardRef<HTMLInputElement, CheckboxInputProps>((props, ref) => {
```
desiprisg, line 8: Can't we avoid the extra primitive here by cloning the `Input` and passing the extra prop at `index.ts` (same for checkbox)?

packages/clerk-js/src/ui/polishedAppearance.ts
```diff
@@ -131,50 +131,19 @@ export const polishedAppearance: Appearance = {
         },
       },
       input: {
-        ...inputStyles(theme),
-        '&[type="radio"]': {
-          boxShadow: 'none',
-          '&:hover': {
-            boxShadow: 'none',
-          },
+        '&[data-variant="default"]': {
```
desiprisg, line 134: Should we do a `&:not([data-variant="unstyled"])` to make this more future proof as well as a better example for usage in other themes? Does it make more sense to you @octoper ?

desiprisg: Actually, should we just rename `unstyled` to something that makes more sense? Can we find common styles by both elements (search & tag input) and group them in a variant? Right now, we pass inline styles to both and the variant has no use, so targeting `unstyled` doesn't really make sense for any customer.

desiprisg: I know this isn't part of the PR so we can do it separately if you don't want to include something here.

desiprisg (approved):

packages/clerk-js/src/ui/primitives/CheckboxInput.tsx
octoper: Just added `defaultProps` option for `makeCustomizable` so we can pass default props to any primitive.

anagstef:
!preview

packages/clerk-js/src/ui/customizables/index.ts
```diff
@@ -32,6 +32,18 @@ export const Image = makeCustomizable(sanitizeDomProps(makeResponsive(Primitives
 export const Alert = makeCustomizable(sanitizeDomProps(Primitives.Alert));
 export const AlertIcon = makeCustomizable(sanitizeDomProps(Primitives.AlertIcon));
 export const Input = makeCustomizable(sanitizeDomProps(Primitives.Input), { defaultDescriptor: descriptors.input });
+export const CheckboxInput = makeCustomizable(sanitizeDomProps(Primitives.Input), {
+  defaultDescriptor: descriptors.checkbox,
+  defaultProps: {
+    type: 'checkbox',
+  },
```
anagstef, line 39: Why did we add the `defaultProps` functionality here? I liked the concept we had the last time I reviewed. That is, we have a `CheckboxInput` element that is an `Input` with a descriptor and `type='checkbox'`. Similar to what we do with `Flex` and `Col`.

octoper: Yeah I did that to avoid having another file, and because we already had `defaultStyles` and `defaultDescriptor` i thought that was a good addition.

desiprisg (approved):
LGTM!

PR #2898, opened 2024-02-29
chore(*): Expose requestState.headers to response and populate debug headers only in backend

dimkl:
## Description

This refactoring task is to populate debug headers in `@clerk/backend` and expose them in the dependent SDKs.
Up until now, we are re-using internals to populate in each SDK the debug headers.
The debug headers are the : status, message, reason

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/lemon-starfishes-buy.md
+++ b/.changeset/lemon-starfishes-buy.md
@@ -1,0 +1,9 @@
+---
+'@clerk/clerk-sdk-node': minor
+'@clerk/backend': minor
+'@clerk/fastify': minor
+'@clerk/nextjs': minor
+'@clerk/remix': minor
+---
+
+Expose debug headers in response for handshake / signed-out states from SDKs using headers returned from `authenticateRequest()`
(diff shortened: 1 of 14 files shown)


panteliselef:
@dimkl Is there any specific reason we are not applying `withDebugHeaders `  to the signed in state as well ?

I believe up until now if we a user was signed in we would have the header for auth status equal to "signed-in".

dimkl:
> @dimkl Is there any specific reason we are not applying `withDebugHeaders ` to the signed in state as well ?
> 
> I believe up until now if we a user was signed in we would have the header for auth status equal to "signed-in".

Yes there is.
The `reason` & `message` are only populated in the signed-out / handshake cases. Also we don't need to debug the signed-in state since it's our happy path.

nikosdouvlis (approved):
👏🏻

PR #2918, opened 2024-03-02
feat(express): Introduce Clerk package for Express

dimkl:
## Description

Introduce Express specific Clerk SDK `@clerk/express`. The SDK exposes the following API:

- clerkClient: default @clerk/backend client initialized from env variables and used to make BAPI requests
- clerkMiddleware: centralized middleware that authenticates all requests without blocking them (also triggers handshake mechanism)
- getAuth: utility to retrieve the auth state from the request (requires clerkMiddleware to executed)
- protect: middleware that returns http 401 response when request is signed-out

Also all the top level exports from `@clerk/backend` are re-exported from `@clerk/express` top level.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## TODOs

- [ ] Add tests
- [ ] Add examples
- [ ] Replace all `TODO(express)` TODOs
- [ ] Use this package to create separate page in [backend-request/handling](https://docs-preview-605.clerkpreview.com/docs/backend-requests/handling/nodejs) docs


.changeset/bright-gorillas-crash.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+Introduce Express specific Clerk SDK `@clerk/express`. The SDK exposes the following API:
+
+- clerkClient: default @clerk/backend client initialized from env variables and used to make BAPI requests
+- clerkMiddleware: centralized middleware that authenticates all requests without blocking them (also triggers handshake mechanism)
+- getAuth: utility to retrieve the auth state from the request (requires clerkMiddleware to executed)
+- protect: middleware that returns http 401 response when request is signed-out
+
+Also all the top level exports from `@clerk/backend` are re-exported from `@clerk/express` top level.
```
LekoArts, line 12:
```suggestion
Introduce [Express](https://expressjs.com/) specific Clerk SDK called `@clerk/express`. The SDK exposes the following APIs:

- `clerkClient`: Default [`@clerk/backend`](https://clerk.com/docs/references/backend/overview) client initialized from environment variables and used to make backend API requests
- `clerkMiddleware`: Centralized middleware that authenticates all requests without blocking them (also triggers handshake mechanism)
- `getAuth`: Utility to retrieve the auth state from a request (requires `clerkMiddleware` to be executed)
- `protect`: Middleware that returns HTTP 401 response when request is signed-out

Also all the top level exports from `@clerk/backend` are re-exported from `@clerk/express`.
```

packages/express/CHANGELOG.md
LekoArts, line 1: This needs to be an empty changelog

packages/express/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022 Clerk, Inc.
```
LekoArts, line 3:
```suggestion
Copyright (c) 2024 Clerk, Inc.
```

packages/express/README.md
```diff
@@ -0,0 +1,102 @@
...
+</div>
+
+---
+
+## Overview
+
+<!-- TODO(express): Update express/getting-started link or create page -->
+
+[Clerk](https://clerk.com?utm_source=github&utm_medium=express) is the easiest way to add authentication and user management to your Node.js application. To gain a better understanding of the Clerk Backend API and SDK, refer to
+the <a href="https://clerk.com/docs/reference/express/getting-started?utm_source=github&utm_medium=express" target="_blank">Node SDK</a> and <a href="https://clerk.com/docs/reference/backend-api" target="_blank">Backend API</a> documentation.
```
LekoArts, line 36:
```suggestion
[Clerk](https://clerk.com?utm_source=github&utm_medium=express) is the easiest way to add authentication and user management to your Node.js application. To gain a better understanding of the Express SDK and Clerk Backend API, refer to
the <a href="https://clerk.com/docs/reference/express/getting-started?utm_source=github&utm_medium=express" target="_blank">Express SDK</a> and <a href="https://clerk.com/docs/reference/backend-api" target="_blank">Backend API</a> documentation.
```

packages/express/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Getting started
+
+### Prerequisites
+
+- Node.js `>=18.17.0` or later
```
LekoArts, line 42:
```suggestion
- Node.js `>=18.17.0` or later
- Express installed (follow their [Getting started](https://expressjs.com/en/starter/installing.html) guide)
```

packages/express/jest.config.js
```diff
@@ -0,0 +1,17 @@
+const { name } = require('./package.json');
+
+module.exports = {
+  displayName: name.replace('@clerk', ''),
+  globals: {
+    PACKAGE_NAME: '@clerk/clerk-sdk-node',
```
LekoArts, line 6:
```suggestion
    PACKAGE_NAME: '@clerk/express',
```

packages/express/package.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "@clerk/express",
+  "version": "0.0.1",
```
LekoArts, line 3: I'd publish at 1.0.0 and not in the zero range. Semver is weird in the 0 range and I don't think we need to keep it in that range since the API seems to be safe and sound.

packages/express/src/clerkClient.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { loadApiEnv, loadClientEnv } from './utils';
+
+let clerkClientSingleton = {} as unknown as ReturnType<typeof createClerkClient>;
+
+export const clerkClient = new Proxy(clerkClientSingleton, {
+  get(_target, property: string) {
+    const hasBeenInitialised = !!clerkClientSingleton.authenticateRequest;
+    if (hasBeenInitialised) {
+      // @ts-expect-error - Element implicitly has an 'any' type because expression of type 'string | symbol' can't be used to index type 'ExtendedClerk'.
```
LekoArts, line 11: It would be nice to fix these errors 👍

packages/express/src/errors.ts
```diff
@@ -0,0 +1,25 @@
...
+const createErrorMessage = (msg: string) => {
+  return `🔒 Clerk: ${msg.trim()}
+  
+  For more info, check out the docs: https://clerk.com/docs,
+  or come say hi in our discord server: https://clerk.com/discord
+  `;
+};
+
+export const middlewareRequired =
+  createErrorMessage(`The "clerkMiddleware" should be registered before using the "getAuth".
```
LekoArts, line 11:
```suggestion
  createErrorMessage(`The "clerkMiddleware" should be registered before using "getAuth".
```

packages/express/src/getAuth.ts
```diff
@@ -0,0 +1,13 @@
+import type { AuthObject } from '@clerk/backend/internal';
+import type { Request as ExpressRequest } from 'express';
+
+import { middlewareRequired } from './errors';
+import { isAuthInRequest } from './utils';
+
+export const getAuth = (req: ExpressRequest): AuthObject => {
```
LekoArts, line 7: JS docs comments would be nice

packages/express/src/protect.ts
```diff
@@ -0,0 +1,35 @@
...
+    const auth = getAuth(request);
+    if (!auth.has('perm')) {
+      reply.status(403).send('Forbidden');
+      return;
+    }
+    return next();
+  }
+ * router.get('/path', protect(hasPermission), getHandler)
+ */
+export const protect = (handler?: RequestHandler): RequestHandler[] => {
```
LekoArts, line 21: The JSDoc comments should also have a description in addition to the examples

packages/express/src/clerkMiddleware.ts
```diff
@@ -0,0 +1,43 @@
...
+import { defaultHandler } from './utils';
+
+const parseHandlerAndOptions = (args: unknown[]) => {
+  return [
+    typeof args[0] === 'function' ? args[0] : undefined,
+    (args.length === 2 ? args[1] : typeof args[0] === 'function' ? {} : args[0]) || {},
+  ] as [RequestHandler | undefined, ClerkMiddlewareOptions];
+};
+
+export const clerkMiddleware: ClerkMiddleware = (...args: unknown[]): RequestHandler[] => {
```
LekoArts, line 16: JSDoc description here would be great

LekoArts (changes requested):
First pass at looking through the changes :)

packages/express/src/clerkMiddleware.ts
dimkl: I have added JSDoc in the `ClerkMiddleware` definition, just like we did in our `@clerk/nextjs` package. Do you think we should move those to the actual implementation?
Our nextjs JSDocs, used as example: https://github.com/clerk/javascript/blob/4565ef29e67d4e39f14c7e61c71fb411fab9aa8e/packages/nextjs/src/server/clerkMiddleware.ts#L41-L63

dimkl: ty for the feedback. I will move it and add a ticket to move this for our nextjs package.
cc: @nikosdouvlis

dimkl: @LekoArts Checking again your screenshot above,  we can see that there is an `@example` so i guess that the JSDoc work.

LekoArts: I'd consider that example useless as it just shows you what you're already using at the moment, not the other options or more details.
It also doesn't have any description or the params explained.

If you put the description and all three examples onto the function, not the type the IntelliSense will be better

dimkl: I have moved the JSDocs above code definition.
Screenshot 2024-03-15 at 15 53 28

dimkl:
@LekoArts, @nikosdouvlis Could you take another look at this PR?

packages/express/src/errors.ts
```diff
@@ -0,0 +1,29 @@
+// https://github.com/clerk/javascript/blob/main/packages/remix/src/errors.ts#L1-L0
```
LekoArts, line 1: That comment isn't helpful, is it?

packages/express/src/requireAuth.ts
```diff
@@ -0,0 +1,39 @@
+import type { RequestHandler } from 'express';
+
+import { middlewareRequired } from './errors';
+import { getAuth } from './getAuth';
+import { isAuthInRequest } from './utils';
+
+/**
+ * Middleware to requireAuth requests for user authenticated or authorized requests.
```
LekoArts, line 8: Seems like a typo?

```suggestion
 * Middleware to require auth requests for user authenticated or authorized requests.
```

packages/express/src/types.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { AuthenticateRequestOptions, AuthObject } from '@clerk/backend/internal';
+import type { Request as ExpressRequest, RequestHandler } from 'express';
+
+export type ExpressRequestWithAuth = ExpressRequest & { auth: AuthObject };
+
+export type ClerkMiddlewareOptions = AuthenticateRequestOptions & { debug?: boolean };
+
+/**
+ * Middleware for Express that handles authentication and authorization with Clerk.
+ * For more details, please refer to the docs: https://beta.clerk.com/docs/backend-requests/handling/express
```
LekoArts, line 12: In other places we link to `https://beta.clerk.com/docs/references/express/overview?utm_source=github&utm_medium=express`, should we also use this here?

LekoArts (approved):
Overall LGTM, some minor comments I'd like to resolve before merging

packages/express/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+[![Chat on Discord](https://img.shields.io/discord/856971667393609759.svg?logo=discord)](https://clerk.com/discord)
+[![Clerk documentation](https://img.shields.io/badge/documentation-clerk-green.svg)](https://beta.clerk.com/docs?utm_source=github&utm_medium=express)
+[![Follow on Twitter](https://img.shields.io/twitter/follow/ClerkDev?style=social)](https://twitter.com/intent/follow?screen_name=ClerkDev)
+
+[Changelog](https://github.com/clerk/javascript/blob/main/packages/express/CHANGELOG.md)
+·
+[Report a Bug](https://github.com/clerk/javascript/issues/new?assignees=&labels=needs-triage&projects=&template=BUG_REPORT.yml)
+·
+[Request a Feature](https://github.com/clerk/javascript/issues/new?assignees=&labels=feature-request&projects=&template=FEATURE_REQUEST.yml)
```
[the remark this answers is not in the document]
dimkl: Nice catch! i will change this in this PR and open a ticket to replace it in all other Readme files

packages/express/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+```sh
+npm install @clerk/express
+```
+
+To build the package locally with the TypeScript compiler, run:
+
+```sh
+npm run build
+```
```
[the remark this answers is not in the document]
dimkl: If we remove it from here, we should probably remove it from the rest of the README files. @LekoArts Wdyt?

packages/express/src/errors.ts
dimkl: it's a indication that the `createErrorMessage` was copied from the remix package in order to replace it with a shared util but i will remove it since we will probably tackle the error handling in a common way for all packages so there is no need to check this use that comment.

packages/express/README.md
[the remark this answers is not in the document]
LekoArts: We can update the READMEs in another PR 👍

packages/express/src/authenticateRequest.ts
```diff
@@ -0,0 +1,97 @@
+import type { RequestState } from '@clerk/backend/internal';
+import { AuthStatus, createClerkRequest } from '@clerk/backend/internal';
+import { handleValueOrFn } from '@clerk/shared/handleValueOrFn';
+import { isDevelopmentFromSecretKey } from '@clerk/shared/keys';
+import { isHttpOrHttps, isProxyUrlRelative, isValidProxyUrl } from '@clerk/shared/proxy';
+import type { Response } from 'express';
```
nikosdouvlis, line 6: ❓ How is this different than the native `Response` class?

packages/express/src/clerkClient.ts
```diff
@@ -0,0 +1,28 @@
+import type { ClerkClient } from '@clerk/backend';
+import { createClerkClient } from '@clerk/backend';
+
+import { loadApiEnv, loadClientEnv } from './utils';
+
+let clerkClientSingleton = {} as unknown as ClerkClient;
+
+export const clerkClient = new Proxy(clerkClientSingleton, {
```
nikosdouvlis, line 8: Can you please remind me why we need the `Proxy` here?

packages/express/src/clerkClient.ts
```diff
@@ -0,0 +1,28 @@
...
+  get(_target, property: keyof ClerkClient) {
```
nikosdouvlis, line 9: Can you please remind me why we need the Proxy here? Is it because of the SK/PK errors we throw during initialization? 

I'd feel much more comfortable if we could remove the Proxy, no worries if its blocking though. However, we should figure out how we can avoid the need for workarounds like this

packages/express/src/clerkMiddleware.ts
```diff
@@ -0,0 +1,59 @@
...
+    typeof args[0] === 'function' ? args[0] : undefined,
+    (args.length === 2 ? args[1] : typeof args[0] === 'function' ? {} : args[0]) || {},
+  ] as [RequestHandler | undefined, ClerkMiddlewareOptions];
+};
+
+export const clerkMiddleware: ClerkMiddleware = (...args: unknown[]) => {
+  const [handler, options] = parseHandlerAndOptions(args);
+
+  const shouldInitializeClerkClient = Object.keys(options).length > 0;
+  const clerkClient = shouldInitializeClerkClient ? createClerkClient(options) : defaultClerkClient;
```
nikosdouvlis, line 27: ❓ Why don't we switch to an injection logic here?
Instead of initializing a clerk client, wouldn't it be better to ask the users to inject an instance? That way we can remove the duplicated logic and option props from here.

nikosdouvlis (approved):

packages/express/src/clerkClient.ts
dimkl: We are using the Proxy to lazy load the envs when the `clerkClient` is actually used instead of loading the envs at the time of the `cerkClient` import.
Before we introduced the Proxy the customers using the `@clerk/clerk-sdk-node` had to import dotenv before importing the `@clerk/clerk-sdk-node`. Also due to autoformatting of files with eslint, the dotenv is moved after the clerk-sdk-node which was throwing errors of uninitialized pk/sk.

packages/express/src/clerkMiddleware.ts
dimkl: We could do that but with this approach the customers don't need to use the `createClerkClient` (which is a low-level function re-exported from `@clerk/backend`) to avoid using envs for some reason.
What do you think will be the pros / cons of injecting a `clerkClient`?

packages/express/src/authenticateRequest.ts
dimkl: It seems they are different types:
Screenshot 2024-04-04 at 22 24 55

packages/express/src/clerkMiddleware.ts
nikosdouvlis: All low-level apis were moved to sub paths of clerk/backend, so I wouldn't consider `createClerkClient` a low-level API right now -  this is the preferred way to use Clerk in CF workers and similar runtimes after all. 

As for injection vs us automatically initializing a client, it might be personal preference, but I think that:
- It's more explicit for the user 
- A method called `clerkMiddleware` should probably not be concerned with creating a clerk client at all. 

The 
```
  const shouldInitializeClerkClient = Object.keys(options).length > 0;
```
makes me think that `clerkMiddlware` does more than it should, as without the named variable I wouldn't be able to understand what `Object.keys(options).length > 0` means

packages/express/src/requireAuth.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @example
+ * hasPermission = (request, response, next) => {
+ *    const auth = getAuth(request);
+ *    if (!auth.has({ permission: 'permission' })) {
+ *      response.status(403).send('Forbidden');
+ *      return;
+ *    }
+ *    return next();
+ * }
+ * router.get('/path', requireAuth, hasPermission, getHandler)
```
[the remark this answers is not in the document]
nikosdouvlis: We could introduce a new protect api in the future similar to auth().protect() -  no need to change anything now though :)

packages/express/src/clerkMiddleware.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  const middleware: RequestHandler = async (request, response, next) => {
+    try {
+      const requestState = await authenticateRequest({
+        clerkClient,
+        request,
+        options,
+      });
```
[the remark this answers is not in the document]
dimkl: resolved offline. `setResponseHeaders` is responsible to inject the headers from authenticateRequest and set handshake redirect. Any content-type response is supported.

packages/express/src/clerkMiddleware.ts
dimkl: I have updated the clerkMiddleware to support `clerkClient` optional instance instead of implicitly creating one.

packages/express/src/clerkClient.ts
dimkl: We are using this pattern in the `@clerk/clerk-sdk-node`. We will discuss it in future refactorings.

PR #2927, opened 2024-03-04
feat(backend): Signing a JWT respects 'iat' claim

gkats:
## Description
Calling the `signJwt` function overrides the 'iat' (issued at) claim of the payload.

```js
import { signJwt } from '@clerk/backend/jwt';

const key = 'the-private-key';
const payload = {
  sub: 'subject',
  // provide the issued_at claim
  iat: 123,
}
const { data } = await signJwt(payload, key, {
  algorithm: 'RS256',
  header: {
    alg: 'RS256',
    typ: 'JWT',
  }
});
data.iat // => returns the current time instead of 123.
```

This PR fixes the problem. 

When calling the signJwt function to generate a JWT the value of the 'iat' claim in the payload will be respected, falling back to the current time if the claim is not present in the payload.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/rare-chefs-invent.md
+++ b/.changeset/rare-chefs-invent.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
+The signJwt function preserves the value of the 'iat' claim, falling back to the current time if the claim is not present in the payload
--- a/packages/backend/src/jwt/__tests__/signJwt.test.ts
+++ b/packages/backend/src/jwt/__tests__/signJwt.test.ts
@@ -47,5 +47,39 @@ export default (QUnit: QUnit) => {
       const { data: verifiedPayload } = await verifyJwt(data, { key: pemEncodedPublicKey });
       assert.deepEqual(verifiedPayload, payload);
     });
+
+    test("respects the provided 'iat' claim value", async assert => {
+      // Pass a value for the 'iat' claim
+      const iat = 123;
+      const { data } = await signJwt(
+        {
+          ...payload,
+          iat,
+        },
+        signingJwks,
+        {
+          algorithm: mockJwtHeader.alg,
+          header: mockJwtHeader,
+        },
+      );
+      assertOk(assert, data);
+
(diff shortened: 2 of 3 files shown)


Nikpolik (approved):
❓ Is there any scenario where we want to have a IAT older then when the token was signed?

Nikpolik:
> > Is there any scenario where we want to have a IAT older then when the token was signed?
> 
> Please allow me to reverse the question a bit. Is there any reason why the caller should not be able to specify the 'iat' claim?
> 
> I expect that it's possible to generate tokens with whatever value we want for the 'iat' claim.
> 
> The ability to set the "issued at" claim could also prevent clock discrepancies between nodes of a distributed system, by guaranteeing that all chronological values are generated from a trusted/central time source.

Yep this makes 100% sense. I guess ensuring that no invalid IAT is added should be part of the code generating the JWT in the first place and not when signing it!

colinclerk:
Clerk's SDK (in particular the Handshake and Interstitial mechanism) depend on the iat claim using the same clock as the __client_uat cookie.  What is motivating the change?

gkats:
> What is motivating the change?

@colinclerk I'd like to reuse the clerk/backend package for generating JWTs at Edge. I would expect that the caller should set the 'iat' claim based on their own "clock". Our FAPI currently sets 'iat' to a value with its own clock for example. 

> Clerk's SDK (in particular the Handshake and Interstitial mechanism) depend on the iat claim using the same clock as the __client_uat cookie.

I see. So the Clerk SDK cannot accept a clock and relies on `Date` on each platform that the code is executed, right? If I understand correctly you are suggesting that callers _should not_ set the 'iat' claim on their own.

I find that kind to be kind of backwards. 

In any case this change allows you to either pass the 'iat' claim or default to `Date.now()` which should preserve the existing behavior of the library.

gkats:
After discussing with @colinclerk and @nikosdouvlis we decided to change the internal implementation of `signJwt`. 

Closing this PR until then.

PR #2942, opened 2024-03-06
fix(nextjs): Middleware protect detects internal navigation in pages router

panteliselef:
## Description

Using auth().protect inside clerkMiddleware will perform a redirection instead of throwing a not found error when internal navigation in pages router occurs and the user is unauthenticated.

Original issue:
Case1: When auth().redirectToSignIn() is used

signing out on the client-side triggers the middleware and the middleware returns a 307 redirect to /sign-in . This is the expected behavior

Case2: When auth().protect() is used

event though signing out on the client-side triggers the middleware, protect() treats the request as an api/route handler request and returns 404, so the redirect does not happen. This should be like case 1.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/quiet-kings-relate.md
+++ b/.changeset/quiet-kings-relate.md
@@ -1,0 +1,5 @@
+---
+'@clerk/nextjs': patch
+---
+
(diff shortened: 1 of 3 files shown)


nikosdouvlis (approved):
👏🏻

PR #2945, opened 2024-03-06
chore(shared): Convert shared package to ES Modules

arjunyel:
## Description

Now that Remix is switching to Vite, ideally all packages used with it are in ES Module format. This is the first step to converting all the Clerk packages to ESM. The module format should be standardized to `NodeNext` as that is the recommended option for libraries.

To land this I need help with one issue, the `clerk-js` tests are failing with `TypeError: (0 , import_swr.default) is not a function at usePagesOrInfinite (/javascript/packages/shared/src/react/hooks/usePagesOrInfinite.ts:102:7)`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/.eslintrc.cjs
+++ b/packages/shared/.eslintrc.cjs
@@ -1,6 +1,7 @@
 module.exports = {
   root: true,
   extends: ['custom/browser', 'custom/typescript', 'custom/jest', 'custom/react'],
+  ignorePatterns: ['jest.config.js'],
   overrides: [
     {
       files: ['./src/utils/globs.ts'],
--- a/packages/shared/jest.config.js
+++ b/packages/shared/jest.config.js
@@ -1,8 +1,12 @@
-const { name } = require('./package.json');
+import packageJson from './package.json' assert { type: 'json' };
 
 /** @type {import('ts-jest').JestConfigWithTsJest} */
 const config = {
-  displayName: name.replace('@clerk', ''),
+  preset: 'ts-jest/presets/default-esm',
+  moduleNameMapper: {
+    '(.+)\\.js': '$1',
+  },
+  displayName: packageJson.name.replace('@clerk', ''),
   injectGlobals: true,
 
   testEnvironment: 'jsdom',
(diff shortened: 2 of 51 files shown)


LekoArts:
Hi, thanks for your PR!

While it is true that the ecosystem is _slowly_ moving towards ESM-only, it's far, far away from that (ideal) future. You can check some progress here for example: https://github.com/wooorm/npm-esm-vs-cjs

Clerk's packages, and `@clerk/shared` in particular, are used in different runtimes, environments, and with different bundlers. As such, Clerk's packages must be distributed as dual packages (CJS + ESM) since we can't switch to ESM-only yet.

If you're seeing problems with our packages in e.g. Remix + Vite, please open a bug report with a reproduction and we can look into it.

Again, thanks for the PR, I can personally empathize with wanting to switch things to ESM but unfortunately we have to support CJS for a little while longer.

arjunyel:
@LekoArts hey friend, this did not remove the cjs part, it just enabled the ESM part :) so CJS environments would still pull the CJS files

arjunyel:
If your look at the build output ESM is .js and CJS is .CJS, type: module lets runtimes know the .js files are ESM, they still can get the CJS if they need to. Also the packages should all be standardized on NodeNext modules because that's what's recommended for libraries

LekoArts:
The ESM part should already work as we use `.mjs` files without a `type: "module"`. Adding `type: "module"` is a breaking change and requires changing our bundling setup to output all CJS files with `.cjs` extension. A dual setup is a real headache and right now it works so if we don't have to we can keep it as-is. As per https://blog.isquaredsoftware.com/2023/08/esm-modernization-lessons/

> Alternately, if you don't have type: "module", .js files are treated as CommonJS. You can use .mjs to mark individual files as ESM.

That's what we do.

For the `tsconfig.json` part we're following this recommendation: https://www.totaltypescript.com/tsconfig-cheat-sheet#not-transpiling-with-typescript

LekoArts:
Thanks! I gave it a read and updated an internal ticket that will track updating our `tsconfig.json` files - because currently they are different in each package and eventually we'll want to align them 👍

arjunyel:
So can we merge this PR if I just take out type: module? :)

LekoArts:
We want to keep things as-is for now since they are working but appreciate you for wanting to help! Once we'd be up for switching things over I'll make an effort to open this up for OSS users :)

PR #2949, opened 2024-03-07
fix(clerk-js): Remove ticket params after they are consumed

desiprisg:
## Description
Remove the **clerk_ticket and **clerk_invitation_token query params after they are consumed.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
+++ b/packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
@@ -2,7 +2,7 @@ import { useClerk } from '@clerk/shared/react';
 import React from 'react';
 
 import { ERROR_CODES } from '../../../core/constants';
-import { getClerkQueryParam } from '../../../utils/getClerkQueryParam';
+import { getClerkQueryParam, removeClerkQueryParam } from '../../../utils/getClerkQueryParam';
 import { buildSSOCallbackURL, withRedirectToAfterSignUp } from '../../common';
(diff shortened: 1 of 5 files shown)


packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
```diff
@@ -109,6 +109,8 @@ function _SignUpStart(): JSX.Element {
     signUp
       .create({ strategy: 'ticket', ticket: formState.ticket.value })
       .then(signUp => {
+        removeClerkQueryParam('__clerk_ticket');
+        removeClerkQueryParam('__clerk_invitation_token');
```
nikosdouvlis, line 113: Is this the only place where the ticket can be consumed? 
Don't we need this to be moved in the resources themselves? Think about users using clerk-js directly :)

nikosdouvlis: Also, don't forget the unit tests

desiprisg: The resource accepts the ticket as a param, it does not read it through the URL.

nikosdouvlis: Cool - let's add a unit test and merge

nikosdouvlis (approved):

PR #1680, opened 2023-09-04
refactor(clerk-js): Pick the redirection prop using a more intuitive approach

desiprisg:
## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Packages affected

- [ ] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [ ] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [ ] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `@clerk/chrome-extension`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

- [x] `npm test` runs as expected.
- [x] `npm run build` runs as expected.

This is a proposed refactor to make the choosing of the redirect url a more intuitive process. The new function accepts keys and the sources to choose them from, in order. A validator and formatter can also be added.

Initially, this was done as part of a PR that also added some new query params for org actions, but since more logic has been added by the orgs team, I'll leave that up to them to implement. It should be relativelt easy to do after this refactor.

PR #2955, opened 2024-03-07
fix(clerk-js,types): Fix the redirection that happens when SSO fails

anagstef:
## Description

Current Behaviour:
If a SignIn or SignUp SSO fails and needs to redirect back to the component, it always redirects to the Account Portal.

Expected Behaviour:
If the component is mounted on the user's app, then it should redirect back to the user's app.

Video coming soon...

Tech Details:
This happens because in the `handleRedirectCallback` we extract the `signInUrl` and `signUpUrl` from the displayConfig, without taking into account the env vars or the props that the user has passed.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
+++ b/packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
@@ -91,7 +91,9 @@ export const useSignUpContext = (): SignUpContextType => {
     displayConfig: displayConfig,
   });
 
-  let signUpUrl = pickRedirectionProp('signUpUrl', { ctx, options, displayConfig }, false);
+  let signUpUrl =
+    (ctx.routing === 'path' ? ctx.path : undefined) ||
+    pickRedirectionProp('signUpUrl', { options, displayConfig }, false);
   if (authQs && ctx.routing !== 'virtual') {
     signUpUrl += `#/?${authQs}`;
   }
@@ -175,7 +177,9 @@ export const useSignInContext = (): SignInContextType => {
     signUpUrl += `#/?${authQs}`;
   }
 
-  let signInUrl = pickRedirectionProp('signInUrl', { ctx, options, displayConfig }, false);
+  let signInUrl =
+    (ctx.routing === 'path' ? ctx.path : undefined) ||
+    pickRedirectionProp('signInUrl', { options, displayConfig }, false);
   if (authQs && ctx.routing !== 'virtual') {
     signInUrl += `#/?${authQs}`;
   }
--- a/packages/elements/src/internals/machines/third-party/actors.ts
+++ b/packages/elements/src/internals/machines/third-party/actors.ts
@@ -88,6 +88,8 @@ export const handleRedirectCallback = fromCallback<AnyEventObject, HandleRedirec
         secondFactorUrl: ClerkJSNavigationEvent.signIn,
         verifyEmailAddressUrl: ClerkJSNavigationEvent.verification,
         verifyPhoneNumberUrl: ClerkJSNavigationEvent.verification,
+        signUpUrl: ClerkJSNavigationEvent.signUp,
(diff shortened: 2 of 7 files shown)


packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
```diff
@@ -91,7 +91,9 @@ export const useSignUpContext = (): SignUpContextType => {
     displayConfig: displayConfig,
   });
 
-  let signUpUrl = pickRedirectionProp('signUpUrl', { ctx, options, displayConfig }, false);
+  let signUpUrl =
+    (ctx.routing === 'path' ? ctx.path : undefined) ||
+    pickRedirectionProp('signUpUrl', { options, displayConfig }, false);
```
anagstef, line 96: Removed the `ctx` from passing it inside the `pickRedirectionProp` here, as it never contains a `signUpUrl` prop from the `SignUp` component. The priority now is:

- The `ctx.path` is the `path` prop that user passes to the `SignUp` component, when using `routing=path`.
- The `options` passed in `pickRedirectionProp` is the ClerkProvider `signUpUrl` prop or the `NEXT_PUBLIC_CLERK_SIGN_UP_URL` env var, with that order.
- Lastly, the `displayConfig` that has the Account Portal value.

cc @nikosdouvlis @dimkl

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
```diff
@@ -175,7 +177,9 @@ export const useSignInContext = (): SignInContextType => {
     signUpUrl += `#/?${authQs}`;
   }
 
-  let signInUrl = pickRedirectionProp('signInUrl', { ctx, options, displayConfig }, false);
+  let signInUrl =
+    (ctx.routing === 'path' ? ctx.path : undefined) ||
+    pickRedirectionProp('signInUrl', { options, displayConfig }, false);
```
anagstef, line 182: Same as above!

packages/elements/src/internals/machines/third-party/actors.ts
```diff
@@ -88,6 +88,8 @@ export const handleRedirectCallback = fromCallback<AnyEventObject, HandleRedirec
         secondFactorUrl: ClerkJSNavigationEvent.signIn,
         verifyEmailAddressUrl: ClerkJSNavigationEvent.verification,
         verifyPhoneNumberUrl: ClerkJSNavigationEvent.verification,
+        signUpUrl: ClerkJSNavigationEvent.signUp,
+        signInUrl: ClerkJSNavigationEvent.signIn,
```
anagstef, line 92: @tmilewski @LekoArts  Added this one here, because the build was breaking. I don't know if that's ok.

LekoArts: Should be fine 👍

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
desiprisg: https://github.com/clerk/javascript/pull/1680 looking tastier

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
nikosdouvlis: I'm a bit skeptical because the logic here has become somewhat complicated.
I think the reason you're not moving the path check inside `pickRedirectionProp` is that `pickRedirectionProp` is too generic  for that?

nikosdouvlis (approved):

dimkl (approved):

PR #2956, opened 2024-03-08
chore(clerk-js,types): Remove unused internal components, rename text color schemes

desiprisg:
## Description
Rename the `primary` color scheme of the Text primitive to `body` and the `neutral` to `secondary`.

Also deleted some internal components that are not used anymore. Did it in the same PR as their implementation relied on the fact that `Link` and `Text` had the same color scheme names, which is no longer the case. No need to refactor something that is not used anymore.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


panteliselef (approved):
The removal of unused components looks all right.

What was the reason for renaming the colorScheme ? Do we want to align it with internal naming patterns ?

anagstef (approved):
🧹🧹🧹

nikosdouvlis (approved):

PR #2965, opened 2024-03-11
feat(backend): Export the JSON types for clerk resources

desiprisg:
## Description
Export the JSON types for clerk resources

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


desiprisg:
Question for reviewers: Is there anything in JSON.ts that might not be beneficial to export?

panteliselef:
Should these be re-exported from the rest of "server" packages ? next, fastify, etc.

desiprisg:
> Should these be re-exported from the rest of "server" packages ? next, fastify, etc.

I think we re-export everything by default, but I'll check,

panteliselef:
@desiprisg we are not. Check [this](https://github.com/clerk/javascript/blob/909ae62ae0ef8b0882df89f1919b667c0e9241ad/packages/nextjs/src/server/index.ts#L22) we are very explicit about which types we return. So far we are just returning the classes as types. We probably need to do the same for the JSON types.

dimkl:
I would suggest we export explicitly only the types we want and avoid `*` top level exports from the `@clerk/backend` since it may pollute all of our other SDKs with internal types / utils exported by the `JSON.ts` meant ONLY to be used internally.
Let's stick to the more disciplined approach for now.

packages/backend/src/index.ts
```diff
@@ -55,6 +55,11 @@ export type { VerifyTokenOptions } from './tokens/verify';
  */
 export type { WebhookEvent, WebhookEventType } from './api/resources';
 
+/**
+ * JSON types
+ */
+export type * from './api/resources/JSON';
```
dimkl, line 61: 🔧  Let's export ONLY the types we want to and avoid `*` as it may result in unwanted exports in the future since the `api/resources/JSON` exports are considered to be consumed internally.

nikosdouvlis: @desiprisg PTAL at this one as we'd like to release today

desiprisg: Done

PR #2970, opened 2024-03-11
feat(clerk-js): Experimental method for authenticating with a passkey

panteliselef:
## Description

This PR exposes the `SignIn.__experimental_authenticateWithPasskey()` method.

### Usage

https://github.com/clerk/javascript/assets/19269911/ceaa87b6-e2f3-40af-b17e-68a2f096961e

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/resources/SignIn.ts
+++ b/packages/clerk-js/src/core/resources/SignIn.ts
@@ -1,5 +1,6 @@
 import { ClerkRuntimeError, deepSnakeToCamel, Poller } from '@clerk/shared';
 import type {
+  __experimental_PasskeyFactor,
   AttemptFirstFactorParams,
   AttemptSecondFactorParams,
   AuthenticateWithRedirectParams,
@@ -7,8 +8,6 @@ import type {
   CreateEmailLinkFlowReturn,
   EmailCodeConfig,
   EmailLinkConfig,
-  PassKeyConfig,
-  PasskeyFactor,
   PhoneCodeConfig,
   PrepareFirstFactorParams,
   PrepareSecondFactorParams,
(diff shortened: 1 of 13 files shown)


packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -74,6 +83,11 @@ export class SignIn extends BaseResource implements SignInResource {
   prepareFirstFactor = (factor: PrepareFirstFactorParams): Promise<SignInResource> => {
     let config;
     switch (factor.strategy) {
+      // @ts-ignore As this is experimental we want to support it at runtime, but not at the type level
```
LekoArts, line 86: I'd like to know what the rest of the team thinks about this because in my opinion we should ship and use these experimental types. The name is clear enough and would allow us to not loose TS support while developing this.

The methods are also public so why not also make the types public?

packages/clerk-js/src/utils/passkeys.ts
```diff
@@ -124,16 +168,37 @@ function convertJSONToPublicKeyCreateOptions(jsonPublicKey: PublicKeyCredentialC
       ...jsonPublicKey.user,
       id: userIdBuffer,
     },
-  } as PublicKeyCredentialCreationOptionsWithoutExtensions;
+  } as __experimental_PublicKeyCredentialCreationOptionsWithoutExtensions;
+}
+
+function convertJSONToPublicKeyRequestOptions(jsonPublicKey: PublicKeyCredentialRequestOptionsJSON) {
```
LekoArts, line 174: Do we have unit tests for this?

packages/clerk-js/src/core/resources/SignIn.ts
panteliselef: 1. They will become public once the feature is out of the experimental stage.
2. I wanted to "hide" them as these will affect the current public and stable APIs we have like `prepareFirstFactor`.

LekoArts (approved):

PR #2976, opened 2024-03-12
feat(clerk-js): Throw errors for invalid mounts/modal opens in development

desiprisg:
## Description
* Throw an error instead of `console.info` in cases where the mounting of a component or opening of a modal is not supposed to happen (in development). Render nothing in production without an error or warning. 
* Cleaned up some cases where we would only return early if in development. That made no sense. We should only show the error in dev and return in either case.
* Removed the `withOrganizationsEnabledGuard` HOC as it did not have any effect since the `clerk` method protections were introduced.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/clerk.ts
```diff
@@ -566,7 +590,9 @@ export class Clerk implements ClerkInterface {
   public mountOrganizationList = (node: HTMLDivElement, props?: OrganizationListProps) => {
     this.assertComponentsReady(this.#componentControls);
     if (disabledOrganizationsFeature(this, this.#environment)) {
-      console.info(warnings.cannotRenderAnyOrganizationComponent('OrganizationList'));
+      if (this.#instanceType === 'development') {
+        throw new Error(warnings.cannotRenderAnyOrganizationComponent('OrganizationList'));
```
panteliselef, line 594: 🙃 We have a "ClerkRuntimeError" class we could take advantage of.

panteliselef (approved):

packages/clerk-js/src/core/clerk.ts
desiprisg: 🔧

desiprisg: `ClerkRuntimeError` also requires a code for localization. This makes no sense in this case as this is a dev only error.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -347,8 +352,13 @@ export class Clerk implements ClerkInterface {
 
   public openSignUp = (props?: SignUpProps): void => {
     this.assertComponentsReady(this.#componentControls);
-    if (sessionExistsAndSingleSessionModeEnabled(this, this.#environment) && this.#instanceType === 'development') {
-      return console.info(warnings.cannotOpenSignUpOrSignUp);
+    if (sessionExistsAndSingleSessionModeEnabled(this, this.#environment)) {
+      if (this.#instanceType === 'development') {
+        throw new ClerkRuntimeError(warnings.cannotOpenSignInOrSignUp, {
+          code: 'cannot_open_single_session_enabled',
```
panteliselef, line 358: I feel that this prefix is better
```suggestion
          code: 'cannot_render_single_session_enabled',
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -362,8 +372,13 @@ export class Clerk implements ClerkInterface {
 
   public openUserProfile = (props?: UserProfileProps): void => {
     this.assertComponentsReady(this.#componentControls);
-    if (noUserExists(this) && this.#instanceType === 'development') {
-      return console.info(warnings.cannotOpenUserProfile);
+    if (noUserExists(this)) {
+      if (this.#instanceType === 'development') {
+        throw new ClerkRuntimeError(warnings.cannotOpenUserProfile, {
+          code: 'cannot_open_no_user',
```
panteliselef, line 378:
```suggestion
          code: 'cannot_render_user_missing',
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -378,11 +393,19 @@ export class Clerk implements ClerkInterface {
...
+        });
+      }
       return;
     }
-    if (noOrganizationExists(this) && this.#instanceType === 'development') {
-      console.info(warnings.cannotRenderComponentWhenOrgDoesNotExist);
+    if (noOrganizationExists(this)) {
+      if (this.#instanceType === 'development') {
+        throw new ClerkRuntimeError(warnings.cannotRenderComponentWhenOrgDoesNotExist, {
+          code: 'cannot_open_no_organization',
```
panteliselef, line 406:
```suggestion
          code: 'cannot_render_organization_missing',
```

panteliselef (commented):
Added some suggestions for the code names

PR #2980, opened 2024-03-12
feat(backend): Create SamlConnectionApi and SamlConnection resource

EmmanouelaPothitou:
## Description

Expose any BAPI SAML Connection endpoints as part of every Clerk BE SDK

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/backend/src/api/resources/SamlConnection.ts
```diff
@@ -0,0 +1,49 @@
...
+    readonly sp_metadata_url: string,
+    readonly active: boolean,
+    readonly provider: string,
+    readonly user_count: number,
+    readonly sync_user_attributes: boolean,
+    readonly allow_subdomains: boolean,
+    readonly allow_idp_initiated: boolean,
+    readonly created_at: number,
+    readonly updated_at: number,
+    readonly attribute_mapping: AttributeMapping,
```
nikosdouvlis, line 24: These class properties need to be camelCased (not snaked_cased). 
In general, the convention we use is to use snake_case for the DTOs (a Data Transfer Object is an object that is used to encapsulate data, and send it from one subsystem of an application to another, eg from our backend to the customer's server using `@clerk/backend) and camelCase within the Clerk package.

To be more precise with an example, even though we use `idp_sso_url` to transfer the sso URL in the network layer, our customers need to be able to access it using:

```ts
const samlConnection = await apiClient.samlConnections.getSamlConnection(...);
samlConnection.idpSsoUrl;
```

packages/backend/src/api/endpoints/SamlConnectionApi.ts
```diff
@@ -0,0 +1,90 @@
...
+
+type UpdateSamlConnectionParams = {
+  name?: string;
+  provider?: SamlIdpSlug;
+  domain?: string;
+  idp_entity_id?: string;
+  idp_sso_url?: string;
+  idp_certificate?: string;
+  idp_metadata_url?: string;
+  attribute_mapping?: {
```
nikosdouvlis, line 37: Same here, these need to be camelCased. See an example here: https://github.com/clerk/javascript/blob/966b312050b572fbbbc07a6f0581cbec21847375/packages/backend/src/api/endpoints/UserApi.ts#L76

packages/backend/src/api/endpoints/SamlConnectionApi.ts
```diff
@@ -0,0 +1,90 @@
...
+type SamlConnectionListParams = {
+  limit?: number;
+  offset?: number;
+};
+type CreateSamlConnectionParams = {
+  name: string;
+  provider: SamlIdpSlug;
+  domain: string;
+  idp_entity_id?: string;
+  idp_sso_url?: string;
```
nikosdouvlis, line 18: Ditto

packages/backend/src/api/endpoints/SamlConnectionApi.ts
```diff
@@ -0,0 +1,90 @@
+import type { SamlIdpSlug } from '@clerk/types';
+
+import { joinPaths } from '../../util/path';
+import type { SamlConnection } from '../resources';
+import { AbstractAPI } from './AbstractApi';
+
+const basePath = '/saml_connections';
+
+type SamlConnectionListParams = {
```
chanioxaris, line 9: 🔧 We also support the optional parameters `query` & `orderBy`

packages/backend/src/api/endpoints/SamlConnectionApi.ts
```diff
@@ -0,0 +1,90 @@
...
+  offset?: number;
+};
+type CreateSamlConnectionParams = {
+  name: string;
+  provider: SamlIdpSlug;
+  domain: string;
+  idpEntityId?: string;
+  idpSsoUrl?: string;
+  idpCertificate?: string;
+  idpMetadataUrl?: string;
```
chanioxaris, line 20: 🔧 We are missing the optional `idpMetadata` which is actually the metadata file content

packages/backend/src/api/endpoints/SamlConnectionApi.ts
```diff
@@ -0,0 +1,90 @@
...
+};
+
+type UpdateSamlConnectionParams = {
+  name?: string;
+  provider?: SamlIdpSlug;
+  domain?: string;
+  idpEntityId?: string;
+  idpSsoUrl?: string;
+  idpCertificate?: string;
+  idpMetadataUrl?: string;
```
chanioxaris, line 36: 🔧 Same comment about `idpMetadata` property

packages/backend/src/api/resources/AttributeMapping.ts
```diff
@@ -0,0 +1,14 @@
+import type { AttributeMappingJSON } from './JSON';
```
chanioxaris, line 1: ❓ Maybe we should move this in the saml connection resource file instead as it's part of it and not an different resource. Not sure how we handle these cases

packages/backend/src/api/resources/SamlConnection.ts
```diff
@@ -0,0 +1,49 @@
+import type { AttributeMapping } from './AttributeMapping';
+import type { SamlConnectionJSON } from './JSON';
+
+export class SamlConnection {
```
chanioxaris, line 4: 🔧 The following properties can be null `idpEntityId`, `idpSsoUrl`, `idpCertificate`, `idpMetadataUrl`

We also missing the optional `idpMetadata` property

chanioxaris (commented):
Great work! Also it seems like there are some commits included not related to our changes here

nikosdouvlis (approved):
Looks good to me!

packages/backend/src/api/resources/SamlConnection.ts
```diff
@@ -0,0 +1,63 @@
...
+    readonly spMetadataUrl: string,
+    readonly active: boolean,
+    readonly provider: string,
+    readonly userCount: number,
+    readonly syncUserAttributes: boolean,
+    readonly allowSubdomains: boolean,
+    readonly allowIdpInitiated: boolean,
+    readonly createdAt: number,
+    readonly updatedAt: number,
+    readonly attributeMapping: AttributeMapping | null,
```
chanioxaris, line 24: This property will always be present
```suggestion
    readonly attributeMapping: AttributeMapping,
```

chanioxaris (approved):
Great work 💯

PR #2981, opened 2024-03-12
test(repo): Add email code integration test into the generic test suite

octoper:
## Description

This PR adds the email code integration tests into the generic test suite 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


integration/testUtils/usersService.ts
```diff
@@ -8,8 +8,8 @@ export type FakeOrganization = ReturnType<ReturnType<Awaited<typeof createUserSe
 
 export const createUserService = (clerkClient: ReturnType<typeof Clerk>) => {
   const self = {
-    createFakeUser: () => {
-      const email = `clerkcookie+${hash()}@mailsac.com`;
+    createFakeUser: (testMode = false) => {
```
anagstef, line 11: @octoper I'd suggest we use an object here, so that the usage will be:
```
createFakeUser({ testMode: true });
```
instead of:
```
createFakeUser(true);
```
which is weird to read, because you don't really know what `true` means.

nikosdouvlis: I agree with @anagstef , let's use
```
type FakeUserOptions = {
/**
* A fictional email is an email that contains +clerk_test and can always be verified using 424242 as the OTP. No email will be sent.
* https://clerk.com/docs/testing/test-emails-and-phones#email-addresses
**/
fictionalEmail: boolean
/**
* A fictional phone number follows the "+1 (XXX) 555-0100 to +1 (XXX) 555-0199" format and can always be verified using `424242` as the OTP. No email will be sent.
* https://clerk.com/docs/testing/test-emails-and-phones#phone-numbers
**/
fictionPhoneNumber: boolean
}
```

nikosdouvlis (approved):

PR #2983, opened 2024-03-12
fix(*): Update dependencies to resolve security alerts

dimkl:
## Description

Investigate and fix dependabot security alerts and results from `npm audit`.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/integration/templates/next-app-router/package.json
+++ b/integration/templates/next-app-router/package.json
@@ -12,7 +12,7 @@
     "@types/node": "^18.17.0",
     "@types/react": "18.2.48",
     "@types/react-dom": "18.2.18",
-    "next": "13",
+    "next": "^13.5",
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "typescript": "5.1.6"
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -113,7 +113,7 @@
     "esbuild": "^0.15.12",
     "esbuild-register": "^3.3.3",
     "eslint-config-custom": "*",
-    "miniflare": "^2.12.1",
+    "miniflare": "^2.14.2",
     "npm-run-all": "^4.1.5",
     "qunit": "^2.19.3",
     "sinon": "^14.0.1",
--- a/packages/elements/package.json
+++ b/packages/elements/package.json
@@ -84,7 +84,7 @@
     "eslint-config-custom": "*",
     "next": "^14.0.4",
     "tslib": "2.4.1",
-    "tsup": "^8.0.1",
+    "tsup": "*",
     "type-fest": "^4.9.0",
     "typescript": "^5.3.3"
   },
--- a/packages/expo/package.json
+++ b/packages/expo/package.json
@@ -43,7 +43,8 @@
     "@clerk/clerk-react": "5.0.0-beta.30",
     "@clerk/shared": "2.0.0-beta.19",
     "base-64": "^1.0.0",
-    "react-native-url-polyfill": "2.0.0"
+    "react-native-url-polyfill": "2.0.0",
+    "tslib": "2.4.1"
   },
   "devDependencies": {
     "@clerk/types": "^4.0.0-beta.20",
@@ -52,16 +53,17 @@
     "@types/react": "*",
     "@types/react-dom": "*",
     "eslint-config-custom": "*",
+    "expo-application": "^4.2.2",
     "expo-auth-session": "^5.4.0",
+    "expo-constants": "^14.5.1",
     "expo-web-browser": "^12.8.2",
-    "tslib": "^2.6.2",
     "typescript": "*"
   },
   "peerDependencies": {
     "expo-application": ">=5",
     "expo-auth-session": ">=4",
     "expo-constants": ">=14",
-    "expo-web-browser": ">=12",
+    "expo-web-browser": ">=12.5.0",
     "react": ">=18",
     "react-dom": ">=18"
   },
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -70,7 +70,8 @@
     "@clerk/backend": "1.0.0-beta.28",
     "@clerk/clerk-react": "5.0.0-beta.30",
     "@clerk/shared": "2.0.0-beta.19",
-    "path-to-regexp": "6.2.1"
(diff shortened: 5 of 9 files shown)


packages/expo/package.json
```diff
@@ -52,16 +53,13 @@
     "@types/react": "*",
     "@types/react-dom": "*",
     "eslint-config-custom": "*",
-    "expo-auth-session": "^5.4.0",
-    "expo-web-browser": "^12.8.2",
```
LekoArts, line 56: You'll still need them in `devDeps` to develop the package locally, otherwise you'll have TS errors in https://github.com/clerk/javascript/blob/71663c568926b1d60b97aa7ccc5298d05b618af2/packages/expo/src/useOAuth.ts

packages/nextjs/package.json
```diff
@@ -78,12 +78,11 @@
     "@types/react": "*",
     "@types/react-dom": "*",
     "eslint-config-custom": "*",
-    "next": "^13.4.19",
```
LekoArts, line 81: Same as with expo, we'll need to keep this as a devDep

nikosdouvlis: I think we can remove this from here if we have it as a `devDependency` in the root `package.json` but what's the benefit in doing that?

packages/backend/package.json
```diff
@@ -113,7 +113,7 @@
     "esbuild": "^0.15.12",
     "esbuild-register": "^3.3.3",
     "eslint-config-custom": "*",
-    "miniflare": "^2.12.1",
+    "miniflare": "^3.20240304.0",
```
anagstef, line 116: Is there a possibility that this one will introduce breaking changes?

packages/nextjs/package.json
dimkl: i will keep it as dev dependency here.

packages/backend/package.json
dimkl: This is a dev dependency only used in our backend tests. I would not consider it a breaking change. Do you have something in mind that could be broken by this?

packages/expo/package.json
dimkl: We will keep them as devDeps to support cases of package managers not installing peer deps by default.

dimkl:
I had some issues with "List node_modules" of our GH action when updating the `miniflare` and `tslib` libraries. I resolved it by installing those packages specifically in each package directory.
Screenshot 2024-03-13 at 15 28 40

Example run: https://github.com/clerk/javascript/actions/runs/8264832463/job/22609252010?pr=2983

dimkl:
I changed the `miniflare` to `2.4.12` instead of `3.20240304.0` since the tests broke due to the following error:
```
Disallowed operation called within global scope. Asynchronous I/O (ex: fetch() or connect())....
```

integration/templates/next-app-router/package.json
```diff
@@ -12,7 +12,7 @@
     "@types/node": "^18.17.0",
     "@types/react": "18.2.48",
     "@types/react-dom": "18.2.18",
-    "next": "13",
+    "next": "13.5.4",
```
anagstef, line 15: I think we should keep the `13` version here, as this is about integration testing and we want the latest 13 version of next to be installed.

dimkl: You are correct, it shouldn't be explicit. Is it okay if i put something like this `^13.5`?
This change is applied to resolve this: https://github.com/clerk/javascript/security/dependabot/234

dimkl: Nice catch!

packages/expo/package.json
```diff
@@ -43,7 +43,8 @@
     "@clerk/clerk-react": "5.0.0-beta.30",
     "@clerk/shared": "2.0.0-beta.19",
     "base-64": "^1.0.0",
-    "react-native-url-polyfill": "2.0.0"
+    "react-native-url-polyfill": "2.0.0",
+    "tslib": "2.4.1"
```
anagstef, line 47: Why we downgraded to 2.4.1 here?

packages/expo/package.json
```diff
@@ -52,16 +53,16 @@
     "@types/react": "*",
     "@types/react-dom": "*",
     "eslint-config-custom": "*",
-    "expo-auth-session": "^5.4.0",
```
anagstef, line 55: ❓  Is it safe to remove this package from the devDependencies? 🤔

dimkl: you are correct. It should not be removed. i missed it during the last rebase. I am trying to resolve another issue. so please wait until i ping you to avoid wasting time reviewing this. :(

packages/expo/package.json
dimkl: To consolidate all the different `tslib` versions. I will drop this change and handle it in a separate PR

tmilewski (approved):

packages/elements/package.json
```diff
@@ -106,6 +106,9 @@
       "optional": true
     }
   },
+  "optionalDependencies": {
+    "@rollup/rollup-linux-x64-gnu": "^4.9.6"
```
LekoArts, line 110: This is a sign that something else is wrong, this shouldn't be here and not necessary. Every other package that uses `tsup` would run into it otherwise. The commited lock file in this PR must be wrong

LekoArts (changes requested):

LekoArts (approved):

packages/nextjs/package.json
```diff
@@ -70,20 +70,20 @@
...
-    "tslib": "2.4.1",
+    "next": "^13.5.4",
     "typescript": "*"
   },
   "peerDependencies": {
-    "next": "^13.0.4 || ^14.0.3",
+    "next": "^13.5.4 || ^14.0.3",
```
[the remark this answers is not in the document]
dimkl: @nikosdouvlis yes, kind of.
if this is for the changeset entry, i guess it's better to change it to `major`. Since we are in a pre-release tag, it's okay.
I think that customers who want to Clerk with NextJS in production would probably have a newer @13 version with all the security updates.
Wdyt?

PR #2993, opened 2024-03-13
fix(clerk-js): Remove ticket related query params at the end of the flow

desiprisg:
## Description
Moving this further down to keep the ticket active flow upon page refresh.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


dimkl (commented):
Could we add tests to verify the removal of the query params?

.changeset/tasty-seals-fail.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
dimkl, line 2: 🙃 I think we should add changelog since this PR is a fix for ticket flows for a bug introduced in the previous version.

desiprisg: Not aware of the bug you're talking about. Do we have any resource on that. I'd like to make sure this fixes it as well? I'm just doing this because the state will be lost on refresh and that seems wrong.

dimkl: initiate a ticket flow and before completing it (eg before completing the sign-up from an invitation), refresh the page. Since the ticket query param is removed, the flow will never be completed.

desiprisg: Seems to be no issue with multi-step flows as the ticket is passed through the `create()` method only. It's just the refresh from what I can tell which should be fixed by this PR.

packages/clerk-js/src/ui/components/SignUp/__tests__/SignUpStart.test.tsx
```diff
@@ -230,12 +231,6 @@ describe('SignUpStart', () => {
         ),
       );
       expect(fixtures.signUp.create).toHaveBeenCalledWith({ strategy: 'ticket', ticket: 'test_ticket' });
-
-      expect(window.history.replaceState).toHaveBeenCalledWith(
-        undefined,
-        '',
-        expect.not.stringContaining('__clerk_ticket'),
-      );
```
nikosdouvlis, line 238: Why did we completely remove this test? 
Let's re-introduce it at the appropriate spot

.changeset/tasty-seals-fail.md
dimkl: probably yes. i haven't checked all the flows. If that is the only flow affected, let's proceed with fixing only that.

packages/clerk-js/src/ui/components/SignUp/__tests__/SignUpStart.test.tsx
desiprisg: This whole test was added in the previous PR. The query param part of the test will have to be moved to a separate test.

desiprisg: Added!

packages/clerk-js/src/ui/components/SignUp/__tests__/SignUpStart.test.tsx
```diff
@@ -231,11 +232,49 @@ describe('SignUpStart', () => {
       );
       expect(fixtures.signUp.create).toHaveBeenCalledWith({ strategy: 'ticket', ticket: 'test_ticket' });
 
+      //don't remove the query param quite yet
```
nikosdouvlis, line 235: 💯

nikosdouvlis (approved):

PR #3000, opened 2024-03-15
feat(clerk-js): Use passkey as first factor in `<SignIn/>`

panteliselef:
### Things to consider
- Unit tests will be part of follow up PR
- This is behind a FF controlled by FAPI.

### Autofill Passkey
https://github.com/clerk/javascript/assets/19269911/c9a49406-73bf-4e75-b652-b484d8522b72

### Discoverable Passkeys
https://github.com/clerk/javascript/assets/19269911/8d05ca93-2eae-49b9-9d7b-1b1fd2c1a7c0

### Sign in with identifier + passkeys as first factor
https://github.com/clerk/javascript/assets/19269911/a36495b6-5e18-4bc6-a2ee-f285bb4b45b5

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/utils/passkeys.ts
```diff
@@ -114,6 +142,7 @@ async function webAuthnGetCredential({
     const credential = (await navigator.credentials.get({
       publicKey: publicKeyOptions,
       mediation: conditionalUI ? 'conditional' : 'optional',
+      signal: __internal_WebAuthnAbortService.createAbortSignal(),
```
panteliselef, line 145: Only one call to retrieve credentials can exist at time.

That means in the case of `autofill` user may never interact with it, this makes sure that every new call to retrieve the credentials will cancel the previews one.

packages/clerk-js/src/ui/components/SignIn/SignInFactorOnePasskey.tsx
```diff
@@ -0,0 +1,92 @@
...
+            <Icon
+              elementDescriptor={descriptors.passkeyIcon}
+              icon={Fingerprint}
+              sx={t => ({
+                color: t.colors.$neutralAlpha500,
+                marginInline: 'auto',
+                paddingBottom: t.sizes.$1,
+                width: t.sizes.$12,
+                height: t.sizes.$12,
+              })}
```
panteliselef, line 53: @desiprisg @anagstef  Please let me know if we have new patterns that this code should align with.

In the designs this icon have a size of `48x48` which is not covered by the current Icon implementation and just adding an `xl` variant does not feel right as the `lg` represents a `20x20` so the difference would be quite large between them.

We may add a new `Header.Icon` component, but as long as this is used in only one place do we need them to ?

packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -330,7 +344,15 @@ export class SignIn extends BaseResource implements SignInResource {
...
-      this.supportedFirstFactors = deepSnakeToCamel(data.supported_first_factors) as SignInFirstFactor[];
+      // TODO: Remove this mock
+      if (data.supported_first_factors) {
+        this.supportedFirstFactors = deepSnakeToCamel([
+          ...data.supported_first_factors,
+          {
+            strategy: 'passkey',
+          },
+        ]) as SignInFirstFactor[];
+      }
```
panteliselef, line 355: Pending for FAPI to catch up

packages/clerk-js/src/ui/components/SignIn/SignInFactorOnePasskey.tsx
panteliselef: Maybe the `Header.Icon` will make sense in order to have a proper descriptor instead of the `passkeyIcon` descriptor. WDYT ?

LekoArts (approved):
Generally LGTM. Left a non-blocking comment.

I'd also make sure that [your question](https://github.com/clerk/javascript/pull/3000/files#r1526146590) is answered 👍

packages/localizations/src/en-US.ts
```diff
@@ -346,6 +347,10 @@ export const enUS: LocalizationResource = {
       subtitle: 'Enter the password associated with your account',
       title: 'Enter your password',
     },
+    passkey: {
+      title: 'Use your passkey',
+      subtitle: "Using your passkey confirms it's you. Your device may ask for your fingerprint, face or screen lock.",
```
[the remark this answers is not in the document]
panteliselef: Hey @LekoArts this text is copied from the designs. I'm sure we can iterate more on this as more people use it during the beta. Yes, by "screen lock" we mean a PIN.

packages/clerk-js/src/ui/components/SignIn/SignInFactorOnePasskey.tsx
anagstef: It's ok to keep the width and height values.

anagstef: I'd say we don't add a `Header.Icon` yet. Let's keep the Icon as is for now. :)

PR #3001, opened 2024-03-15
fix(nextjs,shared): Allow manually passed publishable/secret key to middleware

desiprisg:
## Description

It seems that we weren't taking the passed in `secretKey` and `publishableKey` into account in every scenario.
I've moved the checks to the top and fixed the passed in values where necessary.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/nextjs/src/server/index.ts
```diff
@@ -1,3 +1,5 @@
+import { buildErrorThrower } from '@clerk/shared';
```
LekoArts, line 1:
```suggestion
import { buildErrorThrower } from '@clerk/shared/error';
```

Small optimization for tree-shaking

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -1,17 +1,17 @@
...
-import { AuthStatus, constants, createClerkRequest } from '@clerk/backend/internal';
+import { AuthStatus, constants, createClerkRequest, createRedirect } from '@clerk/backend/internal';
 import { isDevelopmentFromSecretKey } from '@clerk/shared/keys';
 import { eventMethodCalled } from '@clerk/shared/telemetry';
 import type { NextFetchEvent, NextMiddleware, NextRequest } from 'next/server';
 import { NextResponse } from 'next/server';
 
 import { isRedirect, mergeResponses, serverRedirectWithAuth, setHeader, stringifyHeaders } from '../utils';
 import { withLogger } from '../utils/debugLogger';
+import { errorThrower } from '.';
```
dimkl, line 10: ⛔ Please replace it with explicit import since this may cause circular dependency issues.

playground/nextjs/package.json
```diff
@@ -11,8 +11,13 @@
...
   "dependencies": {
-    "@clerk/nextjs": "beta",
-    "@clerk/themes": "beta",
+    "@clerk/backend": "file:.yalc/@clerk/backend",
+    "@clerk/clerk-react": "file:.yalc/@clerk/clerk-react",
+    "@clerk/clerk-sdk-node": "file:.yalc/@clerk/clerk-sdk-node",
+    "@clerk/nextjs": "file:.yalc/@clerk/nextjs",
+    "@clerk/shared": "file:.yalc/@clerk/shared",
+    "@clerk/themes": "file:.yalc/@clerk/themes",
+    "@clerk/types": "file:.yalc/@clerk/types",
```
dimkl, line 20: 🔧 i think we should revert this since it's being used in the preview deployment.
cc: @LekoArts

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -115,12 +120,26 @@ export interface AuthMiddleware {
...
   const [params = {}] = args as [AuthMiddlewareParams?];
-  const { beforeAuth, afterAuth, publicRoutes, ignoredRoutes, apiRoutes, ...options } = params;
+  const publishableKey = params.publishableKey || PUBLISHABLE_KEY;
+  if (!publishableKey) {
+    errorThrower.throwMissingPublishableKeyError();
+  }
+  const secretKey = params.secretKey || SECRET_KEY;
+  if (!secretKey) {
+    errorThrower.throwMissingSecretKeyError();
+  }
```
dimkl, line 130: 🔧  i think we shouldn't add the above 2 validations since they should be triggered by the `clerkClient.authenticateRequest()` of `@clerk/backend`. Having those checks makes them run for all our SDKs depending on the `@clerk/backend` and it makes it easier to maintain them being in a single place.

desiprisg: I think these belong here as well. It's the library's responsibility to ensure that it's parameters are available as arguments or environment variables.

playground/nextjs/package.json
desiprisg: @LekoArts Do we need this? This gets overwritten everytime we execute the yalc command.

packages/nextjs/src/server/authMiddleware.ts
dimkl: From the offline discussion, we have concluded that those checks will be kept since the values of pk/sk are used outside of the `authenticateRequest()` context and it's confusing to depend that the authenticateRequest() should be called before using those values to validate the values.

packages/backend/src/util/assertValidSecretKey.ts
```diff
@@ -1,8 +1,6 @@
 export function assertValidSecretKey(val: unknown): asserts val is string {
   if (!val || typeof val !== 'string') {
-    throw Error(
-      'Missing Clerk Secret Key or API Key. Go to https://dashboard.clerk.com and get your key for your instance.',
-    );
+    throw Error('Missing Clerk Secret Key. Go to https://dashboard.clerk.com and get your key for your instance.');
```
dimkl, line 3: nice!

packages/nextjs/src/server/authMiddleware.test.ts
```diff
@@ -449,19 +436,12 @@ describe('Dev Browser JWT when redirecting to cross origin', function () {
...
     const resp = await authMiddleware({
-      beforeAuth: () => NextResponse.next(),
+      beforeAuth: () => NextResponse.redirect('https://google.com/'),
     })(mockRequest({ url: '/protected', appendDevBrowserCookie: true }), {} as NextFetchEvent);
 
     expect(resp?.status).toEqual(307);
-    expect(resp?.headers.get('location')).toEqual(
-      'https://accounts.included.katydid-92.lcl.dev/sign-in?redirect_url=https%3A%2F%2Fwww.clerk.com%2Fprotected',
-    );
+    expect(resp?.headers.get('location')).toEqual('https://google.com/');
```
dimkl, line 444: 🙃 Could you also assert that the `x-clerk-redirect-to` is not set? This is not a change related to this PR but it will improve the test to match the description.

packages/nextjs/src/server/authMiddleware.test.ts
```diff
@@ -19,20 +28,7 @@ jest.mock('./clerkClient', () => {
...
-    'https://accounts.included.katydid-92.lcl.dev/sign-in?redirect_url=https%3A%2F%2Fwww.clerk.com%2Fprotected',
-  );
-  return setHeader(res, 'x-clerk-redirect-to', 'true');
-});
-
-jest.mock('./redirectHelpers', () => {
-  return {
-    redirectToSignIn: mockRedirectToSignIn,
-  };
-});
```
dimkl, line 33: 📝 nice!. This is not needed since we won't use it in `authMiddleware()`

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -220,20 +237,33 @@ const authMiddleware: AuthMiddleware = (...args: unknown[]) => {
 
 export { authMiddleware };
 
+const redirectAdapter = (url: string) => {
+  const res = NextResponse.redirect(url);
+  return setHeader(res, constants.Headers.ClerkRedirectTo, 'true');
+};
```
dimkl, line 243: Let's move somewhere in utils and use the redirectAdapter from here: https://github.com/clerk/javascript/blob/84c05e6237bf34d5f14e25ac8bc7059a8e684c1e/packages/nextjs/src/server/clerkMiddleware.ts#L159

playground/nextjs/middleware.ts
dimkl, line 1: 🙃 Let's revert this change since this is a playground

dimkl (approved):

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -223,17 +245,25 @@ export { authMiddleware };
...
       if (isApiRoute(req)) {
-        informAboutProtectedRoute(req.experimental_clerkUrl.pathname, params, true);
+        informAboutProtectedRoute(req.experimental_clerkUrl.pathname, enforcedParams, true);
         return apiEndpointUnauthorizedNextResponse();
       } else {
-        informAboutProtectedRoute(req.experimental_clerkUrl.pathname, params, false);
+        informAboutProtectedRoute(req.experimental_clerkUrl.pathname, enforcedParams, false);
       }
-      return redirectToSignIn({ returnBackUrl: req.experimental_clerkUrl.href });
+      return createRedirect({
```
nikosdouvlis, line 258: Good job using `createRedirect` for this :)

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -37,6 +38,10 @@ type ClerkMiddlewareHandler = (
 ) => NextMiddlewareReturn;
 
 export type ClerkMiddlewareOptions = AuthenticateRequestOptions & { debug?: boolean };
+export type EnforcedClerkMiddlewareOptions = {
```
nikosdouvlis, line 41: Same comment about `EnforcedClerkMiddlewareOptions` - we should't rely on types like these as we've found that they are not maintainable in the long run. Can we rely on TS inference more within the functions?

Also let's be cautious about the `export` keyword as this doesn't look like a type we want to make public.

packages/nextjs/src/server/utils.ts
```diff
@@ -194,3 +194,7 @@ export const handleMultiDomainAndProxy = (clerkRequest: ClerkRequest, opts: Auth
     signInUrl,
   };
 };
+
+export const redirectAdapter = (url: string | URL) => {
```
nikosdouvlis, line 198: ✔️

nikosdouvlis (commented):
Tests look good, good job!

.changeset/stupid-coats-smell.md
```diff
@@ -0,0 +1,6 @@
+---
+'@clerk/nextjs': minor
+'@clerk/shared': minor
+---
+
+Fix: allow manually passed publishable/secret key to middleware.
```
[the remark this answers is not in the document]
desiprisg: this is a bug fix, shouldn't it be stated?

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -104,6 +109,11 @@ type AuthMiddlewareParams = AuthenticateRequestOptions & {
   debug?: boolean;
 };
 
+type EnforcedAuthMiddlewareParams = {
+  publishableKey: string;
+  secretKey: string;
+} & AuthMiddlewareParams;
+
```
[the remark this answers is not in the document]
desiprisg: This is just an added internal type to avoid passing unenforced options to other functions that use it.

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -115,12 +125,24 @@ export interface AuthMiddleware {
...
+    errorThrower.throwMissingPublishableKeyError();
+  }
+  const secretKey = params.secretKey || SECRET_KEY;
+  if (!secretKey) {
+    errorThrower.throwMissingSecretKeyError();
+  }
+  const signInUrl = params.signInUrl || SIGN_IN_URL;
+  const signUpUrl = params.signUpUrl || SIGN_UP_URL;
+
+  const enforcedParams: EnforcedAuthMiddlewareParams = { ...params, publishableKey, secretKey, signInUrl, signUpUrl };
```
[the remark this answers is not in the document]
desiprisg: I've made some changes. PTAL

packages/nextjs/src/server/clerkMiddleware.ts
desiprisg: This has been removed.

packages/nextjs/src/server/authMiddleware.ts
```diff
@@ -223,17 +236,25 @@ export { authMiddleware };
 const createDefaultAfterAuth = (
   isPublicRoute: ReturnType<typeof createRouteMatcher>,
   isApiRoute: ReturnType<typeof createApiRoutes>,
-  params: AuthMiddlewareParams,
+  options: { signInUrl: string; signUpUrl: string; publishableKey: string; secretKey: string },
```
nikosdouvlis, line 239: ❓ why are we defining the options here explicitly? I think you're trying to minimize what is visible to this function, which is great, but in this case I think it's better to pass the options through so we don't have to worry about maintaining an extra inline type :)

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -62,7 +69,21 @@ interface ClerkMiddleware {
 
 export const clerkMiddleware: ClerkMiddleware = (...args: unknown[]): any => {
   const [request, event] = parseRequestAndEvent(args);
-  const [handler, options] = parseHandlerAndOptions(args);
+  const [handler, params] = parseHandlerAndOptions(args);
+  const publishableKey = assertKey(params.publishableKey || PUBLISHABLE_KEY, () =>
+    errorThrower.throwMissingPublishableKeyError(),
+  );
+  const secretKey = assertKey(params.secretKey || SECRET_KEY, () => errorThrower.throwMissingSecretKeyError());
```
nikosdouvlis, line 76: 🙃 
```suggestion
  const secretKey = assertKey(params.secretKey || SECRET_KEY, errorThrower.throwMissingSecretKeyError);
```

(same for the rest)

packages/nextjs/src/server/utils.ts
```diff
@@ -194,3 +194,15 @@ export const handleMultiDomainAndProxy = (clerkRequest: ClerkRequest, opts: Auth
...
+  return NextResponse.redirect(url, { headers: { [constants.Headers.ClerkRedirectTo]: 'true' } });
+};
+
+export function assertKey(key: string, onError: () => never): string {
```
nikosdouvlis, line 202: 🙃 / nit: 
The "assert" prefix here made me think this was an assertion function (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) 

Can we make it work with the errorThrower and an `: asserts key is string` return type? 
If not, I'd probably inline the check as it's fairly simple: you abstract an `if()` check with the cost of some extra indirection. This is a 🙃 so it's up to you if it's worth it or not!

nikosdouvlis (approved):

packages/nextjs/src/server/utils.ts
desiprisg: No, because you are already passing in a `string` because we are declaring our constants like `process.env.FOO || ''`.

packages/nextjs/src/server/clerkMiddleware.ts
desiprisg: This results in an eslint error.

packages/nextjs/src/server/authMiddleware.ts
desiprisg: If we were to pass the params type, we would have to assert the necessary keys again to be safe, and that is just not the purpose of this function.

PR #3004, opened 2024-03-15
feat(elements): Add SafeIdentifier & Salutation [SDK-1501]

tmilewski:
## Description

Adds a basic Sign In `SafeIdentifier` and `Salutation` component.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/internals/machines/sign-in/types/router.types.ts
```diff
@@ -93,3 +93,18 @@ export interface SignInRouterSchema {
...
+export type SignInStateValue = any; // TODO: Update
+
+export type SignInRouterSnapshot = MachineSnapshot<
+  SignInRouterContext,
+  SignInRouterEvents,
+  SignInChildren,
+  SignInStateValue,
+  SignInRouterTags,
+  SignInOuptut
+>;
```
BRKalow, line 110: What's this change for? 👀

packages/elements/src/react/sign-in/index.ts
```diff
@@ -14,6 +14,12 @@ export {
   SignInVerification as Verification,
 } from './verifications';
 
+export {
+  SignInName as Name,
```
BRKalow, line 18: Do we need `Name`? Is it used somewhere in our existing flow?

packages/elements/src/react/sign-in/identifiers.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return SignInRouterCtx.useSelector(SignInSafeIdentifierSelector);
+}
+
+export function SignInSalutation(): string {
+  return SignInRouterCtx.useSelector(SignInSalutationSelector);
+}
+
+export function SignInName(): string {
+  return SignInRouterCtx.useSelector(SignInNameSelector);
+}
```
BRKalow, line 19: I'm wondering if we wrap these in a `<span>` to ensure we're consistently rendering markup from our components. The alternative if someone wants to style these now is to wrap them in an element, which I would say isn't consistent with the rest of the component surface at the moment (I think we're always rendering markup for components that render something).

packages/elements/src/internals/machines/utils/formatters.ts
```diff
@@ -0,0 +1,24 @@
...
+  firstName,
+  lastName,
+  identifier,
+}: {
+  firstName: string | undefined;
+  lastName: string | undefined;
+  identifier: string | undefined;
+}): string {
+  return (firstName && formatName(firstName)) || (lastName && formatName(lastName)) || identifier || '';
+}
```
BRKalow, line 24: 💭 this is potentially something that we'd want to get from FAPI instead.

packages/elements/src/internals/machines/sign-in/types/router.types.ts
tmilewski: A central place to retain the type for a SignInRouter snapshot... as used largely by `useSelector`.

packages/elements/src/react/sign-in/index.ts
tmilewski: Felt like a nicety given it's use in `Salutation`, such that it doesn't fall back to a phone number or email.

BRKalow: I'd say let's remove for now unless we get feedback asking for it specifically. My main signal is that we're not using it in the fixture app.

packages/elements/src/react/sign-in/identifiers.tsx
tmilewski: Not always (ex: `SignIn`). That said, I'm torn. 

On one hand, I agree. 
On another, is that really a problem? 

Meaning do we want to surface more code and force them to have additionaly elements around something they'll likely wrap in an alternate way (similarly to how I've used them in the example sign-in page)?

packages/elements/src/internals/machines/utils/formatters.ts
tmilewski: Definitely, though in the mean time... 🤷

BRKalow: Maybe we can leave a comment as a breadcrumb? 😄 e.g.

```
// TODO: ideally the derivation of these values lives in FAPI and comes back directly from the API
```

tmilewski:
@BRKalow Updated

packages/elements/src/react/sign-in/identifiers.tsx
BRKalow: It's not a problem really, just more of a consistency thing and being predictable in how our components behave. We can always adjust down the road though, so happy to ship with just text here.

packages/elements/examples/nextjs/app/sign-in/[[...sign-in]]/page.tsx
```diff
@@ -188,7 +210,9 @@ export default function SignInPage() {
             <Verification name='reset_password_email_code'>
               <H3>Verify your email</H3>
 
-              <P>Please check your email for a verification code...</P>
+              <P className='text-sm'>
+                Please check your <SafeIdentifier /> for a verification code...
```
Comment (BRKalow) [inline packages/elements/examples/nextjs/app/sign-in/[[...sign-in]]/page.tsx:214]: nit: Should this be something like:
```suggestion
                We've sent a verification code to <SafeIdentifier />.
```

BRKalow (approved):

PR #3017, opened 2024-03-20
feat(elements): Adjust FieldState and Field

LekoArts:
## Description

- Adjust `<FieldState>` to return its state directly in the function param, not nested inside an object
- Adjust `<Field>` to allow for a function prop that gives back the field state

```tsx
<Field name="email">
  <FieldState>
    {(state) => (
      <pre>Field state: {state}</pre>
    )}
  </FieldState>
</Field>
```

```tsx
<Field name="emailAddress">
  {(fieldState) => (
    <span>{fieldState === "invalid" ? "Error" : "All fine!"}</span>
  )}
</Field>
```

Fixes SDK-1524
Fixes SDK-1525

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:


packages/elements/examples/nextjs/app/sign-in/[[...sign-in]]/page.tsx
```diff
@@ -126,12 +126,18 @@ export default function SignInPage() {
...
-                    className='bg-[rgb(12,12,12)] border-[rgb(37,37,37)] border rounded data-[invalid=true]:border-red-400 w-full placeholder-[rgb(100,100,100)] px-4 py-2'
-                    placeholder='Enter your email address'
-                  />
-                  <FieldError className='block text-red-400 font-mono w-full' />
+                  {fieldState => (
+                    <>
+                      <Label className='sr-only'>Email</Label>
+                      <Input
+                        className={`bg-[rgb(12,12,12)] border-[rgb(37,37,37)] border rounded w-full placeholder-[rgb(100,100,100)] px-4 py-2 ${
+                          fieldState === 'invalid' && 'border-red-500'
```
Comment (BRKalow) [inline packages/elements/examples/nextjs/app/sign-in/[[...sign-in]]/page.tsx:134]: [nit] Just an observation, I would actually consider this _not_ the proper usage of accessing field state. Given that we have the data attribute for styling purposes, that's probably how we would want to educate folks for basic styling.

For more advanced use cases, like animation, then we have the render prop for programmatic access.

BRKalow (approved):
lgtm!

packages/elements/examples/nextjs/app/sign-in/[[...sign-in]]/page.tsx
Comment (LekoArts) [reply packages/elements/examples/nextjs/app/sign-in/[[...sign-in]]/page.tsx:134]: Yeah, agree. One should use the data attribute - but since this is our Playground and I needed a way to test it + have it used somewhere to see possible future regressions I opted to do that 👍

PR #3026, opened 2024-03-21
feat(elements): Move <Loading> component to /common subpath

LekoArts:
## Description

```diff
- import { Loading } from "@clerk/elements/sign-in"
- import { Loading } from "@clerk/elements/sign-up"
+ import { Loading } from "@clerk/elements/common"
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/react/common/loading.tsx
```diff
@@ -0,0 +1,249 @@
...
+        >
+          {children}
+        </SignUpLoading>
+      );
+    } else {
+      throw new ClerkElementsRuntimeError(`Invalid scope "${scope}" used for <Loading> inside <SignUp>.`);
+    }
+  }
+
+  throw new ClerkElementsRuntimeError('Invalid state for <Loading>. Please open an issue.');
```
LekoArts, line 121: TS didn't infer further up that `signUpRouterRef` must exist so this is just to be safe

packages/elements/src/react/common/loading.tsx
```diff
@@ -0,0 +1,249 @@
...
+}
+
+type SignInLoadingProps = {
+  scope?: LoadingScope<SignInStep>;
+  children: (isLoading: boolean) => React.ReactNode;
+  routerRef: ActorSignIn;
+};
+
+function SignInLoading({ children, scope, routerRef }: SignInLoadingProps) {
```
LekoArts, line 130: There is _some_ shared logic between SignInLoading and SignUpLoading but I felt that trying to DRY things up would complicate things. I think as-is it's more readable albeit we have some duplicate lines.

BRKalow (approved):
In the future I wonder if we could invert control here and have each flow provide loading data in a unique loading context, so we don't have to do the flow checks in the loading component itself. 🤔 

We can iterate on it though, I like that this is coming from common now! Nice work

PR #3030, opened 2024-03-22
fix(elements): Improve OTP selection interaction

BRKalow:
## Description

Was doing some testing with the OTP input and noticed some odd behavior around selection. I refactored this to align with some of the earlier logic, which seemed to be functioning better.

* Tested in iOS, Chrome, Firefox, and Safari
* Validated all selection interactions work as expected
* Validated pasting works across all browsers

Some notes:
* We can likely re-add some of the removed styles, I just reset from where I knew selection was close to working
* I removed the animated cursor to validate multi-select
* Why did we change from `none` to `transparent` in most places initially?
* The keyboard interaction doesn't seem necessary as we're relying on the underlying native input

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/examples/nextjs/app/otp-playground/page.tsx
```diff
@@ -41,13 +43,13 @@ export default function Page() {
...
 
-                  {(status === 'cursor' || status === 'selected') && (
+                  {/* {(status === 'cursor' || status === 'selected') && (
                     <motion.div
                       layoutId='tab-trigger-underline'
                       transition={{ duration: 0.2, type: 'spring', damping: 20, stiffness: 200 }}
                       className='absolute inset-0 border border-sky-400 shadow-[0_0_8px_2px_theme(colors.sky.400/30%)] z-10 rounded-[inherit] bg-sky-400/10'
                     />
-                  )}
+                  )} */}
```
BRKalow, line 52: Commented out so I could see the styles when doing multi-select 👀

packages/elements/src/react/common/form/otp.tsx
```diff
@@ -88,173 +87,70 @@ const OTPInputStandard = React.forwardRef<HTMLInputElement, Omit<OTPInputProps,
...
-      const element = innerRef.current;
-
-      if (!element) return;
-
-      const originalSelect = element.select.bind(element);
-      element.select = () => {
-        originalSelect();
-        setSelectionRange([0, element.value.length]);
-      };
-    }, []);
```
BRKalow, line 120: Why did we use the native `.select()` here?

packages/elements/src/react/common/form/otp.tsx
```diff
@@ -88,173 +87,70 @@ const OTPInputStandard = React.forwardRef<HTMLInputElement, Omit<OTPInputProps,
...
-          setSelectionRange([innerRef.current.selectionStart, innerRef.current.selectionEnd]);
-        }
-      }, 50);
+    // This ensures we can access innerRef internally while still exposing it via the ref prop
+    React.useImperativeHandle(ref, () => innerRef.current as HTMLInputElement, []);
 
-      return () => {
-        clearInterval(interval);
-      };
-    }, [isFocused]);
```
BRKalow, line 135: This seemed heavy-handed and unnecessary, reverted to earlier implementation.

packages/elements/src/react/common/form/otp.tsx
```diff
@@ -88,173 +87,70 @@ const OTPInputStandard = React.forwardRef<HTMLInputElement, Omit<OTPInputProps,
...
         style={wrapperStyle}
       >
         {/* We can't target pseudo-elements with the style prop, so we inject a tag here */}
-        <style>
-          {`
-        input[data-otp-input-segmented]::selection {
+        <style>{`
+      input[data-otp-input]::selection {
         color: transparent;
-          background-color: transparent;
```
BRKalow, line 170: There was probably a good reason to use `transparent` 👀. Can add back!

packages/elements/src/react/sign-in/root.tsx
```diff
@@ -55,6 +55,10 @@ export function SignInRoot({ children, path = SIGN_IN_DEFAULT_BASE_PATH }: SignI
   // TODO: eventually we'll rely on the framework SDK to specify its host router, but for now we'll default to Next.js
   const router = useNextRouter();
 
+  useEffect(() => {
+    console.log('mounted');
+  }, []);
+
```
BRKalow, line 61:
```suggestion
```

packages/elements/src/react/common/form/otp.tsx
LekoArts: `none` is invalid:
CleanShot 2024-03-22 at 09 21 53

packages/elements/src/react/common/form/otp.tsx
```diff
@@ -2,18 +2,17 @@ import { Control as RadixControl } from '@radix-ui/react-form';
...
   'type' | 'autoComplete' | 'minLength' | 'maxLength' | 'inputMode' | 'pattern' | 'spellCheck'
 > & {
-  render?: (props: { value: string; status: 'cursor' | 'selected' | 'none'; index: number }) => React.ReactNode;
+  render?: (props: { value: string; status: OTPInputSegmentStatus; index: number }) => React.ReactNode;
   length?: number;
   autoSubmit?: boolean;
 };
 
-type SelectionRange = readonly [start: number | null, end: number | null];
+type SelectionRange = readonly [start: number, end: number];
```
LekoArts, line 15: I made the type closer to the actual implementation, since according to https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-textarea/input-selectionstart it can be `null`, too

packages/elements/src/react/common/form/otp.tsx
```diff
@@ -88,173 +87,83 @@ const OTPInputStandard = React.forwardRef<HTMLInputElement, Omit<OTPInputProps,
...
-      element.select = () => {
-        originalSelect();
-        setSelectionRange([0, element.value.length]);
-      };
-    }, []);
-
-    // Continuously update the selection range when the input is focused
-    React.useEffect(() => {
-      if (!isFocused) return;
+    const isFocused = () => document.activeElement === innerRef.current;
```
LekoArts, line 96: We should continue to use the `useFocus` hook since it covers more edge cases than this selector

packages/elements/src/react/common/form/otp.tsx
LekoArts: Might have been a left-over from testing, I think it should be covered by the existing listeners

packages/elements/src/react/common/form/otp.tsx
```diff
@@ -88,173 +87,83 @@ const OTPInputStandard = React.forwardRef<HTMLInputElement, Omit<OTPInputProps,
...
-        </style>
+      `}</style>
         <RadixControl
           ref={innerRef}
           {...rest}
-          value={value}
-          onSelect={e => {
-            _select();
-            rest?.onSelect?.(e);
+          onBlur={event => {
```
LekoArts, line 131: `onBlur` is not necessary since the `useFocus` hook is already handling this

packages/elements/src/react/common/form/otp.tsx
```diff
@@ -88,173 +87,83 @@ const OTPInputStandard = React.forwardRef<HTMLInputElement, Omit<OTPInputProps,
...
-              selectionRange[1] > i;
-
+          {Array.from({ length }).map((_, i) => {
             return (
               <React.Fragment key={`otp-segment-${i}`}>
                 {render({
-                  value,
-                  status: isCursor ? 'cursor' : isSelected ? 'selected' : 'none',
+                  value: String(props.value)[i] || '',
+                  status:
```
LekoArts, line 159: This nested ternary logic is hard too read for me so I'd prefer to keep the separation like previously

LekoArts (changes requested):
I've left more comments in Slack

packages/elements/src/react/common/form/otp.tsx
BRKalow: What edge cases?

packages/elements/src/react/common/form/otp.tsx
BRKalow: sounds good 👍

packages/elements/src/react/common/form/otp.tsx
BRKalow: perfect! 😄

packages/elements/src/react/common/form/otp.tsx
LekoArts: Ah, I forgot about SDK-1366 and that I haven't handled that yet. So right now the hook's only difference is checking for the existence of ref, otherwise it should behave the same. Once we use https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/interactions/src/useFocus.ts we can handle e.g. `iframe` edge cases

PR #3032, opened 2024-03-22
fix(elements): Rename internal helper type

LekoArts:
## Description

Small follow-up for https://github.com/clerk/javascript/pull/3026 - rename the types in the Step components to not overwrite the exported function one.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/react/sign-in/step.tsx
```diff
@@ -10,8 +10,8 @@ export const SIGN_IN_STEPS = {
   'choose-strategy': 'choose-strategy',
 } as const;
 
-export type SignInStep = (typeof SIGN_IN_STEPS)[keyof typeof SIGN_IN_STEPS];
-type StepWithProps<N extends SignInStep, T> = { name: N } & T;
+export type TSignInStep = (typeof SIGN_IN_STEPS)[keyof typeof SIGN_IN_STEPS];
```
LekoArts, line 13: Core of the change - renamed it so that it's not the same as the exported function. Previously if you e.g. hovered the type you'll get the JSDoc for the function

BRKalow (approved):

PR #3034, opened 2024-03-22
feat(clerk-js,localizations,shared,types): Prompt user to reset pwned…

yourtallness:
… password at sign-in

## Description

If admin has enabled `password_settings.enforce_on_sign_in` and HIBP is enabled, then a password could potentially be detected as pwned at a subsequent sign-in.

The API will respond with error code `form_password_pwned`, in which case we will show a corresponding error and show the alternative method list, prompting them to reset their password.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/SignIn/AlternativeMethods.tsx
+++ b/packages/clerk-js/src/ui/components/SignIn/AlternativeMethods.tsx
@@ -12,11 +12,13 @@ import { SignInSocialButtons } from './SignInSocialButtons';
 import { useResetPasswordFactor } from './useResetPasswordFactor';
 import { withHavingTrouble } from './withHavingTrouble';
 
+type AlternativeMethodsMode = 'forgot' | 'pwned' | 'default';
+
 export type AlternativeMethodsProps = {
   onBackLinkClick: React.MouseEventHandler | undefined;
   onFactorSelected: (factor: SignInFactor) => void;
   currentFactor: SignInFactor | undefined | null;
-  asForgotPassword?: boolean;
+  mode?: AlternativeMethodsMode;
 };
 
 export type AlternativeMethodListProps = AlternativeMethodsProps & { onHavingTroubleClick: React.MouseEventHandler };
@@ -28,26 +30,24 @@ export const AlternativeMethods = (props: AlternativeMethodsProps) => {
 };
 
 const AlternativeMethodsList = (props: AlternativeMethodListProps) => {
-  const { onBackLinkClick, onHavingTroubleClick, onFactorSelected, asForgotPassword = false } = props;
+  const { onBackLinkClick, onHavingTroubleClick, onFactorSelected, mode = 'default' } = props;
   const card = useCardState();
   const resetPasswordFactor = useResetPasswordFactor();
   const { firstPartyFactors, hasAnyStrategy } = useAlternativeStrategies({
     filterOutFactor: props?.currentFactor,
   });
 
+  const flowPart = determineFlowPart(mode);
+  const cardTitleKey = determineTitle(mode);
+  const isReset = determineIsReset(mode);
+
   return (
-    <Flow.Part part={asForgotPassword ? 'forgotPasswordMethods' : 'alternativeMethods'}>
+    <Flow.Part part={flowPart}>
       <Card.Root>
         <Card.Content>
           <Header.Root showLogo>
-            <Header.Title
-              localizationKey={localizationKeys(
-                asForgotPassword ? 'signIn.forgotPasswordAlternativeMethods.title' : 'signIn.alternativeMethods.title',
-              )}
-            />
-            {!asForgotPassword && (
-              <Header.Subtitle localizationKey={localizationKeys('signIn.alternativeMethods.subtitle')} />
-            )}
+            <Header.Title localizationKey={cardTitleKey} />
+            {!isReset && <Header.Subtitle localizationKey={localizationKeys('signIn.alternativeMethods.subtitle')} />}
           </Header.Root>
           <Card.Alert>{card.error}</Card.Alert>
           {/*TODO: extract main in its own component */}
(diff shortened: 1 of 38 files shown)


packages/clerk-js/src/ui/components/SignIn/SignInFactorOnePasswordCard.tsx
```diff
@@ -58,6 +60,7 @@ export const SignInFactorOnePasswordCard = (props: SignInFactorOnePasswordProps)
   const clerk = useClerk();
 
   const goBack = () => {
+    card.setError(undefined);
```
yourtallness, line 63: This doesn't work, the error is not cleared.

packages/clerk-js/src/ui/components/SignIn/SignInFactorOnePasswordCard.tsx
```diff
@@ -83,6 +86,14 @@ export const SignInFactorOnePasswordCard = (props: SignInFactorOnePasswordProps)
           return clerk.__internal_navigateWithError('..', err.errors[0]);
         }
 
+        if (isPasswordPwnedError(err)) {
+          if (onPasswordPwned) {
+            card.setError({ ...err.errors[0], code: err.errors[0].code + '__sign_in' });
```
yourtallness, line 91: Need to add a suffix to target the l10n key we want.

desiprisg: I'd say just hardcode it here since the type will always be the same.

packages/clerk-js/src/ui/components/SignIn/SignInFactorOnePasswordCard.tsx
desiprisg, line 63: Maybe we ignore `undefined` ? Or maybe there's a function to clear the error? Could you check the implementation?

packages/clerk-js/src/ui/components/SignIn/SignInFactorOnePasswordCard.tsx
```diff
@@ -40,12 +41,13 @@ const usePasswordControl = (props: SignInFactorOnePasswordProps) => {
         : onShowAlternativeMethodsClick
         ? onShowAlternativeMethodsClick
         : () => null,
+      // onPasswordPwned,
```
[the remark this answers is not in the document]
desiprisg: not sure why you need it here?

packages/clerk-js/src/ui/components/SignIn/SignInFactorOne.tsx
```diff
@@ -135,6 +138,10 @@ export function _SignInFactorOne(): JSX.Element {
           }}
           onForgotPasswordMethodClick={resetPasswordFactor ? toggleForgotPasswordStrategies : toggleAllStrategies}
           onShowAlternativeMethodsClick={toggleAllStrategies}
+          onPasswordPwned={() => {
+            setIsPasswordPwned(true);
+            toggleForgotPasswordStrategies();
```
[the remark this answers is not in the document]
desiprisg: agreed

packages/clerk-js/src/ui/components/SignIn/SignInFactorOnePasswordCard.tsx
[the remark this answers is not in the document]
yourtallness: Me neither :-D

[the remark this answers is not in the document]
yourtallness: @desiprisg do we not need to export is like the other handlers?

packages/clerk-js/src/ui/components/SignIn/SignInFactorOnePasswordCard.tsx
yourtallness: Never mind, this was not the correct location to do it.

yourtallness: It worked when I moved it to the proper back handlers.

packages/clerk-js/src/ui/components/SignIn/AlternativeMethods.tsx
```diff
@@ -179,3 +167,35 @@ export function getButtonIcon(factor: SignInFactor) {
...
+      return 'alternativeMethods';
+  }
+}
+
+function determineTitle(mode: AlternativeMethodsMode): LocalizationKey {
+  switch (mode) {
+    case 'forgot':
+      return localizationKeys('signIn.forgotPasswordAlternativeMethods.title');
+    case 'pwned':
+      return localizationKeys('signIn.passwordPwned.title');
```
desiprisg, line 187: Was this localization key already on the types or do we need to add it?

desiprisg (commented):
If we could also get a couple of tests in to make sure this functionality doesn't break in the future, that would be superb!

yourtallness:
> If we could also get a couple of tests in to make sure this functionality doesn't break in the future, that would be superb!

I have added 2 tests (phone + pwned, email + pwned), can you recommend more scenarios?

packages/clerk-js/src/ui/components/SignIn/SignInFactorOnePasswordCard.tsx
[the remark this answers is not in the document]
panteliselef: Seems like we don't need it, can we remove the comment ?

packages/shared/src/error.ts
```diff
@@ -67,6 +67,10 @@ export function isUserLockedError(err: any) {
   return isClerkAPIResponseError(err) && err.errors?.[0]?.code === 'user_locked';
 }
 
+export function isPasswordPwnedError(err: any) {
+  return isClerkAPIResponseError(err) && err.errors?.[0]?.code === 'form_password_pwned';
```
panteliselef, line 71: How come we use the first error item ? Will this always be the case from FAPI ? I see that we are doing the same above, just asking tho.

panteliselef (approved):

PR #3035, opened 2024-03-22
chore(clerk-js): Improve Token.create to handle network failures errors

dimkl:
## Description

Error resolved by returning an empty string or null instead of trying to decode the `jwt` property from the response data. After this fix an `token.getRawString()` will return an empty string and  `session.getToken()` will return `null`  on Network failure and when request fails and browser is offline.

Current error message example:
```
Error: ClerkJS: Token refresh failed (error='Cannot read properties of null (reading 'jwt')')
```

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/tokenCache.ts
+++ b/packages/clerk-js/src/core/tokenCache.ts
@@ -80,6 +80,10 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
 
     entry.tokenResolver
       .then(newToken => {
+        if (!newToken.jwt) {
+          return deleteKey();
+        }
+
         const expiresAt = newToken.jwt.claims.exp;
         const issuedAt = newToken.jwt.claims.iat;
         const expiresIn: Seconds = expiresAt - issuedAt;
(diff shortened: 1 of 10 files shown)


packages/clerk-js/src/core/tokenCache.ts
```diff
@@ -80,6 +80,10 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
 
     entry.tokenResolver
       .then(newToken => {
+        if (!newToken.jwt) {
+          return deleteKey();
+        }
```
nikosdouvlis, line 85: Was this issue caused by an error thrown inside the `tokenResolver` promise?

nikosdouvlis (approved):

packages/clerk-js/src/core/tokenCache.ts
dimkl: This change attempts to fix the case where the network is offline causing the `Token.create` to return a `newToken : null`. Currently an error related to `Cannot read properties of null (reading 'jwt')` is thrown due to code in the following lines. To mitigate that (with the code above) I've added this guard to delete the key (if it exists) and return since there is no cookie to set.
Does this make any sense?

PR #3038, opened 2024-03-22
fix(clerk-js): Bundle the `zxcvbn` library inside the headless variant

anagstef:
## Description

Currently, the `zxcvbn` is lazy loaded on clerk-js due to its size. However, using `import()` is not working in non-browser environments, including Expo.

Considering the assumption that the `clerk.headless.js` variant is only used on native environments, this PR bundles the whole `zxcvbn` inside the headless clerk-js.

This PR will fix the usage of the `validatePassword` on Expo.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


BRKalow:
@anagstef Let's add an e2e here that uses the headless entry and calls `validatePassword()` 👀

panteliselef:
@anagstef I think we should also mention the increase in the bundle size as it is significant.

anagstef:
> @anagstef Let's add an e2e here that uses the headless entry and calls validatePassword() 👀

@BRKalow we should find a way to mimic the non-browser environment. We should give a shot to https://wix.github.io/Detox/
> @anagstef I think we should also mention the increase in the bundle size as it is significant.

@panteliselef you mean on the changeset?

packages/clerk-js/webpack.config.js
```diff
@@ -155,6 +155,9 @@ const commonForProd = () => {
       globalObject: 'globalThis',
     },
     plugins: [
+      new webpack.DefinePlugin({
+        __IS_BROWSER__: variant !== variants.clerkHeadless,
+      }),
```
nikosdouvlis, line 160: My understanding is that this is only used in `clerkHeadless` package, so let's kill the `variant` argument from here and move the plugin definition to the clerkHeadless rules in `prodConfig` (probably line 196?)

packages/clerk-js/src/utils/zxcvbn.ts
```diff
@@ -3,17 +3,32 @@ import type { ZxcvbnResult } from '@clerk/types';
...
+
+// We can only dynamically import zxcvbn in the browser.
+// On the headless variant we need to require it, and
+// this is because import() uses the document object
+// which is not available in Expo or other native environments.
+const loadLib = () => {
+  let core;
+  let languageCommon;
+  if (__IS_BROWSER__) {
+    core = import('@zxcvbn-ts/core');
```
nikosdouvlis, line 27: ❓ Is this a **sync** import? 
Pretty sure this is not widely supported yet - this should probably be `await`ed here.

nikosdouvlis: Is there a webpack config we can leverage where dynamic imports are "ignored" and instead they are bundled as normal?

anagstef: > Is this a sync import?
This is followed by:
```
return Promise.all([core, languageCommon]);
```
so, we don't expect it to be sync.

anagstef: > Is there a webpack config we can leverage where dynamic imports are "ignored" and instead they are bundled as normal?

Didn't find anything. :(

packages/clerk-js/src/utils/zxcvbn.ts
```diff
@@ -2,33 +2,32 @@ import type { ZxcvbnResult } from '@clerk/types';
...
 // which is not available in Expo or other native environments.
-const loadLib = () => {
+const loadLib = async () => {
   let core;
   let languageCommon;
   if (__IS_NATIVE__) {
-    core = require('@zxcvbn-ts/core');
-    languageCommon = require('@zxcvbn-ts/language-common');
+    core = await require('@zxcvbn-ts/core');
+    languageCommon = await require('@zxcvbn-ts/language-common');
```
panteliselef, line 27: Isn't require synchronous ?

packages/clerk-js/src/utils/zxcvbn.ts
```diff
@@ -2,33 +2,32 @@ import type { ZxcvbnResult } from '@clerk/types';
...
   } else {
-    core = import('@zxcvbn-ts/core');
-    languageCommon = import('@zxcvbn-ts/language-common');
+    core = await import('@zxcvbn-ts/core');
+    languageCommon = await import('@zxcvbn-ts/language-common');
   }
-  return Promise.all([core, languageCommon]);
+  return { core, languageCommon };
```
panteliselef: Does this mean the the dependencies will no longer load in parallel ?

packages/clerk-js/src/utils/zxcvbn.ts
LekoArts: @anagstef I also think that the `await` before the require isn't doing anything 👍

anagstef: Yes, I've removed this, but haven't pushed yet! Thank you both for noticing!

packages/clerk-js/src/utils/zxcvbn.ts
anagstef: Hmm. Nice catch! I'll revert this one.

nikosdouvlis:
@anagstef do we have any updates on this one?

nikosdouvlis:
This looks like a stale PR. Closing for now, lets reopen if we need to.

PR #3041, opened 2024-03-22
chore(e2e): Change the domain of fictional emails so we can cleanup user corerctly

octoper:
## Description

This PR updates the domain that is used for the fictional emails we are generating for Integration tests.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


anagstef (approved):

integration/testUtils/usersService.ts
```diff
@@ -67,7 +67,7 @@ export const createUserService = (clerkClient: ClerkClient) => {
         withUsername = false,
       } = options || {};
       const randomHash = hash();
-      const email = fictionalEmail ? `${randomHash}+clerk_test@example.com` : `clerkcookie+${randomHash}@mailsac.com`;
+      const email = fictionalEmail ? `${randomHash}+clerk_test@clerkcook.ie` : `clerkcookie+${randomHash}@mailsac.com`;
```
LekoArts, line 70: Wouldn't `clerkcookie.com` also work? Then you could leave the `query` in the `cleanup.setup.ts` the same

integration/tests/sign-in-flow.test.ts
```diff
@@ -109,8 +109,8 @@ testAgainstRunningApps({ withEnv: [appConfigs.envs.withEmailCodes] })('sign in f
     await u.po.signIn.getForgotPassword().click();
     await u.po.signIn.getResetPassword().click();
     await u.po.signIn.enterTestOtpCode();
-    await u.po.signIn.setPassword(`${fakeUserWithPasword.password}_reset`);
-    await u.po.signIn.setPasswordConfirmation(`${fakeUserWithPasword.password}_reset`);
+    await u.po.signIn.setPassword('password_reset');
```
LekoArts, line 112: Why was this change necessary?

octoper: Yeah it was failing as the new password which is 2 times the email was too long.

integration/testUtils/usersService.ts
octoper: We also use email as password and there is a limit on characteres so I tried to keep it short

LekoArts: How about changing the password to `{email}{randomHash}` instead, then you could use the full `clerkcookie` name and wouldn't need to change the individual test. Better to fix the root cause than to fix it in an individual test. This way no one else will run into it.

What do you think? :)

nikosdouvlis: Agree with Lennart here :)

LekoArts: The hash will be inside the email name so I think you can still figure out the password 👍

LekoArts (approved):

PR #3051, opened 2024-03-26
chore(repo): Configure Dependabot [SDK-1557]

tmilewski:
## Description

Configures Dependabot. See code comments for applied settings.

SDK-1557

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.github/dependabot.yml
```diff
@@ -0,0 +1,82 @@
...
+      timezone: "Etc/UTC"
+    commit-message:
+      prefix: "chore(repo)"
+      include: "scope"
+    groups:
+      "Dependency updates [Core 1]":
+        update-types:
+          - "major"
+          - "minor"
+          - "patch"
```
jescalan, line 66: I feel like we only wanna apply security patches to core 1, yeah?

.github/dependabot.yml
```diff
@@ -0,0 +1,82 @@
...
+
+  # This configuration will:
+  # - Check for updates weekly on Wednesday at 12:00 UTC
+  # - Create one pull request at a time
+  # - Commit changes with the prefix `chore(repo)` and include the scope
```
jescalan, line 71: I'm confused about this one - why is it run on a different day and whats the difference from the other ones?

tmilewski: To offset any required changes, but can standardize. This will be way less frequent than the other two.

jescalan: Still not sure I understand what this means

LekoArts: The difference is the [`package-ecosystem` key](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem). So the first one will update the package.json entries, the second one will update the GitHub actions.

LekoArts (changes requested):
See convo in Slack, I'm not sure if Dependabot is the right tool for our use cases

PR #3054, opened 2024-03-26
fix(elements): Check for controlled values and handle accordingly

alexcarpenter:
## Description

Enables controlled inputs for clerk/elements.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/react/common/form/index.tsx
```diff
@@ -148,12 +148,16 @@ const useInput = ({
...
-      if (!name) return;
+      if (!name || initialValue) return;
       ref.send({ type: 'FIELD.UPDATE', field: { name, value: event.target.value } });
     },
-    [ref, name, onChangeProp],
+    [ref, name, onChangeProp, initialValue],
   );
 
+  React.useEffect(() => {
+    ref.send({ type: 'FIELD.UPDATE', field: { name, value: initialValue } });
```
tmilewski, line 158: ❓ Should we check if name/initialValue are present? No real need to send the event if the initialValue isn't present, right?

```suggestion
    ref.send({ type: 'FIELD.UPDATE', field: { name, value: initialValue } });
```

tmilewski (changes requested):

packages/elements/src/react/common/form/index.tsx
alexcarpenter: Good callout https://github.com/clerk/javascript/pull/3054/commits/08c60f2bf355881549c20fd68708a03a1d7c71a0

tmilewski (approved):
lgtm, thanks!

PR #3057, opened 2024-03-27
fix(backend): Fix typo in jwk-remote-missing error message

dimkl:
## Description

Fix typo

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/chilly-tips-compete.md
+++ b/.changeset/chilly-tips-compete.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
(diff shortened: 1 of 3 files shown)


LekoArts:
It can be a single `kid` or multiple `kid` since it's an array that gets concenated to a comma-delimited list. That's why I chose `are` - if we want to have the pluralization correct all the time we should use `is` / `are` depending on how many items the `cacheValues` array has.

anagstef (approved):

PR #3062, opened 2024-03-28
fix(clerk-js): Correct chunk loading logic in clerk.headless.browser variant

LekoArts:
## Description

We were missing the `setWebpackChunkPublicPath` util in the `clerk.headless.browser` variant, making this variant break when there was a difference between latest published version and the pinned version in the cloudflare proxy. The util is necessary to replace chunk loading snippets with the fixed version of its entry file version.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/lazy-games-destroy.md
+++ b/.changeset/lazy-games-destroy.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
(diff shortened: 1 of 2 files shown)


BRKalow (approved):
good catch!

PR #3092, opened 2024-04-02
fix(clerk-js): Avoid depending on `count` as it can be zero but invitations may still exist

panteliselef:
## Description

**Background**: In order to provide a better UX, we are manually updating the cache where userInvitations are stored. This allows us to update the invitation in place and avoid layout shifts, if we would simply revalidate this would not be possible.

We are also manually updating `count` in the local cache in order to display the correct number of invitations available, without revalidating as this would cause a layout shift.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/five-shoes-promise.md
+++ b/.changeset/five-shoes-promise.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Avoid depending on `count` as it can be zero but invitations may still exist.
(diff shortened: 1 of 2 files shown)


octoper:
!preview

dimkl (commented):
❓ what is the actual use case where we want to have count `0` with invitations?
If there is no case (product wise) i think we need to open a bug ticket to the backend team to fix this.

clerk-cookie:

| :-- | :-- | :-- |
| :cook: _Deploying..._ | [Inspect](https://javascript-preview-gu3oscv38.clerkstage.dev) | Apr 02, 2024 01:38 PM |

panteliselef:
> ❓ what is the actual use case where we want to have count `0` with invitations? If there is no case (product wise) i think we need to open a bug ticket to the backend team to fix this.

@dimkl I updated the description. Generally `count` is very useful, for this particular case and because we manually update the cache instead of revalidating we need to use this differently.

octoper (approved):
💯

dimkl:
@panteliselef Is the `userInvitations` or `userSuggestions` paginated responses? If so, is the `count` related to the total count data of the response?
If the `count` is related to the total count of the paginated response, i think we still need to depend on that instead of the `length`. Wdyt?

PR #3103, opened 2024-04-03
Feat/Manage plan screen

AlexNti:
## Description

In this commit we are adding the `Manage Plans` screen of the new Billing page. This screen is going to be used from both the User and Organization Profile.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/common/Billing.tsx
```diff
@@ -0,0 +1,204 @@
...
+      setIsLoading(true);
+      await action();
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  return (
+    <Button
+      elementDescriptor={descriptors.billingChangePlanAction}
```
panteliselef, line 160: As discussed, probably this does not provide much value

packages/clerk-js/src/ui/common/Billing.tsx
```diff
@@ -0,0 +1,204 @@
...
+      variant='outline'
+      sx={t => ({ color: t.colors.$colorText })}
+      localizationKey={localizationKeys('billing.managePlanScreen.action__changePlan')}
+    />
+  );
+};
+
+const GoToPlanAndBilling = () => {
+  return (
+    <Box sx={t => ({ marginBottom: t.space.$4 })}>
```
panteliselef, line 172: Is it possible to remove this div and use `gap` from the parent `<Col/>` ?

packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationBilling.tsx
```diff
@@ -1,24 +1,69 @@
...
+            elementDescriptor={descriptors.spinner}
+          />
+        </Box>
+      </Box>
+    );
   }
 
-  if ((data?.data.length || 0) < 1) {
-    return <>No data</>;
-  }
```
panteliselef, line 17: Couldn't be possible to have 0 available plans ? What will happen if you have a "current" plan but the app owner removes all the entries from the dashboard. Will "available plans" still be populated with the current plan at least ?

packages/clerk-js/src/ui/common/Billing.tsx
```diff
@@ -0,0 +1,204 @@
...
+import { Header, IconButton } from '../elements';
+import { ArrowLeftIcon, Check } from '../icons';
+
+export type OrganizationPlanCardProps = Pick<BillingPlanResource, 'name' | 'features' | 'priceInCents'> & {
+  isCurrentPlan: boolean;
+  planKey: string;
+  changePlanAction: React.ReactNode;
+};
+
+export const OrganizationPlanCard = (params: OrganizationPlanCardProps) => {
```
panteliselef, line 14: User and Org plan cards will be identical, right ?

packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationBilling.tsx
```diff
@@ -1,24 +1,69 @@
...
+      <OrganizationPlanCard
+        isCurrentPlan={currentPlan?.key === key}
+        planKey={key}
+        key={id}
+        name={name}
+        features={features}
+        priceInCents={priceInCents / 100}
+        changePlanAction={<ChangePlanButton action={() => handleChangePlan(key)} />}
+      />
+    );
```
panteliselef, line 65: Based on the assumption that we will not have different cards between users and orgs, can't we move this logic inside `ManagePlaceScreen` ?

I'd either keep the same API but `plans` should be of type "availablePlan", or simple accept children
```tsx
<MangePlanScreen>
  {plans}
</MangePlanScreen>

```

packages/clerk-js/src/ui/customizables/elementDescriptors.ts
```diff
@@ -261,6 +261,9 @@ export const APPEARANCE_KEYS = containsAllElementsConfigKeys([
   'buttonArrowIcon',
   'providerIcon',
   'spinner',
+
+  'billingPlanCard',
+  'billingChangePlanAction',
```
panteliselef, line 266: > As discussed, probably this does not provide much value

`billingChangePlanAction` to be removed

packages/clerk-js/src/ui/common/Billing.tsx
```diff
@@ -0,0 +1,204 @@
...
+  return (
+    <Col
+      elementDescriptor={descriptors.billingPlanCard}
+      elementId={descriptors.billingPlanCard.setId(params.planKey)}
```
AlexNti, line 18: @anagstef @octoper Do you have any objection on using dynamic ID for the element descriptor?

anagstef: @AlexNti unfortunately, elementId is used to generate Public API classnames in the form of `elementDescriptor__elemendId`. So, this is considered Public API that any user can target. :(

anagstef: After discussing with @AlexNti , these plan keys will be set by the developer, so this is totally ok!

packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationBilling.tsx
AlexNti: Initially like this, but the `changePlanAction` will be different for orgs and users. I moved this outside in order to avoid prop drilling of the  `handleChangePlan` function. 

I can change the `ManagePlanScreen` to accept children instead of plans.

packages/clerk-js/src/ui/common/Billing.tsx
AlexNti: Yes they will be identical.

packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationBilling.tsx
AlexNti: It isn't possible to have 0 available plans as when we have the Billing enabled there will be always at least one available plan

panteliselef (approved):

PR #3121, opened 2024-04-04
feat(nextjs): Add signature verification of token passed from middleware

BRKalow:
## Description

Introduce signature verification of the JWT that is passed from middleware to the Next.js origin. We're only introducing this functionality into the Next.js SDK as it's the only framework (currently) that involves a network hop between verification of the token and usage.

The goal here was to retain the sync API of `auth()`, and so we're not using the native async crypto modules at this time. Instead, we're using `crypto-js`, which exposes sync crpyto APIs.

To generate a signature, we're passing the verified JWT and the app's clerk secret key to generate an HMAC with SHA1. The performance impact of this is negligible, with a <1ms measured impact locally. SHA1 is acceptable for use here, as the token has already been verified and we aren't concerned with a collision. We're validating that the token we verified in the middleware is the same one we're using at the app's origin.

This also further mitigates the IDOR vulnerability from earlier this year, ensuring we are always using the previously verified token.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


panteliselef:
@BRKalow is this something that only concerns nextjs ? Can't we move this to backend ?

BRKalow:
@panteliselef Yeah it's only a Next.js concern, due to the handoff from middleware -> app backend.

packages/nextjs/package.json
```diff
@@ -70,11 +70,13 @@
     "@clerk/backend": "1.0.0-beta.33",
     "@clerk/clerk-react": "5.0.0-beta.36",
     "@clerk/shared": "2.0.0-beta.21",
+    "crypto-js": "^4.2.0",
```
nikosdouvlis, line 73: Let's pin the version here

packages/nextjs/package.json
```diff
@@ -70,11 +70,13 @@
...
     "path-to-regexp": "6.2.1",
     "tslib": "2.4.1"
   },
   "devDependencies": {
     "@clerk/types": "4.0.0-beta.25",
+    "@types/crypto-js": "^4.2.2",
```
nikosdouvlis, line 79: And here as well

nikosdouvlis (approved):
👏🏻

PR #3091, opened 2024-04-02
fix(shared): Automatic hook revalidation v4

panteliselef:
### Before

https://github.com/clerk/javascript/assets/19269911/e4da2dff-971d-4d9a-8ec3-1e483085e798

### After

https://github.com/clerk/javascript/assets/19269911/0c583fc2-7aab-4737-b678-0e09cdf38938

## Checklist

- [ ] `npm test` runs as expected.
- [ ] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #3126, opened 2024-04-05
fix(clerk-js,shared,react): Use common cache for swr

panteliselef:
## Description

Alternative solution to #3091 that aims to solve the same issue: How to sync data between hooks and UI components.

A common issue with Clerk so far is that the react hooks that are bundled with a host application will have a different cache than the once used with the bundled UI components from clerk-js.

This PR solves that issue by introducing a single common cache that is created a top level and is passed down to any dependencies.

### Scenarios
1. clerk-react has its own cache and it is passed down to clerk-js via the `load` method (discarding the clerk-js default cache).
2. clerk-js is used directly and it uses its own cache.

This ensures that when calling `useOrganizationList({memberships:true})` from the host app, if the hook has already run and fetched data due to a UI component having been rendered then this will be a cache hit (if data are not stale). Previous that would be a cache miss.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/brave-panthers-eat.md
+++ b/.changeset/brave-panthers-eat.md
@@ -1,0 +1,8 @@
+---
+'@clerk/clerk-js': minor
+'@clerk/shared': minor
+'@clerk/clerk-react': minor
+'@clerk/types': minor
+---
+
+Create a common cache between clerk-react and clerk-js in order for hooks to mounted inside Clerk UI components and a host application to share data.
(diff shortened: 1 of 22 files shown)


panteliselef:
Closing this as backporting this to v4 is not a priority at the moment

PR #3154, opened 2024-04-09
feat(clerk-js): Support different CAPTCHA widget types

anagstef:
## Description
This PR adds support for different CAPTCHA widgets coming from the FAPI.

### Current behavior:
Currently, we support only an invisible captcha widget. This widget gets rendered in an injected div element at the end of the document body.

### After this PR:
For backward compatibility, we will support the previous functionality when the user has selected the `"invisible"` widget. This is the setting to which all current users will be migrated.

This PR adds support for a visible challenge if needed on sign-up. This challenge gets rendered on the `<div id="clerk-captcha" />` element in the SignUp form.

Users with custom sign-up flows must have the `<div id="clerk-captcha" />` element inside their DOM when calling the `signUp.create` function.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/eighty-pens-mix.md
+++ b/.changeset/eighty-pens-mix.md
@@ -1,0 +1,6 @@
+---
+'@clerk/clerk-js': minor
+'@clerk/types': minor
+---
+
+Add support for different CAPTCHA widget types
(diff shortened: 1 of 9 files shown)


dimkl (approved):

BRKalow:
@anagstef can this be styled by consumers? If yes, how will we support that?

anagstef:
@BRKalow it can receive a theme between `light` and `dark`, a locale, and a size property. Currently, all of these default to `auto` which gets data from the browser. We can expose these variables through the `appearance` prop, at a later time.

panteliselef (approved):
LGTM 💯

PR #3156, opened 2024-04-09
feat(clerk-react,nextjs): Speed up clerk-js loading by using a `<script/>` tag

panteliselef:
### Metrics (Fast 3G)
Previous 4120ms
New Nextjs 2907ms
-30% in loading time

### Metrics (Cable)
Previous 613ms
New Nextjs 617ms
Almost no change

### Nextjs
Benefits of using the `next/script` is that it moves scripts inside `<head/>` in Pages router. It also has build it cache to avoid loading 2 scripts at the same time.

Using `next/script` with our ClerkProvide that targets the App Router was not possible as `next/script` need to be used inside the "document" which means inside `<html/>`.

### More SSR Frameworks
- Remix does not a similar component as Next.js and it will not automatically move scripts inside `<head/>`.
- Gatsby seems to offer one, but it lacks the "beforeInteraction" strategy which is the one that will add the script on generated html before hydration.
For these I would offer a dedicated `<ClerkJSScript/>`.

### CSR React
- Supporting a script tag in a CSR app does not improve loading times.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/react/src/utils/loadClerkJsScript.ts
```diff
@@ -11,17 +11,31 @@ import { versionSelector } from './versionSelector';
...
+      existingScript.addEventListener('load', () => {
+        resolve(existingScript);
+      });
+
+      existingScript.addEventListener('error', () => {
+        reject(FAILED_TO_LOAD_ERROR);
+      });
+    });
+  }
+
```
panteliselef, line 38: This also allows anyone to add a script without waiting for an official support.

packages/nextjs/src/utils/clerk-js-script.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const Script = props.router === 'app' ? 'script' : NextScript;
+
+  return (
+    <Script
+      src={scriptUrl}
+      data-clerk-js-script
+      async
+      crossOrigin='anonymous'
+      {...buildClerkJsScriptAttributes(options)}
+      {...(props.router === 'pages' ? { strategy: 'beforeInteractive' } : {})}
```
BRKalow, line 40: Do we want `beforeInteractive`? This will delay the app's hydration, based on the docs: https://nextjs.org/docs/pages/building-your-application/optimizing/scripts#strategy

I'd say we use the default setting, `afterInteractive`. We don't want to optimize clerk's loading at the expense of the host app's performance.

BRKalow (approved):
One comment about the strategy passed to `next/script`, which I think we should leave as the default. Otherwise this lgtm. Nice work.

packages/nextjs/src/utils/clerk-js-script.tsx
panteliselef: I've tested both, `beforeInteractive` does not seem to have such effect. On the other hand if we use `afterInteractive` fetching of clerk-js will start after hydration making the changes of this PR to not have any effect.


image

nikosdouvlis: I thought what @BRKalow would happen as well, but after testing it seems that `beforeInteractive` does not delay hydration. This is a little surprising to me, I wonder if there's another test we could run to validate  it

nikosdouvlis, line 40:
```suggestion
 strategy: props.router === 'pages' ? 'beforeInteractive' : undefined
```

nikosdouvlis (approved):
👏🏻

dimkl (approved):

packages/nextjs/src/utils/clerk-js-script.tsx
BRKalow: Makes sense! Nice job with the thorough testing 😃

PR #3158, opened 2024-04-10
chore(clerk-js): Only allow permission authorizations internally

panteliselef:
## Description

Internally we should **not** depend on roles. Since custom roles was introduced we have updated our internal clerk-js component to work with permissions only.

There are no references to `<Protect role="..."/>` in clerk-js.

This PR narrows the types in order to avoid confusion when protect is used inside clerk-js. Public APIs are not affected.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


nikospapcom (approved):

dimkl:
@panteliselef Why did we remove the `role`? I think that we should either provide both `role` & `permission` or none since most of the time `role` is modeled as a group of permissions.

panteliselef:
@dimkl i've updated the description, it should be clear why we can do this.

dimkl (approved):

AlexNti (approved):

PR #3161, opened 2024-04-10
Emmanouela/sdk 832 add test files in   tests   directories

EmmanouelaPothitou:
Make the tests location consistent across all packages, using __tests__ folders.


packages/chrome-extension/src/__tests__/exports.test.ts
```diff
@@ -1,4 +1,4 @@
-import * as publicExports from './index';
+import * as publicExports from '..';
```
nikosdouvlis, line 1:
```suggestion
import * as publicExports from '../index';
```

nikosdouvlis (approved):

PR #3162, opened 2024-04-10
Introduce force and fallback redirect urls

nikosdouvlis:
### The problem
Our middleware helpers (`authMiddleware` and the newer `clerkMiddleware`) will automatically redirect unauthenticated requests for protected routes to `SIGN_IN_URL` while appending the current `req.url` as the `redirect_url` param, as most of the time, redirecting the user to the page they requested is the most desirable behavior. This, however, makes the redirect props on the `SignIn` and `SignUp` components almost impossible to use for more advanced use cases. 

For example, let's imagine a SignUp flow where users need to go through onboarding after a successful sign-up.

```ts
// middleware.ts
export default clerkMiddleware()

// sign-up.ts
<SignUp afterSignUpUrl='/onboarding' />
```

The setup shown about will not work out of the box right now. If a user tries to access protected `/app` page, then:
- the middleware will redirect to `/sign-in?redirect_url=/app`
- the user will click "sign up" and they will be redirected to `/sign-up?redirect_url=/app`
- the `redirect_url` search param has priority over all other props and env vars, so after a successful sign-up, the user will be redirected to `/app` instead of `/onboarding`

There are two potential workarounds here:
- manually rewrite the URL and remove the `redirect_url` search param on `/sign-up`
- disable the default "return back url" behavior in the middleware for the specific route

Neither workarounds provide acceptable DX though.

### Proposed solution
Replace `afterSignInUrl` with `signInForceRedirectUrl`/`signInFallbackRedirectUrl` and `afterSignUpUrl` with `signUpForceRedirectUrl`/`signUpFallbackRedirectUrl`, completely remove the existing `afterSign*Url` props and alter the relative specifity of the redirect props.

More specifically:
- Introduce `CLERK_SIGN_IN_FORCE_REDIRECT_URL` and `CLERK_SIGN_IN_FALLBACK_REDIRECT_URL` env vars
- Introduce `CLERK_SIGN_UP_FORCE_REDIRECT_URL` and `CLERK_SIGN_UP_FALLBACK_REDIRECT_URL` env vars
- Remove `CLERK_AFTER_SIGN_IN_URL` and `CLERK_AFTER_SIGN_UP_URL` env vars
- Remove `afterSignInUrl`, `afterSignUpUrl`, `redirectUrl` props from `SignIn`, `SignUp` and `ClerkProvider` components
- Introduce `signInForceRedirectUrl`, `signInFallbackRedirectUrl`, `signUpForceRedirectUrl`, `signUpFallbackRedirectUrl` props on `ClerkProvider`
- Introduce `forceRedirectUrl`, `fallbackRedirectUrl`, `signUpForceRedirectUrl`, `signUpFallbackRedirectUrl` props on `SignIn`
- Introduce `forceRedirectUrl`, `fallbackRedirectUrl`, `signInForceRedirectUrl`, `signInFallbackRedirectUrl` props on `SignUp`

The new priorities (as described by the `RedirectUrls` class in the code) would be:
1. forceRedirectUrl search param || forceRedirectUrl prop || forceRedirectUrl env var
2. redirect_url search param
3. fallbackRedirectUrl search param || fallbackRedirectUrl prop || fallbackRedirectUrl env vars
(env vars are automatically passed to ClerkProvider for supported frameworks, so ClerkProvider props and env vars are interchangeable in the above list)

Even though this setup is more complex, it provides a lot more flexibility and control over the redirect behavior. We don't expect most users to need all the new props, so the basic use case will still be simple and straightforward (see the example below).

### Example usage
In a Next.js app, you could set up the following:

```ts
// .env
NEXT_PUBLIC_CLERK_SIGN_UP_FORCE_REDIRECT_URL=/onboarding
NEXT_PUBLIC_CLERK_SIGN_IN_FALLBACK_REDIRECT_URL=/app

// layout.tsx
import { ClerkProvider } from '@clerk/clerk-react'
<ClerkProvider>
    {...}
</ClerkProvider>

// app/sign-in/[[...rest]]/page.tsx
import {SignIn} from '@clerk/nextjs'
export default function SignInPage() {
    return (
        <SignIn />
    )
}

// app/sign-up/[[...rest]]/page.tsx
import {SignUp} from '@clerk/nextjs'
export default function SignUpPage() {
    return (
        <SignUp />
    )
}
```

With the new redirect props, if a user tries to access protected `/app` page, then:
- the middleware will redirect to `/sign-in?redirect_url=/app`
- the user will click "sign up" and they will be redirected to `/sign-up?redirect_url=/app`
- the `redirect_url` no longer has priority, so after a successful sign-up, the user will be redirected to `/onboarding` as defined by the env vars

If, however, the user simply signs in, they will be redirected to `/app` as the fallback URL has lower priority than the `redirect_url` search param. 

### Expected behavior
- The props will not be transferred as search params when navigating between `/sign-up` <> `/sign-in`, except for the `redirect_url`.
- The props will not be transferred as search params when navigating from localhost/app to AP and back, except for the `redirect_url`. AP will respect the `redirect_url` param or fall back to the dashboard values.

Let's now examine some common setups. The following examples assume that no env vars are defined, so all config is based on `ClerkProvider` and the components themselves.

Example 1: interaction between mounted components
```jsx
// main.ts
<ClerkProvider
    signInUrl='/sign-in'
    signUpUrl='/sign-up'
/>

<SignIn
    forceRedirectUrl='/a-force'
    signUpForceRedirectUrl='/a-su-force'
/>

<SignUp
    forceRedirectUrl='/b-force'
    signUpForceRedirectUrl='/b-si-force'
/>
```

Flow:
- user lands on /sign-in
- user clicks on "sign up" link

Behavior: 
- User signs up with email/pass or oauth
  - will be redirected to `/b-force`
- User signs in (existing user) via oauth
  - will be redirected to `/b-si-force`
- User signs in (existing user) via oauth + 2fa (transfer to `/sign-in/continue`)
  - will be redirected to `/a-force`

Example 2: interaction between mounted SignIn and AP SignUp
```jsx
// main.ts
<ClerkProvider
    signInUrl='/sign-in'
/>

<SignIn
    forceRedirectUrl='/a-force'
    signUpForceRedirectUrl='/a-su-force'
/>
```

Flow:
- user lands on `/sign-in?redirect_url=/hello`
- user clicks on "sign up" link
- user gets redirected to `AP/sign-up?redirect_url=https://localhost:3000/hello`

Behavior:
- After sign up
  - will be redirected to `https://localhost:3000/hello`

If the `redirect_url` search param did not exist, AP would use the dashboard values.

#### Regarding onboarding flows
Even though we're using onboarding as an example flow above, we should note that the best way to enforce a user goes through onboarding after sign-up is to add an attribute using the metadata API and then enforce a redirect to `/onboarding` (through middleware or the framework router) unless a user has completed the onboarding. This way, the user can't skip the onboarding by manually changing the URL.

In this case, the `signUpForceRedirectUrl` prop can be used to further optimize the flow, as the user will be redirected to the onboarding URL immediately without the need for additional middleware-triggered redirects. 

#### Account portal - TODO
Account Portal needs to be updated to use the above props instead of the current afterSignInUrl and afterSignUpUrl. The staging deployment will happen when the above changes are merged and a new beta is released.
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/constants.ts
```diff
@@ -1,4 +1,11 @@
-export const PRESERVED_QUERYSTRING_PARAMS = ['after_sign_in_url', 'after_sign_up_url', 'redirect_url'];
+// TODO: Do we still have a use for this or can we simply preserve all params?
+export const PRESERVED_QUERYSTRING_PARAMS = [
```
nikosdouvlis, line 2: This is the list of search params we preserve when navigating using our internal router. I don't recall any reason that'd block us from simply preserving all params (including user-passed params)

packages/clerk-js/src/ui/components/SignIn/SignIn.tsx
```diff
@@ -44,9 +44,8 @@ function SignInRoutes(): JSX.Element {
           <SignInSSOCallback
             signUpUrl={signInContext.signUpUrl}
             signInUrl={signInContext.signInUrl}
-            afterSignInUrl={signInContext.afterSignInUrl}
-            afterSignUpUrl={signInContext.afterSignUpUrl}
-            redirectUrl={signInContext.redirectUrl}
+            signInForceRedirectUrl={signInContext.afterSignInUrl}
+            signUpForceRedirectUrl={signInContext.afterSignUpUrl}
             continueSignUpUrl={signInContext.signUpContinueUrl}
```
nikosdouvlis, line 49: I am unsure of the API we want her - this is the component that calls `Clerk.handleRedirectCallback` internally so even if this is an internal change, the API change is still reflected  on `Clerk.handleRedirectCallback` 

Does this function need all 4 URL options or can we simply pass `redirectUrl`?

packages/clerk-js/src/ui/components/SignIn/SignIn.tsx
```diff
@@ -58,7 +57,7 @@ function SignInRoutes(): JSX.Element {
         </Route>
         <Route path='verify'>
           <SignInEmailLinkFlowComplete
-            redirectUrlComplete={signInContext.afterSignInUrl || signInContext.redirectUrl || undefined}
+            redirectUrlComplete={signInContext.afterSignInUrl}
```
nikosdouvlis, line 60: The URL is already calculated in the signIn/signUp context

packages/clerk-js/src/ui/components/SignIn/__tests__/SignInStart.test.tsx
```diff
@@ -199,7 +199,7 @@ describe('SignInStart', () => {
       expect(fixtures.signIn.create).toHaveBeenCalled();
       expect(fixtures.signIn.authenticateWithRedirect).toHaveBeenCalledWith({
         strategy: 'saml',
-        redirectUrl: 'http://localhost/#/sso-callback?redirect_url=http%3A%2F%2Flocalhost%2F',
+        redirectUrl: 'http://localhost/sso-callback',
```
nikosdouvlis, line 202: This is probably a mistake as the sso-callback needs to have a redirectURL passed-in.. investigating

packages/clerk-js/src/ui/components/SignUp/SignUp.tsx
```diff
@@ -42,9 +42,8 @@ function SignUpRoutes(): JSX.Element {
           <SignUpSSOCallback
             signUpUrl={signUpContext.signUpUrl}
             signInUrl={signUpContext.signInUrl}
-            afterSignUpUrl={signUpContext.afterSignUpUrl}
-            afterSignInUrl={signUpContext.afterSignInUrl}
-            redirectUrl={signUpContext.redirectUrl}
+            signUpForceRedirectUrl={signUpContext.afterSignUpUrl}
+            signInForceRedirectUrl={signUpContext.afterSignInUrl}
```
nikosdouvlis, line 46: Same comment as the SignInSSOCallback

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
```diff
@@ -57,54 +61,41 @@ export const useSignUpContext = (): SignUpContextType => {
   const options = useOptions();
   const clerk = useClerk();
 
+  assertNoLegacyProp(options);
+  assertNoLegacyProp(ctx);
```
nikosdouvlis, line 65: We decided that we're not going to support only the new props going forwards - we're throwing if we detect any of the old props

jescalan:
Just want to double check - in the description it says:

> Remove afterSignInUrl, afterSignUpUrl, redirectUrl props from SignIn and SignUp components

Should this also include the ClerkProvider component?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1644,31 +1642,13 @@ export class Clerk implements ClerkInterface {
     });
   };
 
-  #buildUrl = (key: 'signInUrl' | 'signUpUrl', options?: SignInRedirectOptions | SignUpRedirectOptions): string => {
-    if (!this.loaded || !this.#environment || !this.#environment.displayConfig) {
+  #buildUrl = (key: 'signInUrl' | 'signUpUrl', params?: Record<string, string>): string => {
```
panteliselef, line 1645: Shouldn't this be
```suggestion
  #buildUrl = (key: 'signInUrl' | 'signUpUrl', params?: RedirectOptions): string => {
```

packages/nextjs/src/global.d.ts
```diff
@@ -17,6 +17,10 @@ declare global {
       NEXT_PUBLIC_CLERK_SIGN_UP_URL: string | undefined;
       NEXT_PUBLIC_CLERK_AFTER_SIGN_IN_URL: string | undefined;
       NEXT_PUBLIC_CLERK_AFTER_SIGN_UP_URL: string | undefined;
```
panteliselef, line 19: Shall we drop those ?

packages/types/src/redirects.ts
```diff
@@ -1,5 +1,23 @@
...
+/**
+ * Redirect URLs for different actions.
+ * Mainly used to be used to type internal Clerk functions.
+ */
+export type RedirectOptions = {
+  signInForceRedirectUrl?: string | null;
+  signInFallbackRedirectUrl?: string | null;
+  signUpForceRedirectUrl?: string | null;
+  signUpFallbackRedirectUrl?: string | null;
+};
```
panteliselef, line 19: Wouldn't it be safer to do 
```suggestion
export type RedirectOptions = SignUpForceRedirectUrl & SignUpFallbackRedirectUrl & SignInFallbackRedirectUrl & SignInForceRedirectUrl
```

packages/types/src/clerk.ts
```diff
@@ -466,57 +475,47 @@ export interface Clerk {
...
-   * Full URL or path to navigate after requesting email verification.
-   */
-  verifyEmailAddressUrl?: string | null;
-
-  /**
-   * Full URL or path to navigate after requesting phone verification.
-   */
-  verifyPhoneNumberUrl?: string | null;
-};
+export type HandleOAuthCallbackParams = AfterSignOutUrl &
```
panteliselef, line 478: Does `AfterSignOutUrl` actually affect this API ?

packages/clerk-js/src/utils/url.ts
```diff
@@ -369,3 +388,16 @@ export function createAllowedRedirectOrigins(
...
   return origins;
 }
+
+export const isCrossOrigin = (url: string | URL, origin: string | URL = window.location.origin): boolean => {
+  try {
+    if (isRelativeUrl(url)) {
+      return false;
+    }
+    const urlOrigin = new URL(url).origin;
+    const originOrigin = new URL(origin).origin;
```
panteliselef, line 398: Shouldn't this be always `window.location.origin` ? why do we need to grab it as param ?

panteliselef (approved):

packages/types/src/redirects.ts
nikosdouvlis: Nice one, thanks!

packages/types/src/clerk.ts
nikosdouvlis: Nice catch, this existed before but I checked and we never had a use for it ✅

nikosdouvlis:
> Just want to double check - in the description it says:
> 
> > Remove afterSignInUrl, afterSignUpUrl, redirectUrl props from SignIn and SignUp components
> 
> Should this also include the ClerkProvider component?

@jescalan , correct, updating the description

packages/clerk-js/src/utils/url.ts
nikosdouvlis: To allow for easier testing without having to mock window.location :)

nikosdouvlis:
@panteliselef comments resolved, appreciate the review 🙏🏻

PR #3168, opened 2024-04-11
feat(elements): Support fallback prop for flow roots

BRKalow:
## Description

Supports a `fallback` prop on the flow `Root` components. This will be rendered before Clerk has loaded. As part of this, I also adjusted the flows to support some rendering of the Elements components without a Clerk instance. This requires that we're safer when accessing data off of `context.clerk`, as the flow providers will now render before Clerk has loaded and so `context.clerk` might be undefined. I added a `SET_CLERK` event to each router machine to ensure the loaded clerk instance gets passed on load.

I also refactored the third party provider machine and hook to remove the derived `thirdPartyProviders` state. I moved this logic into a hook and fixed a bug in how we detect whether or not a provider is enabled. The machine didn't rely on the data, and with the changes here we remove the need to pass environment data through a few machines.

fallback prop:
```tsx
<SignIn.Root fallback={<SignInLoading />}>
  ...
</SignIn.Root>
```

SDK-1626

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/internals/machines/sign-up/machines/router.machine.ts
```diff
@@ -68,8 +68,8 @@ export const SignUpRouterMachine = setup({
     transfer: ({ context }) => context.router?.push(context.clerk.buildSignInUrl()),
   },
   guards: {
-    areFieldsMissing: ({ context }) => context.clerk.client.signUp.missingFields.length > 0,
-    areFieldsUnverified: ({ context }) => context.clerk.client.signUp.unverifiedFields.length > 0,
+    areFieldsMissing: ({ context }) => context.clerk?.client?.signUp?.missingFields?.length > 0,
```
BRKalow, line 71: Might need to be a bit more thorough with applying optional chaining here. Will do another pass.

packages/elements/examples/nextjs/app/sign-up/[[...sign-up]]/page.tsx
```diff
@@ -21,7 +21,70 @@ function CustomSubmit({ children }: ComponentProps<'button'>) {
 
 export default function SignUpPage() {
   return (
-    <SignUp>
+    <SignUp
```
Comment (tmilewski) [inline packages/elements/examples/nextjs/app/sign-up/[[...sign-up]]/page.tsx:24]: Is there potential of the start step acting as the fallback?

Comment (BRKalow) [reply packages/elements/examples/nextjs/app/sign-up/[[...sign-up]]/page.tsx:24]: We could, in practice there are some edge cases though, like if we're landing on a callback route. So I think the current guidance would be to abstract start and re-use it as your fallback if this is the desired behavior.

packages/elements/src/react/sign-up/root.tsx
```diff
@@ -26,6 +28,11 @@ function SignUpFlowProvider({ children }: SignUpFlowProviderProps) {
   useEffect(() => {
     if (!clerk || !router) return;
 
+    // @ts-expect-error -- This is actually an IsomorphicClerk instance
+    clerk.addOnLoaded(() => {
+      ref.send({ type: 'SET_CLERK', clerk });
```
tmilewski, line 33: 👍

tmilewski (approved):

packages/elements/src/internals/machines/sign-in/machines/router.machine.ts
```diff
@@ -144,6 +144,11 @@ export const SignInRouterMachine = setup({
         },
       })),
     },
+    SET_CLERK: {
```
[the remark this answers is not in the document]
BRKalow: SGTM!

PR #3169, opened 2024-04-11
fix(clerk-js): Add payment method redirect fix

nikospapcom:
In this pr we are adding a fix also for useFetch hook. We are planning to add in main but after introducing some tests first

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/components/OrganizationProfile/OrganizationBillingPage.tsx
```diff
@@ -8,7 +8,7 @@ export const OrganizationBillingPage = withCardStateProvider(() => {
   const { organization } = useOrganization();
   const { data: availablePlans, isLoading: isLoadingAvailablePlans } = useFetch(
     organization?.getAvailablePlans,
-    `organization-available-plans-${organization?.id}`,
+    `organization-available-plans`,
```
nikospapcom, line 11: we don't need to pass `organization?.id` because available plans is the same for all orgs

packages/clerk-js/src/ui/hooks/useFetch.ts
```diff
@@ -77,6 +77,7 @@ export const useFetch = <K, T>(
   const cached = useSyncExternalStore(subscribeCache, getCache);
 
   useEffect(() => {
+    fetcherRef.current = fetcher; // Only update fetcher when params change
```
nikospapcom, line 80: Adding this after pairing with @panteliselef 
Next steps I'm planning to add some tests for `useFetch` in the main and then I will add this again

PR #3191, opened 2024-04-15
feat(clerk-js): Fallback to invisible CAPTCHA if the element is not found

anagstef:
## Description

This PR:

- Expects a `captcha_public_key_invisible` CAPTCHA key alongside with the main key, so it can fallback to `invisible` CAPTCHA if the `clerk-captcha` element is not found in the DOM.
- Utilizes the `unsupported-callback` prop of the CF Turnstile, to show error when the browser is not supported.
- Sends a `captcha_widget_type` prop in the `/sign_ups` POST request to FAPI to let it know what widget type was used.
- Removes the support to `always_visible` widget type option.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/types/src/displayConfig.ts
```diff
@@ -2,7 +2,7 @@ import type { DisplayThemeJSON } from './json';
 import type { ClerkResource } from './resource';
 
 export type PreferredSignInStrategy = 'password' | 'otp';
-export type CaptchaWidgetType = 'smart' | 'always_visible' | 'invisible' | null;
```
dimkl, line 5: ❓ Did we drop this option? If so, shouldn't we mention it in changelog that the `always_visible` is removed?

dimkl (approved):
❓ How did the `captchaPublicKeyInvisible` was populated up until now? Wasn;t the `invisible` mode the default up until now?
I am mostly asking because it seems that this is a fix, not a feat :/

packages/types/src/displayConfig.ts
dimkl: since it wasn't introduced publicly, you are correct, we shouldn't add it in the changelog. nice!

anagstef:
> ❓ How did the captchaPublicKeyInvisible was populated up until now? Wasn;t the invisible mode the default up until now?

@dimkl you are right, the default one until now was Invisible. This `captchaPublicKeyInvisible` new property is actually a fallback that will be used if the main captcha strategy cannot be used. I'll add a comment to explain this.

panteliselef:
I'm confused as to why we need 2 site keys at the same time

anagstef:
@panteliselef The `captchaPublicKeyInvisible` is a fallback to invisible, if the `clerk-captcha` element is not found in the DOM.

panteliselef (approved):

nikosdouvlis (approved):

PR #3204, opened 2024-04-16
feat(nextjs): Throw when component is mounted on non-catch-all route

nikosdouvlis:
## Description
When a component is using path-based routing, we need it to be mounted on an optional catch-all route page so the internal router can handle all navigations that affect the part of the URL after the path the component is mounted at (eg `/sign-in/` and after)

Pages router: 
We can simply try to match the `[[....*]]` pattern in the `pathname`

App router:
There is no easy way to do this. Instead, we fire a request to a random subpath (based on the path the component is mounted at). If we get a 404, it means that the current router is NOT a catch-all route 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


panteliselef (approved):
lgtm

octoper:
❓  @nikosdouvlis Should we also do this for our Remix SDK as it's also has path-based routing strategy by default?

nikosdouvlis:
> ❓ @nikosdouvlis Should we also do this for our Remix SDK as it's also has path-based routing strategy by default?

Great point - absolutely :)

PR #3214, opened 2024-04-17
fix(types,localizations): Added translation keys and Hu translations

ezkomoly:
## Description

I've found that after applying translations there are a bunch of english text left on profile, login pages etc.
Not sure if these were kept english on purpose, but I added them, so they can be translated.
I've also changed some of the translations to be better.

## Checklist

- [ X ] `npm test` runs as expected.
- [ X ] `npm run build` runs as expected.

## Type of change

- [ X ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/types/src/localization.ts
+++ b/packages/types/src/localization.ts
@@ -179,6 +179,7 @@ type _LocalizationResource = {
       subtitle_phone: LocalizationValue;
       formTitle: LocalizationValue;
       resendButton: LocalizationValue;
+      formSubtitle_email: LocalizationValue;
     };
     resetPassword: {
       title: LocalizationValue;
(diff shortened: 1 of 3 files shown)


packages/types/src/localization.ts
anagstef, line 1: All these new keys are not used anywhere. We cannot add them. :(

anagstef (changes requested):
Hello! Thank you for contributing. 🙏 
Unfortunately, there is no reason to add new localization keys that are not used anywhere. Can you explain the reasoning behind this decision? 
Maybe you are using the Core 1 (v4) SDKs?

packages/types/src/localization.ts
ezkomoly: After applying any of the translations there are still english text left on the components.
I've added the keys for these. 

### For example using my translation: 
Screenshot 2024-04-25 at 23 06 51

### Or the french translation: 
Screenshot 2024-04-25 at 23 08 57

### However using the translation I provided in this PR, with they new keys, these can be translated also: 
Screenshot 2024-04-25 at 23 10 07

ezkomoly: Okay, I've realized my mistake, I was using Core 1 instead of Core 2 where the new keys are pointless.

Sorry.

PR #3246, opened 2024-04-23
chore(backend): Use devBrowser from auth context instead of retrieving from clerkUrl

dimkl:
## Description

Minor refactor to re-use authContext.devBrowser instead of retrieving it manually from the query params of the url. With this change we will keep the retrieval of the dev browser token centralized (in the authContext).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


dimkl:
This change should not be applied as the condition should specifically check if the dev browser is in the URL to cover cases like AP redirects to application.

PR #3249, opened 2024-04-24
fix(clerk-js): Hide "Create Organization" on OrganizationList when disabled for user

dimkl:
## Before 
Screenshot 2024-04-24 at 13 19 21

## After
Screenshot 2024-04-24 at 13 18 57

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/clean-boats-invite.md
+++ b/.changeset/clean-boats-invite.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Hide "Create Organization" in <OrganizationList/> when it's disabled for user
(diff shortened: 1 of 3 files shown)


anagstef (approved):

panteliselef:
I don't wanna request changes as this can be patched later, but if the user does not have any memberships, invitations or suggestions we fallback to create organization. We need to handle that case as well.

dimkl:
@panteliselef what is the expected behaviour in that case if the user cannot create an organization?

panteliselef:
@dimkl I forgot to mention that the scenario i described would occur when the component has the `hidePersonal` prop

PR #3257, opened 2024-04-24
chore(clerk-react): Send telemetry events for React hooks

LauraBeatris:
## Description

Resolves SDK-1672

Sends telemetry events for React hooks. This will be useful to track regarding the creation of custom flows with hooks such as `useSignUp`. 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/react/src/hooks/useAuth.ts
+++ b/packages/react/src/hooks/useAuth.ts
@@ -1,3 +1,4 @@
+import { eventMethodCalled } from '@clerk/shared/telemetry';
 import type {
   ActJWTClaim,
   CheckAuthorizationWithCustomPermissions,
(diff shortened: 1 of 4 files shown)


packages/react/src/hooks/useAuth.ts
```diff
@@ -141,6 +142,10 @@ export const useAuth: UseAuth = () => {
     [orgId, orgRole, userId, orgPermissions],
   );
 
+  useEffect(() => {
```
LauraBeatris, line 145: We need to be careful that those events will be emitted on the React tree render. Especially during local development, this could be dispatched in a huge volume due to hot reloads.

I'm considering passing a sampling rate to decrease the fraction of all events that should be recorded. We already do so for `COMPONENT_MOUNTED`.

LauraBeatris:
I don't think we should emit events for all hooks. We need to think about what particular insights each hook would give us, and if it's useful. 

The reason I advocate for sending events from `useSignIn` and `useSignUp`, is cause those give us insights of custom flows, eg: How this would react when Clerk Elements get released.

LauraBeatris:
Blocking this PR as I'm still deciding on how to move forward with recordings events in code hot-paths.

LauraBeatris:
Also closing this one in order to open later directly from the repo instead of a forked branch.

PR #3267, opened 2024-04-25
feat(shared): Introduce client-side caching for telemetry events

LauraBeatris:
## Description

Currently blocking SDK-1672 (https://github.com/clerk/javascript/pull/3257)

Introduces client-side caching for telemetry events. This prevents event flooding in frequently executed code paths, such as for React hooks or components.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/src/__tests__/telemetry.test.ts
+++ b/packages/shared/src/__tests__/telemetry.test.ts
@@ -8,6 +8,10 @@ jest.useFakeTimers();
 const TEST_PK = 'pk_test_Zm9vLWJhci0xMy5jbGVyay5hY2NvdW50cy5kZXYk';
 
 describe('TelemetryCollector', () => {
+  beforeEach(() => {
+    localStorage.clear();
+  });
+
   test('does nothing when disabled', async () => {
     const fetchSpy = jest.spyOn(global, 'fetch');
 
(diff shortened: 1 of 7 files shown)


packages/shared/src/telemetry/events/method-called.ts
```diff
@@ -1,6 +1,8 @@
+import { TelemetryClientCache } from '../clientCache';
 import type { TelemetryEventRaw } from '../types';
 
 const EVENT_METHOD_CALLED = 'METHOD_CALLED' as const;
+const EVENT_SAMPLING_RATE = 0.1;
```
LauraBeatris, line 5: Also introducing a sampling rate lower than the default for this event, to handle the server-side scenarios.

packages/shared/src/__tests__/telemetry.test.ts
```diff
@@ -170,4 +171,67 @@ describe('TelemetryCollector', () => {
...
+      });
+
+      jest.runAllTimers();
+
+      expect(fetchSpy).toHaveBeenCalled();
+
+      fetchSpy.mockRestore();
+    });
+
+    test('does not send event when it is in the cache', () => {
```
LauraBeatris, line 204: Still need to write tests for the following cases:
- When storage isn't available, fallbacks to the event-specific sampling rate
- When there's a `QuotaExceededError`, fallbacks to the event-specific sampling rate

packages/shared/src/telemetry/types.ts
```diff
@@ -71,5 +71,6 @@ export type TelemetryEvent = {
 export type TelemetryEventRaw<Payload = TelemetryEvent['payload']> = {
   event: TelemetryEvent['event'];
   eventSamplingRate?: number;
+  clientCacheKey?: `${TelemetryEvent['event']}:${string}`;
```
LauraBeatris, line 74: I'm not 100% decided if it's better to apply it as a required property. Depending on its nature, there's the likelihood of an event being called on the client side and server side.

`clerkMiddleware` gets only called on the server side, using an event-specific sampling rate. However, a React hook could be called on a client component.

panteliselef:
This PR also enables #3279 (adding telemetry to clerk-elements)

packages/shared/src/telemetry/types.ts
LauraBeatris: Alright, changing the approach here - we're going to generate that cache key based on the event payload, instead of having to pass it as a property.

packages/shared/src/telemetry/clientCache.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (!cacheEntry) {
+      localStorage.setItem(
+        this.#storageKey,
+        JSON.stringify({
+          [key]: now,
+        }),
+      );
+    }
+
+    const hasExpired = cacheEntry && now - cacheEntry > this.#cacheTtl;
```
LauraBeatris, line 29: Need to add a test for this scenario.

LauraBeatris:
Here's is how the cache is going to look like from the local storage - a map between unit event keys and their TTL in milliseconds: 

CleanShot 2024-04-29 at 13 02 13

LauraBeatris:
Re-opening it on https://github.com/clerk/javascript/pull/3287, this branch is from a forked repo so I think this is the cause of not allowing to run integration tests.

PR #3274, opened 2024-04-26
feat(backend,clerk-js,shared): Introduce suffixed / un-suffixed cookies 

dimkl:
## Description

Introduce suffixed / un-suffixed cookies to support multiple apps in the same domain.

## TODOs

- [ ] add integration test
- [ ] add tests 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/tokens/authenticateContext.ts
+++ b/packages/backend/src/tokens/authenticateContext.ts
@@ -19,6 +19,7 @@ interface AuthenticateContextInterface extends AuthenticateRequestOptions {
   // cookie-based values
   sessionTokenInCookie: string | undefined;
   clientUat: number;
+  suffixedCookies: boolean;
   // handshake-related values
   devBrowserToken: string | undefined;
   handshakeToken: string | undefined;
@@ -45,12 +46,17 @@ class AuthenticateContext {
     return this.sessionTokenInCookie || this.sessionTokenInHeader;
   }
 
+  private get cookieSuffix() {
+    return this.publishableKey?.split('_').pop();
+  }
+
   public constructor(private clerkRequest: ClerkRequest, options: AuthenticateRequestOptions) {
     // Even though the options are assigned to this later in this function
     // we set the publishableKey here because it is being used in cookies/headers/handshake-values
     // as part of getMultipleAppsCookie
     this.initPublishableKeyValues(options);
     this.initHeaderValues();
+    // initCookieValues should be used before initHandshakeValues because the it depends on suffixedCookies
     this.initCookieValues();
     this.initHandshakeValues();
     Object.assign(this, options);
@@ -70,15 +76,6 @@ class AuthenticateContext {
     this.frontendApi = pk.frontendApi;
   }
 
-  private initHandshakeValues() {
-    this.devBrowserToken =
-      this.getQueryParam(constants.QueryParameters.DevBrowser) ||
-      this.getMultipleAppsCookie(constants.Cookies.DevBrowser);
-    // Using getCookie since we don't suffix the handshake token cookie
-    this.handshakeToken =
-      this.getQueryParam(constants.QueryParameters.Handshake) || this.getCookie(constants.Cookies.Handshake);
-  }
-
   private initHeaderValues() {
     this.sessionTokenInHeader = this.stripAuthorizationHeader(this.getHeader(constants.Headers.Authorization));
     this.origin = this.getHeader(constants.Headers.Origin);
@@ -93,8 +90,19 @@ class AuthenticateContext {
   }
 
   private initCookieValues() {
-    this.sessionTokenInCookie = this.getMultipleAppsCookie(constants.Cookies.Session);
-    this.clientUat = Number.parseInt(this.getMultipleAppsCookie(constants.Cookies.ClientUat) || '') || 0;
+    // suffixedCookies needs to be set first because it's used in getMultipleAppsCookie
+    this.suffixedCookies = this.getSuffixedCookie(constants.Cookies.SuffixedCookies) === 'true';
+    this.sessionTokenInCookie = this.getSuffixedOrUnSuffixedCookie(constants.Cookies.Session);
+    this.clientUat = Number.parseInt(this.getSuffixedOrUnSuffixedCookie(constants.Cookies.ClientUat) || '') || 0;
+  }
+
+  private initHandshakeValues() {
+    this.devBrowserToken =
+      this.getQueryParam(constants.QueryParameters.DevBrowser) ||
+      this.getSuffixedOrUnSuffixedCookie(constants.Cookies.DevBrowser);
+    // Using getCookie since we don't suffix the handshake token cookie
+    this.handshakeToken =
+      this.getQueryParam(constants.QueryParameters.Handshake) || this.getCookie(constants.Cookies.Handshake);
   }
 
   private stripAuthorizationHeader(authValue: string | undefined | null): string | undefined {
@@ -113,9 +121,15 @@ class AuthenticateContext {
     return this.clerkRequest.cookies.get(name) || undefined;
   }
 
-  private getMultipleAppsCookie(cookieName: string) {
-    const suffix = this.publishableKey?.split('_').pop();
-    return this.getCookie(`${cookieName}_${suffix}`) || this.getCookie(cookieName) || undefined;
+  private getSuffixedCookie(name: string) {
+    return this.getCookie(`${name}_${this.cookieSuffix}`) || undefined;
+  }
+
+  private getSuffixedOrUnSuffixedCookie(cookieName: string) {
+    if (this.suffixedCookies) {
+      return this.getSuffixedCookie(cookieName);
+    }
+    return this.getCookie(cookieName);
   }
 }
 
--- a/packages/backend/src/tokens/authenticateContext.ts
+++ b/packages/backend/src/tokens/authenticateContext.ts
@@ -1,4 +1,7 @@
+import { parsePublishableKey } from '@clerk/shared/keys';
+
 import { constants } from '../constants';
+import { assertValidPublishableKey } from '../util/optionsAssertions';
 import type { ClerkRequest } from './clerkRequest';
 import type { AuthenticateRequestOptions } from './types';
 
@@ -23,6 +26,10 @@ interface AuthenticateContextInterface extends AuthenticateRequestOptions {
   clerkUrl: URL;
   // cookie or header session token
   sessionToken: string | undefined;
+  // enforce existence of the following props
+  publishableKey: string;
+  instanceType: string;
+  frontendApi: string;
 }
 
 interface AuthenticateContext extends AuthenticateContextInterface {}
@@ -39,6 +46,10 @@ class AuthenticateContext {
   }
 
   public constructor(private clerkRequest: ClerkRequest, options: AuthenticateRequestOptions) {
+    // Even though the options are assigned to this later in this function
+    // we set the publishableKey here because it is being used in cookies/headers/handshake-values
+    // as part of getMultipleAppsCookie
+    this.initPublishableKeyValues(options);
     this.initHeaderValues();
     this.initCookieValues();
     this.initHandshakeValues();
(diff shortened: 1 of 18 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -162,8 +158,8 @@ export class Clerk implements ClerkInterface {
   #authService: SessionCookieService | null = null;
   #broadcastChannel: LocalStorageBroadcastChannel<ClerkCoreBroadcastChannelEvent> | null = null;
   #componentControls?: ReturnType<MountComponentRenderer> | null;
-  #devBrowser: DevBrowser | null = null;
-  #environment?: EnvironmentResource | null;
+  // converted to protected environment to support `updateEnvironment` type assertion
+  protected environment?: EnvironmentResource | null;
```
dimkl, line 162: ℹ️ @nikosdouvlis ☝️

packages/clerk-js/src/core/clerk.ts
```diff
@@ -741,7 +738,13 @@ export class Clerk implements ClerkInterface {
...
     }
 
-    await this.#authService?.setAuthCookiesFromSession(newSession);
+    // Sync __session and __client_uat to cookies using events.TokenUpdate dispatched event
+    // only for newSession is null since the getToken will not be executed. Since getToken
+    // triggers internally a events.TokenUpdate there is no need to trigger it when the newSession exists.
+    const token = await newSession?.getToken();
+    if (!token) {
+      eventBus.dispatch(events.TokenUpdate, { token: null });
+    }
```
dimkl, line 747: ℹ️ @nikosdouvlis ☝️

packages/backend/src/tokens/authenticateContext.ts
```diff
@@ -39,44 +47,76 @@ class AuthenticateContext {
...
+  private getHeader(name: string) {
+    return this.clerkRequest.headers.get(name) || undefined;
+  }
+
+  private getCookieWithVariants(cookieName: string) {
+    const pipeline = new Pipeline<string | undefined, Map<string, string>>()
+      .addOperation(new ReadSuffixedCookieOperation(cookieName, this.publishableKey))
+      .addOperation(new ReadCookieOperation(cookieName));
+    // get first result
+    return pipeline.invoke(this.clerkRequest.cookies).find(Boolean);
```
BRKalow, line 118: 🤔 Instead of introducing this `Pipeline` primitive, can we create some sort of cookies abstraction that handles dealing with the suffix?

This isn't a pattern we have elsewhere and I'm finding myself having to jump to the Pipeline definition to understand what's happening here. If we want to keep this abstraction, some comments might help explaining how it works.

packages/shared/src/pipelines.ts
```diff
@@ -0,0 +1,95 @@
...
+  set(data: D, options: O): void {
+    this.operations.map(operation => operation.set(data, options));
+  }
+
+  remove(): void {
+    this.operations.map(operation => operation.remove());
+  }
+}
+
+const getCookieSuffixFromPk = (publishableKey: string): string => {
```
BRKalow, line 45: I wouldn't expect these cookie-specific methods to live in this file, just the pipeline primitives. I see the IOOperations are defined in shared/cookies, maybe that would be a better place for these other operations as well.

packages/shared/src/pipelines.ts
```diff
@@ -0,0 +1,95 @@
+export interface Operation<T, D = unknown> {
+  invoke(data: D): T;
+}
+
+export class Pipeline<T, D> implements Operation<T[]> {
```
BRKalow, line 5: If we're going to introduce these new functional primitives, can we make sure they're clearly documented?

packages/backend/src/tokens/request.ts
```diff
@@ -66,6 +66,10 @@ export async function authenticateRequest(
   const authenticateContext = createAuthenticateContext(createClerkRequest(request), options);
   assertValidSecretKey(authenticateContext.secretKey);
 
+  const cookiePipeline = new Pipeline<string, string>()
+    .addOperation(new SuffixCookieOperation(authenticateContext.publishableKey))
+    .addOperation(new UnSuffixCookieOperation(authenticateContext.publishableKey));
+
```
nikosdouvlis, line 72: Why do we define the pipeline here instead of closer to where we call it? Is it used by something in between?

packages/backend/src/tokens/request.ts
```diff
@@ -109,13 +115,15 @@ export async function authenticateRequest(
...
-      headers.append('Set-Cookie', x);
-      if (x.startsWith(`${constants.Cookies.Session}=`)) {
-        sessionToken = x.split(';')[0].substring(10);
-      }
+      cookiePipeline.invoke(x).forEach(xVariants => {
+        if (xVariants.startsWith(`${constants.Cookies.Session}=`)) {
+          sessionToken = x.split(';')[0].split('=')[1];
+        }
+        headers.append('Set-Cookie', xVariants);
+      });
```
nikosdouvlis, line 123: What would be the alternative way to write this without pipelines?

packages/shared/src/pipelines.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  addOperation(operation: Operation<T>) {
+    this.operations.push(operation);
+    return this;
+  }
+
+  invoke(data: D): T[] {
+    return this.operations.map(operation => operation.invoke(data));
+  }
+}
```
nikosdouvlis, line 16: I'm trying to understand the pattern used here but I do need to do some more reading. This, however, looks like we're simply chaining methods while keeping a reference to their return value. Is this simply it?

packages/shared/src/pipelines.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export interface IOOperation<T, D = unknown, O = unknown> {
+  get(): T;
+  set(data: D, options: O): void;
+  remove(): void;
+}
```
nikosdouvlis, line 22: (This looks like the missing interface I was talking about in #3362)

This feels a little too complicated to me because it looks like we're defining 4 interfaces (and some duplicates across this PR and #3362 ) to simply invoke 2 methods. However, I want to think more about it as I'm not familiar with the pattern.

Could we achieve the same results by defining a method/helper with the following signature?

```
function writeCookie(handlers: CookieHandler[]) {
  handlers.forEach(h => h.write(...))
}
```

or similar? Would that be simpler?

packages/shared/src/pipelines.ts
dimkl: The Pipeline pattern is a mix of 2 more commonly known patterns:
- [Chain of Responsibility](https://refactoring.guru/design-patterns/chain-of-responsibility)
- [Command](https://refactoring.guru/design-patterns/command)

>This, however, looks like we're simply chaining methods while keeping a reference to their return value.
(kind of) yes.

Article about Pipeline pattern: https://medium.com/@bonnotguillaume/software-architecture-the-pipeline-design-pattern-from-zero-to-hero-b5c43d8a4e60

packages/shared/src/pipelines.ts
dimkl: This PR handles both generating multiple values based on a single input (case of backend SDK that uses a single cookie directive and produces the suffixed and un-suffixed variants) and performing multiple actions triggering a single pipeline method (eg write both suffixed and un-suffixed cookies using a single cookie key and value).
That's why there are multiple interfaces that look similar.
To achieve the same using method / helpers we should:
- write 3 cookie handlers (eg session, clientUat, devBrowser) for ClerkJS
- write 3 functions to generate the cookie key variants (eg session, clientUat, devBrowser)  for backend SDK

For me this seems like the same amount of class or methods but different implementation. Using a pattern (even though it would require some time to learn it) would make the handling of all those cases consistent and expected instead of trying to read and verify what it's separate handler or function does.
If you think that it's more complex than it should, let's drop the pattern usage and simplify things. @nikosdouvlis Feel free to provide any suggestion on how we could make things simpler?

packages/shared/src/pipelines.ts
dimkl: i agree, it should be moved to a more suited place. This is a draft PR, If we agree on the implementation, i will move them to a more appropriate location.

packages/shared/src/pipelines.ts
dimkl: If we are going to introduce them i will add JSDoc with examples for each primitive :)

dimkl:
Would you consider this example of implementing the same pattern more readable and preferrable?

```typescript
//
// @clerk/shared
//
type Operation<T> = (...args: any[]) => T;
const pipeline = <T>(operations: Operation<T>[], ...args: any[]) =>
  operations.map((o) => o(...args));

//
// @clerk/backend
//
const getCookieSuffixFromPk = (publishableKey: string): string => {
  return publishableKey.split("_").pop() || "";
};

const unSuffixCookie = (publishableKey: string) => {
  return (cookieDirective: string) => {
    //...
  };
};

const suffixCookie = (publishableKey: string) => {
  return (cookieDirective: string) => {
    //...
  };
};

// Example of executing the pipeline

const cookieDirectives = ["__session=aloha"];
cookieDirectives.map((cookieDirective) =>
  pipeline(
    [suffixCookie("pk_test_"), unSuffixCookie("pk_test_")],
    cookieDirective
  )
);

//
// @clerk/clerk-js
//

import { createCookieHandler } from "@clerk/shared/cookie";

const getCookie = () => {
  return (cookieName: string) => {
    //...
  };
};

const getSuffixedCookie = (publishableKey: string) => {
  return (cookieName: string) => {
    //...
  };
};

// Operations

const setCookie = () => {
  return (cookieName: string, value: string) => {
    //...
  };
};

const setSuffixedCookie = (publishableKey: string) => {
  return (cookieName: string, value: string) => {
    //...
  };
};

const removeCookie = () => {
  return (cookieName: string, value: string) => {
    //...
  };
};

const removeSuffixedCookie = (publishableKey: string) => {
  return (cookieName: string, value: string) => {
    //...
  };
};

// pipelines

const getPipeline = (cookieName: string) =>
  pipeline([getCookie(), getSuffixedCookie("pk_test_")], cookieName);
const setPipeline = (cookieName: string, value: string) =>
  pipeline([setCookie(), setSuffixedCookie("pk_test_")], cookieName, value);
const removePipeline = (cookieName: string) =>
  pipeline([removeCookie(), removeSuffixedCookie("pk_test_")], cookieName);

// Example of executing the pipeline

getPipeline("__session");
setPipeline("__session", "value");
removePipeline("__session");
```

cc: @BRKalow , @nikosdouvlis

packages/backend/src/tokens/request.ts
BRKalow, line 123: I wish there was a way to return the session token in the handshake payload that was more directly accessible, e.g. a separate key `handshakePayload.sessionToken`, but that would bloat the payload size 😢

packages/backend/src/tokens/request.ts
```diff
@@ -86,12 +90,14 @@ export async function authenticateRequest(
...
-    const frontendApiNoProtocol = pk.frontendApi.replace(/http(s)?:\/\//, '');
+    const frontendApiNoProtocol = authenticateContext.frontendApi.replace(/http(s)?:\/\//, '');
 
     const url = new URL(`https://${frontendApiNoProtocol}/v1/client/handshake`);
     url.searchParams.append('redirect_url', redirectUrl?.href || '');
+    // Request ONLY suffixed cookies from FAPI handshake
+    url.searchParams.append('__clerk_suffixed_cookies', 'true');
 
-    if (pk?.instanceType === 'development' && authenticateContext.devBrowserToken) {
+    if (authenticateContext?.instanceType === 'development' && authenticateContext.devBrowserToken) {
```
BRKalow, line 100: good thinking moving this to the context 👍

packages/shared/src/pipelines.ts
dimkl: i have moved any code related to cookie suffixes to the `@clerk/shared/keys`

dimkl:
Closing this in favor of another implementation : https://github.com/clerk/javascript/pull/3506

PR #3279, opened 2024-04-29
feat(elements): Add telemetry

LekoArts:
## Description

Adds telemetry tracking to better understand what people build with Elements. Initially we'll be able to see:

- If a user creates a sign-in and/or sign-up flow
- Which steps they create
- Which inputs they use inside their steps

Fixes SDK-1658

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/react/common/form/index.tsx
```diff
@@ -528,6 +530,10 @@ type FormInputProps =
  */
 const Input = React.forwardRef<React.ElementRef<typeof RadixControl>, FormInputProps>(
   (props: FormInputProps, forwardedRef) => {
+    const telemetry = useTelemetry();
+
+    telemetry?.record(eventComponentMounted('Elements_Input', props));
+
```
panteliselef, line 536: ❓
1. Wouldn't this fire on every re-render ?
2. Seeing the implementation of `eventComponentMounted` the props you seem to be interested will be ignored.

LekoArts: 1. Yes. https://github.com/clerk/javascript/pull/3267 is fixing this
2. Thanks! I fixed it by introducing a new helper function

packages/shared/src/telemetry/events/component-mounted.ts
```diff
@@ -11,8 +11,19 @@ type EventComponentMounted = {
...
 /**
  * Fired when one of the Clerk components is mounted.
+ *
+ * @param component - The name of the component.
+ * @param props - The props passed to the component. Will be filtered to a known list of props.
+ *
+ * @example
+ * telemetry.record(eventComponentMounted('SignUp', props));
  */
 export function eventComponentMounted(
```
BRKalow, line 28: can we change this one to `eventPrebuiltComponentMounted` (or something similar) and expose your currently named `eventComponentMountedRaw` as `eventComponentMounted`?

LekoArts: Done 👍

packages/elements/src/react/common/form/index.tsx
LekoArts: They will probably be released around the same time so it's not blocking

BRKalow (approved):
⭐

tmilewski (approved):

PR #3287, opened 2024-04-29
feat(shared): Introduce client-side caching for telemetry events

LauraBeatris:
## Description

Originally raised (https://github.com/clerk/javascript/pull/3267/files) but had to re-open here due GitHub permissions issues on running integrations from a forked branch.

Currently blocking SDK-1672 (https://github.com/clerk/javascript/pull/3257) and https://github.com/clerk/javascript/pull/3279

Introduces client-side caching for telemetry events. This prevents event flooding in frequently executed code paths, such as for React hooks or components.

Event sampling by rate will be executed on the server-side, or when neither the browser storage or window are available. 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/metal-bulldogs-destroy.md
+++ b/.changeset/metal-bulldogs-destroy.md
@@ -1,0 +1,5 @@
+---
+'@clerk/shared': patch
+---
+
+Introduce client-side caching to TelemetryCollector
--- a/.changeset/metal-bulldogs-destroy.md
+++ b/.changeset/metal-bulldogs-destroy.md
@@ -2,4 +2,4 @@
 '@clerk/shared': patch
 ---
 
-Introduce client-side caching to TelemetryCollector
+Internal change to add client-side caching to Clerk's telemetry. This prevents event flooding in frequently executed code paths, such as for React hooks or components. Gracefully falls back to the old behavior if e.g. `localStorage` is not available. As such, no public API changes and user changes are required.
--- a/packages/shared/src/__tests__/telemetry.test.ts
+++ b/packages/shared/src/__tests__/telemetry.test.ts
@@ -1,6 +1,9 @@
 import 'cross-fetch/polyfill';
 
+import assert from 'assert';
+
 import { TelemetryCollector } from '../telemetry';
+import type { TelemetryEvent } from '../telemetry/types';
 
 jest.useFakeTimers();
 
@@ -220,6 +223,11 @@ describe('TelemetryCollector', () => {
         payload,
       });
 
+      collector.record({
+        event,
+        payload,
+      });
+
       jest.runAllTimers();
 
       expect(fetchSpy).toHaveBeenCalledTimes(2);
@@ -275,6 +283,37 @@ describe('TelemetryCollector', () => {
       expect(fetchSpy).not.toHaveBeenCalled();
 
       randomSpy.mockRestore;
+    });
+
+    test('generates unique key without credentials based on event payload', () => {
+      const collector = new TelemetryCollector({
+        publishableKey: TEST_PK,
+      });
+
+      const event: TelemetryEvent = {
+        sk: '123',
+        pk: TEST_PK,
+        it: 'development',
+        event: 'TEST_EVENT',
+        cv: '0.1',
+        sdkv: '0.1',
+        payload: {
+          foo: true,
+        },
+      };
+
+      collector.record(event);
+      collector.record(event);
+
+      jest.runAllTimers();
+
+      const item = localStorage.getItem('clerk_telemetry_throttler');
+      assert(item);
+      const expectedKey = '["","TEST_EVENT",true,"development",null,null]';
+
+      expect(JSON.parse(item)[expectedKey]).toEqual(expect.any(Number));
+
+      fetchSpy.mockRestore();
     });
   });
 });
--- a/packages/shared/src/__tests__/telemetry.test.ts
+++ b/packages/shared/src/__tests__/telemetry.test.ts
@@ -212,6 +212,42 @@ describe('TelemetryCollector', () => {
       fetchSpy.mockRestore();
     });
 
+    test('sends event when it is in the cache but has expired', () => {
+      const fetchSpy = jest.spyOn(global, 'fetch');
+      const originalDateNow = Date.now;
+      const cacheTtl = 86400000;
+
+      let now = originalDateNow();
+      jest.spyOn(Date, 'now').mockImplementation(() => now);
+
+      const collector = new TelemetryCollector({
+        publishableKey: TEST_PK,
+        maxBufferSize: 1,
+      });
+
+      const event = 'TEST_EVENT';
+      const payload = {
+        foo: true,
+      };
+
+      collector.record({
+        event,
+        payload,
+      });
+
+      // Move time forward beyond the cache TTL
+      now += cacheTtl + 1;
+
+      collector.record({
+        event,
+        payload,
+      });
+
+      expect(fetchSpy).toHaveBeenCalledTimes(2);
+
+      fetchSpy.mockRestore();
+    });
+
     test('does not send event when it is in the cache', () => {
       const fetchSpy = jest.spyOn(global, 'fetch');
 
--- a/packages/shared/src/__tests__/telemetry.test.ts
+++ b/packages/shared/src/__tests__/telemetry.test.ts
@@ -8,10 +8,6 @@ jest.useFakeTimers();
 const TEST_PK = 'pk_test_Zm9vLWJhci0xMy5jbGVyay5hY2NvdW50cy5kZXYk';
 
 describe('TelemetryCollector', () => {
-  beforeEach(() => {
-    localStorage.clear();
-  });
-
   test('does nothing when disabled', async () => {
     const fetchSpy = jest.spyOn(global, 'fetch');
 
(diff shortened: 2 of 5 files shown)


packages/shared/src/telemetry/clientCache.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (!entry) {
+      const updatedCache = {
+        ...this.#cache,
+        [key]: now,
+      };
+
+      localStorage.setItem(this.#storageKey, JSON.stringify(updatedCache));
+    }
+
+    const hasExpired = entry && now - entry > this.#cacheTtl;
```
LauraBeatris, line 29: Realized this logic is incorrect. When an entry expires, we should update `clerk_telemetry` cache object and remove that key from it. 

When returning here, an event is cached when it isn't expired & it's entry exists.

packages/shared/src/telemetry/clientCache.ts
```diff
@@ -0,0 +1,97 @@
+import type { TelemetryEventRaw } from './types';
+
+type TtlInMilliseconds = number;
+
+const DEFAULT_CACHE_TTL_MS = 1000; // 24 hours
```
BRKalow, line 5: Is this value just for testing? Or is the comment wrong? 👀

packages/shared/src/telemetry/clientCache.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Manages caching for telemetry events using the browser's localStorage to
+ * mitigate event flooding in frequently executed code paths.
+ */
+export class TelemetryClientCache {
```
BRKalow, line 11: nit: Do you think like `TelemetryEventThrottler`, or something similar, would be a better name? I feel like that more clearly communicates the intent 💭

packages/shared/src/telemetry/clientCache.ts
```diff
@@ -0,0 +1,97 @@
...
+  #storageKey = 'clerk_telemetry';
+  #cacheTtl = DEFAULT_CACHE_TTL_MS;
+
+  cacheAndRetrieve(event: TelemetryEventRaw): boolean {
```
BRKalow, line 15: if we change the naming this could be: `isEventThrottled`

packages/shared/src/telemetry/collector.ts
```diff
@@ -110,7 +114,7 @@ export class TelemetryCollector {
 
     this.#logEvent(preparedPayload.event, preparedPayload);
 
-    if (!this.#shouldRecord(event.eventSamplingRate)) {
+    if (!this.#shouldRecord(event)) {
```
BRKalow, line 117: We might want to pass `preparedPayload` here so we can include it as part of the throttled key, as if any of the "global" metadata that is provided in that method changes the event would still get throttled.

packages/shared/src/__tests__/telemetry.test.ts
```diff
@@ -145,29 +146,168 @@ describe('TelemetryCollector', () => {
...
+      });
+
+      jest.runAllTimers();
+
+      expect(fetchSpy).toHaveBeenCalled();
+
+      fetchSpy.mockRestore();
+    });
+
+    test('sends event when it is in the cache but has expired', () => {
```
BRKalow, line 215: Nice tests 🥇

BRKalow (commented):
Looking good! Left a comment about the naming of the cache class. I wonder if calling it a "throttler" (or something) is more accurate, as opposed to a cache.

packages/shared/src/telemetry/collector.ts
```diff
@@ -119,13 +123,23 @@ export class TelemetryCollector {
...
-    const shouldBeSampled =
-      randomSeed <= this.#config.samplingRate &&
-      (typeof eventSamplingRate === 'undefined' || randomSeed <= eventSamplingRate);
+    const clientCache = this.#clientCache;
 
-    return this.isEnabled && !this.isDebug && shouldBeSampled;
+    if (clientCache.isStorageSupported) {
+      const isCached = clientCache.cacheAndRetrieve(event);
+      return !isCached;
+    }
```
BRKalow, line 137: Can we include the `isStorageSupported` check in `cacheAndRetrieve()`? That would simplify usage here

packages/shared/src/telemetry/clientCache.ts
```diff
@@ -0,0 +1,97 @@
...
+   * If any of these operations fail, it indicates that localStorage is either
+   * not supported or not writable (e.g., in cases where the storage is full or
+   * the browser is in a privacy mode that restricts localStorage usage).
+   */
+  get isStorageSupported(): boolean {
+    if (typeof window === 'undefined') {
+      return false;
+    }
+
+    const storage = window['localStorage'];
```
BRKalow, line 72: ❓ why did we do property access with brackets instead of `window.localStorage`?

packages/shared/src/telemetry/clientCache.ts
LauraBeatris: Ooops... I was testing it and ended pushing the wrong value.

packages/shared/src/telemetry/clientCache.ts
LauraBeatris: Yeah, it makes sense to me! `ClientCache` is indeed a bit generic, it leads to the question: "caching for what"?

packages/shared/src/telemetry/collector.ts
LauraBeatris: Yeah, it makes sense! Better rely on the `preparedPayload` than the raw one.

.changeset/metal-bulldogs-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/shared': patch
+---
+
+Introduce client-side caching to TelemetryCollector
```
LekoArts, line 5:
```suggestion
Internal change to add client-side caching to Clerk's telemetry. This prevents event flooding in frequently executed code paths, such as for React hooks or components. Gracefully falls back to old behavior if e.g. `localStorage` is not available. As such, no public API changes and user changes required.
```

You can follow [these guidelines](https://www.notion.so/clerkdev/Code-collaboration-4d32380fb78c4226abd9e1bcb7d55abf?pvs=4#62d500ed4eea46c0b8732f8fe4db6056) to write more thorough changesets. I'm also not always following them 1:1 but they are good general guardrails.

packages/shared/src/telemetry/clientCache.ts
panteliselef: Yeap this name feels more appropriate, my understanding here is that if a key needs to be invalidated we just delete and expect the next attempt to cache it again

packages/shared/src/__tests__/telemetry.test.ts
```diff
@@ -145,29 +146,168 @@ describe('TelemetryCollector', () => {
...
+    afterEach(() => {
+      windowSpy.mockRestore();
+    });
 
-    jest.runAllTimers();
+    test('does not send events if the random seed does not exceed the event-specific sampling rate', async () => {
+      windowSpy.mockImplementation(() => undefined);
 
-    expect(fetchSpy).not.toHaveBeenCalled();
+      const fetchSpy = jest.spyOn(global, 'fetch');
```
LekoArts, line 171: Nit: You could also move the `fetchSpy` init to a `let` variable and create the mock in `beforeEach` like with `window.get`

panteliselef (approved):
This looks good to me, I'm approving in order not to block this PR further, suggestions from @BRKalow probably should be addressed before merging. I've added a few questions myself. 

Amazing stuff

LekoArts (commented):
Great work on this!

packages/shared/src/telemetry/clientCache.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    return !!entry;
+  }
+
+  #generateKey({ event, payload }: TelemetryEventRaw): string {
+    const payloadUniqueKey = JSON.stringify(
+      Object.keys(payload)
+        .sort()
+        .map(key => payload[key]),
+    );
```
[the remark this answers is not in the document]
LekoArts: By sorting it we can ensure that payloads with the same content (but with different ordering) are still cached in the same key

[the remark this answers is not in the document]
LauraBeatris: Will add a JSDocs on this method to clarify about that!

packages/shared/src/__tests__/telemetry.test.ts
LauraBeatris: I've moved some spies to the root `describe` block!

packages/shared/src/telemetry/throttler.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    return !!entry;
+  }
+
+  /**
+   * Generates a consistent unique key for telemetry events by sorting payload properties.
+   * This ensures that payloads with identical content in different orders produce the same key.
+   */
+  #generateKey({ event, payload }: TelemetryEventRaw): string {
```
LauraBeatris, line 48: Just need to refactor this according to @BRKalow feedback, we should use the `preparedPayload` rather than the raw one.

BRKalow (approved):
⭐

PR #3291, opened 2024-04-30
fix(clerk-js): Show email filed when trying to edit email after trying to sign up

octoper:
## Description

Right now there is an issue that when the app has email required but the user tries to sign up with a method that does not provides us an email like Metamask, they will be prompted to add the email, they can add the email the first time but if the try to edit the email, now the form does not show the email field. This PR fixes this by checking if the previous email has been already verified, because if the user has click the verification link or entered the verification code with the first inserted email they would be logged in right away.

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/components/SignUp/__tests__/SignUpContinue.test.tsx
```diff
@@ -43,10 +43,27 @@ describe('SignUpContinue', () => {
...
+      f.startSignUpWithEmailAddress({
+        emailVerificationStatus: 'verified',
+      });
+    });
+    render(<SignUpContinue />, { wrapper });
+    // This will have to be show if the email is already set but it is not verified
+    // because if the users tries to edit the email before the first one has verified
+    // the email field will be lost
     expect(screen.queryByText(/email address/i)).not.toBeInTheDocument();
-    screen.getByText(/password/i);
```
panteliselef, line 49: Was this failing ? Seems like the new statement check for the same thing

panteliselef (approved):

packages/clerk-js/src/ui/components/SignUp/__tests__/SignUpContinue.test.tsx
octoper: No it was not failing I just wanted to align that with the above line

desiprisg (approved):

PR #3292, opened 2024-04-30
fix(elements): Only call useSearchParams() on the client

BRKalow:
## Description

Only call `useSearchParams()` on the client to ensure that we don't de-opt pages into client-side rendering. This shouldn't be an issue in practice as we only use the router data on the client currently.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


tmilewski (approved):
:shipit:

PR #3299, opened 2024-05-01
[sdk-node] Export authenticateRequest function

greguintow:
## Description

- [sdk-node] Export `authenticateRequest` function

**Context**
- Currently using a nitro server which isn't compatible with the clerk express middleware when is not authenticated. I was able to create a workaround locally, but thought would make sense to export this to make it easier for others.

For others reference, this is how I did it:
```ts
// server/middleware/auth.ts

import { clerkClient } from '@clerk/clerk-sdk-node'
import { createClerkRequest } from '@clerk/backend/internal'

const incomingMessageToRequest = req => {
  const headers = Object.keys(req.headers).reduce((acc, key) => Object.assign(acc, { [key]: req?.headers[key] }), {})
  const protocol = req.connection?.encrypted ? 'https' : 'http'
  const dummyOriginReqUrl = new URL(req.url || '', `${protocol}://clerk-dummy`)
  return new Request(dummyOriginReqUrl, {
    method: req.method,
    headers: new Headers(headers)
  })
}

export default defineEventHandler(async event => {
  const requestState = await clerkClient.authenticateRequest(
    createClerkRequest(incomingMessageToRequest(event.node.req))
  )
  const auth = requestState.toAuth()

  event.context.user = auth
})
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/lazy-crabs-exist.md
+++ b/.changeset/lazy-crabs-exist.md
@@ -1,0 +1,5 @@
+---
+"@clerk/clerk-sdk-node": patch
+---
+
+[sdk-node] Export authenticateRequest function
--- a/packages/sdk-node/src/index.ts
+++ b/packages/sdk-node/src/index.ts
@@ -24,3 +24,4 @@ export { createClerkExpressRequireAuth, createClerkExpressWithAuth };
 
 export { requireAuth } from './requireAuth';
 export { withAuth } from './withAuth';
+export { authenticateRequest } from './authenticateRequest';


nikosdouvlis:
Hello @greguintow , many thanks for the PR!
We have plans to deprecate `@clerk/clerk-sdk-node` and replace it with `@clerk/express` soon as the "node SDK" abstraction no longer feels valid - there are many JS runtimes introduced the past 2 years that need to be supported.

The isomorphic `@clerk/backend` package supports all JS runtimes and except for some low-level utils, it should export the same helpers as `@clerk/clerk-sdk-node` except for the helpers that are specific to Express. 

Would this maybe handle your use case better? If not, it'd be great if you could share any insights to help us better understand your requirements :)

greguintow:
> Hello @greguintow , many thanks for the PR! We have plans to deprecate `@clerk/clerk-sdk-node` and replace it with `@clerk/express` soon as the "node SDK" abstraction no longer feels valid - there are many JS runtimes introduced the past 2 years that need to be supported.
> 
> The isomorphic `@clerk/backend` package supports all JS runtimes and except for some low-level utils, it should export the same helpers as `@clerk/clerk-sdk-node` except for the helpers that are specific to Express.
> 
> Would this maybe handle your use case better? If not, it'd be great if you could share any insights to help us better understand your requirements :)

hey @nikosdouvlis , yeah I think that would be good, thaanks!

nikosdouvlis:
> > Hello @greguintow , many thanks for the PR! We have plans to deprecate `@clerk/clerk-sdk-node` and replace it with `@clerk/express` soon as the "node SDK" abstraction no longer feels valid - there are many JS runtimes introduced the past 2 years that need to be supported.
> > The isomorphic `@clerk/backend` package supports all JS runtimes and except for some low-level utils, it should export the same helpers as `@clerk/clerk-sdk-node` except for the helpers that are specific to Express.
> > Would this maybe handle your use case better? If not, it'd be great if you could share any insights to help us better understand your requirements :)
> 
> hey @nikosdouvlis , yeah I think that would be good, thaanks!

Great :) I'm going to close this one for now - please reopen it if you think the existing solution is not enough! 
Many thanks for opening the PR in the first place 🙏🏻

PR #3302, opened 2024-05-01
fix(clerk-js): UI fixes for phone input and OTP inputs

octoper:
## Description

This PR fixes 2 UI issues that were found when using `@tailwindcss/forms`, the first problem was that phone number had the country code flowing over the phone input and the other one was that the OTP where huge due to `box-sizing: content-box`

## Before
CleanShot 2024-05-01 at 21 17 01@2x

CleanShot 2024-05-01 at 21 16 23@2x

# After
CleanShot 2024-05-01 at 21 31 45@2x

CleanShot 2024-05-01 at 21 15 53@2x

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
+++ b/packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
@@ -190,7 +190,7 @@ const PhoneInputBase = forwardRef<HTMLInputElement, PhoneInputProps & { feedback
                 borderWidth: 0,
                 borderTopLeftRadius: 0,
                 borderBottomLeftRadius: 0,
-                paddingLeft: `${`+${selectedCountryOption.country.code}`.length + 1}ch`,
+                paddingLeft: `${`+${selectedCountryOption.country.code}`.length + 1.5}ch`,
                 '&:focus': {
                   borderColor: 'unset',
                   boxShadow: 'unset',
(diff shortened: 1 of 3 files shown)


desiprisg:
Did we check if these issues appear only without `simpleStyles: true` ? If they appear with or without it, I'd suggest doing a check with both settings and see if both are fixed.

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
```diff
@@ -180,14 +179,22 @@ const PhoneInputBase = forwardRef<HTMLInputElement, PhoneInputProps & { feedback
...
               transitionTimingFunction: t.transitionTiming.$common,
               transitionDuration: t.transitionDuration.$focusRing,
+              // This ensures that this input will never have any border or boxShadow styles.
+              '&[type=tel]': {
+                borderRadius: t.radii.$md,
+                borderWidth: 0,
+                borderTopLeftRadius: 0,
+                borderBottomLeftRadius: 0,
+                paddingLeft: t.space.$1,
+                '&:focus': {
```
desiprisg, line 193: What about hover and active states?

octoper:
@desiprisg Yeah I have checked both and they had the same issues but those changes fixes them both

packages/clerk-js/src/ui/elements/PhoneInput/PhoneInput.tsx
octoper: Hover and active states are not changed by default but we can include them as well

desiprisg (approved):

anagstef (approved):

PR #3305, opened 2024-05-01
fix(elements): Prevent multiple verification codes from being sent

BRKalow:
## Description

Fixes an issue where multiple prepare calls were being fired, resulting in multiple verification codes being sent. We're doing this by relying on the response from FAPI, which contains an `expireAt` property indicating when the code is expired. If it's not expired, we won't send a new prepare call by default.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/elements/src/react/sign-in/context/router.context.ts
+++ b/packages/elements/src/react/sign-in/context/router.context.ts
@@ -35,6 +35,8 @@ export function useSignInRouteRegistration<
     isMounted.current = true;
 
     return () => {
+      if (isMounted.current) return;
+
       routerRef.send({
         type: 'ROUTE.UNREGISTER',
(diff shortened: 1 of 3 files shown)


packages/elements/src/react/sign-in/context/router.context.ts
```diff
@@ -35,6 +35,8 @@ export function useSignInRouteRegistration<
     isMounted.current = true;
 
     return () => {
+      if (isMounted.current) return;
+
```
BRKalow, line 39:
```suggestion
```

tmilewski (approved):

PR #3318, opened 2024-05-03
fix(clerk-js): Set eTLD+1 as the domain for client_uat

BRKalow:
## Description

Set an explicit domain value for our `clientUat`  cookie. This mirrors logic in FAPI. Without this, we were seeing duplicate `__client_uat` cookies, which causes issues when they get out of sync. I also removed an unused helper method, `getAllETLDs()`, that I stumbled across.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/utils/cookies/getCookieDomain.ts
+++ b/packages/clerk-js/src/utils/cookies/getCookieDomain.ts
@@ -13,6 +13,10 @@ export function getCookieDomain(hostname = window.location.hostname) {
     return eTLDPlusOne;
   }
 
+  if (hostname === 'localhost') {
+    return hostname;
+  }
+
   const hostnameParts = hostname.split('.');
 
   // we know for sure that the first entry is definitely a TLD, skip it
--- a/packages/clerk-js/src/utils/cookies/getCookieDomain.ts
+++ b/packages/clerk-js/src/utils/cookies/getCookieDomain.ts
@@ -7,13 +7,14 @@ import { createCookieHandler } from '@clerk/shared/cookie';
  */
 let eTLDPlusOne: string;
 const eTLDCookie = createCookieHandler('__clerk_test_etld');
-export function getCookieDomain(hostname = window.location.hostname) {
+
+export function getCookieDomain(hostname = window.location.hostname, cookieHandler = eTLDCookie) {
   // only compute it once per session to avoid unnecessary cookie ops
   if (eTLDPlusOne) {
     return eTLDPlusOne;
   }
 
-  if (hostname === 'localhost') {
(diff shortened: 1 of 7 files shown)


packages/clerk-js/src/utils/url.ts
```diff
@@ -43,19 +43,6 @@ export function getETLDPlusOneFromFrontendApi(frontendApi: string): string {
...
-  const memo = [];
-  const domains = [];
-
-  for (let i = parts.length - 1; i > 0; i--) {
-    memo.unshift(parts[i]);
-    domains.push(memo.join('.'));
-  }
-
-  return domains;
-}
```
BRKalow, line 57: Stumbled on this, and turns out it's unused, so removed 🧹

packages/clerk-js/src/utils/cookies/getCookieDomain.ts
```diff
@@ -0,0 +1,32 @@
...
+export function getCookieDomain(hostname = window.location.hostname) {
+  // only compute it once per session to avoid unnecessary cookie ops
+  if (eTLDPlusOne) {
+    return eTLDPlusOne;
+  }
+
+  const hostnameParts = hostname.split('.');
+
+  // we know for sure that the first entry is definitely a TLD, skip it
+  for (let i = hostnameParts.length - 2; i >= 0; i--) {
```
BRKalow, line 19: Walking backwards here (from the TLD) so we can construct progressively more specific domains. It should be more efficient going from the full domain to the TLD.

colinclerk: How do we solve for localhost?

BRKalow: Fixed!

packages/clerk-js/src/utils/cookies/getCookieDomain.ts
LekoArts, line 1: Can you add unit tests for this? 🙏

LekoArts (changes requested):

packages/clerk-js/src/utils/cookies/getCookieDomain.ts
nikosdouvlis: @BRKalow would this work for addresses poitning to localhost like `0.0.0.0`?

packages/clerk-js/src/utils/cookies/getCookieDomain.ts
nikosdouvlis: Done!

PR #3321, opened 2024-05-03
chore(backend): Fix JWKS cache

BRKalow:
## Description

Fixes a bug in the JWKS cache where it was getting cleared but not treated as expired.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/tokens/keys.ts
+++ b/packages/backend/src/tokens/keys.ts
@@ -73,7 +73,7 @@ export function loadClerkJWKFromLocal(localKey?: string): JsonWebKey {
         n: modulus,
         e: 'AQAB',
       },
-      -1, // local key never expires in cache
+      false, // local key never expires in cache
     );
   }
 
@@ -158,9 +158,7 @@ export async function loadClerkJWKFromRemote({
     } else {
       throw new TokenVerificationError({
         action: `Go to your Dashboard and validate your secret and public keys are correct. ${TokenVerificationErrorAction.ContactSupport} if the issue persists.`,
-        message: `Unable to find a signing key in JWKS that matches the kid='${kid}' of the provided session token. Please make sure that the __session cookie or the HTTP authorization header contain a Clerk-generated session JWT.${
-          jwkKeys ? ` The following kid is available: ${jwkKeys}` : ''
-        }`,
+        message: `Unable to find a signing key in JWKS that matches the kid='${kid}' of the provided session token. Please make sure that the __session cookie or the HTTP authorization header contain a Clerk-generated session JWT. The following kid is available: ${jwkKeys}`,
         reason: TokenVerificationErrorReason.JWKKidMismatch,
       });
     }
--- a/packages/backend/src/tokens/keys.ts
+++ b/packages/backend/src/tokens/keys.ts
@@ -1,4 +1,4 @@
-import { API_URL, API_VERSION, JWKS_CACHE_TTL_MS, MAX_CACHE_LAST_UPDATED_AT_SECONDS } from '../constants';
+import { API_URL, API_VERSION, MAX_CACHE_LAST_UPDATED_AT_SECONDS } from '../constants';
 import {
   TokenVerificationError,
   TokenVerificationErrorAction,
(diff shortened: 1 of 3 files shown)


packages/backend/src/tokens/keys.ts
```diff
@@ -26,19 +26,9 @@ function getCacheValues() {
...
-
-  if (jwksCacheTtlInMs >= 0) {
-    setTimeout(() => {
-      if (jwk) {
-        delete cache[jwk.kid];
-      } else {
-        cache = {};
-      }
-    }, jwksCacheTtlInMs);
-  }
```
BRKalow, line 41: We already have a hard-coded 5 minute TTL for the values in the jwks cache. This setTimeout appears to be resulting in a scenario where the cache is cleared by the timeout here immediately after the 5 minute TTL is hit, so the cache remains empty for the remainder of the 5 minute TTL.

packages/backend/src/tokens/keys.ts
```diff
@@ -116,11 +106,10 @@ export async function loadClerkJWKFromRemote({
   apiUrl = API_URL,
   apiVersion = API_VERSION,
   kid,
-  jwksCacheTtlInMs = JWKS_CACHE_TTL_MS,
```
BRKalow, line 119: This doesn't seem to be functionally doing anything if beyond 5 minutes, as we have the hardcoded 5 minute max cache time.

packages/backend/src/tokens/__tests__/keys.test.ts
```diff
@@ -184,5 +185,38 @@ export default (QUnit: QUnit) => {
         }
       }
     });
+
+    test('cache TTLs do not conflict', async assert => {
```
BRKalow, line 189: This a really specific test case for the observed bug. I confirmed that this test fails on `main` and passes on this branch. Not sure if it's worthwhile to keep around for a long time with the removal of the timer-based cache eviction. 🤔

tmilewski (approved):

packages/backend/src/tokens/keys.ts
```diff
@@ -151,13 +142,21 @@ export async function loadClerkJWKFromRemote({
...
-      reason: TokenVerificationErrorReason.RemoteJWKMissing,
-    });
+    if (!jwkKeys) {
+      throw new TokenVerificationError({
+        action: TokenVerificationErrorAction.ContactSupport,
+        message: `No signing key found in JWKS. This is likely a bug in Clerk. (cache=${JSON.stringify(
+          cache,
+        )}, lastUpdatedAt=${lastUpdatedAt}, kid=${kid})`,
+        reason: TokenVerificationErrorReason.RemoteJWKMissing,
+      });
```
panteliselef, line 152: Why do we need this ? Are we not confident in the fix, or is this another known complex issue ?

BRKalow: Oops, I think this logic is wrong. The thought is that the cache should never be empty at this point, so it helps distinguish between failure states. It shouldn't be possible though.

I feel confident in the fix though, happy to remove.

integration/tests/handshake.test.ts
```diff
@@ -160,7 +160,9 @@ test.describe('Client handshake @generic', () => {
       }),
       redirect: 'manual',
     });
+    console.log(res.headers);
     expect(res.status).toBe(307);
+    expect(res.headers.get('x-clerk-auth-reason')).toBe('session-token-outdated');
```
BRKalow, line 165:
```suggestion
    expect(res.status).toBe(307);
```

packages/backend/src/tokens/keys.ts
```diff
@@ -208,7 +196,19 @@ async function fetchJWKSFromBAPI(apiUrl: string, key: string, apiVersion: string
...
+  if (lastUpdatedAt === -1) {
+    return false;
+  }
+
+  // If the cache has expired, clear the value so we don't attempt to make decisions based on stale data
+  const isExpired = Date.now() - lastUpdatedAt >= MAX_CACHE_LAST_UPDATED_AT_SECONDS * 1000;
+
+  if (isExpired) {
+    cache = {};
+  }
```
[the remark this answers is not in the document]
nikosdouvlis: I agree that a predicate function shouldn't mutate external state

packages/backend/src/tokens/keys.ts
```diff
@@ -208,7 +196,19 @@ async function fetchJWKSFromBAPI(apiUrl: string, key: string, apiVersion: string
 }
 
 function cacheHasExpired() {
-  return Date.now() - lastUpdatedAt >= MAX_CACHE_LAST_UPDATED_AT_SECONDS * 1000;
```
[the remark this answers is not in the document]
dimkl: @nikosdouvlis I agree that we should align all of them to be in MS. It was capped at 5 minutes because we considered that it would be safe to consider that in 5 minutes all the systems related to the JWKs would have been updated to include a new instance or an update to the JWKs initiated by our Clerk API (eg Edge cache, read-only replicas, ...)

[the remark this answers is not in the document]
nikosdouvlis: That makes sense but we probably need to document jwksCacheTtlInMs better if we don't deprecated it as the prop is somewhat confusing

PR #3329, opened 2024-05-06
refactor(types,shared,elements): Assign `telemetry` property to main Clerk interface

LauraBeatris:
## Description

Unblocks SDK-1672

Assigns `telemetry` property to main Clerk instance to access it within React hooks from `shared` package, like so:

````ts
const clerk = useClerkInstanceContext();
clerk.telemetry?.record(eventMethodCalled('useOrganization'));
````

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### CODE REVIEW

File: packages/clerk-js/src/core/clerk.ts [hunk 1]
```diff
@@ -151,7 +151,7 @@ export class Clerk implements ClerkInterface {
   public organization: OrganizationResource | null | undefined;
   public user: UserResource | null | undefined;
   public __internal_country?: string | null;
-  public telemetry?: TelemetryCollector;
+  public telemetry: TelemetryCollector | undefined;
```
LauraBeatris, line 154: Updated to `| undefined` instead of `|`, otherwise it leads to a build error: Property 'telemetry' is optional in type 'Clerk' but required in type 'Clerk'.

I think it's also good to keep it consistent with the other attributes in here.

packages/types/src/telemetry.ts
```diff
@@ -0,0 +1,42 @@
+import type { InstanceType } from 'instance';
+
+export type TelemetryEvent = {
```
LauraBeatris, line 3: I realized I haven't imported this within `shared/telemetry` - applying some changes here.

packages/types/src/telemetry.ts
```diff
@@ -0,0 +1,42 @@
...
+   */
+  sdkv?: string;
+  payload: Record<string, string | number | boolean>;
+};
+
+export type TelemetryEventRaw<Payload = TelemetryEvent['payload']> = {
+  event: TelemetryEvent['event'];
+  eventSamplingRate?: number;
+  payload: Payload;
+};
```
panteliselef, line 36: Now that we are exposing these as public API i feel that we need to do the following

Rename `TelemetryEventRaw` to `TelemetryEvent`
Rename `TelemetryEvent` to `TelemetryRecordedEvent` or `TelemetryStoredEvent`

LauraBeatris: I agree that `Raw` seems a bit generic here 🤔 - @BRKalow do you have other suggestions on that naming?

panteliselef:
@LauraBeatris With this change, [this piece of code](https://github.com/clerk/javascript/blob/7ef7bea4e8f1e361bdc86e8fffe99cfa0c108bda/packages/elements/src/react/utils/telemetry.ts) can be updated as well

LauraBeatris:
> @LauraBeatris With this change, [this piece of code](https://github.com/clerk/javascript/blob/7ef7bea4e8f1e361bdc86e8fffe99cfa0c108bda/packages/elements/src/react/utils/telemetry.ts) can be updated as well

@panteliselef I've just removed the assertion & `ts-ignore` comment. I'd prefer to remove the entire hook tho, but don't have a strong opinion on this. 

It seems that it was only added to override the `telemetry` type, and the actual benefit is that you don't have to call `useClerk()` and access it every time.

packages/types/src/telemetry.ts
BRKalow: 🤔 I wouldn't _really_ consider these part of the public API, it's an internal type that we're forced to expose due to how our packages are structured...

Raw feels accurate because it's the "raw" event that we're collecting before transforming it to send to our backend. `TelemetryRecordedEvent` is fine if we want to rename that.

We could also put a JSDoc with `@internal` to make it clear these aren't meant for public usage:
```ts
/**
 * @internal
 */
export type TelemetryEventRaw = ...
```

LauraBeatris: > Raw feels accurate because it's the "raw" event that we're collecting before transforming it to send to our backend. TelemetryRecordedEvent is fine if we want to rename that.

@BRKalow ACK - thanks for the context!  I'm going to move forward with adding `@internal` to those interfaces.

BRKalow (approved):
💯

.changeset/heavy-kings-fix.md
```diff
@@ -0,0 +1,9 @@
+---
+'@clerk/shared': patch
+'@clerk/types': patch
+'@clerk/elements': patch
+---
+
+The following are all internal changes and not relevant to any end-user:
+
+Create type interface for `TelemetryCollector` on `@clerk/types`. This allows to assign `telemetry` on the main Clerk SDK object, while inhering from the actual `TelemetryCollector` implementation.
```
LekoArts, line 9:
```suggestion
Create type interface for `TelemetryCollector` on `@clerk/types`. This allows to assign `telemetry` on the main Clerk SDK object, while inheriting from the actual `TelemetryCollector` implementation.
```

LekoArts (approved):

PR #3334, opened 2024-05-07
fix(nextjs): url-based session syncing for auth() helpers

BRKalow:
## Description

Helper methods from `auth()` that redirect don't currently pass along the dev browser from the current request. This causes apps that use account portal in development to not receive the current dev browser, causing other redirect issues after a user authenticates.

To address this, I've updated `createRedirect()` to accept `devBrowserToken` and apply it to the sign in / up URL when redirecting cross-origin.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


LekoArts (approved):

panteliselef (approved):
lgtm

packages/backend/src/createRedirect.ts
```diff
@@ -12,6 +18,10 @@ const buildUrl = (_baseUrl: string | URL, _targetUrl: string | URL, _returnBackU
   if (returnBackUrl) {
     res.searchParams.set('redirect_url', returnBackUrl.toString());
   }
+  // For cross-origin redirects, we need to pass the dev browser token for URL session syncing
+  if (_devBrowserToken && baseUrl.hostname !== res.hostname) {
+    res.searchParams.set(constants.Cookies.DevBrowser, _devBrowserToken);
```
dimkl, line 23: @BRKalow i think we should use the QueryParameters constants when adding search params instead of Cookies contants even though they both have the same value.
PR to resolve this: https://github.com/clerk/javascript/pull/3355

PR #3341, opened 2024-05-07
feat(clerk-react,shared): Add telemetry events for React hooks

LauraBeatris:
## Description

Resolves SDK-1672

Add telemetry events for `useSignIn`, `useSignUp`, `useOrganizations` and `useOrganizationList`. This will allow us to track the usage of hooks for custom flows.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/react/src/hooks/useSignIn.ts
+++ b/packages/react/src/hooks/useSignIn.ts
@@ -1,5 +1,7 @@
 import { useClientContext } from '@clerk/shared/react';
+import { eventMethodCalled } from '@clerk/shared/telemetry';
 import type { SetActive, SignInResource } from '@clerk/types';
+import { useEffect } from 'react';
 
 import { useIsomorphicClerkContext } from '../contexts/IsomorphicClerkContext';
 import { useAssertWrappedByClerkProvider } from './useAssertWrappedByClerkProvider';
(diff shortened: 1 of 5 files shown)


packages/react/src/hooks/useSignIn.ts
```diff
@@ -24,6 +26,10 @@ export const useSignIn: UseSignIn = () => {
   const isomorphicClerk = useIsomorphicClerkContext();
   const client = useClientContext();
 
+  useEffect(() => {
+    isomorphicClerk.telemetry?.record(eventMethodCalled('useSignIn'));
```
LauraBeatris, line 30: Hooks aren't necessary methods per se. However, I don't see the benefits of creating another event type for it, cause it wouldn't change much in terms of querying that data. 

Anyway, open for feedback here!

packages/react/src/hooks/useSignUp.ts
```diff
@@ -24,6 +26,10 @@ export const useSignUp: UseSignUp = () => {
   const isomorphicClerk = useIsomorphicClerkContext();
   const client = useClientContext();
 
+  useEffect(() => {
+    isomorphicClerk.telemetry?.record(eventMethodCalled('useSignUp'));
+  }, [isomorphicClerk.telemetry]);
```
panteliselef, line 31: Based on the new cache mechanism, i believe we can skip the usage of useEffect

panteliselef (approved):
Maybe in the future we can have `eventHookCalled` where we track the params passed to those hooks, useful for useOrganization and useOrganizationList, but i'm happy with this for now

packages/react/src/hooks/useSignUp.ts
LauraBeatris: Makes sense. Even without the `useEffect`, it'll only record the event once for multiple function executions. 

Removed on https://github.com/clerk/javascript/pull/3341/commits/e41aaad55ee6b22573e6d1ba11d3a7a76085af80

PR #3343, opened 2024-05-07
feat(elements): Complete and Reset Updates

tmilewski:
## Description

This PR contains the results of multiple similar PRs. As-such, this PR ensures the following:

- [x] The sign-up/in fallback prop isn't presented if the user isn't visiting the root (start) route.
- [x] Upon completion of sign-up/in, we no longer render the Start step.
- [x] Fixes the Sign In/Sign Up machine not appropriately attaching new form machines, thus breaking the flow when returning to an already completed flow.
- [x] Stops the `STRATEGY.UNREGISTER` event from being called on stopped machines.
- [x] Updates `xstate` to `v5.13.0` 
 

Fixes SDK-1745

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/react/sign-in/root.tsx
```diff
@@ -94,6 +94,7 @@ export function SignInRoot({
 
   // TODO: eventually we'll rely on the framework SDK to specify its host router, but for now we'll default to Next.js
   const router = useNextRouter();
+  const isRootPath = path === router.pathname();
```
tmilewski, line 97: This is required, on both Sign Up and Sign In, otherwise, in the case of callbacks, the fallback will show regardless.

packages/elements/src/internals/machines/sign-in/router.machine.ts
```diff
@@ -429,7 +434,7 @@ export const SignInRouterMachine = setup({
       },
     },
     Callback: {
-      tags: 'route:callback',
+      tags: 'route:complete',
```
tmilewski, line 437: Do we want a callback state? It won't show for any meaningful amount of time.

packages/elements/src/react/sign-in/root.tsx
```diff
@@ -44,7 +44,9 @@ function SignInFlowProvider({ children, exampleMode }: SignInFlowProviderProps)
         actor.send(evt);
       }
     });
-  }, [clerk, exampleMode, formRef, router]);
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [clerk, exampleMode, formRef?.id, !!router]);
```
tmilewski, line 49: I saw that formRef and router were causing many unnecessary re-renders. I narrowed the checks down here and in sign up to optimize.

packages/elements/src/react/sign-in/verifications.tsx
```diff
@@ -80,7 +80,7 @@ export function SignInStrategy({ children, name }: SignInStrategyProps) {
     }
 
     return () => {
-      if (factorCtx) {
+      if (factorCtx?.getSnapshot().status === 'active') {
```
tmilewski, line 83: Ensure we only run this event on an active machine.

LekoArts (approved):

PR #3345, opened 2024-05-08
fix(remix): Ensure headers with the same name do not get overwritten

BRKalow:
## Description

When authenticating a request, it's possible we may pass multiple headers with the same name (e.g. multiple `Set-Cookie` headers). By using `headers.set()`, we were only accepting the last header of that name. Switching to `headers.append()` should fix this.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/empty-cobras-sit.md
+++ b/.changeset/empty-cobras-sit.md
@@ -1,0 +1,5 @@
+---
+'@clerk/remix': patch
+---
+
+Fixes a bug in the Remix SDK where headers passed from Clerk with the same name would get overwritten.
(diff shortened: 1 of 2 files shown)


BRKalow:
!snapshot

.changeset/empty-cobras-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/remix': patch
+---
+
+Fixes a bug in the Remix SDK where headers passed from Clerk with the same name would get overwritten.
```
LekoArts, line 5:
```suggestion
Fixes a bug where headers passed from Clerk with the same name would get overwritten.
```

LekoArts (approved):

PR #3351, opened 2024-05-09
fix(clerk-js): Fix Smart captcha on Sign Up ticket flow

anagstef:
## Description

This PR fixes the logic determining if the captcha should fallback to 'Invisible' if the element for the 'Smart' captcha is not found in the DOM. The widget is now selected and rendered synchronously.

### Explanation:

Previously the order of the events was the following:
1. Determine if clerk-js should use the `Smart` or the `Invisible` Captcha widget.
2. Download the CF Turnstile script.
3. Render the Captcha.

The 2nd step above runs asynchronously and sometimes loading states can remove the `#clerk-captcha` element from the DOM.

After this change, we move the 2nd step to be done first and synchronously decide the widget and render it (step 1 and 3).

The addition of the captcha element on the loading card may solve our current issue. However, the change described above is a better practice for avoiding future issues.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/utils/captcha.ts
+++ b/packages/clerk-js/src/utils/captcha.ts
@@ -100,7 +100,7 @@ export const getCaptchaToken = async (captchaOptions: {
   const { siteKey, scriptUrl, widgetType, invisibleSiteKey } = captchaOptions;
   let captchaToken = '',
     id = '';
-  let invisibleWidget = !widgetType || widgetType === 'invisible';
+  let isInvisibleWidget = !widgetType || widgetType === 'invisible';
   let turnstileSiteKey = siteKey;
 
   let widgetDiv: HTMLElement | null = null;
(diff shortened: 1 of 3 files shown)


packages/clerk-js/src/utils/captcha.ts
```diff
@@ -112,29 +112,29 @@ export const getCaptchaToken = async (captchaOptions: {
...
+          if (visibleDiv) {
+            visibleDiv.style.display = 'block';
+            widgetDiv = visibleDiv;
+          } else {
+            console.error('Captcha DOM element not found. Using invisible captcha widget.');
+            widgetDiv = createInvisibleDOMElement();
+            isInvisibleWidget = true;
+            turnstileSiteKey = invisibleSiteKey;
+          }
+        }
```
panteliselef, line 135: ❓ It's not obvious to me, why moving this code is beneficial, could you add more details ?

anagstef: @panteliselef added more context in the description of the PR

panteliselef (approved):

PR #3354, opened 2024-05-09
fix(clerk-expo): Use atob polyfill when using Hermes JS engine

desiprisg:
## Description

Use a polyfill for the `atob` function to prevent errors when using the Hermes JS engine, since the engine's implementation is stricter than it should be and errors when the publishable key's base64 encoded part's length is not divisible by 4.

Related issue by the default JS engine for expo 51: https://github.com/facebook/hermes/issues/1379

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/expo/src/polyfills/base64Polyfill.ts
+++ b/packages/expo/src/polyfills/base64Polyfill.ts
@@ -1,9 +1,14 @@
 import { decode, encode } from 'base-64';
 
-if (!global.btoa) {
+//@ts-expect-error HermesInternal is added by the Hermes JS Engine
+const isHermes = !!global.HermesInternal;
+
+// See Default Expo 51 engine Hermes' issue: https://github.com/facebook/hermes/issues/1379
+if (!global.btoa || isHermes) {
   global.btoa = encode;
 }
 
-if (!global.atob) {
+// See Default Expo 51 engine Hermes' issue: https://github.com/facebook/hermes/issues/1379
+if (!global.atob || isHermes) {
   global.atob = decode;
 }
--- a/.changeset/plenty-comics-matter.md
+++ b/.changeset/plenty-comics-matter.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-expo': patch
+---
+
+Use a polyfill for the `atob` function to prevent errors when using the Hermes JS engine, since the engine's `atob` implementation is stricter than it should be.


packages/expo/src/polyfills/base64Polyfill.ts
```diff
@@ -4,6 +4,7 @@ if (!global.btoa) {
   global.btoa = encode;
```
anagstef, line 4: ❓ do you think we should also always use the `encode` polyfill in case there are differences between them and cause issues?

dimkl: i think we should be consistent and use `encode`, just like Stefanos indicated.

nikosdouvlis:
Can we detect the expo version used and polyfil for >51?

packages/expo/src/polyfills/base64Polyfill.ts
desiprisg: We can, but I can't see how that would result in an error. This is supposed to be a temporary fix for our customers that are encountering this issue due to the Hermes JS engien. A fix for that is already merged by them.

desiprisg:
Opted in checking for hermes and only applying the polyfill in that case. https://reactnative.dev/docs/hermes#confirming-hermes-is-in-use

fredrikburmester:
Just adding a comment to say that I got this error when using `"@clerk/clerk-expo": "^1.0.7",` in Expo 51.

```
 ERROR  Error: Not a valid base64 encoded string length

This error is located at:
    in ClerkProvider (created by RootLayout)
```

**This PR seems to have solved my issue. Thanks!**

packages/expo/src/polyfills/base64Polyfill.ts
anagstef: @desiprisg in a rare case that someone is using both `encode` and `decode` they may get issues. These functions work together, they better come from the same library to be extra safe. Better safe than sorry.

nikosdouvlis: After thinking more about this, I think @anagstef is correct here. Please ignore my offline comment about this and lets polyfill both

nikosdouvlis (approved):

desiprisg:
!snapshot

clerk-cookie:
Hey @desiprisg - the snapshot version command generated the following package versions:
| Package | Version |
| --- | --- |
| @clerk/clerk-expo | 1.0.8-snapshot.v7e206ff |

`@clerk/clerk-expo`
```sh
npm i @clerk/clerk-expo@1.0.8-snapshot.v7e206ff --save-exact
```
`gatsby-plugin-clerk`
```sh
```

anagstef (approved):
💯

PR #3355, opened 2024-05-09
fix(backend,nextjs): Replace DevBrowser constant usage for query params

dimkl:
## Description

Instead of using `Cookies.DevBrowser` we should use `QueryParameters.DevBrowser` when adding dev browser token to the url for consistency even though both constants have the same value.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


nikosdouvlis (approved):

panteliselef (approved):

BRKalow (approved):
Good catch!

PR #3359, opened 2024-05-10
fix(elements): Types for step component

LekoArts:
## Description

While writing the E2E tests I noticed a bug:
I couldn't apply `className` to `SignIn.Step name="choose-strategy"` and `SignIn.Step name="forgot-password"`. Looking at the types, yeah, they only accepted `children`. They both didn't render any markup and were only Context wrappers.

This PR adds `<div>` as markup around these two steps and let's one apply their props to that.

While running the build process I got warnings about circular deps for Elements so I removed the barrel file for the sign-in step component and moved files around.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/react/sign-in/verifications.tsx
```diff
@@ -18,7 +18,7 @@ import {
 } from '~/react/sign-in/context';
 import { createContextFromActorRef } from '~/react/utils/create-context-from-actor-ref';
 
-export type SignInVerificationsProps = { preferred?: ClerkSignInStrategy; children: React.ReactNode } & FormProps;
+export type SignInVerificationsProps = { preferred?: ClerkSignInStrategy } & FormProps;
```
LekoArts, line 21: `FormProps` already marks `children` as `required`

panteliselef (approved):
nice catch

tmilewski (approved):

PR #3362, opened 2024-05-10
chore(clerk-js): Reorganize cookies code and fix TokenUpdate event

dimkl:
## Description

Re-organize cookie codebase into a central place, fix `TokenUpdate` event to be triggered on sign-out and drop duplicate event on refreshing token.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/shared/src/cookie.ts
```diff
@@ -13,8 +13,8 @@ export function createCookieHandler(cookieName: string) {
     /**
      * Setting a cookie will use some defaults such as path being set to "/".
      */
-    set(newValue: string, options: Cookies.CookieAttributes = {}) {
-      return Cookies.set(cookieName, newValue, options);
+    set(newValue: string, options: Cookies.CookieAttributes = {}): void {
+      Cookies.set(cookieName, newValue, options);
```
dimkl, line 17: This could be a breaking change but since it's in the `@clerk/shared` package and it's supposed to be used internally i think we are good to go.

packages/clerk-js/src/core/auth/SessionCookieService.ts
```diff
@@ -0,0 +1,134 @@
...
+import { createClientUatCookie } from './cookies/clientUat';
+import type { SessionCookieHandler } from './cookies/session';
+import { createSessionCookie } from './cookies/session';
+import type { DevBrowser } from './devBrowser';
+import { createDevBrowser } from './devBrowser';
+import { SessionCookiePoller } from './SessionCookiePoller';
+
+// TODO: make SessionCookieService singleton since it handles updating cookies using a poller
+// and we need to avoid updating them concurrently.
+export class SessionCookieService {
```
BRKalow, line 18: It's hard to see the diff of this file as it was moved...can you outline the changes made here briefly?

BRKalow, line 18: Can we provide a brief JSDoc comment here outlining what this class is responsible for? I believe it has expanded a bit with this PR. Maybe obvious now, but helpful for the future if more folks need to contribute.

packages/clerk-js/src/core/auth/cookies/clientUat.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const CLIENT_UAT_COOKIE_NAME = '__client_uat';
+
+export type ClientUatCookieHandler = {
+  set: (client: ClientResource | undefined) => void;
+  get: () => number;
+};
+
+export const createClientUatCookie = (): ClientUatCookieHandler => {
+  const clientUatCookie = createCookieHandler(CLIENT_UAT_COOKIE_NAME);
```
BRKalow, line 16: There's a comment in `session.ts` that mentions what the cookie is used for, maybe we can do the same here and in `devBrowser`?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -741,8 +738,13 @@ export class Clerk implements ClerkInterface {
...
 
-    await this.#authService?.setAuthCookiesFromSession(newSession);
-
+    // Sync __session and __client_uat to cookies using events.TokenUpdate dispatched event
+    // only for newSession is null since the getToken will not be executed. Since getToken
+    // triggers internally a events.TokenUpdate there is no need to trigger it when the newSession exists.
+    const token = await newSession?.getToken();
+    if (!token) {
+      eventBus.dispatch(events.TokenUpdate, { token: null });
+    }
```
BRKalow, line 747: 💭  given the comment, this feels like it should live in `getToken()`, if we're already dispatching the event when the token updates anyways. What do you think?

dimkl: This change cannot live in the `getToken()` since this covers the case of the `newSession` being null. In that case the `getToken()` will not be executed but the code will work due to `?.` operator.

packages/clerk-js/src/core/auth/SessionCookieService.ts
dimkl: We probably need to rename the service to `AuthCookieService`. My intention is to create an interface that can be used both for Browser / Native apps and use 2 different implementations. The `SessionCookieService` will be used for Browser.

Changes:
- move code cookies related with auth to this service
- introduce auth related helper methods that use auth cookies to avoid exposing cookies handling outside of this module
- apply changes due to refactoring cookies exposed interface

packages/clerk-js/src/core/auth/SessionCookieService.ts
```diff
@@ -52,33 +87,24 @@ export class SessionCookieService {
   }
 
   private async refreshSessionToken(): Promise<void> {
-    if (!inBrowser()) {
-      return;
-    }
-
```
dimkl: ℹ️ since this class will only be initialized in standard browsers, there is no need to guard for non browsers.

packages/clerk-js/src/core/auth/SessionCookieService.ts
```diff
@@ -52,33 +87,24 @@ export class SessionCookieService {
...
     if (!this.clerk.session) {
       return;
     }
 
     try {
-      this.updateSessionCookie(await this.clerk.session?.getToken());
+      await this.clerk.session.getToken();
```
dimkl, line 95: ℹ️ Since the `this.clerk.session.getToken()` triggers the `events.TokenUpdate` event which updates the session cookie there is no need to invoke `this.updateSessionCookie`.

packages/clerk-js/src/core/auth/SessionCookieService.ts
```diff
@@ -52,33 +87,24 @@ export class SessionCookieService {
...
   }
 
-  private updateSessionCookie(token: TokenResource | string | undefined | null) {
-    const rawToken = typeof token === 'string' ? token : token?.getRawString();
-
-    if (rawToken) {
-      return setSessionCookie(rawToken);
-    }
-    return removeSessionCookie();
+  private updateSessionCookie(token: string | null) {
```
dimkl, line 101: ℹ️ simplified the function signature (dropped the `TokenResource`) since it's being used internally this is not a breaking change.

packages/clerk-js/src/core/auth/SessionCookieService.ts
```diff
@@ -1,35 +1,74 @@
...
+  public setupProduction() {
+    this.devBrowser.clear();
+  }
+
+  public async handleUnauthenticatedDevBrowser() {
+    this.devBrowser.clear();
+    await this.devBrowser.setup();
+  }
+
+  public urlWithAuth(url: URL): URL {
```
dimkl, line 65: ℹ️ i will add JSDoc for this.

packages/clerk-js/src/core/auth/SessionCookieService.ts
```diff
@@ -0,0 +1,134 @@
...
+// TODO: make SessionCookieService singleton since it handles updating cookies using a poller
+// and we need to avoid updating them concurrently.
+export class SessionCookieService {
+  private environment: EnvironmentResource | undefined;
+  private poller: SessionCookiePoller | null = null;
+  private clientUat: ClientUatCookieHandler;
+  private sessionCookie: SessionCookieHandler;
+  private devBrowser: DevBrowser;
+
+  constructor(private clerk: Clerk, fapiClient: FapiClient) {
```
BRKalow, line 25: Parameter properties ✨

BRKalow (approved):
This looks good to me! Approving so you can merge tomorrow, but would like to see some inline comments added based on my initial review.

Nice cleanup!

packages/shared/src/cookie.ts
nikosdouvlis: ❓ why did we remove the `return` statement here?

packages/clerk-js/src/core/auth/cookies/devBrowser.ts
```diff
@@ -0,0 +1,37 @@
...
+import { addYears } from '@clerk/shared/date';
+import { DEV_BROWSER_JWT_KEY } from '@clerk/shared/devBrowser';
+
+import { inCrossOriginIframe } from '../../../utils';
+
+export type DevBrowserCookieHandler = {
+  set: (jwt: string) => void;
+  get: () => string | undefined;
+  remove: () => void;
+};
```
nikosdouvlis, line 11: 🙃 Nit picky, I'd prefer we had a common `CookieHandler` interface for these I think

packages/clerk-js/src/core/auth/SessionCookieService.ts
```diff
@@ -0,0 +1,134 @@
...
+  public setupProduction() {
+    this.devBrowser.clear();
+  }
+
+  public async handleUnauthenticatedDevBrowser() {
+    this.devBrowser.clear();
+    await this.devBrowser.setup();
+  }
+
+  public urlWithAuth(url: URL): URL {
```
nikosdouvlis, line 65:
```suggestion
  public decorateUrlWithDevBrowserToken(url: URL): URL {
```

packages/clerk-js/src/core/auth/SessionCookieService.ts
```diff
@@ -0,0 +1,134 @@
...
+    });
+  }
+
+  public setEnvironment(environment: EnvironmentResource) {
+    this.environment = environment;
+    this.setClientUatCookieForDevelopmentInstances();
+  }
+
+  public isSignedOut() {
+    return this.clientUat.get() <= 0;
```
nikosdouvlis, line 49: Where is this used from?
The check here feels a little off to me because its the first time (afaik) that we infer `signedOut` based on the client_uat cookie in clerk-js.

dimkl:
@BRKalow Could you take a look at the following commits (PR comments are addressed):
- [Rename SessionCookieService to AuthCookieService for clarity](https://github.com/clerk/javascript/pull/3362/commits/d911c3f24f1ee71dcdbc8255b2112b8cc79ed6f9)
- [Add JSDoc for auth services and cookies](https://github.com/clerk/javascript/pull/3362/commits/d9d1b7fc1b3ec6cfac4b62ec76030152f0973901)

packages/clerk-js/src/core/auth/SessionCookieService.ts
dimkl: @nikosdouvlis https://github.com/clerk/javascript/blob/d911c3f24f1ee71dcdbc8255b2112b8cc79ed6f9/packages/clerk-js/src/core/clerk.ts#L1464

packages/clerk-js/src/core/auth/SessionCookieService.ts
dimkl: I think that we should better keep it more abstract since in the future we may be adding more parameters to the url. Wdyt?

packages/clerk-js/src/core/auth/cookies/devBrowser.ts
dimkl: All 3 CookieHandlers do not have any conflict. I have no objection to merge them and use 1 CookieHandler for all. I think the current approach was used to use different parameters names and types for each of those handlers (eg clientUat has `set(client: ClientResource | undefined)` ) and to keep only the used operation for each handler (eg session handler does not have `remove()` method).
@nikosdouvlis do you think we should merge them?

dimkl:
@nikosdouvlis We removed the return statement because we didn't use it and it would make the CookieHandler method signatures used in clerk-js return values the same as the `shared/cookie` cookieHandler which will allow us in the future to merge them if less changes.
We can drop that change if you think it's gonna cause issues or if it does not give us enough value to change it.
ref: https://github.com/clerk/javascript/pull/3362#discussion_r1599821034

packages/clerk-js/src/core/auth/AuthCookieService.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  // TODO(@dimkl): Replace this method call with an event listener to decouple Clerk with setEnvironment
+  public setEnvironment(environment: EnvironmentResource) {
+    this.environment = environment;
+    this.setClientUatCookieForDevelopmentInstances();
+  }
+
+  public isSignedOut() {
+    return this.clientUat.get() <= 0;
+  }
```
BRKalow, line 68: @nikosdouvlis what's the "normal" way to check for signed out?

nikosdouvlis: In clerk-js/clerk-react the usual way to do this would be `!!Clerk.user` I believe. 
clerk-js has a mutable state so we've never had to check a *cookie* for the signed in status. Since we have an in-memory state, I don't think we should rely to the cookie for this.

Could it by that the `isSignedOut` method name is confusing here? 

I see that before this PR, the caller used to do: 
```
  #shouldSyncWithPrimary = (): boolean => {
    if (this.#hasJustSynced()) {
      this.#clearJustSynced();
      return false;
    }

    if (!this.isSatellite) {
      return false;
    }

    return getClientUatCookie() <= 0;
  };
```

If this is the only usage, what if we changed the name to something mre specific?

packages/clerk-js/src/core/auth/SessionCookieService.ts
nikosdouvlis: See my comment here: https://github.com/clerk/javascript/pull/3362#discussion_r1600721576

packages/clerk-js/src/core/auth/AuthCookieService.ts
dimkl: The cookie is being used here because the `#shouldSyncWithPrimary()` is being called before the ClerkJS is loaded so there is no access to the user.
We could enhance the `isSignedOut()` method to cover both cases or we could rename it. @nikosdouvlis What would you prefer? If we decide to rename it, could you suggest an alternative name?
```typescript
  public isSignedOut() {
    if (!this.clerk.loaded) {
      return this.clientUat.get() <= 0;
    }
    return !!this.clerk.user;
  }
  ```

File: packages/clerk-js/src/core/auth/SessionCookieService.ts (hunk 12 above)
Comment (dimkl) [reply packages/clerk-js/src/core/auth/SessionCookieService.ts]: resolving this. let's continue the discussion from the mentioned comment.

File: packages/clerk-js/src/core/auth/SessionCookieService.ts (hunk 11 above)
Comment (dimkl) [reply packages/clerk-js/src/core/auth/SessionCookieService.ts]: I will rename it as suggested.

File: packages/clerk-js/src/core/auth/cookies/clientUat.ts (hunk 3 above)
Comment (dimkl) [reply packages/clerk-js/src/core/auth/cookies/clientUat.ts:22]: https://github.com/clerk/javascript/pull/3362#issuecomment-2110029255

File: packages/clerk-js/src/core/auth/SessionCookieService.ts (hunk 2 above)
Comment (dimkl) [reply packages/clerk-js/src/core/auth/SessionCookieService.ts]: https://github.com/clerk/javascript/pull/3362#issuecomment-2110029255

File: packages/shared/src/cookie.ts (hunk 1 above)
Comment (dimkl) [reply packages/shared/src/cookie.ts:17]: https://github.com/clerk/javascript/pull/3362#issuecomment-2110070935

File: packages/clerk-js/src/core/auth/AuthCookieService.ts (hunk 13 above)
Comment (nikosdouvlis) [reply packages/clerk-js/src/core/auth/AuthCookieService.ts:68]: Talked with Dimitris offline and it seems that my concerns where not valid, lets go with the initial proposal :)

File: packages/clerk-js/src/core/auth/cookies/clientUat.ts [hunk 14]
```diff
@@ -0,0 +1,57 @@
...
+ * Create a long-lived JS cookie to store the client last updated_at timestamp
+ * for development instances (for production instance is set by FAPI).
+ * The cookie is used as hint from the Clerk Backend packages to identify
+ * if the user is authenticated or not.
+ */
+export const createClientUatCookie = (): ClientUatCookieHandler => {
+  const clientUatCookie = createCookieHandler(CLIENT_UAT_COOKIE_NAME);
+
+  const get = (): number => {
+    return parseInt(clientUatCookie.get() || '0', 10);
```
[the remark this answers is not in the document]
dimkl: nice! I will keep the existing code but i will note it for future refactoring.

PR #3373, opened 2024-05-13
fix(clerk-js,shared,clerk-react): Use common cache for swr

panteliselef:
## Description

Alternative solution to #3091 (but for Core-2) that aims to solve the same issue: How to sync data between hooks and UI components.

A common issue with Clerk so far is that the react hooks that are bundled with a host application will have a different cache than the once used with the bundled UI components from clerk-js.

This PR solves that issue by introducing a single common cache that is created a top level and is passed down to any dependencies.

### Scenarios
1. clerk-react has its own cache and it is passed down to clerk-js via the `load` method (discarding the clerk-js default cache).
2. clerk-js is used directly and it uses its own cache.

This ensures that when calling `useOrganizationList({memberships:true})` from the host app, if the hook has already run and fetched data due to a UI component having been rendered then this will be a cache hit (if data are not stale). Previous that would be a cache miss.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/clerk.ts
```diff
@@ -136,6 +136,13 @@ const defaultOptions: ClerkOptions = {
   signUpForceRedirectUrl: undefined,
 };
 
+class ClerkMap extends Map {
+  name = 'clerk-map';
+
+  constructor() {
+    super();
+  }
+}
```
BRKalow, line 145: What's the thought process behind extending `Map` here? In case we want to add unique functionality? If that's the cache, I would maybe call it `ClerkDataCache` or something. If we don't have plans to extend it, I'd maybe drop the `ClerkMap` class altogether.

packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -33,6 +33,7 @@ export function ClerkContextProvider(props: ClerkContextProvider): JSX.Element |
 
   const derivedState = deriveState(clerkLoaded, state, initialState);
   const clerkCtx = React.useMemo(() => ({ value: clerk }), [clerkLoaded]);
+  const clerkCache = React.useMemo(() => ({ provider: () => clerk.cache }), []);
```
BRKalow, line 36: Is clerk guaranteed to be loaded here?

packages/react/src/isomorphicClerk.ts
```diff
@@ -171,6 +181,17 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
...
 
+  // @ts-ignore
+  get cache(): ClerkMap {
+    console.log('this.clerkjs', this.clerkjs?.cache);
+    return cache;
+    // if (this.clerkjs) {
+    //   return this.clerkjs.cache;
+    // } else {
+    //   return null;
+    // }
```
BRKalow, line 192: Some leftover comments here

packages/react/src/isomorphicClerk.ts
```diff
@@ -137,6 +137,16 @@ type IsomorphicLoadedClerk = Without<
...
 
+class ClerkMap extends Map {
+  name = 'clerk-react';
+
+  constructor() {
+    super();
+  }
+}
+
+const cache = new ClerkMap();
```
BRKalow, line 148: If we're gonna have this Map wrapper, maybe we can define it in `shared`?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -156,6 +163,7 @@ export class Clerk implements ClerkInterface {
   protected internal_last_error: ClerkAPIError | null = null;
 
   #publishableKey: string = '';
+  #cache = new ClerkMap();
```
BRKalow, line 166: I do like the idea of this 👍 Seems like it'll be a really useful tool given the split between UI rendered by Clerk and the host.

That said, I'm not super excited about adding more to the public API of our core Clerk class. Given that this cache is pretty squarely a UI concern / helper, I wonder if it's a good time to introduce a `Clerk.ui` module to encapsulate these concerns. What do you think?

packages/clerk-js/src/core/clerk.ts
panteliselef: This will be replaced with a simple `Map`, I was using this for logging purposes in order to identify if the correct "cache" was being used

packages/react/src/isomorphicClerk.ts
panteliselef: i will drop this, as i mentioned above

packages/clerk-js/src/core/clerk.ts
panteliselef: Not sure how the Clerk.ui module will help us here.

Maybe we should simply introduce `Clerk.__internal_setRequestCache()` or `Clerk.__internal_requestCache` (setter/getter)

in order to avoid passing the cache as an option when calling `Clerk.load`.

To answer to your question, i do see the point of having another way of grouping these APIs, but I would not delay the PR for this.

BRKalow: It helps by giving us a way to not continue to grow the top-level API surface area of our core Clerk class. That's fine if we don't want to delay 👍

panteliselef:
A bug in the implementation was discovered, closing and it is fixed, I will reopen.

PR #3374, opened 2024-05-13
fix(clerk-js): Improve logging for CAPTCHA script loading errors

anagstef:
## Description

This PR adds a `console.error` on every CAPTCHA error.

Also, when script loading fails, we log it in the FAPI, meaning that the `/sign_ups` API call will be executed and will include the following value:
```
captchaError: 'captcha_script_failed_to_load'
```

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/resources/SignUp.ts
+++ b/packages/clerk-js/src/core/resources/SignUp.ts
@@ -82,7 +82,6 @@ export class SignUp extends BaseResource implements SignUpResource {
         paramsWithCaptcha.captchaToken = captchaToken;
         paramsWithCaptcha.captchaWidgetType = captchaWidgetTypeUsed;
       } catch (e) {
-        console.error('Clerk CAPTCHA Error: ', e.captchaError || e.message || e);
         if (e.captchaError) {
           paramsWithCaptcha.captchaError = e.captchaError;
         } else {
--- a/packages/clerk-js/src/core/resources/SignUp.ts
+++ b/packages/clerk-js/src/core/resources/SignUp.ts
@@ -82,6 +82,7 @@ export class SignUp extends BaseResource implements SignUpResource {
         paramsWithCaptcha.captchaToken = captchaToken;
         paramsWithCaptcha.captchaWidgetType = captchaWidgetTypeUsed;
       } catch (e) {
+        console.error('Clerk CAPTCHA Error: ', e.captchaError || e.message || e);
         if (e.captchaError) {
           paramsWithCaptcha.captchaError = e.captchaError;
         } else {
--- a/packages/clerk-js/src/utils/captcha.ts
+++ b/packages/clerk-js/src/utils/captcha.ts
@@ -1,8 +1,6 @@
 import { loadScript } from '@clerk/shared/loadScript';
 import type { CaptchaWidgetType } from '@clerk/types';
 
-import { clerkFailedToLoadThirdPartyScript } from '../core/errors';
-
 interface RenderOptions {
   /**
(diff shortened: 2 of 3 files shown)


packages/clerk-js/src/utils/captcha.ts
```diff
@@ -76,9 +74,12 @@ export async function loadCaptcha(url: string) {
...
     try {
       await loadScript(url, { defer: true });
-    } catch (_) {
+    } catch {
       // Rethrow with specific message
-      clerkFailedToLoadThirdPartyScript('Cloudflare Turnstile');
+      console.error('Clerk: Failed to load the CAPTCHA script from the URL: ', url);
+      throw {
+        captchaError: 'captcha_script_failed_to_load',
+      };
```
dimkl, line 82: 🙃 I would suggest we throw a `ClerkError` (an existing one or create a new one) instead of throwing an object. In order to provide a better error handling in the future i think that we will need to use our Error classes.

packages/clerk-js/src/utils/captcha.ts
```diff
@@ -76,9 +74,12 @@ export async function loadCaptcha(url: string) {
   if (!window.turnstile) {
     try {
       await loadScript(url, { defer: true });
-    } catch (_) {
+    } catch {
       // Rethrow with specific message
-      clerkFailedToLoadThirdPartyScript('Cloudflare Turnstile');
```
dimkl, line 81: 🙃 Why did we drop the usage of `clerkFailedToLoadThirdPartyScript` ? Couldn't we enhance it with an optional `url` parameter?

packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -82,6 +82,7 @@ export class SignUp extends BaseResource implements SignUpResource {
         paramsWithCaptcha.captchaToken = captchaToken;
         paramsWithCaptcha.captchaWidgetType = captchaWidgetTypeUsed;
       } catch (e) {
+        console.error('Clerk CAPTCHA Error: ', e.captchaError || e.message || e);
```
panteliselef, line 85: Not a big fan of this, but if we have a big issue and this will help devs with debugging, then i'm ok with it.

agis:
I think it's worth noting that, even more important that the console log that this PR achieves, is the request that we'll make to FAPI's signUp.create in case there's an error loading the script - this was not the case before this PR. I'm mentioning this for reviewers, because it might not be apparent how this is achieved by looking at the code changes.

packages/clerk-js/src/utils/captcha.ts
anagstef: It's a bit complicated! We don't want everything to throw here. We actually send the `captchaError` error value to the FAPI.

packages/clerk-js/src/utils/captcha.ts
anagstef: As described above, the `clerkFailedToLoadThirdPartyScript` throws an Error object which is not what we want to do here.

packages/clerk-js/src/core/resources/SignUp.ts
anagstef: Not a big fan either, but I added this because in some cases we replace the actual error value with a generic error message. We could remove this of course.

anagstef: So, I'll remove this one and leave only the specific one about the script failing.

panteliselef (approved):

PR #3377, opened 2024-05-13
fix(elements): Fix follow-on initialization discrepancies

tmilewski:
## Description

This is an addendum to the sign in/sign up updates seen in #3343.

This PR does the following:

- [x] Removes the need to transition to an alternate "Complete" state when using `.setActive`. The machine retains the current state throughout the `setActive` flow and resets after that's done.
- [x] Fixes the Sign In/Sign Up machine not appropriately attaching new form machines, thus breaking the flow.
- [x] Stops the `STRATEGY.UNREGISTER` event from being called on stopped machines.

Fixes SDK-1746

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/elements/src/react/sign-in/root.tsx
+++ b/packages/elements/src/react/sign-in/root.tsx
@@ -44,7 +44,9 @@ function SignInFlowProvider({ children, exampleMode }: SignInFlowProviderProps) 
         actor.send(evt);
       }
     });
-  }, [clerk, exampleMode, formRef, router]);
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [clerk, exampleMode, formRef?.id, !!router]);
 
   return <SignInRouterCtx.Provider actorRef={actor}>{children}</SignInRouterCtx.Provider>;
 }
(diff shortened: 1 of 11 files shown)


packages/elements/src/react/sign-in/verifications.tsx
```diff
@@ -80,7 +80,7 @@ export function SignInStrategy({ children, name }: SignInStrategyProps) {
     }
 
     return () => {
-      if (factorCtx) {
+      if (factorCtx?.getSnapshot().status === 'active') {
```
tmilewski, line 83: Ensure we only run this event on an active machine.

packages/elements/src/react/sign-in/root.tsx
```diff
@@ -44,7 +44,9 @@ function SignInFlowProvider({ children, exampleMode }: SignInFlowProviderProps)
         actor.send(evt);
       }
     });
-  }, [clerk, exampleMode, formRef, router]);
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [clerk, exampleMode, formRef?.id, !!router]);
```
tmilewski, line 49: I saw that `formRef` and `router` were causing many unnecessary re-renders. I narrowed the checks down here and in sign up to optimize.

PR #3383, opened 2024-05-15
feat(clerk-js,shared): Custom logger to enable logging once per session

desiprisg:
## Description

Without:
Screenshot 2024-05-15 at 11 54 35

With:
Screenshot 2024-05-15 at 11 55 09

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/src/logger.ts
+++ b/packages/shared/src/logger.ts
@@ -1,0 +1,24 @@
+const loggedMessages: Set<string> = new Set();
+
+export const logger = {
+  /**
+   * A custom logger that ensures messages are logged only once.
+   * Reduces noise and duplicated messages when logs are in a hot codepath.
+   */
+  warnOnce: (msg: string) => {
+    if (loggedMessages.has(msg)) {
+      return;
+    }
+
+    loggedMessages.add(msg);
+    console.warn(msg);
+  },
+  logOnce: (msg: string) => {
+    if (loggedMessages.has(msg)) {
+      return;
+    }
+
+    console.log(msg);
+    loggedMessages.add(msg);
+  },
+};
--- a/.changeset/clean-kings-brake.md
+++ b/.changeset/clean-kings-brake.md
@@ -1,0 +1,6 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/shared': patch
+---
+
+Add a custom logger to allow logging a message or warning to the console once per session, in order to avoid consecutive identical logs due to component rerenders.
--- a/packages/clerk-js/src/utils/__tests__/url.test.ts
+++ b/packages/clerk-js/src/utils/__tests__/url.test.ts
@@ -1,3 +1,4 @@
+import { logger } from '@clerk/shared/logger';
 import type { SignUpResource } from '@clerk/types';
 
 import {
@@ -473,7 +474,7 @@ describe('isAllowedRedirectOrigin', () => {
     ['https://test.clerk.com/foo?hello=1', [/https:\/\/www\.clerk\.com/], false],
   ];
 
-  const warnMock = jest.spyOn(global.console, 'warn').mockImplementation();
+  const warnMock = jest.spyOn(logger, 'warnOnce');
 
   beforeEach(() => warnMock.mockClear());
   afterAll(() => warnMock.mockRestore());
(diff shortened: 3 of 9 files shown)


packages/shared/src/logger.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const addToLoggedMessages = (msg: string) => {
+  loggedMessages = [...loggedMessages, msg];
+};
+
+export const logger = {
+  warnOncePerSession: (msg: string) => {
+    if (loggedMessages.find(message => message === msg)) {
+      return;
+    }
```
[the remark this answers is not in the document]
desiprisg: Do you mean `includes` ?

packages/shared/src/logger.ts
```diff
@@ -0,0 +1,24 @@
+let loggedMessages: string[] = [];
+
+const addToLoggedMessages = (msg: string) => {
+  loggedMessages = [...loggedMessages, msg];
+};
+
+export const logger = {
+  warnOncePerSession: (msg: string) => {
```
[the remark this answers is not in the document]
desiprisg: I'll remove it for now, just wanted to be more explicit.

packages/shared/src/logger.ts
```diff
@@ -0,0 +1,24 @@
+let loggedMessages: string[] = [];
+
+const addToLoggedMessages = (msg: string) => {
+  loggedMessages = [...loggedMessages, msg];
+};
```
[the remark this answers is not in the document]
octoper: I agree with @dimkl to use `Set` as will reduce time complexity, `Array.prototype.includes` has `O(n)` where `Set` can have `O(1)` or `O(log(N))`

packages/shared/src/logger.ts
LekoArts, line 1: We should use multiple `Set` here (for warnings, logs, etc.)

We pretty much have most of the logic already here: https://github.com/clerk/javascript/blob/main/packages/shared/src/deprecated.ts

So we can DRY things up here

LekoArts (changes requested):
If you add a new file to the `shared` package you'll need to adjust the https://github.com/clerk/javascript/blob/main/packages/shared/subpaths.mjs array and add `logger` to the `files` array of https://github.com/clerk/javascript/blob/main/packages/shared/package.json

packages/shared/src/logger.ts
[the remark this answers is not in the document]
dimkl: 💯

packages/shared/src/logger.ts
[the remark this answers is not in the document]
dimkl: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has

packages/shared/src/logger.ts
desiprisg: Not sure if we want multiple sets. Would you want the same message to be printed in a warning and log ?

BRKalow: I'm fine with a single set here 👍 It seems unlikely we would log the same message using different levels. If so, only displaying it for the first one feels okay.

packages/shared/src/logger.ts
```diff
@@ -0,0 +1,20 @@
+const loggedMessages: Set<string> = new Set();
+
+export const logger = {
```
BRKalow, line 3:
```suggestion
/**
 * A custom logger that ensures messages are logged only once. Reduces noise and duplicated messages when logs are in a hot codepath.
 */
export const logger = {
```

BRKalow (commented):
👍  let's add a JSDoc, otherwise LGTM!

packages/shared/src/logger.ts
LekoArts: Yeah, the reason was exactly that (different levels). But I guess it's fine then 👍

packages/shared/src/deprecated.ts
```diff
@@ -28,7 +29,7 @@ export const deprecated = (fnName: string, warning: string, key?: string): void
   }
   displayedWarnings.add(messageId);
 
-  console.warn(
+  logger.warnOnce(
```
LekoArts, line 32: This change isn't correct, we'd be tracking the message in two separate sets. The functionality of `warnOnce` is already given from the code above this code block. If you don't want to combine the `deprecated` function and the new logger in this PR, let's leave it for now and it can be refactored later

LekoArts (changes requested):

packages/shared/src/deprecated.ts
desiprisg: Ah yeah I didn't think that's what you meant and I missed that. I'll leave it for another PR.

LekoArts (approved):

PR #3392, opened 2024-05-16
feat(clerk-js,clerk-react,nextjs,types): Drop experimental tag from GoogleOneTap

panteliselef:
## Description

Add support for GoogleOneTap. New APIs listed:
### React component
- `<GoogleOneTap/>`

Customize the UX of the prompt

```tsx
<GoogleOneTap
  cancelOnTapOutside={false}
  itpSupport={false}
  fedCmSupport={false}
/>
```

### Use the component from with Vanilla JS
- `Clerk.openGoogleOneTap(props: GoogleOneTapProps)`
- `Clerk.closeGoogleOneTap()`
### Low level APIs for custom flows
- `await Clerk.authenticateWithGoogleOneTap({ token: 'xxxx'})`
- `await Clerk.handleGoogleOneTapCallback()`

We recommend using this two methods together in order and let Clerk to perform the correct redirections.
```tsx
google.accounts.id.initialize({
  callback: async response => {
    const signInOrUp = await Clerk.authenticateWithGoogleOneTap({ token: response.credential})
    await Clerk.handleGoogleOneTapCallback(signInOrUp, {
      signInForceRedirectUrl: window.location.href,
    })
  },
});
```

In case you want to handle the redirection and session management yourself you can do so like this
```tsx
google.accounts.id.initialize({
  callback: async response => {
    const signInOrUp = await Clerk.authenticateWithGoogleOneTap({ token: response.credential})
    if(signInOrUp.status === 'complete') {
        await Clerk.setActive({
          session: signInOrUp.createdSessionId
        })
    }
  },
});
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -225,27 +223,6 @@ export class SignIn extends BaseResource implements SignInResource {
...
-          // TODO-ONETAP: Add new types when feature is ready for public beta
-          // @ts-expect-error
-          strategy: 'google_one_tap',
-          googleOneTapToken: params.token,
-        });
-      }
-      throw err;
-    }) as Promise<SignInResource | SignUpResource>;
-  };
-
```
panteliselef, line 248: This is moved to Clerk.ts

packages/react/src/isomorphicClerk.ts
```diff
@@ -947,6 +985,13 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
     }
   };
 
+  authenticateWithGoogleOneTap = async (
+    params: AuthenticateWithGoogleOneTapParams,
+  ): Promise<SignInResource | SignUpResource> => {
+    const clerkjs = await this.#waitForClerkJS();
+    return clerkjs.authenticateWithGoogleOneTap(params);
+  };
```
panteliselef, line 993: I want to highlight this change. This goes against the pattern we already have. But I cannot think of a reason why we cannot do this.

chanioxaris (approved):

PR #3395, opened 2024-05-16
chore(testing): Add experimental console warning on Cypress setup

anagstef:
## Description

This PR adds the following console message on the node process that runs the `clerkSetup` function from `@clerk/testing/cypress`:

```
@clerk/testing: Cypress is an experimental project and subject to change in the future.
```

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/testing/README.md
+++ b/packages/testing/README.md
@@ -89,6 +89,8 @@ test("sign up", async ({ page }) => {
 
 ### Cypress
 
+⚠️ **Using Cypress to test Clerk is still in the experimental stage. Please be aware that there may be limitations and potential issues when using this approach.**
+
 Firstly, add your Clerk keys (`CLERK_PUBLISHABLE_KEY` and `CLERK_SECRET_KEY`) to your environment variables file (e.g. `.env.local` or `.env.`).
 You can find these keys in your Clerk Dashboard.
 
--- a/packages/testing/README.md
+++ b/packages/testing/README.md
@@ -89,7 +89,7 @@ test("sign up", async ({ page }) => {
 
 ### Cypress
 
-⚠️ **Using Cypress to test Clerk is still in the experimental stage. Please be aware that there may be limitations and potential issues when using this approach.**
+⚠️ **Please note:** Support for Cypress is still experimental. Be aware that there are limitations and potential issues at this stage. Please open an issue with a minimal reproduction so that these issues can be fixed. Thanks!
 
 Firstly, add your Clerk keys (`CLERK_PUBLISHABLE_KEY` and `CLERK_SECRET_KEY`) to your environment variables file (e.g. `.env.local` or `.env.`).
 You can find these keys in your Clerk Dashboard.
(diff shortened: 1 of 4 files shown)


panteliselef (approved):
I am fine with this, but why didn't we lunch this in `@clerk/testing/experimental/cypress` ?

anagstef:
> I am fine with this, but why didn't we lunch this in `@clerk/testing/experimental/cypress`?

@panteliselef it's a bit complicated. The actual package is not experimental, the Clerk + Cypress integration is. It will be easily removed when we validate it's fully supported.

panteliselef:
@anagstef 
an idea would be to have both
- `@clerk/testing/experimental/cypress`
- `@clerk/testing/playwrite`

Just because i think warning in tests are usually overlooked

packages/testing/src/cypress/setup.ts
```diff
@@ -23,6 +24,8 @@ type ClerkSetupParams = {
  * @throws An error if the Cypress config object is not provided.
  */
 export const clerkSetup = async ({ config, options }: ClerkSetupParams) => {
+  experimentalConsoleWarning();
```
LekoArts, line 27: Can we inline this instead of creating a separate function? If we don't use it in another place it's not really necessary

LekoArts (dismissed):
Let's add a note about this to the README, too

packages/testing/src/cypress/setup.ts
anagstef: We can, although I did it this way so it would be cleaner, but I can change it.

anagstef:
> @anagstef an idea would be to have both
> 
> * `@clerk/testing/experimental/cypress`
> * `@clerk/testing/playwrite`
> 
> Just because i think warning in tests are usually overlooked

@panteliselef The point I was trying to make is that when we fully support Cypress, no actions will be needed for this package. This package just implements testing tokens, which works as expected. This warning is just informative about the Cypress support in general.

packages/testing/src/cypress/setup.ts
```diff
@@ -23,6 +23,11 @@ type ClerkSetupParams = {
  * @throws An error if the Cypress config object is not provided.
  */
 export const clerkSetup = async ({ config, options }: ClerkSetupParams) => {
+  console.log(
+    '\x1b[33m%s\x1b[0m',
+    '@clerk/testing: Cypress is an experimental project and subject to change in the future.',
```
LekoArts, line 28:
```suggestion
    '@clerk/testing: Support for Cypress is experimental and subject to change in the future.',
```

packages/testing/README.md
```diff
@@ -89,6 +89,8 @@ test("sign up", async ({ page }) => {
 
 ### Cypress
 
+⚠️ **Using Cypress to test Clerk is still in the experimental stage. Please be aware that there may be limitations and potential issues when using this approach.**
```
LekoArts, line 92:
```suggestion
⚠️ **Please note:** Support for Cypress is still experimental. Be aware that there are limitations and potential issues at this stage. Please open an issue with a minimal reproduction so that these issues can be fixed. Thanks!
```

PR #3401, opened 2024-05-17
fix(clerk-js): Update existing `SignUp` object on `SignUpContinue` OAuth

LauraBeatris:
## Description

Resolves SDK-1689 - fixes Web3 sign-up flow with social connections.

The SDK was creating a new `SignUp` object on OAuth callback when continuing a signup flow, overriding properties such as `web3_wallet`. 

Instead of creating a new `SignUp` object, it should update the existing one. 

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/SignUp/SignUpContinue.tsx
+++ b/packages/clerk-js/src/ui/components/SignUp/SignUpContinue.tsx
@@ -175,6 +175,7 @@ function _SignUpContinue() {
                 <SignUpSocialButtons
                   enableOAuthProviders={showOauthProviders}
                   enableWeb3Providers={showWeb3Providers}
+                  continueSignUp
                 />
               )}
               <SignUpForm
--- a/.changeset/loud-tables-rush.md
+++ b/.changeset/loud-tables-rush.md
@@ -1,0 +1,9 @@
+---
+'@clerk/clerk-js': patch
+---
+
+A bug was fixed to not override the existing sign-up state on the OAuth callback.
+
+When continuing a sign-up flow with social connections, `@clerk/clerk-js` was creating a new `SignUpResource` object, instead of patching the existing one.
+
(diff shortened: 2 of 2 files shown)


packages/clerk-js/src/ui/components/SignUp/SignUpContinue.tsx
```diff
@@ -175,6 +175,7 @@ function _SignUpContinue() {
                 <SignUpSocialButtons
                   enableOAuthProviders={showOauthProviders}
                   enableWeb3Providers={showWeb3Providers}
+                  continueSignUp
```
LauraBeatris, line 178: https://github.com/clerk/javascript/blob/ff52e7905c0f492c2cd2eac8bb8ece3e726237b7/packages/clerk-js/src/core/resources/SignUp.ts#L207-L208

Since `continueSignUp` wasn't being provided, then a new `SignUp` object got created on the redirect.

LauraBeatris: I might be missing something here and this could ended up introducing side effects with other flows, so please let me know 🙇🏻

panteliselef: ❓Shouldn't the absence of `continueSignUp` create issues for the rest of oath providers ? Did we verify that those were failing as well, because if they were successful that would seem weird.

LauraBeatris: Good point, I haven't verified yet - what are other flows that someone could fall on `ContinueSignUp` so that I can verify that?

panteliselef: We have verified (offline) that flows work as expected

panteliselef (approved):
💯

PR #3379, opened 2024-05-14
feat(clerk-js,clerk-react): Replace mount with open for GoogleOneTap

panteliselef:
## Description

The interaction with Google One Tap resembles a modal like experience and that is why we decided to drop the "mount" method and use "open" for the low level clerk-js API.

We still want to support the `GoogleOneTap` react component that a user can simply mount in their application.

To achieve both things mentioned above, a few changes were necessary to be made in the internal `Portal` components of `clerk-js` and `clerk-react`

This PR introduces:
- `Clerk.__experimental_openGoogleOneTap(props)`
- `Clerk.__experimental_closeGoogleOneTap()`
and drops
- `Clerk.__experimental_mountGoogleOneTap(element, props)`
- `Clerk.__experimental_unmountGoogleOneTap(element)`

## Checklist

- [ ] `npm test` runs as expected.
- [ ] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #3250, opened 2024-04-24
feat(clerk-js): Handle PSU and MFA flows for `<GoogleOneTap/>`

panteliselef:
## Public API updates

By default we are returning back to the location where the flow started.
To accomplish that internally we will use the `redirect_url` query parameter to build the url.
```tsx
<__experimental_GoogleOneTap />
```

In the above example if there is a `SIGN_UP_FORCE_REDIRECT_URL` or `SIGN_IN_FORCE_REDIRECT_URL` set then the developer would need to pass new values as props like this

```tsx
<__experimental_GoogleOneTap
  signInForceRedirectUrl=""
  signUpForceRedirectUrl=""
/>
```

Let the developer configure the experience they want to offer. (All these values are true by default)
```tsx
<__experimental_GoogleOneTap
  cancelOnTapOutside={false}
  itpSupport={false}
  fedCmSupport={false}
/>
```

Moved authenticateWithGoogleOneTap to Clerk singleton
```ts
Clerk.__experimental_authenticateWithGoogleOneTap
```

Created the handleGoogleOneTapCallback in Clerk singleton
```ts
Clerk.__experimental_handleGoogleOneTapCallback
```

## Checklist

- [ ] `npm test` runs as expected.
- [ ] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #3409, opened 2024-05-21
Support for `<GoogleOneTap/>`

panteliselef:
## Description

This PR backports into Core 1
- #3198 
- #3250 
- #3379 
- #3392 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/clerk.ts
```diff
@@ -1159,6 +1220,25 @@ export default class Clerk implements ClerkInterface {
...
+    params: AuthenticateWithGoogleOneTapParams,
+  ): Promise<SignInResource | SignUpResource> => {
+    return this.client?.signIn
+      .create({
+        strategy: 'google_one_tap',
+        token: params.token,
+      })
+      .catch(err => {
+        if (isClerkAPIResponseError(err) && err.errors[0].code === 'external_account_not_found') {
+          return this.client?.signUp.create({
```
anagstef, line 1233: @panteliselef I think I've mentioned this before, but I'm double-checking that this `signUp.create` will trigger a Captcha functionality, and if the user has selected the "Smart" option it will try to find the `clerk-captcha` in the DOM. Thankfully, it won't break, because we fallback to "Invisible", but the fallback mechanism may change in the future, we should not depend on it.

anagstef (approved):

packages/clerk-js/src/core/clerk.ts
panteliselef: ACK

PR #3410, opened 2024-05-21
fix(clerk-js): Use the correct returnBack url when GoogleOneTap remains open across navigations

panteliselef:
## Description

Bug fix: Use the correct returnBack url when GoogleOneTap remains open across navigations.
Previously it will only use the url that existed in the browser url bar at the time the component was initially rendered.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/components/GoogleOneTap/one-tap-start.tsx
```diff
@@ -16,35 +16,14 @@ function _OneTapStart(): JSX.Element | null {
...
-          signUpUrl,
-          continueSignUpUrl,
-          secondFactorUrl,
-          firstFactorUrl,
-          signUpForceRedirectUrl,
-          signInForceRedirectUrl,
-        },
-        navigate,
-      );
+      await clerk.handleGoogleOneTapCallback(res, ctx.generateCallbackUrls(window.location.href), navigate);
```
panteliselef, line 26: ❗️I couldn't find a robust way to listen to url changes (history) so i went with the synchronous solution to pass the current url when the callback runs

desiprisg: Can we do something like:
```
window.history.pushState = new Proxy(window.history.pushState, {
  apply: (target, thisArg, argArray) => {
    // trigger here what you need
    return target.apply(thisArg, argArray);
  },
});
```

desiprisg (approved):
Looks good but let's try to do it with a proxy on `pushState` and see if that simplifies things?!

panteliselef:
We will attempt to find another solution in another PR

PR #3411, opened 2024-05-21
chore(repo): Update commitlint and use globby

LekoArts:
## Description

On the PR https://github.com/clerk/javascript/pull/3058 the action for commitlint failed with:

```shell
Error: Cannot find module 'glob'
```

Because we don't have `glob` in the root `package.json` and the GitHub action was missing the setup step. Since we already have `globby` installed I opted to use that instead. Since it's ESM-only, I switched the config format of commitlint to TypeScript. For that I had to update commitlint itself. The two breaking changes were:

- [Make the package ESM-only](https://github.com/conventional-changelog/commitlint/blob/master/CHANGELOG.md#1900-2024-02-27)
- [Require Node 18](https://github.com/conventional-changelog/commitlint/blob/master/CHANGELOG.md#1800-2023-10-20)

Which is both fine for us.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


BRKalow (approved):

.github/workflows/pr-title-linter.yml
```diff
@@ -16,9 +16,15 @@ jobs:
...
+      - name: Setup
+        id: config
+        uses: ./.github/actions/init
         with:
-          fetch-depth: 0
-          show-progress: false
+          turbo-signature: ${{ secrets.TURBO_REMOTE_CACHE_SIGNATURE_KEY }}
+          turbo-team: ${{ vars.TURBO_TEAM }}
+          turbo-token: ${{ secrets.TURBO_TOKEN }}
+          playwright-enabled: true # Must be present to enable caching on branched workflows
```
BRKalow, line 27: (future thought) I feel like we could make the name for this more explicit, based on the comment.

PR #3428, opened 2024-05-22
chore(clerk-react,nextjs,shared): Allow react and react-dom 19.0.0-beta

nikosdouvlis:
With the next major release, NextJS@15 will deped on react and react-dom v19 that is still in beta. We are updating our peer dependencies for the related packages, otherwise the package managers will fail to resolve the package versions when users try to install or update the Clerk packages.

 Due to the semver specification for prerelease tags (https://semver.org/#spec-item-11), we can't be very exact here. Adding `19.0.0-beta` to the list will also allow `19.0.0-canary` and `19.0.0-rc` releases as well, but this is not something we can or need to solve at the moment.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -86,8 +86,8 @@
   },
   "peerDependencies": {
     "next": "^13.5.4 || ^14.0.3",
-    "react": ">=18",
-    "react-dom": ">=18"
+    "react": ">=18 || >=19.0.0-beta",
+    "react-dom": ">=18 || >=19.0.0-beta"
   },
   "engines": {
     "node": ">=18.17.0"
--- a/.changeset/hot-phones-serve.md
+++ b/.changeset/hot-phones-serve.md
@@ -1,0 +1,7 @@
+---
+"@clerk/nextjs": patch
+"@clerk/clerk-react": patch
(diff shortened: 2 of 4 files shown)


packages/nextjs/package.json
```diff
@@ -86,8 +86,8 @@
   },
   "peerDependencies": {
     "next": "^13.5.4 || ^14.0.3",
-    "react": ">=18",
-    "react-dom": ">=18"
+    "react": ">=18 || 19.0.0-beta",
```
BRKalow, line 89: given that `>=18` has no upper bound, is this change necessary? I'm guessing it might be because of the prerelease identifier.

In any case,  I think this should be:
```
>=18 || >=19.0.0-beta
```

Per https://semver.npmjs.com

nikosdouvlis: Oh damn it, I think I messed up here. 
It 100% needs to be `>=` or the check on the prerelease will always be false.
Let me test quick to make sure

BRKalow (approved):

PR #3430, opened 2024-05-23
chore(clerk-js): Replace qs library with custom implementation

EmmanouelaPothitou:
## Description
Removing the qs library and using the native URLSearchParams API instead reduced the bundle size of clerk.browser.js by 13.5% (from 70.14KB to 60.62KB) and clerk.headless.js by 22% (from 46.5KB to 36.5KB).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/fapiClient.ts
```diff
@@ -32,8 +31,21 @@ export type FapiRequestCallback<T> = (
...
+  return type === 'key' ? camelToSnake(str) : customDefaultEncoder(str);
+};
+const stringifyBody = obj => {
+  return Object.keys(obj)
+    .map(key => {
+      const encodedKey = camelToSnakeEncoder(key, 'key');
+      const encodedValue = obj[key] && camelToSnakeEncoder(obj[key], 'value');
+      return encodedValue && `${encodedKey}=${encodedValue}`;
+    })
+    .join('&');
```
nikosdouvlis, line 48: The map + join operation here makes sense,however, I think we can replace this snippet using the native [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) class

nikosdouvlis: Also, is this something we could use the commented-out `stringifyQueryParams` for?

packages/clerk-js/src/core/fapiClient.ts
```diff
@@ -125,7 +137,18 @@ export function createFapiClient(clerkInstance: Clerk): FapiClient {
...
+    objParams &&
+      Object.keys(objParams).forEach(key => {
+        const value = objParams[key];
+        if (Array.isArray(value)) {
+          value.forEach(item => queryParams.append(key, item));
+        } else {
+          queryParams.append(key, value);
+        }
+      });
+    return queryParams.toString();
```
nikosdouvlis, line 151: Same comment as above, can we replace this with `stringifyQueryParams`?

packages/clerk-js/src/core/fapiClient.ts
```diff
@@ -193,8 +216,11 @@ export function createFapiClient(clerkInstance: Clerk): FapiClient {
     // Currently, this is needed only for form-urlencoded, so that the values reach the server in the form
     // foo=bar&baz=bar&whatever=1
     // @ts-ignore
+
     if (requestInit.headers.get('content-type') === 'application/x-www-form-urlencoded') {
-      requestInit.body = qs.stringify(body, { encoder: camelToSnakeEncoder, indices: false });
+      if (body && typeof body === 'object') {
```
nikosdouvlis, line 221: ❓ This check should probably be moved into `stringifyQueryParams`!

packages/clerk-js/src/ui/router/index.tsx
```diff
@@ -6,3 +6,6 @@ export * from './Route';
 export * from './Switch';
 
 export type { ParsedQs } from 'qs';
+export type ParsedQueryString = {
+  [key: string]: undefined | string | string[] | ParsedQueryString | ParsedQueryString[];
```
nikosdouvlis, line 10: ❓ Are there any cases where the parsed queries would be anything different than `Record<string,string>`?
Can we confirm the above and simplify the type accordingly?

packages/clerk-js/src/utils/querystring.ts
```diff
@@ -1,11 +1,14 @@
-import qs from 'qs';
-
 export const getQueryParams = (queryString: string) => {
-  return qs.parse(queryString || '', {
-    ignoreQueryPrefix: true,
-  }) as Record<string, string>;
-};
+  const queryParamsObject: { [key: string]: string } = {};
+  const queryParams = new URLSearchParams(queryString);
+  queryParams?.forEach((value, key) => {
```
nikosdouvlis, line 4: `queryParams` is always defined here, no need for `?.`
```suggestion
  queryParams.forEach((value, key) => {
```

packages/clerk-js/src/utils/querystring.ts
```diff
@@ -1,11 +1,14 @@
...
+    queryParamsObject[key] = value;
+  });
 
-export const stringifyQueryParams = (params: Record<string, unknown> | Array<unknown>) => {
-  return qs.stringify(params || {});
+  return queryParamsObject as Record<string, string>;
```
nikosdouvlis, line 8: Good job keeping the return value the same while refactoring the internals, that is going to save time refactoring the rest of the usages.

However, I think we only use this in a few places right now. I think it's worth it to change the function to return an instance of `URLSearchParams` and update the places where this is used accordingly. What do you think?

nikosdouvlis:
Good job so far. Please go through the PR comments and once you're ready, don't forget to:
1. Update the PRs description to explain why we're doing this change. Also, can you calculate the change in bundle size?
2. Add a few unit tests that cover fapiClient and the encoders used

packages/clerk-js/src/core/fapiClient.ts
```diff
@@ -193,8 +188,9 @@ export function createFapiClient(clerkInstance: Clerk): FapiClient {
     // Currently, this is needed only for form-urlencoded, so that the values reach the server in the form
     // foo=bar&baz=bar&whatever=1
     // @ts-ignore
+
     if (requestInit.headers.get('content-type') === 'application/x-www-form-urlencoded') {
-      requestInit.body = qs.stringify(body, { encoder: camelToSnakeEncoder, indices: false });
+      requestInit.body = stringifyQueryParams(body, camelToSnake);
```
nikosdouvlis, line 193:
```suggestion
      requestInit.body = stringifyQueryParams(body, { keyEncoder: camelToSnake });
```

packages/clerk-js/src/utils/__tests__/querystring.test.ts
```diff
@@ -1,16 +1,55 @@
...
 });
 
 describe('stringifyQueryParams(object)', () => {
+  // it('converts an object to querystring', () => {
+  //   expect(stringifyQueryParams({})).toEqual('');
+  //   expect(stringifyQueryParams({ foo: '42', bar: '43' })).toBe('foo=42&bar=43');
+  //   expect(stringifyQueryParams({ foo: ['42', '22'], bar: '43' })).toBe('foo=42&foo=22&bar=43');
+  //   expect(stringifyQueryParams({ foo: ['42', '22'], bar: undefined })).toBe('foo=42&foo=22');
+  //   expect(stringifyQueryParams({ unsafeMetadata: { bar: '1' } })).toBe('unsafe_metadata=%7B%22bar%22%3A%221%22%7D');
+  // });
```
nikosdouvlis, line 34: 🔧 Lets remove the comments here

packages/clerk-js/src/utils/__tests__/querystring.test.ts
```diff
@@ -1,16 +1,55 @@
...
+  it('converts an object to querystring when value is an object', () => {
+    expect(stringifyQueryParams({ unsafe_metadata: { bar: '1' } })).toBe('unsafe_metadata=%7B%22bar%22%3A%221%22%7D');
+  });
+
+  it('converts an object to querystring when key is camelCase', () => {
+    expect(stringifyQueryParams({ barFoo: '1' }, { encoder: camelToSnake })).toBe('bar_foo=1');
+    expect(stringifyQueryParams({ unsafeMetadata: { bar: '1' } }, { encoder: camelToSnake })).toBe(
+      'unsafe_metadata=%7B%22bar%22%3A%221%22%7D',
+    );
+  });
```
nikosdouvlis, line 54: Lets add the equivalent test for `getQueryParams` and test that we correctly parse `unsafe_metadata=%7B%22bar%22%3A%221%22%7D` to an object

packages/clerk-js/src/utils/querystring.ts
```diff
@@ -1,11 +1,47 @@
...
+      }
+    } else {
+      queryParamsObject[key] = value;
+    }
+  });
+  return queryParamsObject as Record<string, string>;
+};
+
+type StringifyQueryParamsOptions = {
+  encoder?: (key: string) => string;
```
nikosdouvlis, line 21: Let's name this `keyEncoder` as it only deals with object keys and not their values.

packages/clerk-js/src/utils/__tests__/querystring.test.ts
EmmanouelaPothitou: We talked offline with @nikosdouvlis and we decided we don't need this.

> Lets add the equivalent test for `getQueryParams` and test that we correctly parse `unsafe_metadata=%7B%22bar%22%3A%221%22%7D` to an object

nikosdouvlis:
@EmmanouelaPothitou This looks good to me. Can you include the size reduction in the description?

packages/clerk-js/src/utils/querystring.ts
```diff
@@ -1,11 +1,54 @@
...
+    | undefined
+    | string,
+  opts: StringifyQueryParamsOptions = {},
+) => {
+  if (params === null || params === undefined) {
+    return '';
+  }
+
+  const queryParams = new URLSearchParams();
+  if (params && typeof params === 'object') {
```
dimkl, line 37: 🙃 I think we can simplify the above and this condition with the following:
```typescript
if (!params || typeof params !== 'object') {
  return ''
}
const queryParams = new URLSearchParams();
Object.keys(params).forEach(key => {
   //...
```

packages/clerk-js/src/utils/querystring.ts
```diff
@@ -1,11 +1,54 @@
...
+      const encodedKey = opts.keyEncoder ? opts.keyEncoder(key) : key;
+      const value = params[key];
+      if (Array.isArray(value)) {
+        value.forEach(v => v !== undefined && queryParams.append(encodedKey, v || ''));
+      } else if (value === undefined) {
+        return;
+      } else if (typeof value === 'object' && value !== null) {
+        queryParams.append(encodedKey, JSON.stringify(value));
+      } else {
+        queryParams.append(encodedKey, value || '');
```
dimkl, line 48: ❓ In the case of `{key: null}`  a `key=` will be added in the querysting. Are we sure about this? Also, can we add a test for that?

dimkl (commented):
🤔 In case the `stringifyQueryParams` is being used in the request url, i think we should also urlEncode the values to avoid cases of strings containing invalid url symbols (eg `=`).
Could we test how the `qs` behaves in those cases?

packages/clerk-js/src/utils/querystring.ts
nikosdouvlis: This was the behavior of `qs` and we decided to keep it the same. @EmmanouelaPothitou please add a test for this case as Dimitris pointed out

EmmanouelaPothitou:
> 🤔 In case the `stringifyQueryParams` is being used in the request url, i think we should also urlEncode the values to avoid cases of strings containing invalid url symbols (eg `=`). Could we test how the `qs` behaves in those cases?

Using the append method on a URLSearchParams instance automatically encodes the value, so I think we don't need to use urlEncode.

EmmanouelaPothitou:
!preview

nikosdouvlis:
!preview

EmmanouelaPothitou:
!snapshot

nikosdouvlis:
!snapshot

EmmanouelaPothitou:
!snapshot

clerk-cookie:
Hey @EmmanouelaPothitou - the snapshot version command generated the following package versions:
| Package | Version |
| --- | --- |
| @clerk/chrome-extension | 1.0.18-snapshot.v11a5fcc |
| @clerk/clerk-js | 5.6.1-snapshot.v11a5fcc |
| @clerk/clerk-expo | 1.2.1-snapshot.v11a5fcc |

`@clerk/chrome-extension`
```sh
npm i @clerk/chrome-extension@1.0.18-snapshot.v11a5fcc --save-exact
```
`@clerk/clerk-js`
```sh
npm i @clerk/clerk-js@5.6.1-snapshot.v11a5fcc --save-exact
```
`@clerk/clerk-expo`
```sh
npm i @clerk/clerk-expo@1.2.1-snapshot.v11a5fcc --save-exact
```
`gatsby-plugin-clerk`
```sh
```

nikosdouvlis:
Will be released once #3518 is out

nikosdouvlis (approved):

nikosdouvlis:
Great job @EmmanouelaPothitou!

packages/clerk-js/src/core/__tests__/fapiClient.test.ts
```diff
@@ -90,20 +90,45 @@ describe('buildUrl(options)', () => {
...
+          array: ['item1', 'item2'],
+        },
+      }).href,
+    ).toBe('https://clerk.example.com/v1/foo?array=item1&array=item2&_clerk_js_version=42.0.0');
+  });
 
+  // The return value isn't as expected.
+  // The buildUrl function converts an undefined value to the string 'undefined'
+  // and includes it in the search parameters.
+  it.skip('parses search params when value is undefined', () => {
```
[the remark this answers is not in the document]
EmmanouelaPothitou: The buildUrl function converts an undefined value to the string 'undefined' and includes it in the search parameters. This issue existed before removing the qs library, so I didn't address it in this PR. I will now create a task in Linear to investigate it further. Therefore, I think we should skip this test for now.

PR #3442, opened 2024-05-24
feat(remix): Support new redirect urls as environment variables or options to rootAuthLoader

panteliselef:
## Description

As options
- signInForceRedirectUrl
- signUpForceRedirectUrl
- signInFallbackRedirectUrl
- signUpFallbackRedirectUrl

As environment variables
- CLERK_SIGN_IN_FORCE_REDIRECT_URL
- CLERK_SIGN_UP_FORCE_REDIRECT_URL
- CLERK_SIGN_IN_FALLBACK_REDIRECT_URL
- CLERK_SIGN_UP_FALLBACK_REDIRECT_URL

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/remix/src/ssr/rootAuthLoader.ts
```diff
@@ -49,7 +49,8 @@ export const rootAuthLoader: RootAuthLoader = async (
 
   const loadedOptions = loadOptions(args, opts);
   // Note: authenticateRequest() will throw a redirect if the auth state is determined to be handshake
-  const requestState = await authenticateRequest(args, loadedOptions);
+  const _requestState = await authenticateRequest(args, loadedOptions);
+  const requestState = { ...loadedOptions, ..._requestState };
```
panteliselef, line 53: Should @clerk/backend be aware of these new redirect urls ? Seems like `afterXUrl` are being returned as part of the SignedIn/OutState

dimkl: ❓ Aren't those properties only available in the client? I believe that the only reason the `afterXUrl` exist in the backend is for debugging.
@nikosdouvlis Do you have any more context?

panteliselef: Yes, i would expect them to only be used in the client as well.
Do you think we should update add the new urls in the request state for usage in debugging as well ? Or is it ok to leave the PR as it is now ?

packages/remix/src/client/RemixClerkProvider.tsx
```diff
@@ -80,6 +84,10 @@ export function ClerkProvider({ children, ...rest }: ClerkProviderPropsWithState
     signUpUrl: __signUpUrl,
     afterSignInUrl: __afterSignInUrl,
     afterSignUpUrl: __afterSignUpUrl,
```
octoper, line 86: ❓ Do we still need these or can we mark them as deprecated?

packages/remix/src/ssr/rootAuthLoader.ts
nikosdouvlis: Why wouldn't we want them in `loadedOptions`? I think we can safely add them - improved debuggability is a welcome side effect

nikosdouvlis (approved):

Issue #1043, opened 2023-04-06
Should rootAuthLoader support the loadUser option?


tmcw:
I recently integrated Clerk into my application, which also uses a relational database, so it needed to use Clerk's externalId setting to cross-reference. So, I then used rootAuthLoader with Remix, and the `loadUser` option to get the user object, to get the externalId.

This worked in development but is a hazard for production: Clerk's rate limit is 5 reqs/second, so once any website goes above that, you start randomly showing users logged-out states and throwing errors.

The eventual solution - helpfully shown by Clerk's support, which has been great - is to store the necessary data in the user session to limit database access. However, I think it's probably a good idea to guard against other users falling into this trap: 5 reqs / second is a very surprisingly low rate limit, and it very much makes it so that you shouldn't ever require a request to clerk for each pageload.

- [x] Review the documentation: https://clerk.com/docs
- [x] Go through package changelog files.

dimkl:
Hello @tmcw, thank you for making this comment.
We have an internal ticket to tackle some issues with the options passed in our `middlewares`/ `loaders` / ... across all of our supported SDKs. It seems that there are some inconsistencies based on framework limitations and other reasons.
I leave this issue open and reply when we have finalized our decisions and our API designs to provide a better answer to this.

tmcw:
Thanks. Trying to dodge the rate limit is really our main struggle with Clerk right now. I totally understand why it'd be a good idea to rate-limit login attempts, but for a method like getting a user object or a jwt token, such a brutal rate limit means that we're needing to add caches and workarounds that we didn't anticipate having to build.

perkinsjr:
Spoke to tmcw regarding the rate limiting so we should be good on that front.

clerk-cookie:
This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 10 days.

tmcw:
Unless I'm missing something, I still think this issue is valid: if you use `rootAuthLoader` with the `loadUser` endpoint, your site will break quickly because of the rate limit. We switched to not using `loadUser` and trying to rely on the session object, but if `loadUser` is a footgun, it should probably be documented as such or removed.

dimkl:
It's still missing. I have marked it as not stale.

clerk-cookie:
This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 10 days.

tmcw:
Keeping this open for the same reasons as above.

[...]

PR #3443, opened 2024-05-24
fix(remix): Deprecate `loadUser`, `loadSession` and `loadOrganization` on rootAuthLoader

anagstef:
Currently, `rootAuthLoader` types suggest the usage of `loadUser`, `loadSession` and `loadOrganization` boolean options. Using any of these will result to a call to Clerk Backend API on each root loader execution. This results in rate limiting issues for apps with many users and affects scalability.

This PR deprecates these 3 options so that they're not misused.

Relevant issue: #1043 

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


octoper (approved):
🔥

panteliselef:
@anagstef @octoper I feel like these should have been removed with the Core 2 launch, was this an oversight on our part ?

anagstef:
@panteliselef Yes, we should have removed them.

.changeset/hip-pets-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/remix': patch
+---
+
+Deprecate loadUser, loadSession and loadOrganization on rootAuthLoader options
```
desiprisg, line 5:
```suggestion
Deprecate `loadUser`, `loadSession` and `loadOrganization` on `rootAuthLoader` options
```

desiprisg (approved):

dimkl:
I think we should also use the `import { deprecated } from "@clerk/shared"` to add a deprecation warning in the server logs.
cc: @anagstef

PR #3458, opened 2024-05-28
feat(ui): Add theme builder functionality

alexcarpenter:
## Description

- Reorganize the UI components into a `primitives` and `components` folders
  - `src/primitives/*` - the building blocks for AIO components void of any Clerk Elements/logic
  - `src/components/*` - the AIO components built using primitives and Clerk Elements 
- Adds `<ComponentSelect />` to easily render different components within theme-builder
- Adds `<AppearanceToggle />` and hooks up light/dark controls to `generateColors`
- Adds `<ThemeDialog />` to render the generated CSS

> [!NOTE]  
> theme-builder styling is purposely very basic atm. There will be a design rev for this so I am not investing to much time in making it look nice, more focused on making it work for what we need to rebuild the AIO components.

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/README.md
```diff
@@ -10,9 +10,12 @@
 
 # @clerk/ui
 
-> [!WARNING]  
+> [!WARNING]
 > This is an experimental package and work-in-progress. Please come back later 👋
 
+`src/primitives/*` - the building blocks for AIO components void of any Clerk Elements/logic
```
kylemac, line 16:
```suggestion
- `src/primitives/*` - the building blocks for AIO components void of any Clerk Elements/logic
```

packages/ui/README.md
```diff
@@ -10,9 +10,12 @@
...
+`src/components/*` - the AIO components built using primitives and Clerk Elements
```
kylemac, line 17:
```suggestion
- `src/components/*` - the AIO components built using primitives and Clerk Elements
```

packages/ui/src/components/sign-in/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+            <Field.Input />
+            <Field.Message intent='error'>Identifier is invalid.</Field.Message>
+          </Field.Root>
+          <Button>Continue</Button>
+        </Card.Body>
+      </Card.Content>
+      <Card.Footer>
+        <Card.FooterAction>
+          <Card.FooterActionText>
+            Don&apos;t have an account? <Card.FooterActionLink href='/sign-up'>Sign up</Card.FooterActionLink>
```
kylemac, line 32: we should probably consider i18n even from this early stage. it doesn't have to be a blocker for this PR but we should consider it a fast'ish follow imo

joe-bell: Adding a ticket to Linear for follow-up 🫡

packages/ui/src/components/sign-up/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+            <Field.Message intent='error'>Identifier is invalid.</Field.Message>
+          </Field.Root>
+          <Button>Continue</Button>
+        </Card.Body>
+      </Card.Content>
+      <Card.Footer>
+        <Card.FooterAction>
+          <Card.FooterActionText>
+            Have an account? <Card.FooterActionLink href='/sign-in'>Sign in</Card.FooterActionLink>
+          </Card.FooterActionText>
```
joe-bell, line 33: Not super important, but wondering if it makes sense to break this down into a separate component. Just thinking aloud, not a blocker

packages/ui/src/primitives/field.tsx
```diff
@@ -0,0 +1,124 @@
...
+      >
+        {path}
+      </svg>
+    );
+  }
+
+  return null;
+}
+
+export { Root, Label, Input, Message };
```
joe-bell, line 124: Non-blocker: what's your thinking here instead of inline `export const`?

packages/ui/tsup.config.ts
```diff
@@ -43,13 +43,13 @@ export default defineConfig(overrideOptions => {
...
-      input: 'src/input.tsx',
-      label: 'src/label.tsx',
-      seperator: 'src/seperator.tsx',
+      'primitives/button': 'src/primitives/button.tsx',
+      'primitives/card': 'src/primitives/card.tsx',
+      'primitives/connection': 'src/primitives/connection.tsx',
+      'primitives/field': 'src/primitives/field.tsx',
+      'primitives/seperator': 'src/primitives/seperator.tsx',
+      'components/sign-in/index': 'src/components/sign-in/index.tsx',
+      'components/sign-up/index': 'src/components/sign-up/index.tsx',
```
joe-bell, line 52: Is the `/index` necessary here?

```suggestion
      'components/sign-in': 'src/components/sign-in/index.tsx',
      'components/sign-up': 'src/components/sign-up/index.tsx',
```

packages/ui/theme-builder/app/theme-builder.tsx
joe-bell, line 1: Any reason we couldn't write this directly in `page.tsx`? or is this just to save faffing about with `use client` everywhere?

joe-bell (approved):

packages/ui/src/primitives/field.tsx
alexcarpenter: Updated 👍🏼

packages/ui/theme-builder/app/theme-builder.tsx
alexcarpenter: Yeah, mainly just to keep as simple as possible for now knowing will likely be rebuilding the theme-builder in some fashion.

packages/ui/src/components/sign-up/index.tsx
alexcarpenter: Yeah, I could see a Text component emerging possibly from this. Can definitely revisit.

PR #3472, opened 2024-05-30
feat(elements): Add support for sign in with passkey

panteliselef:
## Description

This PR add support for passkey usage within a SignIn flow

APIs introduced:
- `<SignIn.Passkey />`
- `<SignIn.SupportedStrategy name='passkey'>`
- `<SignIn.Strategy name='passkey'>`
- `<Clerk.Input type='text' passkeyAutofill>`

Usage Examples:
- `<SignIn.Passkey />`
```tsx
<SignIn.Step name='start'>
  <SignIn.Passkey>
    <Clerk.Loading>
      {isLoading => (isLoading ? <Spinner /> : 'Use passkey instead')}. 
    </Clerk.Loading>
  </SignIn.Passkey>
</SignIn.Step>
```

- `<SignIn.SupportedStrategy name='passkey'>`
```tsx
<SignIn.SupportedStrategy asChild name='passkey' >
  <Button>use passkey</Button>
</SignIn.SupportedStrategy>
```

- `<SignIn.Strategy name='passkey'>`
```tsx
<SignIn.Strategy name='passkey'>
  <p className='text-sm'>
    Welcome back <SignIn.Salutation />!
  </p>

  <CustomSubmit>Continue with Passkey</CustomSubmit>
</SignIn.Strategy>
```

- `<Clerk.Input type='text' passkeyAutofill>`
```tsx
<SignIn.Step name='start'>
  <Clerk.Field name='identifier'>
    <Clerk.Label className='sr-only'>Email</Clerk.Label>
    <Clerk.Input passkeyAutofill placeholder='Enter your email address' />
    <Clerk.FieldError />
  </Clerk.Field>
</SignIn.Step/>
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/internals/machines/sign-in/utils/starting-factors.ts
```diff
@@ -1,9 +1,10 @@
 // These utilities are ported from: packages/clerk-js/src/ui/components/SignIn/utils.ts
 // They should be functionally identical.
+import { isWebAuthnSupported } from '@clerk/clerk-js/src/utils/passkeys';
```
panteliselef, line 3: Is this even allowed? Do we need to move this to shared ?

packages/elements/src/internals/machines/sign-in/verification.machine.ts
```diff
@@ -348,7 +349,7 @@ export const SignInFirstFactorMachine = SignInVerificationMachine.provide({
       const currentVerificationExpiration = clerk.client.signIn.firstFactorVerification.expireAt;
       const needsPrepare = resendable || !currentVerificationExpiration || currentVerificationExpiration < new Date();
 
-      if (!params?.strategy || params.strategy === 'password' || !needsPrepare) {
+      if (!params?.strategy || params.strategy === 'password' || params.strategy === 'passkey' || !needsPrepare) {
```
panteliselef, line 352: Skipping prepare for passkeys

packages/elements/src/internals/machines/sign-in/verification.machine.ts
```diff
@@ -368,6 +369,9 @@ export const SignInFirstFactorMachine = SignInVerificationMachine.provide({
       const password = fields.get('password')?.value as string | undefined;
 
       switch (strategy) {
+        case 'passkey': {
+          return await parent.getSnapshot().context.clerk.client.signIn.authenticateWithPasskey();
+        }
```
panteliselef, line 374: When "Submit" is hit we don't want to call attempt, instead call `authenticateWithPasskey`

packages/elements/src/react/common/form/index.tsx
```diff
@@ -270,6 +273,29 @@ const useInput = ({
...
+      setIsSupported(_isSupported);
+    }
+
+    // @ts-expect-error - Depending on type the props can be different
+    if (passthroughProps?.passkeyAutofill) {
+      runAutofillPasskey();
+    }
+
+    // @ts-expect-error - Depending on type the props can be different
+  }, [passthroughProps?.passkeyAutofill]);
```
panteliselef, line 289: Should this be handled inside an actor ?

packages/elements/src/react/common/form/index.tsx
```diff
@@ -306,10 +332,19 @@ const useInput = ({
...
+      autoComplete: 'webauthn',
+    };
+  }
+
   // Filter out invalid props that should not be passed through
   // @ts-expect-error - Doesn't know about type narrowing by type here
   const { validatePassword: _1, ...rest } = passthroughProps;
 
+  // @ts-expect-error - Depending on type the props can be different
+  delete rest['passkeyAutofill'];
```
panteliselef, line 346: Do we have a better way for sanitize props ?

packages/elements/src/internals/machines/sign-in/start.machine.ts
```diff
@@ -97,5 +115,53 @@ export const SignInStartMachine = setup({
...
+          guard: not('isExampleMode'),
+          target: 'AttemptingPasskey',
+          reenter: true,
+        },
+        SUBMIT: {
+          guard: not('isExampleMode'),
+          target: 'Attempting',
+          reenter: true,
+        },
+      },
```
panteliselef, line 149: Enabling "autofill" is kind of fire and forget action, so we should be able to recover from this state and not get stuck. Promise will await until user interact with the field and select a passkey and that may never happen.

packages/elements/src/internals/machines/sign-in/router.machine.ts
```diff
@@ -285,6 +286,22 @@ export const SignInRouterMachine = setup({
         },
       },
       on: {
+        'AUTHENTICATE.PASSKEY': {
+          actions: sendTo(
+            'start',
+            ({ event }) =>
+              ({
+                type: 'AUTHENTICATE.PASSKEY',
+                flow: event.flow,
```
panteliselef, line 295: flow is redundant here, i'll remove it

.changeset/fuzzy-bees-doubt.md
```diff
@@ -0,0 +1,51 @@
+---
+'@clerk/elements': minor
+---
+
+Support passkeys in SignIn flows.
+APIs introduced:
+- `<SignIn.Action passkey />`
```
tmilewski, line 7: I'm more keen on `<SignIn.Passkey>`, but I'm open to what @BRKalow and @LekoArts think.

LekoArts: Gut feeling-wise I'd probably also go with `<SignIn.Passkey>`

packages/elements/src/internals/machines/sign-in/utils/starting-factors.ts
LekoArts: I don't think this will work in the built package, so you have two options:

1. Copy/paste into this file
2. Move this to shared and use from there

packages/elements/src/react/common/form/index.tsx
```diff
@@ -1,3 +1,4 @@
+import { isWebAuthnAutofillSupported } from '@clerk/clerk-js/src/utils/passkeys';
```
LekoArts, line 1: Same as in your https://github.com/clerk/javascript/pull/3472/files#r1621275247 comment

packages/elements/src/react/common/form/index.tsx
```diff
@@ -270,6 +273,29 @@ const useInput = ({
     ref.send({ type: 'FIELD.UPDATE', field: { name, value: initialValue } });
   }, [name, ref, initialValue]);
 
+  const [isSupported, setIsSupported] = React.useState(false);
```
LekoArts, line 276: That name is too generic 😅 Further below I didn't know anymore what `isSupported` refers to

packages/elements/src/react/common/form/index.tsx
LekoArts: Which part are you referring to?

packages/elements/src/react/common/form/index.tsx
LekoArts: Directly above it, you could move that there

packages/elements/src/react/sign-in/action/passkey.tsx
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Resend verification codes during the sign-in process.
+ * This component must be used within the <Step name="start">.
+ *
+ * @note This component is not intended to be used directly. Instead, use the <Action resend> component.
+ *
+ * @example
+ * import { Action } from '@clerk/elements/sign-in';
+ * <Action passkey>Use passkey instead</Action>;
+ */
```
LekoArts, line 24: Needs an update

packages/elements/src/react/common/form/index.tsx
BRKalow: Yeah, passkey specific logic shouldn't live in the common input element like this. Let's try to move it to an actor.

Why does `passkeyAutofill` need to be on the input?

packages/elements/src/internals/machines/sign-in/verification.machine.ts
```diff
@@ -145,7 +145,8 @@ const SignInVerificationMachine = setup({
   },
   guards: {
     isResendable: ({ context }) => context.resendable || context.resendableAfter === 0,
-    isNeverResendable: ({ context }) => context.currentFactor?.strategy === 'password',
+    isNeverResendable: ({ context }) =>
+      context.currentFactor?.strategy === 'password' || context.currentFactor?.strategy === 'passkey',
```
BRKalow, line 149: Consider consolidating this logic and the `nonPreparable` logic so we don't have to maintain separate checks.

.changeset/fuzzy-bees-doubt.md
BRKalow: The action should ideally be a verb. If it's more similar to `SignIn.Connection` in its function, then I agree it should be `<SignIn.Passkey>`

BRKalow (changes requested):
Looking good. I'd like to see the passkey specific logic moved out of the input logic.

packages/elements/src/internals/machines/sign-in/utils/starting-factors.ts
panteliselef: Shame, I will add it to shared

packages/elements/src/react/common/form/index.tsx
panteliselef: This is only necessary if we want to support [passkey autofill](https://web.dev/articles/passkey-form-autofill#annotate_form_input_field).

BRKalow: Should we rely on "the platform" here? It doesn't feel necessary to create a separate abstraction if all it takes is specifying a native property on the element, unless we need to execute some specific logic to enable autofill.

packages/elements/src/react/common/form/index.tsx
panteliselef: I've now removed that state

packages/elements/src/react/common/form/index.tsx
```diff
@@ -170,9 +179,12 @@ const useInput = ({
   onChange: onChangeProp,
   onBlur: onBlurProp,
   onFocus: onFocusProp,
+  // @ts-expect-error - Depending on type the props can be different
+  passkeyAutofill,
```
LekoArts, line 183: If you look at `validatePassword` (and e.g. line 241) you'll see the same problem.

I chose to not add the optional prop to the destructuring here but instead typecast the `passthroughProps` and then get `validatePassword` from there.

Then you need to typecast only once and don't need to add ts-expect-error in multiple places. Makes more sense to me 🤔

packages/elements/src/react/sign-in/action/action.tsx
```diff
@@ -16,6 +16,7 @@ export type SignInActionProps = { asChild?: boolean } & FormSubmitProps &
         submit?: never;
       } & Omit<SignInNavigateProps, 'to'>)
     | { navigate?: never; resend?: never; submit: true }
+    | { navigate?: never; resend?: never; submit?: never }
```
LekoArts, line 19: Arent't you missing `passkey` here?

packages/elements/src/react/sign-in/passkey.tsx
```diff
@@ -0,0 +1,46 @@
...
+const SIGN_IN_PASSKEY_NAME = 'SignInPasskey';
+
+/**
+ * Prompt users to select a passkey from their device in order to sign in.
+ * This component must be used within the <Step name="start">.
+ *
+ *
+ * @example
+ * import { Passkey } from '@clerk/elements/sign-in';
+ * <Passkey>Use passkey instead</Passkey>;
```
LekoArts, line 22: We're leaving out the imports in those examples.
And since we want people to use namespaces we should use `<SignIn.Passkey>` here instead

```suggestion
 * <SignIn.Passkey>Use passkey instead</SignIn.Passkey>;
```

packages/elements/src/react/sign-in/action/action.tsx
panteliselef: Correct this needs to be dropped

packages/elements/src/react/common/form/index.tsx
```diff
@@ -546,7 +580,43 @@ const Input = React.forwardRef<React.ElementRef<typeof RadixControl>, FormInputP
...
+const SignInInput = React.forwardRef<React.ElementRef<typeof RadixControl>, FormInputProps>(
+  (props: FormInputProps, forwardedRef) => {
+    const signInRouterRef = SignInRouterCtx.useActorRef(true);
+    const passkeyAutofillSupported = useSignInPasskeyAutofill();
+
+    React.useEffect(() => {
+      if (passkeyAutofillSupported) {
+        signInRouterRef?.send({ type: 'AUTHENTICATE.PASSKEY.AUTOFILL' });
+      }
+    }, [passkeyAutofillSupported, signInRouterRef]);
```
BRKalow, line 616: Can this be in the `useInput()` hook? Ideally the components here are just thin wrappers around the hook, which contains the business logic

packages/elements/src/react/common/form/index.tsx
```diff
@@ -533,6 +565,8 @@ type FormInputProps =
 const Input = React.forwardRef<React.ElementRef<typeof RadixControl>, FormInputProps>(
   (props: FormInputProps, forwardedRef) => {
     const clerk = useClerk();
+    const passkeyAutofillProp = (props as PasskeyInputProps).passkeyAutofill;
+    const signInRouterRef = SignInRouterCtx.useActorRef(true);
```
BRKalow, line 569: I'm not crazy about referencing the sign in actor directly in these components, but not sure if there's another way. 🤔

Maybe @tmilewski has an idea?

BRKalow (commented):
Looking good. Let's move the logic from the component into `useInput()` to keep the component layer thin.

Any ideas for not referencing the sign in flow actor directly in the common input components?

packages/elements/src/react/common/form/index.tsx
panteliselef: Probably yes, but if felt like something only this component should care about.

because we are using refs here and not the context, it will not error, but i thought it would ok to not litter useInput with sign in specific logic.

packages/elements/src/react/common/form/index.tsx
panteliselef: Is this better  ?
```tsx
<SignIn.Passkey>
   <Clerk.Input /> 
</SignIn.Passkey>
```

panteliselef: I can see issues with that as well, because SignIn.Passkey wouldn't know whether it needs to pass the `autofill` prop or not. To avoid leaking the prop to the dom for everything else, but only pass it to Clerk.Input.

Another thought is to have a `SignIn.PasskeyInput`, but this does not feel right.

panteliselef: We are now simply detecting `autoComplete="webauthn"`

packages/elements/src/react/common/form/index.tsx
panteliselef: Turns out with the current setup this is a bit hard. `useSignInPasskeyAutofill` will use the `useSelector` underneath which would error when the SignInRouter context is not available.

packages/elements/src/react/common/form/index.tsx
```diff
@@ -552,6 +556,11 @@ type FormInputProps =
 const Input = React.forwardRef<React.ElementRef<typeof RadixControl>, FormInputProps>(
   (props: FormInputProps, forwardedRef) => {
     const clerk = useClerk();
+    const field = useInput(props);
+
+    const hasPasskeyAutofillProp = Boolean(field.props.autoComplete?.includes('webauthn'));
+    const allowedTypeForPasskey = (['text', 'email'] as FormInputProps['type'][]).includes(field.props.type);
```
BRKalow, line 562: Should we consider `tel` here for phone numbers?

BRKalow (approved):
👍

packages/elements/src/react/common/form/index.tsx
panteliselef: I think you are correct

PR #3476, opened 2024-05-30
feat(ui): Add theme-builder deploy

BRKalow:
## Description

Add a small script to facilitate deploying the theme-builder app to Vercel as needed.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/ui/theme-builder/scripts/deploy.mjs
+++ b/packages/ui/theme-builder/scripts/deploy.mjs
@@ -1,0 +1,36 @@
+#!/usr/bin/env zx
+/* eslint-disable */
+import 'zx/globals';
+
+process.env.FORCE_COLOR = '1';
+
+if ((await which('vercel', { nothrow: true })) === null) {
+  echo('Vercel CLI not found. Please install it by running `npm i -g vercel`');
+  process.exit(1);
+}
+
+// build top-level
+within(async () => {
+  // repo root
+  cd('../../../');
+
+  // turbo build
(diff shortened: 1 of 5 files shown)


packages/ui/turbo.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "extends": ["//"],
+  "pipeline": {
+    "build": {
+      "inputs": ["src/**"],
+      "outputs": ["dist/**"]
```
BRKalow, line 7: The `theme-builder/package.json` file was getting included as a cache output from the root turbo config for some reason, so I created the nested one here to remove that.

packages/ui/theme-builder/scripts/deploy.mjs
```diff
@@ -0,0 +1,29 @@
...
+  process.exit(1);
+}
+
+// build top-level
+await within(async () => {
+  // repo root
+  cd('../../../');
+
+  // turbo build
+  await spinner('Building dependencies...', () => $`npx turbo build --filter=ui...`);
```
BRKalow, line 18: The `spinner()` usage in this file is because these commands execute subcommands themselves, so the output isn't getting fully piped through to this script. The spinner gives _some_ feedback.

alexcarpenter (approved):
Let's go

PR #3480, opened 2024-05-31
fix(nextjs): Use separate buffers for each awaitable navigation type

panteliselef:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/nextjs/src/app-router/client/useInternalNavFun.ts
```diff
@@ -6,31 +6,48 @@ import type { NextClerkProviderProps } from '../../types';
...
+  name: string;
 }) => {
-  const { windowNav, routerNav } = props;
+  const { windowNav, routerNav, name } = props;
   const pathname = usePathname();
   const [isPending, startTransition] = useTransition();
 
   if (windowNav) {
-    window.__clerk_internal_navFun = (to, opts) => {
+    registerNavigationType(name).fun = (to, opts) => {
```
octoper, line 42:
```suggestion
		const registerNavigation = registerNavigationType(name)
    registerNavigation.fun = (to, opts) => {
```

❓ Is there any issue changing to the above, so we don't always running the function as it is called many times?

packages/nextjs/src/app-router/client/useInternalNavFun.ts
```diff
@@ -6,31 +6,48 @@ import type { NextClerkProviderProps } from '../../types';
...
-        if (!window.__clerk_internal_navPromisesBuffer) {
+        if (!registerNavigationType(name).promisesBuffer) {
           // We need to use window to store the reference to the buffer,
           // as ClerkProvider might be unmounted and remounted during navigations
           // If we use a ref, it will be reset when ClerkProvider is unmounted
-          window.__clerk_internal_navPromisesBuffer = [];
+          registerNavigationType(name).promisesBuffer = [];
         }
-        window.__clerk_internal_navPromisesBuffer.push(res);
+        registerNavigationType(name).promisesBuffer!.push(res);
```
octoper, line 50: ❓ Does this need to be always push, shouldn't we calling push/replace depending the strategy? (Not exactly part of what the PR changes here but I'm curious)

cc @nikosdouvlis @desiprisg

BRKalow: This is separate from the concept of router `push`/`replace`, we're just pushing a promise onto the internal buffer here. I think this should be okay.

packages/nextjs/src/app-router/client/useInternalNavFun.ts
```diff
@@ -6,31 +6,48 @@ import type { NextClerkProviderProps } from '../../types';
...
+    window.__clerk_internal_navigations = {};
+  }
+
+  if (!(name in window.__clerk_internal_navigations)) {
+    // @ts-ignore
+    window.__clerk_internal_navigations[name] = {};
+  }
+
+  return window.__clerk_internal_navigations[name];
+};
```
BRKalow, line 30: nit, this feels more like a getter:

```suggestion
const getClerkNavigationObject = (name: string) => {
    window.__clerk_internal_navigations ??= {};
    window.__clerk_internal_navigations[name] ??= {}
  	return window.__clerk_internal_navigations[name];
};
```

packages/nextjs/src/app-router/client/useInternalNavFun.ts
```diff
@@ -6,31 +6,48 @@ import type { NextClerkProviderProps } from '../../types';
...
-    window.__clerk_internal_navFun = (to, opts) => {
+    registerNavigationType(name).fun = (to, opts) => {
       return new Promise<void>(res => {
-        if (!window.__clerk_internal_navPromisesBuffer) {
+        if (!registerNavigationType(name).promisesBuffer) {
           // We need to use window to store the reference to the buffer,
           // as ClerkProvider might be unmounted and remounted during navigations
           // If we use a ref, it will be reset when ClerkProvider is unmounted
-          window.__clerk_internal_navPromisesBuffer = [];
+          registerNavigationType(name).promisesBuffer = [];
```
BRKalow, line 48: Could be simplified with:
```suggestion
          registerNavigationType(name).promisesBuffer ??= [];
```

Can remove the wrapped if that way.

BRKalow (approved):

PR #3507, opened 2024-06-03
feat(ui): Implements `--font-size` variable

alexcarpenter:
Implements `--font-size` variable that we use to derive the font sizes used throughout the components. Defaults to `0.8125rem` = `13px`.

Font scale was pulled from https://github.com/clerk/javascript/blob/388070e8605dfc924dd913653c1763b2267ddd25/packages/clerk-js/src/ui/customizables/parseVariables.ts#L84-L97

https://linear.app/clerk/issue/SDKI-16/implement-font-size-variable

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/primitives/field.tsx
```diff
@@ -44,7 +44,7 @@ export const Input = React.forwardRef<HTMLInputElement, React.InputHTMLAttribute
       ref={forwardedRef}
       {...props}
       className={cn(
-        "block w-full bg-white text-gray-12 rounded-md bg-clip-padding py-1.5 px-2.5 border border-gray-a6 outline-none focus:ring-[0.1875rem] focus:ring-gray-a3 data-[invalid='true']:border-destructive data-[invalid='true']:focus:ring-destructive/30 focus:border-gray-a8 hover:border-gray-a8 disabled:opacity-50 disabled:cursor-not-allowed text-[0.8125rem]/[1.125rem]",
+        "block w-full bg-white text-gray-12 rounded-md bg-clip-padding py-1.5 px-2.5 border border-gray-a6 outline-none focus:ring-[0.1875rem] focus:ring-gray-a3 data-[invalid='true']:border-destructive data-[invalid='true']:focus:ring-destructive/30 focus:border-gray-a8 hover:border-gray-a8 disabled:opacity-50 disabled:cursor-not-allowed text-base",
```
joe-bell, line 47: Wondering if there's a way to make sure this never goes lower than `1rem` – wdyt?

joe-bell (approved):

packages/ui/src/primitives/field.tsx
alexcarpenter: Something to consider, the current appearance prop accepts below 1rem so I kept as is for now. Not sure how we'd do that in only css variables though.

PR #3508, opened 2024-06-03
fix(clerk-react): Fix forceRedirectUrl and fallbackRedirectUrl when passed to button components

BRKalow:
## Description

When passing `forceRedirectUrl` and `fallbackRedirectUrl` to `SignInButton` and `SignUpButton`, the props were getting renamed and subsequently ignored by the flow when opening in a modal. I fixed the property names and added a few test cases to cover the behavior.

Note that the non-modal flow wasn't broken as it expects the prefixed props (`signInForceRedirectUrl`, `signInFallbackRedierctUrl`).

Writing the tests also uncovered a bug with our modal implementation where the entire modal DOM tree was wrapped in an element that had `aria-hidden`, which I fixed.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/react/src/components/SignInButton.tsx
```diff
@@ -11,21 +12,27 @@ export const SignInButton = withClerk(({ clerk, children, ...props }: WithClerkP
   const child = assertSingleChild(children)('SignInButton');
 
   const clickHandler = () => {
-    const opts = {
+    const opts: SignInProps = {
+      forceRedirectUrl,
+      fallbackRedirectUrl,
```
BRKalow, line 17: The modal mode relies on the non-prefixed props, adding the type here revealed the issue.

packages/react/src/components/SignInButton.tsx
```diff
@@ -11,21 +12,27 @@ export const SignInButton = withClerk(({ clerk, children, ...props }: WithClerkP
...
     };
 
     if (mode === 'modal') {
       return clerk.openSignIn(opts);
     }
-    return clerk.redirectToSignIn(opts);
+    return clerk.redirectToSignIn({
+      ...opts,
+      signInFallbackRedirectUrl: fallbackRedirectUrl,
+      signInForceRedirectUrl: forceRedirectUrl,
```
BRKalow, line 28: when calling the redirect methods, the prefixed props are necessary. Added here.

packages/react/src/components/SignInButton.tsx
```diff
@@ -11,21 +12,27 @@ export const SignInButton = withClerk(({ clerk, children, ...props }: WithClerkP
...
+    });
   };
 
   const wrappedChildClickHandler: React.MouseEventHandler = async e => {
-    await safeExecute((child as any).props.onClick)(e);
+    if (child && typeof child === 'object' && 'props' in child) {
+      await safeExecute(child.props.onClick)(e);
+    }
```
BRKalow, line 35: Some auto-fix of a lint rule caused a type error, so I tweaked the logic a bit here as a result.

packages/clerk-js/src/ui/elements/Modal.tsx
```diff
@@ -50,15 +50,13 @@ export const Modal = withFloatingTree((props: ModalProps) => {
     >
       <ModalContext.Provider value={modalCtx}>
         <Flex
-          aria-hidden
```
BRKalow, line 53: I don't think we want this, as it removes the entire modal from the accessibility tree.

panteliselef (approved):

tmilewski (approved):

packages/react/src/components/__tests__/SignUpButton.test.tsx
```diff
@@ -63,6 +63,7 @@ describe('<SignUpButton/>', () => {
     userEvent.click(btn);
     await waitFor(() => {
       expect(mockRedirectToSignUp).toHaveBeenCalledWith({
+        fallbackRedirectUrl: url,
         signUpFallbackRedirectUrl: url,
```
panteliselef, line 67: @BRKalow  why do we need both ?

packages/clerk-js/src/ui/elements/Modal.tsx
JacobMGEvans: Definitely not 😅

packages/react/src/components/SignInButton.tsx
JacobMGEvans: Interesting.

packages/react/src/components/__tests__/SignUpButton.test.tsx
panteliselef: I believe types only require the prefixed variant, so we are good

dimkl (approved):

PR #3510, opened 2024-06-04
fix(repo): Use correct scope for renovate PR commit msg

LekoArts:
## Description

The PR title checker for https://github.com/clerk/javascript/pull/3486 was failing because it used the scope `chore(sdk-node)`. Right now the renovate bot used the folder names for the scope names - but our commitlint uses the `name` inside each `package.json`. And unfortunately these are not the same 🙈 

So this PR adjusts the renovate config to use the `name` from `package.json` so that the PR title checker will be happy in the future!

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


dimkl (approved):
❓ What are all those changes in the `renovate.json5` ?
Also couldn't we create an object mapper (folder name -> scope name) in the `scripts/renovate-config-generator.mjs` instead of reading the package.json.
I think that we should be more strict In the future and use the same folder and package name in order to be consistent and use the folder names in our scripts.

PR #3513, opened 2024-06-04
feat(ui): Render enabled connections and implement loading states

alexcarpenter:
## Description

- Introduces hooks folder with `getEnabledSocialConnectionsFromEnvironment`
- Render enabled connections from `getEnabledSocialConnectionsFromEnvironment`
- Replace inline busy/loading states with Elements Loading component
- Adds object support within transformer

https://linear.app/clerk/issue/SDKI-22/convert-buttons-to-make-use-of-elements-loading-component
https://linear.app/clerk/issue/SDKI-14/implement-connections-rendering

Screenshot 2024-06-04 at 2 54 47 PM

Follow up ticket to handle rendering connection layout based on enabled count https://linear.app/clerk/issue/SDKI-24/implement-connection-layout-rendering

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/hooks/getEnabledSocialConnectionsFromEnvironment.ts
```diff
@@ -0,0 +1,19 @@
...
+  const { socialProviderStrategies, web3FirstFactors, authenticatableSocialStrategies } = environment.userSettings;
+
+  const enabledConnections = new Set([
+    ...socialProviderStrategies,
+    ...web3FirstFactors,
+    ...authenticatableSocialStrategies,
+  ]);
+
+  return [...OAUTH_PROVIDERS, ...WEB3_PROVIDERS].filter(provider => enabledConnections.has(provider.strategy));
+}
```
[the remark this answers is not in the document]
alexcarpenter: Possibly yeah, hesitant to do any of that abstraction atm but could see that as an option once we get farther along in the project.

packages/ui/src/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+  coinbase: 'Coinbase',
+  spotify: 'Spotify',
+  xero: 'Xero',
+  box: 'Box',
+  slack: 'Slack',
+  linear: 'Linear',
+  x: 'Twitter',
+  enstall: 'Enstall',
+  metamask: 'MetaMask',
+};
```
alexcarpenter, line 32: @joe-bell curious what you think we should do here? Needed this mapping to connect from provider => our icon component. There still is a type issue in usage that I ignored for the time being. Wasn't sure if this was the right approach, or create a Connection specific Icon that maps all of these?

packages/ui/src/primitives/icon.tsx
```diff
@@ -249,6 +249,17 @@ export const Linear = createIcon({
   ),
 });
 
+export const LinkedIn = createIcon({
```
alexcarpenter, line 252: Maybe we remove `LinkedInOIDC` and just use `LinkedIn` since they are both the same icon?

joe-bell: Yeah that makes sense to me; I only did it initially because I was using the icon name as a key

packages/ui/src/utils.ts
joe-bell: I reckon we could probably merge together with `constants` – what do you reckon?

BRKalow (approved):

PR #3431, opened 2024-05-23
ci(repo): Add next v13 variant to e2e tests

panteliselef:
## Description

Previously we were only running e2e tests for the next version defined in the package.json file of its template.
With this PR we are adding a configuration to always run the test with an app with latest next@13 and another app with latest next@14

## Checklist

- [ ] `npm test` runs as expected.
- [ ] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #3516, opened 2024-06-05
fix(repo): Actually run Next.js integration for v13 & v14

LekoArts:
## Description

The PR https://github.com/clerk/javascript/pull/3431 was in the correct state at some point but when it was merged the functionality of running the integration tests of Next.js both in Next 13 and 14 actually didn't work.

After reading https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-expanding-configurations I think what we had previously won't work.

Proof that Next 13 is used:
https://github.com/clerk/javascript/actions/runs/9379903450/job/25825965337?pr=3516#step:7:91

Proof that Next 14 is used:
https://github.com/clerk/javascript/actions/runs/9379903450/job/25825965564?pr=3516#step:7:89

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,7 +129,7 @@ jobs:
 
     strategy:
       matrix:
-        test-name: [ 'generic', 'nextjs', 'express', 'quickstart', 'ap-flows', 'elements' ]
+        test-name: ['generic', 'express', 'quickstart', 'ap-flows', 'elements']
         test-project: ['chrome']
         include:
           - test-name: 'nextjs'


dimkl:
@LekoArts based on the [ref](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-expanding-configurations) posted in the description the example contains the `windows-latest` in os and also includes a variation for the `windows-latest` os.
```yaml
        os: [windows-latest, ubuntu-latest]
        node: [14, 16]
        include:
          - os: windows-latest
            node: 16
            npm: 6
```
I would expect in our case either the 13 or the 14 to be executed twice instead of none 🤔 . Are we sure about this change?
Also the `integration tests (nextjs, chrome)` in the status checks seem to be stuck.

dimkl (approved):

nikosdouvlis (approved):

LekoArts:
I think this happened:

> For each object in the include list, the key:value pairs in the object will be added to each of the matrix combinations if none of the key:value pairs overwrite any of the original matrix values. If the object cannot be added to any of the matrix combinations, a new matrix combination will be created instead. Note that the original matrix values will not be overwritten, but added matrix values can be overwritten.

Anyways, we saw it in the PRs not running and in this PR they work, so should be all fine

PR #3517, opened 2024-06-05
fix(elements): Display `assertIsDefined` errors during development

LekoArts:
## Description

So inside our sign-in verification machine we use `assertIsDefined` a bunch. Those errors weren't displayed anywhere, you could only see them if you've used `NEXT_PUBLIC_CLERK_ELEMENTS_DEBUG=true`.

In SDK-1470 I recorded how I got into such a state some time ago. Actually, I wasn't able to reproduce now because our conditional logic is better now. However, if for _some reason_ a user gets into the state of `assertIsDefined` throwing, they'll now see it at least 👍 

Fixes SDK-1470

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/internals/machines/sign-in/verification.machine.ts
```diff
@@ -352,14 +362,14 @@ export const SignInFirstFactorMachine = SignInVerificationMachine.provide({
         return Promise.resolve(clerk.client.signIn);
       }
 
-      assertIsDefined(params);
+      assertIsDefined(params, 'First factor params');
```
LekoArts, line 365: I've added a `label` to all instances of `assertIsDefined`. If `params` would be `undefined` it would say:

undefined is not defined

With the label it'll say

First factor params is not defined

👍

anagstef (approved):

PR #3519, opened 2024-06-05
chore(clerk-js): Lazily load  VirtualBodyRootPortal

panteliselef:
## Description

With the introduction of `GoogleOneTap` in `clerk-js@5.4.0` the bundle for clerk-js.browser.js was increased from 69KB to 166KB. This PR fixes that.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/soft-beers-hunt.md
+++ b/.changeset/soft-beers-hunt.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Bug fix: statically loading a resource would result in an increase in bundle size for `clerk.browser.js`.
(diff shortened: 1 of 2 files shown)


anagstef (approved):

dimkl (approved):
🤔  I think we had bundlewatch enabled with a hard limit of around 70KB. I think we should add a check to avoid raising the size of the clerk browser without being noticed.

panteliselef:
AFAIK  we will investigate why bundlewatch is not running and we will re introduce it soon in another PR

PR #3524, opened 2024-06-05
feat(ui): Implement `<PhoneNumberField />` component

alexcarpenter:
## Description

- Implements `<PhoneNumberField />`
- Pulls data and hook from clerk-js https://github.com/clerk/javascript/blob/49e7886f1f24a8ac95514e8fe8b0006dca715a79/packages/clerk-js/src/ui/elements/PhoneInput/useFormattedPhoneNumber.ts#L17

https://github.com/clerk/javascript/assets/825855/931db79e-14c2-4fc1-8ed8-f9108604ce85

https://linear.app/clerk/issue/SDKI-10/build-phone-country-code-combobox

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/common/PhoneNumberInput/index.tsx
joe-bell, line 1: Perhaps a nit but does it make more sense to call this `PhoneNumberField` and include the `label` prop?

alexcarpenter: Thinking about how we'd swap between email and phone number like this is why I built the input on its own for now. Likely multiple components that make use of the field internally. e.g. `<PhoneNumberField />` `<EmailOrPhoneNumberField />`?


https://github.com/clerk/javascript/assets/825855/7cbd03f8-9a11-44ad-850a-ee9ffcbb8f2c

joe-bell:
Testing locally it seems that the prefix gets added; but my assumption is the dropdown will take care of that already?
CleanShot 2024-06-06 at 16 17 43@2x

packages/ui/src/common/PhoneNumberInput/index.tsx
joe-bell: Hmmm. What if:

1. We had a field for each
2. We handle how to display said fields in a separate component; which we wrap in `relative` class and absolutely position the specific caption

What do you reckon?

joe-bell: Gotcha, in that case I think let's add a `alternative` slot or something like you suggested; but we create a separate component to handle the logic of which one to display?

alexcarpenter: Sounds good, I will wrap in a field and label.

joe-bell:
Sorry Alex, got my PRs muddled 😅

alexcarpenter:
> Testing locally it seems that the prefix gets added; but my assumption is the dropdown will take care of that already? CleanShot 2024-06-06 at 16 17 43@2x

As you add more numbers, it removes the plus to my knowledge.

alexcarpenter:
> Testing locally it seems that the prefix gets added; but my assumption is the dropdown will take care of that already? CleanShot 2024-06-06 at 16 17 43@2x

The jump from 2-3 in gray scale is fairly drastic.

## `gray-2`
Screenshot 2024-06-06 at 1 44 14 PM


## `gray-3`
Screenshot 2024-06-06 at 1 44 33 PM

joe-bell:
Ooof I agree this is tough, I think we really need a gray in the middle here 🤔 

Anyway, let's loop back. Otherwise this looks great!

joe-bell (approved):

PR #3525, opened 2024-06-05
feat(nextjs,backend,integration): Introduce dynamic keys from `clerkMiddleware` 

LauraBeatris:
## Description

Resolves SDK-1253, SDK-1811, SDK-1770

This PR introduces dynamic keys from `clerkMiddleware`, solving the following problems:
- `auth()` cannot assert the session token signature due to an undefined secret key, leading to a runtime error. 
- `auth.createRedirect` doesn't redirect to `signInUrl` or `signUpUrl` since those aren't propagated as well. 

The propagation is done by encrypting an `x-clerk-request-data` header and sending it from the middleware runtime to the application runtime, where it's going to be decrypted using a key. The key used for encryption is resolved based on the following conditions:
- If security-sensitive options, such as `secretKey` are provided, then `CLERK_ENCRYPTION_KEY` is required
- Otherwise, fallbacks to `CLERK_SECRET_KEY`, which is going to have backward compatibility. 

## Checklist

  - [x] Merge [Docs PR](https://github.com/clerk/clerk-docs/pull/1164) with this PR

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/nextjs/src/app-router/server/ClerkProvider.tsx
```diff
@@ -6,7 +6,9 @@ import { mergeNextClerkPropsWithEnv } from '../../utils/mergeNextClerkPropsWithE
 import { ClientClerkProvider } from '../client/ClerkProvider';
 import { initialState } from './auth';
 
-export function ClerkProvider(props: Without<NextClerkProviderProps, '__unstable_invokeMiddlewareOnAuthStateChange'>) {
+export function ClerkProvider(
+  props: Without<NextClerkProviderProps, '__unstable_invokeMiddlewareOnAuthStateChange' | 'signInUrl' | 'signUpUrl'>,
```
LauraBeatris, line 10: I've deliberately removed `signInUrl` and `signUpUrl` from the allowed props of `ClerkProvider` on the server. 

IMO limiting it only to be passed via the `clerkMiddleware` would cause less confusion, one single source of truth for those options to be used by server-side helpers. 

If we propagate within `ClerkProvider` on the server to the request, then there's the question of which would would take precedence: Middleware options or provider options?

nikosdouvlis: This makes lots of sense to me, but unfortunately, we can't do it at the moment as it's a breaking change :( 

I agree that the options of `clerkMiddleware` should take precedence over the ones defined on `ClerkProvider`.

What would happen if we keep both right now and make sure that the `clerkMiddleware` ones override all else?

packages/nextjs/src/app-router/server/auth.ts
```diff
@@ -28,15 +28,18 @@ export const auth = (): Auth => {
       clerkRequest.clerkUrl.searchParams.get(constants.QueryParameters.DevBrowser) ||
       clerkRequest.cookies.get(constants.Cookies.DevBrowser);
 
+    const requestData = getHeader(request, constants.Headers.ClerkRequestData);
+    const decryptedRequestData = requestData ? decryptClerkRequestData(requestData) : {};
```
nikosdouvlis, line 32: 🙃 (nitpicky comment)
It looks like `requestData` is only being used in the ternary operator here. Could we maybe move the conditional logic in `decryptClerkRequestData`, tweak the helper so it handles undefined/null values gracefully (by returning `{}`) and do this instead?

```
const decryptedRequestData = decryptClerkRequestData(getHeader(request, constants.Headers.ClerkRequestData))
```

packages/nextjs/src/server/errors.ts
```diff
@@ -97,3 +97,5 @@ For additional information about middleware, please visit https://clerk.com/docs
 };
 
 export const authSignatureInvalid = `Clerk: Unable to verify request, this usually means the Clerk middleware did not run. Ensure Clerk's middleware is properly integrated and matches the current route. For more information, see: https://clerk.com/docs/nextjs/middleware. (code=auth_signature_invalid)`;
+
+export const signingKeyInvalid = `Clerk: Unable to decrypt request data, this usually means the signing key is invalid. Ensure the signing key is properly set. (code=signing_key_invalid)`;
```
nikosdouvlis, line 101: Shall we include a short example or instruct the users to check whether the `CLERK_ENCRYPTION_KEY` env var is defined?

packages/nextjs/src/server/utils.ts
```diff
@@ -155,6 +157,7 @@ export function decorateRequest(
       [constants.Headers.AuthMessage]: message || '',
       [constants.Headers.AuthReason]: reason || '',
       [constants.Headers.ClerkUrl]: req.clerkUrl.toString(),
```
nikosdouvlis, line 159: (not a concern of this PR)

I wonder if we'd like to refactor how we propagate these values using the mechanism Laura introduces in this PR.
We don't need encryption for these, but I think Laura's way is cleaner: create an object and append it to the request, parse the request and get an object back that contains all the values needed without dealing with details (such as the header keys used).

Let's discuss offline about a potential refactor to simplify this :)

nikosdouvlis:
I know that this is still a draft PR, so don't worry about this comment yet. Once everything's is in place, let's also add an e2e test covering this flow. Feel free to ping anyone from team-sdk about this :)

packages/nextjs/src/app-router/server/ClerkProvider.tsx
BRKalow: How would these options propagate to components that are not server rendered? We would need to pass the options from middleware to the client (middlware -> server -> client). My recommendation would be to not attempt that, scope it to the properties needed by our server-side helpers for now, and document the constraints clearly.

What do y'all think?

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -76,6 +76,11 @@ export const clerkMiddleware: ClerkMiddleware = withLogger('clerkMiddleware', lo
     logger.enable();
   }
 
+  const hasProvidedRuntimeOptions = params.secretKey || params.signInUrl || params.signInUrl;
+  if (hasProvidedRuntimeOptions) {
+    assertKey(ENCRYPTION_KEY, () => errorThrower.throwMissingEncryptionKeyError());
+  }
```
BRKalow, line 82: Maybe it's not necessary, but we had talked about defaulting to using the secret key as the encryption key **if** no secret key is provided as an option to the middleware. This prevents everyone from needing to provide an encryption key.

So I think we should adjust the check to only error if secret key is provided and no encryption key is found

packages/nextjs/src/server/utils.ts
```diff
@@ -234,3 +237,23 @@ export function assertTokenSignature(token: string, key: string, signature?: str
     throw new Error(authSignatureInvalid);
   }
 }
+
+/**
+ * Encrypt request data using signing key.
+ */
+export function encryptClerkRequestData(options: Partial<AuthenticateRequestOptions>): string {
+  return AES.encrypt(JSON.stringify(options), ENCRYPTION_KEY).toString();
```
BRKalow, line 245: We should fallback to the secret key if no encryption key is provided. This makes it backwards compat and avoids the need for the most common use case to provide an `ENCRYPTION_KEY`.

```
if secretKey is provided as an option, then required ENCRYPTION_KEY
if no secretKey is provided as an option, ENCRYPTION_KEY is optional and falls back to SECRET_KEY
```

packages/nextjs/src/server/errors.ts
LauraBeatris: +1, we should link to the docs where mention regarding the use case of this key.

packages/nextjs/src/server/clerkMiddleware.ts
LauraBeatris: +1, considering the secret key is the only bit from the sensitive payload and therefore requires encryption.

We need to make sure that the SDK will throw in an error if there's not a `CLERK_SECRET_KEY` defined as well.

packages/nextjs/src/app-router/server/ClerkProvider.tsx
LauraBeatris: > scope it to the properties needed by our server-side helpers for now, and document the constraints clearly

+1 on that! I'll start documenting the constraints for each helper/component.

.changeset/young-pigs-live.md
```diff
@@ -0,0 +1,12 @@
...
+'@clerk/nextjs': minor
+'@clerk/shared': minor
+---
+
+Propagate Next.js middleware options to the application server, solving the following issues:
+- `auth` couldn't assert the session token signature due to not having access to `secretKey` passed as an option.
+- `auth.redirectToSignIn` doesn't redirect to `signInUrl` provided as option.
+
+`CLERK_ENCRYPTION_KEY` must be defined when providing `secretKey` as an option, in order to securely propagated this value
+between servers. For more information regarding options propagation, refer to (TODO - Add docs URL here)
```
LauraBeatris, line 12: I'll spin up the docs PRs to merge alongside this feature.

packages/nextjs/src/server/errors.ts
```diff
@@ -97,3 +97,6 @@ For additional information about middleware, please visit https://clerk.com/docs
 };
 
 export const authSignatureInvalid = `Clerk: Unable to verify request, this usually means the Clerk middleware did not run. Ensure Clerk's middleware is properly integrated and matches the current route. For more information, see: https://clerk.com/docs/nextjs/middleware. (code=auth_signature_invalid)`;
+
+export const encryptionKeyInvalid = `Clerk: Unable to decrypt request data, this usually means the encryption key is invalid. Ensure the encryption key is properly set. (code=encryption_key_invalid)
+For additional information, please visit https://clerk.com/docs/references/nextjs/clerk-middleware#clerk-middleware-options.`;
```
LauraBeatris, line 102: This should link to a new section specific to options propagation on `clerkMiddleware`, I'll update this PR later when writing the docs.

LauraBeatris: Is it reasonable for us to quickly deploy the docs changes first, then update this PR and merge it? I believe it's better to link documentation here.

integration/presets/envs.ts
```diff
@@ -60,26 +63,30 @@ const withAPCore1ClerkLatest = environmentConfig()
...
+  .setEnvVariable('private', 'CLERK_ENCRYPTION_KEY', constants.CLERK_ENCRYPTION_KEY);
 
 const withAPCore2ClerkLatest = environmentConfig()
   .setId('withAPCore2ClerkLatest')
   .setEnvVariable('public', 'CLERK_TELEMETRY_DISABLED', true)
   .setEnvVariable('private', 'CLERK_SECRET_KEY', envKeys['core-2-all-enabled'].sk)
   .setEnvVariable('public', 'CLERK_PUBLISHABLE_KEY', envKeys['core-2-all-enabled'].pk)
-  .setEnvVariable('public', 'CLERK_JS_URL', constants.E2E_APP_CLERK_JS || 'http://localhost:18211/clerk.browser.js');
+  .setEnvVariable('public', 'CLERK_JS_URL', constants.E2E_APP_CLERK_JS || 'http://localhost:18211/clerk.browser.js')
+  .setEnvVariable('private', 'CLERK_ENCRYPTION_KEY', constants.CLERK_ENCRYPTION_KEY);
```
LauraBeatris, line 82: I need to add this one to the Vercel project in which runs the integration tests against.

packages/nextjs/src/server/utils.ts
```diff
@@ -234,3 +246,45 @@ export function assertTokenSignature(token: string, key: string, signature?: str
...
+ **/
+export function encryptClerkRequestData(options: Partial<AuthenticateRequestOptions>) {
+  /**
+   * Warns if encryption key is missing when secret key is provided to prevent breaking changes.
+   * Prepares users for a potential escalation to an error in a future major version.
+   */
+  if (options.secretKey && !ENCRYPTION_KEY) {
+    logger.warnOnce(
+      'Clerk: Missing `CLERK_ENCRYPTION_KEY`. Required for propagating `secretKey` middleware option. See docs: https://clerk.com/docs/references/nextjs/clerk-middleware#server-side-options-propagation',
+    );
```
BRKalow, line 262:
```suggestion
    // TODO: change this to an error in the next major version of `@clerk/nextjs`
    logger.warnOnce(
      'Clerk: Missing `CLERK_ENCRYPTION_KEY`. Required for propagating `secretKey` middleware option. See docs: https://clerk.com/docs/references/nextjs/clerk-middleware#server-side-options-propagation',
    );
```

LauraBeatris:
`clerkMiddleware` doesn't have access to the latest env var value, therefore it leads to a decryption error in case that value gets updated/removed and it differs from the initial key value used for encryption. 

It could be the case where a developer updates their env var locally, or in the cloud, and the errors out for fresh requests. 

With that said, I'm thinking of moving this to the BE client and then the encryption key can be accessed within the request state in the singleton 🤔 

CleanShot 2024-06-17 at 13 48 55

LauraBeatris:
> `clerkMiddleware` doesn't have access to the latest env var value, therefore it leads to a decryption error in case that value gets updated/removed and it differs from the initial key value used for encryption.
> 
> It could be the case where a developer updates their env var locally, or in the cloud, and the errors out for fresh requests.
> 
> With that said, I'm thinking of moving this to the BE client and then the encryption key can be accessed within the request state in the singleton 🤔
> 
> CleanShot 2024-06-17 at 13 48 55

Created an [issue](https://github.com/vercel/next.js/issues/66962) on the Next.js repo! Let's wait on the discussion there before moving forward here.

LauraBeatris:
FYI - We're working on sharing the dynamic keys to the BAPI client exported within `@clerk/nextjs`. Otherwise, if `secretKey` is being provided, but not shared to `createClerkClient` singleton, it leads to the following error:

CleanShot 2024-06-18 at 17 35 19

LauraBeatris:
Unit tests for `clerkMiddleware` are failing since we're using AsyncLocalStorage now, I'll update those tomorrow 🫡

LauraBeatris:
Integration tests for `elements` are failing. It's passing on the `main` branch tho, so I need to investigate the root cause here.

packages/nextjs/src/server/clerkClient.ts
```diff
@@ -27,6 +33,54 @@ const clerkClient = createClerkClient({
...
+ */
+const clerkClientForRequest = () => {
+  let requestData;
+
+  try {
+    const request = buildRequestLike();
+    const encryptedRequestData = getHeader(request, constants.Headers.ClerkRequestData);
+    requestData = decryptClerkRequestData(encryptedRequestData);
+  } catch (err) {
+    // When outside of middleware runtime, fallbacks to access request data from `NextRequest`
```
BRKalow, line 59: I _think_ this comment is outdated based on the logic 👀  We're falling back to using our internal async-local store.

packages/nextjs/src/server/errors.ts
```diff
@@ -97,3 +97,6 @@ For additional information about middleware, please visit https://clerk.com/docs
 };
 
 export const authSignatureInvalid = `Clerk: Unable to verify request, this usually means the Clerk middleware did not run. Ensure Clerk's middleware is properly integrated and matches the current route. For more information, see: https://clerk.com/docs/nextjs/middleware. (code=auth_signature_invalid)`;
+
+export const encryptionKeyInvalid = `Clerk: Unable to decrypt request data, this usually means the encryption key is invalid. Ensure the encryption key is properly set. (code=encryption_key_invalid)
+For additional information, please visit https://clerk.com/docs/references/nextjs/clerk-middleware#dynamic-keys.`;
```
BRKalow, line 102:
```suggestion
export const encryptionKeyInvalid = `Clerk: Unable to decrypt request data, this usually means the encryption key is invalid. Ensure the encryption key is properly set. For more information, see: https://clerk.com/docs/references/nextjs/clerk-middleware#dynamic-keys. (code=encryption_key_invalid)`;
```

.changeset/young-pigs-live.md
```diff
@@ -0,0 +1,12 @@
+---
+'@clerk/backend': minor
+'@clerk/nextjs': minor
+'@clerk/shared': minor
+---
+
+Introduces dynamic keys from `clerkMiddleware`, allowing access by server-side helpers like `auth`. Keys such as `signUpUrl`, `signInUrl`, `publishableKey` and `secretKey` are securely encrypted using AES algorithm.
```
BRKalow, line 7:
```suggestion
Introduces dynamic keys from `clerkMiddleware`, allowing access by server-side helpers like `auth`. Options such as `signUpUrl`, `signInUrl`, `publishableKey` and `secretKey` are securely encrypted using AES algorithm and made available to the Next.js backend.
```

.changeset/young-pigs-live.md
```diff
@@ -0,0 +1,12 @@
...
+
+- When providing `secretKey`, `CLERK_ENCRYPTION_KEY` is required as the encryption key. If `secretKey` is not provided, `CLERK_SECRET_KEY` is used by default.
+- `clerkClient` from `@clerk/nextjs` should now be called as a function, and it's singleton is deprecated, in favor of reading dynamic keys from Next.js runtime.
```
BRKalow, line 10:
```suggestion
- `clerkClient` from `@clerk/nextjs` should now be called as a function, and its singleton form is deprecated. This change allows the Clerk backend client to read keys from the current request, which is necessary to support dynamic keys.
```

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -91,82 +94,88 @@ export const clerkMiddleware: ClerkMiddleware = withLogger('clerkMiddleware', lo
...
+      const protect = createMiddlewareProtect(clerkRequest, authObject, redirectToSignIn);
+      const authObjWithMethods: ClerkMiddlewareAuthObject = Object.assign(authObject, { protect, redirectToSignIn });
+
+      let handlerResult: Response = NextResponse.next();
+      try {
+        handlerResult =
+          (await clerkMiddlewareRequestDataStore.run(
+            options,
+            async () => await handler?.(() => authObjWithMethods, request, event),
+          )) || handlerResult;
```
BRKalow, line 142: This logic is getting a bit hard to follow, maybe we can split it apart for clarity?

```suggestion
        const userHandlerResult = await clerkMiddlewareRequestDataStore.run(
          options,
          async () => handler?.(() => authObjWithMethods, request, event),
        )
        handlerResult = userHandlerResult || handlerResult
```

BRKalow (commented):
Looking good!

.changeset/young-pigs-live.md
LauraBeatris: Nice one on providing context regarding the change!

.changeset/young-pigs-live.md
LauraBeatris: I was getting confused between "Keys" and "Options", nice feedback!

packages/nextjs/src/server/errors.ts
LauraBeatris: A bit of wrong formatting, nice catch!

packages/nextjs/src/server/clerkClient.ts
BRKalow: @LauraBeatris I'm wondering if we need to check the error here to make sure it's not something we should re-throw. I think it's possible that `headers()` and `cookies()` throw errors related to static generation bail-out that might need to be rethrown...

LauraBeatris: @BRKalow I've restructured to check if the store is defined first, instead of relying on the error being thrown to determine the runtime. 

```ts
const middlewareStore = clerkMiddlewareRequestDataStore.getStore();
if (Object.values(middlewareStore ?? {}).length) {
  requestData = middlewareStore;
} else {
  // When outside of middleware runtime, fallbacks to access request data from `NextRequest`
  const request = buildRequestLike();
  const encryptedRequestData = getHeader(request, constants.Headers.ClerkRequestData);
  requestData = decryptClerkRequestData(encryptedRequestData);
}
```

.changeset/young-pigs-live.md
```diff
@@ -0,0 +1,12 @@
+---
+'@clerk/backend': minor
+'@clerk/nextjs': minor
+'@clerk/shared': minor
```
BRKalow, line 4: `@clerk/shared` looks untouched 👀 

```suggestion
'@clerk/backend': minor
'@clerk/nextjs': minor
```

BRKalow (approved):
🚢

dimkl:
❓ Why did we choose the approach of the `clerkClient()` Proxy instead of exposing a `getClerkClient()`?
Even though we have already introduced the Proxy in the `@clerk/clerk-sdk-node` to load dynamically the env variables, I think that we could avoid using it here, and introduce a more specific API to avoid any confusion over the `clerkClient`.

cc: @LauraBeatris , @nikosdouvlis , @BRKalow

PR #3535, opened 2024-06-07
feat(ui): OTPField

joe-bell:
## Description

Adds an `OTPField` component; with updated styles to `Field` to accommodate

Took a while to try and accommodate forced states alongside native ones; but fairly confident that CSS variables solves our challenges here.

I didn't wire up the field to a `Step` yet for ease-of-development

https://github.com/clerk/javascript/assets/7349341/e45a6179-3179-49ef-9192-3535538f056b

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/common/OTPField.tsx
```diff
@@ -0,0 +1,68 @@
...
+                        selected: 'focus-visible',
+                        hovered: 'hover',
+                        none: 'native',
+                      } as const
+                    )[status]
+                  }
+                  {...props}
+                  asChild
+                >
+                  <span>{value}</span>
```
alexcarpenter, line 46: Can we implement the cursor when the segment is in the cursor status?

packages/ui/src/common/OTPField.tsx
```diff
@@ -0,0 +1,68 @@
...
+            return (
+              <Field.Message
+                justify='center'
+                intent='error'
+              >
+                {message}
+              </Field.Message>
+            );
+          }}
+        </Common.FieldError>
```
alexcarpenter, line 64: Lets add the resend-able button here as well https://clerk.com/docs/elements/reference/sign-up#action

packages/ui/src/common/OTPField.tsx
```diff
@@ -0,0 +1,68 @@
...
+          {({ state }) => (
+            <Common.Input
+              type='otp'
+              autoSubmit
+              className='flex gap-2'
+              render={({ value, status }) => (
+                <Field.Input
+                  type='otp'
+                  data-status={status}
+                  intent={state}
```
alexcarpenter, line 32: Would this be `data-intent`? or do we need this since data-status is above?

packages/ui/src/common/OTPField.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+        <Common.FieldState>
+          {({ state }) => (
+            <Common.Input
+              type='otp'
+              autoSubmit
+              className='flex gap-2'
+              render={({ value, status }) => (
+                <Field.Input
+                  type='otp'
```
alexcarpenter, line 30: I don't think we need this since this is being asChilded onto a span.

packages/ui/src/primitives/field.tsx
```diff
@@ -40,31 +48,73 @@ export const Label = React.forwardRef<HTMLLabelElement, React.HTMLAttributes<HTM
...
       ref={forwardedRef}
       className={cn(
-        'block w-full bg-white text-gray-12 rounded-md bg-clip-padding py-1.5 px-2.5 border border-gray-a6 outline-none text-base',
-        'focus-visible:ring-[0.1875rem] disabled:opacity-50 disabled:cursor-not-allowed',
+        'flex w-full min-w-0 bg-white text-gray-12 rounded-md bg-clip-padding py-1.5 px-2.5 border outline-none text-base',
+        'disabled:opacity-50 disabled:cursor-not-allowed',
+        // type-specific overrides
+        props.type === 'otp'
+          ? 'min-h-10 text-[calc(var(--cl-font-size)*1.4)] font-semibold justify-center'
+          : 'min-h-8 justify-start',
```
alexcarpenter, line 76: 🤔 hm, mixed feelings on adding type specific logic like this within the input component. Since we are forwarding and input component onto a span. We need to make sure no input specific props are added in here to avoid invalid markup being added where used. Lets move forward as is for now.

packages/ui/src/common/OTPField.tsx
alexcarpenter: Screenshot 2024-06-07 at 9 58 51 AM

packages/ui/src/primitives/field.tsx
```diff
@@ -40,31 +48,73 @@ export const Label = React.forwardRef<HTMLLabelElement, React.HTMLAttributes<HTM
...
       data-field-input=''
       ref={forwardedRef}
       className={cn(
-        'block w-full bg-white text-gray-12 rounded-md bg-clip-padding py-1.5 px-2.5 border border-gray-a6 outline-none text-base',
-        'focus-visible:ring-[0.1875rem] disabled:opacity-50 disabled:cursor-not-allowed',
+        'flex w-full min-w-0 bg-white text-gray-12 rounded-md bg-clip-padding py-1.5 px-2.5 border outline-none text-base',
+        'disabled:opacity-50 disabled:cursor-not-allowed',
+        // type-specific overrides
+        props.type === 'otp'
+          ? 'min-h-10 text-[calc(var(--cl-font-size)*1.4)] font-semibold justify-center'
```
alexcarpenter, line 75: Looks like these were 40x40 and centered. Should we keep them sized this way or keep them fluid like this?

Screenshot 2024-06-07 at 9 58 51 AM

packages/ui/src/components/sign-in/sign-in.tsx
```diff
@@ -62,6 +63,7 @@ export function SignInComponent() {
                     <div className='space-y-4'>
                       <EmailField disabled={isGlobalLoading} />
                       <PasswordField disabled={isGlobalLoading} />
+                      <OTPField disabled={isGlobalLoading} />
```
alexcarpenter, line 66: When tabbing into the OTPInput, the card gets offset. Likely due to the overflow width added to the OTP input to allow space for the one password icons.


https://github.com/clerk/javascript/assets/825855/361a69d6-d972-4f0d-9c8f-0736372db14a

packages/ui/src/common/OTPField.tsx
joe-bell: In this case we're passing the `intent` down to the `Field` which will handle the appropriate error/success/warning styles

packages/ui/src/common/OTPField.tsx
joe-bell: We do in this case because it's how I'm setting the styles in the `Field.Input` component

packages/ui/src/common/OTPField.tsx
joe-bell: Could be misunderstanding but I thought I achieved this here?
https://github.com/clerk/javascript/pull/3535/files#r1632939660

packages/ui/src/primitives/field.tsx
joe-bell: I guess my thinking here is that I didn't want to create some arbitrary abstraction that could be misused, but totally get you on the props part. Wondering if there's a way to circumvent this prop application behaviour somehow?

packages/ui/src/common/OTPField.tsx
joe-bell: oh right or are you saying we should have a blinking cursor to mimic other inputs?

packages/ui/src/primitives/field.tsx
joe-bell: This way deffo looks better!

packages/ui/src/components/sign-in/sign-in.tsx
joe-bell: Should've fixed this 🤞

packages/ui/src/common/OTPField.tsx
joe-bell: Done!

packages/ui/src/common/OTPField.tsx
alexcarpenter: Can we name as `variant` or something else? Or if we're keen to keep as type make sure that type doesn't get added to the span?

joe-bell: Sure!

joe-bell: Done

alexcarpenter (approved):

PR #3538, opened 2024-06-07
feat(tanstack-start): Introduce @clerk/tanstack-start

octoper:
## Description

Introducing an experimental version of Clerk SDK for TanStack Start.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/tanstack-start/src/client/useAwaitableNavigate.ts
```diff
@@ -0,0 +1,31 @@
+import type { NavigateOptions } from '@tanstack/react-router';
+import { useLocation, useNavigate } from '@tanstack/react-router';
+import React from 'react';
+
+type Resolve = (value?: unknown) => void;
+
+export const useAwaitableNavigate = () => {
+  const navigate = useNavigate();
+  const location = useLocation();
+  const resolveFunctionsRef = React.useRef<Resolve[]>([]);
```
panteliselef, line 10: ❓@octoper I remember explictly not using refs for awaitable navigate for nextjs. Is this case different ?

octoper: I followed what we already have in the Remix SDK as it's similar

packages/tanstack-start/tsup.config.ts
```diff
@@ -0,0 +1,47 @@
...
+    esbuildPlugins: [esbuildPluginFilePathExtensions({ esmExtension: 'js' })],
+  };
+
+  const cjs: Options = {
+    ...common,
+    format: 'cjs',
+    outDir: './dist/cjs',
+    dts: true,
+    esbuildPlugins: [esbuildPluginFilePathExtensions({ esmExtension: 'cjs' })],
+  };
```
panteliselef, line 41: ❓(just curious mostly) Do we need to support CommonJS ?

octoper: We probably don't need it but I wanted to keep it for now as we can't use server functions yet and not sure if having only ESM causes any trouble in specific platforms.

nikosdouvlis: Given that this is a new package, we could drop cjs support completely and see if we get any requests to add it back

packages/tanstack-start/src/utils/index.ts
```diff
@@ -0,0 +1,9 @@
+import { buildErrorThrower } from '@clerk/shared';
+
+export const isClient = typeof window !== 'undefined';
```
nikosdouvlis, line 3: This looks fine as you rely on vite to instantiate the module once for the client and once for the server. Just curious - is there any other reason you prefer constants instead of functions here?

packages/tanstack-start/src/server/middlewareHandler.ts
```diff
@@ -0,0 +1,44 @@
...
+import { getResponseClerkState } from './utils';
+
+export type RequestHandler<TRouter extends AnyRouter> = (ctx: {
+  request: Request;
+  router: TRouter;
+  responseHeaders: Headers;
+}) => Promise<Response>;
+export type CustomizeRequestHandler<TRouter extends AnyRouter> = (cb: RequestHandler<TRouter>) => EventHandler;
+
+export function clerkMiddlewareHandler<TRouter extends AnyRouter>(eventHandler: CustomizeRequestHandler<TRouter>) {
```
nikosdouvlis, line 15: Could someone pass options to this one? For example, is it possible to set a PK/SK pair during runtime instead of simply relying on env vars?

packages/tanstack-start/README.md
```diff
@@ -0,0 +1,194 @@
...
+import { getRouterManifest } from '@tanstack/start/router-manifest';
+import { createRouter } from './router';
+import { clerkMiddlewareHandler } from '@clerk/tanstack-start/server';
+
+const handler = createRequestHandler({
+  createRouter,
+  getRouterManifest,
+});
+
+export default clerkMiddlewareHandler(handler)(defaultStreamHandler);
```
nikosdouvlis, line 137: Can you show an example of passing options to clerkMiddleware?

packages/tanstack-start/src/client/ClerkProvider.tsx
```diff
@@ -0,0 +1,94 @@
...
+      disabled: __telemetryDisabled,
+      debug: __telemetryDebug,
+    },
+  };
+
+  return (
+    <>
+      {/* TODO: revisit window.__clerk_init_state */}
+      <Asset tag='script'>{`window.__clerk_init_state = ${JSON.stringify(contextRouter?.clerkStateContext)}`}</Asset>
+      <ClerkOptionsProvider options={mergedProps}>
```
nikosdouvlis, line 69: Could you please explain why we need this context?

packages/tanstack-start/src/utils/index.ts
octoper: No reason, you are right this should be a function!

packages/tanstack-start/src/client/ClerkProvider.tsx
octoper: This is the initial state passed to the `ClerkProvider`, it's probably name confusion I will change it!

packages/tanstack-start/src/server/middlewareHandler.ts
octoper: I didn't want to introduce anything for now as we don't know if this is the final solution.

nikosdouvlis: Good thinking :) Im raising this because it could result in an public API change so lets see an example of passing options to the helper first. Im thinking that this:

```ts
const handler = createRequestHandler({
  createRouter,
  getRouterManifest,
});

export default clerkMiddlewareHandler(handler, { secretKey: '....' })(defaultStreamHandler);
```

might turn out to be a bit too complicated for us. If we want to accept options, we could maybe turn the helper into a factory function, ex:

```
const handler = createRequestHandler({
  createRouter,
  getRouterManifest,
});

const clerkHandler = createClerkHandler(handler, { secretKey: '...' })

export default clerkHandler(defaultStreamHandler);
```
or similar

octoper: Yeah it looks good, I will implement this today and see if we can go with a better solution

nikosdouvlis (approved):
Amazing job @octoper :)

PR #3540, opened 2024-06-07
feat(ui): Implement connection layout rendering

alexcarpenter:
## Description

Given the amount of enabled connections distribute the connection buttons evenly with a max of 6 per row.

Follows [`distributeStrategiesIntoRows`](https://github.com/clerk/javascript/blob/main/packages/clerk-js/src/ui/elements/utils.ts#L4) but uses flexbox as to not need to render in each row.

Screenshot 2024-06-07 at 5 21 08 PM

https://linear.app/clerk/issue/SDKI-24/implement-connection-layout-rendering

## TODO
- [ ] Conditionally show/hide label text based on enabled connections length.
- [ ] Move column count variable into CSS class
- [ ] Ensure height is preserved when icon only is shown

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/primitives/connection.tsx
```diff
@@ -3,20 +3,37 @@ import * as React from 'react';
...
+  const numRows = Math.ceil(items.length / maxCols);
+  const itemsPerRow = Math.ceil(items.length / numRows);
+  const rows: T[][] = Array.from({ length: numRows }, () => []);
+  let currentArrayIndex = 0;
+  for (const item of items) {
+    rows[currentArrayIndex].push(item);
+    if (rows[currentArrayIndex].length === itemsPerRow) {
+      currentArrayIndex++;
+    }
+  }
```
alexcarpenter, line 16: Likely can be simplified and just take in a count and give back a column value to be used.

packages/ui/src/primitives/connection.tsx
```diff
@@ -3,20 +3,36 @@ import * as React from 'react';
...
       ref={forwardedRef}
       {...props}
       className={cn(
-        '[--cl-connection-gap:theme(spacing.2)]',
-        '[--cl-connection-cols:2]',
-        'flex gap-[--cl-connection-gap] flex-wrap items-center justify-center',
+        'flex items-center justify-center flex-wrap [--cl-connection-gap:theme(spacing.2)] gap-y-[--cl-connection-gap] -mx-[calc(var(--cl-connection-gap)/2)]',
         className,
       )}
+      style={{ '--cl-connection-cols': cols } as React.CSSProperties}
```
alexcarpenter, line 35: Transformer currently doesn't account for template literals. Will follow up a fix.

kylemac: love that we're collecting different uses and making the transformer better each time 📈

packages/ui/src/primitives/connection.tsx
```diff
@@ -3,20 +3,36 @@ import * as React from 'react';
...
+  const cols = getColumnCount(React.Children.toArray(children));
   return (
     <div
       ref={forwardedRef}
       {...props}
       className={cn(
-        '[--cl-connection-gap:theme(spacing.2)]',
-        '[--cl-connection-cols:2]',
-        'flex gap-[--cl-connection-gap] flex-wrap items-center justify-center',
+        'flex items-center justify-center flex-wrap [--cl-connection-gap:theme(spacing.2)] gap-y-[--cl-connection-gap] -mx-[calc(var(--cl-connection-gap)/2)]',
```
joe-bell, line 32: Perhaps a nit, but wondering if we should just use:

```suggestion
        'flex items-center justify-center flex-wrap [--cl-connection-gap:theme(spacing.2)] -m-[calc(var(--cl-connection-gap)/2)]',
```

and then use `p-` instead of `px-` within the child

packages/ui/src/primitives/connection.tsx
alexcarpenter: Updated https://github.com/clerk/javascript/pull/3540/commits/f4d1e9c4c012555c595518d8cd8f3e109371b921

packages/ui/src/primitives/connection.tsx
alexcarpenter: Sorry I meant at the usage level like this: https://play.tailwindcss.com/KiPqalm1Xa

The wrapping card body has `space-y-6` and the `<Connection.Root />` is a direct child with the negative margin applied. If we include the negative vertical margin, they'd be conflicting with each other.

packages/ui/src/primitives/connection.tsx
joe-bell: @alexcarpenter any chance we could use this internally for `dashboard/new` as a follow-up task? Our current implementation is pretty rough (myself and Austin just hard-coded the conditions!)

packages/ui/src/primitives/connection.tsx
```diff
@@ -3,22 +3,31 @@ import * as React from 'react';
 
 import { Spinner } from './spinner';
 
-export const Root = React.forwardRef<HTMLDivElement, React.HTMLAttributes<HTMLDivElement>>(function Root(
-  { children, className, ...props },
-  forwardedRef,
-) {
+function getColumnCount(itemsLength: number, maxCols: number = 6): number {
```
joe-bell, line 6: cc @austincalvelage (@alexcarpenter solved it!)

joe-bell (approved):
Super nice work 🔥

packages/ui/src/primitives/connection.tsx
alexcarpenter: For sure, no reason you wouldn't be able to use.

packages/ui/src/primitives/connection.tsx
austincalvelage: Legend!

PR #3543, opened 2024-06-10
fix(backend,shared,clerk-js): Improve cookie suffixes

dimkl:
## Description

Improve the cookie suffixes from the publishableKey to a fixed length 8 char string (hashed substring of publishableKey) to reduce the cookie size.
This change will also reduce the handshake payload by ~ 120 bytes.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/tokens/authenticateContext.ts
+++ b/packages/backend/src/tokens/authenticateContext.ts
@@ -1,9 +1,9 @@
-import { parsePublishableKey } from '@clerk/shared/keys';
 import type { Jwt } from '@clerk/types';
 
 import { constants } from '../constants';
 import { decodeJwt } from '../jwt/verifyJwt';
 import { assertValidPublishableKey } from '../util/optionsAssertions';
+import { getCookieSuffix, getSuffixedCookieName, parsePublishableKey } from '../util/shared';
 import type { ClerkRequest } from './clerkRequest';
 import type { AuthenticateRequestOptions } from './types';
 
@@ -48,11 +48,11 @@ class AuthenticateContext {
     return this.sessionTokenInCookie || this.sessionTokenInHeader;
   }
 
-  private get cookieSuffix() {
-    return this.publishableKey?.split('_').pop();
-  }
-
-  public constructor(private clerkRequest: ClerkRequest, options: AuthenticateRequestOptions) {
+  public constructor(
+    private cookieSuffix: string,
+    private clerkRequest: ClerkRequest,
+    options: AuthenticateRequestOptions,
+  ) {
     // Even though the options are assigned to this later in this function
     // we set the publishableKey here because it is being used in cookies/headers/handshake-values
     // as part of getMultipleAppsCookie
@@ -124,7 +124,7 @@ class AuthenticateContext {
   }
 
   private getSuffixedCookie(name: string) {
-    return this.getCookie(`${name}_${this.cookieSuffix}`) || undefined;
+    return this.getCookie(getSuffixedCookieName(name, this.cookieSuffix)) || undefined;
   }
 
   private getSuffixedOrUnSuffixedCookie(cookieName: string) {
@@ -228,8 +228,10 @@ class AuthenticateContext {
 
 export type { AuthenticateContext };
 
-export const createAuthenticateContext = (
-  ...args: ConstructorParameters<typeof AuthenticateContext>
-): AuthenticateContext => {
-  return new AuthenticateContext(...args);
+export const createAuthenticateContext = async (
+  clerkRequest: ClerkRequest,
+  options: AuthenticateRequestOptions,
+): Promise<AuthenticateContext> => {
+  const cookieSuffix = options.publishableKey ? await getCookieSuffix(options.publishableKey) : '';
+  return new AuthenticateContext(cookieSuffix, clerkRequest, options);
 };
--- a/packages/clerk-js/src/core/auth/AuthCookieService.ts
+++ b/packages/clerk-js/src/core/auth/AuthCookieService.ts
@@ -1,5 +1,6 @@
 import { setDevBrowserJWTInURL } from '@clerk/shared/devBrowser';
 import { is4xxError, isClerkAPIResponseError, isNetworkError } from '@clerk/shared/error';
+import { getCookieSuffix } from '@clerk/shared/keys';
 import type { Clerk, EnvironmentResource } from '@clerk/types';
 
 import { clerkCoreErrorTokenRefreshFailed, clerkMissingDevBrowserJwt } from '../errors';
@@ -39,7 +40,12 @@ export class AuthCookieService {
   private sessionCookie: SessionCookieHandler;
   private devBrowser: DevBrowser;
 
-  constructor(private clerk: Clerk, fapiClient: FapiClient) {
+  public static async create(clerk: Clerk, fapiClient: FapiClient) {
+    const cookieSuffix = await getCookieSuffix(clerk.publishableKey);
+    return new AuthCookieService(clerk, fapiClient, cookieSuffix);
+  }
+
+  private constructor(private clerk: Clerk, fapiClient: FapiClient, cookieSuffix: string) {
     // set cookie on token update
     eventBus.on(events.TokenUpdate, ({ token }) => {
       this.updateSessionCookie(token && token.getRawString());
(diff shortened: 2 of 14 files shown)


packages/shared/src/__tests__/keys.test.ts
```diff
@@ -164,3 +165,19 @@ describe('isProductionFromSecretKey(key)', () => {
...
+    ['pk_test_Y2xlcmsuY2xlcmsuZGV2JA', 'QvfNY2dr'],
+  ];
+
+  test.each(cases)('given %p pk, returns %p cookie suffix', async (pk, expected) => {
+    expect(await getCookieSuffix(pk)).toEqual(expected);
+  });
+
+  test('omits special characters from the cookie suffix', async () => {
+    const pk = 'pk_test_ZW5vdWdoLWFscGFjYS04Mi5jbGVyay5hY2NvdW50cy5sY2xjbGVyay5jb20k';
+    expect(await getCookieSuffix(pk)).toEqual('jtYvytHq');
```
nikosdouvlis, line 181: Is there any scenario where we can end up with a suffix whose length is less than 8 chars? For example, if there's an edge case where we need to remove many special characters? 

I'm almost sure that there's no such case, but asking just to be safe.

packages/shared/jest.setup.ts
```diff
@@ -32,3 +36,10 @@ Object.defineProperty(global.window, 'navigator', {
   value: navigatorMock,
   writable: true,
 });
+
+// polyfill TextDecoder, TextEncoder for jsdom >= 16
+Object.assign(global, { TextDecoder, TextEncoder });
```
nikosdouvlis, line 41: 💯

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1504,7 +1505,8 @@ export class Clerk implements ClerkInterface {
   };
 
   #loadInStandardBrowser = async (): Promise<boolean> => {
-    this.#authService = new AuthCookieService(this, this.#fapiClient);
+    const cookieSuffix = await getCookieSuffix(this.publishableKey);
+    this.#authService = new AuthCookieService(this, this.#fapiClient, cookieSuffix);
```
nikosdouvlis, line 1509: Do you think we should deal with cookie suffixes inside `AuthCookieService`?
Does `Clerk` need to know about this? We don't seem to use it anywhere else here. `AuthCookieService` does hold a reference to `this`, so we can read the publishableKey from within. 

If you want to leverage DI for testing, maybe pass the `getCookieSuffix` function instead?

packages/clerk-js/src/core/auth/AuthCookieService.ts
```diff
@@ -39,7 +39,7 @@ export class AuthCookieService {
   private sessionCookie: SessionCookieHandler;
   private devBrowser: DevBrowser;
 
-  constructor(private clerk: Clerk, fapiClient: FapiClient) {
+  constructor(private clerk: Clerk, fapiClient: FapiClient, cookieSuffix: string) {
```
nikosdouvlis, line 42: See my comment below

packages/backend/src/tokens/authenticateContext.ts
```diff
@@ -228,8 +227,11 @@ class AuthenticateContext {
...
-export const createAuthenticateContext = (
-  ...args: ConstructorParameters<typeof AuthenticateContext>
-): AuthenticateContext => {
-  return new AuthenticateContext(...args);
+export const createAuthenticateContext = async (
+  clerkRequest: ClerkRequest,
+  options: AuthenticateRequestOptions,
+): Promise<AuthenticateContext> => {
+  const cookieSuffix = options.publishableKey ? await getCookieSuffix(options.publishableKey) : '';
+  const ctx = new AuthenticateContext(cookieSuffix, clerkRequest, options);
```
nikosdouvlis, line 235: Same comment about moving cookieSuffix inside AuthenticateContext, but I don't feel strongly about this.

nikosdouvlis, line 235:
```suggestion
return new AuthenticateContext(cookieSuffix, clerkRequest, options);
```

nikosdouvlis (approved):

packages/shared/src/__tests__/keys.test.ts
dimkl: For such case to exist, we would have to create a hash digest with more than 32 (out of 40) bytes with a value that will be encoded to special char.
I will search for the sha1 RFC and try to identify if and when would that be possible

packages/clerk-js/src/core/clerk.ts
dimkl: The reason i am passing the cookieSuffix value in `AuthCookieService` is because we cannot initiliaze the cookieSuffix value in the AuthCookieService contrustor where the cookie handlers are initialized and use the cookie suffix since the generation of the cookie suffix is async due to the crypto sha1 addition.

packages/backend/src/tokens/authenticateContext.ts
dimkl: closing as it's the same response to the above. Let's discuss if we want to use another approach and i will apply i to all your related comments.

packages/clerk-js/src/core/clerk.ts
nikosdouvlis: You're right! The alternative here would be to defer the init logic into an `init` method instead, something like 

```
this.#authService = new AuthCookieService(this, this.#fapiClient);
await this.#authService.init()
```

Would that be better for our case? Its a minor difference, but the singleton would have one less dependency to worry about.

packages/shared/src/__tests__/keys.test.ts
nikosdouvlis: Great 💯 
Resolving this one as it shouldn't block you. Its very unlikely that this could happen, but if you find anything just add another comment here.

packages/clerk-js/src/core/clerk.ts
dimkl: I am not against that approach. The risky part in this implementation is that we should be very careful to call the `init()` method exactly after the service initialization or enforce somehow (eg throwing an error) that the `init()` method is called before the authService is used.
If we decide to use the approach of introducing a method to initialize the authService instead of depending in the constructor, i would suggest we expose a static method to initialize the instance and make the constructor private.
cc: @nikosdouvlis

nikosdouvlis: Agreed, I think its better in this case. We can always revise this decision later

dimkl: Implementation updated to use private constructor and static method to initialize the AuthCookieService.
Reasons behind updating implementation as suggested by Nikos:
- reduce dependencies in core/clerk.ts
- AuthCookieService contains all logic for cookie and cookie suffixes
- a way to bypass the limitation of un-supported async constructors

BRKalow (approved):

packages/shared/src/keys.ts
```diff
@@ -110,10 +110,14 @@ export function isProductionFromSecretKey(apiKey: string): boolean {
...
 
-export const getCookieSuffix = (publishableKey: string): string => {
-  return publishableKey.split('_').pop() || '';
-};
+export async function getCookieSuffix(publishableKey: string): Promise<string> {
+  const data = new TextEncoder().encode(publishableKey);
+  const digest = await globalThis.crypto.subtle.digest('sha-1', data);
+  const stringDigest = String.fromCharCode(...new Uint8Array(digest));
+  // Base 64 Encoding with URL and Filename Safe Alphabet: https://datatracker.ietf.org/doc/html/rfc4648#section-5
+  return isomorphicBtoa(stringDigest).replace(/\+/gi, '-').replace(/\//gi, '_').substring(0, 8);
```
nikosdouvlis, line 118: This could result in cookie names that look like `__session_a3f_bd3f`
This is not going to play well with `split('_')` from the base branch because:

```
"__session_asdfasdf".split('_').pop();
> 'asdfasdf'
"__session_asdfa_sdf".split('_').pop();
> 'sdf'
```

In your research, did you hit any case where we can end up with cookie suffixes less than 8 characters long?
If not, I suggest we stick with the initial proposal of simply stripping `+`, `/` and `=`. Or we can replace them with an alphanumeric char like `0`. This is going to lower entropy but just marginally so we shouldn't worry about it

dimkl: As discussed online, since we don't use the `.split('_')` and if choose to use a delimiter we would prefer a symbol outside of the currently used in cookie names (eg `:` because `_` is used as word separator in cookie names) we will keep the url-safe base64 solution.
cc: @nikosdouvlis

nikosdouvlis: Agreed, marking this as resolved :)

PR #3518, opened 2024-06-05
fix(nextjs): Replace router.refresh() with cookies().delete()

nikosdouvlis:
## Description

In this PR we replace `router.refresh` with a server action that calls `cookies().delete`. According to the [NextJS docs](https://nextjs.org/docs/app/building-your-application/caching), using the `cookies().delete` API invalidates the FE client cache, similar to what `router.refresh` does.

The main difference between the two is that `router.refresh` will then attempt to fetch and re-mount the current page. This causes our components to unmount and mount again, losing state before `setActive` completes. This usually leads to navigation/redirect issues, issues completing oauth flows etc. 

Caveat: 
We can only use server action directly for next@14 and above, as server actions were an experimental feature in next@13.
For next@13 we still call `router.refresh`. Users that are on next@13 should upgrade to at least `next@14` if they notice any of the issues described above.

## Checklist

- [x] `npm test` runs as expected.
- [x] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #3544, opened 2024-06-10
fix(clerk-js): Deprecate `afterSignOutUrl` from UserButton

panteliselef:
## Description

Fixes: DASH-183

This PR address the following issues:
- In a single session app, user opens user profile through the user button, the developer has set `afterSignOutUrl` but when user deletes their account, the provided `afterSignOutUrl` is not respected.
- In a multi session app, user goes to user profile, developer has set `afterMultiSessionSingleSignOutUrl` in user button but the user when they delete one of their accounts is not redirected to the appropriate url defined as prop.

The main issue is that we don't propagate `afterSignOutUrl` and `afterMultiSessionSingleSignOutUrl` from the UserButton to UserProfile.

This PR addresses that by positioning those props at the root lever (ClerkProvider or Clerk.load). We considered simply introducing those props to UserProfile, but the naming would not be ideal. We also considered renaming them, but at that point  you want the components to always be in sync, so we positioned them at the root.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/components/UserProfile/DeleteUserForm.tsx
```diff
@@ -36,7 +38,12 @@ export const DeleteUserForm = withCardStateProvider((props: DeleteUserFormProps)
       }
 
       await user.delete();
-      await navigateAfterSignOut();
+
+      // TODO: Investigate if we need to call `setActive` with {session: null}
+      if (otherSessions.length === 0) {
+        return navigateAfterSignOut();
+      }
+      await navigateAfterMultiSessionSingleSignOutUrl();
```
BRKalow, line 46: 💭  I wouldn't necessarily expect that this method would need to be aware of multi-session. I understand why it was done this way as part of this PR, but it feels like a better internal abstraction might be to consolidate the multi-session logic into `navigateAfterSignOut()`.

packages/clerk-js/src/ui/components/UserProfile/DeleteUserForm.tsx
```diff
@@ -36,7 +38,12 @@ export const DeleteUserForm = withCardStateProvider((props: DeleteUserFormProps)
       }
 
       await user.delete();
-      await navigateAfterSignOut();
+
+      // TODO: Investigate if we need to call `setActive` with {session: null}
```
BRKalow, line 42: Do we need to do this investigation before merge?

panteliselef: I was waiting for #3518 to be merged, and i will check again, but simply calling setActive would not refresh the router.

packages/clerk-js/src/ui/components/UserProfile/DeleteUserForm.tsx
panteliselef: I think creating an abstraction here would make our lifer harder, we have `useMultisessionActions` which needs to know the difference between single session sign out and multisession sign out. We can refactor but i don't see a clear benefit here.

packages/clerk-js/src/ui/components/UserProfile/DeleteUserForm.tsx
panteliselef: So i investigated further, even with #3518  the issue persisted. I've identified this as a FAPI issue, because after `user.delete()` client_uat comes back with a valid timestamp instead of being `0`. This will need a separate ticket to be address and should not be part of this PR

packages/clerk-js/src/ui/components/UserProfile/DeleteUserForm.tsx
BRKalow: Benefit would be not having to spread some of this logic across many different methods. That's fine to leave it for now, let's jut keep an eye on these code paths and we can abstract if it feels like multi-session logic is too difficult to manage.

.changeset/selfish-ladybugs-smile.md
```diff
@@ -0,0 +1,6 @@
+---
+'@clerk/clerk-js': minor
+'@clerk/types': minor
+---
+
+Deprecate `afterSignOutUrl` from UserButton.
```
BRKalow, line 6: Can you specify what actions a user needs to take here w.r.t. the deprecation? What do they also need to change in their code?

BRKalow (approved):
Left a comment on the changelog, but otherwise looks good.

anagstef (approved):
LGTM 💯

.changeset/selfish-ladybugs-smile.md
panteliselef: makes sense, i updated it

PR #3548, opened 2024-06-11
feat(ui): Extract `lineHeights`

joe-bell:
## Description

Following up on #3535 and chats with @alexcarpenter; this PR extracts `lineHeight` for self/external reference

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/common/OTPField.tsx
```diff
@@ -52,7 +52,7 @@ export function OTPField({
                           data-otp-field-cursor=''
                           // Note:
                           // Opportunity to extract lineHeight to theme
-                          className='motion-safe:animate-blink self-center rounded-full bg-current w-[2px] h-[calc(theme(fontSize.base)*1.38462)]'
+                          className='motion-safe:animate-blink self-center rounded-full bg-current w-[2px] h-[calc(theme(fontSize.base)*theme(lineHeight.base))]'
```
joe-bell, line 55: This is what I was hoping to do in #3535 @alexcarpenter (during our tuple)

packages/ui/src/tailwind.config.ts
```diff
@@ -61,11 +61,14 @@ const config = {
...
-        small: '1.38462',
-        medium: '1.41176',
-        large: '1.45455',
+        DEFAULT: '1.38462',
+        base: '1.38462',
+        xs: '1.38462',
+        sm: '1.38462',
+        md: '1.38462',
+        lg: '1.38462',
+        xl: '1.38462',
```
alexcarpenter, line 70: Should these be the values that were previously here. For example you removed the medium line height that was being used on the card title component which was intentional.

The font sizes would all just default to `theme('lineHeight.DEFAULT')`? and allow for overriding via `leading-md` in usage.

joe-bell: I swapped it out for the sake of consistency to our tokens (I think that `md` was only being used in one place which lead to my confusion)

Open to adding a `leading-[]` override, but it kinda felt like a negligible difference for me to opt for consistency

alexcarpenter: Got it, these were pulled from https://github.com/clerk/javascript/blob/da0f50725f91dda0cbbef123ebf3686cb1426354/packages/clerk-js/src/ui/foundations/typography.ts#L8-L14

Seems like will want to preserve these so we can make use of them vs needing to do arbitrary values, wdyt?

alexcarpenter: I am not 100% following what you mean. My thinking was:

```js
fontSize: ({ theme }) => ({
  DEFAULT: ['var(--cl-font-size)', theme('lineHeight.DEFAULT')],
  base: ['var(--cl-font-size)', theme('lineHeight.DEFAULT')],
  xs: ['calc(var(--cl-font-size) * 0.8)', theme('lineHeight.DEFAULT')],
  sm: ['calc(var(--cl-font-size) * 0.9)', theme('lineHeight.DEFAULT')],
  md: ['var(--cl-font-size)', theme('lineHeight.DEFAULT')],
  lg: ['calc(var(--cl-font-size) * 1.3)', theme('lineHeight.DEFAULT')],
  xl: ['calc(var(--cl-font-size) * 1.85)', theme('lineHeight.DEFAULT')],
  'icon-sm': ['calc(var(--cl-font-size) * 1.23)', '0'],
}),
lineHeight: {
  DEFAULT: '1.38462',
  normal: 'normal',
  xs: '1.33333',
  sm: '1.38462',
  md: '1.41176',
  lg: '1.45455',
},
```

Then default line height would be used but still allow `leading` overrides. So for the card title it'd be `className="text-lg leading-md"`.

Are you suggesting it should be `className="text-lg leading-lg"`?

joe-bell: My thinking here is that within `fontSize` we'll still reference the respective `lineHeight` key – as that feels like a more common default for me compared to using the same lineHeight for all 

I think what got me stumped here is that they're not named the same in our current implementation (`sm` **and** `small`)

joe-bell: I'm going to park this and come back to it

PR #3557, opened 2024-06-12
feat(elements): Infer base path

LekoArts:
## Description

The `path` prop on the `<SignIn.Root>` and `<SignUp.Root>` component is now automatically inferred. Previously, the default values were `/sign-in` and `/sign-up`, on other routes you had to explicitly define your route.

The new heuristic for determining the path where `<SignIn.Root>` and `<SignUp.Root>` are mounted is:

1. `path` prop
2. Automatically inferred
3. If it can't be inferred, fallback to `CLERK_SIGN_IN_URL` and `CLERK_SIGN_UP_URL` env var
4. Fallback to `/sign-in` and `/sign-up`

Fixes SDK-1791

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/elements/src/react/sign-in/root.tsx
```diff
@@ -62,22 +63,29 @@ function SignInFlowProvider({ children, exampleMode }: SignInFlowProviderProps)
...
+  /**
+   * Fallback markup to render while Clerk is loading
+   */
   fallback?: React.ReactNode;
   /**
-   * The base path for your sign-in route. Defaults to `/sign-in`.
-   *
-   * TODO: re-use usePathnameWithoutCatchAll from the next SDK
+   * The base path for your sign-in route.
+   * Will be automatically inferred in Next.js.
```
LekoArts, line 72: In the future we can adjust this line e.g. with Remix once we support this in other frameworks, too

packages/elements/src/react/sign-in/root.tsx
```diff
@@ -89,18 +97,21 @@ export type SignInRootProps = SignInFlowProviderProps & {
  */
 export function SignInRoot({
   children,
-  exampleMode,
+  exampleMode = false,
   fallback = null,
-  path = SIGN_IN_DEFAULT_BASE_PATH,
-  routing,
+  path: pathProp,
+  routing = ROUTING.path,
```
LekoArts, line 103: Made the falsy `exampleMode` and `routing` props more explicit by setting the default values they'd be inferred to anyways

packages/elements/src/react/utils/path-inference/next.tsx
```diff
@@ -0,0 +1,61 @@
+import { useRouter } from 'next/compat/router';
+import { useParams, usePathname } from 'next/navigation';
```
LekoArts, line 2: So in `packages/nextjs/src/client-boundary/hooks/usePathnameWithoutCatchAll.tsx` we `require` these two hooks since the inline comment there says that they'd error out in pages router.

I found this papertrail:

1. https://github.com/vercel/next.js/discussions/62880#discussioncomment-9730647
2. https://github.com/vercel/next.js/issues/54242
3. https://github.com/vercel/next.js/pull/55280
4. https://github.com/vercel/next.js/blob/4398e348ee724d88282881137311680ee60fb23c/packages/next/navigation-types/compat/navigation.d.ts#L13-L20

So I think we're good with importing it directly since the version that got introduced is below our minimum Next.js version

packages/elements/tsup.config.ts
```diff
@@ -21,7 +21,7 @@ export default defineConfig(overrideOptions => {
       'react/sign-in/index': 'src/react/sign-in/index.ts',
       'react/sign-up/index': 'src/react/sign-up/index.ts',
     },
-    external: ['react', 'react-dom', '@statelyai/inspect'],
+    external: ['react', 'react-dom', 'next', '@statelyai/inspect'],
```
LekoArts, line 24: While https://tsup.egoist.dev/#excluding-packages explains

> but dependencies and peerDependencies in your package.json are always excluded

I guess it doesn't hurt explicitly adding it here if we have `react` and `react-dom` here already

packages/elements/src/react/utils/path-inference/next.tsx
panteliselef: hmm, does that mean we can update the one within the nextjs package as well ?

tmilewski (approved):

PR #3570, opened 2024-06-13
fix(repo): Correct issues with NPM deduping

tmilewski:
## Description

Discovered when running `make up/fe`.

- Turborepo interactive UI wasn't playing nicely, and thus is now disabled.
- NPM was deduping incorrectly causing a nested dependency to resolve to a breaking version.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


package.json
```diff
@@ -94,6 +94,7 @@
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "rimraf": "^5.0.5",
+    "statuses": "^1.4.0",
```
panteliselef, line 97: Is this necessary?

panteliselef (approved):

package.json
tmilewski: Without, it doesn't dedupe correctly.

PR #3585, opened 2024-06-18
feat(tailwindcss-transformer): Add support for cx, cva and clsx

joe-bell:
## Description

Currently, classes within a `cva`/`cx`/`clsx` function don't get transformed in the same way as `cn`; this PR aims to fix that

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/tailwindcss-transformer/src/__tests__/__snapshots__/transform-cva.test.ts.snap
```diff
@@ -0,0 +1,41 @@
...
+        {
+          intent: "cl-986a1b71",
+          size: "cl-c082456a",
+          className: "cl-d2cf63c7"
+        },
+      ],
+      defaultVariants: {
+        intent: "cl-986a1b71",
+        size: "cl-c082456a",
+      },
```
bradlc, line 31: These shouldn't be transformed 🙈 – I wonder if we can special-case `cva` and only transform certain keys (i.e. `base`, `variants` in this case)

joe-bell: Oh shoot good catch; I completely missed that bit

packages/tailwindcss-transformer/src/index.ts
```diff
@@ -73,12 +73,37 @@ export function transform(code: string, ctx: { styleCache: StyleCache }) {
...
-      if (node.callee.type === 'Identifier' && node.callee.name === 'cn') {
+      // `className` concatenation functions
+      if (node.callee.type === 'Identifier' && ['cn', 'cx', 'clsx'].includes(node.callee.name)) {
         visitNode(node, ctx);
       }
+      // cva functions (note: only compatible with cva@1.0)
+      if (
+        node.callee.type === 'Identifier' &&
+        node.callee.name === 'cva' &&
+        node.arguments[0].type === 'ObjectExpression'
```
bradlc, line 95: Although in practice you wouldn't call `cva` without arguments, I'm thinking to be safe we should use optional chaining here to prevent a crash if you do `cva()`:

```suggestion
        node.arguments[0]?.type === 'ObjectExpression'
```

bradlc (approved):

joe-bell:
Blocked by https://github.com/clerk/javascript/pull/3586

PR #3599, opened 2024-06-19
test(clerk-js): Upgrade @testing-library packages

panteliselef:
## Description

This PR addresses the following deprecation warning in our unit tests

image

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/elements/__tests__/PlainInput.test.tsx
```diff
@@ -120,7 +120,7 @@ describe('PlainInput', () => {
     expect(getByRole('link', { name: /take action/i })).toBeInTheDocument();
     expect(getByRole('link', { name: /take action/i })).not.toHaveAttribute('rel');
     expect(getByRole('link', { name: /take action/i })).not.toHaveAttribute('target');
-    expect(getByRole('link', { name: /take action/i })).toHaveAttribute('href', '');
+    expect(getByRole('link', { name: /take action/i })).toHaveAttribute('href', '#');
```
panteliselef, line 123: Seems like the upgraded package will not match the `link` role to an anchor tag with an empty href attribute.
A quick test verified that if we use `#` as the href value, the tests will pass and the intended usage of `a` will not change within our component.

Feels like this anchor tag needs to be replaced with a button, but i can see how this could be considered a breaking change.

panteliselef: Alternatively we can change the tests to something like
```tsx
expect(getByText(/take action/i, { selector: 'a' })).toBeInTheDocument();
```

LekoArts: `href="#"` typically scrolls the page to the top, that could be an unintended side-effect of this change.

I also noticed while working on the E2E tests for Elements that this shouldn't be a `<link>` but a `<button>`. Can you add a `TODO(v6)` to make this change?

dimkl (commented):
❓ Does this deprecation warning exist in our minimum supported react version or is it introduced in the latest React version? Also is the fix applicable in our minimum supported version?

.changeset/cold-singers-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/clerk-js": minor
+---
+
+Change the value of href of field action achor tags from empty to `#` in order for it to be.
```
anagstef, line 5: ❓ Something is missing from the end of the sentence? 🤔

packages/clerk-js/src/ui/elements/__tests__/PlainInput.test.tsx
panteliselef: Yeah, i agree that this should be a button.

panteliselef:
@dimkl the latest `@testing-library/react` requires `^18.0.0` which is the same as the minimum supported react version we also support for both clerk-js and clerk-react.

packages/clerk-js/src/ui/elements/__tests__/PlainInput.test.tsx
panteliselef: I think i will proceed with just updating the tests then.

.changeset/cold-singers-suffer.md
panteliselef: I've going to completely remove this, because i'm reverting that change.

LekoArts (approved):

anagstef (approved):

PR #3614, opened 2024-06-21
feat(ui): Extract some logic into reusable hooks

alexcarpenter:
## Description

Introduce three new hooks based on current usage across `<SignIn />` `<SignUp />` components:

- `useLocalization` - access both `t` and `translateError` functions throughout components
- `useAttributes` - access field attributes such as `enabled` or `required`
- `useDisplayConfig` - access display config settings such as `applicationName`, `homeUrl`, `logoImageUrl`

Currently I am just updating the usage within the sign-in/up flows. Likely in a follow up PR we may refactor the common components to access this data internally using these hooks vs needs to pass them in as props within the sign-in/up components.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/ui/src/hooks/use-attributes.ts
+++ b/packages/ui/src/hooks/use-attributes.ts
@@ -1,8 +1,9 @@
 import { useClerk } from '@clerk/clerk-react';
-import type { Attribute, EnvironmentResource } from '@clerk/types';
+import type { Attribute, AttributeData, EnvironmentResource } from '@clerk/types';
 
-export function useAttributes(attribute?: Attribute) {
+export function useAttributes(attribute: Attribute): AttributeData {
   const clerk = useClerk();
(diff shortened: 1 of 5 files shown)


tmilewski (approved):

packages/ui/src/hooks/use-attributes.ts
```diff
@@ -0,0 +1,9 @@
+import { useClerk } from '@clerk/clerk-react';
+import type { Attribute, AttributeData, EnvironmentResource } from '@clerk/types';
+
+export function useAttributes(attribute: Attribute): AttributeData {
+  const clerk = useClerk();
+  const userSettingsAttributes = ((clerk as any)?.__unstable__environment as EnvironmentResource)?.userSettings
```
alexcarpenter, line 6: Open to feedback on how to properly type these as to not need to use any.

PR #3615, opened 2024-06-21
refactor(ui): Move logic into email-field

alexcarpenter:
## Description

Move rendering and attribute fetching logic into the `<EmailField />` component.

Instead of needing to pass the fetched data from the `<SignUp />` component, move as much rendering logic into the component while offering escape hatches via props. This is to make it easier to implement fields across both SignIn/Up and reduce inconsistencies that may arise with needing to manually provide optional props across steps and flows.

Usage update:

```diff
- {emailAddressEnabled ? (
-   <EmailField
-     label={t('formFieldLabel__emailAddress')}
-     hintText={t('formFieldHintText__optional')}
-     required={emailAddressRequired}
-     disabled={isGlobalLoading}
-     error={translateError}
-   />
- ) : null}
+ <EmailField disabled={isGlobalLoading} />
```

Each prop can be manually overridden at usage level if needed.

```tsx
<EmailField
  label="Custom label"
  hintText="Custom hint text"
  enabled={true}
  required={false}
  disabled={isGlobalLoading}
/>
```

Pending feedback will also update the following fields with similar changes.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/common/email-field.tsx
```diff
@@ -50,7 +70,7 @@ export function EmailField({
         </Common.FieldState>
         <Common.FieldError asChild>
           {({ message, code }) => {
-            return <Field.Message intent='error'>{error ? error(message, code, ERROR_NAME) : message}</Field.Message>;
+            return <Field.Message intent='error'>{translateError(message, code, DEFAULT_ERROR_NAME)}</Field.Message>;
```
alexcarpenter, line 73: Can easily enable a custom translateError prop if/when its needed in the future.

tmilewski (approved):
💯, seems to be some minor type issues though

packages/ui/src/common/email-field.tsx
alexcarpenter: add for backwards compat for the time being https://github.com/clerk/javascript/pull/3615/commits/476a0656df6f4a43cb450b5fac79356c0d4b5dac

PR #3617, opened 2024-06-21
feat(nextjs): Read dynamic keys from `clerkClient` within middleware runtime 

LauraBeatris:
## Description

Resolves SDK-1253, SDK-1811, SDK-1770

PR on top of https://github.com/clerk/javascript/pull/3525 to facilitate review on `clerkClient` changes. 

Refactoring `clerkClient` to be called as a function, instead of singleton, in order to read request data. It reads based on the following:
- If `clerkClient` gets accessed on middleware runtime, then use ALS store to access dynamic keys data 
- If `clerkClient` gets accessed on application server, then use `NextRequest` to access dynamic keys data 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/nextjs/src/server/clerkClient.ts
```diff
@@ -27,6 +33,63 @@ const clerkClient = createClerkClient({
...
+    return createClerkClientWithOptions({
+      secretKey: requestData.secretKey,
+      publishableKey: requestData.publishableKey,
+    });
+  }
+
+  return clerkClient;
+};
+
+const clerkClient: ClerkClient & typeof clerkClientForRequest = Object.assign(
```
LauraBeatris, line 89: Must update docs to call `clerkClient()` instead of `clerkClient` https://clerk.com/docs/references/nextjs/route-handlers#interact-with-clerks-backend-api

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -91,82 +91,88 @@ export const clerkMiddleware: ClerkMiddleware = withLogger('clerkMiddleware', lo
...
+      logger.debug('auth', () => ({ auth: authObject, debug: authObject.debug() }));
+
+      const redirectToSignIn = createMiddlewareRedirectToSignIn(clerkRequest);
+      const protect = createMiddlewareProtect(clerkRequest, authObject, redirectToSignIn);
+      const authObjWithMethods: ClerkMiddlewareAuthObject = Object.assign(authObject, { protect, redirectToSignIn });
+
+      let handlerResult: Response = NextResponse.next();
+      try {
+        handlerResult =
+          (await clerkClientStorage.run(
```
LauraBeatris, line 136: It was necessary to place another ALS context wrapped around the returned handler, otherwise `clerkClient` calls within the middleware handler (👇🏻 ) wouldn't have access to the store:
```ts
export default clerkMiddleware((auth)=> {
  clerkClient().users().get("...")
})
```

packages/nextjs/src/server/clerkClient.ts
```diff
@@ -27,6 +34,63 @@ const clerkClient = createClerkClient({
...
+const createClerkClientWithOptions: typeof createClerkClient = options =>
+  createClerkClient({ ...clerkClientDefaultOptions, ...options });
+
+const clerkClientSingleton = createClerkClient(clerkClientDefaultOptions);
+
+/**
+ * @deprecated
+ * This singleton is deprecated and will be removed in a future release. Please use `clerkClient()` as a function instead.
+ */
+const clerkClientSingletonProxy = new Proxy(clerkClientSingleton, {
```
LauraBeatris, line 50: Calling out for JS help 🙇🏻 We want to log a deprecation warning only if `clerkClient` gets called as a singleton. 

The `Proxy` is being placed around the singleton reference here, but in line 90, it's properties are being merged with `clerkClient()` on `Object.assign`, therefore the deprecation warnings is being logged for both cases. 

It's important to keep exporting `clerkClient` and allow it to be either a singleton or a function but to do so, it requires that `Object.assign` bit - is there another way to do so while performing a proxy for the deprecation warning?

packages/nextjs/src/server/clerkClient.ts
```diff
@@ -27,6 +34,63 @@ const clerkClient = createClerkClient({
...
+  get(target, prop, receiver) {
+    deprecated('clerkClient object', 'Use `clerkClient()` as a function instead.');
+
+    return Reflect.get(target, prop, receiver);
+  },
```
BRKalow, line 55: How does this work with nested properties? E.g. if I access `clerkClient.users.getUser()` will this still log the deprecation warning?

packages/nextjs/src/server/clerkClient.ts
BRKalow: Do you mean that the usage of `Object.assign()` is causing the deprecation warning to be logged? Or because we're using the proxied client object under the hood for the non-dynamic keys case, it logs for both:
```
clerkClient().users.getUser()

clerkClient.users.getUser()
```

packages/nextjs/src/server/clerkClient.ts
```diff
@@ -27,6 +34,63 @@ const clerkClient = createClerkClient({
...
 });
 
+/**
+ * Constructs a BAPI client that accesses request data within the runtime.
+ * Necessary if middleware dynamic keys are used.
```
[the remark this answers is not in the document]
BRKalow: Using the public-facing name in the code can help others connect the code to the product feature, so I think it's okay 👍

packages/nextjs/src/server/clerkClient.ts
LauraBeatris: It's logging for both cases, and I think this is due to `Object.assign` since it's executing `Proxy.get` for every single property.

integration/tests/dynamic-keys.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    await app.setup();
+    await app.withEnv(appConfigs.envs.withEmailCodes);
+    await app.dev();
+  });
+
+  test.afterAll(async () => {
+    await app.teardown();
+  });
+
+  // test('redirects to `signInUrl` on `auth.protect`')
```
LauraBeatris, line 71: Still have to implement those tests here, leaving this as a WIP, will continue to work today/tomorrow!

integration/tests/handshake.test.ts
```diff
@@ -8,7 +8,7 @@ import { generateConfig, getJwksFromSecretKey } from '../testUtils/handshake';
 
 const PORT = 4199;
 
-test.skip('Client handshake @generic', () => {
+test.describe('Client handshake @generic', () => {
```
LauraBeatris, line 11: @nikosdouvlis I've went ahead and removed the `skip` from handshake tests here.

PR #3620, opened 2024-06-21
fix(repo): Bump Node to latest version

dstaley:
## Description

This PR updates our Verdaccio process to unblock the upgrade to Node v18.20.3, which includes npm v10.7.0.

In modern versions of npm, you need to have an `authToken` in your `.npmrc` to be able to run `npm publish`. We achieved this previously with the use of the packages `npm-cli-adduser` and `npm-cli-login`. These packages were last updated six years ago. npm v10.7.0 included a rework of the CLI output, which breaks those packages since they relied on a specific output structure. We can achieve the same end result as those packages by simply adding an `_authToken` entry to `.npmrc` manually. Our Verdaccio configuration doesn't rely on the `_authToken` being any specific value due to the use of `$all`, which [allows all users, both anonymous and logged in, to publish packages](https://verdaccio.org/docs/authentication#understanding-groups).

As part of this PR, I've removed all the `TODO` comments and temporary version pins. We now default to the latest version of Node v18, which is 18.20.3 as of writing.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] other: repo hygiene

### DIFF
--- a/.github/actions/init/action.yml
+++ b/.github/actions/init/action.yml
@@ -4,9 +4,7 @@ inputs:
   node-version:
     description: 'The node version to use'
     required: false
-    # This is pinned to `18.20.2` because `18.20.3` bumps npm from `10.5.0` to `10.7.0`, which causes our verdaccio setup to fail (see SDK-1793).
-    # TODO: Update workflows/ci.yml when this isn't pinned anymore.
-    default: '18.20.2'
+    default: '18'
   playwright-enabled:
     description: 'Enable Playwright?'
     required: false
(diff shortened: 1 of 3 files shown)


tmilewski (approved):

dimkl (commented):
Nice work and PR description!!
Could we also remove the `npm-cli-adduser` and `npm-cli-login` packages from our dependencies since we no longer need them?

PR #3628, opened 2024-06-27
fix(clerk-js): Clear session after deleting account

octoper:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


octoper:
!preview

clerk-cookie:

| :-- | :-- | :-- |
| :cook: _Deploying..._ | [Inspect](https://javascript-preview-4gmrtq2ol.clerkstage.dev) | Jun 27, 2024 08:10 PM |

octoper:
!preview

packages/clerk-js/src/core/clerk.ts
```diff
@@ -732,7 +732,10 @@ export class Clerk implements ClerkInterface {
     }
 
     // getToken syncs __session and __client_uat to cookies using events.TokenUpdate dispatched event.
-    await newSession?.getToken();
+    const token = await newSession?.getToken();
+    if (!token) {
+      eventBus.dispatch(events.TokenUpdate, { token: null });
+    }
```
panteliselef, line 738: Don't we need to broadcast the sign out as well ?
```suggestion
    if (!token) {
      this.#broadcastSignOutEvent();
      eventBus.dispatch(events.TokenUpdate, { token: null });
    }
```

packages/clerk-js/src/ui/components/UserProfile/DeleteUserForm.tsx
```diff
@@ -35,8 +36,13 @@ export const DeleteUserForm = withCardStateProvider((props: DeleteUserFormProps)
...
       }
 
+      // Store the session so it can be ended after the user is deleted
+      const session = client.lastActiveSessionId;
       await user.delete();
-      await navigateAfterSignOut();
+      await setActive({
+        session,
+        beforeEmit: navigateAfterSignOut,
+      });
```
panteliselef, line 45: ❓Why don't we do 
```suggestion
      await setActive({
        session: null,
        beforeEmit: navigateAfterSignOut,
      });
```
and update the code in clerk.ts to be ?
```tsx
    const shouldSignOutSession = newSession === null;
    if (shouldSignOutSession) {
      this.#broadcastSignOutEvent();
      eventBus.dispatch(events.TokenUpdate, { token: null });
    }
```

panteliselef (commented):
💯 Seems like we were working on the same thing, sorry for not communicating it. I asked some questions mainly because i went with a different direction

packages/clerk-js/src/core/clerk.ts
octoper: This was a recent deletion we did so I brought it back, nice observation not 100% sure that we should broadcast the sign out as we mostly dispatching the `TokenUpdate` event to keep cookies clear if there is no token at that point

packages/clerk-js/src/ui/components/UserProfile/DeleteUserForm.tsx
octoper: I explicitly added the session just to handle multi-sessions

octoper:
!preview

octoper:
!preview

octoper:
!preview

packages/clerk-js/src/ui/components/UserProfile/DeleteUserForm.tsx
```diff
@@ -37,13 +38,16 @@ export const DeleteUserForm = withCardStateProvider((props: DeleteUserFormProps)
...
-
       // TODO: Investigate if we need to call `setActive` with {session: null}
-      if (otherSessions.length === 0) {
-        return navigateAfterSignOut();
-      }
-      await navigateAfterMultiSessionSingleSignOutUrl();
+      const navigationCallback =
+        otherSessions.length === 0 ? navigateAfterSignOut : navigateAfterMultiSessionSingleSignOutUrl;
+      return await setActive({
+        session,
```
dimkl, line 48: ⛔ We should not use the previous session since it's (probably) already deleted. We should pass `{ session: null }`

packages/clerk-js/src/core/clerk.ts
dimkl: I would expect to keep this without the `#broadcastSignOutEvent();` and probably add to this event a listener in clerk-js that call the `#broadcastSignOutEvent()` when it's the token is null.

dimkl (changes requested):
I have left some comments. Also we need to add a test to verify the expected behaviour!

.changeset/chatty-feet-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/clerk-js": patch
+---
+
+Fix bug where session cookie wasn't cleared after account deletion
```
panteliselef, line 5:
```suggestion
Bug fix: Clear session cookie after a user deletes their account
```

panteliselef (approved):
LGTM, added a comment for an updated changeset

octoper:
@dimkl Added the change and E2E tests as you suggested

dimkl (approved):

PR #3646, opened 2024-07-01
feat(astro): Introduce Astro SDK

panteliselef:
## Description
This PR is tackling ECO-6 and ports the original code from [astro-clerk-auth](https://github.com/panteliselef/astro-with-clerk-auth) inside this monorepo as a new package called `@clerk/astro`.

`@clerk/astro` will become the official Astro SDK from Clerk.

While porting the original code the following things happened
- Hotload is now the default method of loading clerk-js on the client.
- Dropped the deprecated internal stores, now the stores are only available from `/client/stores`.
- Dropped the `/v0` module
- Dropped the exported `clerkClient` as a variable, now you need to call `clerkClient(context)` and pass the astro context.

In follow up PRs, we will introduce a playground application for Astro, and e2e tests for the node runtime

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/astro/src/server/get-auth.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const jwt = decodeJwt(authToken as string);
+    // @ts-expect-error - TODO: Align types
+    return signedInAuthObject({ ...options, sessionToken: jwt.raw.text }, jwt.payload);
+  };
+};
+
+const authAuthHeaderMissing = (helperName = 'auth') =>
+  `Clerk: ${helperName}() was called but Clerk can't detect usage of authMiddleware(). Please ensure the following:
+    - authMiddleware() is used in your Next.js Middleware.
```
wobsoriano, line 60: We forgot to update the middleware and SDK name here

packages/astro/src/server/clerk-middleware.ts
```diff
@@ -0,0 +1,343 @@
...
+   NEXT_PUBLIC_CLERK_IS_SATELLITE='true'
+   `;
+
+// Duplicate from '@clerk/nextjs'
+export const missingSignInUrlInDev = `
+Invalid signInUrl. A satellite application requires a signInUrl for development instances.
+Check if signInUrl is missing from your configuration or if it is not an absolute URL
+
+1) With middleware
+   e.g. export default authMiddleware({signInUrl:'SOME_URL', isSatellite:true});
```
wobsoriano, line 209: We forgot to update the middleware and SDK names here as well

packages/astro/src/astro-components/interactive/OrganizationList.astro
wobsoriano, line 1: I am thinking if there's a way to reduce duplication of code from the interactive components. Could making a component that does all of the logic work? 🤔  Say we have an `InteractiveComponent` that accepts the element ID prefix + the props

```astro
---
import type { SignInProps } from "@clerk/types";
type Props = SignInProps;

import InteractiveLayout from './InteractiveLayout.astro'

const props = {
  ...Astro.props
}
---

<InteractiveLayout {...props} prefix="sign-in" />
```

wobsoriano (commented):
@panteliselef  and I reviewed this code last week, and the changes look good to me.

I've left a few suggestions for improvement.

packages/astro/src/astro-components/interactive/OrganizationList.astro
panteliselef: I remember me trying to do something similar, unfortunately i cannot recall what was the wall i was hitting. Probably we should have another look and either implement the suggestion or document why we can't.

packages/astro/src/astro-components/control/Protect.astro
```diff
@@ -0,0 +1,38 @@
...
+  | {
+      condition: (has: CheckAuthorizationWithCustomPermissions) => boolean;
+      role?: never;
+      permission?: never;
+    }
+  | {
+      condition?: never;
+      role?: never;
+      permission?: never;
+    };
```
BRKalow, line 28: These are common with the types for `Protect` from our Next.js SDK right? Maybe an opportunity to consolidate 👀

packages/astro/src/astro-components/control/Protect.astro
```diff
@@ -0,0 +1,38 @@
...
+
+const { has, userId } = Astro.locals.auth();
+const isUnauthorized =
+  !userId ||
+  (typeof Astro.props.condition === "function" &&
+    !Astro.props.condition(has)) ||
+  ((Astro.props.role || Astro.props.permission) && !has(Astro.props));
```
BRKalow, line 35: nit: What do you think about reversing the logic here? `isAuthorized` feels clearer than `isUnauthorized` (default to unauthorized in the logic).

packages/astro/src/astro-components/interactive/OrganizationList.astro
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    window.__astro_clerk_component_props.get(category)?.set(id, props);
+  };
+
+  setOrCreatePropMap({
+    category: "organization-list",
+    id: `clerk-organization-list-${safeId}`,
+    props,
+  });
```
BRKalow, line 35: Can you explain what's happening here? Maybe it's my unfamiliarity with Astro, but it's not immediately clear what we're doing with this inline script. Alternatively, is it possible to consolidate this somehow into a single source of truth? Or is there some limitation that requires us to duplicate it in all of the interactive component definitions?

packages/astro/src/stores/external.ts
```diff
@@ -0,0 +1,31 @@
+import { computed } from 'nanostores';
+
+import { $clerk, $csrState, $initialState } from './internal';
+import { deriveState } from './utils';
+
+export const $authStore = computed([$csrState, $initialState], (state, initialState) => {
```
BRKalow, line 6: Is this part of the public API? If so, can we add a JSDoc comment?

packages/astro/src/astro-components/control/Protect.astro
BRKalow: I see this is declared as `ProtectComponentDefaultProps` elsewhere, can we reuse that?

packages/astro/src/server/build-clerk-hotload-script.ts
```diff
@@ -0,0 +1,64 @@
...
+type BuildClerkJsScriptOptions = {
+  proxyUrl: string;
+  domain: string;
+  clerkJSUrl?: string;
+  clerkJSVariant?: 'headless' | '';
+  clerkJSVersion?: string;
+  publishableKey: string;
+};
+
+const clerkJsScriptUrl = (opts: BuildClerkJsScriptOptions) => {
```
BRKalow, line 20: This file is mostly duplicative of the logic in clerk-react right? What would it take to re-use the logic from there?

packages/astro/src/server/clerk-middleware.ts
```diff
@@ -0,0 +1,343 @@
...
+  res.headers.set(constants.Headers.AuthReason, reason || '');
+
+  /**
+   * Populate every page with the authObject. This allows for SSR to work properly
+   * without the developer having to wrap manually each page with `ClerkLayout.astro`
+   * ^ ClerkLayout is still needed in order to populate the ssrState store, but it not responsible for passing the data to a page.
+   */
+  if (res.headers.get('content-type') === 'text/html') {
+    const reader = res.body?.getReader();
+    const stream = new ReadableStream({
```
BRKalow, line 243: 💭 Would [TransformStream](https://developer.mozilla.org/en-US/docs/Web/API/TransformStream) be more appropriate here? That's kind of what we're doing, in a way.

packages/astro/src/server/clerk-middleware.ts
```diff
@@ -0,0 +1,343 @@
...
+        let { value, done } = await reader!.read();
+        const encoder = new TextEncoder();
+        const decoder = new TextDecoder();
+        while (!done) {
+          const decodedValue = decoder.decode(value);
+
+          /**
+           * Hijack html response to position `__CLERK_ASTRO_DATA__` before the closing `head` html tag
+           */
+          if (decodedValue.includes('</head>')) {
```
BRKalow, line 254: This could likely be made more efficient if we check against the encoded value instead of decoding & encoding the entire body.

e.g.
```ts
const ENCODED_END_HEAD_MARKER = new TextEncoder().encode('</head>');

...

if (value.includes(ENCODED_END_HEAD_MARKER)) {
  const [p1, p2] = value.split(ENCODED_END_HEAD_MARKER);
  controller.enqueue(p1);
  ...
  controller.enqueue(p2);
} else {
  controller.enqueue(value)
}
```

packages/astro/src/server/clerk-middleware.ts
```diff
@@ -0,0 +1,343 @@
...
+              ),
+            );
+
+            controller.enqueue(
+              encoder.encode(
+                `<script id="__CLERK_ASTRO_SAFE_VARS__" type="application/json">${JSON.stringify(getClientSafeEnv(locals))}</script>\n`,
+              ),
+            );
+
+            if (__HOTLOAD__) {
```
BRKalow, line 269: ❓ why are we providing hotloading as an option?

packages/astro/src/server/get-auth.ts
```diff
@@ -0,0 +1,66 @@
...
+export const createGetAuth = ({ noAuthStatusMessage }: { noAuthStatusMessage: string }) => {
+  return (
+    req: Request,
+    locals: APIContext['locals'],
+    opts?: { secretKey?: string },
+  ): AuthObjectWithoutResources<SignedInAuthObject> | AuthObjectWithoutResources<SignedOutAuthObject> => {
+    // When the auth status is set, we trust that the middleware has already run
+    // Then, we don't have to re-verify the JWT here,
+    // we can just strip out the claims manually.
+    const authToken = locals.authToken || getAuthKeyFromRequest(req, 'AuthToken');
```
BRKalow, line 30: how does Astro middleware work? Does it run in the same process as the app server itself, or is it like Next where it runs somewhere else? Do we need to worry about the validity of the token here or are we safe to assume it is the token we previously verified in the middleware?

packages/astro/src/internal/utils/versionSelector.ts
```diff
@@ -0,0 +1,7 @@
+export const versionSelector = (clerkJSVersion: string | undefined): string => {
+  if (clerkJSVersion) {
+    return clerkJSVersion;
+  }
+
+  return '5';
```
BRKalow, line 6: ❓ Consider moving this to a constant so it's easier to find and change in the future.

packages/astro/src/integration/create-integration.ts
```diff
@@ -0,0 +1,112 @@
...
+           */
+
+          injectScript(
+            'page',
+            `
+            ${command === 'dev' ? `console.log("${packageName}","Initialize Clerk: page")` : ''}
+            import { runInjectionScript } from "${buildImportPath}";
+            await runInjectionScript(${JSON.stringify(params)});`,
+          );
+        },
```
BRKalow, line 106: Will these two both run if a UI framework is present? Is that problematic?

packages/astro/src/integration/index.ts
```diff
@@ -0,0 +1,5 @@
+import { createIntegration } from './create-integration';
+
+export default createIntegration({
+  mode: 'hotload',
```
BRKalow, line 4: is this overridable by the user? If yes, why?

BRKalow (commented):
A self-review would be helpful here given the size of the PR and most folks unfamiliarity with some of the more Astro-specific logic.

packages/astro/src/astro-components/interactive/OrganizationList.astro
panteliselef: Niceee 💯

packages/astro/src/astro-components/control/Protect.astro
panteliselef: So the idea was, to do minimal changes on the first PR and start removing or cleaning up, afterwards. There is already a ticket for handling/improving sharing logic. I believe what I should do is do another pass identify potential "shared" code, mark it, and document it in the ticket. Once we have e2e tests or a few example apps using the official package then we can safely start improving the shared logic.

packages/astro/src/astro-components/interactive/OrganizationList.astro
panteliselef: @wobsoriano seems to have a way in order to address the multiple duplicates. I remember myself, hitting a limitation but we will revisit this one.

I will add a comment in this file that describes what is happening here.

packages/astro/src/stores/external.ts
panteliselef: Yes it is, i will add them

packages/astro/src/server/clerk-middleware.ts
panteliselef: So in the community sdk, hotloading was opt-in, now it is opt-out, but we are still allowing devs to use the astro integration with clerk-js as a direct dependency bundled with the rest of the app.

packages/astro/LICENCE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022 Clerk, Inc.
```
anagstef, line 3: Let's go with `2024`
```suggestion
Copyright (c) 2024 Clerk, Inc.
```

packages/astro/src/server/get-auth.ts
panteliselef: It runs on the same machine, where SSR takes place. Astro supports deployment to vercel where you can choose to use the edge middleware, but we are kind of blocking that because our middleware is using an AsyncLocalStorage

packages/astro/README.md
```diff
@@ -0,0 +1,369 @@
...
+  </a>
+  <br />
+</p>
+
+# @clerk/astro
+
+<div align="center">
+
+[![Chat on Discord](https://img.shields.io/discord/856971667393609759.svg?logo=discord)](https://clerk.com/discord)
+[![Clerk documentation](https://img.shields.io/badge/documentation-clerk-green.svg)](https://clerk.com/docs?utm_source=github&utm_medium=clerk_nextjs)
```
anagstef, line 16:
```suggestion
[![Clerk documentation](https://img.shields.io/badge/documentation-clerk-green.svg)](https://clerk.com/docs?utm_source=github&utm_medium=clerk_astro)
```

packages/astro/README.md
```diff
@@ -0,0 +1,369 @@
...
+});
+```
+
+## Support
+
+You can get in touch with us in any of the following ways:
+
+- Join our official community [Discord server](https://clerk.com/discord)
+- Create a [GitHub Discussion](https://github.com/clerk/javascript/discussions)
+- Contact options listed on [our Support page](https://clerk.com/support?utm_source=github&utm_medium=clerk_nextjs)
```
anagstef, line 348:
```suggestion
- Contact options listed on [our Support page](https://clerk.com/support?utm_source=github&utm_medium=clerk_astro)
```

packages/astro/src/integration/index.ts
panteliselef: A developer cannot manipulate this code, or call `createIntegration` explictly.
what the can do though is import the clerk integration from `@clerk/astro/bunlded` instead of `@clerk/astro`

The default export of `@clerk/astro` uses the clerk integration with hotloading.

packages/astro/src/integration/create-integration.ts
panteliselef: There will be no problem, because `runInjectionScript` will only execute clerk-js initialization once

panteliselef: I will write more comments about why we need to have both.

packages/astro/src/client/react/uiComponents.tsx
```diff
@@ -0,0 +1,134 @@
+import type {
```
anagstef, line 1: ❓ This file is copy pasted from `@clerk/clerk-react` right?

packages/astro/src/client/react/utils.tsx
```diff
@@ -0,0 +1,39 @@
+import type { LoadedClerk } from '@clerk/types';
```
anagstef, line 1: ❓ also copy pasted?

anagstef (approved):
🚀 ⭐

packages/astro/README.md
```diff
@@ -0,0 +1,369 @@
+<p align="center">
+  <a href="https://clerk.com?utm_source=github&utm_medium=clerk_nextjs" target="_blank" rel="noopener noreferrer">
```
LekoArts, line 2:
```suggestion
  <a href="https://clerk.com?utm_source=github&utm_medium=clerk_astro" target="_blank" rel="noopener noreferrer">
```

packages/astro/README.md
```diff
@@ -0,0 +1,369 @@
...
+
+To build the package in watch mode, run the following:
+
+```sh
+npm run dev
+```
+
+## Report Issues
+
+If you are experiencing issues please submit them via the [Issues page in GH](https://github.com/panteliselef/astro-with-clerk-auth/issues). As this SDK is not officially suppported by Clerk or Astro, contacting them directly for issues regarding this package might cause your requests to be unanswered.
```
LekoArts, line 80: This needs change

packages/astro/package.json
```diff
@@ -0,0 +1,136 @@
+{
+  "name": "@clerk/astro",
+  "description": "Clerk SDK for Astro",
+  "version": "0.0.0",
+  "type": "module",
+  "private": false,
```
LekoArts, line 6: You can remove that line

packages/astro/package.json
```diff
@@ -0,0 +1,136 @@
...
+  "devDependencies": {
+    "@clerk/eslint-config-custom": "*",
+    "react": "*",
+    "typescript": "*"
+  },
+  "peerDependencies": {
+    "astro": "^3.2.0 || ^4.0.0"
+  },
+  "engines": {
+    "node": ">=18.17.0"
```
LekoArts, line 131: The README says that Node 20 is required?

packages/astro/src/astro-components/control/Protect.astro
LekoArts: Agree, that this is hard to grok

panteliselef: That's a fair point, I will added it as a todo if that's ok

packages/astro/src/astro-components/interactive/OrganizationList.astro
```diff
@@ -0,0 +1,36 @@
...
+const safeId = customAlphabet(urlAlphabet, 10)();
+
+const props = {
+  ...Astro.props,
+}
+---
+
+<div id={`clerk-organization-list-${safeId}`}></div>
+
+<script is:inline define:vars={{ props: Astro.props, safeId }}>
```
LekoArts, line 16: The `safeId` and `setOrCreatePropMap` are repeated in multiple Astro components. Can we consolidate that into shared utils?

packages/astro/src/integration/create-integration.ts
```diff
@@ -0,0 +1,112 @@
...
+    const clerkJSUrl = (params as any)?.clerkJSUrl as string | undefined;
+    const clerkJSVariant = (params as any)?.clerkJSVariant as string | undefined;
+    const clerkJSVersion = (params as any)?.clerkJSVersion as string | undefined;
+
+    return {
+      name: '@clerk/astro/integration',
+      hooks: {
+        'astro:config:setup': ({ config, injectScript, updateConfig, logger, command }) => {
+          if (config.output === 'static') {
+            logger.error(`${packageName} requires SSR to be turned on. Please update output to "server"`);
```
LekoArts, line 32: Why is SSR required? Can't it also work with CSR?

LekoArts (changes requested):
First pass

packages/astro/src/client/react/uiComponents.tsx
panteliselef: Yes, but `withClerk` is a custom implementation.

packages/astro/src/client/react/utils.tsx
panteliselef: Yes, but `withClerk` is a custom implementation.

packages/astro/package.json
panteliselef: Are you referring to  `private: false` ?

packages/astro/README.md
panteliselef: I dropped it.

packages/astro/src/astro-components/interactive/OrganizationList.astro
panteliselef: We will resolve this in a follow up PR, i've documented it

packages/astro/package.json
panteliselef: I had it original to node 20, but because our ci runs with v18 i had to update it. No worries i've checked the Astro minimum node version and this is compatible, also the API for `AsyncLocalStorage` that we are using has been stable for many versions prior to v18.

packages/astro/src/integration/create-integration.ts
panteliselef: Currently we support only SSR, because we are injecting the clerk-js script from the middleware. But i will add a ticket in order to explore this.

packages/astro/src/astro-components/control/Protect.astro
panteliselef: In this case i think i will just use the already existed type as you recommended.

packages/astro/src/server/build-clerk-hotload-script.ts
panteliselef: This will be address with ECO-4, but thanks for mentioning it

packages/astro/package.json
LekoArts: Yes

LekoArts (approved):

LekoArts (approved):

PR #3655, opened 2024-07-02
fix(clerk-react): Fix race condition on updating ClerkProvider props

anagstef:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/react/src/isomorphicClerk.ts
+++ b/packages/react/src/isomorphicClerk.ts
@@ -356,9 +356,6 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
 
   #waitForClerkJS(): Promise<HeadlessBrowserClerk | BrowserClerk> {
     return new Promise<HeadlessBrowserClerk | BrowserClerk>(resolve => {
-      if (this.#loaded) {
-        resolve(this.clerkjs!);
-      }
       this.addOnLoaded(() => resolve(this.clerkjs!));
     });
   }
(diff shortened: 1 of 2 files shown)


packages/react/src/isomorphicClerk.ts
```diff
@@ -356,9 +356,6 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
 
   #waitForClerkJS(): Promise<HeadlessBrowserClerk | BrowserClerk> {
     return new Promise<HeadlessBrowserClerk | BrowserClerk>(resolve => {
-      if (this.#loaded) {
-        resolve(this.clerkjs!);
-      }
```
BRKalow, line 361: ❓ Why are we removing this check? Does `addOnLoaded` to the same internally?

anagstef: Exactly! This condition may create another race condition, because this `resolve` may run first, before the ones that are already on the callback array. It's better to always use the callback array so the order is always the same.

BRKalow (approved):

dimkl (approved):
❓ Could we add a test for this?

anagstef:
@dimkl Added a unit test!

Also, I updated a type!

packages/react/src/isomorphicClerk.ts
```diff
@@ -579,13 +577,13 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
...
-    } else {
-      return undefined;
-    }
+  __unstable__updateProps = (props: any): void => {
+    void this.#waitForClerkJS().then(clerkjs => {
+      // Handle case where accounts has clerk-react@4 installed, but clerk-js@3 is manually loaded
+      if (clerkjs && '__unstable__updateProps' in clerkjs) {
+        (clerkjs as any).__unstable__updateProps(props);
+      }
+    });
```
panteliselef, line 586: Aren't we the only ones that we will be using `__unstable__updateProps` ?
Is this public API ?
I propose to update this to be 
```suggestion
  __unstable__updateProps = (props: any): Promise<void> => {
    return this.#waitForClerkJS().then(clerkjs => {
      // Handle case where accounts has clerk-react@4 installed, but clerk-js@3 is manually loaded
      if (clerkjs && '__unstable__updateProps' in clerkjs) {
        (clerkjs as any).__unstable__updateProps(props);
      }
    });
```

panteliselef (approved):

packages/react/src/isomorphicClerk.ts
anagstef: I was thinking exactly the same thing, but I think we have users that use this property... But, I guess it does not really change something for them, as it was `void` before.

Plus, this way we can await it correctly 🤔

anagstef: Done! ✔️

PR #3656, opened 2024-07-02
chore(astro): Add unstyled components

wobsoriano:
## Description

This PR is tackling ECO-2 and adds unstyled sign in, sign up, and sign out components for both Astro and React.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/astro/src/astro-components/react/SignInButton.astro
wobsoriano, line 1: The unstyled components from `/astro-components/react` directory are basically re-exported unstyled components from `/astro-components/unstyled/` but with a `client:load` directive already added so users can easily import in Astro components without having to add a client directive. It will then be automatically hydrated.

packages/astro/src/astro-components/unstyled/SignOutButton.astro
```diff
@@ -0,0 +1,27 @@
+---
+import type { HTMLTag, Polymorphic } from 'astro/types'
+type Props<Tag extends HTMLTag> = Polymorphic<{ as: Tag; id?: string; redirectUrl?: string }>
+
+import { customAlphabet, urlAlphabet } from "nanoid";
+
+const safeId = customAlphabet(urlAlphabet, 10)();
```
wobsoriano, line 7: Starting to notice we're repeating this piece of logic in all our Astro components. I'll make a ticket that will added this

packages/astro/src/astro-components/unstyled/SignUpButton.astro
```diff
@@ -0,0 +1,51 @@
+---
+import type { HTMLTag, Polymorphic } from 'astro/types'
+import type { SignUpProps } from "@clerk/types";
+type Props<Tag extends HTMLTag> = Polymorphic<SignUpProps & { as: Tag; id?: string; mode?: 'redirect' | 'modal' }>
```
wobsoriano, line 4: The [`Polymorphic`](https://docs.astro.build/en/guides/typescript/#polymorphic-type) type here is a helper that ensure that passed element is a valid HTML element.

packages/astro/src/astro-components/unstyled/SignInButton.astro
```diff
@@ -0,0 +1,49 @@
+---
+import type { HTMLTag, Polymorphic } from 'astro/types'
+import type { SignInProps } from "@clerk/types";
+type Props<Tag extends HTMLTag> = Polymorphic<SignInProps & { as: Tag; id?: string; mode?: 'redirect' | 'modal' }>
```
wobsoriano, line 4: The [Polymorphic](https://docs.astro.build/en/guides/typescript/#polymorphic-type) type here is a helper that ensure that passed element is a valid HTML element.

packages/astro/src/astro-components/unstyled/SignInButton.astro
```diff
@@ -0,0 +1,49 @@
...
+
+import { customAlphabet, urlAlphabet } from "nanoid";
+
+const safeId = customAlphabet(urlAlphabet, 10)();
```
wobsoriano, line 8: Starting to notice we're repeating this piece of logic in all our Astro components. I'll make a ticket that will added this

packages/astro/src/astro-components/unstyled/SignInButton.astro
```diff
@@ -0,0 +1,49 @@
...
+  signUpForceRedirectUrl,
+};
+---
+
+<Tag {...elementProps} id={id}>
+  <slot>Sign in</slot>
+</Tag >
+
+<script is:inline define:vars={{ signInOptions, mode, safeId }}>
+  const btn = document.querySelector(`#${safeId}`);
```
wobsoriano, line 34: Question about the usage of `id` here - is it safe to use the `id` as selector or should we use a `data-*` attribute? `id` has a big chance of getting replaced here since this is an unstyled component.

packages/astro/src/astro-components/unstyled/SignInButton.astro
```diff
@@ -0,0 +1,49 @@
...
+  forceRedirectUrl,
+  fallbackRedirectUrl,
+  signUpFallbackRedirectUrl,
+  signUpForceRedirectUrl,
+};
+---
+
+<Tag {...elementProps} id={id}>
+  <slot>Sign in</slot>
+</Tag >
```
wobsoriano, line 31: For Astro components, we're not using the `asChild` and children props to render a custom element like what we have in the React SDK. We're using the built-in slots. More info on ECO-2 ticket

packages/astro/src/astro-components/unstyled/SignInButton.astro
wobsoriano: @panteliselef If we go this way we will need to update the [mounting function](https://github.com/clerk/javascript/blob/5c3adbbb3e358182ae307beec5f67f35a79ca51c/packages/astro/src/client/mount-clerk-astro-js-components.ts#L18) since it looks for all `data-clerk-id` elements.

What about

- `data-clerk-interactive-ui` for the interactive components
- `data-clerk-unstyled-ui` for the unstyled

or we can drop the `ui` in there as well

packages/astro/src/astro-components/react/SignInButton.astro
```diff
@@ -0,0 +1,15 @@
...
+
+const props = {
+  ...Astro.props
+}
+---
+
+<SignInButtonReact
+  client:load
+  {...props}
+/>
```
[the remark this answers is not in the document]
wobsoriano: I think we should just drop it. Unstyled components won't benefit here anyway since islands don't support `React.cloneElement`

packages/astro/src/client/react/types.ts
```diff
@@ -0,0 +1,20 @@
...
+    'fallbackRedirectUrl' | 'forceRedirectUrl' | 'signUpForceRedirectUrl' | 'signUpFallbackRedirectUrl'
+  >;
+
+export type SignUpButtonProps = {
+  unsafeMetadata?: SignUpUnsafeMetadata;
+} & ButtonProps &
+  Pick<
+    SignUpProps,
+    'fallbackRedirectUrl' | 'forceRedirectUrl' | 'signInForceRedirectUrl' | 'signInFallbackRedirectUrl'
+  >;
```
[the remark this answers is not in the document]
wobsoriano: Updated, thanks!

packages/astro/src/astro-components/unstyled/SignInButton.astro
wobsoriano: No actually, `data-clerk-ui` makes sense since it's a UI component. `data-clerk-unstyled` sounds more better

wobsoriano: Updated 🚀

packages/astro/src/astro-components/unstyled/SignOutButton.astro
```diff
@@ -0,0 +1,27 @@
...
+<Tag {...elementProps} id={id}>
+  <slot>Sign out</slot>
+</Tag >
+
+<script is:inline define:vars={{ redirectUrl, safeId }}>
+  const btn = document.querySelector(`#${safeId}`);
+
+  btn.addEventListener("click", () => {
+    window.Clerk.signOut({ redirectUrl })
+  });
```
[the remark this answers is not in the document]
wobsoriano: Nice catch. I think we need to respect the props like `sessionId` since it's required for signing out of a specific session.

Also, looking at the `SignOutButton` [reference](https://clerk.com/docs/components/unstyled/sign-out-button#sign-out-button-properties), the callback is never stated and it's recommended to migrate to `redirectUrl` in the [Core 2 migration doc](https://clerk.com/docs/upgrade-guides/core-2/remix#other-breaking-changes).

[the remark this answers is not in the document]
wobsoriano: I'm still waiting for a response to this but the `signOutCallback` is being deprecated

panteliselef (approved):

PR #3658, opened 2024-07-03
fix(chrome-extension): Ensure synced client cookie URLs resolve appropriately

tmilewski:
## Description

Ensures that there's a valid URL as some don't resolve with a protocol

Fixes SDK-1867

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


tmilewski:
!snapshot

clerk-cookie:
Hey @tmilewski - the snapshot version command generated the following package versions:
| Package | Version |
| --- | --- |
| @clerk/chrome-extension | 1.1.2-snapshot.vd49286e |

`@clerk/chrome-extension`
```sh
npm i @clerk/chrome-extension@1.1.2-snapshot.vd49286e --save-exact
```
`gatsby-plugin-clerk`
```sh
```

dimkl (dismissed):
Instead of making this change, why don't we fix the code where it's being called to pass a valid URL with protocol?
This issue is caused by incorrectly passing the `clerk.frontendApi` (which is not a full URL) as param in the JWTHandler.

ref: https://github.com/clerk/javascript/blob/1a704ed9de3ad222ff17d10cb383639b2fc038b9/packages/chrome-extension/src/singleton.ts#L48

tmilewski:
> Instead of making this change, why don't we fix the code where it's being called to pass a valid URL with protocol? This issue is caused by incorrectly passing the `clerk.frontendApi` (which is not a full URL) as param in the JWTHandler.
> 
> ref:
> 
> https://github.com/clerk/javascript/blob/1a704ed9de3ad222ff17d10cb383639b2fc038b9/packages/chrome-extension/src/singleton.ts#L48

@dimkl Yup, I just prepended it. I'm going to leave the other method as there are other cases where that will be necessary and ultimately extended.

packages/chrome-extension/src/singleton.ts
```diff
@@ -45,7 +45,7 @@ export async function buildClerk({
   // Set up cookie params based on environment
   const getClientCookieParams: GetClientCookieParams = isProd
     ? {
-        urls: key.frontendApi,
+        urls: `https://${key.frontendApi}`,
```
anagstef, line 48: Can we merge the logic we have here and in `ensureFormattedUrl` somehow? 
For example, use here the `ensureFormattedUrl` function.

tmilewski: @anagstef `ensureFormattedUrl` is going to handle a lot more use-cases soon. This PR is largely to solve a singular current issue.

anagstef (approved):
LGTM

dimkl:
> @dimkl Yup, I just prepended it. I'm going to leave the other method as there are other cases where that will be necessary and ultimately extended.

@tmilewski Could you provide example of other cases?
Based on the code, the only case where the url as string is used is for a production instance where we explicitly pass the frontendApiKey (prefixed with `https` after this PR). In the case of the development instance, an array of URLs is passed and this code is not executed.
Could we revert the changes in the `packages/chrome-extension/src/utils/cookies.ts` as it seems unnecessary?

Issue #3660, opened 2024-07-03
`@clerk/nextjs@5.2.0` cause a build warning in NextJS middleware


arvinxx:
### Preliminary Checks

- [X] I have reviewed the documentation: https://clerk.com/docs
- [X] I have searched for existing issues: https://github.com/clerk/javascript/issues
- [X] I have not already reached out to Clerk support via email or Discord (if you have, no need to open an issue here)
- [X] This issue is not a question, general help request, or anything other than a bug report directly related to Clerk. Please ask questions in our Discord community: https://clerk.com/discord.

### Reproduction

https://github.com/lobehub/lobe-chat/tree/reproduction/clerk-nextjs-5.2.0

### Publishable key

no need, it's a build time issue

### Description

here is the build log:
```
   Creating an optimized production build ...
 ✓ (pwa) Compiling for server...
 ✓ (pwa) Compiling for server...
 ✓ (pwa) Compiling for client (static)...
 ○ (pwa) Service worker: /vercel/path1/public/sw.js
 ○ (pwa)   URL: /sw.js
 ○ (pwa)   Scope: /
 ⚠ Compiled with warnings
./node_modules/@clerk/nextjs/dist/esm/server/clerkMiddleware.js
A Node.js module is loaded ('async_hooks' at line 3) which is not supported in the Edge Runtime.
Learn More: https://nextjs.org/docs/messages/node-module-in-edge-runtime
Import trace for requested module:
./node_modules/@clerk/nextjs/dist/esm/server/clerkMiddleware.js
./node_modules/@clerk/nextjs/dist/esm/server/index.js
./node_modules/@clerk/shared/dist/chunk-RSOCGYTF.mjs
A Node.js API is used (MessageEvent at line: 28) which is not supported in the Edge Runtime.
Learn more: https://nextjs.org/docs/api-reference/edge-runtime
Import trace for requested module:
./node_modules/@clerk/shared/dist/chunk-RSOCGYTF.mjs
./node_modules/@clerk/shared/dist/index.mjs
./node_modules/@clerk/nextjs/dist/esm/server/utils.js
./node_modules/@clerk/nextjs/dist/esm/server/clerkClient.js
./node_modules/@clerk/nextjs/dist/esm/server/index.js
```

you can see the error from `@clerk/nextjs`. please remove this usage: 

[...]

PR #3661, opened 2024-07-03
fix(nextjs): Update imports to avoid Node.js build warnings on edge runtime

LauraBeatris:
## Description

Fixes https://github.com/clerk/javascript/issues/3660

- Uses `AsyncLocalStorage` available as a global polyfill for the edge runtime, instead of importing from Node.js: https://nextjs.org/docs/pages/api-reference/edge#nextjs-specific-polyfills
- Imports `logger` from nested import

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/src/server/clerkMiddleware.ts
+++ b/packages/nextjs/src/server/clerkMiddleware.ts
@@ -1,3 +1,5 @@
+import { AsyncLocalStorage } from 'node:async_hooks';
+
 import type {
   AuthenticateRequestOptions,
   AuthObject,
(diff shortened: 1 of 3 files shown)


BRKalow (approved):

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -1,3 +1,5 @@
+import { AsyncLocalStorage } from 'node:async_hooks';
```
LauraBeatris, line 1: By using `node:async_hooks`, it uses the `AsyncLocalStorage` polyfill from Next.js. TBH it's a controversy from what Vercel docs mention, in which `async_hooks` would also be polyfilled: https://vercel.com/docs/functions/runtimes/edge-runtime#compatible-node.js-modules

CleanShot 2024-07-04 at 10 26 57

PR #3662, opened 2024-07-03
chore(astro): Add telemetry and SDK Metadata

wobsoriano:
## Description

This PR is tackling ECO-3 and records telemetry event of server-side middleware usage, as well as nanostores client usage.

As stated in the initial Astro PR, playground and tests will follow.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


turbo.json
```diff
@@ -25,7 +25,8 @@
     "VERCEL",
     "VITE_CLERK_*",
     "EXPO_PUBLIC_CLERK_*",
-    "REACT_APP_CLERK_*"
+    "REACT_APP_CLERK_*",
+    "MODE"
```
wobsoriano, line 29: This is added because we've added a global env var (`import.meta.env.MODE`) in the server SDK Metadata

packages/astro/src/stores/external.ts
```diff
@@ -124,3 +121,19 @@ export const $signInStore = computed([$clientStore], client => client?.signIn);
...
+ */
+const recordTelemetryEvent = (store: Store, method: string) => {
+  onMount(store, () => {
+    $clerk.get()?.telemetry?.record(eventMethodCalled(method));
+  });
+};
+
+recordTelemetryEvent($signInStore, '$signInStore');
+recordTelemetryEvent($signUpStore, '$signUpStore');
+recordTelemetryEvent($organizationStore, '$organizationStore');
```
wobsoriano, line 139: We're still unsure about this. Do we have any criteria on recording telemetry events in the client?

This nanostores are basically 1:1 of React hooks. When a store is used, the `onMount` callback gets invoked and that's the time we record the event

packages/astro/src/stores/external.ts
```diff
@@ -1,4 +1,5 @@
-import { computed } from 'nanostores';
+import { eventMethodCalled } from '@clerk/shared/dist/telemetry';
```
panteliselef, line 1:
```suggestion
import { eventMethodCalled } from '@clerk/shared/telemetry';
```

dimkl (approved):

packages/astro/src/stores/external.ts
panteliselef: it's kind of tough to find a common solution for hooks and stores here. I think we can continue with this, and then later in analytics calculate the actual usage of stores by subtracting the usage of its respective hook.

something among the lines of 
```ts
userStoreUsage = getCount(`$userStore`) - `getCount(`$useUser`)
```

does this make sense to you ? Do you have an alternative ?

Also we can always use new computed stores for our hooks in order to have a different reference.

turbo.json
panteliselef: 🙃 Is it possible to somehow bypass this with an eslint disable rule ? I'd avoid adding this here, because it's not clear what it does, unless we add an inline comment inside `turbo.json`

packages/astro/src/stores/external.ts
panteliselef, line 139: ❓ Shall we add do this for all stores ?

turbo.json
wobsoriano: Yes it's possible to bypass this. Now we have 2 instances of `import.meta.env.MODE`, one is inside integration. I agree to bypass it for now, it's unclear what it does. Are we okay with bypassing for now @panteliselef ?

packages/astro/src/stores/external.ts
wobsoriano: The idea is to match the recording of the React hooks which only records `useSignIn`, `useSignUp`, `useOrganization` and `useOrganizationList`. But if it makes sense to record all of the stores, I'm not sure 🤔

wobsoriano: Yeah it makes sense, so that if we update later we can track the exact usage of stores (not the wrapping hooks). Let's go with it then for now

turbo.json
wobsoriano: Updated

packages/astro/src/stores/external.ts
wobsoriano: What if we just do it to all stores in this file, and if it gets too crowded in the analytics later, update it to only log which makes sense?

panteliselef: As discussed offline, we will open a new PR if we need to add telemetry to the remaining stores

panteliselef (approved):

PR #3664, opened 2024-07-03
chore(astro): Remove duplicate logic in interactive components

wobsoriano:
## Description

This PR addresses ECO-16 by consolidating the logic of interactive components into a single component used across all of them.

As said in the initial Astro PR, playground and tests will follow.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/astro/src/astro-components/interactive/InteractiveUI.astro
wobsoriano, line 1: There's probably a better component name for this

panteliselef: 🙃 `InternalUIComponentRenderer` ?

packages/astro/src/astro-components/interactive/InteractiveUI.astro
```diff
@@ -0,0 +1,42 @@
...
+import { customAlphabet, urlAlphabet } from "nanoid";
+
+const safeId = customAlphabet(urlAlphabet, 10)();
+
+const props = {
+  ...Astro.props,
+}
+---
+
+<div id={`clerk-${props.component}-${safeId}`}></div>
```
panteliselef, line 16: Do you think this needs to change to something like `data-clerk-id` ?

packages/astro/src/astro-components/interactive/InteractiveUI.astro
```diff
@@ -0,0 +1,42 @@
...
+  component: 'sign-in' | 'sign-up' | 'organization-list' | 'organization-profile' | 'organization-switcher' | 'user-button' | 'user-profile'
+}
+
+import { customAlphabet, urlAlphabet } from "nanoid";
+
+const safeId = customAlphabet(urlAlphabet, 10)();
+
+const props = {
+  ...Astro.props,
+}
```
panteliselef, line 13: Do we need this ? I think we are accessing `Astro.props` below directly

packages/astro/src/astro-components/interactive/InteractiveUI.astro
wobsoriano: Yeah that's more more sense, I have a question like that [here](https://github.com/clerk/javascript/pull/3656#discussion_r1665978552) in the unstyled components PR. We can go with that 👍🏼

wobsoriano: Also need to update the mounting function to use that attribute in the query selector

panteliselef: that's correct

packages/astro/src/astro-components/interactive/InteractiveUI.astro
wobsoriano: Still using this in new update, but destructured since the `component` prop is not a clerk UI element option

packages/astro/src/astro-components/interactive/InteractiveUI.astro
wobsoriano: Updated! I was about to use `InteractiveUIMounter` lol but this suggestion is better

panteliselef (approved):
🚢 it

PR #3673, opened 2024-07-08
chore(astro): Optimize stream modification

wobsoriano:
## Description

This PR tackles ECO-19 and ECO-18 and optimizes stream modification by working directly with the `Uint8Array` instead of decoding and re-encoding a chunk to check for string (`</head`>) availability. Introduces `TransformStream`.

A benchmark between the current implementation and the proposed is available [in this gist](https://gist.github.com/wobsoriano/1415371d490cc956cab4f56cf3d2a341).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.changeset/olive-foxes-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/astro": patch
+---
+
+Introduce `TransformStream` and work directly with encoded chunks when modifiying stream
```
panteliselef, line 5: Generally we prefer to write more "product" oriented changesets when possilble. In this case I'd mentioned the performance boost, instead of mentioning the "how" with `TransformStream`

wobsoriano: I'll keep that in mind, thanks! and updated

panteliselef (approved):
Nice work 💯

PR #3676, opened 2024-07-08
feat(astro): Add Google One Tap

panteliselef:
## Description
Introduces Google One Tap to Astro, as an interactive Astro component and as a React component that can be used inside your custom JSX.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


wobsoriano (approved):
Thank you 🚀

packages/astro/src/astro-components/interactive/InternalUIComponentRenderer.astro
```diff
@@ -1,7 +1,7 @@
 ---
 interface Props {
   [key: string]: unknown
-  component: 'sign-in' | 'sign-up' | 'organization-list' | 'organization-profile' | 'organization-switcher' | 'user-button' | 'user-profile'
+  component: 'sign-in' | 'sign-up' | 'organization-list' | 'organization-profile' | 'organization-switcher' | 'user-button' | 'user-profile' | 'google-one-tap'
```
wobsoriano, line 4: Sorry for this late follow up question. Would it be beneficial if we use the keys of the clerk components in the [mounting function](https://github.com/clerk/javascript/blob/5c3adbbb3e358182ae307beec5f67f35a79ca51c/packages/astro/src/client/mount-clerk-astro-js-components.ts#L7) as type so we don't have to dupe it?

panteliselef: In order to achieve that we would have to make this union available in a type exported from a submodule which makes it a public API and I am not sure if we wanna do that.

All files inside `astro-components` are basically just copied as they are in the correct location during the build and before publish. Later they will be bundled into the host app by Astro.

panteliselef: In this particular case because it's only types the host app seems to not break from the invalid relative path that will exist in this file, but still i believe the argument above stands

wobsoriano: Totally valid argument, no sense exposing the type. Thanks for answering!

PR #3681, opened 2024-07-09
chore(astro,shared): Move functions that can be reused by different SDKs

wobsoriano:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/src/deriveState.ts
+++ b/packages/shared/src/deriveState.ts
@@ -1,0 +1,71 @@
+import type {
+  ActiveSessionResource,
+  InitialState,
+  OrganizationCustomPermissionKey,
+  OrganizationCustomRoleKey,
+  OrganizationResource,
+  Resources,
+  UserResource,
+} from '@clerk/types';
+
+export const deriveState = (clerkLoaded: boolean, state: Resources, initialState: InitialState | undefined) => {
+  if (!clerkLoaded && initialState) {
+    return deriveFromSsrInitialState(initialState);
+  }
+  return deriveFromClientSideState(state);
+};
+
+const deriveFromSsrInitialState = (initialState: InitialState) => {
+  const userId = initialState.userId;
+  const user = initialState.user as UserResource;
+  const sessionId = initialState.sessionId;
+  const session = initialState.session as ActiveSessionResource;
+  const organization = initialState.organization as OrganizationResource;
+  const orgId = initialState.orgId;
+  const orgRole = initialState.orgRole as OrganizationCustomRoleKey;
+  const orgPermissions = initialState.orgPermissions as OrganizationCustomPermissionKey[];
+  const orgSlug = initialState.orgSlug;
+  const actor = initialState.actor;
+
+  return {
+    userId,
+    user,
+    sessionId,
+    session,
+    organization,
+    orgId,
+    orgRole,
+    orgPermissions,
+    orgSlug,
+    actor,
+  };
+};
+
+const deriveFromClientSideState = (state: Resources) => {
+  const userId: string | null | undefined = state.user ? state.user.id : state.user;
+  const user = state.user;
+  const sessionId: string | null | undefined = state.session ? state.session.id : state.session;
+  const session = state.session;
+  const actor = session?.actor;
+  const organization = state.organization;
+  const orgId: string | null | undefined = state.organization ? state.organization.id : state.organization;
+  const orgSlug = organization?.slug;
+  const membership = organization
+    ? user?.organizationMemberships?.find(om => om.organization.id === orgId)
+    : organization;
+  const orgPermissions = membership ? membership.permissions : membership;
+  const orgRole = membership ? membership.role : membership;
+
+  return {
+    userId,
+    user,
+    sessionId,
+    session,
+    organization,
+    orgId,
+    orgRole,
+    orgSlug,
+    orgPermissions,
+    actor,
+  };
+};
--- a/packages/astro/src/stores/external.ts
+++ b/packages/astro/src/stores/external.ts
@@ -1,8 +1,8 @@
+import { deriveState } from '@clerk/shared/deriveState';
 import { eventMethodCalled } from '@clerk/shared/telemetry';
 import { computed, onMount, type Store } from 'nanostores';
 
 import { $clerk, $csrState, $initialState } from './internal';
-import { deriveState } from './utils';
 
 /**
  * A client side store that is prepopulated with the authentication context during SSR.
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -51,6 +51,7 @@
     "cookie",
     "date",
     "deprecated",
+    "deriveState",
     "error",
     "file",
     "globs",
--- a/packages/shared/src/__tests__/deriveState.test.ts
+++ b/packages/shared/src/__tests__/deriveState.test.ts
@@ -1,0 +1,28 @@
+import type { InitialState, Resources } from '@clerk/types';
+
+import { deriveState } from '../deriveState';
+
+describe('deriveState', () => {
+  const mockInitialState = {
+    userId: 'user_2U330vGHg3llBga8Oi0fzzeNAaG',
+    sessionId: 'sess_2j1R7g3AUeKMx9M23dBO0XLEQGY',
+    orgId: 'org_2U330vGHg3llBga8Oi0fzzeNAaG',
+  } as InitialState;
+
+  const mockResources: Resources = {
+    client: {} as Resources['client'],
+    user: { id: mockInitialState.userId } as Resources['user'],
+    session: { id: mockInitialState.sessionId } as Resources['session'],
+    organization: { id: mockInitialState.orgId } as Resources['organization'],
+  };
+
+  it('uses SSR state when !clerkLoaded and initialState is provided', () => {
+    expect(deriveState(false, {} as Resources, mockInitialState)).toEqual(mockInitialState);
+  });
+
+  it('uses CSR state when clerkLoaded', () => {
+    const result = deriveState(true, mockResources, undefined);
+    expect(result.userId).toBe(mockInitialState.userId);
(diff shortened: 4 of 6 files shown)


packages/shared/src/deriveState.ts
```diff
@@ -0,0 +1,71 @@
...
+export const deriveState = (clerkLoaded: boolean, state: Resources, initialState: InitialState | undefined) => {
+  if (!clerkLoaded && initialState) {
+    return deriveFromSsrInitialState(initialState);
+  }
+  return deriveFromClientSideState(state);
+};
+
+const deriveFromSsrInitialState = (initialState: InitialState) => {
+  const userId = initialState.userId;
+  const user = initialState.user as UserResource;
```
LekoArts, line 20: Why do we need to typecast some properties in this function?

packages/shared/src/loadClerkJsScript.ts
LekoArts, line 1: Can we add unit tests for these functionalities? 🙏

wobsoriano: Hi Lennart! It's still in WIP but I'll definitely add one 🫡

packages/shared/src/deriveState.ts
wobsoriano: Good question. Thanks for pointing this out! It was adapted from the React SDK, and I maintained the original structure (including typecasts).

My guess is we want to narrow the types if `InitialState` is provided? And also some of them are typed as strings (i.e. `orgRole`, `orgPermissions`) and typecasting it just for semantics / future-proofihng? Maybe @panteliselef 
 can help me answer this question 😄

packages/shared/src/loadClerkJsScript.ts
wobsoriano: Tests added 🧪

wobsoriano, line 1: This function is extracted directly from the [React SDK](https://github.com/clerk/javascript/blob/main/packages/react/src/utils/loadClerkJsScript.ts). The only difference is that the [`versionSelector`](https://github.com/clerk/javascript/blob/15f5980089eff8dbd8e18fd5733b3702df48c28d/packages/react/src/utils/loadClerkJsScript.ts#L65) is removed, which may not be needed for most cases (?) and uses a compile-time env variable (`PACKAGE_VERSION`).

I'm thinking of using and it moving to shared as well, but remove the compile-time variable. What do you think @LekoArts @panteliselef?

packages/astro/src/server/build-clerk-hotload-script.ts
wobsoriano, line 1: Notice that we're not using the shared `loadClerkJsScript` function here but instead the other script url and attribute building functions, which comes from the shared `loadClerkJsScript` export as well, to clean up some duplicate logic. We did this because we're building the script server-side so we can safely pass env vars.

If we want to use the main `loadClerkJsScript` function, we will have to refactor and exclude the hotload script in the server response transform.

packages/astro/src/server/clerk-middleware.ts
```diff
@@ -279,10 +277,6 @@ async function decorateRequest(
             controller.enqueue(clerkAstroData);
             controller.enqueue(clerkSafeEnvVariables);
 
-            if (__HOTLOAD__) {
-              controller.enqueue(hotloadScript);
-            }
-
```
wobsoriano, line 285: We are no longer attaching the hotload script to the Response since we're using the `loadClerkJsScript` shared function in client instead

packages/astro/src/client/index.ts
```diff
@@ -1,11 +1,14 @@
-import { waitForClerkScript } from '../internal/utils/loadClerkJSScript';
+import { loadClerkJsScript } from '@clerk/shared';
```
LekoArts, line 1:
```suggestion
import { loadClerkJsScript } from '@clerk/shared/loadClerkJsScript';
```

packages/astro/src/client/index.ts
```diff
@@ -14,7 +17,10 @@ export const createClerkInstance = runOnce(createClerkInstanceInternal);
 export async function createClerkInstanceInternal(options?: AstroClerkIntegrationParams) {
   let clerkJSInstance = window.Clerk;
   if (!clerkJSInstance) {
-    await waitForClerkScript();
+    await loadClerkJsScript({
+      clerkJSVersion: HARDCODED_LATEST_CLERK_JS_VERSION,
```
LekoArts, line 21: We'll need to remove this, I'll explain it in https://github.com/clerk/javascript/pull/3681/files#r1672868119

packages/astro/src/client/index.ts
```diff
@@ -14,7 +17,10 @@ export const createClerkInstance = runOnce(createClerkInstanceInternal);
...
+      ...(options as any),
```
LekoArts, line 22: The `as any` is a code smell, ideally the `options` has the correct type, could be a footgun in the future

packages/astro/src/internal/merge-env-vars-with-params.ts
```diff
@@ -14,13 +14,21 @@ const mergeEnvVarsWithParams = (params?: AstroClerkIntegrationParams & { publish
...
+  // We have an eslint config that requires us to declare env variables in the turbo.json file.
+  // We're skipping/disabling it below as we use it only for the Astro integration.
   return {
-    signInUrl: paramSignIn || import.meta.env.PUBLIC_ASTRO_APP_CLERK_SIGN_IN_URL,
-    signUpUrl: paramSignUp || import.meta.env.PUBLIC_ASTRO_APP_CLERK_SIGN_UP_URL,
-    isSatellite: paramSatellite || import.meta.env.PUBLIC_ASTRO_APP_CLERK_IS_SATELLITE,
-    proxyUrl: paramProxy || import.meta.env.PUBLIC_ASTRO_APP_CLERK_PROXY_URL,
-    domain: paramDomain || import.meta.env.PUBLIC_ASTRO_APP_CLERK_DOMAIN,
-    publishableKey: paramPublishableKey || import.meta.env.PUBLIC_ASTRO_APP_CLERK_PUBLISHABLE_KEY || '',
+    // eslint-disable-next-line turbo/no-undeclared-env-vars
```
LekoArts, line 20: Nit: Use block comments (https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1) because then it's more concise

packages/shared/src/deriveState.ts
LekoArts: I asked because then I think `initialState` should be typed correctly in the first place 😊 

Maybe spend a little bit of time checking if we can fix this on a holistic level, otherwise the type-casts will be fine for now (if other things break)

packages/shared/src/loadClerkJsScript.ts
LekoArts: The `loadClerkJsScript` function needs to behave exactly like the the React SDK one, including the `versionSelector` functionality. Otherwise things like our `snapshot` command won't work since the SDK will load the incorrect version of `clerk-js` on the client.

With some changes it'll also ensure that the correct major version of `clerk-js` is loaded for the major version of the SDK. With the current hardcoded `5` version you could run into the problem of mismatched versions.

If `versionSelector` is moved to the `shared` package we'll need to change the `packageVersion = PACKAGE_VERSION` fallback in the args and **most likely** need to change it to `JS_PACKAGE_VERSION`. Test this!
Reason: `clerk-react` currently has the same major version as clerk-js, so the logic is sound here. But `shared` is separate from that, so we should probably replace it to that

LekoArts (changes requested):

packages/astro/src/internal/merge-env-vars-with-params.ts
```diff
@@ -14,13 +14,21 @@ const mergeEnvVarsWithParams = (params?: AstroClerkIntegrationParams & { publish
     ...rest
   } = params || {};
 
+  // We have an eslint config that requires us to declare env variables in the turbo.json file.
+  // We're skipping/disabling it below as we use it only for the Astro integration.
```
LekoArts, line 18: You could also add `'turbo/no-undeclared-env-vars': ['error', { allowList: ['import.meta.env'] }],` or something like that to the ESLint config

packages/shared/src/loadClerkJsScript.ts
wobsoriano: Thanks for taking some time to answer this @LekoArts! This makes sense to me.

I have a question for this one - is there gonna be a time that `clerk-js` and `clerk-react` will have different versions (other than snapshots)?

packages/shared/src/versionSelector.ts
wobsoriano, line 1: As @LekoArts suggested [here](https://github.com/clerk/javascript/pull/3681#discussion_r1673544525), the `loadClerkJsScript` should behave the same way like the React SDK because commands like `snapshot` won't work since the SDK will load an incorrect version of `clerk-js`.

The default `packageVersion` prop is now set to the `clerk-js` version (`JS_PACKAGE_VERSION`) instead of the `clerk-react` version (`PACKAGE_VERSION`).

With this, all SDKs should behave the same with the `clerk-js` version. Thanks!

packages/astro/src/server/clerk-middleware.ts
panteliselef: You can still use `loadClerkJsScript` in the client but we will expect to read the existing script.

packages/shared/src/loadClerkJsScript.ts
panteliselef: Given the fact that `versionSelector` will be framework specific i think we need to come up with a better way of setting it. I believe each framework should be able to specific its own logic.

@wobsoriano It could happen, for example let's say we wanna a new major for clerk-react because we are switching to react19 and using some of the new features. In that scenario we wouldn't have to cut a new major for clerk-js.

wobsoriano:
Closing this but it will be tackled in the future at a more relaxing pace.

PR #3682, opened 2024-07-09
chore(astro): Drop convenience Astro wrappers for React

wobsoriano:
## Description

We have 3 ways of importing Astro/React Clerk components (docs WIP):

1. `import { } from '@clerk/astro/components/{control|interactive|unstyled}'` - Fully Astro components.
2. `import { } from '@clerk/astro/client/react'` - React components. Needs the [`client:load`](https://docs.astro.build/en/reference/directives-reference/#clientload) directive when imported in an Astro component to be hydrated.
3. `import { } from '@clerk/astro/components/react` - Astro components that wrap option 2 with the `client:load` directive for convenience.

Multiple import methods can confuse users. Some features, like the`asChild` prop for unstyled components, don't work when React components are used in Astro due to Astro's [slot](https://docs.astro.build/en/basics/astro-components/#slots) system and islands architecture. As a result, we're discontinuing option 3.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


panteliselef (approved):
🧹

PR #3693, opened 2024-07-11
chore(e2e): Test control components for Astro and Astro + React

panteliselef:
## Description

This PR also changes `astro.node.withEmailCodes` to `astro.node.withCustomRoles`. Additionally the `withCustomRoles` now will use the production Backend API

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


integration/templates/astro-node/src/components/page-with-user.tsx
wobsoriano, line 1: Could we use the [custom store hook](https://github.com/clerk/javascript/blob/d006499c28a493573443d53900a84ba564b61922/packages/astro/src/client/react/utils.tsx#L11) that was previously created for this? Or is the intentional?

integration/templates/astro-node/src/components/page-with-user.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { type PropsWithChildren, useSyncExternalStore } from 'react';
+import { UserProfile } from '@clerk/astro/client/react';
+import { $userStore } from '@clerk/astro/client/stores';
+
+export function PageWithUser({ children }: PropsWithChildren) {
+  const user = useSyncExternalStore($userStore.listen, $userStore.get, $userStore.get);
```
wobsoriano, line 6: Could we use the [custom store hook](https://github.com/clerk/javascript/blob/d006499c28a493573443d53900a84ba564b61922/packages/astro/src/client/react/utils.tsx#L11) that was previously created for this? Or is the intentional?

panteliselef: We are not exporting that in user land tho. I would expect people that use `@clerk/astro` to create their own hook or use `@nanostores/react`. In the future we could have `useUser` exported from `@clerk/astro`

nikosdouvlis (approved):

PR #3698, opened 2024-07-11
chore(astro): Move safe id generation to a shared function

wobsoriano:
## Description

This PR tackles ECO-22 and consolidates ID generation with a custom utility function

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


anagstef (approved):

packages/astro/src/astro-components/unstyled/SignInButton.astro
```diff
@@ -3,9 +3,9 @@ import type { HTMLTag, Polymorphic } from 'astro/types'
 import type { SignInProps } from "@clerk/types";
 type Props<Tag extends HTMLTag = HTMLTag> = Polymorphic<SignInProps & { as: Tag; mode?: 'redirect' | 'modal' }>
 
-import { customAlphabet, urlAlphabet } from "nanoid";
+import { generateSafeId } from '@clerk/astro/internal';
```
dimkl, line 6: ❓ Since this import is already inside the astro package why do we use the `'@clerk/astro/internal';` instead of relative path?

dimkl (approved):

packages/astro/src/astro-components/unstyled/SignInButton.astro
panteliselef: This is a very nice question. Anything that is inside `astro-components/*` will be just copied to `components/*` and will be accessible by `@clerk/astro/components/*`.

Because there is no building step here, the relative paths will not be resolvable by the host application.

panteliselef (approved):

PR #3702, opened 2024-07-12
fix(elements): ensure updated provided values are sent to the machine

alexcarpenter:
## Description

Fix an issue when the elements input has a controlled value, it was not getting updated when the value changed. Causing stale values to persist.

Renamed the prop to `providedValue` as its not only used for the `initialValue`.

Example usage in AIO components rebuild to support the country code being separated from the formatted phone number.

```tsx
function Example() {
  const [val, setVal] = useState('')
  return (
    <>
      // Hidden elements input
      <Common.Input hidden value={val} />
      <input value={val} onChange={e => setValue(e.target.value}} />
    </>
  )
}
```

BEFORE

https://github.com/user-attachments/assets/79554a60-5ede-4b36-a133-b78b3e45d831

AFTER

https://github.com/user-attachments/assets/e0b833d8-8e49-49a5-a83b-d6c925300d11

https://linear.app/clerk/issue/SDKI-136/ensure-controlled-inputs-value-is-up-to-date-and-sent-to-machine-on

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


octoper (approved):
💯

tmilewski (approved):
💯

PR #3711, opened 2024-07-15
feat(astro): Support `auth().redirectToSignIn()` at page level

panteliselef:
## Description

Support `Astro.locals.auth().redirectToSignIn()`
This allows for redirectingToSignIn at the page level and provides parity with Nextjs.

Note: We are not exposing `auth().protect()` because DX is taking a hit, because we cannot just throw a response in astro, we need to return it. And we are not including protect at the middleware level so we don't confuse people that would expect it to work at page level.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/astro/src/server/clerk-middleware.ts
```diff
@@ -218,14 +224,38 @@ Check if signInUrl is missing from your configuration or if it is not an absolut
...
+
+      return createRedirect({
+        redirectAdapter,
+        devBrowserToken: devBrowserToken,
+        baseUrl: clerkUrl.toString(),
+        publishableKey: getSafeEnv(context).pk!,
+        signInUrl: requestState.signInUrl,
+        signUpUrl: requestState.signUpUrl,
+      }).redirectToSignIn({
+        returnBackUrl: opts.returnBackUrl === null ? '' : opts.returnBackUrl || clerkUrl.toString(),
```
wobsoriano, line 251: Does the `redirectToSignUp` function commonly used?

wobsoriano (approved):
Looks straightforward, just left a question 🚀

packages/astro/src/server/clerk-middleware.ts
panteliselef: We've build it, but no public API consumes it. and this implementation aims to mirror the API from @clerk/nextjs.

wobsoriano: Thanks for answering!

PR #3714, opened 2024-07-15
fix(ui): Inconsistent focus behaviour

joe-bell:
## Description

- Fixes `ToggleButton`s displaying `focus-visible` styles on click within Safari (switching to RAC's attributes solved the issue here)
  - I chose to preserve `native` as the default until we revisit in the polishing milestone
- Removes `:focus` in favour of `:focus-visible` where present
- Adjusts the `Card.Logo` to match other elements  

image

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


alexcarpenter (approved):
👍🏼 yeah, ideally we use the same strategy for focus visible across all components. We could use the useFocusVisble hook for non RAC components https://react-spectrum.adobe.com/react-aria/useFocusVisible.html#:~:text=useFocusVisible%20handles%20focus%20interactions%20for,then%20focus%20is%20not%20visible.

PR #3715, opened 2024-07-15
feat(ui): Add email link strategy for SignIn

joe-bell:
## Description

Enable `<SignIn.Strategy name='email_link'>` with resend functionality

CleanShot 2024-07-15 at 15 50 04@2x
CleanShot 2024-07-15 at 15 49 47@2x

> [!NOTE]
>
> Whilst the UI works as expected, no email will be sent due to the lack of a `redirect_url`. This will be tackled in a follow-up ticket once addressed upstream

### Additional Changes

I switched all the `aria-label`s for the pencil icon to "Start again" as this felt more appropriate; this might be a good candidate for extraction later. I'll create a follow-up ticket to translate in our polishing milestone

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/components/sign-in/sign-in.tsx
```diff
@@ -208,7 +208,7 @@ export function SignInComponentLoaded() {
                             <button
                               type='button'
                               className='text-accent-9 focus-visible:ring-default size-4 rounded-sm outline-none focus-visible:ring-2'
-                              aria-label='Edit email address'
+                              aria-label='Start again'
```
alexcarpenter, line 211: Are there localized values for these already by chance?

alexcarpenter (approved):
Non-blocking comment on localization but LGTM

packages/ui/src/components/sign-in/sign-in.tsx
joe-bell: Not yet! Already created a follow-up ticket to come back to this (and other areas) before we ship!

https://linear.app/clerk/issue/DES-1118/common-ensure-all-componentscommonprimitives-are-compatible-with-i18n

PR #3723, opened 2024-07-15
fix(clerk-js): Improve UX for recoverable actions in ConnectedAccounts

panteliselef:
### Before
image

### Now, when user needs to reauthorize/reconnect
Unify the errors for reauthorization vs attempting to create a new external account for the user. UX-wise the user will have to take the same action in order to recover, so we should not overwhelm them with information that is clearly more useful to developers than end users
image

and on mobile
image

### Now, not recoverable state
If an error indicate that the user cannot recover, only then show the text in red. Without the "requires action" badge
image

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/localizations/src/en-US.ts
```diff
@@ -458,6 +458,7 @@ export const enUS: LocalizationResource = {
       'Sign up unsuccessful due to failed security validations. Please refresh the page to try again or reach out to support for more assistance.',
     captcha_unavailable:
       'Sign up unsuccessful due to failed bot validation. Please refresh the page to try again or reach out to support for more assistance.',
+    external_account_missing_refresh_token: 'This account has been disconnected',
```
chanioxaris: A better message might be the following:
`This account is missing a refresh token` or `There is no associated refresh token with this account`

Maybe we need to provide the steps that the user needs to do in order to resolve this e.g. Try to reconnect your account or sign in

packages/localizations/src/ar-SA.ts
```diff
@@ -305,6 +305,10 @@ export const arSA: LocalizationResource = {
       title: 'التحقق من البريد الإلكتروني',
     },
     emailLink: {
+      clientMismatch: {
```
chanioxaris, line 308: ❓ Are these changes related to connected accounts?

panteliselef: No, i had to run the script that autogenerates all other localization with `undefined` but seems like when `clientMismatch` was added we never run this. It will be a lot manual work to revert them, and they are not causing any harm.

packages/localizations/src/en-US.ts
panteliselef: I understand were you are coming from, but we are explicitly using a very simple message here, because this is something the end user will see. I am confident that unless you are a developer you are not aware of what a "refresh token" is, so the error message would not be that useful to you.

nikospapcom (approved):

PR #3724, opened 2024-07-15
feat(astro): Introduce Clerk loaded and loading React components

wobsoriano:
## Description

This PR tackles ECO-43 and adds `<ClerkLoaded />` and `<ClerkLoading />` React components.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/astro/src/react/controlComponents.tsx
```diff
@@ -1,8 +1,9 @@
 import type { CheckAuthorizationWithCustomPermissions } from '@clerk/types';
 import type { PropsWithChildren } from 'react';
 
+import { $csrState } from '../stores/internal';
```
wobsoriano, line 4: Importing this from internal unless we decide to export the Clerk loading status as another store to user

wobsoriano: I actually think exporting the loading status (`$isLoadingClerkStore`) would help userland implement their own stuff and not do hacky stuff to check if Clerk loaded status

integration/tests/astro/react/components.test.ts
```diff
@@ -67,5 +67,13 @@ testAgainstRunningApps({ withPattern: ['astro.node.withCustomRoles'] })(
       await expect(u.page.getByText('Go to this page to see your profile')).toBeVisible();
       await expect(u.page.getByText('Sign out!')).toBeVisible();
     });
+
+    test('render content based on Clerk loaded status', async ({ page, context }) => {
+      const u = createTestUtils({ app, page, context });
+      await u.page.goToRelative('/utility');
+      await expect(u.page.getByText('Clerk is loading')).toBeVisible();
+      await u.page.waitForClerkJsLoaded();
+      await expect(u.page.getByText('Clerk is loaded')).toBeVisible();
```
panteliselef, line 76: Shall we do
```suggestion
      await u.page.goToRelative('/utility');
      await expect(u.page.getByText('Clerk is loading')).toBeVisible();
      await expect(u.page.getByText('Clerk is loaded')).not.toBeVisible();
      await u.page.waitForClerkJsLoaded();
      await expect(u.page.getByText('Clerk is loaded')).toBeVisible();
      await expect(u.page.getByText('Clerk is loading')).not.toBeVisible();
```

packages/astro/src/react/controlComponents.tsx
panteliselef: in user land can't they do 

```tsx
import { $clerkStore } from '@clerk/astro/client';

const loadedClerk = () => {
  return new Promise((res) => {
    $clerkStore.subscribe((clerk) => {
      if(clerk.loaded) res(clerk)
    } )
  });
}
```

or


```tsx

 // outside the component
 $isLoadingClerkStore = computed($clerkStore,(clerk) => clerk.loaded)
 // inside the component
 const { isLoaded } = useStore($isLoadingClerkStore);

```

packages/astro/src/react/controlComponents.tsx
```diff
@@ -21,6 +22,22 @@ export function SignedIn(props: PropsWithChildren) {
...
+};
+
+export const ClerkLoading = ({ children }: React.PropsWithChildren<unknown>): JSX.Element | null => {
+  const { isLoaded } = useStore($csrState);
+  if (isLoaded) {
+    return null;
+  }
+  return <>{children}</>;
+};
+
```
panteliselef, line 40: Seems like we are making use of the wrong `useStore` hook here. That one is using authAsyncStorage during SSR which may cause issues during hydration as i am not expecting `$csrState` and `authAsyncStorage.getStore()` to match. Also `isLoaded` from useAuth() is not an indication that clerk-js has loaded, but an indication that `useAuth()` can be safely accessed. In a SSR app `isLoaded` will always be true even when clerk-js has not loaded yet.

panteliselef (changes requested):
Added a comment which explains why we cannot use that `useStore` hook

integration/tests/astro/react/components.test.ts
wobsoriano: Yep, makes sense

packages/astro/src/react/controlComponents.tsx
wobsoriano: Thanks for explaining this again. Also I realized that we dont really need the features of this custom `useStore` since clerk loading and loaded components expects to be working on client only.

I'm exporting the `useStore` have in the react utils file and use inside control components, which will also be used by other control components if we decide to add more

packages/astro/src/react/controlComponents.tsx
wobsoriano: Thanks. I remember we talked about this in ECO-4

packages/astro/src/react/controlComponents.tsx
```diff
@@ -1,8 +1,11 @@
 import type { CheckAuthorizationWithCustomPermissions } from '@clerk/types';
+import { computed } from 'nanostores';
 import type { PropsWithChildren } from 'react';
 
+import { $clerkStore } from '../client';
```
panteliselef, line 5:
```suggestion
import { $clerkStore } from '../stores/external';
```

panteliselef (approved):
I added a minor correction. 💯

.changeset/thick-buttons-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/astro": patch
+---
+
+Introduce Clerk loaded and loading React components
```
panteliselef, line 5:
```suggestion
Introduce `<ClerkLoaded/>` and `<ClerkLoading/>` React components
```

packages/astro/src/react/controlComponents.tsx
```diff
@@ -24,7 +24,7 @@ export function SignedIn(props: PropsWithChildren) {
   return props.children;
 }
 
-const $isLoadingClerkStore = computed($clerkStore, clerk => clerk?.loaded);
+const $isLoadingClerkStore = computed($csrState, state => state.isLoaded);
```
wobsoriano, line 27: @panteliselef I replaced this to use the `isLoaded` property from the CSR state, since it gives more accurate status and we set the value right after Clerk has been loaded [here](https://github.com/clerk/javascript/blob/944bb2e23ccd6187bf1ad623b48629e8c2b71442/packages/astro/src/internal/create-clerk-instance.ts#L35).

The `$clerkStore` gives unreliable loaded status, maybe because we're setting it only once [here](https://github.com/clerk/javascript/blob/944bb2e23ccd6187bf1ad623b48629e8c2b71442/packages/astro/src/internal/create-clerk-instance.ts#L27)? And for some reason, nanostore freezes the object?

Even though the tests are passing on my local, it is still failing on CI 🤔

PR #3728, opened 2024-07-16
fix(ui): Redirect landing page buttons to new UI

joe-bell:
## Description

Clicking on these landing page buttons will open up the current UI rather than our rebuilt components; this PR amends that behaviour 

In the future, we'll seek to re-add these components, but it makes more sense to keep within our own flow for now

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/ui/theme-builder/package.json
+++ b/packages/ui/theme-builder/package.json
@@ -21,7 +21,7 @@
     "@radix-ui/react-tooltip": "^1.1.1",
     "bezier-easing": "^2.1.0",
     "change-case": "^5.4.4",
-    "clsx": "^2.1.1",
+    "cva": "^1.0.0-beta.1",
     "colorjs.io": "^0.5.0",
     "next": "14.2.4",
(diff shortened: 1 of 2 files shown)


packages/ui/theme-builder/package.json
```diff
@@ -21,7 +21,7 @@
     "@radix-ui/react-tooltip": "^1.1.1",
     "bezier-easing": "^2.1.0",
     "change-case": "^5.4.4",
-    "clsx": "^2.1.1",
```
joe-bell, line 24: We aren't actually using `clsx` in the `theme-builder`, so we can safely remove

joe-bell:
Fixing here instead:
https://clerkinc.slack.com/archives/C076A87HCM6/p1721136457685309?thread_ts=1721135307.131459&cid=C076A87HCM6

PR #3751, opened 2024-07-17
fix(ui): Format safe identifier phone number during sign in verification step

alexcarpenter:
## Description

Add formatSafeIdentifier util to format phone number during sign in verification step

BEFORE:
Screenshot 2024-07-17 at 5 29 29 PM

AFTER:
Screenshot 2024-07-17 at 5 29 13 PM

https://linear.app/clerk/issue/SDKI-121/signin-verifications-phone-number-should-be-formatted-according-to

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/ui/src/components/sign-in/sign-in.tsx
+++ b/packages/ui/src/components/sign-in/sign-in.tsx
@@ -10,7 +10,6 @@ import { EmailOrUsernameOrPhoneNumberField } from '~/common/email-or-username-or
 import { OTPField } from '~/common/otp-field';
 import { PasswordField } from '~/common/password-field';
 import { PhoneNumberField } from '~/common/phone-number-field';
-import { parsePhoneString } from '~/common/phone-number-field/utils';
 import { PhoneNumberOrUsernameField } from '~/common/phone-number-or-username-field';
 import { UsernameField } from '~/common/username-field';
 import { useAttributes } from '~/hooks/use-attributes';
@@ -25,7 +24,7 @@ import * as Icon from '~/primitives/icon';
 import { LinkButton } from '~/primitives/link-button';
 import { SecondaryButton } from '~/primitives/secondary-button';
 import { Seperator } from '~/primitives/seperator';
-import { isPhoneNumber } from '~/utils/is-phone-number';
+import { formatSafeIdentifier } from '~/utils/format-safe-identifier';
 
 export function SignInComponent() {
   return (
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/sign-in/sign-in.tsx
```diff
@@ -204,7 +205,11 @@ export function SignInComponentLoaded() {
                       <Card.Description>{t('signIn.password.subtitle')}</Card.Description>
                       <Card.Description>
                         <span className='flex items-center justify-center gap-2'>
-                          <SignIn.SafeIdentifier />
+                          <SignIn.SafeIdentifier
+                            transform={val => {
+                              return formatSafeIdentifier(val) || val;
```
dstaley, line 210: If we add some overloads to the `formatSafeIdentifier` function to tell TypeScript that it will always return a string if provided a string, we can remove this logical OR which won't ever be invoked since `SafeIdentifier` falls back to an empty string in the worse case.

[TS Playground](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABMOAnAtgQygZU8AUwEkATAsWYGA1ACgGcpUAuRMEAGw4EpX2uA3AChQkWAmRosufMTIUYVGgyatwZKmAIleidQU3bho6PCQoM2PIVLlK1OoxaInMMAHNdrj8fCmJFtLWcnaKDirO3u6IAD5snByxemAabtrciADeQoi5iIqItACETrFxTgB0bhAcIGT0tADkAAKNZS5MFdIQABa0APQA2pgAtABeAIIjAFoAuv3ci1k5eauoBFAgqEhOwqsAvkIrueub24iNjcKHQhAIjPn0EzhMbtEAvJKWMjby9sqXbjCO5gB4wegAOQSiE+gSsslsCiUdH4PGB9ygjwAqikDGkSDCvkEEX8wsp9IYdAIgA)

dstaley (approved):
approved, but if you want to get rid of that logical OR we can by adding a few overloads to `formatSafeIdentifier`

packages/ui/src/utils/format-safe-identifier.ts
```diff
@@ -0,0 +1,15 @@
+import { stringToFormattedPhoneString } from '~/common/phone-number-field/utils';
```
joe-bell, line 1: Nit: I think we should move this to `~/utils` if it's something we plan to use across different areas of the codebase; this'll avoid the likelihood of circular imports (and just a little tidier)

I may go ahead and just make this change to avoid the back-and-forth

joe-bell: Ooof actually just seen how this also relies on `data` etc. too. Let's come back to this

joe-bell (approved):

nikosdouvlis (approved):

PR #3758, opened 2024-07-18
chore(nextjs): Introduce build-time error for server-only helpers on client-side

LauraBeatris:
## Description

Resolves https://github.com/clerk/javascript/issues/3660

Uses the `server-only` package to throw a build-time error if Clerk server-side helpers get accidentally imported on the client side. 

CleanShot 2024-07-18 at 09 57 12@2x

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/src/app-router/server/auth.ts
+++ b/packages/nextjs/src/app-router/server/auth.ts
@@ -1,3 +1,5 @@
+import 'server-only';
+
 import type { AuthObject, RedirectFun } from '@clerk/backend/internal';
 import { constants, createClerkRequest, createRedirect } from '@clerk/backend/internal';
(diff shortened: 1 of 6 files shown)


packages/nextjs/src/app-router/server/auth.ts
```diff
@@ -1,3 +1,5 @@
+import 'server-only';
```
LauraBeatris, line 1: I've import it separately on `auth` and `currentUser` only instead of importing it once on `server` index. 

My concern is that the error would be thrown even when importing type definitions, but I'd like to minimize the surface here.

tmilewski (approved):

PR #3774, opened 2024-07-19
fix(clerk-js): Set __session and __client_uat before invalidating cache

nikosdouvlis:
## Description

Fixes the following flow that mainly affected nextjs apps:

Setup:
- Dev instance
- Latest next, clerk/nextjs and clerk-js
- Multi-app-same-domain FF disabled
Steps:
- Go to app
- Enter email and hit continue (signIn.create):  client_uat is 0
- Enter password and hit continue (signIn.attemptFirstFactor) : client_uat > 0 (for accounts)
- Clerk-js invalidates cache : **client_uat is 0 because  clerk-js does not set it correctlly**

The main issue in the clerk-js codebase was that during sign-in/sign-up, we set the cookies **implicitly** when we update the current Client resource (Base -> updateClient -> Client.fromJson() -> new Session -> hydrate token cache).  However, the Clerk.client only gets updated AFTER the token cache is hydrated, so setting the client_uat token failed because when it runs, Clerk.client.activeSessions is still 0 (its still the previous value)

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/poor-rockets-peel.md
+++ b/.changeset/poor-rockets-peel.md
@@ -1,0 +1,5 @@
+---
+"@clerk/clerk-js": patch
+---
+
+Set **session and **client_uat before invalidating cache during setActive flow
--- a/.changeset/poor-rockets-peel.md
+++ b/.changeset/poor-rockets-peel.md
@@ -2,4 +2,4 @@
 "@clerk/clerk-js": patch
 ---
 
-Set **session and **client_uat before invalidating cache during setActive flow
+Set `session` and `client_uat` before invalidating cache during `setActive()` flow
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -720,6 +720,10 @@ export class Clerk implements ClerkInterface {
       eventBus.dispatch(events.TokenUpdate, { token: null });
     }
 
+    if (session?.lastActiveToken) {
+      eventBus.dispatch(events.TokenUpdate, { token: session.lastActiveToken });
+    }
+
     await onBeforeSetActive();
 
     //1. setLastActiveSession to passed user session (add a param).
@@ -755,6 +759,7 @@ export class Clerk implements ClerkInterface {
     }
 
     this.#setAccessors(newSession);
+
     this.#emit();
     await onAfterSetActive();
     this.#resetComponentsState();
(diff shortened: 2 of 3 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -720,6 +720,10 @@ export class Clerk implements ClerkInterface {
       eventBus.dispatch(events.TokenUpdate, { token: null });
     }
 
+    if (session?.lastActiveToken) {
+      eventBus.dispatch(events.TokenUpdate, { token: session.lastActiveToken });
+    }
+
```
dimkl, line 726: I don't think this will fix the issue because the `TokenUpdate` event is already triggered during session instance initialization when the `lastActiveToken` exists: https://github.com/clerk/javascript/blob/3c8fda28cfc99d6266397b75ebe1e6161192283c/packages/clerk-js/src/core/resources/Session.ts#L44

nikosdouvlis: I've updated the description, please take a look

anagstef (approved):

.changeset/poor-rockets-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/clerk-js": patch
+---
+
+Set **session and **client_uat before invalidating cache during setActive flow
```
LekoArts, line 5:
```suggestion
Set `session` and `client_uat` before invalidating cache during `setActive()` flow
```

packages/clerk-js/src/core/clerk.ts
dimkl: ok. it makes more sense now that this will fix the current issue.
Can we add some comments above this change that describes why we need this line of code?
Also instead of applying this change, why don't we introduce a separate event  `ClientUpdate` to handle the clientUat cookie when the `Clerk.client` is updated?

dimkl: Also let's make sure that there is a test to cover that the issue has been fixed before merging this change

nikosdouvlis: > Also instead of applying this change, why don't we introduce a separate event ClientUpdate to handle the clientUat cookie when the Clerk.client is updated?

I'd say that the current implementation needs a holistic refactor so minor tweaks won't make a difference at the momennt

nikosdouvlis: Added a test 👍🏻

PR #3777, opened 2024-07-22
Proof of Concept: Draft DX for setActive taking an org slug

izaaklauer:
## What's the goal?

update setActive to allow it to take an organization slug as an identifier, rather than just the ID

### DIFF
--- a/packages/types/src/clerk.ts
+++ b/packages/types/src/clerk.ts
@@ -673,6 +673,13 @@ export type SignUpRedirectOptions = RedirectOptions &
     initialValues?: SignUpInitialValues;
   };
 
+/**
+ * OrganizationIdentifier uniquely identifies an organization within an instance
+ */
+export type OrganizationIdentifier =
+  | { id: string; slug?: never } // Via id
+  | { id?: never; slug: string }; // Via slug
+
 export type SetActiveParams = {
   /**
    * The session resource or session id (string version) to be set as active.
@@ -681,10 +688,16 @@ export type SetActiveParams = {
   session?: ActiveSessionResource | string | null;
 
   /**
-   * The organization resource or organization id (string version) to be set as active in the current session.
+   * @deprecated The organization resource or organization id (string version) to be set as active in the current session.
    * If `null`, the currently active organization is removed as active.
    */
   organization?: OrganizationResource | string | null;
+
+  /**
+   * The organization, identified via slug or ID, to be set as active in the current session.
+   * If `null`, the currently active organization is removed as active.
+   */
+  organizationIdentifier?: OrganizationIdentifier | null;
 
   /**
    * Callback run just before the active session and/or organization is set to the passed object.


packages/types/src/clerk.ts
```diff
@@ -673,6 +673,13 @@ export type SignUpRedirectOptions = RedirectOptions &
     initialValues?: SignUpInitialValues;
   };
 
+/**
+ * OrganizationIdentifier uniquely identifies an organization within an instance
+ */
+export type OrganizationIdentifier =
+  | { id: string; slug?: never } // Via id
+  | { id?: never; slug: string }; // Via slug
```
izaaklauer, line 681: This is my attempt at saying via typescript: "You can identify an org via a slug alone, or an ID alone, but not a slug and an ID, and not neither." Something like the excellent protobuf [oneof](https://github.com/hashicorp/waypoint/blob/b81cf848b44f57b95cb2e9766839582cc625d9c2/pkg/server/proto/server.proto#L1316-L1324).

I don't love this though - it doesn't read very clearly to me. I'm open to the argument that a plan-old object with both params as optional might make for a better api due to readibility.

alexcarpenter: I've used this util a couple of times recently from [type-fest](https://github.com/sindresorhus/type-fest/tree/main) RequireOneOrNone - https://github.com/sindresorhus/type-fest/blob/main/source/require-one-or-none.d.ts

tmilewski: I'm with @alexcarpenter. `type-fest` is solid.

mzhong9723, line 681: ❓ Is there a downside of supporting the case where _both_ organization id and slug and provided? Maybe to avoid dealing with the case where the slug and ID don't match?

izaaklauer: In my mind, the best thing would be to clearly communicate to the end-user that we need one or the other, ideally through the type system. 

I can't think of a reason why a caller would ever want to send both, apart from misunderstanding our docs and taking a belt-and-suspenders approach to specifying things. If we think that's likely to happen, I could see adding logic to check if they're both pointing to the same underlying org, and succeeding quietly. But that'd be my plan B!

packages/types/src/clerk.ts
```diff
@@ -681,11 +688,17 @@ export type SetActiveParams = {
...
    * If `null`, the currently active organization is removed as active.
    */
   organization?: OrganizationResource | string | null;
 
+  /**
+   * The organization, identified via slug or ID, to be set as active in the current session.
+   * If `null`, the currently active organization is removed as active.
+   */
+  organizationIdentifier?: OrganizationIdentifier | null;
+
```
izaaklauer, line 701: Here's an idea thanks to @gkats : https://github.com/clerk/clerk_go/pull/7174#discussion_r1687485284

It looks like bapi accepts a string that can be the slug _or_ the org id in other places, and it works fine. If we do that here, we don't need to change the API at all! 

I wouldn't have expected this to work, but as @mzhong9723 pointed out here: https://github.com/clerk/clerk_go/pull/7174#discussion_r1687162721 we have validation logic already that prevents a slug from sharing a name with an org ID. 

It strikes me that there's slightly elevated potential for both user confusion (e.g. a user is trying to specify an org ID that doesn't exist, but a slug that shares the name _does_ exist, so we return that), and bugs on our side (we need to make sure that we always test the string against both the org id _and_ the slug every time). But it still strikes me as the best path forward, seeing as we already have the validation logic and a precedent and it makes for a much smaller change!

izaaklauer:
We found precedent for taking an id OR a slug in the same string parameter, and we have validation to make sure that's safe on the backend, so we're now aiming for that route!

More context here: https://linear.app/clerk/issue/ORGS-17/accept-slug-from-setactive#comment-f0bf7438

Thank you @alexcarpenter  and @tmilewski for the tips! Type-fest seems great, and i'm looking forward to using it.

PR #3783, opened 2024-07-23
refactor(ui): Unified button component

joe-bell:
## Description

Merges our `Button` and `SecondaryButton` into a single `cva`-powered component

> [!NOTE]
> I intentionally decided not to extract variant-based values into custom properties at this stage; it felt better to cross that bridge when we need to 

I intentionally didn't try to extract out styles into custom properties 

### Screenshots

CleanShot 2024-07-23 at 16 52 50@2x
CleanShot 2024-07-23 at 16 52 37@2x
CleanShot 2024-07-23 at 16 52 27@2x
CleanShot 2024-07-23 at 16 52 11@2x

### Additional changes

- Rename `link-button` to `link` to align with other components (we don't have a `<Link />` just yet, but it felt right to avoid confusion with `Button`)
- Introduce a TailwindCSS fix that allows for default ring/width values (which were conflicting with variant-based values at the same specificity)

### Follow-up tasks

Completing this task opened up other areas for polish

- [ ] Consistent `ring` color rather than adhoc (see Linear)
- [ ] Include a `ghost` Button variant (see Linear)

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/primitives/button.tsx
```diff
@@ -1,41 +1,85 @@
...
     >
       {busy ? (
-        <Spinner className='min-w-0 shrink-0 text-[1.125rem]'>Loading…</Spinner>
+        <Spinner className='shrink-0 text-[1.125rem]'>Loading…</Spinner>
       ) : (
         <>
-          <span className='min-w-0 truncate'>{children}</span>
-          {icon && <span className='ms-2 min-w-0 shrink-0 text-[0.625rem] opacity-60'>{icon}</span>}
+          <span className='truncate'>{children}</span>
+          {icon && <span className='shrink-0 text-[0.625rem] text-inherit opacity-60'>{icon}</span>}
```
alexcarpenter, line 82: Secondary button persists the children text on busy and places the icon position at the start of the children text. Are we looking to keep parity there?

alexcarpenter: Screenshot 2024-07-23 at 12 49 17 PM

alexcarpenter (commented):
Looks good. One comment on the difference in icon position and busy state rendering between the primary and secondary variants.

packages/ui/src/primitives/button.tsx
joe-bell: Oof good spot, maybe we need an additional `hideChildrenOnBusy` prop or something – what do you think?

I can roll that out tomorrow if you approve, but equally happy if you'd rather I explore this further and allow you approve tomorrow

alexcarpenter: Thought on naming but couldn't come up with anything better. So use your best judgement. Approving ✅

alexcarpenter (approved):

PR #3786, opened 2024-07-23
fix(clerk-js): Use focus event to trigger session touch

BRKalow:
## Description

When tabs have different selected organizations, we rely on the `session.touch()` call to update the session with the tab's selected org. Previously, this relied on the `visibilitychange` event, which works when switching between tabs, but doesn't work if a user has two windows that are visible simultaneously.

This change shifts to using the `focus` event, which seems to be a better indication of a tab being interacted with. This guarantees that any time a user shifts back to interacting with a tab, its previously selected organization is set as active.

Additionally, this updates `Session.getToken()` to accept an `organizationId` parameter, allowing session tokens to be created with a specific organization ID, not only the session's active organization. This ensures calls to `getToken()` will pull the tab's selected organization in memory.

I've also updated the session cookie update flow to only update the cookie _if_ the tabs selected organization matches the session's active organization, and ensure that the session cookie is updated immediately on window focus to ensure the session cookie contains an org ID matching the tab's selected org.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


BRKalow:
!snapshot

clerk-cookie:
Hey @BRKalow - the snapshot version command generated the following package versions:
| Package | Version |
| --- | --- |
| @clerk/astro | 1.0.1-snapshot.v259851a |
| @clerk/backend | 1.4.0-snapshot.v259851a |
| @clerk/chrome-extension | 1.1.5-snapshot.v259851a |
| @clerk/clerk-js | 5.10.0-snapshot.v259851a |
| @clerk/elements | 0.10.5-snapshot.v259851a |
| @clerk/clerk-expo | 1.2.8-snapshot.v259851a |
| @clerk/express | 0.0.18-snapshot.v259851a |
| @clerk/fastify | 1.0.20-snapshot.v259851a |
| @clerk/localizations | 2.5.0-snapshot.v259851a |
| @clerk/nextjs | 5.2.5-snapshot.v259851a |
| @clerk/clerk-react | 5.2.9-snapshot.v259851a |
| @clerk/remix | 4.2.4-snapshot.v259851a |
| @clerk/clerk-sdk-node | 5.0.17-snapshot.v259851a |
| @clerk/shared | 2.4.0-snapshot.v259851a |
| @clerk/tanstack-start | 0.1.8-snapshot.v259851a |
| @clerk/testing | 1.2.0-snapshot.v259851a |
| @clerk/themes | 2.1.13-snapshot.v259851a |
| @clerk/types | 4.9.0-snapshot.v259851a |

`@clerk/astro`
```sh
npm i @clerk/astro@1.0.1-snapshot.v259851a --save-exact
```
`@clerk/backend`
```sh
npm i @clerk/backend@1.4.0-snapshot.v259851a --save-exact
```
`@clerk/chrome-extension`
```sh
npm i @clerk/chrome-extension@1.1.5-snapshot.v259851a --save-exact
```
`@clerk/clerk-js`
```sh
npm i @clerk/clerk-js@5.10.0-snapshot.v259851a --save-exact
```
`@clerk/elements`
```sh
npm i @clerk/elements@0.10.5-snapshot.v259851a --save-exact
```
`@clerk/clerk-expo`
```sh
npm i @clerk/clerk-expo@1.2.8-snapshot.v259851a --save-exact
```
`@clerk/express`
```sh
npm i @clerk/express@0.0.18-snapshot.v259851a --save-exact
```
`@clerk/fastify`
```sh
npm i @clerk/fastify@1.0.20-snapshot.v259851a --save-exact
```
`gatsby-plugin-clerk`
```sh
```
`@clerk/localizations`
```sh
npm i @clerk/localizations@2.5.0-snapshot.v259851a --save-exact
```
`@clerk/nextjs`
```sh
npm i @clerk/nextjs@5.2.5-snapshot.v259851a --save-exact
```
`@clerk/clerk-react`
```sh
npm i @clerk/clerk-react@5.2.9-snapshot.v259851a --save-exact
```
`@clerk/remix`
```sh
npm i @clerk/remix@4.2.4-snapshot.v259851a --save-exact
```
`@clerk/clerk-sdk-node`
```sh
npm i @clerk/clerk-sdk-node@5.0.17-snapshot.v259851a --save-exact
```
`@clerk/shared`
```sh
npm i @clerk/shared@2.4.0-snapshot.v259851a --save-exact
```
`@clerk/tanstack-start`
```sh
npm i @clerk/tanstack-start@0.1.8-snapshot.v259851a --save-exact
```
`@clerk/testing`
```sh
npm i @clerk/testing@1.2.0-snapshot.v259851a --save-exact
```
`@clerk/themes`
```sh
npm i @clerk/themes@2.1.13-snapshot.v259851a --save-exact
```
`@clerk/types`
```sh
npm i @clerk/types@4.9.0-snapshot.v259851a --save-exact
```

BRKalow:
!snapshot

.changeset/new-donkeys-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/clerk-js": patch
+---
+
+Fixes a bug where multiple tabs with different active organizations would not always respect the selected organization. Going forward, when a tab is focused the active organization will immediately be updated to the tab's last active organization.
```
BRKalow, line 5:
```suggestion
Fixes a bug where multiple tabs with different active organizations would not always respect the selected organization. Going forward, when a tab is focused the active organization will immediately be updated to the tab's last active organization.

Additionally, `Clerk.session.getToken()` now accepts an `organizationId` option. The provided organization ID will be used to set organization-related claims in the generated session token.
```

packages/clerk-js/src/ui/utils/test/mockHelpers.ts
```diff
@@ -13,20 +13,33 @@ type DeepJestMocked<T> = T extends FunctionLike
       }
     : T;
 
-const mockProp = <T>(obj: T, k: keyof T) => {
+type MockMap<T = any> = {
+  [K in { [K in keyof T]: T[K] extends (...args: any[]) => any ? K : never }[keyof T]]?: jest.Mock<
+    // @ts-expect-error -- the typing seems to be working in practice
+    T[K]
+  >;
+};
```
BRKalow, line 21: This type gives us a type-safe way to provide specific mock implementations to `mockMethodsOf`, see below for usage.

clerk-cookie:
Hey @BRKalow - the snapshot version command generated the following package versions:
| Package | Version |
| --- | --- |
| @clerk/astro | 1.0.5-snapshot.vf8fb0f6 |
| @clerk/backend | 1.5.0-snapshot.vf8fb0f6 |
| @clerk/chrome-extension | 1.1.8-snapshot.vf8fb0f6 |
| @clerk/clerk-js | 5.11.0-snapshot.vf8fb0f6 |
| @clerk/dev-cli | 0.0.4-snapshot.vf8fb0f6 |
| @clerk/elements | 0.11.0-snapshot.vf8fb0f6 |
| @clerk/clerk-expo | 2.0.1-snapshot.vf8fb0f6 |
| @clerk/express | 0.0.22-snapshot.vf8fb0f6 |
| @clerk/fastify | 1.0.24-snapshot.vf8fb0f6 |
| @clerk/localizations | 2.5.3-snapshot.vf8fb0f6 |
| @clerk/nextjs | 5.2.9-snapshot.vf8fb0f6 |
| @clerk/clerk-react | 5.3.0-snapshot.vf8fb0f6 |
| @clerk/remix | 4.2.8-snapshot.vf8fb0f6 |
| @clerk/clerk-sdk-node | 5.0.21-snapshot.vf8fb0f6 |
| @clerk/shared | 2.4.2-snapshot.vf8fb0f6 |
| @clerk/tanstack-start | 0.1.12-snapshot.vf8fb0f6 |
| @clerk/testing | 1.2.4-snapshot.vf8fb0f6 |
| @clerk/themes | 2.1.15-snapshot.vf8fb0f6 |
| @clerk/types | 4.10.0-snapshot.vf8fb0f6 |

`@clerk/astro`
```sh
npm i @clerk/astro@1.0.5-snapshot.vf8fb0f6 --save-exact
```
`@clerk/backend`
```sh
npm i @clerk/backend@1.5.0-snapshot.vf8fb0f6 --save-exact
```
`@clerk/chrome-extension`
```sh
npm i @clerk/chrome-extension@1.1.8-snapshot.vf8fb0f6 --save-exact
```
`@clerk/clerk-js`
```sh
npm i @clerk/clerk-js@5.11.0-snapshot.vf8fb0f6 --save-exact
```
`@clerk/dev-cli`
```sh
npm i @clerk/dev-cli@0.0.4-snapshot.vf8fb0f6 --save-exact
```
`@clerk/elements`
```sh
npm i @clerk/elements@0.11.0-snapshot.vf8fb0f6 --save-exact
```
`@clerk/clerk-expo`
```sh
npm i @clerk/clerk-expo@2.0.1-snapshot.vf8fb0f6 --save-exact
```
`@clerk/express`
```sh
npm i @clerk/express@0.0.22-snapshot.vf8fb0f6 --save-exact
```
`@clerk/fastify`
```sh
npm i @clerk/fastify@1.0.24-snapshot.vf8fb0f6 --save-exact
```
`gatsby-plugin-clerk`
```sh
```
`@clerk/localizations`
```sh
npm i @clerk/localizations@2.5.3-snapshot.vf8fb0f6 --save-exact
```
`@clerk/nextjs`
```sh
npm i @clerk/nextjs@5.2.9-snapshot.vf8fb0f6 --save-exact
```
`@clerk/clerk-react`
```sh
npm i @clerk/clerk-react@5.3.0-snapshot.vf8fb0f6 --save-exact
```
`@clerk/remix`
```sh
npm i @clerk/remix@4.2.8-snapshot.vf8fb0f6 --save-exact
```
`@clerk/clerk-sdk-node`
```sh
npm i @clerk/clerk-sdk-node@5.0.21-snapshot.vf8fb0f6 --save-exact
```
`@clerk/shared`
```sh
npm i @clerk/shared@2.4.2-snapshot.vf8fb0f6 --save-exact
```
`@clerk/tanstack-start`
```sh
npm i @clerk/tanstack-start@0.1.12-snapshot.vf8fb0f6 --save-exact
```
`@clerk/testing`
```sh
npm i @clerk/testing@1.2.4-snapshot.vf8fb0f6 --save-exact
```
`@clerk/themes`
```sh
npm i @clerk/themes@2.1.15-snapshot.vf8fb0f6 --save-exact
```
`@clerk/types`
```sh
npm i @clerk/types@4.10.0-snapshot.vf8fb0f6 --save-exact
```

packages/clerk-js/src/utils/pageLifecycle.ts
```diff
@@ -16,22 +16,22 @@ const noop = () => {
...
+    return { onPageFocus: noop };
   }
 
   const callbackQueue: Record<string, Array<() => void>> = {
-    'visibilitychange:visible': [],
+    focus: [],
   };
 
-  document.addEventListener('visibilitychange', () => {
+  window.addEventListener('focus', () => {
```
BRKalow, line 26: `focus` seems to be a better indicator of activity, and avoids an edge case with visibility where two windows are open side-by-side, both being visible.

packages/clerk-js/src/core/auth/AuthCookieService.ts
```diff
@@ -52,11 +52,16 @@ export class AuthCookieService {
   ) {
     // set cookie on token update
     eventBus.on(events.TokenUpdate, ({ token }) => {
+      // only update session cookie from active tab
+      if (!document.hasFocus()) {
+        return;
+      }
```
BRKalow, line 58: We might want more nuance here, e.g. checking if there is only a single tab open (even if in the background). The token is already refreshed on refocus of the tab, so the session cookie should quickly get refreshed.

Direct access of the cookie is discouraged, so in practice this _shouldn't_ be an issue...

dimkl: ☁️ I think we should not add this guard here since it will disable updating the cookie on background which may result in increasing the HTTP 401 errors shown in customers due to requests triggered in background of customers applications.
Also wouldn't we want to have the session updated on the following cases:
- different multiple browser windows open at the same time (or using split screen)
- having the tab open without having the document focused
- having the developer tools open (`document.hasFocus()` returns false)

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -151,28 +151,38 @@ export class Session extends BaseResource implements SessionResource {
...
     if (!template && Number(leewayInSeconds) >= 60) {
       throw new Error('Leeway can not exceed the token lifespan (60 seconds)');
     }
 
-    const tokenId = this.#getCacheId(template);
+    const tokenId = this.#getCacheId(template, organizationId);
     const cachedEntry = skipCache ? undefined : SessionTokenCache.get({ tokenId }, leewayInSeconds);
 
+    // Dispatch tokenUpdate only for __session tokens with the session's active organization ID, and not JWT templates
+    const shouldDispatchTokenUpdate = !template && options?.organizationId === Session.clerk.organization?.id;
```
dimkl, line 169: 🤔 I am kind of sceptic about using the ` Session.clerk.organization?.id` here :/.
Also why do we enforce that the token:update event should be dispatched only for the same organization id?

BRKalow: @dimkl We don't want the session cookie to contain an org ID that isn't the active one. So if someone provides a different org ID that does not match the active org for the session, we don't want the cookie to be updated.

packages/clerk-js/src/core/auth/AuthCookieService.ts
BRKalow: @dimkl Right, I think the check needs to be more robust, but I think we do need to include something. The issue is that two tabs / windows with different selected orgs would be updating the cookie at the same time. So the org ID in the cookie would not be consistent.

packages/clerk-js/src/core/resources/Session.ts
BRKalow: Do you have a recommendation for a better way to access the currently selected org?

dimkl: Could we take it from the session claims or from the `this.lastActiveOrganizationId`?
Also have we checked that the switch organization scenario updates the session cookie correctly or is the `token:update` event skipped and the cookie stays the same?

BRKalow: switching the organization via `setActive()` dispatches `TokenUpdate` separately, and I confirmed it updates the cookie as expected!

packages/clerk-js/bundlewatch.config.json
```diff
@@ -1,6 +1,6 @@
 {
   "files": [
-    { "path": "./dist/clerk.browser.js", "maxSize": "62kB" },
+    { "path": "./dist/clerk.browser.js", "maxSize": "63kB" },
```
BRKalow, line 3: I hit the limit 😢 bumping a single kB...

.changeset/new-donkeys-sneeze.md
```diff
@@ -0,0 +1,7 @@
+---
+"@clerk/clerk-js": patch
```
BRKalow, line 2:
```suggestion
"@clerk/clerk-js": minor
```

packages/clerk-js/src/core/auth/AuthCookieService.ts
```diff
@@ -110,27 +110,47 @@ export class AuthCookieService {
     this.poller.startPollingForSessionToken(() => this.refreshSessionToken());
   }
 
-  private refreshTokenOnVisibilityChange() {
-    document.addEventListener('visibilitychange', () => {
+  private refreshTokenOnFocus() {
+    window.addEventListener('focus', () => {
```
nikosdouvlis, line 114: We should either replace this event listener call here with a `PageLifecycle`  helper or ditch the `PageLifecycle` 
helper altogether as it feels a little unnecessary right now

packages/clerk-js/src/core/auth/AuthCookieService.ts
```diff
@@ -110,27 +110,47 @@ export class AuthCookieService {
...
       if (document.visibilityState === 'visible') {
-        void this.refreshSessionToken();
+        // Certain data-fetching libraries that refetch on focus (such as swr) use setTimeout(cb, 0) to schedule a task on the event loop.
+        // This gives us an opportunity to ensure the session cookie is updated with a fresh token before the fetch occurs, but it needs to
+        // be done with a microtask. Promises schedule microtasks, and so by using `updateCookieImmediately: true`, we ensure that the cookie
+        // is updated as part of the scheduled microtask. Our existing event-based mechanism to update the cookie schedules a task, and so the cookie
+        // is updated too late and not guaranteed to be fresh before the refetch occurs.
+        void this.refreshSessionToken({ updateCookieImmediately: true });
```
nikosdouvlis, line 121: Very helpful context here 👏🏻 
For anyone visiting this PR in the future, here are the main differences between tasks and microtasks:

> Tasks vs. microtasks
A task is anything scheduled to be run by the standard mechanisms such as initially starting to execute a script, asynchronously dispatching an event, and so forth. Other than by using events, you can enqueue a task by using [setTimeout()](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout) or [setInterval()](https://developer.mozilla.org/en-US/docs/Web/API/setInterval).
The difference between the task queue and the microtask queue is simple but very important:
When a new iteration of the event loop begins, the runtime executes the next task from the task queue. Further tasks and tasks added to the queue after the start of the iteration will not run until the next iteration.
Whenever a task exits and the execution context stack is empty, all microtasks in the microtask queue are executed in turn. The difference is that execution of microtasks continues until the queue is empty—even if new ones are scheduled in the interim. In other words, microtasks can enqueue new microtasks and those new microtasks will execute before the next task begins to run, and before the end of the current event loop iteration.

Taken from https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide/In_depth#run_javascript_run:~:text=Tasks%20vs.%20microtasks (which I recommend everyone to read)

packages/clerk-js/src/core/auth/AuthCookieService.ts
```diff
@@ -110,27 +110,47 @@ export class AuthCookieService {
...
+  } = {}): Promise<void> {
     if (!this.clerk.session) {
       return;
     }
 
     try {
-      await this.clerk.session.getToken();
+      const token = await this.clerk.session.getToken();
+      if (updateCookieImmediately) {
+        this.updateSessionCookie(token);
```
nikosdouvlis, line 138: Again, for any future readers: The reason the token update gets scheduled as a microtask here because it's part of the promise that was initialized when the **async** `refreshSessionToken` was called 

@BRKalow please correct me if I'm wrong here

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -151,28 +151,38 @@ export class Session extends BaseResource implements SessionResource {
...
-      if (!template) {
+      if (shouldDispatchTokenUpdate) {
         eventBus.dispatch(events.TokenUpdate, { token: cachedToken });
       }
       // Return null when raw string is empty to indicate that there it's signed-out
       return cachedToken.getRawString() || null;
     }
     const path = template ? `${this.path()}/tokens/${template}` : `${this.path()}/tokens`;
-    const tokenResolver = Token.create(path);
+    // TODO: update template endpoint to accept organizationId
```
nikosdouvlis, line 180: Is there a ticket for this? Just making sure we don't forget about it as it sounds pretty important :)

nikosdouvlis (approved):

packages/clerk-js/src/core/resources/Session.ts
BRKalow: Added! https://linear.app/clerk/issue/SDKI-620/update-v1clientsessionssession-idtokenstemplate-name-to-accept

packages/clerk-js/src/core/auth/AuthCookieService.ts
BRKalow: Agreed, good idea 👍

imsherrill:
@dimkl @nikosdouvlis 
is it possible that this is causing some unintended downstream consequences?  I'm using clerk with react, and every time my window gains focus, and clerk makes a request to `?touch`, it destabilizes the reference to the `user` state stored within clerk and all the parts of the app that depend on the `user` object from clerk are recomputed/triggered/rendered.

The18thWarrior:
> @dimkl @nikosdouvlis is it possible that this is causing some unintended downstream consequences? I'm using clerk with react, and every time my window gains focus, and clerk makes a request to `?touch`, it destabilizes the reference to the `user` state stored within clerk and all the parts of the app that depend on the `user` object from clerk are recomputed/triggered/rendered.

I also am seeing this behavior. @BRKalow @dimkl is there a know workaround for this?

The18thWarrior:
I was able to wrap the useUser in a hook to solve this issue. May be worth adding this to docs. 

`import { useState, useEffect, useMemo } from 'react';
import { useUser } from '@clerk/nextjs';

export function useClerkUser() {
  const {user: clerkUser} = useUser();
  const [user, setUser] = useState(clerkUser);
  const [balance, setBalance] = useState<number | null>(null);
  
  const syncUser = async () => {
    if (!clerkUser) return;
    if (user && user.id === clerkUser.id) return;
    setUser(clerkUser);
  };

  useEffect(() => {
    syncUser();
  }, [clerkUser]);

  const memoizedValue = useMemo(() => ({
    user,
    refresh: () => syncUser(),
  }), [balance]);

  return memoizedValue;
}
`

The18thWarrior:
Also the touch flag on ClerkProvider helps

dimkl:
Sorry for the late response but i am not a maintainer of this project any more. I don't know if there things has changed but one of the reason of triggering a `touch` request and updating the user object was to propagate user changes and also update the auth state of the user. For example if the user has performed signout from another tab, or has been updated we wanted to enforce that those changes are propagated as soon as possible and avoid showing restricted information or incorrect data to the end user.
cc: @imsherrill, @The18thWarrior

PR #3792, opened 2024-07-23
fix(astro): Middleware return type

wobsoriano:
## Description

The middleware contains logic to check for a passed handler. When no handler is provided, it automatically executes the [next function](https://github.com/clerk/javascript/blob/195fd46bd5cff3b54161ede02848d1c08e695116/packages/astro/src/server/clerk-middleware.ts#L108).  However, the return type doesn't handle this scenario and still expects us to call the `next` function. It should be optional.

Screenshot 2024-07-23 at 3 27 01 PM

This small PR addresses the issue.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/ninety-squids-appear.md
+++ b/.changeset/ninety-squids-appear.md
@@ -2,4 +2,4 @@
 "@clerk/astro": patch
 ---
 
-Fix middleware return type.
+Allow the handler of `clerkMiddleware` to return undefined. When undefined is returned, `clerkMiddleware` implicitly calls `await next()`.
--- a/.changeset/ninety-squids-appear.md
+++ b/.changeset/ninety-squids-appear.md
@@ -1,0 +1,5 @@
+---
+"@clerk/astro": patch
+---
+
+Fix middleware return type.
--- a/packages/astro/src/server/clerk-middleware.ts
+++ b/packages/astro/src/server/clerk-middleware.ts
@@ -40,7 +40,7 @@ type ClerkAstroMiddlewareHandler = (
   auth: () => ClerkMiddlewareAuthObject,
   context: AstroMiddlewareContextParam,
   next: AstroMiddlewareNextParam,
-) => AstroMiddlewareReturn;
+) => AstroMiddlewareReturn | undefined;
 
 type ClerkAstroMiddlewareOptions = AuthenticateRequestOptions;
 


anagstef (approved):

nikospapcom (approved):

.changeset/ninety-squids-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/astro": patch
+---
+
+Fix middleware return type.
```
panteliselef, line 5: Allow the handler of `clerkMiddleware` to return undefined. When undefined is returned `clerkMiddleware` implicitly calls `await next()`

panteliselef (changes requested):

.changeset/ninety-squids-appear.md
wobsoriano: thanks Pantelis

panteliselef (approved):

softbeehive:
Hi folks, 

I'm integrating clerk with Astro v4.12.2. May I ask **why do you use default export**? Astro documentation states that middleware is defined by exporting `onRequest`

I used this doc as a guide
https://clerk.com/docs/references/astro/clerk-middleware#protect-routes-based-on-user-authentication-status

I had an impression it was copied from next.js without much testing. Default export was throwing errors, because Astro expects a Response. I ended up doing the following:

```ts
import { defineMiddleware, sequence } from 'astro:middleware'
import { clerkMiddleware } from '@clerk/astro/server'

const authMiddleware = defineMiddleware(async (context, next) => {
  // Protect these pages
  const protectedPages = ['/plan']
  const { url } = context

  // Skip auth for public pages
  if (!protectedPages.some((pagePath) => url.pathname.startsWith(pagePath))) {
    return next()
  }

  const auth = await context.locals?.auth()
  if (auth.userId == null) {
    return auth.redirectToSignIn()
  }

  return next()
})

export const onRequest = sequence(clerkMiddleware(), authMiddleware)

```

I also dug deeper to discover that Astro intergration telemetry collection is not configurable. I recognize this integration is fresh. Consider making it respect CLERK_TELEMETRY_DEBUG. I learned new things but I'd appreciate the seamless quickstart experience, hopefully other folks can benefit from my struggle.

wobsoriano:
Thank you for reporting the incorrect exports @softbeehive. I've opened a pull request to address this docs issue [here](https://github.com/clerk/clerk-docs/pull/1325).

The correct way to export and use the middleware is as follows:

```ts
import { clerkMiddleware, createRouteMatcher } from '@clerk/astro/server';

const isProtectedRoute = createRouteMatcher(['/plan(.*)']);

export const onRequest = clerkMiddleware((auth, context) => {
  const { redirectToSignIn, userId } = auth();

  if (!userId && isProtectedRoute(context.request)) {
    return redirectToSignIn();
  }
});
```

A [PR](https://github.com/clerk/javascript/pull/3808) is also open for opting-out of telemetry.

softbeehive:
Cheers, +1 ⭐

PR #3793, opened 2024-07-24
chore(dev-cli): Adjust package.json

LekoArts:
## Description

The `package.json` was missing some keys that we typically have in our other packages.

### DIFF
--- a/.changeset/tiny-glasses-teach.md
+++ b/.changeset/tiny-glasses-teach.md
@@ -1,0 +1,5 @@
+---
+"@clerk/dev-cli": patch
+---
+
+Adjust package.json to align with the other published packages inside clerk/javascript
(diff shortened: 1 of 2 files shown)


.changeset/tiny-glasses-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/dev-cli": patch
+---
+
+Adjust package.json to align with the other published packages inside clerk/javascript
```
anagstef, line 5:
```suggestion
---
---
```
I think we don't need to release a version for these changes

LekoArts: But otherwise they wouldn't get published (which is the intention) 😅

nikosdouvlis (approved):

.changeset/tiny-glasses-teach.md
anagstef: Ok!

panteliselef (approved):

PR #3801, opened 2024-07-24
feat(ui): Sign out of other sessions checkbox

joe-bell:
## Description

Went back-and-forth on how to implement this – it opens up another future discussion around component composition within Elements – but ultimately decided to go down the route of using `accent-color` for basic styling at the moment

https://github.com/user-attachments/assets/86dd337e-1098-4463-85fd-497974e95ad6

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/primitives/field.tsx
```diff
@@ -80,6 +85,21 @@ export const Hint = React.forwardRef(function Hint(
...
+export const Checkbox = React.forwardRef(function FieldCheckbox(
+  props: React.InputHTMLAttributes<HTMLInputElement>,
+  forwardedRef: React.ForwardedRef<HTMLInputElement>,
+) {
+  return (
+    <input
+      ref={forwardedRef}
+      type='checkbox'
+      data-field-checkbox
+      className={cx('accent-accent-9 mt-[0.1875em] size-3 cursor-pointer')}
```
alexcarpenter, line 97: whats the reasoning to have margin top on the component here?

alexcarpenter (approved):
Nice. One small non-blocking question around mt usage on the checkbox component.

packages/ui/src/primitives/field.tsx
joe-bell: This is so that when the `Label` goes beyond one line, it'll align nicely to the top line, rather than awkwardly in the middle

PR #3804, opened 2024-07-24
fix(dev-cli): Call mkdir before attempting to write the config file

BRKalow:
## Description

Ensure the directory where we write the config file exists.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/dev-cli/src/commands/init.js
+++ b/packages/dev-cli/src/commands/init.js
@@ -1,6 +1,6 @@
 import { spawn } from 'node:child_process';
-import { writeFile } from 'node:fs/promises';
-import { join } from 'node:path';
+import { mkdir, writeFile } from 'node:fs/promises';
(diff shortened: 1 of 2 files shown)


packages/dev-cli/src/commands/init.js
```diff
@@ -26,6 +26,8 @@ export async function init() {
     },
   };
 
+  await mkdir(dirname(CONFIG_FILE), { recursive: true });
```
BRKalow, line 29: This should resolve successfully if it already exists for some reason:
```
Calling fsPromises.mkdir() when path is a directory that exists results in a rejection only when recursive is false.
```

dstaley (approved):

thiskevinwang (approved):
LGTM

PR #3812, opened 2024-07-24
feat(elements,ui): Add support for password error localization

alexcarpenter:
## Description

This PR adds a new `codes` argument to the render prop for `<Common.FieldState>` containing error codes corresponding to failed validations. This allows consumers to generate appropriate error messages based on the provided codes.

It also updates our UI package to demonstrate usage of the `codes` prop to generate localized error messages.

## Screenshots

Screen Shot 2024-07-25 at 12 06 21 PM
Screen Shot 2024-07-25 at 12 05 59 PM
Screen Shot 2024-07-25 at 12 05 41 PM

## Checklist

- [x] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated (https://github.com/clerk/clerk-docs/pull/1330)

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/utils/makeLocalizable.ts
```diff
@@ -220,3 +220,95 @@ export const makeLocalizeable = (resource: LocalizationResource) => {
...
+};
+
+const errorMessages: Record<keyof Omit<ComplexityErrors, 'allowed_special_characters'>, [string, string] | string> = {
+  max_length: ['unstable__errors.passwordComplexity.maximumLength', 'length'],
+  min_length: ['unstable__errors.passwordComplexity.minimumLength', 'length'],
+  require_numbers: 'unstable__errors.passwordComplexity.requireNumbers',
+  require_lowercase: 'unstable__errors.passwordComplexity.requireLowercase',
+  require_uppercase: 'unstable__errors.passwordComplexity.requireUppercase',
+  require_special_char: 'unstable__errors.passwordComplexity.requireSpecialCharacter',
+};
```
panteliselef, line 271: I know this is a draft, so feel free to ignore.
Is is possible to moved this under @clerk/shared ?

dstaley: absolutely! where else would we want to use it? I can easily see the case for the first three functions, but I'm curious where else we'd want to use the `errorMessages` object.

panteliselef: I believe this code originated from clerk-js, so the goal would be to eliminate duplication across packages

packages/elements/src/react/common/form/index.tsx
```diff
@@ -554,8 +565,9 @@ function FieldState({ children }: FieldStateRenderFn) {
 
   const message = feedback?.message instanceof ClerkElementsFieldError ? feedback.message.message : feedback?.message;
   const codes = feedback?.codes;
+  const config = feedback && 'config' in feedback ? feedback?.config : undefined;
```
panteliselef, line 568: Maybe I'm missing something, but what does this config achieve ? Should this be fetched from the environment

dstaley: Alex and I thought the same thing, but the config is currently at `Clerk.__unstable__environment.userSettings.passwordSettings`, so we felt it was better to provide it as a prop rather than have users explicitly access an unstable prop in their code.

dstaley: The config contains important values such as minimum and maximum length values, which we need access to in order to build error messages such as `Your password needs to be at least <n> characters long.`.

packages/ui/src/utils/makeLocalizable.ts
dstaley: Ah gotcha! Given that the UI package will eventually replace the logic within `clerk-js` I think it's probably fine to leave this duplicated for the moment. That being said we should definitely take a look at what we've pulled in and see if there's a case for it existing in a shared package after the merger.

tmilewski: Can we at least add a comment that this was pulled from that code? Just so that it's clear that this was duplicated and, eventually, the other work should be replaced. 🫶

packages/elements/src/internals/machines/form/form.context.ts
```diff
@@ -29,8 +29,15 @@ export const fieldValueSelector = (name: string | undefined) => (state: Snapshot
...
   Boolean(fieldValueSelector(name)(state));
 
+type MapValue<A> = A extends Map<any, infer V> ? V : never;
+
 /**
  * Selects field-specific feedback, if they exist
  */
-export const fieldFeedbackSelector = (name: string | undefined) => (state: SnapshotState) =>
-  name ? state.context.fields.get(name)?.feedback : undefined;
+// We declare an explicit return type here because TypeScript's inference results in the subtype reduction of the union
```
tmilewski, line 37: Let's play more into JSDoc here.

```js
/**
 * Selects field-specific feedback, if they exist.
 * 
 * @note We declare an explicit return type here because TypeScript's inference 
 *       results in the subtype reduction of the union used for 
 *       feedback. Explicitly declaring the return type allows for 
 *       all members of the union to be included in the return type.
 */
```

packages/ui/src/utils/makeLocalizable.ts
dstaley: Aye aye! Added a note in [ed1f767](https://github.com/clerk/javascript/pull/3812/commits/ed1f767ec7380adb9368f7393f5ff3e7a8e4a352)

packages/elements/src/internals/machines/form/form.context.ts
dstaley: Added in [48572c4](https://github.com/clerk/javascript/pull/3812/commits/48572c46cefa35730182bb60849108413a9f5bae)!

tmilewski (approved):

packages/elements/src/react/common/form/index.tsx
panteliselef: After this change would a developer have to pass the config prop for this to work as expected ? I don't advocate for devs to have to pass `Clerk.__unstable__environment.userSettings.passwordSettings` but it seems that this can be abstracted away as element already have access to environment internally. 

Saying this because i saw a PR for docs, and i'm not sure why we are thinking this should be public API

alexcarpenter: `config` was already being destructured in the `usePassword` hooks for use in the `generatePasswordErrorText` usage within elements. This PR is just exposing the usage of `config` within fieldstate for users to be able to use and translate the errors in usage of elements.

We are not requiring devs to access `Clerk.__unstable__environment.userSettings.passwordSettings` if that is what you are asking.

alexcarpenter: If there is a concern about exposing the config here, we can remove and access the config within UI package via `useClerk` and circle back and figure out the best solution.

panteliselef: Maybe I've not understood the usage of config here. I don't want to block this. Is this config for providing the text that will be used for localization or is this config used for (i.e.) setting the minimum password length ?

alexcarpenter: @panteliselef no worries, appreciate the review. You can see the config usage here in our UI usage of elements https://github.com/clerk/javascript/pull/3812/files#diff-44ae29a5976ef7ae53f3bda6592ff474783e38e79d777b25066f0b664be58cf4R307

tmilewski (approved):

PR #3820, opened 2024-07-25
refactor(ui): Use current accent color, switch to system font

joe-bell:
## Description

Caught up with Alvish today and he made a few recommendations that were relatively quick to fix:

1. Use our current (purple) accent color to reduce confusion between our accent and grayscale palette
2. Default to system font – current behaviour – rather than inherit

Setting the `--cl-font-family` has no effect currently as we haven't yet decided how to approach "global" styles within components, so for now I'm going to simply remove Inter (we can reassess later if needed)

**Before**

CleanShot 2024-07-25 at 16 23 22@2x

**After**

CleanShot 2024-07-25 at 16 23 01@2x

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/theme-builder/app/theme-builder.tsx
```diff
@@ -8,11 +8,11 @@ import { generateColors, getPreviewStyles } from './generate-colors';
 import { ThemeDialog } from './theme-dialog';
 import { ToggleGroup } from './toggle-group';
 
-const lightAccentDefault = '#2F3037';
+const lightAccentDefault = '#6C47FF';
```
alexcarpenter, line 11: I would assume we'd not want our default accent color to be the clerk purple and default to the current gray.

alexcarpenter (approved):
Sorry I missed the review here on friday.

I would not assume we'd update the primary accent color to be our purple but something we can circle back on and not a blocker.

packages/ui/theme-builder/app/theme-builder.tsx
joe-bell: No worries on the delay!

I went with Alvish's recommendation for now to help distinguish between the gray/accent (which I agree has proven tricky to differentiate at times)

PR #3823, opened 2024-07-25
feat(backend): Add param in User.update()

NicolasLopes7:
## Description
This PR adds a new field for the `user.update()` called `createOrganizationsLimit` which can be either number or null. This field is used to control how many orgs an user can create.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/api/endpoints/UserApi.ts
+++ b/packages/backend/src/api/endpoints/UserApi.ts
@@ -95,6 +95,7 @@ type UpdateUserParams = {
   externalId?: string;
   createdAt?: Date;
   createOrganizationEnabled?: boolean;
+  createOrganizationsLimit?: number;
 } & UserMetadataParams &
   (UserPasswordHashingParams | object);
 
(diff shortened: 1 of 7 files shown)


packages/backend/src/api/endpoints/UserApi.ts
dimkl, line 1: 🔧 It seems that this param is also supported in the `createUser` based on our backend-api docs: https://clerk.com/docs/reference/backend-api/tag/Users#operation/CreateUser!c=200&path=create_organizations_limit&t=response
Let's add it there too.

dimkl (approved):

packages/backend/src/api/endpoints/UserApi.ts
NicolasLopes7: Hey, I've just took a look at the BAPI implementation and it seems that is not taking the fields and assigning them, I'll fix that and get back here adding the params in a new PR 👍 I've created this [ticket](https://linear.app/clerk/issue/ORGS-112/fix-bapi-post-users-does-not-accept-user-actions-even-docs-says-it) to track it.

PR #3825, opened 2024-07-25
feat(clerk-js): Add ability to set active organization by slug

wobsoriano:
## Description

This PR introduces ability to set an active organization by slug and updates existing org tests to use a proper Org ID format (`org_*`).

```ts
setActive({ organization: 'some-cool-slug' })
```

Fixes ORGS-17

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -712,7 +712,7 @@ export class Clerk implements ClerkInterface {
     if (newSession && shouldSwitchOrganization) {
       const organizationIdOrSlug = typeof organization === 'string' ? organization : organization?.id;
 
-      if (isOrganizationId(organizationIdOrSlug!)) {
+      if (isOrganizationId(organizationIdOrSlug)) {
         newSession.lastActiveOrganizationId = organizationIdOrSlug || null;
       } else {
         const matchingOrganization = newSession.user.organizationMemberships.find(
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -75,6 +75,7 @@ import {
   inBrowser,
   isDevAccountPortalOrigin,
   isError,
+  isOrganizationId,
   isRedirectForFAPIInitiatedFlow,
   noOrganizationExists,
   noUserExists,
@@ -707,9 +708,18 @@ export class Clerk implements ClerkInterface {
     // However, if the `organization` parameter is not given (i.e. `undefined`), we want
     // to keep the organization id that the session had.
     const shouldSwitchOrganization = organization !== undefined;
+
     if (newSession && shouldSwitchOrganization) {
-      const organizationId = typeof organization === 'string' ? organization : organization?.id;
-      newSession.lastActiveOrganizationId = organizationId || null;
+      const organizationIdOrSlug = typeof organization === 'string' ? organization : organization?.id;
+
+      if (isOrganizationId(organizationIdOrSlug!)) {
+        newSession.lastActiveOrganizationId = organizationIdOrSlug || null;
+      } else {
+        const matchingOrganization = newSession.user.organizationMemberships.find(
+          mem => mem.organization.slug === organizationIdOrSlug,
+        );
+        newSession.lastActiveOrganizationId = matchingOrganization?.organization.id || null;
+      }
     }
 
     // If this.session exists, then signOut was triggered by the current tab
--- a/packages/clerk-js/src/utils/organization.ts
+++ b/packages/clerk-js/src/utils/organization.ts
@@ -1,8 +1,8 @@
 /**
  * Checks and assumes a string is an organization ID if it starts with 'org_', specifically for
- disambiguating with slugs. `_` is a disallowed character in slug names, so slugs cannot 
+ disambiguating with slugs. `_` is a disallowed character in slug names, so slugs cannot
  start with `org_`.
  */
-export function isOrganizationId(id: string) {
-  return id.startsWith('org_');
+export function isOrganizationId(id: string | null | undefined): boolean {
+  return typeof id === 'string' && id.startsWith('org_');
 }
--- a/packages/clerk-js/src/utils/organization.ts
+++ b/packages/clerk-js/src/utils/organization.ts
@@ -1,7 +1,7 @@
 /**
  * Checks and assumes a string is an organization ID if it starts with 'org_', specifically for
- disambiguating with slugs. `_` is a disallowed character in slug names, so slugs cannot
- start with `org_`.
+ * disambiguating with slugs. `_` is a disallowed character in slug names, so slugs cannot
+ * start with `org_`.
  */
 export function isOrganizationId(id: string | null | undefined): boolean {
   return typeof id === 'string' && id.startsWith('org_');
--- a/packages/clerk-js/src/utils/organization.ts
+++ b/packages/clerk-js/src/utils/organization.ts
@@ -1,0 +1,6 @@
+/**
+ * Checks and assumes a string is an organization ID if it starts with 'org_'.
+ */
+export function isOrganizationId(id: string) {
+  return id.startsWith('org_');
+}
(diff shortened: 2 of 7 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -708,8 +708,16 @@ export class Clerk implements ClerkInterface {
...
-      newSession.lastActiveOrganizationId = organizationId || null;
+      const organizationIdOrSlug = typeof organization === 'string' ? organization : organization?.id;
+
+      if (organizationIdOrSlug?.startsWith('org_')) {
+        newSession.lastActiveOrganizationId = organizationIdOrSlug || null;
+      } else {
+        const matchingOrganization = session?.user?.organizationMemberships.find(
+          mem => mem.organization.slug === organizationIdOrSlug,
+        );
+        newSession.lastActiveOrganizationId = matchingOrganization?.id || null;
```
wobsoriano, line 719: Get the `organization_id` by checking the memberships array using the string value passed which is likely a `slug`

packages/clerk-js/src/core/clerk.ts
```diff
@@ -708,8 +708,16 @@ export class Clerk implements ClerkInterface {
     // to keep the organization id that the session had.
     const shouldSwitchOrganization = organization !== undefined;
     if (newSession && shouldSwitchOrganization) {
-      const organizationId = typeof organization === 'string' ? organization : organization?.id;
-      newSession.lastActiveOrganizationId = organizationId || null;
+      const organizationIdOrSlug = typeof organization === 'string' ? organization : organization?.id;
+
+      if (organizationIdOrSlug?.startsWith('org_')) {
```
wobsoriano, line 713: If the string value starts with `org_`, then it's automatically an `organization_id` and we don't have to do extra checks.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -708,8 +708,16 @@ export class Clerk implements ClerkInterface {
...
+        newSession.lastActiveOrganizationId = organizationIdOrSlug || null;
```
wobsoriano, line 714: If the string value starts with `org_`, then it's automatically an `organization_id` and we don't have to do extra checks.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -708,8 +708,16 @@ export class Clerk implements ClerkInterface {
...
+      } else {
+        const matchingOrganization = session?.user?.organizationMemberships.find(
```
izaaklauer, line 716: Worth noting that this won't work if the slug is for an organization that they're not a member of, but the backend would reject that case anyway!

On reflection, if we fail to find an organization here (maybe because the user isn't a member), we might want to throw a helpful error here.

izaaklauer: Also: what happens here if session.user is null, and under what circumstances would that happen?

packages/clerk-js/src/core/clerk.ts
izaaklauer: This is likely worth breaking out into it's own utility function. Something like `isOrganizationId()`.

izaaklauer (commented):
Looking good I think!

I wonder if there are any tests we could add for passing a slug into setActive.

packages/clerk-js/src/core/clerk.ts
wobsoriano: > Also: what happens here if session.user is null, and under what circumstances would that happen?

This is a good catch! We have a condition in the beginning of the `setActive` method that checks if there's an available session already, and we should be using the `session` property from the `Clerk` class instead of the local one which will always have a value when it reaches the organization ID check:

https://github.com/clerk/javascript/blob/bb2e8720e8f6bb955d4f100d4f998065fec52727/packages/clerk-js/src/core/clerk.ts#L681-L685

So instead of `session?.user` we should be doing `this.session?.user` instead. Thanks!

wobsoriano: Will add tests!

packages/clerk-js/src/core/clerk.ts
wobsoriano: Applied!

izaaklauer (commented):
Overall this looks good to me!

🔧  As a part of this PR, I think we should also update the [docs for SetActiveParams](https://github.com/clerk/javascript/blob/bb2e8720e8f6bb955d4f100d4f998065fec52727/packages/types/src/clerk.ts#L684) to clarify that it can take a slug.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -707,9 +708,18 @@ export class Clerk implements ClerkInterface {
...
+
     if (newSession && shouldSwitchOrganization) {
-      const organizationId = typeof organization === 'string' ? organization : organization?.id;
-      newSession.lastActiveOrganizationId = organizationId || null;
+      const organizationIdOrSlug = typeof organization === 'string' ? organization : organization?.id;
+
+      if (isOrganizationId(organizationIdOrSlug!)) {
+        newSession.lastActiveOrganizationId = organizationIdOrSlug || null;
+      } else {
+        const matchingOrganization = newSession.user.organizationMemberships.find(
```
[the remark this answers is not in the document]
izaaklauer: I'd be happy to pair again to trace through this with you @wobsoriano . And I think it's likely that someone with more experience may know the answer to these off-hand - they strike me as foundational-sdk-philosophy questions 😄

packages/clerk-js/src/utils/organization.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Checks and assumes a string is an organization ID if it starts with 'org_'.
+ */
+export function isOrganizationId(id: string) {
```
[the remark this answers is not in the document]
wobsoriano: We are not exporting any utility function in the `clerk-js` package, so it's definitely not part of the public API

packages/clerk-js/src/core/clerk.ts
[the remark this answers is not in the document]
izaaklauer: I happened to chat with @BRKalow , and I got my answers!

1: The magic here is that, when someone calls useClerk, they're getting IsomorphicClerk, who'se implementation of setActive is here: 
https://github.com/clerk/javascript/blob/427fcdeaaba4e77273be29b4d7cca43f9aa18693/packages/react/src/isomorphicClerk.ts#L602-L608

If clerk hasn't loaded the client by the time the user calls setActive, it'll return `return Promise.reject();`. And if clerk has loaded, we'll have that session data. So this is safe?

2: This isn't client piggybacking supplying this data. Client piggybacking is us returning the latest client data on api calls _other than_ the explicit call to load the client, but it's the explicit call that we're making when clerk loads.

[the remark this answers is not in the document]
wobsoriano: This is a good client piggybacking intro. Thanks izaak!

izaaklauer (approved):

packages/clerk-js/src/core/clerk.ts
```diff
@@ -707,9 +708,18 @@ export class Clerk implements ClerkInterface {
     // However, if the `organization` parameter is not given (i.e. `undefined`), we want
     // to keep the organization id that the session had.
     const shouldSwitchOrganization = organization !== undefined;
+
     if (newSession && shouldSwitchOrganization) {
-      const organizationId = typeof organization === 'string' ? organization : organization?.id;
-      newSession.lastActiveOrganizationId = organizationId || null;
+      const organizationIdOrSlug = typeof organization === 'string' ? organization : organization?.id;
+
+      if (isOrganizationId(organizationIdOrSlug!)) {
```
LekoArts, line 715: I'd probably opt for removing the `!` and add `falsy` handling to `isOrganizationId` + add tests for passing in `undefined`

wobsoriano: Updated!

LekoArts (approved):

PR #3833, opened 2024-07-29
Fix return types for backend resource deletions

kylekz:
## Description

When deleting resources, for example an Organization, the backend client is typed to return `Organization`, whereas the actual HTTP API returns a `DeletedObject`, resulting in a type mismatch. This PR:
- updates `DeletedObjectJSON` to accept a generic that extends `string`, with a default of `string`
  - i initially wanted this to extend `ObjectType` but then that would break `DeletedObject.fromJSON()`
- changes the generics passed into `this.request()` for the following resource delete methods:
  - `Allowlist`
  - `Domain`
  - `EmailAddress`
  - `Organization`
  - `PhoneNumber`
  - `RedirectUrl`
  - `SamlConnection`
    - for this one i'm just assuming `ObjectType.SamlAccount` is the correct type, as there's no `SamlConnection`
  - `User`

I'm not sure what the convention would be for handling this `DeletedObjectJSON` `type` property so this just seemed like the most straight forward way that wouldn't break anything 😅

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


dimkl (changes requested):
Hello @kylekz and sorry for the late response and thank you very much for the contribution but i am afraid that it will require some work in order to be merged.

Since you took the time to contribute i will try to provide some context for the rest of the required changes.
The API methods of our backend SDK are expected to return a `*Resource` type. The `request<T>` method used is expected to be passed a `*Resource` type which will be the returned type of the API method.
Based on the above instead of using the `this.request<DeletedObjectJSON<typeof ObjectType.*>>` we should be using the `this.request<DeletedObject<typeof ObjectType.SamlAccount>>`

Finally the `saml_account` and `saml_connection` are different resources. As you have noticed the `ObjectType.SamlConnection` is missing from our codebase so we need to add it 😅

kylekz:
@dimkl no problem, thanks for the response! I've updated to add `ObjectType.SamlConnection`, added a generic to `DeletedObject` and made sure to use that instead of `DeletedObjectJSON`

I don't want to be wasting dev time with reviews etc so I'm off the mark here then I'm cool with closing and just opening an issue :)

clerk-cookie:
Hello 👋

PR #3844, opened 2024-07-30
chore(backend): Expose `AuthObject` type

kduprey:
## Description

This PR moves the AuthObjectDebugData, AuthObjectDebug, SignedInAuthObject, SignedOutAuthObject, and AuthObject type definitions in favour of exporting from @clerk/types. This is to enable clients to easily access this object type across our JavaScript SDKs.

https://linear.app/clerk/issue/SDK-1894/export-authobject-as-requested-by-client-in-clerktypes-package

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/internal.ts
+++ b/packages/backend/src/internal.ts
@@ -9,12 +9,7 @@ export { debugRequestState } from './tokens/request';
 
 export type { AuthenticateRequestOptions } from './tokens/types';
 
-export type {
-  SignedInAuthObjectOptions,
-  SignedInAuthObject,
-  SignedOutAuthObject,
-  AuthObject,
-} from './tokens/authObjects';
+export type { SignedInAuthObjectOptions } from './tokens/authObjects';
 export { makeAuthObjectSerializable, signedOutAuthObject, signedInAuthObject } from './tokens/authObjects';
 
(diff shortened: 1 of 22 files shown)


LekoArts (approved):
image

nikosdouvlis (approved):

dimkl (approved):
☁️ I think that this kind of change should be a minor since we are adding a new export.

packages/backend/src/internal.ts
```diff
@@ -9,12 +9,7 @@ export { debugRequestState } from './tokens/request';
 
 export type { AuthenticateRequestOptions } from './tokens/types';
 
-export type {
-  SignedInAuthObjectOptions,
-  SignedInAuthObject,
-  SignedOutAuthObject,
-  AuthObject,
-} from './tokens/authObjects';
+export type { SignedInAuthObjectOptions, SignedInAuthObject, SignedOutAuthObject } from './tokens/authObjects';
```
dimkl, line 12: 🔧 This change will break existing application importing from `/internal`. Even though they should not use it, should we keep it and add a deprecation warning to remove it later?
cc: @nikosdouvlis

PR #3846, opened 2024-07-30
fix(astro): Protect component prop types

wobsoriano:
## Description

This PR addresses an issue where the `<Protect />` component was accepting any prop without TypeScript raising errors. All the Astro components does not have a build step, and directly copied to `components/*`, so relative paths will not be resolvable.

Screenshot 2024-07-30 at 3 29 41 PM

Screenshot 2024-07-30 at 4 34 14 PM

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/astro/src/react/controlComponents.tsx
+++ b/packages/astro/src/react/controlComponents.tsx
@@ -1,10 +1,14 @@
-import type { CheckAuthorizationWithCustomPermissions, HandleOAuthCallbackParams } from '@clerk/types';
+import type {
+  CheckAuthorizationWithCustomPermissions,
+  HandleOAuthCallbackParams,
+  OrganizationCustomPermissionKey,
+  OrganizationCustomRoleKey,
+} from '@clerk/types';
 import { computed } from 'nanostores';
 import type { PropsWithChildren } from 'react';
 import React, { useEffect, useState } from 'react';
 
 import { $csrState } from '../stores/internal';
-import type { ProtectComponentDefaultProps } from '../types';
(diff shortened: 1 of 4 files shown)


packages/astro/src/react/controlComponents.tsx
```diff
@@ -70,7 +74,28 @@ export const ClerkLoading = ({ children }: React.PropsWithChildren): JSX.Element
...
+        condition: (has: CheckAuthorizationWithCustomPermissions) => boolean;
+        role?: never;
+        permission?: never;
+      }
+    | {
+        condition?: never;
+        role?: never;
+        permission?: never;
+      }
+  ) & {
```
wobsoriano, line 98: We need to separate the prop types of the Astro component and React component since the Astro one has no build step and imported types will not be resolvable.

dimkl (approved):

packages/astro/src/react/controlComponents.tsx
alexcarpenter: Would the astro prop types need `fallback: never;` added?

wobsoriano: If the purpose of adding `fallback: never` is to show an error whenever unexpected props are used in Astro `<Protect />`, we don't need it it since it should already show type errors for random props

Also just FYI, Astro does not support [typed slots](https://github.com/withastro/roadmap/discussions/414)

PR #3848, opened 2024-07-30
chore(repo): Retry handshake in case of handshake cookie collision

nikosdouvlis:
## Description
 We want the `__clerk_handshake` cookie to always be set on the ETLD+1 domain in order to support multiple apps running on the same domain, even if the apps are running on different same-level subdomains, for example, `sub1.a.example.com` and `sub2.a.example.com`

Of course, with multiple apps running, we risk collisions if more than one app needs to handshake at the same time (e.g., reopen a browser window with multiple tabs open). 

We can deal with collisions in 3 different ways: 
1. Set the handshake cookie on the domain the app is running on. This however, will break for the scenario where two apps using the same PK/SK run on same-level subdomains. Imagine two apps, a.example.com and b.example.com, with the latter being an 'alias' of the former. Visiting b.example.com with an expired session token will result in a handshake redirect to a.example.com, the handshake cookie will be set on a.example.com, and the handshake will never finish because b.example.com won't be able to read it. Also, if a.example.com tries to set a cookie on b.example.com, the browser will reject it as the two do not domain-match.
2. Suffix the cookies. The handshake cookie is rather big so we risk hitting header size limits with multiple suffixed handshake tokens on the etld+1
3. Let it collide, but handle the collisions gracefully

This PR implements solution 3. We let the handshake cookies collide. In the extremely rare case of a collision, the app that accidentally reads a handshake cookie of a different instance will simply ignore it (signature verification will fail) and retry the handshake again if needed. 

This PR also adds an infinite redirect loop protection mechanism to protect against the case where a customer accidentally added a PK/SK pair that does not match. 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/constants.ts
+++ b/packages/backend/src/constants.ts
@@ -19,6 +19,7 @@ const Cookies = {
   ClientUat: '__client_uat',
   Handshake: '__clerk_handshake',
   DevBrowser: '__clerk_db_jwt',
+  InfiniteRedirectionLoopCookie: '__clerk_redirection_loop',
 } as const;
 
 const QueryParameters = {
--- a/packages/backend/src/constants.ts
+++ b/packages/backend/src/constants.ts
@@ -19,7 +19,7 @@ const Cookies = {
   ClientUat: '__client_uat',
   Handshake: '__clerk_handshake',
   DevBrowser: '__clerk_db_jwt',
-  InfiniteRedirectionLoopCookie: '__clerk_redirection_loop',
+  RedirectCount: '__clerk_redirect_count',
 } as const;
 
 const QueryParameters = {
--- a/packages/backend/src/tokens/request.ts
+++ b/packages/backend/src/tokens/request.ts
@@ -179,6 +179,8 @@ ${error.getFullMessage()}`,
       // proceed with triggering handshake.
       const isRedirectLoop = setHandshakeInfiniteRedirectionLoopHeaders(handshakeHeaders);
       if (isRedirectLoop) {
+        const msg = `Clerk: Refreshing the session token resulted in an infinite redirect loop. This usually means that your Clerk instance keys do not match - make sure to copy the correct publishable and secret keys from the Clerk dashboard.`;
+        console.log(msg);
         return signedOut(authenticateContext, reason, message);
       }
       return handshake(authenticateContext, reason, message, handshakeHeaders);
@@ -210,7 +212,7 @@ ${error.getFullMessage()}`,
     }
 
     const newCounterValue = authenticateContext.handshakeRedirectLoopCounter + 1;
-    const cookieName = constants.Cookies.InfiniteRedirectionLoopCookie;
+    const cookieName = constants.Cookies.RedirectCount;
     headers.append('Set-Cookie', `${cookieName}=${newCounterValue}; SameSite=Lax; HttpOnly; Max-Age=3`);
     return false;
   }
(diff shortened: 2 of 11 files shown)


packages/backend/src/tokens/request.ts
```diff
@@ -193,6 +193,45 @@ ${error.getFullMessage()}`,
...
+      error.reason === TokenVerificationErrorReason.InvalidSecretKey ||
+      error.reason === TokenVerificationErrorReason.JWKKidMismatch ||
+      error.reason === TokenVerificationErrorReason.JWKFailedToResolve
+    ) {
+      // Let the request go through and eventually retry another handshake
+      // TODO: set a cookie so break the infinite loop
+      return;
+    }
+    // TODO: if N retries reached, return signedOut
+    const msg = `Clerk: Handshake token verification failed with "${error.reason}"`;
```
dimkl, line 231: 🔧 i think we should add use the `getFullMessage()` (as we did before this change) instead of error to include all the information causing the error.

packages/backend/src/tokens/request.ts
```diff
@@ -193,6 +200,57 @@ ${error.getFullMessage()}`,
...
+    if (authenticateContext.handshakeRedirectLoopCounter === 3) {
+      return false;
+    }
+
+    const newCounterValue = authenticateContext.handshakeRedirectLoopCounter + 1;
+    const cookieName = constants.Cookies.InfiniteRedirectionLoopCookie;
+    headers.append('Set-Cookie', `${cookieName}=${newCounterValue}; SameSite=Lax; HttpOnly; Max-Age=3`);
+
+    return true;
+  }
```
dimkl, line 216: 🙃 add an empty line

packages/backend/src/tokens/request.ts
```diff
@@ -173,9 +173,17 @@ ${error.getFullMessage()}`,
...
       // In the future if we want to decorate the handshake redirect with additional headers per call we need to tweak this logic.
-      return handshake(authenticateContext, reason, message, headers ?? buildRedirectToHandshake());
+      const handshakeHeaders = headers ?? buildRedirectToHandshake();
+      // Introduce the mechanism to protect for infinite handshake redirect loops
+      // using a cookie and returning true if it's infinite redirect loop or false if we can
+      // proceed with triggering handshake.
+      const isRedirectLoop = setHandshakeInfiniteRedirectionLoopHeaders(handshakeHeaders);
+      if (isRedirectLoop) {
+        return signedOut(authenticateContext, reason, message);
+      }
```
BRKalow, line 183: Would be a good idea to log something here, or adjust the provided message to explain what has happened. 👀

packages/backend/src/constants.ts
```diff
@@ -19,6 +19,7 @@ const Cookies = {
   ClientUat: '__client_uat',
   Handshake: '__clerk_handshake',
   DevBrowser: '__clerk_db_jwt',
+  InfiniteRedirectionLoopCookie: '__clerk_redirection_loop',
```
BRKalow, line 22: nit: maybe a better name

```suggestion
  RedirectCount: '__clerk_redirect_count',
```

nikosdouvlis: yup that looks better 👍🏻

packages/backend/src/tokens/request.ts
nikosdouvlis: ✔️ 
Added a simple console.log - this is only shown in production and we shouldn't show anything to the end user

nikosdouvlis:
!snapshot multiapp

PR #3858, opened 2024-07-31
feat(clerk-react): Add 'nonce' prop to ClerkProvider

jescalan:
## Description

When using a [strict-dynamic](https://content-security-policy.com/strict-dynamic/) content security policy, Clerk currently will not work, as we do not pass a nonce to the hotloaded clerk-js script. This PR adds a `nonce` prop to `ClerkProvider` that can be used to thread the nonce value through to the clerk-js script load. It also includes an extra feature for next.js where the nonce will be automatically pulled from the CSP header and threaded through without needing any props so long as the provider is server-rendered.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/src/app-router/server/utils.ts
+++ b/packages/nextjs/src/app-router/server/utils.ts
@@ -77,7 +77,7 @@ export function getScriptNonceFromHeader(cspHeaderValue: string): string | undef
   // extra layer.
   if (/[&><\u2028\u2029]/g.test(nonce)) {
     throw new Error(
-      'Nonce value from Content-Security-Policy contained HTML escape characters.\nLearn more: https://nextjs.org/docs/messages/nonce-contained-invalid-characters',
+      'Nonce value from Content-Security-Policy contained invalid HTML escape characters, which is disallowed for security reasons. Make sure that your nonce value does not contain the following characters: `<`, `>`, `&`',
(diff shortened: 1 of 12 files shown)


packages/nextjs/src/app-router/server/utils.ts
```diff
@@ -38,3 +38,48 @@ export const buildRequestLike = () => {
...
+  if (!nonce) {
+    return;
+  }
+
+  // Don't accept the nonce value if it contains HTML escape characters.
+  // Technically, the spec requires a base64'd value, but this is just an
+  // extra layer.
+  if (/[&><\u2028\u2029]/g.test(nonce)) {
+    throw new Error(
+      'Nonce value from Content-Security-Policy contained HTML escape characters.\nLearn more: https://nextjs.org/docs/messages/nonce-contained-invalid-characters',
```
BRKalow, line 80: Let's expand the error message and change the link here 👀

BRKalow (commented):
Needs a changeset!

nikosdouvlis:
Added a changeset to let the tests run :)

BRKalow (approved):
👏

PR #3878, opened 2024-08-01
refactor(ui): Break SignUp into smaller components based on step

alexcarpenter:
## Description

Begin reducing the fairly large signup component into smaller most digestible step components.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/components/sign-up/index.tsx
```diff
@@ -1 +1,15 @@
...
+
+export function SignUp() {
+  return (
+    <SignUpRoot>
+      <SignUpStart />
+      <SignUpVerifications />
+      <SignUpContinue />
+    </SignUpRoot>
+  );
+}
```
joe-bell, line 15: Nit: my preference would be to keep this `sign-up.tsx` and re-export from `index.ts`

joe-bell (approved):

packages/ui/src/components/sign-up/index.tsx
alexcarpenter: Works for me, I've been cautious to introduce any barrel files though.

PR #3887, opened 2024-08-02
refactor(ui): Break down SignIn into separate steps

joe-bell:
## Description

Follows upon #3878; breaking down `SignIn` into separate step components

See more detailed comments below 👇 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/components/sign-in/hooks/use-get-help.ts
joe-bell, line 1: Not sure how I feel about `components/<name>/hooks` but it felt better to separate the hook from the `SignInGetHelp` UI (avoiding any chance of circular dependencies)

packages/ui/src/components/sign-up/sign-up.tsx
joe-bell, line 1: @alexcarpenter just wanted to call this out rather than let it slide through your radar! My goal here was to match the behaviour of `SignIn` where we currently (and previously) define the component in `sign-in.tsx` – I wanted to match that choice here

packages/ui/src/components/sign-up/steps/start.tsx
```diff
@@ -16,7 +16,6 @@ import { useDisplayConfig } from '~/hooks/use-display-config';
 import { useEnabledConnections } from '~/hooks/use-enabled-connections';
 import { useEnvironment } from '~/hooks/use-environment';
 import { useLocalizations } from '~/hooks/use-localizations';
-import { Alert } from '~/primitives/alert';
```
joe-bell, line 19: This is unused; some hangover from my attempt to fix conflicts in #3878

packages/ui/src/components/sign-up/indentifiers.tsx
joe-bell, line 1: As this isn't being used anywhere outside of `verifications` – it felt better to define inline to reduce cognitive overhead and avoid "over componentizing" things

packages/ui/src/components/sign-in/first-factor-connections.tsx
joe-bell, line 1: Similarly to [above](https://github.com/clerk/javascript/pull/3887/files#r1701599502), we aren't using this anywhere outside of `choose-strategy` – let's define inline

packages/ui/src/components/sign-up/sign-up.tsx
alexcarpenter: works for me!

alexcarpenter (approved):
💯

PR #3907, opened 2024-08-06
refactor(ui): Use button for connections

joe-bell:
## Description

Removes our `Connection` primitive, and instead extends upon `Button` – with Alvish confirming secondary button icon colors should be the same darker tint

Instead, all rendering logic for `Connection` now lives in `Connections`

> [!NOTE]
> - I made the intentional decision to **not** respect the `socialButtonsVariant` for the "forgot password" `Step` because it felt very odd having a full-width button empty of space. This does contradict with current behaviour a little, but it feels like the right design choice to me. Happy to adjust this thinking though!
> - It's worth noting that the `gap` value matching our other buttons in the "forgot password" and "choose-strategy" steps is very much a happy accident. We might not need to ever worry about adding a `gap` prop, but wanted to flag it anyway  

### Before

CleanShot 2024-08-06 at 11 08 12@2x
CleanShot 2024-08-06 at 10 52 45@2x

### After

CleanShot 2024-08-06 at 11 12 12@2x
CleanShot 2024-08-06 at 10 42 51@2x

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


alexcarpenter (commented):
We seemed to have lost the loading "busy" state for the connection buttons?

packages/ui/src/primitives/button.tsx
```diff
@@ -81,13 +94,31 @@ export const Button = React.forwardRef(function Button(
       type={type}
       {...props}
     >
-      {busy && spinnerWhenBusy ? (
-        <Spinner className='shrink-0 text-[1.125rem]'>Loading…</Spinner>
+      {busy && intent === 'primary' ? (
```
alexcarpenter, line 97: Seeing a slight change in positioning of the text when we go into busy state. Would this be handled in icon adjustments ticket or apart of this work?


https://github.com/user-attachments/assets/440c8d6b-a2a5-46e3-bf2d-f84a08d8189e

joe-bell: Happy to fix as a part of this one! If you approve I'll get it fixed first thing tomorrow, but no pressure

alexcarpenter (approved):

PR #3920, opened 2024-08-08
fix(clerk-js): Passkeys should not be a candidate for the dynamic fie…

EmmanouelaPothitou:
…ld (v4)

## Description
Passkeys should not be a candidate for the dynamic field.
Allow users to display the email address field after selecting to input a phone number. Previously that was not possible when passkeys were enabled.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/heavy-wombats-sing.md
+++ b/.changeset/heavy-wombats-sing.md
@@ -1,0 +1,6 @@
+---
+"@clerk/clerk-js": patch
+"@clerk/types": patch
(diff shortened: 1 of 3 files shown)


panteliselef (approved):
Seems like we forgot to backport this one, nice work

PR #3934, opened 2024-08-12
fix(elements): Use window.location to navigate for absolute URLs

dstaley:
## Description

Adds support for navigating to absolute URLs via `context.router.push`. This allows `redirect_url` to be an absolute URL.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/elements/src/react/router/next.ts
+++ b/packages/elements/src/react/router/next.ts
@@ -1,6 +1,7 @@
 import { usePathname, useRouter, useSearchParams } from 'next/navigation';
 
 import { NEXT_WINDOW_HISTORY_SUPPORT_VERSION } from '~/internals/constants';
+import { isAbsoluteUrl } from '~/utils/is-absolute-url';
 
 import type { ClerkHostRouter } from './router';
 
@@ -21,7 +22,13 @@ export const useNextRouter = (): ClerkHostRouter => {
   return {
     mode: 'path',
     name: 'NextRouter',
-    push: (path: string) => router.push(path),
+    push: (path: string) => {
+      if (isAbsoluteUrl(path)) {
+        window.location.href = path;
+      } else {
+        router.push(path);
+      }
+    },
     replace: (path: string) =>
       canUseWindowHistoryAPIs ? window.history.replaceState(null, '', path) : router.replace(path),
     shallowPush(path: string) {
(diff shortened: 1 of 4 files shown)


packages/elements/src/react/router/next.ts
```diff
@@ -21,7 +22,13 @@ export const useNextRouter = (): ClerkHostRouter => {
...
     mode: 'path',
     name: 'NextRouter',
-    push: (path: string) => router.push(path),
+    push: (path: string) => {
+      if (isAbsoluteUrl(path)) {
+        window.location.href = path;
+      } else {
+        router.push(path);
+      }
+    },
```
panteliselef, line 31: Maybe i am missing some context here, but is this different from this [windowNavigate](https://github.com/clerk/javascript/blob/b6f0613dc9d8b0bab41cfabbaa8621b126e3bdf5/packages/clerk-js/src/utils/windowNavigate.ts#L11) within clerk-js

panteliselef: Also I believe the nextjs router's `push` will respect the absolute path and it will perform the navigation.

dstaley: I totally thought the same thing about `push`, but in my testing it did not (at least with cross-origin domains, not sure about same-origin). I know the `<Link>` component handles cross-origin absolute URLs though, so I think that's where I got confused.

We can definitely bring `windowNavigate` over into Elements for the additional functionality. Good call!

panteliselef: I tested with App router `router.push('https://google.com')` and seems to work. i've not tested with absolute url and same origin tho.

If you see something different, it may be an inconsistency across nextjs versions. so having the above logic will be required in that case.

dstaley: You're absolutely right. In Elements we have two routers, `ClerkRouter` and `FrameworkRouter`, which `ClerkRouter` wraps. I was calling `ClerkRouter.push(absoluteUrl)`, which doesn't work because it transforms the URL into `/?redirect_url=absoluteUrl`, and then calls `FrameworkRouter.push("/?redirect_url=absoluteUrl")`, which doesn't work. Calling `FrameworkRouter.push(absoluteUrl)` totally works as expected.

I'll close this PR in favor of a new one that reworks `ClerkRouter` to prevent the transformation of absolute URLs.

dstaley:
Closing in favor of https://github.com/clerk/javascript/pull/3947

PR #3974, opened 2024-08-16
fix(elements): Fix `SignIn.SafeIdentifier` potentially outputting an incorrect value

tmilewski:
## Description

Fix `SignIn.SafeIdentifier` potentially outputting an incorrect identifier when using similar multi-session sign in strategies.

Fixes: SDKI-609

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/elements/src/internals/machines/sign-in/router.selectors.ts
+++ b/packages/elements/src/internals/machines/sign-in/router.selectors.ts
@@ -8,18 +8,24 @@ export function SignInSafeIdentifierSelectorForStrategy(
 ): (s: SignInRouterSnapshot) => string {
   return (s: SignInRouterSnapshot) => {
     const signIn = s.context.clerk?.client.signIn;
-    const identifier = signIn.identifier || '';
 
     if (strategy) {
-      const matchingFactor = [...(signIn.supportedFirstFactors ?? []), ...(signIn.supportedSecondFactors ?? [])].find(
-        f => f.strategy === strategy,
-      );
+      const matchingFactors = [
+        ...(signIn.supportedFirstFactors ?? []),
+        ...(signIn.supportedSecondFactors ?? []),
+      ].filter(f => f.strategy === strategy);
+
+      const matchingFactor =
+        signIn.identifier && matchingFactors.length > 0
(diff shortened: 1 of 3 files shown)


packages/elements/src/internals/machines/sign-in/router.selectors.ts
```diff
@@ -8,18 +8,24 @@ export function SignInSafeIdentifierSelectorForStrategy(
...
-        f => f.strategy === strategy,
-      );
+      const matchingFactors = [
+        ...(signIn.supportedFirstFactors ?? []),
+        ...(signIn.supportedSecondFactors ?? []),
+      ].filter(f => f.strategy === strategy);
+
+      const matchingFactor =
+        signIn.identifier && matchingFactors.length > 0
+          ? matchingFactors.find(f => 'safeIdentifier' in f && f.safeIdentifier === signIn.identifier)
```
dstaley, line 20: Requiring the factor's safeIdentifier to match the user provided identifier results in the user provided identifier being returned for factors in which it's not valid, for example rendering a phone number for `email_code` and `email_link`. Should we maybe break this logic up into two pieces?

```ts
const matchingFactorForIdentifier =
  signIn.identifier && matchingFactors.length > 0
    ? matchingFactors.find(
        (f) => "safeIdentifier" in f && f.safeIdentifier === signIn.identifier
      )
    : null;

const matchingFactorForStrategy = matchingFactors[0];

if (
  matchingFactorForIdentifier &&
  "safeIdentifier" in matchingFactorForIdentifier
) {
  return matchingFactorForIdentifier.safeIdentifier;
} else if (
  matchingFactorForStrategy &&
  "safeIdentifier" in matchingFactorForStrategy
) {
  return matchingFactorForStrategy.safeIdentifier;
}
```

Screen Shot 2024-08-16 at 10 37 18 AM

alexcarpenter (approved):

dstaley (changes requested):

nme22 (approved):
Approving. I think @dstaley 's mention of breaking this logic up into two spots is a good call.

dstaley (approved):

PR #3977, opened 2024-08-16
feat(nextjs): Handle URL <> Session Org Mismatch in Middleware

izaaklauer:
## What problem is this solving?

Today, when developers use organization slugs in URLs, they need to go through contortions to be able to depend on the organization data in the clerk session, especially in server-rendered pages. In short, they need to detect the mismatch between the URL and the session on the server component, punt back to client-side javascript that can call setActive for the org as specified by the URL, and then re-render the server component. See some of that workflow here: https://clerk.com/docs/guides/force-organizations#set-an-active-organization-based-on-the-url

## What changed?

This PR introduces a new NextJS middleware option that allows developers to specify their URL structure to the middleware - specifically which url patterns indicate a desire to activate the personal workspace or an organization, and if so which one (by slug or ID). 

The middleware then checks the actual URL against that pattern, and if it detects a mismatch between the desired-active organization (from the url) and the actually-active organization (from the session), it performs a handshake with a new query param, which will activate the new organization.

## What does the new API look like? 

Imagine an application that supports organizations and the personal workspace, but includes both in their URLs. The might have URLs like:

| URL                      | Indicates                               |
| ------------------------ | --------------------------------------- |
| `/orgs/bcorp`              | org with slug "bcorp" should be active  |
| `/orgs/bcorp/settings`     | org with slug "bcorp" should be active  |
| `/personal-workspace/home` | the personal workspace should be active |

Currently, that application would require special handling in both server and client javascript to detect and resolve an organization mismatch. 

Now, they can add middleware config like this:

```ts
import { clerkMiddleware, createRouteMatcher } from "@clerk/nextjs/server";

const isProtectedRoute = createRouteMatcher(["(.*)"]);

export default clerkMiddleware((auth, req) => {
  if (isProtectedRoute(req)) auth().protect();
}, {
  organizationSyncOptions: {
    organizationPatterns: [
      "/orgs/:slug",
      "/orgs/:slug/(.*)",
    ],
    personalWorkspacePatterns: [
      "/personal-workspace",
      "/personal-workspace/(.*)"
    ],
  },
});
```

## What's next?
- A public-facing guide explaining the new guidance for managing an organization via the URL
- A public-facing sample repository demonstrating the new option

## Further Reading:
Clerk internal DX guide: https://www.notion.so/clerkdev/Sync-Org-from-URL-to-Session-via-Middleware-f41f00865390480ab2279391c5625b27?pvs=4


packages/backend/src/tokens/request.ts
```diff
@@ -381,10 +394,36 @@ ${error.getFullMessage()}`,
...
+          mustActivate = true;
+        }
+        if (mustActivate) {
+          return handleMaybeHandshakeStatus(authenticateContext, AuthErrorReason.ActiveOrganizationMismatch, '');
+        }
+      }
+
+    }
+  }
+  return requestState
```
izaaklauer, line 426: I'm not love with doing all this work after the `authenticateRequestWithTokenIn...` functions run. It also seems fine to do this right before we call `signedIn`, which would allow us to avoid handshaking in the case that we just resolved a handshake, which might be important (i'll think that through).

packages/backend/src/tokens/request.ts
```diff
@@ -96,6 +99,16 @@ export async function authenticateRequest(
...
 
+    if (options.organizationSync) {
+      const toActivate = getActivationEntity(requestURL, options.organizationSync);
+      if (toActivate) {
+        const params = getActivationParam(toActivate);
+        Object.entries(params).forEach(([key, value]) => {
+          url.searchParams.append(key, value);
+        });
+      }
+    }
```
izaaklauer, line 110: This allows us to resolve a potential mismatch during the original auth-renewing handshake.

integration/tests/handshake.test.ts
```diff
@@ -72,7 +73,7 @@ test.describe('Client handshake @generic', () => {
     await new Promise<void>(resolve => jwksServer.close(() => resolve()));
   });
 
-  test('Test standard signed-in - dev', async () => {
```
izaaklauer, line 75: My IDE automatically deleted all of these `Test` prefixes, citing this:

```
ESLint: should not have duplicate prefix(playwright/valid-title)
```

I can add them back if we like them though!

integration/tests/handshake.test.ts
```diff
@@ -885,3 +886,290 @@ test.describe('Client handshake @generic', () => {
...
+      },
+      then: {
+        expectStatus: 200,
+        fapiOrganizationIdParamValue: null,
+      },
+    },
+  ];
+
+  for (const testCase of cookieAuthCases) {
+    ['Cookie', 'Authorization'].forEach(authHeader => {
```
izaaklauer, line 1132: Note: This means we have ~20 test cases (10 for each authorization type), which takes about 1.5 mins to execute on my machine in parallel.

I don't feel like we _really_ need all of these test cases for both Cookie and Authorization, because in the actual implementation the logic is all shared. So we could get away with just doing one or two test cases for one of these auth types if we feel like we need to get the test runtime down.

packages/backend/src/tokens/request.ts
```diff
@@ -188,6 +203,53 @@ ${error.getFullMessage()}`,
     return signedOut(authenticateContext, reason, message);
   }
 
+  // NOTE(izaak): I don't love that null return implies signed-in...
```
izaaklauer, line 206: Another  option is I could take _and_ sometimes return a `SignedInState`, which also feels bad. I'll think about how to improve the structure, but ideas welcome.

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -187,7 +187,35 @@ const parseHandlerAndOptions = (args: unknown[]) => {
   ] as [ClerkMiddlewareHandler | undefined, ClerkMiddlewareOptions];
 };
 
-export const createAuthenticateRequestOptions = (clerkRequest: ClerkRequest, options: ClerkMiddlewareOptions) => {
+export const createAuthenticateRequestOptions: (clerkRequest: ClerkRequest, options: ClerkMiddlewareOptions) => {
+  /* TODO(izaak): I get this error unless I specify the type explicitly:
```
izaaklauer, line 191: Ideas welcome from the typescript wizards here

integration/templates/next-app-router/src/app/organizations-by-id/[id]/page.tsx
```diff
@@ -0,0 +1,18 @@
+import { auth } from '@clerk/nextjs/server';
+
+export default function Home({ params }): {} {
```
Comment (LauraBeatris) [inline integration/templates/next-app-router/src/app/organizations-by-id/[id]/page.tsx:3]: What's the goal for shipping those pages within the app-router template?

Comment (izaaklauer) [reply integration/templates/next-app-router/src/app/organizations-by-id/[id]/page.tsx:3]: Exactly! For the integration tests, I needed an app that had:

- an org ID as a path token
- an org slug as a path token
- an org prefix that indicated the personal workspace


And I needed sub-pages under each of those. So here we are!

packages/backend/src/tokens/request.ts
```diff
@@ -395,3 +485,93 @@ export const debugRequestState = (params: RequestState) => {
...
+    try {
+      orgResult = matcher(url.pathname);
+    } catch (e) {
+      // Intentionally not logging the path to avoid potentially leaking anything sensitive
+      console.error(`Failed to apply organization pattern "${options.organizationPatterns}" to a path`, e);
+      return null;
+    }
+
+    if (orgResult) {
+      const { slug = '', id = '' } = orgResult.params as { slug?: string; id?: string };
```
LauraBeatris, line 542: Instead of forcing a assertion on an unknown value, let's verify during runtime:

```ts
  if (orgResult && 'params' in orgResult) {
      const params = orgResult.params;

      if ('id' in params && typeof params.id === 'string') {
        return { type: 'organization', organizationId: params.id };
      }
```

packages/backend/src/tokens/request.ts
```diff
@@ -395,3 +485,93 @@ export const debugRequestState = (params: RequestState) => {
...
+      );
+      return null;
+    }
+
+    let personalResult: boolean | MatchResult<object>;
+    try {
+      personalResult = matcher(url.pathname);
+    } catch (e) {
+      // Intentionally not logging the path to avoid potentially leaking anything sensitive
+      console.error(`Failed to apply personal workspace pattern "${options.personalWorkspacePatterns}" to a path`, e);
```
LauraBeatris, line 511: I think we don't need to catch exceptions here, cause `match` doesn't thrown on failure, it returns `false`:

> A match is either `false` (no match) or a match result.

https://github.com/pillarjs/path-to-regexp/blob/df39d6c30b2a8bdc9b8bda186f658cc89d47e44c/src/index.ts#L437-L442

packages/backend/src/tokens/request.ts
```diff
@@ -395,3 +485,93 @@ export const debugRequestState = (params: RequestState) => {
...
+      console.warn(
+        'Detected an organization pattern match, but no organization ID or slug was found in the URL. Does the pattern include `:id` or `:slug`?',
+      );
+    }
+  }
+  return null;
+}
+
+// ActivatibleEntity is an entity that can be activated by the handshake API.
+type ActivatibleEntity = {
```
LauraBeatris, line 558: How about `OrganizationSyncTarget` or `OrganizationSyncEntity` - to be consistent with "organization sync" here.

Also, let's use `type` as a discriminator union here to make it really explicit that it's an impossible state to have `personalWorkspace` with `organizationId` or `organizationSlug`:

```ts
/** OrganizationSyncTarget is an entity that can be activated by the handshake API. **/
type OrganizationSyncTarget =
  | {
      type: 'personalWorkspace';
    }
  | {
      type: 'organization';
      organizationId: string;
    }
  | {
      type: 'organization';
      organizationSlug: string;
    };
```

It's also not possible to have both slug and ID at the same time, so let's keep those separate.

packages/backend/src/tokens/request.ts
```diff
@@ -395,3 +485,93 @@ export const debugRequestState = (params: RequestState) => {
...
+  type: 'personalWorkspace' | 'organization' | 'none';
+  organizationId?: string;
+  organizationSlug?: string;
+};
+
+// getHandshakeActivationParam takes an activatibatle entity (organization or personal workspace)
+// and generates the query parameters that FAPI expects on the handshake API to activate that entity.
+function getHandshakeActivationParam(toActivate: ActivatibleEntity): Map<string, string> {
```
LauraBeatris, line 566:
```suggestion
function `getOrganizationSyncQueryParams`
```

Let's be more explicit regarding query params here. 

Still on that function, I'd suggest to move to a switch/case to avoid a lot of nested `ifs`, becomes more idiomatic: 

```ts
  switch (entity.type) {
    case 'organization': {
      return {
        organization_id: 'organizationId' in entity ? entity.organizationId : entity.organizationSlug,
      };
    }
    default:
      return null;
  }

```

packages/backend/src/tokens/request.ts
```diff
@@ -188,6 +203,59 @@ ${error.getFullMessage()}`,
...
+    if (!mustActivate) {
+      return null;
+    }
+    if (authenticateContext.handshakeRedirectLoopCounter > 0) {
+      // We have an organization that needs to be activated, but this isn't our first time redirecting.
+      // This is because we attempted to activate the organization previously, but the organization
+      // must not have been valid (either not found, or not valid for this user), and gave us back
+      // a null organization. We won't re-try the handshake, and leave it to the server component to handle.
+      console.warn(
+        'Clerk: Organization activation handshake loop detected. This is likely due to an invalid organization ID or slug. Skipping organization activation.',
```
LauraBeatris, line 243: Let's move this to `errors.ts`.

packages/backend/src/tokens/request.ts
LauraBeatris: Let's leave as `null` but clarify on some JSDocs!

packages/backend/src/tokens/request.ts
```diff
@@ -197,7 +265,16 @@ ${error.getFullMessage()}`,
...
+      const signedInRequestState = await signedIn(authenticateContext, data, undefined, sessionTokenInHeader!);
+      // Org sync if necessary
+      const handshakeRequestState = handleMaybeOrganizationSyncHandshake(
+        authenticateContext,
+        signedInRequestState.toAuth(),
+      );
+      if (handshakeRequestState) {
+        return handshakeRequestState;
+      }
+      return signedInRequestState;
```
LauraBeatris, line 277:
```suggestion
      return handshakeRequestState || signedInRequestState;
```

packages/backend/src/tokens/request.ts
izaaklauer: We chatted about this, and decided to leave it. Rationalle:
- Other logs in this file (and in the backend package generally) aren't declaring error consts
- This error is very specific and unlikely to be reused

packages/backend/src/tokens/request.ts
```diff
@@ -96,6 +100,17 @@ export async function authenticateRequest(
       url.searchParams.append(constants.QueryParameters.DevBrowser, authenticateContext.devBrowserToken);
     }
 
+    if (options.organizationSync) {
```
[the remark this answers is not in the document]
izaaklauer: 👍  d0d545f8fd19f506a67c366c618ffe472064f3c1

packages/backend/src/tokens/request.ts
```diff
@@ -188,6 +203,56 @@ ${error.getFullMessage()}`,
     return signedOut(authenticateContext, reason, message);
   }
 
+  // handleMaybeOrganizationSyncHandshake determines if a handshake must occur to resolve a mismatch between
+  // the organization as specified by the URL (according to the options) and the actual active organization
+  // on the session.
+  // NOTE(izaak): I don't love that null return implies signed-in...
+  function handleMaybeOrganizationSyncHandshake(
```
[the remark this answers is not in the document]
izaaklauer: To me, the difference is that `is...` implies a boolean return. `isRequestEligibleForHandshake` returns a boolean.

I'm more modeling after `handleMaybeHandshakeStatus`, which "handles" the situation. by adding the appropriate handshake headers to the response.

packages/backend/src/tokens/__tests__/request.test.ts
```diff
@@ -165,6 +165,181 @@ export default (QUnit: QUnit) => {
...
+          type: 'organization',
+          organizationSlug: 'org_bar',
+        },
+      },
+    ];
+
+    testCases.forEach(testCase => {
+      test(testCase.name, assert => {
+        const path = new URL(`http://localhost:3000${testCase.whenAppRequestPath}`);
+        const toActivate = getOrganizationSyncTarget(path, testCase.whenOrgSyncOptions);
```
izaaklauer, line 337: To get this test to work, I had to export `getOrganizationSyncTarget` (see [here](https://github.com/clerk/javascript/pull/3977/files#diff-50844e9765ddaece7cd9885aa9a6b4556d93d6ccaf92ef2484399cec67a09405R642)), which I don't love. Any ideas on how to avoid that are welcome.

integration/tests/handshake.test.ts
```diff
@@ -885,3 +886,524 @@ test.describe('Client handshake @generic', () => {
...
+
+    // The middleware should redirect to fapi with this query param value:
+    fapiOrganizationIdParamValue: string | null;
+  };
+
+  const cookieAuthCases: testCase[] = [
+    // ---------------- Session active vs expired tests ----------------
+    // Note: it would be possible to run _every_ test with both active and expired initial states
+    //       and expect the same results, but we're avoiding that to save some test execution time.
+    {
```
izaaklauer, line 990: There are a lot of test cases here, but there could be more! The tests here include every (anticipated) permutation of every input parameter in isolation, but not every combination of permutations (to keep a reasonable execution time)

See this (clerk internal) [spreadsheet](https://docs.google.com/spreadsheets/d/15OVcrHz_tj1K9XSfgPDzEQbjQLqn7XotJaZdZbLrLI4/edit?usp=sharing) that I used to track which permutations are tested.

integration/tests/handshake.test.ts
```diff
@@ -885,3 +886,524 @@ test.describe('Client handshake @generic', () => {
     expect(res.status).toBe(200);
   });
 });
+
+test.describe('Client handshake with organization activation @nextjs', () => {
```
izaaklauer, line 890: Scrutiny here: These tests take almost exactly one minute to run on my machine, in parallel with seven worker (via `npm run test:integration:nextjs -- handshake.test.ts`). For comparison, the above generic tests take 32 seconds on my machine.

If we think that's unacceptable, I can do an optimization pass. The lowest-handing-fruit that I see is that each test is starting it's own application server to control it's middleware config. It's likely possible to make the tests serial, overwrite the middleware config per-test, and reload the application.

packages/backend/src/tokens/request.ts
izaaklauer: The internal fork can throw! https://github.com/clerk/javascript/blob/6275c242cd8bcb6f7766934059967e0fe775a0c1/packages/shared/src/pathToRegexp.ts#L31

I do see your point that `match` is documented to return false, and I suppose I could read through the `match` implementation (or thoroughly test) to ensure that it won't throw for any config. But it seems safer to me to just add the error handling!

packages/backend/src/tokens/request.ts
```diff
@@ -395,3 +485,93 @@ export const debugRequestState = (params: RequestState) => {
...
+
+// getActivationEntity determines if the given URL and settings indicate a desire to activate a specific organization or personal workspace.
+function getActivationEntity(url: URL, options: OrganizationSyncOptions): ActivatibleEntity | null {
+  // Check for personal workspace activation
+  if (options.personalWorkspacePatterns) {
+    let matcher: MatchFunction<object>;
+    try {
+      matcher = match(options.personalWorkspacePatterns);
+    } catch (e) {
+      console.error(
```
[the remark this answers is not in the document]
izaaklauer: > Let's replace match to use the vendor path-to-regexp introduced https://github.com/clerk/javascript/issues/4145. We'll have to re-export this match function within @clerk/shared.

Done!

> In this case makes more sense to re-thrown to catch invalid patterns during development, since developers might not notice the console.error on the terminal.

Done! Good idea.

c2028328caaf9cc198ea36a14aa42112ebf55df5

packages/backend/src/tokens/request.ts
izaaklauer: Good idea - Done!

6c8f99f02a9f6285e89e7f133bf39499441509ec

package-lock.json
```diff
@@ -14799,6 +14799,12 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
+    "node_modules/astro/node_modules/path-to-regexp": {
+      "version": "6.3.0",
+      "resolved": "https://registry.npmjs.org/path-to-regexp/-/path-to-regexp-6.3.0.tgz",
+      "integrity": "sha512-Yhpw4T9C6hPpgPeA28us07OJeqZ5EzQTkbfwuhsUg0c237RomFoETJgmp2sa3F/41gfLE6G5cqcYwznmeEeOlQ==",
+      "peer": true
+    },
```
izaaklauer, line 14807: I'm note sure why npm added this - I can delete it if we think that's prudent.

packages/backend/src/tokens/request.ts
```diff
@@ -531,3 +631,106 @@ export const debugRequestState = (params: RequestState) => {
...
+  }
+
+  // Check for personal workspace activation
+  if (options.personalWorkspacePatterns) {
+    let matcher: MatchFunction<Partial<Record<string, string | string[]>>>;
+    try {
+      matcher = match(options.personalWorkspacePatterns);
+    } catch (e) {
+      // Likely to be encountered during development, so throwing the error is more prudent than logging
+      throw new Error(`Invalid personal workspace pattern "${options.personalWorkspacePatterns}": "${e}"`);
```
izaaklauer, line 655: Note that in practice, this never seems to throw - any string appears to be accepted.

.changeset/olive-trainers-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/nextjs": patch
+---
+
+New nextjs middleware option `OrganizationSyncOptions`, which syncs an active organization or personal workspace from a URL to the clerk session.
```
LauraBeatris, line 5:
```suggestion
Introduces `organizationSyncOptions` option to `clerkMiddleware`, which syncs an active organization or personal workspace from a URL to the clerk session.
```

packages/backend/src/tokens/request.ts
```diff
@@ -299,8 +376,13 @@ ${error.getFullMessage()}`,
       if (errors) {
         throw errors[0];
       }
-      // use `await` to force this try/catch handle the signedIn invocation
-      return signedIn(authenticateContext, data, undefined, sessionTokenInHeader!);
+      const signedInRequestState = signedIn(authenticateContext, data, undefined, sessionTokenInHeader!);
+      // Org sync if necessary
+      const handshakeRequestState = handleMaybeOrganizationSyncHandshake(
```
colinclerk, line 381: AFAIK, we do not currently support handshake for token in header, so I think this should be removed.

Token in header implies that the request is a fetch() rather than a navigation (which would rely on cookies).

The issue is that a fetch request will not normally come up with `credentials: true`, and then redirecting to handshake fails because the __client cookie is not forwarded along to FAPI.

Rather than support this and try to educate on credentials: true, we believe the proper the solution is to include the appropriate token in the fetch(), and I think that logic holds for org sync.

izaaklauer: 👍  Understood. To paraphrase and check my own understanding:

Handshake isn't supported on fetch requests, because a non-browser fetch-er won't automatically include the `__client` cookie, so the handshake response wouldn't be useful. We are currently never handshaking when the auth is defined via headers, and it doesn't make sense org sync either.

I'll pull the handshake detection out of this pathway.

izaaklauer: One other interesting note: while pairing, @LauraBeatris , @wobsoriano  and I figured out that we will return a handshake response when the auth header is present IF the `sec-fetch-dest: document` header is applied. 

I think that's fine, because no fetch client would naturally include that, but it's an interesting hole in the "we don't redirect if there's header auth" assumption.

colinclerk: Can you clarify the hole? `sec-fetch-dest: document` implies it's a page navigation, and the `Authorization` header cannot be set on page navigation (with small exception of Basic auth, but that's not relevant to clerk's header auth that relies on Bearer keys).

izaaklauer: Just a hole in my own assumptions about when we would and wouldn't handshake. I was assuming we were using the auth method (header vs cookie) to determine if we handshake, but in fact it's the `sec-fetch-dest` and `accept` headers that make the ultimate decision: https://github.com/clerk/javascript/blob/19d3808d4672234944226d6709ec51214e8d6e1d/packages/backend/src/tokens/request.ts#L65-L71

But you're right - the auth header and sec-fetch-dest:document should be mutually exclusive, so the distinction doesn't matter in practice.

packages/backend/src/tokens/request.ts
```diff
@@ -106,9 +109,11 @@ export async function authenticateRequest(
   }
 
   function buildRedirectToHandshake({
+    requestURL,
     handshakeReason,
     refreshError,
   }: {
+    requestURL: URL;
```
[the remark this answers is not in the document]
izaaklauer: 👍  905c3a917cb0928cf56b9d67c15ac84f5f4d015e

packages/backend/src/tokens/request.ts
izaaklauer: In any case, fixed: 889097c2eba9af33778ab5b87cd0f3ea9e81b93d

packages/backend/src/tokens/request.ts
```diff
@@ -291,6 +308,63 @@ ${error.getFullMessage()}`,
...
 
+  /**
+   * Determines if a handshake must occur to resolve a mismatch between the organization as specified
+   * by the URL (according to the options) and the actual active organization on the session.
+   *
+   * @returns {HandshakeState | SignedOutState | null} - The function can return the following:
+   *   - {HandshakeState}: If a handshake is needed to resolve the mismatched organization.
+   *   - {SignedOutState}: If a handshake is required but cannot be performed.
+   *   - {null}:           If no action is required.
+   */
```
LauraBeatris, line 319:
```suggestion
   * @returns {HandshakeState | SignedOutState | null} - The function can return the following:
   *   - {HandshakeState}: If a handshake is needed to resolve the mismatched organization.
   *   - {SignedOutState}: If a handshake is required but cannot be performed.
   *   - {null}:           If no action is required.
   */
```
We can remove the `@returns` tag here since TypeScript already handles that for us.

packages/backend/src/tokens/request.ts
```diff
@@ -531,3 +621,106 @@ export const debugRequestState = (params: RequestState) => {
   const { isSignedIn, proxyUrl, reason, message, publishableKey, isSatellite, domain } = params;
   return { isSignedIn, proxyUrl, reason, message, publishableKey, isSatellite, domain };
 };
+
+/**
+ * Determines if the given URL and settings indicate a desire to activate a specific
+ * organization or personal workspace.
+ *
+ * @returns {OrganizationSyncTarget | null} - The function can return the following:
+ *  - {OrganizationSyncTarget}: If the URL and settings indicate a desire to activate an organization or personal workspace.
```
LauraBeatris, line 630:
```suggestion

```

packages/backend/src/tokens/request.ts
```diff
@@ -291,6 +308,63 @@ ${error.getFullMessage()}`,
...
+   * @returns {HandshakeState | SignedOutState | null} - The function can return the following:
+   *   - {HandshakeState}: If a handshake is needed to resolve the mismatched organization.
+   *   - {SignedOutState}: If a handshake is required but cannot be performed.
+   *   - {null}:           If no action is required.
+   */
+  function handleMaybeOrganizationSyncHandshake(
+    authenticateContext: AuthenticateContext,
+    auth: SignedInAuthObject,
+  ): HandshakeState | SignedOutState | null {
+    const toActivate = getOrganizationSyncTarget(authenticateContext.clerkUrl, options.organizationSyncOptions);
```
LauraBeatris, line 324:
```suggestion
    const organizationSyncTarget = getOrganizationSyncTarget(authenticateContext.clerkUrl, options.organizationSyncOptions);
```

@izaaklauer How about renaming this to organization target, it's confusing a bit from `toActivate` to `mustActivate` below those almost look the same thing.

LauraBeatris (approved):

packages/backend/src/tokens/request.ts
izaaklauer: Is there a better place for me to give details on what each of those return types mean?

packages/backend/src/tokens/request.ts
```diff
@@ -574,6 +662,106 @@ export const debugRequestState = (params: RequestState) => {
...
+  }
+  return null;
+}
+
+/**
+ * Represents an organization or a personal workspace - e.g. an
+ * entity that can be activated by the handshake API.
+ */
+export type OrganizationSyncTarget =
+  | { type: 'personalWorkspace' }
```
panteliselef, line 742: @izaaklauer I remember us labelling this as "Personal Account". Can you confirm with @colinclerk on this ?

packages/backend/src/tokens/types.ts
```diff
@@ -10,5 +10,60 @@ export type AuthenticateRequestOptions = {
...
+ * and handle mismatches appropriately (e.g. by returning a 404).
+ */
+export type OrganizationSyncOptions = {
+  /**
+   * URL patterns that are organization-specific and contain an organization ID or slug as a path token.
+   * If a request matches this path, the organization identifier will be extracted and activated before rendering.
+   *
+   * WARNING: If the organization cannot be activated either because it does not exist or the user lacks access,
+   * organization-related fields will be set to null. The server component must detect this and respond
+   * with an appropriate error (e.g., notFound()).
```
panteliselef, line 37: I think this is nice to have here, but it definitely needs to reach the customer facing documentation that we have as well.

packages/backend/src/tokens/types.ts
```diff
@@ -10,5 +10,60 @@ export type AuthenticateRequestOptions = {
...
+  /**
+   * URL patterns for resources in the context of a clerk personal workspace (user-specific, outside any organization).
+   * If the route also matches the organizationPattern, this takes precedence.
+   *
+   * Common examples:
+   * - ["/user", "/user/(.*)"]
+   * - ["/user/:any", "/user/:any/(.*)"]
+   */
+  personalWorkspacePatterns?: Array<Pattern>;
+};
```
panteliselef, line 60: Maybe there is a benefit of redefining those for the `nextjs` package and take advantage of our  [RouteMatcherWithNextTypedRoutes](https://github.com/clerk/javascript/blob/7f873916ed95ed852acfc82680fe71fbc79ec662/packages/nextjs/src/server/routeMatcher.ts#L9)

panteliselef: And just to double check here, we don't wanna use the `createRouteMatcher` pattern here, because we are not expecting for folks to use more than one path ? Although the name of the properties hints that you can. If not, I think supporting the route matcher here makes sense.

cc @nikosdouvlis

packages/backend/src/tokens/types.ts
```diff
@@ -10,5 +10,60 @@ export type AuthenticateRequestOptions = {
...
+   *
+   * Must have a path token named either ":id" (matches a clerk organization ID) or ":slug" (matches a clerk
+   * organization slug).
+   *
+   * Common examples:
+   * - ["/orgs/:slug", "/orgs/:slug/(.*)"]
+   * - ["/orgs/:id", "/orgs/:id/(.*)"]
+   * - ["/app/:any/orgs/:slug", "/app/:any/orgs/:slug/(.*)"]
+   */
+  organizationPatterns?: Array<Pattern>;
```
BRKalow, line 49: I _think_ this is generally the canonical syntax when defining an array type:

```suggestion
  organizationPatterns?: Pattern[];
```

There's no functional difference either way, this is just a shorthand for `Array<Pattern>`

packages/backend/src/tokens/request.ts
izaaklauer: 👍 Internal conversation around this here: https://clerkinc.slack.com/archives/C078JFZLLQ3/p1727818679671599

izaaklauer: tl;dr: Agreed that account is canonical, and i'll update the references accordingly. Thank you!

izaaklauer: Fixed in 5858a24cb1215314f3adf73bac1f8e397d1aeefc

integration/tests/handshake.test.ts
izaaklauer: Oh good idea! Since I wrote this initially, @LauraBeatris landed https://github.com/clerk/javascript/pull/4160, which I think I could use to pull middleware options out of headers and avoid starting a new app every time. I'll give that a shot.

izaaklauer: Done: 3551d2f331c1620a6c2994fc93b52753a2a10c28

It's odd to me though that this hasn't significantly reduced load or test time on my machine. It's still 56 seconds 🤔 .

packages/backend/src/tokens/types.ts
izaaklauer: Good idea! I didn't realize we already have [RouteMatcherParams](https://github.com/search?q=repo%3Aclerk%2Fjavascript+%22export+type+RouteMatcherParam%22&type=code), which serves the same purpose.

It looks like there's some duplication between the astro and nextjs RouteMatcherParams. My inclination here is to make the astro version (which doesn't depend on anything astro-specific) shared, and import that for use here. But let me know if you think another path would be better!

r/e createRouteMatcher - We could take a createRouteMatcher-returned function here too, but my inclination is to start with just accepting the pattern syntax for simplicity and expand the the more complex types if necessary in the future - I explored that a bit here (internal only): https://www.notion.so/Sync-Org-from-URL-to-Session-via-Middleware-f41f00865390480ab2279391c5625b27?d=f3738cb7f62c48bb90f8924c653a83a9&pvs=4#c0b57bf937694dd6bf510aeb4dde12af. Again, differing opinions welcome!

izaaklauer: Oh you know, digging in a bit further, I think the main difference here between this and `RouteMatcherWithNextTypedRoutes`/`RouteMatcherRoutes` is that they're geared at returning true/false, not pulling information out of the path. For the organization pattern, we need it to include a group (which I have here as `:slug` or `:id`). I don't see a clear way to extend the RouteMatcherParams to encompass that without making the type too complex for my taste. I could have the personal workspace pattern use RouteMatcherParams, but I think the mixed types would be worse DX. 

And to answer your question:
>  we are not expecting for folks to use more than one path ?

I am expecting more than one path! It's very tricky to get most practical examples down to just one path expression, so I think allowing multiples is the way to go.

packages/backend/src/tokens/request.ts
```diff
@@ -574,6 +662,106 @@ export const debugRequestState = (params: RequestState) => {
...
+    if (personalResult) {
+      return { type: 'personalWorkspace' };
+    }
+  }
+
+  // Check for organization activation
+  if (options.organizationPatterns) {
+    let matcher: MatchFunction<Partial<Record<string, string | string[]>>>;
+    try {
+      matcher = match(options.organizationPatterns);
```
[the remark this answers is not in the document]
izaaklauer: 👍  9f36a37c0fdfc1103a709e6fed88de29ec695cc4

(will close this and carry the convo on in https://github.com/clerk/javascript/pull/3977#discussion_r1781281505)

integration/tests/handshake.test.ts
```diff
@@ -885,3 +886,525 @@ test.describe('Client handshake @generic', () => {
...
+    secFetchDestHeader: string | null;
+  };
+
+  type then = {
+    // A handshake should (or should not) occur:
+    expectStatus: number;
+
+    // The middleware should redirect to fapi with this query param value:
+    fapiOrganizationIdParamValue: string | null;
+  };
```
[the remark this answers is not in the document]
izaaklauer: 👍  5780554d1fe66ee8c0d0238e0c7f6e8edd307b17

packages/backend/src/tokens/types.ts
izaaklauer: 👍  4076c031d66ac56955f56e8b44dff0db2d7af6b9

BRKalow (approved):
lgtm!

PR #3984, opened 2024-08-19
feat(ui): Add useCard hook

alexcarpenter:
## Description

Extract `useCard` hook for accessing `logoProps` and `footerProps` data.

Fixes SDKI-612

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/hooks/use-card.ts
```diff
@@ -0,0 +1,26 @@
+import { useAppearance } from '~/contexts';
```
dstaley, line 1: Can you add some comments to this file describing what the hook does, when you'd want to use it, etc? Also what do you think about calling it `useCardDecoration`? With good docs I don't really think the name matters, but it feels like these values aren't important for the actual usage of the card, so calling the hook `useCard` feels a bit bigger than what the use case actually is.

alexcarpenter: Let me know what you think here https://github.com/clerk/javascript/pull/3984/commits/5083de9c9869a5892421a8d21815c55a75ca9404

packages/ui/src/hooks/use-card.ts
```diff
@@ -0,0 +1,61 @@
...
+ *   return (
+ *     <Card>
+ *       <Card.Logo {...logoProps} />
+ *       <Card.Footer {...footerProps} />
+ *     </Card>
+ *   );
+ * }
+ * ```
+ *
+ * @returns {{
```
dstaley, line 25: I don't think we need the `@returns` annotation here since TypeScript is already inferring a return type, right?

alexcarpenter: 👍🏼 it is yeah

dstaley (approved):

PR #3995, opened 2024-08-21
chore(nextjs): Fix server actions detection

panteliselef:
## Description

`isServerActionRequest` was still depending on the `next-url` header which no longer is available (removed at `14.0.4` for server actions and in `14.2.3` when requesting RSCs ([Related PR](https://github.com/clerk/javascript/pull/3529))

This PR fixes the issue and prepares the ground for future work that would require these utilities to work properly.

`pagePath` from __nextGetStaticStore is available since `next@13.5.4` until `next@14.2.5` which is the latest stable release

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/src/server/protect.ts
+++ b/packages/nextjs/src/server/protect.ts
@@ -112,10 +112,8 @@ export const createProtect = (opts: {
 
 const isServerActionRequest = (req: Request) => {
   return (
-    !!req.headers.get(nextConstants.Headers.NextUrl) &&
-    (req.headers.get(constants.Headers.Accept)?.includes('text/x-component') ||
-      req.headers.get(constants.Headers.ContentType)?.includes('multipart/form-data') ||
-      !!req.headers.get(nextConstants.Headers.NextAction))
+    req.headers.get(constants.Headers.Accept)?.includes('text/x-component') &&
+    !!req.headers.get(nextConstants.Headers.NextAction)
   );
 };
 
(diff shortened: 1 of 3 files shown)


packages/nextjs/src/server/protect.ts
```diff
@@ -112,10 +112,8 @@ export const createProtect = (opts: {
 
 const isServerActionRequest = (req: Request) => {
   return (
-    !!req.headers.get(nextConstants.Headers.NextUrl) &&
-    (req.headers.get(constants.Headers.Accept)?.includes('text/x-component') ||
-      req.headers.get(constants.Headers.ContentType)?.includes('multipart/form-data') ||
```
panteliselef, line 117: Checking against multipart is not realistic. The following example will fire a request with Content-type `text/plain;charset=UTF-8`

```tsx
'use client';

import { addItem2 } from '../actions';

export default function AddToCart() {
  return (
    <button
      type='button'
      onClick={() => {
        addItem2({ wowowwo: 'dwa' });
      }}
    >
      Somethj
    </button>
  );
}

```

packages/nextjs/src/server/protect.ts
```diff
@@ -112,10 +112,8 @@ export const createProtect = (opts: {
 
 const isServerActionRequest = (req: Request) => {
   return (
-    !!req.headers.get(nextConstants.Headers.NextUrl) &&
```
panteliselef, line 115: Next-url as a header has been dropped from request to server actions since `14.1.0` way before `14.2.2` where our page detection started to fail.

PR #3976, opened 2024-08-16
feat(ui): Initial AppearanceProvider support

dstaley:
## Description

This PR contains a complete implementation of the `appearance` prop for the new AIO components. Notable changes from the existing `appearance` prop include the dropping of arbitrary CSS (such as pseudo elements) and the removal of component-specific styles like `appearance={{ elements: { signIn: { ... } } }}`.

This PR also contains the following sub-PRs:
- https://github.com/clerk/javascript/pull/4012
- https://github.com/clerk/javascript/pull/4015
- https://github.com/clerk/javascript/pull/3999
- https://github.com/clerk/javascript/pull/3988

## Checklist

- [ ] `npm test` runs as expected.
- [ ] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #3999, opened 2024-08-22
feat(ui): Add support for element descriptors

dstaley:
## Description

This PR adds support for element descriptors, via the `mergeDescriptors` helper utility. The utility is used to apply the correct `className` and `style` attributes for the given descriptors.

This PR also contains the following sub-PRs that setup Vitest:
- https://github.com/clerk/javascript/pull/4012
- https://github.com/clerk/javascript/pull/4015

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/primitives/alert.tsx
```diff
@@ -1,29 +1,61 @@
...
+  type ParsedElementsFragment,
+  type ParsedDescriptor,
+} from '~/contexts/AppearanceContext';
 import * as Icon from './icon';
 
+export const layoutStyle = {
+  alert: {
+    className: 'border px-4 py-3',
+  },
+  alert__warning: {},
```
dstaley, line 15: These empty objects were necessary when `buildTheme` wanted to have every descriptor. With that restriction removed we wouldn't need these. We would need some runtime-validation though because otherwise `parsedElements` would return `undefined` if you tried to access `elements.alert__warning` with just the `layoutTheme`.

packages/ui/src/contexts/AppearanceContext.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+    });
+  });
+
+  it('propogates the provided baseTheme', () => {
+    const testTheme = structuredClone(fullTheme);
+    testTheme.alert__warning.className = 'alert-test-classname';
+
+    const wrapper = ({ children }: { children: ReactNode }) => (
+      <AppearanceProvider
+        theme={testTheme}
```
dstaley, line 88: In a near-future refactor this will be moved into the `appearance` prop instead of being a prop on the provider directly.

packages/ui/src/primitives/alert.tsx
BRKalow: I think we can either rely on `useAppearance()` to provide defaults, or let `mergeDescriptors()` gracefully handle `undefined`

packages/ui/src/themes/buildTheme.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export function buildTheme(p: PartialTheme): ParsedElements {
+  const theme: Partial<ParsedElements> = {};
+
+  for (const descriptor in p) {
+    if (p[descriptor as keyof PartialTheme]) {
+      const { className, style } = p[descriptor as keyof PartialTheme]!;
+      theme[descriptor as keyof ParsedElements] = {
+        descriptor: `cl-${descriptor}`,
+        className: className ?? 'debug',
```
BRKalow, line 14: This is default unnecessary after [this PR](https://github.com/clerk/javascript/pull/4022) right?

packages/ui/src/themes/full.ts
```diff
@@ -0,0 +1,6 @@
+import { visualStyle as alertVisualStyle } from '~/primitives/alert';
+import { buildTheme, mergeTheme } from './buildTheme';
+import { layoutTheme } from './layout';
+
+const visualTheme = buildTheme({ ...alertVisualStyle });
```
BRKalow, line 5: So this will become a large spread of all of our internal appearance objects right?

packages/ui/src/themes/buildTheme.ts
```diff
@@ -0,0 +1,33 @@
...
+        style: style ?? {},
+      };
+    }
+  }
+
+  return theme as ParsedElements;
+}
+
+export function mergeTheme(a: ParsedElements, b: ParsedElements): ParsedElements {
```
BRKalow, line 23: 👀 JSDoc

packages/ui/src/themes/buildTheme.ts
```diff
@@ -0,0 +1,33 @@
...
+  const mergedTheme = { ...a };
+
+  for (const d in mergedTheme) {
+    const descriptor = d as keyof ParsedElements;
+    mergedTheme[descriptor].className = [mergedTheme[descriptor].className, b[descriptor].className].join(' ');
```
BRKalow, line 28: Should we account for empty here? Or is it happening elsewhere?
```suggestion
    mergedTheme[descriptor].className = [mergedTheme[descriptor].className, b[descriptor].className].filter(Boolean).join(' ');
```

BRKalow (commented):
lgtm! left a few minor comments

packages/ui/src/themes/buildTheme.ts
dstaley: Yup, that's in `main` so we'll pick it up once we're in #3976 which targets `main`.

packages/ui/src/themes/full.ts
dstaley: Yup, this was in order to support type checking the argument to `buildTheme` since the individual objects have a `satisfies` constraint that lets them retain their individual keys. When we drop the "complete theme" requirement we won't have to spread the objects.

packages/ui/src/themes/buildTheme.ts
dstaley: yeah this is another place where we're assuming we're provided with two complete themes. When we remove that constraint we will need to add logic to handle situations where a descriptor is in a but not in b, and vice versa.

packages/ui/src/themes/buildTheme.ts
dstaley: 👍 2c629156f508d50d678f00071649dd1c7e61c18e

BRKalow (approved):

PR #3911, opened 2024-08-06
feat(astro): Add CSR support

wobsoriano:
## Description

This PR adds support for `static` and `hybrid` outputs in Astro by making our control components isomorphic. A basic e2e test has been added as well.

Limitations:

- For [`hybrid`](https://docs.astro.build/en/guides/server-side-rendering/#configure-server-or-hybrid) output, a `isStatic` prop is available to control which component to use. It's going to use CSR components by default (since pages are prerendered by default in hybrid output).
- ~~Have to separate the web components to their own files instead of putting them inside the `*CSR.astro` components because of some Vite issues. Please see ECO-17 for more info.~~ Fixed [here](https://github.com/clerk/javascript/pull/3911/commits/0ee00fff12f86173c32ec3edf2347c7b277e03d8) 

Closes **ECO-17**

## Checklist

- [x] `npm test` runs as expected.
- [x] `npm run build` runs as expected.
- [x] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #4000, opened 2024-08-22
fix(astro): Use batched stores to avoid race conditions in auth store

wobsoriano:
## Description

This PR fixes an issue where the `$authStore` nanostore goes back to its initial values (`null/undefined`) after clerk-js has been loaded. See what's happening in the screen recording below:

https://github.com/user-attachments/assets/66a73350-34eb-4d76-9b0a-db91683b0df4

This is happening because `computed` stores react each time any of their dependencies get updated. In our case, the `$csrStore` and `$initialState` stores. It resets back to `undefined` even though we have an initial value because of how the [`derivedStore`](https://github.com/clerk/javascript/blob/main/packages/shared/src/deriveState.ts) function works:

1. We set the `$initialState` coming from SSR, we have an `$authStore` value [available](https://github.com/clerk/javascript/blob/ed7baa0488df0ee4c48add2aac934ffb47e4a6d2/packages/shared/src/deriveState.ts#L13).
2. Next, after clerk-js has been loaded, we set the `isLoaded` property of `$csrStore` to `true`, and `derivedState` will recalculate and [give us the client state and not SSR state](https://github.com/clerk/javascript/blob/ed7baa0488df0ee4c48add2aac934ffb47e4a6d2/packages/shared/src/deriveState.ts#L15), which at this time is still undefined since [it didn't reach the listener for resources yet](https://github.com/clerk/javascript/blob/ed7baa0488df0ee4c48add2aac934ffb47e4a6d2/packages/astro/src/internal/create-clerk-instance.ts#L55). This is where we get `undefined` values again.
3. Next, we are setting each key of the resources (client, user, org, session), and each action will trigger the computed `$authStore` again, giving us different values.

This is where the `batched` store comes in. This store will wait until the end of a tick to update itself. This is what we want for `derivedState` to give a correct output in SSR and CSR:

https://github.com/user-attachments/assets/a89dc4fb-6e99-488c-96b8-da5501e95354

All other stores should work properly as they depend on the `$authStore`.

Fixes ECO-162

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -83,7 +83,7 @@
     "@clerk/shared": "2.5.2",
     "@clerk/types": "4.14.0",
     "nanoid": "5.0.7",
-    "nanostores": "0.10.3",
+    "nanostores": "0.11.2",
     "path-to-regexp": "6.2.2"
   },
   "devDependencies": {
--- a/packages/astro/src/stores/external.ts
+++ b/packages/astro/src/stores/external.ts
@@ -1,6 +1,6 @@
 import { deriveState } from '@clerk/shared/deriveState';
 import { eventMethodCalled } from '@clerk/shared/telemetry';
-import { computed, onMount, type Store } from 'nanostores';
+import { batched, computed, onMount, type Store } from 'nanostores';
 
 import { $clerk, $csrState, $initialState } from './internal';
 
@@ -13,7 +13,7 @@ import { $clerk, $csrState, $initialState } from './internal';
  *
  * $authStore.subscribe((auth) => console.log(auth.userId))
  */
-export const $authStore = computed([$csrState, $initialState], (state, initialState) => {
+export const $authStore = batched([$csrState, $initialState], (state, initialState) => {
   return deriveState(
     state.isLoaded,
     {
--- a/packages/astro/src/stores/internal.ts
+++ b/packages/astro/src/stores/internal.ts
@@ -16,10 +16,10 @@ export const $csrState = map<{
   organization: OrganizationResource | undefined | null;
 }>({
   isLoaded: false,
-  client: null,
-  user: null,
-  session: null,
-  organization: null,
+  client: undefined,
+  user: undefined,
+  session: undefined,
+  organization: undefined,
 });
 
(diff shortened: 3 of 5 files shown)


packages/astro/src/stores/internal.ts
```diff
@@ -16,10 +16,10 @@ export const $csrState = map<{
...
 }>({
   isLoaded: false,
-  client: null,
-  user: null,
-  session: null,
-  organization: null,
+  client: undefined,
+  user: undefined,
+  session: undefined,
+  organization: undefined,
```
wobsoriano, line 22: See my comment [here](https://github.com/clerk/javascript/pull/3911/files#r1722623607). Putting the update here as they're related

packages/astro/package.json
```diff
@@ -83,7 +83,7 @@
     "@clerk/shared": "2.5.2",
     "@clerk/types": "4.14.0",
     "nanoid": "5.0.7",
-    "nanostores": "0.10.3",
+    "nanostores": "0.11.2",
```
wobsoriano, line 86: Version bump with bug fixes and perf improvements. No breaking change. See [changelog](https://github.com/nanostores/nanostores/releases/tag/0.11.0)

packages/astro/src/stores/external.ts
```diff
@@ -13,7 +13,7 @@ import { $clerk, $csrState, $initialState } from './internal';
  *
  * $authStore.subscribe((auth) => console.log(auth.userId))
  */
-export const $authStore = computed([$csrState, $initialState], (state, initialState) => {
+export const $authStore = batched([$csrState, $initialState], (state, initialState) => {
```
panteliselef, line 16: Does this return a read only store similar to computed ?

panteliselef (approved):

packages/astro/src/stores/external.ts
wobsoriano: Yes readonly too

PR #4003, opened 2024-08-22
fix(clerk-js): Update MFA second factor default UI logic

alexcarpenter:
## Description

Currently, when you have both an TOTP + sms code two-step verification, we show the ability to set an SMS code as default but TOTP always takes precedence which makes for confusing behavior where users think they can set the SMS code as the default.

There is also a scenario where a user doesn't have a TOTP code setup, and two sms codes but they are not able to set the default SMS code due to TOTP code not available https://github.com/clerk/javascript/blob/22f19d3bfa987268610e93d2028815d12b374110/packages/clerk-js/src/ui/components/UserProfile/MfaSection.tsx#L96

This PR updates the `MfaPhoneCodeMenu` logic to only allow setting sms code as default when there is no TOTP code set, and disables set default on sms codes when current sms code is already set as default.

https://github.com/user-attachments/assets/a370209b-eda4-4ba2-a322-ff8a27401605

fixes SDKI-615

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/three-feet-nail.md
+++ b/.changeset/three-feet-nail.md
@@ -1,0 +1,5 @@
+---
+"@clerk/clerk-js": patch
+---
+
+Fix issue where MFA two-factor UI was incorrectly rendering set default action in SMS code when TOTP strategy was defined.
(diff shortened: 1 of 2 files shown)


panteliselef (approved):
LGTM

PR #4004, opened 2024-08-22
fix(clerk-js): Fallback to window location for modal sign in/up button flows

alexcarpenter:
## Description

When `<SignInButton />` and `<SignUpButton />` are used in `modal` mode, fallback to current window location when no `fallbackRedirectUrl` prop is provided. This feels like the right behavior default for the usage of `modals` keeping the user on the page, the modal was initiated from.

fixes SDKI-616

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -365,7 +365,7 @@ export class Clerk implements ClerkInterface {
     void this.#componentControls.ensureMounted({ preloadHint: 'SignIn' }).then(controls =>
       controls.openModal('signIn', {
         ...props,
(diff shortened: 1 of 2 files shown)


nikosdouvlis (commented):
👋🏻 
What's the default behavior at the moment? Are we going to change the default or do we simply want to deprecate the prop now?

alexcarpenter:
> 👋🏻 What's the default behavior at the moment? Are we going to change the default or do we simply want to deprecate the prop now?

More context in the ticket here with link to slack thread https://linear.app/clerk/issue/SDKI-616/button-components-in-modal-mode-should-automatically-redirect-back-to

PR #4009, opened 2024-08-22
fix(repo): Adjust npm cache

BRKalow:
## Description

Currently we are caching the `node_modules` folder in CI, however this is totally useless when used in conjunction with `npm ci`, as the command will remove the folder if it exists. Instead, we want to cache npm's internal cache. `actions/setup-node` handles this out of the box with some configuration, which I've added as part of this PR.

We should see some modest speed improvements for two reasons:
* We are no longer caching and restoring `node_modules`, only for it to be immediately deleted by `npm ci`
* We are now caching npm's internal cache, which `npm ci` will use if available.

I've also removed some debug logging and a stray `--verbose` flag, which should also provide some small perf improvement.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


octoper (approved):
🔥

nikosdouvlis (approved):

PR #4010, opened 2024-08-23
fix(clerk-js): Handle multiple `addListener` calls

wobsoriano:
## Description

Closes SDK-1552

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/react/src/isomorphicClerk.ts
```diff
@@ -168,6 +168,14 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
...
   private premountOrganizationListNodes = new Map<HTMLDivElement, OrganizationListProps>();
   private premountMethodCalls = new Map<MethodName<BrowserClerk>, MethodCallback>();
+  // A separate Map of `addListener` method calls to handle multiple listeners.
+  private premountAddListenerCalls = new Map<
+    ListenerCallback,
+    {
+      unsubscribe: UnsubscribeCallback;
+      nativeUnsubscribe?: UnsubscribeCallback;
+    }
+  >();
```
wobsoriano, line 178: The reason we want to add a separate map for `addListener` method calls is that we are already [calling `addListener` internally](https://github.com/clerk/javascript/blob/edbb1d4696a58f4321e50c15d311ca925e81b840/packages/react/src/contexts/ClerkContextProvider.tsx#L31) in our `<ClerkContextProvider>` component to set the auth state, and I think the current implementation supports only one pre-mount listener (`addListener` being the unique key to the `premountMethodCalls` map) and it leads to overwrites/missed events/race condition.

wobsoriano: This is the same idea as the `loadedListeners` array

BRKalow (approved):
👏

PR #4012, opened 2024-08-23
test(ui): Add Vitest with in-source testing

dstaley:
## Description

This PR adds initial Vitest support to `packages/ui`. It does not (yet) support component testing, but is sufficient for testing utility functions. Several changes were necessary to `tsconfig.json` and `tsup.config.ts`. Those changes have been annotated in the PR.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/tsconfig.json
```diff
@@ -8,9 +8,9 @@
     "importHelpers": true,
     "isolatedModules": true,
     "jsx": "react-jsx",
-    "lib": ["es2021.intl", "dom"],
-    "module": "NodeNext",
-    "moduleResolution": "NodeNext",
+    "lib": ["es6", "es2021.intl", "dom"],
```
dstaley, line 11: I'm not completely sure why this wasn't causing errors before, but our `lib` config didn't include a standard library. So I added the `es6` as a reasonable baseline.

packages/ui/tsconfig.json
```diff
@@ -8,9 +8,9 @@
...
+    "module": "es2020",
+    "moduleResolution": "Bundler",
```
dstaley, line 13: TypeScript complained about `import.meta` usage under the `NodeNext` module resolution algorithm. `Bundler` seems to be the right choice anyway since we're bundling with esbuild. I also had to change the `module` to something other than `NodeNext`.

packages/ui/tsconfig.json
```diff
@@ -23,7 +23,8 @@
     "paths": {
       "~/*": ["src/*"]
     },
-    "typeRoots": ["../../node_modules/@types", "./src/types"]
+    "typeRoots": ["../../node_modules/@types", "../../node_modules", "./src/types"],
```
dstaley, line 26: I added `../../node_modules` since we need to be able to resolve the types from the `vitest` package. I'm pretty sure we can drop the `node_modules/@types` since it should be covered by `node_modules`, but I left it here for simplicity.

packages/ui/tsconfig.json
```diff
@@ -23,7 +23,8 @@
...
+    "types": ["vitest/importMeta"]
```
dstaley, line 27: This adds the global types for `import.meta`.

packages/ui/tsup.config.ts
```diff
@@ -40,6 +40,7 @@ export default defineConfig(overrideOptions => {
       PACKAGE_NAME: `"${name}"`,
       PACKAGE_VERSION: `"${version}"`,
       __DEV__: `${!isProd}`,
+      'import.meta.vitest': 'undefined',
```
dstaley, line 43: Defining this to `undefined` during build allows rollup to correctly tree shake the test code out of the source files.

packages/ui/tsup.config.ts
```diff
@@ -52,5 +53,6 @@ export default defineConfig(overrideOptions => {
     minify: false,
     sourcemap: true,
     esbuildPlugins: [tailwindcssTransformerCode],
+    treeshake: true,
```
dstaley, line 56: This enables rollup's tree shaking instead of just relying on the limited tree shaking built into esbuild.

packages/ui/src/contexts/AppearanceContext.tsx
```diff
@@ -302,3 +302,52 @@ const AppearanceProvider = (props: AppearanceProviderProps) => {
 };
 
 export { AppearanceProvider, useAppearance };
+
+if (import.meta.vitest) {
```
BRKalow, line 306: fancy ✨

packages/ui/tsconfig.json
BRKalow: I wonder if we can we remove typeRoots altogether? 🤔

dstaley: it's dead!

BRKalow (approved):

PR #4092, opened 2024-09-03
chore(clerk-js,types,clerk-react,elements): Drop coinbase web3 provider

chanioxaris:
## Description

The current 'coinbase' web3 provider has been replaced by a generic 'coinbase_wallet' provider which supports both Smart Wallets and the Coinbase Wallet browser extension. The feature is not yet released, so it's safe to remove them

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/utils/injectedWeb3Providers.ts
+++ b/packages/clerk-js/src/utils/injectedWeb3Providers.ts
@@ -1,4 +1,4 @@
-import type { CoinbaseWeb3Provider, MetamaskWeb3Provider } from '@clerk/types';
+import type { MetamaskWeb3Provider } from '@clerk/types';
 
 //https://eips.ethereum.org/EIPS/eip-6963
 
@@ -27,12 +27,11 @@ interface EIP6963ProviderDetail {
 }
 
 type EIP6963AnnounceProviderEvent = CustomEvent;
-type InjectedWeb3Provider = MetamaskWeb3Provider | CoinbaseWeb3Provider;
+type InjectedWeb3Provider = MetamaskWeb3Provider;
 
 class InjectedWeb3Providers {
   #providers: EIP6963ProviderDetail[] = [];
   #providerIdMap: Record<InjectedWeb3Provider, string> = {
-    coinbase: 'Coinbase Wallet',
     metamask: 'MetaMask',
   } as const;
   static #instance: InjectedWeb3Providers | null = null;
--- a/packages/elements/src/internals/machines/sign-in/verification.machine.ts
+++ b/packages/elements/src/internals/machines/sign-in/verification.machine.ts
@@ -375,7 +375,7 @@ export const SignInFirstFactorMachine = SignInVerificationMachine.provide({
       );
     }),
     prepare: fromPromise(async ({ input }) => {
-      const { params, parent, resendable } = input as PrepareFirstFactorInput;
+      const { params, parent, resendable } = input;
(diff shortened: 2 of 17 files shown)


packages/elements/src/internals/machines/sign-in/verification.machine.ts
```diff
@@ -448,17 +448,6 @@ export const SignInFirstFactorMachine = SignInVerificationMachine.provide({
...
-          const signature = fields.get('signature')?.value as string | undefined;
-          assertIsDefined(signature, 'Web3 Coinbase signature');
-
-          attemptParams = {
-            strategy,
-            signature,
-          } satisfies Web3Attempt;
-
-          break;
-        }
```
panteliselef, line 461: Are we removing the elements support for Coinbase completely or are we replacing it with Coinbase Wallet ? Will this be done in a follow up PR ?

panteliselef (approved):
Minor feedback about elements

.changeset/afraid-dogs-brake.md
```diff
@@ -0,0 +1,2 @@
+---
```
octoper, line 1: Shouldn't we add a changeset about these changes?

packages/clerk-js/src/utils/injectedWeb3Providers.ts
```diff
@@ -27,12 +27,11 @@ interface EIP6963ProviderDetail {
 }
 
 type EIP6963AnnounceProviderEvent = CustomEvent;
-type InjectedWeb3Provider = MetamaskWeb3Provider | CoinbaseWeb3Provider;
+type InjectedWeb3Provider = MetamaskWeb3Provider;
```
octoper, line 30: ❓ We don't need `CoinbaseWalletWeb3Provider` here?

packages/elements/src/internals/machines/sign-in/verification.machine.ts
chanioxaris: In this PR we are removing the support for the deprecated Coinbase completely. There is a separate PR https://github.com/clerk/javascript/pull/4103 to handle the new Coinbase Wallet provider

.changeset/afraid-dogs-brake.md
chanioxaris: I was thinking to avoid adding one as we never really released the Coinbase provider, so in some way we revert the changes. But sure let me add one

packages/clerk-js/src/utils/injectedWeb3Providers.ts
chanioxaris: No because CoinbaseWallet is not an injected provider but it's handled via the coinbase wallet sdk

packages/elements/src/internals/machines/sign-in/verification.machine.ts
```diff
@@ -375,7 +375,7 @@ export const SignInFirstFactorMachine = SignInVerificationMachine.provide({
       );
     }),
     prepare: fromPromise(async ({ input }) => {
-      const { params, parent, resendable } = input as PrepareFirstFactorInput;
+      const { params, parent, resendable } = input;
```
chanioxaris, line 378:
```suggestion
      const { params, parent, resendable } = input as PrepareFirstFactorInput;
```

PR #4094, opened 2024-09-03
fix(ui): Ensure card take up their max available space

alexcarpenter:
## Description

Up until this point each step has a wrapping `div` or `form` element that comes from the Elements step component. This caused some width issues with the max width on the card and adds an extra element in the dom that we could merge into the card root component.

This PR adds `asChild` props where missing on steps in elements and adds support for an `as` prop on the card root component to render as either a `div` or `form` element. This allows us to use asChild on the step onto the card root reducing the extra element. Reducing the amount of descriptors needed ultimately.

Example:

```tsx
<SignIn.Step asChild name="start">
  <Card.Root as="form">...</Card.Root>
</SignIn.Step>
```

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2024-09-04 at 12 47 32 PM | Screenshot 2024-09-04 at 12 46 27 PM |

Fixes SDKI-641

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


joe-bell (approved):
All working for me! Nice work @alexcarpenter

PR #4095, opened 2024-09-03
fix(clerk-js): Ensure we do not access `addEventListener` and `dispatchEvent` in non-browser environments

wobsoriano:
## Description

This PR makes sure to **NOT** call browser specific methods (`addEventListener`, `dispatchEvent`) implemented in [4059](https://github.com/clerk/javascript/pull/4059) in non-browser environments. Even though we are trying to block calls by [checking `window` object](https://github.com/clerk/javascript/blob/c63a5adf0ba4b99252146f168318f51b709bb5dd/packages/clerk-js/src/utils/injectedWeb3Providers.ts#L40), it'll still pass because React Native does have some polyfills to the `window` object.

Fixes SDK-1922

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/utils/injectedWeb3Providers.ts
+++ b/packages/clerk-js/src/utils/injectedWeb3Providers.ts
@@ -35,13 +35,18 @@ export class InjectedWeb3Providers {
     coinbase: 'Coinbase Wallet',
     metamask: 'MetaMask',
   } as const;
+  #hasSetupEventListeners = false;
 
   constructor() {
     if (typeof window === 'undefined') {
       return;
     }
-    window.addEventListener('eip6963:announceProvider', this.#onAnnouncement as EventListener);
-    window.dispatchEvent(new Event('eip6963:requestProvider'));
+
+    if (!this.#hasSetupEventListeners) {
+      window.addEventListener('eip6963:announceProvider', this.#onAnnouncement as EventListener);
+      window.dispatchEvent(new Event('eip6963:requestProvider'));
+      this.#hasSetupEventListeners = true;
+    }
   }
 
   get = (provider: InjectedWeb3Provider) => {
--- a/packages/clerk-js/src/utils/injectedWeb3Providers.ts
+++ b/packages/clerk-js/src/utils/injectedWeb3Providers.ts
@@ -29,7 +29,7 @@ interface EIP6963ProviderDetail {
 type EIP6963AnnounceProviderEvent = CustomEvent;
 type InjectedWeb3Provider = MetamaskWeb3Provider | CoinbaseWeb3Provider;
 
-class InjectedWeb3Providers {
+export class InjectedWeb3Providers {
   #providers: EIP6963ProviderDetail[] = [];
   #providerIdMap: Record<InjectedWeb3Provider, string> = {
     coinbase: 'Coinbase Wallet',
@@ -55,5 +55,3 @@ class InjectedWeb3Providers {
     this.#providers.push(event.detail);
   };
 }
-
-export const injectedWeb3Providers = new InjectedWeb3Providers();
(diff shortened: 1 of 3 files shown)


packages/clerk-js/src/utils/injectedWeb3Providers.ts
```diff
@@ -55,5 +55,3 @@ class InjectedWeb3Providers {
     this.#providers.push(event.detail);
   };
 }
-
-export const injectedWeb3Providers = new InjectedWeb3Providers();
```
wobsoriano, line 59: Exporting this makes it global, and errors out in non-browser environments

packages/clerk-js/src/utils/web3.ts
```diff
@@ -80,5 +80,7 @@ async function getEthereumProvider(provider: Web3Provider) {
     const sdk = new CoinbaseWalletSDK({});
     return sdk.makeWeb3Provider({ options: 'all' });
   }
+
+  const injectedWeb3Providers = new InjectedWeb3Providers();
```
wobsoriano, line 84: Makes sense to just instantiate it here

packages/clerk-js/src/utils/injectedWeb3Providers.ts
```diff
@@ -29,19 +29,24 @@ interface EIP6963ProviderDetail {
...
       return;
     }
-    window.addEventListener('eip6963:announceProvider', this.#onAnnouncement as EventListener);
-    window.dispatchEvent(new Event('eip6963:requestProvider'));
+
+    if (!this.#hasSetupEventListeners) {
+      window.addEventListener('eip6963:announceProvider', this.#onAnnouncement as EventListener);
+      window.dispatchEvent(new Event('eip6963:requestProvider'));
+      this.#hasSetupEventListeners = true;
+    }
```
wobsoriano, line 49: Make sure we only add a listener once, since we moved the instantiation [here](https://github.com/clerk/javascript/pull/4095/files#r1742721324)

packages/clerk-js/src/utils/injectedWeb3Providers.ts
```diff
@@ -56,4 +64,4 @@ class InjectedWeb3Providers {
   };
 }
 
-export const injectedWeb3Providers = new InjectedWeb3Providers();
+export const getInjectedWeb3Providers = () => InjectedWeb3Providers.getInstance();
```
wobsoriano, line 67: Exporting it now as a function but making sure it's still instantiated once

packages/clerk-js/src/utils/injectedWeb3Providers.ts
```diff
@@ -35,15 +35,23 @@ class InjectedWeb3Providers {
     coinbase: 'Coinbase Wallet',
     metamask: 'MetaMask',
   } as const;
+  static #instance: InjectedWeb3Providers | null = null;
 
-  constructor() {
+  private constructor() {
     if (typeof window === 'undefined') {
```
BRKalow, line 41: Odd that this didn't catch it. Does this mean expo is defining `window`, but _not_ `addEventListener`? 😵

BRKalow (approved):
👏

packages/clerk-js/src/utils/injectedWeb3Providers.ts
wobsoriano: React Native is defining `window` [here](https://github.com/facebook/react-native/blob/12b64b78247f2db929c6d421262bdec1f6402e0b/packages/react-native/Libraries/Core/setUpGlobals.js#L18) 🤯

wobsoriano: and looks like expo too

Screenshot 2024-09-03 at 3 10 26 PM

PR #4097, opened 2024-09-04
feat(nextjs,backend,types): adds accountless signup to nextjs app router (opt in only for now)

davidlumley:
## Description

Before, users needed to sign up to Clerk to copy and paste environment keys before they got started building. This is a bit of friction when you're trying to focus on learning a new framework or building your business logic.

This feature attempts to eliminate this requirement when ran in development mode.

It does this by first trying to load ephemeral keys from a local cache. If these keys do not exist, it fetches them from Clerk's API and stores them locally before using them for any future Clerk requests.

To enable this, the environment variable `CLERK_FEATURE_FLAG_EPHEMERAL_ACCOUNTS` must be set to true.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -22,6 +22,10 @@
         "import": "./dist/runtime/node/crypto.mjs"
       },
       "default": "./dist/runtime/browser/crypto.mjs"
+    },
+    "#ephemeral": {
+      "node": "./dist/runtime/node/ephemeral.js",
+      "default": "./dist/runtime/browser/ephemeral.js"
     }
   },
   "exports": {
--- a/packages/backend/src/runtime/node/ephemeral.js
+++ b/packages/backend/src/runtime/node/ephemeral.js
@@ -1,0 +1,154 @@
+const fs = require('node:fs');
+const path = require('node:path');
+
+const { isPublishableKey } = require('@clerk/shared');
+
+async function createEphemeralAccount() {
+  const body = await postJSON('https://api.clerkstage.dev/v1/public/demo_instance');
+
+  const account = {
+    publishableKey: body.frontend_api_key,
+    secretKey: body.backend_api_key,
+    expiresAt: daysFromNow(1),
+  };
+
+  write(account);
+
+  return account;
+}
+
+async function fetchEphemeralAccount() {
+  const account = read();
+
+  if (account) {
+    const verified = await verifyEphemeralAccount(account);
+
+    if (verified) {
+      return account;
+    }
+  }
+
+  return await createEphemeralAccount();
+}
+
+async function verifyEphemeralAccount(account) {
+  // TODO: API not implemented yet.
+  //
+  // try {
+  //   await postJSON('https://api.clerkstage.dev/v1/ephemeral-account/verify', {
+  //     publishable_key: account.publishableKey,
+  //     secret_key: account.secretKey,
+  //   })
+  //
+  //   return true
+  //
+  // } catch (err) {
+  //   if (err instanceof ClientError) {
+  //     return false;
+  //   }
+  //
+  //   throw err;
+  // }
+
+  if (!isPublishableKey(account.publishableKey)) {
+    return false;
+  }
+
+  // NOTE: Simulate failed verification after 15 minutes
+  if (account.expiresAt < daysFromNow(1) - 60 * 15) {
+    return false;
+  }
+
+  return true;
+}
+
+const PATH = path.join(process.cwd(), 'node_modules', '.cache', 'clerkjs', 'ephemeral.json');
+
+function read() {
+  try {
+    if (fs.existsSync(PATH)) {
+      const config = JSON.parse(fs.readFileSync(PATH, { encoding: 'utf-8' }));
+
+      if (config.expiresAt < now()) {
+        return null;
+      }
+
+      return config;
+    }
+  } catch (error) {
+    if (error instanceof Error && error.message.includes('ENOENT')) {
+      return null;
+    }
+
+    throw error;
+  }
+}
+
+function write(obj) {
+  fs.mkdirSync(path.dirname(PATH), { recursive: true });
+
+  const content = JSON.stringify(obj);
+
+  fs.writeFileSync(PATH, content, {
+    encoding: 'utf8',
+    mode: '0777',
+    flag: 'w',
+  });
+}
+
+function postJSON(url, body) {
+  return apiResult(
+    fetch(url, {
+      method: 'POST',
+      body: body ? JSON.stringify(body) : undefined,
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    }),
+  );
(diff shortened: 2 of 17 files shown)


packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -178,6 +191,45 @@ export const clerkMiddleware: ClerkMiddleware = (...args: unknown[]): any => {
...
+        const options = {
+          expires: Number(ephemeralAccount.expiresAt) * 1000,
+        };
+
+        response.cookies.set(constants.Cookies.EphemeralExpiresAt, ephemeralAccount.expiresAt.toString(), options);
+        response.cookies.set(constants.Cookies.EphemeralPublishableKey, ephemeralAccount.publishableKey, options);
+        response.cookies.set(constants.Cookies.EphemeralSecretKey, ephemeralAccount.secretKey, options);
+
+        return response;
+      }
```
BRKalow, line 218: Can we get rid of the query param -> redirect -> cookies -> redirect flow? We do have a way to pass keys from the middleware to the app server as of a few months back, so potentially we can have a flow that originates in ClerkProvider, does a single redirect, and the middleware sets up the options so they're available to the subsequent request.

packages/nextjs/src/app-router/server/auth.ts
```diff
@@ -51,6 +50,25 @@ export const auth = (): Auth => {
...
+  // NOTE: This is a workaround to allow the development environment to use the clerk key
+  // without having to set the environment variables.
+  try {
+    return baseAuth();
+  } catch (e: any) {
+    if (isClerkKeyError(e)) {
+      return initialState();
+    }
+    throw e;
+  }
```
BRKalow, line 69: This is necessary to allow the initial request w/o keys to proceed right? During the flow to fetch and make the ephemeral keys available

packages/nextjs/src/app-router/server/ClerkProvider.tsx
```diff
@@ -8,14 +10,43 @@ import { ClientClerkProvider } from '../client/ClerkProvider';
...
   const state = initialState()?.__clerk_ssr_state as InitialState;
   const cspHeader = headers().get('Content-Security-Policy');
 
+  const providerProps = { ...mergeNextClerkPropsWithEnv(rest) };
+
+  if (!providerProps.publishableKey) {
+    const ephemeralAccount = await fetchEphemeralAccount();
+    const cookieExpiresAt = cookies().get(constants.QueryParameters.EphemeralExpiresAt)?.value;
+    const cookiePublishableKey = cookies().get(constants.QueryParameters.EphemeralPublishableKey)?.value;
+    const cookieSecretKey = cookies().get(constants.QueryParameters.EphemeralSecretKey)?.value;
```
BRKalow, line 26: ❓ If we assume that we no longer automatically fetch fresh keys, do we need to store the keys in cookies? I think no. We'll have our API return the status of the instance / keys (expired or not) so we can render some sort of callout.

packages/backend/src/runtime/node/ephemeral.js
```diff
@@ -0,0 +1,154 @@
+const fs = require('node:fs');
+const path = require('node:path');
+
+const { isPublishableKey } = require('@clerk/shared');
+
+async function createEphemeralAccount() {
+  const body = await postJSON('https://api.clerkstage.dev/v1/public/demo_instance');
```
BRKalow, line 7: Once implemented: also fetch a claim token of some sort so we can build a URL to sign up and claim the instance.

BRKalow (commented):
Walked through this with @mwickett and left some notes, no action needed at the moment necessarily. I think once we define the API changes necessary and what the responses will look like there we'll be able to finalize this piece of the implementation.

Great start!

packages/nextjs/src/server/clerkMiddleware.ts
davidlumley: Yeah, I think we can probably get this down to a single redirect - will do some testing to make sure that works as intended.

packages/nextjs/src/app-router/server/ClerkProvider.tsx
davidlumley: I don't think we need to use cookies at all, will do some testing this week.

packages/backend/src/runtime/node/ephemeral.js
```diff
@@ -0,0 +1,154 @@
...
+      method: 'POST',
+      body: body ? JSON.stringify(body) : undefined,
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    }),
+  );
+}
+
+async function apiResult(whenResponse) {
```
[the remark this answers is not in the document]
davidlumley: Yes! I'll take a look and see if we can't get this working.

packages/nextjs/src/app-router/server/ClerkProvider.tsx
```diff
@@ -8,14 +10,43 @@ import { ClientClerkProvider } from '../client/ClerkProvider';
...
+
+    if (stale) {
+      const params = new URLSearchParams({
+        [constants.QueryParameters.EphemeralExpiresAt]: String(ephemeralAccount.expiresAt),
+        [constants.QueryParameters.EphemeralPublishableKey]: ephemeralAccount.publishableKey,
+        [constants.QueryParameters.EphemeralSecretKey]: ephemeralAccount.secretKey,
+      });
+
+      redirect(`?${params}`);
+    }
```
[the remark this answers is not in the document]
davidlumley: Yeah, think this was some edge case handling @ayrton had added to prevent the front end from being in a different state to the backend.

I think we can probably remove this in favor of the state that is propagated from the middleware.

packages/backend/package.json
```diff
@@ -22,6 +22,10 @@
         "import": "./dist/runtime/node/crypto.mjs"
       },
       "default": "./dist/runtime/browser/crypto.mjs"
+    },
+    "#ephemeral": {
+      "node": "./dist/runtime/node/ephemeral.js",
+      "default": "./dist/runtime/browser/ephemeral.js"
```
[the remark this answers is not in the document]
davidlumley: Great question - this was something I added in order to get the `fs` functions that are only supported in the `node` runtime environment working. If the general mechanism of us writing / reading feels good, I can spend some time trying to find a place in the monorepo that makes more sense and builds without errors.

packages/nextjs/src/app-router/server/auth.ts
davidlumley: Yeah exactly — this may be different with the keys being propagated through the middleware, so will do some testing and we may be able to remove.

clerk-cookie:
Hello 👋

PR #4112, opened 2024-09-05
fix(clerk-js): Ensure getToken() reads correct organization ID

BRKalow:
## Description

Fixes a bug where `getToken()` would read stale organization data from `clerk.organization`. This is problematic because in the `setActive()` flow we call `getToken()` before `clerk.organization` is updated, and so we were reading a stale org ID when attempting to fetch a token for a new session.

For impersonation specifically, this was problematic as the impersonator's session would be in memory while `setActive()` was being called with the impersonated session.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/bright-falcons-move.md
+++ b/.changeset/bright-falcons-move.md
@@ -1,0 +1,5 @@
+---
+"@clerk/clerk-js": patch
+---
+
+Fix bug where session.getToken() was reading a stale organization ID.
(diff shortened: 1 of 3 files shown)


izaaklauer (approved):

NicolasLopes7 (approved):
👍

PR #4118, opened 2024-09-06
feat(clerk-js,clerk-react,types,nextjs): Support assurance in `has()`

panteliselef:
## Description

This PR adds experimental support for `has()` with assurance.
```ts
has({ 
  __experimental_assurance: {
    level: 'L2.secondFactor',
    maxAge: 'A1.10min'
  }
})
```

Also I introduced a new shared utility called `createCheckAuthorization` exported from `@clerk/shared` that all other packages can use.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/react/src/hooks/__tests__/useAuth.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      expectTypeOf({} as const).toMatchTypeOf<ParamsOfHas>();
+    });
+
+    it('has({role: string}) is allowed', () => {
+      expectTypeOf({ role: 'org:admin' }).toMatchTypeOf<ParamsOfHas>();
+    });
+
+    it('has({role: string, permission: string}) is NOT allowed', () => {
+      expectTypeOf({ role: 'org:admin', permission: 'some-perm' }).not.toMatchTypeOf<ParamsOfHas>();
+    });
```
panteliselef, line 20: making sure this did not break

packages/react/src/hooks/useAuth.ts
```diff
@@ -53,7 +54,7 @@ type UseAuthReturn =
       orgId: null;
       orgRole: null;
       orgSlug: null;
-      has: CheckAuthorizationWithoutOrgOrUser;
+      has: CheckAuthorizationWithCustomPermissions;
```
panteliselef, line 57: Assurance can apply to users that do not have an active org

packages/react/src/hooks/useAuth.ts
```diff
@@ -199,7 +191,7 @@ export const useAuth: UseAuth = () => {
       orgId: null,
       orgRole: null,
       orgSlug: null,
-      has: () => false,
+      has,
```
panteliselef, line 194: Assurance can apply to users that do not have an active org

packages/shared/src/authorization.ts
```diff
@@ -0,0 +1,61 @@
...
+  'A5.1wk': 10080, //7 * 24 * 60,
+};
+
+export const createCheckAuthorization = (options: {
+  userId: string | null | undefined;
+  orgId: string | null | undefined;
+  orgRole: string | null | undefined;
+  orgPermissions: string[] | null | undefined;
+  __experimental_factorVerificationAge: [number, number] | null;
+}): CheckAuthorizationWithCustomPermissions => {
```
panteliselef, line 17: A new utility to avoid duplication between clerk-js/react/backend

packages/shared/src/authorization.ts
```diff
@@ -0,0 +1,61 @@
...
+      if (params.__experimental_assurance.level === 'L1.firstFactor') {
+        stepUpAuthorization = hasValidFactorOne;
+      } else if (params.__experimental_assurance.level === 'L2.secondFactor') {
+        stepUpAuthorization = hasValidFactorTwo;
+      } else {
+        stepUpAuthorization = hasValidFactorOne && hasValidFactorTwo;
+      }
+    }
+
+    return [orgAuthorization, stepUpAuthorization].filter(Boolean).some(a => a === true);
```
LauraBeatris, line 59: I don't get this part. So if the role/permissions check fails, but step-up check succeeds, then `has` would return `true` either way?

LauraBeatris (approved):
step-up auth coming soon ❤️

packages/shared/src/authorization.ts
panteliselef: Very good catch, this would come back and bite us. I've updated this logic, and added a bunch of unit tests to make sure each case has the expected outcome.

packages/shared/src/authorization.ts
```diff
@@ -0,0 +1,132 @@
...
+      return factor2Age !== -1 ? isValidFactor2 : isValidFactor1;
+    case 'L3.multiFactor':
+      return factor2Age === -1 ? isValidFactor1 : isValidFactor1 && isValidFactor2;
+  }
+};
+
+/**
+ * Creates a function for comprehensive user authorization checks.
+ * Combines organization-level and step-up authentication checks.
+ * The returned function authorizes if both checks pass, or if at least one passes
```
LauraBeatris, line 114: @panteliselef Makes sense now! 🫡

LauraBeatris (approved):

PR #4134, opened 2024-09-10
fix(clerk-expo): Ensure user-agent is set correctly for iOS requests

wobsoriano:
## Description

WIP, please do not merge. Not sure yet why the lockfile is that big and we might have a better way of retrieving the `user-agent` in iOS devices.

This PR fixes an issue where `isMobile` value of a [`SessionActivity`](https://clerk.com/docs/references/javascript/session-with-activities#session-activity) is always `false` on iOS devices. The backend is likely using a library to determine device type via user-agent string. It expects a "Mobile" word to indicate a mobile device.

Resolves ECO-189

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46206,6 +46206,7 @@
       },
       "peerDependencies": {
         "expo-auth-session": ">=5",
+        "expo-constants": ">=15.4.5",
         "expo-local-authentication": ">=13.5.0",
         "expo-secure-store": ">=12.4.0",
         "expo-web-browser": ">=12.5.0",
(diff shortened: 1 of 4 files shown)


package-lock.json
wobsoriano, line 1: why is this big 😠

packages/expo/package.json
```diff
@@ -79,11 +80,12 @@
   "peerDependencies": {
     "expo-auth-session": ">=5",
     "expo-local-authentication": ">=13.5.0",
+    "expo-constants": ">=15.4.5",
```
wobsoriano, line 83: tagged on npm as `sdk50`

PR #4161, opened 2024-09-13
chore(clerk-js): Use getWeb3ProviderData() instead of duplicating logic

chanioxaris:
## Description

Instead of duplicating the logic to get a Web3Provider from a Web3Strategy, we make sure to DRY them by using the getWeb3ProviderData() method from the types package

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/clerk.ts
```diff
@@ -1418,7 +1419,7 @@ export class Clerk implements ClerkInterface {
     if (!this.client || !this.environment) {
       return;
     }
-    const provider = strategy.replace('web3_', '').replace('_signature', '') as Web3Provider;
+    const provider = getWeb3ProviderData({ strategy })?.provider as Web3Provider;
```
panteliselef, line 1422: Importing from a function from types feels very wrong. Let's discuss offline how we can resolve this.

panteliselef (changes requested):

packages/clerk-js/src/core/clerk.ts
alexcarpenter: There already looks to be a similar function in clerk-js but unused https://github.com/clerk/javascript/blob/main/packages/clerk-js/src/ui/common/constants.ts#L101

chanioxaris: So that's a helper we expose from the start of web3 support and it's useful to map strategy/provider name to a Web3 provider. We even use on other Clerk products (e.g. Dashboard). Happy to discuss if we would like to follow a different path here, if we don't like what we have right now

chanioxaris:
Closing this in order to handle the methods duplication as well as part of a separate effort and move any utilities out of the types package and into the shared one

PR #4162, opened 2024-09-13
chore(e2e): Add tests for custom pages and items from UserProfile and UserButton

panteliselef:
## Description

We were missing tests for any kind of functionality for Custom pages and custom items. Given the fact that we may introduce new APIs for UB, having these tests will make sure that nothing breaks.

Test cases introduced:
**For UB and UP Custom pages**
- Custom pages and link are set in the correct order
- Label, Icons, and content is correctly displayed
- Wrapping a page with context will work, and the custom page will get updated when state changes
- Absolute urls in Link will be respected
- Relative url in Link will be respected

**For UB Custom items**
- items are specified in the correct order
- Label, Icons, and content is correctly displayed
- Absolute urls in Link will be respected
- Relative url in Link will be respected
- Custom actions on Click work as expected

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


tmilewski (approved):

integration/templates/react-vite/src/custom-user-button/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+        </UserButton.UserProfilePage>
+        <UserButton.UserProfilePage label={'security'} />
+        <UserButton.UserProfilePage
+          label={'Page 2'}
+          labelIcon={<p data-label-icon={'page-2'}>🙃</p>}
+          url='page-2'
+        >
+          <h1>Page 2</h1>
+        </UserButton.UserProfilePage>
+        🌐
```
alexcarpenter, line 41: Is this purposefully in between items here?

panteliselef: Yes. Although i can't think of a way we can break this. But be safe than sorry.

alexcarpenter: Got it, if you are actively testing that the component doesn't render this, it'd be helpful to know that in some fashion, wrap in a button that has a text documenting that it shouldn't be rendered vs just having an emoji dropped here. Non-blocking thought.

alexcarpenter (approved):

Issue #4142, opened 2024-09-10
Having a complex Astro config throws "Converting circular structure to JSON" when using Clerk components.


ernestoalejo:
### Preliminary Checks

- [X] I have reviewed the documentation: https://clerk.com/docs

- [X] I have searched for existing issues: https://github.com/clerk/javascript/issues

- [X] I have not already reached out to Clerk support via email or Discord (if you have, no need to open an issue here)

- [X] This issue is not a question, general help request, or anything other than a bug report directly related to Clerk. Please ask questions in our Discord community: https://clerk.com/discord.

### Reproduction

https://stackblitz.com/edit/github-rdq1n6?file=astro.config.mjs

### Publishable key

pk_test_YW1hemVkLXR1cmtleS02Mi5jbGVyay5hY2NvdW50cy5kZXYk

### Description

**Steps to reproduce:**

1. Initialize a new Astro project with Vue.
2. Add Clerk with the official quickstart: https://clerk.com/docs/quickstarts/astro

If the vue plugin, or any other complex enough config, is added before the Clerk plugin runs and reads the config then a `Converting circular structure to JSON` throws when trying to use the components.

This occurs because the Astro config is not guaranteed to be JSON serializable here: https://github.com/clerk/javascript/blob/b97b2c1cae5cb1e569708a8745c13d203beb81d9/packages/astro/src/integration/vite-plugin-astro-config.ts#L33

This also means that the plugin sees an incomplete vision of the config object depending on when it runs so it's not useful anyway.

From the reproduction:

```typescript
export default defineConfig({
  adapter: node({ mode: 'standalone' }),
  output: 'server',

  // Works
  integrations: [clerk()],

  // Works:
  // integrations: [clerk(), vue()],

  // Doesn't work, circular structure:
  // integrations: [vue(), clerk()],
});
```

**Expected behavior:**

It works with any Astro config file.

**Actual behavior:**

![image](https://github.com/user-attachments/assets/9726ff00-a525-472c-bf23-288bd735e37e)

### Environment

```Shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
[...]

PR #4163, opened 2024-09-13
fix(astro): Prevent circular JSON error with complex Astro configs

wobsoriano:
## Description

Previously, the integration was [stringifying the entire Astro config](https://github.com/clerk/javascript/blob/d8954948d08e3d8090b2be41b4670d76341c6dfd/packages/astro/src/integration/vite-plugin-astro-config.ts#L33), which causes a a "Converting circular structure to JSON" error when the Clerk integration was last in the integrations config. This PR fix removes the full config stringification and instead only uses the `output` property from the Astro config, as that's the only part needed.

Resolves SDK-1930
Fixes https://github.com/clerk/javascript/issues/4142

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/astro/src/integration/vite-plugin-astro-config.ts
+++ b/packages/astro/src/integration/vite-plugin-astro-config.ts
@@ -30,10 +30,10 @@ export function vitePluginAstroConfig(astroConfig: AstroConfig): VitePlugin {
     load(id) {
       if (id === resolvedVirtualModuleId) {
         return `
-          export const astroConfig = ${JSON.stringify(astroConfig)};
+          const configOutput = '${astroConfig.output}';
 
           export function isStaticOutput(forceStatic) {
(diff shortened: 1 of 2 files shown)


packages/astro/src/integration/vite-plugin-astro-config.ts
```diff
@@ -30,10 +30,10 @@ export function vitePluginAstroConfig(astroConfig: AstroConfig): VitePlugin {
     load(id) {
       if (id === resolvedVirtualModuleId) {
         return `
-          export const astroConfig = ${JSON.stringify(astroConfig)};
```
wobsoriano, line 33: We actually are not using this anywhere in the astro codebase, and this is internal, so it should be safe too remove the export and just use the needed property

panteliselef (approved):

PR #4171, opened 2024-09-16
fix(backend): Do not cache 307 redirects

nikosdouvlis:
## Description

Chrome aggressively caches inactive tabs. If we don't set the header here, all 307 redirects will be cached and the handshake will end up in an infinite loop.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/tokens/request.ts
+++ b/packages/backend/src/tokens/request.ts
@@ -121,7 +121,7 @@ export async function authenticateRequest(
       newUrl.searchParams.delete(constants.QueryParameters.Handshake);
       newUrl.searchParams.delete(constants.QueryParameters.HandshakeHelp);
       headers.append(constants.Headers.Location, newUrl.toString());
-      headers.append(constants.Headers.CacheControl, 'no-store');
+      headers.set(constants.Headers.CacheControl, 'no-store');
     }
 
     if (sessionToken === '') {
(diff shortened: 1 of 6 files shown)


nikosdouvlis:
!snapshot

packages/backend/src/tokens/request.ts
```diff
@@ -120,7 +120,8 @@ export async function authenticateRequest(
       const newUrl = new URL(authenticateContext.clerkUrl);
       newUrl.searchParams.delete(constants.QueryParameters.Handshake);
       newUrl.searchParams.delete(constants.QueryParameters.HandshakeHelp);
-      headers.append('Location', newUrl.toString());
+      headers.append(constants.Headers.Location, newUrl.toString());
+      headers.append(constants.Headers.CacheControl, 'no-store');
```
anagstef, line 124: @nikosdouvlis can we extract `no-store` value as a constant?

packages/backend/src/tokens/request.ts
```diff
@@ -174,6 +175,13 @@ ${error.getFullMessage()}`,
       // Right now the only usage of passing in different headers is for multi-domain sync, which redirects somewhere else.
       // In the future if we want to decorate the handshake redirect with additional headers per call we need to tweak this logic.
       const handshakeHeaders = headers ?? buildRedirectToHandshake();
+
+      // Chrome aggressively caches inactive tabs. If we don't set the header here,
+      // all 307 redirects will be cached and the handshake will end up in an infinite loop.
+      if (handshakeHeaders.get(constants.Headers.Location)) {
+        handshakeHeaders.append(constants.Headers.CacheControl, 'no-store');
```
agis, line 182: 🔧 I believe it's safer to overwrite any existing values here (i.e. `set`), just in case some code earlier in the path (e.g. another middleware) already sets `Cache-Control: public`.

nikosdouvlis: Good point

agis (approved):

anagstef (approved):

PR #4172, opened 2024-09-16
fix(shared): Add missing entries to "files" array in package.json

LekoArts:
## Description

The last couple of PRs merged did miss adding the new subpaths to the `files` array in package.json. The script that should have checked it miss it, it'll catch it now.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/subpaths.mjs
+++ b/packages/shared/subpaths.mjs
@@ -1,5 +1,5 @@
 // This file is a helper for the "subpath-workaround.mjs" script
-// We have to polyfill our "exports" subpaths :cry:
+// When adding an entry to "subpathNames" also add it to "files" in package.json
 
 export const subpathNames = [
   'authorization',
--- a/scripts/subpath-workaround.mjs
+++ b/scripts/subpath-workaround.mjs
@@ -30,6 +30,11 @@ async function run() {
     ...subpathHelperFile.ignoredFolders,
     'dist',
   ];
+
+  if (pkgFile.files.length !== allFilesNames.length) {
+    throw new Error('The package.json "files" array length does not match the subpaths.mjs');
+  }
+
   const hasAllSubpathsInFiles = pkgFile.files.every(name => allFilesNames.includes(name));
 
   if (!hasAllSubpathsInFiles) {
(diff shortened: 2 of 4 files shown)


scripts/subpath-workaround.mjs
```diff
@@ -30,6 +30,11 @@ async function run() {
     ...subpathHelperFile.ignoredFolders,
     'dist',
   ];
+
+  if (pkgFile.files.length !== allFilesNames.length) {
+    throw new Error('The package.json "files" array length does not match the subpaths.mjs');
+  }
+
```
anagstef, line 37: ❓ is this almost the same error as the one that exists right below?

packages/shared/subpaths.mjs
```diff
@@ -1,5 +1,5 @@
 // This file is a helper for the "subpath-workaround.mjs" script
-// We have to polyfill our "exports" subpaths :cry:
+// When adding an entry to "subpathNames" also add it to "files" in package.json
 
 export const subpathNames = [
```
panteliselef, line 4: Can't we just replace this with package.files ? In order to use package.json as the source of truth ?

LekoArts: The `files` array contains directories (like `dist`, `scripts`, `react`) that need special handling and/or shouldn't get subdirectories created. So one way or another you'd need to add special handling. I can adjust the logic in another PR probably so that `files` array is the source of truth

scripts/subpath-workaround.mjs
LekoArts: The `.every` error won't fire if the length is not the same, so early return here and make the error clearer

panteliselef (approved):

scripts/subpath-workaround.mjs
anagstef: We could merge these 2 checks and throw a single error if any of these checks fail.

tmilewski (approved):

PR #4185, opened 2024-09-18
fix(clerk-js): Restore behavior of MetaMask compatible Web3 wallets

chanioxaris:
## Description

Before, even if a user didn't use the MetaMask browser extension but a compatible one, such as Rabby Wallet, it was possible to use it as they share the same API to authenticate themselves. This behavior stopped working when we added support for EIP6963 regarding handling multiple injected providers. This commit restores the previous behavior by using the existing injected provider if there is a single one

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/utils/injectedWeb3Providers.ts
```diff
@@ -52,6 +52,12 @@ class InjectedWeb3Providers {
...
 
   get = (provider: InjectedWeb3Provider) => {
+    // In case there is a single injected provider, use that directly.
+    // This is needed in order to support other compatible Web3 Wallets (e.g. Rabby Wallet)
+    // and maintain the previous behavior
+    if (this.#providers.length === 1) {
+      return this.#providers[0].provider;
+    }
     return this.#providers.find(p => p.info.name === this.#providerIdMap[provider])?.provider;
   };
```
panteliselef, line 62: Just confirming here: If a user has a single wallet installed that is not metamask we accept it. But if the have Metamask and Rabby only metamask will be accepted ?

chanioxaris: That's correct and hopefully that will solve the issues that some users had after this change

panteliselef (approved):

PR #4187, opened 2024-09-18
chore(shared): Exports `match` utility from the `path-to-regexp` lib

izaaklauer:
### What changed?

This surfaces a new path-to-regexp function, [match](https://github.com/pillarjs/path-to-regexp/tree/6.x?tab=readme-ov-file#match), that upcoming functionality depends on.

### How did you generate the minified js?

Thanks to @nikosdouvlis for giving me the run-down. The procedure goes:

To create the new minified code (high-level)
- Create a minimal typescript project with an index.ts that imports the functionality that you need
- Use tsup to bundle it into a minified file
- Beautify the minified file (e.g. add newlines)

More details for internal folks here: https://clerkinc.slack.com/archives/C07LN8YS9HB/p1726666913614259?thread_ts=1726240424.533199&cid=C07LN8YS9HB

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/big-elephants-add.md
+++ b/.changeset/big-elephants-add.md
@@ -1,0 +1,5 @@
+---
+"@clerk/shared": patch
+---
+
+Expose match from `path-to-regexp`.
--- a/.changeset/big-elephants-add.md
+++ b/.changeset/big-elephants-add.md
@@ -2,4 +2,4 @@
 "@clerk/shared": patch
 ---
 
-Expose match from `path-to-regexp`.
+Exports `match` utility from the `path-to-regexp` lib.
--- a/packages/shared/src/pathToRegexp.ts
+++ b/packages/shared/src/pathToRegexp.ts
@@ -1,9 +1,17 @@
+import type {
+  Match,
+  MatchFunction,
+  ParseOptions,
+  Path,
+  RegexpToFunctionOptions,
+  TokensToRegexpOptions,
+} from './compiled/path-to-regexp';
 import { match as matchBase, pathToRegexp as pathToRegexpBase } from './compiled/path-to-regexp';
 
 export const pathToRegexp = (path: string) => {
   try {
     // @ts-ignore no types exists for the pre-compiled package
-    return pathToRegexpBase(path) as RegExp;
+    return pathToRegexpBase(path);
   } catch (e: any) {
     throw new Error(
       `Invalid path: ${path}.\nConsult the documentation of path-to-regexp here: https://github.com/pillarjs/path-to-regexp/tree/6.x\n${e.message}`,
@@ -17,7 +25,7 @@ export function match<P extends object = object>(
 ): MatchFunction<P> {
   try {
     // @ts-ignore no types exists for the pre-compiled package
-    return matchBase(str, options) as MatchFunction<P>;
+    return matchBase(str, options);
   } catch (e: any) {
     throw new Error(
       `Invalid path and options: Consult the documentation of path-to-regexp here: https://github.com/pillarjs/path-to-regexp/tree/6.x\n${e.message}`,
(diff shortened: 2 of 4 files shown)


packages/shared/src/pathToRegexp.ts
```diff
@@ -1,12 +1,110 @@
...
+    throw new Error(
+      `Invalid path and options: Consult the documentation of path-to-regexp here: https://github.com/pillarjs/path-to-regexp/tree/6.x\n${e.message}`,
+    );
+  }
+}
+
+/**
+ * A match is either `false` (no match) or a match result.
+ */
+export type Match<P extends object = object> = false | MatchResult<P>;
```
izaaklauer, line 31: I pulled in the input and output types for match, and all the types they depend on.

.changeset/big-elephants-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/shared": patch
+---
+
+Expose match from `path-to-regexp`.
```
LauraBeatris, line 5: Let's also highlight that this is an internal change without user-facing impact. Suggestion, feel free to tweak it:

```suggestion
Exports `match` utility from the `path-to-regexp` lib.
```

packages/shared/src/pathToRegexp.ts
```diff
@@ -1,12 +1,110 @@
...
+
+/**
+ * The match function takes a string and returns whether it matched the path.
+ */
+export type MatchFunction<P extends object = object> = (path: string) => Match<P>;
```
LauraBeatris, line 36: Which types are we using on `request.ts`? Let's just export the high-level ones that we rely on for the function return type.

LauraBeatris:
On another note, I wonder if we could the types into the compiled version - in the [Next.js repo](https://github.com/vercel/next.js/blob/ac4a1c6730672bf2783414e51f64b49519ddffdb/packages/next/src/lib/try-to-parse-path.ts#L1) I see types being exported without needing to be redeclared.

LauraBeatris (approved):

packages/shared/src/pathToRegexp.ts
izaaklauer: 👍  https://github.com/clerk/javascript/pull/4187/commits/deb5b9de697cfa9bd7e790af9e603e5ae1e88748

izaaklauer:
> On another note, I wonder if we could import the types into the compiled version - in the [Next.js repo](https://github.com/vercel/next.js/blob/ac4a1c6730672bf2783414e51f64b49519ddffdb/packages/next/src/lib/try-to-parse-path.ts#L1) I see types being exported without needing to be redeclared.

Good idea - I found the `--dts-resolve` flag on tsup, which generated the type declarations that we need, then I copied over those types.

deb5b9de697cfa9bd7e790af9e603e5ae1e88748

packages/shared/src/pathToRegexp.ts
```diff
@@ -1,12 +1,36 @@
...
 
 export const pathToRegexp = (path: string) => {
   try {
     // @ts-ignore no types exists for the pre-compiled package
-    return pathToRegexpBase(path) as RegExp;
+    return pathToRegexpBase(path);
   } catch (e: any) {
     throw new Error(
-      `Invalid path: ${path}.\nConsult the documentation of path-to-regexp here: https://github.com/pillarjs/path-to-regexp\n${e.message}`,
+      `Invalid path: ${path}.\nConsult the documentation of path-to-regexp here: https://github.com/pillarjs/path-to-regexp/tree/6.x\n${e.message}`,
```
izaaklauer, line 17: Calling this out: I think it's more useful to point people toward the docs for the specific version that we've embedded, because the syntax on latest is different.

packages/shared/src/pathToRegexp.ts
```diff
@@ -1,12 +1,36 @@
...
+  Path,
+  RegexpToFunctionOptions,
+  TokensToRegexpOptions,
+} from './compiled/path-to-regexp';
+import { match as matchBase, pathToRegexp as pathToRegexpBase } from './compiled/path-to-regexp';
 
 export const pathToRegexp = (path: string) => {
   try {
     // @ts-ignore no types exists for the pre-compiled package
     return pathToRegexpBase(path) as RegExp;
```
izaaklauer, line 14: Interesting: the pre-commit hooks (likely `npx eslint --fix`) removes the `as RegExp`. I had to skip the pre-commit hooks to avoid that diff. Is this necessary?

izaaklauer: Ah I see now - the type declaration is in `index.d.ts` now, so we don't need it here! Fixed.

061c5557bfd8f65efb08b2c5554194fabb841091

packages/shared/src/pathToRegexp.ts
nikosdouvlis: Good point :check

nikosdouvlis (approved):
LGTM :)

PR #4114, opened 2024-09-05
feat(clerk-js): Render `@clerk/ui`

BRKalow:
## Checklist

- [ ] `npm test` runs as expected.
- [ ] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #4199, opened 2024-09-20
feat(clerk-js): Add bundle-check script

dstaley:
## Description

This PR adds a new script, `bundle-check.mjs`, to `clerk-js`. This allows us to confirm which chunks are dynamically loaded during a particular invocation of Clerk, and to see their gzipped sizes. This will be useful when comparing the total loaded JavaScript sizes of the new versus the old components when they're both in the `clerk-js` package.

Example output:
```
➜  clerk-js CLERK_PUBLISHABLE_KEY=pk_test_... node bundle-check.mjs

/sign-in
┌────────────────────────────┬─────────┐
│ (index)                    │ size    │
├────────────────────────────┼─────────┤
│ clerk.browser.js           │ '68KB'  │
│ framework_79b6b2_5.22.3.js │ '43KB'  │
│ vendors_79b6b2_5.22.3.js   │ '27KB'  │
│ ui-common_79b6b2_5.22.3.js │ '100KB' │
│ 215_79b6b2_5.22.3.js       │ '2.8KB' │
│ signin_79b6b2_5.22.3.js    │ '9.4KB' │
│ 7_79b6b2_5.22.3.js         │ '925B'  │
│ (total)                    │ '252KB' │
└────────────────────────────┴─────────┘

/sign-up
┌────────────────────────────┬─────────┐
│ (index)                    │ size    │
├────────────────────────────┼─────────┤
│ clerk.browser.js           │ '68KB'  │
│ framework_79b6b2_5.22.3.js │ '43KB'  │
│ vendors_79b6b2_5.22.3.js   │ '27KB'  │
│ ui-common_79b6b2_5.22.3.js │ '100KB' │
│ signup_79b6b2_5.22.3.js    │ '6.2KB' │
│ 7_79b6b2_5.22.3.js         │ '925B'  │
│ (total)                    │ '246KB' │
└────────────────────────────┴─────────┘
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/bundle-check.mjs
```diff
@@ -0,0 +1,238 @@
...
+        </script>
+        <script type="text/javascript">${script}</script>
+    </body>
+</html>
+`;
+}
+
+function signIn() {
+  const script = `
+window.Clerk.load({ router: window.VIRTUAL_ROUTER }).then(() => {
```
dstaley, line 94: This `router` option is not actually used in this version of `clerk-js`. It's being introduced in #4114 but it doesn't hurt anything to have it here now. Happy to remove it if it gives anyone heartburn!

packages/clerk-js/bundle-check.mjs
```diff
@@ -0,0 +1,238 @@
...
+  .listen(4000);
+
+const byteFormatter = Intl.NumberFormat('en', {
+  notation: 'compact',
+  style: 'unit',
+  unit: 'byte',
+  unitDisplay: 'narrow',
+});
+/**
+ * Format bytes into kilobytes
```
LauraBeatris, line 154: `Intl.NumberFormat` with `notation: 'compact'` and `unit: 'byte'` choose the most appropriate unit and scale the number accordingly, so not necessarily kilobytes:


```js
const formatter = Intl.NumberFormat("en", {
  notation: "compact",
  style: "unit",
  unit: "byte",
  unitDisplay: "narrow",
});

console.log(formatter.format(10)); // 10B
console.log(formatter.format(200000)); // 200KB
console.log(formatter.format(50000000)); // 50MB
```

```suggestion
 * Format bytes into a human-readable string with appropriate units
```

dstaley: ooo great catch!

packages/clerk-js/bundle-check.mjs
```diff
@@ -0,0 +1,238 @@
...
+      if (err) {
+        res.end();
+        console.error(err);
+      }
+    };
+
+    if (req.url && req.url in routes) {
+      res.writeHead(200, { 'content-type': 'text/html' });
+      res.end(routes[req.url]);
+    } else {
```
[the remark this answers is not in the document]
dstaley: Yeah this is actually because the TypeScript types are not great. `req.url` is _always_ defined for incoming messages received by the HTTP server. But I agree that there's no harm in still checking! I'll add that.

packages/clerk-js/bundle-check.mjs
```diff
@@ -0,0 +1,238 @@
...
+function template(script) {
+  return `
+<!DOCTYPE html>
+<html>
+    <body>
+        <div id="app"></div>
+        <script
+          type="text/javascript"
+          src="/clerk.browser.js"
+          data-clerk-publishable-key="${process.env.CLERK_PUBLISHABLE_KEY}"
```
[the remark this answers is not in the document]
dstaley: This script is purely for our own use, and doesn't come into play for any runtime behaviors. We'll be manually running it like `node bundle-check.mjs`. I honestly want to just embed a key into the file, but we've never done that (and the few that are embedded point to `lclclerk.dev` domains, so the AIO components won't render with them). I don't know of a reason why embedding a publishable key would be bad, but I wanted to get confirmation before I did that. I would strongly prefer we not have to set an environment variable to run the script.

panteliselef (approved):

PR #4201, opened 2024-09-20
feat(express): Support Express v5

wobsoriano:
## Description

This PR sets a minimum Express 4 version and support for Express 5.

For Express `^4.17.0` - we went with this version as it's a more stable version, to avoid issues with earlier 4.x.x releases. We didn't want to start all the way back at 4.0.0 and risk running into old bugs or outdated features. 

Here's the total downloads of `4.17.x`:

Screenshot 2024-09-20 at 8 56 36 AM

For v5, there is not really a big breaking change and it aims to support backward compatibility with Express 4. Here's a [small breaking note](https://github.com/expressjs/express/blob/5.x/History.md#500--2024-09-10) that should not affect our SDK. Full changelog can be found [here](https://github.com/expressjs/express/releases/tag/v5.0.0) if interested.

Note that Express v5 is still tagged as `next` and not `latest` even though it's stable. But we want to add support to it now as we don't want to release another major change in the future.

I tested v5 locally with our current CI test and it passed.

_This is part of making our Express SDK ready._

Resolves ECO-192

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/express/package.json
+++ b/packages/express/package.json
@@ -68,6 +68,9 @@
     "tsup": "*",
     "typescript": "*"
   },
+  "peerDependencies": {
+    "express": "^4.17.0 || ^5.0.0"
+  },
   "engines": {
     "node": ">=18.17.0"
   },


packages/express/package.json
```diff
@@ -68,6 +68,9 @@
     "tsup": "*",
     "typescript": "*"
   },
+  "peerDependencies": {
+    "express": "^4.17.0 || ^5.0.0"
```
wobsoriano, line 72: we support v5 out-of-the-box actually but we want to set a more stable minimum v4 version to prevent any breaking change in earlier versions

.changeset/three-falcons-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/express": patch
```
panteliselef, line 2: I think minor is more appropriate
```suggestion
"@clerk/express": minor
```

packages/express/package.json
panteliselef: `@clerk/express` is still  v0. Are we doing v1 after this PR, or we wait for Express 5 to be cut as latest and then do v1 ?

nikosdouvlis: Lets do v1 now unless we know that express v5 is coming out *soon*

nikosdouvlis (approved):
✅

packages/express/package.json
wobsoriano: For reference, here's the pending items before they tag v5 as latest:

Screenshot 2024-09-23 at 7 52 45 AM

Mostly docs update, so I think `latest` tag will come soon. Let's go v1 major?

.changeset/three-falcons-glow.md
wobsoriano: Updated to `major` since Express v5 tagged as `latest` will come soon anyway and that we can start deprecating `@clerk/clerk-node-sdk` too. What y'all think?

cc @nikosdouvlis

tmilewski (approved):

PR #4210, opened 2024-09-23
feat(clerk-js,types): Supports default role on `OrganizationProfile` invitations

LauraBeatris:
## Description

Resolves ORGS-211

Prepares `OrganizationProfile` for the upcoming `default_role` changes. When inviting a member, the `default_role` will be automatically selected, otherwise it falls back to the only available role. 

Currently behind a feature flag in the Dashboard, we're now allowing updating `default_role` regardless of whether verified domains are enabled. 

---

https://github.com/user-attachments/assets/4750a26a-105c-4b10-9586-b2cb952bbe48

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
+++ b/packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
@@ -4,6 +4,7 @@ import type { ClerkAPIError } from '@clerk/types';
 import type { FormEvent } from 'react';
 import { useState } from 'react';
 
+import { useEnvironment } from '../../contexts';
 import { Flex } from '../../customizables';
 import { Form, FormButtonContainer, TagInput, useCardState } from '../../elements';
 import { useFetchRoles } from '../../hooks/useFetchRoles';
@@ -182,8 +183,20 @@ export const InviteMembersForm = (props: InviteMembersFormProps) => {
   );
 };
 
+/**
+ * Determines default role from the organization settings or fallsback to
+ * the only available role.
+ */
+const useDefaultRole = () => {
+  const { options } = useFetchRoles();
+  const { organizationSettings } = useEnvironment();
+
+  return organizationSettings.domains.default_role ?? options?.[0]?.value ?? undefined;
+};
+
 const AsyncRoleSelect = (field: ReturnType<typeof useFormControl<'role'>>) => {
   const { options, isLoading } = useFetchRoles();
+  const defaultRole = useDefaultRole();
 
   const { t } = useLocalizations();
 
@@ -195,6 +208,7 @@ const AsyncRoleSelect = (field: ReturnType<typeof useFormControl<'role'>>) => {
       >
         <RoleSelect
           {...field.props}
+          value={field.props.value ?? defaultRole}
           roles={options}
           isDisabled={isLoading}
           onChange={value => field.setValue(value)}
--- a/packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
+++ b/packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
@@ -42,9 +42,7 @@ export const InviteMembersForm = (props: InviteMembersFormProps) => {
     label: localizationKeys('formFieldLabel__emailAddresses'),
   });
 
-  const defaultRole = useDefaultRole();
-
-  const roleField = useFormControl('role', defaultRole ?? '', {
+  const roleField = useFormControl('role', '', {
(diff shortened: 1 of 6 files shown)


packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
```diff
@@ -182,8 +183,26 @@ export const InviteMembersForm = (props: InviteMembersFormProps) => {
...
+ * Determines default role from the organization settings or fallback to
+ * the only available role.
+ */
+const useDefaultRole = () => {
+  const { options } = useFetchRoles();
+  const { organizationSettings } = useEnvironment();
+
+  let defaultRole = organizationSettings.domains.defaultRole;
+
+  if (!defaultRole && options?.length === 1) {
```
panteliselef, line 196: Shouldn't this be >1 ?

"If default role doesn't exist and fetched roles exist, use the first fetched role"

LauraBeatris:
Totally forgot that but I need to update some unit tests to handle the default role, will dive into it now.

packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
```diff
@@ -195,6 +214,7 @@ const AsyncRoleSelect = (field: ReturnType<typeof useFormControl<'role'>>) => {
       >
         <RoleSelect
           {...field.props}
+          value={field.props.value || (defaultRole ?? '')}
```
panteliselef, line 217: I think think default role should simply be the initial value, when we call useFormControl

panteliselef (approved):
Left a few comments. If you feel they are not significant feel free to merge.

packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
LauraBeatris: Updated on https://github.com/clerk/javascript/pull/4210/commits/8b634a70edec0ad7f76a03e97b2ce45af42472d1

packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
LauraBeatris: We're checking if there's only one role returned from FAPI here, instead of selecting the first one. This avoids making an assumption on behalf of the user, as there could be cases where they wouldn't want that one the first one as default.

packages/clerk-js/src/ui/components/OrganizationProfile/__tests__/OrganizationProfile.test.tsx
```diff
@@ -72,4 +72,6 @@ describe('OrganizationProfile', () => {
     expect(queryByText('Members')).not.toBeInTheDocument();
     expect(queryByText('General')).toBeInTheDocument();
   });
+
+  it.todo("automatically select the organization's default role");
```
LauraBeatris, line 76: FYI I'm leaving this here, I'll come back to it today before merging.

mzhong9723 (approved):

packages/clerk-js/src/ui/components/OrganizationProfile/__tests__/InviteMembersPage.test.tsx
```diff
@@ -65,6 +202,17 @@ describe('InviteMembersPage', () => {
...
             updatedAt: new Date(),
           },
+          {
+            pathRoot: '',
+            reload: jest.fn(),
+            id: 'admin',
+            key: 'admin',
+            name: 'Admin',
+            description: '',
+            permissions: [],
```
LauraBeatris, line 212: Those tests were mocking with a single role, therefore initiating the field with that one instead and the assertion for "Select role" started to fail.

packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
LauraBeatris: @panteliselef I've actually reverted this previous commit initializing with `useFormControl`. 

I only noticed that the unit tests fail because of that, but the field value doesn't get updated when the roles get fetched. 

The default role gets resolved as undefined on a couple of renders until it gets resolved, but `useFormControl` doesn't seem to update its internal state. 

I've added the `value` prop back and it's re-rendering `RoleSelect` when that changes.

PR #4234, opened 2024-09-26
chore(express): Refactor `requireAuth` and `clerkMiddleware` middlewares

wobsoriano:
## Description

This PR refactors both `clerkMiddleware` and `requireAuth` Express middlewares to simplify DX and reduce confusion when choosing between middlewares:

1. `clerkMiddleware`

Removed option to pass a handler. It does not do anything at all, which can be replaced by a one-liner from userland:

```ts
// Before
app.use(clerkMiddleware((req, res, next) => {
  // req.auth here is undefined so it's not useful at all
}));

// After
app.use(clerkMiddleware());
app.use((req, res, next) => {
  // do stuff
})
```

2. `requireAuth`

Is now using the same logic as `clerkMiddleware` (use `authenticateRequest` and attach `auth` to `request`), except it will redirect to `signInUrl` if unauthenticated instead of calling `next(new Error())`. So you can have `requireAuth` middleware without `clerkMiddleware`:

```ts
// Before - sends error to error middleware
app.use(clerkMiddleware());
app.get('/protected', requireAuth, (req, res) => {
  res.send('This is a protected route');
});

app.use((err, req, res, next) => {
  if (err instanceof UnauthorizedError) {
    res.status(401).send('Unauthorized');
  } else {
    next(err);
  }
});


// After - redirects to signInUrl

// Apply centralized middleware
app.use(requireAuth());

// Apply middleware to a specific route
app.get('/protected', requireAuth({ signInUrl: '/sign-in }), (req, res) => {
  res.send('This is a protected route');
});
```

Other removed exports:

- `UnauthorizedError` (not used anywhere)
- `ForbiddenError` (not used anywhere)

Documentation for this one is still being worked on, we didn't announce the SDK formally, so I think `minor` update is good. The documentation will have migration guide from `@clerk/clerk-sdk-node` to `@clerk/express`. anyway.

Resolves ECO-201

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/express/src/__tests__/clerkMiddleware.test.ts
+++ b/packages/express/src/__tests__/clerkMiddleware.test.ts
@@ -83,35 +83,6 @@ describe('clerkMiddleware', () => {
     assertSignedOutDebugHeaders(response);
   });
 
-  it('supports usage with request handler: app.use(clerkMiddleware(requestHandler))', async () => {
-    const handler: RequestHandler = (_req, res, next) => {
-      res.setHeader('x-clerk-auth-custom', 'custom-value');
-      return next();
-    };
-
-    const response = await runMiddleware(clerkMiddleware(handler, { enableHandshake: true }), {
-      Cookie: '__clerk_db_jwt=deadbeef;',
-    }).expect(200, 'Hello world!');
-
-    expect(response.header).toHaveProperty('x-clerk-auth-custom', 'custom-value');
-    assertSignedOutDebugHeaders(response);
-  });
-
-  it('supports usage with parameters and request handler: app.use(clerkMiddleware(requestHandler, options))', async () => {
-    const handler: RequestHandler = (_req, res, next) => {
-      res.setHeader('x-clerk-auth-custom', 'custom-value');
-      return next();
-    };
-    const options = { publishableKey: 'pk_test_Y2xlcmsuZXhhbXBsZS5jb20k', enableHandshake: true };
-
-    const response = await runMiddleware(clerkMiddleware(handler, options), {
-      Cookie: '__clerk_db_jwt=deadbeef;',
-    }).expect(200, 'Hello world!');
-
-    expect(response.header).toHaveProperty('x-clerk-auth-custom', 'custom-value');
-    assertSignedOutDebugHeaders(response);
-  });
-
   it('throws error if clerkMiddleware is not executed before requireAuth', async () => {
     const customMiddleware: RequestHandler = (_request, response, next) => {
       response.setHeader('x-custom-middleware', 'custom');
@@ -165,7 +136,7 @@ describe('clerkMiddleware', () => {
     const res = {} as Response;
     const mockNext = jest.fn();
 
-    clerkMiddleware()[0](req, res, mockNext);
+    clerkMiddleware()(req, res, mockNext);
 
     expect(mockNext.mock.calls[0][0].message).toBe('Invalid URL');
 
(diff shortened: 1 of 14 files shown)


packages/express/src/__tests__/clerkMiddleware.test.ts
```diff
@@ -83,47 +82,6 @@ describe('clerkMiddleware', () => {
...
-    const customMiddleware: RequestHandler = (_request, response, next) => {
-      response.setHeader('x-custom-middleware', 'custom');
-      return next();
-    };
-
-    const response = await runMiddleware([requireAuth, customMiddleware]).expect(500);
-
-    assertNoDebugHeaders(response);
-    expect(response.header).not.toHaveProperty('x-clerk-auth-custom', 'custom-value');
-  });
```
wobsoriano, line 125: Removed this tests as `requireAuth` does not rely on `clerkMiddleware` anymore

wobsoriano, line 125: Removed this tests as `requireAuth` does not rely on `clerkMiddleware` anymore

wobsoriano, line 125: Removed this tests as `requireAuth` does not rely on `clerkMiddleware` anymore

packages/express/src/__tests__/clerkMiddleware.test.ts
```diff
@@ -83,47 +82,6 @@ describe('clerkMiddleware', () => {
...
-    };
-    const options = { publishableKey: 'pk_test_Y2xlcmsuZXhhbXBsZS5jb20k', enableHandshake: true };
-
-    const response = await runMiddleware(clerkMiddleware(handler, options), {
-      Cookie: '__clerk_db_jwt=deadbeef;',
-    }).expect(200, 'Hello world!');
-
-    expect(response.header).toHaveProperty('x-clerk-auth-custom', 'custom-value');
-    assertSignedOutDebugHeaders(response);
-  });
```
wobsoriano, line 113: Removed as we removed option to pass a handler in `clerkMiddleware`

packages/express/src/authenticateRequest.ts
```diff
@@ -95,3 +96,38 @@ const absoluteProxyUrl = (relativeOrAbsoluteUrl: string, baseUrl: string): strin
   }
   return new URL(relativeOrAbsoluteUrl, baseUrl).toString();
 };
+
+export const authenticateAndDecorateRequest = (options: ClerkMiddlewareOptions = {}) => {
+  const clerkClient = options.clerkClient || defaultClerkClient;
+  const enableHandshake = options.enableHandshake || false;
+
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  const middleware: RequestHandler = async (request, response, next) => {
```
jescalan, line 105: Have we covered for the scenario where someone uses both middlewares? A little conditional at the top of this helper to just call next if `auth` is already on the request I think would take care of that. May be worth adding a test for this as well.

jescalan (commented):
This is beautiful I love it. Such a quick turnaround!

packages/express/src/authenticateRequest.ts
wobsoriano: Nope, but added just now and a test too!

jescalan: Do we need to error if there are both? Could we just add a conditional check for `req.auth` at the top of the function instead?

wobsoriano: Ah I get you now, we want to allow usage of both, but check for `req.auth` first then call `next()` if exists so we dont have to run the same logic etc 👍🏼 Perfect for scenario like

```ts
app.use(clerkMiddleware());

app.get('/protected', requireAuth(), (req, res) => {
  res.send('This is a protected route');
});
```

LauraBeatris:
I love those changes, the DX is much better now ❤️

jescalan (approved):
This looks great to me! Nice work Rob 👏

PR #4243, opened 2024-09-30
chore(clerk-js): Allow single-character usernames in `<UserProfile />` validation

nikospapcom:
## Description

In this pr we are allowing single-character usernames in `<UserProfile />` validation

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/UserProfile/UsernameForm.tsx
+++ b/packages/clerk-js/src/ui/components/UserProfile/UsernameForm.tsx
@@ -27,7 +27,7 @@ export const UsernameForm = withCardStateProvider((props: UsernameFormProps) => 
   const isUsernameRequired = userSettings.attributes.username.required;
 
   const canSubmit =
-    (isUsernameRequired ? usernameField.value.length > 1 : true) && user.username !== usernameField.value;
+    (isUsernameRequired ? usernameField.value.length > 0 : true) && user.username !== usernameField.value;
 
   const updatePassword = async () => {
     try {
--- a/.changeset/cold-mayflies-happen.md
+++ b/.changeset/cold-mayflies-happen.md
@@ -1,0 +1,5 @@
+---
+"@clerk/clerk-js": patch
+---
+
+Allow single-character usernames in `<UserProfile />` validation


packages/clerk-js/src/ui/components/UserProfile/UsernameForm.tsx
```diff
@@ -26,8 +26,7 @@ export const UsernameForm = withCardStateProvider((props: UsernameFormProps) =>
 
   const isUsernameRequired = userSettings.attributes.username.required;
 
-  const canSubmit =
-    (isUsernameRequired ? usernameField.value.length > 1 : true) && user.username !== usernameField.value;
+  const canSubmit = user.username !== usernameField.value;
```
chanioxaris, line 29: ❓ It seems like we have the userSettings available and thus the username min and max length. Is it possible to use them on the client side validation in place of the hardcoded value?

nikospapcom: We have it but we rely only on the backend validation if we have username as missing field during sign-up. 
I removed the client side validation in order to have consistency and have the same behaviour in both cases

ss from missing fields screen

Screenshot 2024-09-30 at 12 02 48 PM

chanioxaris (approved):

packages/clerk-js/src/ui/components/UserProfile/UsernameForm.tsx
chanioxaris: Great thanks!

anagstef (approved):

packages/clerk-js/src/ui/components/UserProfile/UsernameForm.tsx
panteliselef: @nikospapcom I don't see why this change improves things.

Currently we want to disable the button, and don't let you submit if you don't have a value in the field or the value that you matches your current username.

panteliselef (changes requested):

packages/clerk-js/src/ui/components/UserProfile/UsernameForm.tsx
panteliselef: Maybe only changing `usernameField.value.length > 0` is enough ?

octoper (approved):

panteliselef (approved):

PR #4244, opened 2024-09-30
fix(backend): Handle "Cannot convert argument to a ByteString" errors

nikosdouvlis:
## Description
Headers.set can throw a "Cannot convert argument to a ByteString" when trying to set/append a string containing unicode characters as headers can only accept ASCII chars in the range of [0-255]. Under certain cases, verifyToken can throw an error containing the unicode character that caused the error which then we would try to include in the debug header.

This would cause the *only current request* to fail so we are adding extra error handling to prevent this edge scenario.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/tokens/authStatus.ts
+++ b/packages/backend/src/tokens/authStatus.ts
@@ -158,13 +158,27 @@ const withDebugHeaders = <T extends RequestState>(requestState: T): T => {
   const headers = new Headers(requestState.headers || {});
 
   if (requestState.message) {
-    headers.set(constants.Headers.AuthMessage, requestState.message);
+    try {
+      headers.set(constants.Headers.AuthMessage, requestState.message);
+    } catch (e) {
+      // headers.set can throw if unicode strings are passed to it. In this case, simply do nothing
+    }
   }
+
   if (requestState.reason) {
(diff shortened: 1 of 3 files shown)


packages/backend/src/tokens/authStatus.ts
```diff
@@ -158,13 +158,27 @@ const withDebugHeaders = <T extends RequestState>(requestState: T): T => {
   const headers = new Headers(requestState.headers || {});
 
   if (requestState.message) {
-    headers.set(constants.Headers.AuthMessage, requestState.message);
+    try {
+      headers.set(constants.Headers.AuthMessage, requestState.message);
+    } catch (e) {
+      // headers.set can throw if unicode strings are passed to it. In this case, simply do nothing
```
anagstef, line 164: @nikosdouvlis shouldn't we fall back to a specific value? or we don't really care about it?

nikosdouvlis: It's a very rare case that would otherwise cause the current request to fail - I don't think we need additional information in this case as the only way to reach this state is by using a forged/invalid token in the first place

anagstef (approved):

PR #4250, opened 2024-09-30
feat(clerk-js): Enable persisted clients by default

panteliselef:
## Description
We would like to keep around the `persistClient` flag under the `experimental` settings for a bit longer as an escape hatch in case folks encounter any weird behaviour.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/shy-sloths-laugh.md
+++ b/.changeset/shy-sloths-laugh.md
@@ -2,4 +2,15 @@
 "@clerk/clerk-js": minor
 ---
 
-Enable persisted clients by default. To disable this setting pass `persistClient: false` to `experimental` in ClerkProvider.
+We recently shipped an experimental feature to persist the Clerk client (under `persistClient` flag) as an opt-in. This allows for matching a user's device with a client. We want to test this behavior with more users, so we're making it opt-out as the next step. After more successful testing we'll remove the experimental flag and enable it by default.
+
+If you're encountering issues, please open an issue. You can disable this new behavior like so:
+
+```js
+// React
+<ClerkProvider experimental={{ persistClient: false }} />
+
+// Vanilla JS
+await clerk.load({ experimental: { persistClient: false } })
+```
+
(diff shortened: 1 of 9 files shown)


.changeset/shy-sloths-laugh.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/clerk-js": minor
+---
+
+Enable persisted clients by default. To disable this setting pass `persistClient: false` to `experimental` in ClerkProvider.
```
LekoArts, line 5: Let's explain why we're doing this 👍 

````suggestion
We recently shipped an experimental feature to persist the Clerk client (under `persistClient` flag) as an opt-in. This allows for matching a user's device with a client. We want to test this behavior with more users, so we're making it opt-out as the next step. After more successful testing we'll remove the experimental flag and enable it by default.

If you're encountering issues, please open an issue. You can disable this new behavior like so:

```js
// React
<ClerkProvider experimental={{ persistClient: false }} />

// Vanilla JS
await clerk.load({ experimental: { persistClient: false } })
```

````

File: integration/presets/envs.ts [hunk 2]
```diff
@@ -35,9 +35,9 @@ const withEmailCodes = base
   .setEnvVariable('public', 'CLERK_PUBLISHABLE_KEY', instanceKeys.get('with-email-codes').pk)
   .setEnvVariable('private', 'CLERK_ENCRYPTION_KEY', constants.E2E_CLERK_ENCRYPTION_KEY || 'a-key');
 
-const withEmailCodes_persist_client = withEmailCodes
+const withOutEmailCodes_persist_client = withEmailCodes
```
LekoArts, line 38: Why the name change? You're still using the `withEmailCodes` parent for cloning

LekoArts (approved):

integration/presets/envs.ts
panteliselef: i renamed it to `withEmailCodes_destroy_client`

PR #4269, opened 2024-10-01
chore(repo): Add major version check workflow

anagstef:
## Description

This PR adds a GHA that checks if there is a `major` changeset in a PR and if there is one it expects a `!allow-major` comment to exist from an org member in order to pass.

This is to prevent accidental major bumps of our packages.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


anagstef:
!allow-major

anagstef:
!allow-major

.github/workflows/major-version-check.yml
```diff
@@ -0,0 +1,100 @@
...
+          script: |
+            const prNumber = context.payload.pull_request.number;
+            const org = context.repo.owner;
+
+            // Get all comments on the PR
+            const { data: comments } = await github.rest.issues.listComments({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: prNumber,
+            });
```
nikosdouvlis, line 69: Nit: Instead of using a comment, how about using a label here instead?

nikosdouvlis (approved):

.github/workflows/major-version-check.yml
anagstef: Thought about it, but we want an explicit action done by a Clerk member for the check to pass. If we can make sure that a label is added by a Clerk member, it could work. 🤷

PR #4277, opened 2024-10-03
feat(ui): Add `<RouterLink />` component

alexcarpenter:
## Description

Implements `<RouterLink />` component to allow navigations without full page reloads when navigating between `SignIn`/`SignUp` flows.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -29,13 +29,13 @@
     "./*": {
       "import": {
         "types": "./dist/components/*.d.mts",
-        "default": "./dist/components/*.mjs",
-        "browser": "./dist/components/*.mjs"
+        "browser": "./dist/components/*.mjs",
+        "default": "./dist/components/*.mjs"
       },
       "require": {
         "types": "./dist/components/*.d.ts",
-        "default": "./dist/components/*.js",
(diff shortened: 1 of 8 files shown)


packages/ui/package.json
```diff
@@ -29,13 +29,13 @@
...
-        "browser": "./dist/components/*.mjs"
+        "browser": "./dist/components/*.mjs",
+        "default": "./dist/components/*.mjs"
       },
       "require": {
         "types": "./dist/components/*.d.ts",
-        "default": "./dist/components/*.js",
-        "browser": "./dist/components/*.js"
+        "browser": "./dist/components/*.js",
+        "default": "./dist/components/*.js"
```
alexcarpenter, line 38: This order mattered for rendering the components within theme-builder app.

panteliselef (approved):

packages/ui/package.json
jacekradko: According to node docs, the `default` conditional export should always come last, so this is the correct order: https://nodejs.org/api/packages.html#conditional-exports

PR #4281, opened 2024-10-03
fix(clerk-js): Initialize default role with form controls

LauraBeatris:
## Description

Fixes issue where "Send invitation" button is getting disabled even tho there's a default role loaded: 

https://github.com/user-attachments/assets/fe18d474-5cf3-4396-94ec-a008db7a99cf

The issue relies on the fact that we're overriding the `field.value` property when rendering it, instead of passing it to the form controls state: 

https://github.com/user-attachments/assets/c3a6696c-2e62-40b8-8e55-bb346fb528d6

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
+++ b/packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
@@ -2,7 +2,7 @@ import { isClerkAPIResponseError } from '@clerk/shared/error';
 import { useOrganization } from '@clerk/shared/react';
 import type { ClerkAPIError } from '@clerk/types';
 import type { FormEvent } from 'react';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 
 import { useEnvironment } from '../../contexts';
 import { Flex } from '../../customizables';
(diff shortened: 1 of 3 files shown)


packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
```diff
@@ -42,10 +42,19 @@ export const InviteMembersForm = (props: InviteMembersFormProps) => {
     label: localizationKeys('formFieldLabel__emailAddresses'),
   });
 
+  const defaultRole = useDefaultRole();
   const roleField = useFormControl('role', '', {
     label: localizationKeys('formFieldLabel__role'),
   });
 
+  useEffect(() => {
+    if (roleField.value || !defaultRole) {
```
LauraBeatris, line 51: Not sure if that's the expected pattern on other places, but `defaultRole` takes time to be defined, therefore `roleField` can get initialized with an empty value. 

The way that I got to fix it was by introducing this effect here which runs once the default role gets updated, sets the value, and doesn't execute anymore once the field has a value.

izaaklauer (approved):

panteliselef (approved):

FranciscoMoretti:
Hi, thanks for fixing this! I see that `@clerk/clerk-js@5.26.1` was released with this update. 

How does this reach `@clerk/clerk-react`?

LauraBeatris:
@FranciscoMoretti It's deployed now, you will see the latest fix on your app!

FranciscoMoretti:
@LauraBeatris thank you so much for the detailed explanation and update! 💯

PR #4290, opened 2024-10-04
feat(types,ui): Add support for `__experimental.appearance` prop

dstaley:
## Description

This PR adds support for the `__experimental.appearance` prop which is distinct from the existing `appearance` prop. If the new `SignIn`/`SignUp` component is rendered with `__experimental.newComponents`, it will pull the appearance from `__experimental.appearance` instead of `appearance`. If the components are rendered directly (outside of `clerk-js` such as in Theme Builder) they will continue to read from `appearance`.

This PR also adds descriptors to the `Card.Logo` component since we want to target it in Dashboard.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/types/src/clerk.ts
+++ b/packages/types/src/clerk.ts
@@ -837,12 +837,7 @@ export type SignInProps = RoutingOptions & {
   /**
    * Enable experimental flags to gain access to new features. These flags are not guaranteed to be stable and may change drastically in between patch or minor versions.
    */
-  __experimental?: Autocomplete<
-    {
-      newComponents: boolean;
-    },
-    Record<string, any>
-  >;
+  __experimental?: Record<string, any> & { newComponents?: boolean };
 } & TransferableOption &
   SignUpForceRedirectUrl &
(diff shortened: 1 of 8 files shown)


packages/types/src/clerk.ts
```diff
@@ -837,12 +837,7 @@ export type SignInProps = RoutingOptions & {
   /**
    * Enable experimental flags to gain access to new features. These flags are not guaranteed to be stable and may change drastically in between patch or minor versions.
    */
-  __experimental?: Autocomplete<
-    {
-      newComponents: boolean;
-    },
-    Record<string, any>
-  >;
+  __experimental?: Record<string, any> & { newComponents?: boolean };
```
dstaley, line 840: This was changed because I misunderstood what the `Autocomplete` type was for. It's for autocompleting string unions, not for autocompleting arbitrary objects.

tmilewski (approved):

PR #4292, opened 2024-10-06
Update sv-SE.ts

MarcusT96:
## Description

Edited the swedish localization package to more match the en-US package that is managed by the clerk team. This to make us swedes have a more complete authentication system in swedish and not mix swedish and english as before.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


anagstef (approved):
Thank you for contributing! 🙏

packages/localizations/src/sv-SE.ts
```diff
@@ -511,51 +514,56 @@ export const svSE: LocalizationResource = {
...
     captcha_unavailable:
       'Registrering misslyckades på grund av misslyckad bot-validering. Vänligen uppdatera sidan och försök igen eller kontakta supporten för mer hjälp.',
-    form_code_incorrect: 'Koden är felaktig.',
-    form_identifier_exists: 'Identifieraren finns redan.',
-    form_identifier_exists__email_address: 'E-postadressen finns redan.',
-    form_identifier_exists__phone_number: 'Telefonnumret finns redan.',
-    form_identifier_exists__username: 'Användarnamnet finns redan.',
-    form_identifier_not_found: 'Identifieraren hittades inte.',
-    form_param_format_invalid: 'Formatet är ogiltigt.',
+    form_code_incorrect: undefined,
```
tmilewski, line 537: Thanks for your contribution! Is there any reason that some of these are now `undefined`? The current translations _seem_ to be correct.

tmilewski (changes requested):

packages/localizations/src/sv-SE.ts
MarcusT96: Ah, i was changing a lot to match the en-US one so it might have been that i made some of them undefined since they were that in the en-US.ts file. I was afraid I would break something if those were there. But if it dont those changes can be reverted and keep the translations that were correct.

packages/localizations/src/sv-SE.ts
```diff
@@ -511,51 +514,56 @@ export const svSE: LocalizationResource = {
...
+    form_identifier_exists: undefined,
+    form_identifier_exists__email_address: 'Denna e-postadress är taget. Vänligen prova ett annat.',
+    form_identifier_exists__phone_number: 'Detta telefonnummer är taget. Vänligen prova ett annat.',
+    form_identifier_exists__username: 'Detta användarnamn är taget. Vänligen prova ett annat.',
+    form_identifier_not_found: undefined,
+    form_param_format_invalid: undefined,
```
MarcusT96, line 543: form_code_incorrect can be "koden är felaktig" still and the undefined ones can revert as they were since they are correct. The ones I changed are more descriptive with "please try somehting else", so if that is fine they can be changed to the new format otherwise they can be reverted.

packages/localizations/src/sv-SE.ts
```diff
@@ -511,51 +514,56 @@ export const svSE: LocalizationResource = {
...
-    not_allowed_access: 'Åtkomst nekad.',
-    organization_domain_blocked: undefined,
-    organization_domain_common: undefined,
-    organization_membership_quota_exceeded: undefined,
-    organization_minimum_permissions_needed: undefined,
+    not_allowed_access: undefined,
+    organization_domain_blocked: 'Domänen är blockerad.',
+    organization_domain_common: 'Domänen är vanlig.',
+    organization_membership_quota_exceeded: 'Medlemskapet är fullt.',
+    organization_minimum_permissions_needed: 'Du måste ha tillräckligt med behörigheter.',
```
MarcusT96, line 566: Same here, I must have been blind or something, the ones before was correct and can be reverted if I made them undefined

MarcusT96 (commented):
Left notes on the ones were I might have gone a bit bananas, sorry!

packages/localizations/src/sv-SE.ts
MarcusT96: Fixed in latest commit

packages/localizations/src/sv-SE.ts
```diff
@@ -99,31 +100,31 @@ export const svSE: LocalizationResource = {
...
+  formFieldInputPlaceholder__lastName: 'Ange ditt efternamn',
+  formFieldInputPlaceholder__organizationDomain: 'Ange organisationsdomän',
+  formFieldInputPlaceholder__organizationDomainEmailAddress: 'Ange e-postadress för verifiering',
+  formFieldInputPlaceholder__organizationName: 'Ange organisationsnamn',
+  formFieldInputPlaceholder__organizationSlug: 'min-organisation',
+  formFieldInputPlaceholder__password: 'Ange ditt lösenord',
+  formFieldInputPlaceholder__phoneNumber: 'Ange ditt telefonnummer',
+  formFieldInputPlaceholder__username: 'Ange ditt användarnamn',
+  formFieldLabel__automaticInvitations: 'Aktivera automatiska inbjudningar för denna domän',
+  formFieldLabel__backupCode: undefined,
```
tmilewski, line 126: Let's revert this, if it makes sense.

packages/localizations/src/sv-SE.ts
```diff
@@ -511,37 +514,42 @@ export const svSE: LocalizationResource = {
...
     form_param_max_length_exceeded__first_name: 'Förnamnet får inte överskrida 256 tecken.',
     form_param_max_length_exceeded__last_name: 'Efternamnet får inte överskrida 256 tecken.',
     form_param_max_length_exceeded__name: 'Namnet får inte överskrida 256 tecken.',
-    form_param_nil: 'Parametern får inte vara tom.',
+    form_param_nil: undefined,
     form_param_value_invalid: undefined,
-    form_password_incorrect: 'Lösenordet är felaktigt.',
-    form_password_length_too_short: 'Lösenordet är för kort.',
+    form_password_incorrect: undefined,
+    form_password_length_too_short: undefined,
```
tmilewski, line 552: Let's revert these as well, if it makes sense. They seem correct.

tmilewski (changes requested):
@MarcusT96 Maybe I'm misunderstanding some of your comments, but I think it makes sense to keep some of the original translations I called out here.

Thanks again!

packages/localizations/src/sv-SE.ts
MarcusT96: I reverted 'formFieldLabel__backupCode' and kept the rest.

MarcusT96:
Fixed the reverts, we should be good to go now!

tmilewski (approved):

PR #4301, opened 2024-10-08
fix(clerk-js): Bundle CSS into CJS and ESM bundles

dstaley:
## Description

This PR updates the CJS and ESM builds of `clerk-js` to bundle imported CSS files instead of emitting a URL to the file. This is required since we don't have the capability to dynamically load anything on account of not knowing the public path (like we do in our browser bundles). This impacts consumers who directly consume the `clerk.js` and `clerk.mjs` bundles of `clerk-js`, which includes Chrome extensions (and possibly others).

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/webpack.config.js
+++ b/packages/clerk-js/webpack.config.js
@@ -170,12 +170,51 @@ const typescriptLoaderDev = () => {
   };
 };
 
-/** @type { () => (import('webpack').RuleSetRule) } */
+/**
+ * Used in outputs that utilize chunking, and returns a URL to the stylesheet.
+ * @type { () => (import('webpack').RuleSetRule) }
+ */
 const clerkUICSSLoader = () => {
-  // This emits a module exporting the URL to the styles.css file.
+  // This emits a module exporting a URL to the styles.css file.
   return {
     test: /packages\/ui\/dist\/styles\.css/,
     type: 'asset/resource',
+  };
+};
+
+/**
+ * Used in outputs that _do not_ utilize chunking, and returns the contents of the stylesheet.
+ * @type { () => (import('webpack').RuleSetRule) }
+ */
+const clerkUICSSSourceLoader = () => {
+  // This emits a module exporting the contents of the styles.css file.
+  return {
+    test: /packages\/ui\/dist\/styles\.css/,
+    type: 'asset/source',
+  };
+};
+
+/**
+ * Used for production builds that have dynamicly loaded chunks.
+ * @type { () => (import('webpack').Configuration) }
+ * */
+const commonForProdChunked = () => {
+  return {
+    module: {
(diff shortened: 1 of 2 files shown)


packages/clerk-js/webpack.config.js
```diff
@@ -170,22 +170,58 @@ const typescriptLoaderDev = () => {
   };
 };
 
-/** @type { () => (import('webpack').RuleSetRule) } */
+/**
+ * Used in outputs that utilize chunking, and returns a URL to the stylesheet.
+ * @type { () => (import('webpack').RuleSetRule) }
+ */
 const clerkUICSSLoader = () => {
```
dstaley, line 177: These could be one loader with a boolean argument for `supportsChunks`. I didn't have a strong opinion one way to the other, so happy to change this if anyone else has a stronger opinion!

packages/clerk-js/webpack.config.js
```diff
@@ -170,22 +170,58 @@ const typescriptLoaderDev = () => {
...
-  // This emits a module exporting the URL to the styles.css file.
+  // This emits a module exporting a URL to the styles.css file.
```
dstaley, line 178: These could be one loader with a boolean argument for `supportsChunks`. I didn't have a strong opinion one way to the other, so happy to change this if anyone else has a stronger opinion!

packages/clerk-js/webpack.config.js
```diff
@@ -170,22 +170,58 @@ const typescriptLoaderDev = () => {
...
       rules: [svgLoader(), typescriptLoaderProd(), clerkUICSSLoader()],
     },
+  };
+};
+
+/**
+ * Used for production builds that combine all files into one single file (such as for Chrome Extensions).
+ * @type { () => (import('webpack').Configuration) }
+ * */
+const commonForProdBundled = () => {
```
dstaley, line 213: We could also do something like `commonForProd({ bundled: true })` and do everything within the `commonForProd` function.

packages/clerk-js/webpack.config.js
```diff
@@ -283,13 +326,19 @@ const prodConfig = ({ mode }) => {
     ],
   });
 
-  const clerkCjs = merge(clerkEsm, {
+  const clerkCjs = merge(entryForVariant(variants.clerk), common({ mode }), commonForProd(), commonForProdBundled(), {
```
dstaley, line 329: I removed the `merge(clerkEsm` portion because the experimental ESM output was confusing Webpack. It was emitting references to `import.meta.url` within the CJS output, which caused ATTW to complain.

BRKalow (approved):

PR #4304, opened 2024-10-08
chore(clerk-js): Update `OrganizationSwitcher` popover action to include label

LauraBeatris:
## Description

Resolves ORGS-246

Replaces `OrganizationSwitcher` small gear icon action with a labeled button to make it less confusing to end-users.

Now:

CleanShot 2024-10-08 at 19 35 37

Previously:

CleanShot 2024-10-08 at 19 44 44

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/OrganizationSwitcher/OrganizationSwitcherPopover.tsx
+++ b/packages/clerk-js/src/ui/components/OrganizationSwitcher/OrganizationSwitcherPopover.tsx
@@ -8,7 +8,6 @@ import { useEnvironment, useOrganizationSwitcherContext } from '../../contexts';
 import { descriptors, Flex, localizationKeys } from '../../customizables';
 import {
   Actions,
-  ExtraSmallAction,
   OrganizationPreview,
   PersonalWorkspacePreview,
   PopoverCard,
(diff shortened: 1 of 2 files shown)


kylemac:
Q to consider: should we offer an option to reverse this and just go to the gear icon? (aka does this break a contract we have with our users?)

LauraBeatris:
> Q to consider: should we offer an option to reverse this and just go to the gear icon? (aka does this break a contract we have with our users?)

@kylemac I think this won't be needed, cause we're still keeping the same UI hierarchy here with a `menu` -> `menuitem` roles, but the underlying icon now is rendered within a label. 

CleanShot 2024-10-09 at 10 10 04

CleanShot 2024-10-09 at 10 09 42

Depending on how folks are relying on this for things such as UI tests, then it could break when bumping the version, but still, I wouldn't consider that valuable in order to introduce a property for this behavior 🤔

packages/clerk-js/src/ui/components/OrganizationSwitcher/OrganizationSwitcherPopover.tsx
```diff
@@ -105,20 +104,6 @@ export const OrganizationSwitcherPopover = React.forwardRef<HTMLDivElement, Orga
       });
     };
 
-    const manageOrganizationSmallIconButton = (
```
LauraBeatris, line 108: Now we're reusing the action that used to be displayed only when there was a `__unstable_manageBillingUrl`.

jacekradko (approved):

LekoArts (approved):

PR #4313, opened 2024-10-10
feat(localization): Add en-GB

ijxy:
## Description

Replaces a couple of "Americanisms" with their British versions:
- authorize with authorise
- organization with organisation

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/fuzzy-carrots-push.md
+++ b/.changeset/fuzzy-carrots-push.md
@@ -1,0 +1,5 @@
+---
+"@clerk/localizations": minor
+---
+
+Added support for en-GB localization
--- a/packages/localizations/package.json
+++ b/packages/localizations/package.json
@@ -59,6 +59,7 @@
     "da-DK",
     "de-DE",
     "el-GR",
+    "en-GB",
     "en-US",
     "es-ES",
     "fi-FI",
--- a/packages/localizations/src/en-GB.ts
+++ b/packages/localizations/src/en-GB.ts
@@ -1,54 +1,921 @@
-/*
- * =====================================================================================
- * DISCLAIMER:
- * =====================================================================================
- * This localization file is a community contribution and is not officially maintained
- * by Clerk. It has been provided by the community and may not be fully aligned
- * with the current or future states of the main application. Clerk does not guarantee
- * the accuracy, completeness, or timeliness of the translations in this file.
- * Use of this file is at your own risk and discretion.
- * =====================================================================================
- */
-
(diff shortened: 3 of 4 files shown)


.changeset/fuzzy-carrots-push.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/localizations": major
```
tmilewski, line 2:
```suggestion
"@clerk/localizations": minor
```

tmilewski (changes requested):
@ijxy Thank you for this! Please just update the major changeset to minor, and we should be good to go!

.changeset/fuzzy-carrots-push.md
ijxy: Happy to continue as is, but it has occurred to me that _most_ of this file is identical to `en-US` and it would be a lot easier to maintain if only the changed values were present in this file.

As is, the `en-GB` locale would not automatically benefit from changes to the `en-US` values on the keys where they should be the same.

What do you think? Happy to strip out the identical fields if you agree.

tmilewski: @ijxy That’s a good point! Yeah, let’s strip out the duplicates and we’ll merge from there.

Thanks again!

packages/localizations/src/utils/generate.ts
```diff
@@ -164,6 +165,7 @@ run(beBY, 'be-BY');
 run(csCZ, 'cs-CZ');
 run(daDK, 'da-DK');
 run(deDE, 'de-DE');
+run(enGB, 'en-GB');
```
anagstef, line 168: Auto-generation will not work as expected if the file has custom logic inside it. 

Let's drop the changes of `generate.ts` file.

anagstef (changes requested):
Let's drop the changes of `generate.ts` file.

ijxy:
@anagstef @tmilewski Apologies for the delay–I have updated as per comments and extracted the changed words into a little dictionary object so it is easier to see what is actually changed.

The tl;dr is that now the enGB locale data should automatically stay up to date with any changes to enUS. 🎉

tmilewski (approved):
Hey @ijxy! We had a discussion internally and thought it'd be better to keep things consistent as well as not increase run-time work. As-such, I've applied the values in your work to the existing `en-US` file.

Thanks for your contribution!

PR #4318, opened 2024-10-11
fix(elements): Progressive sign up ticket flow

BRKalow:
## Description

Fixed a bug in the sign up flow where the sign up form was unable to accept additional requirements when initiated from a ticket.

fixes SDKI-697

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/elements/src/internals/machines/form/form.machine.ts
+++ b/packages/elements/src/internals/machines/form/form.machine.ts
@@ -169,7 +169,7 @@ export const FormMachine = setup({
 
           if (field) {
             field.checked = event.field.checked;
-            field.disabled = event.field.disabled || false;
+            field.disabled = event.field.disabled ?? field.disabled;
             field.value = event.field.value;
 
             context.fields.set(event.field.name, field);
--- a/packages/elements/src/internals/machines/sign-up/router.machine.ts
+++ b/packages/elements/src/internals/machines/sign-up/router.machine.ts
@@ -16,6 +16,7 @@ import { ClerkElementsError, ClerkElementsRuntimeError } from '~/internals/error
 import { ThirdPartyMachine, ThirdPartyMachineId } from '~/internals/machines/third-party';
 import { shouldUseVirtualRouting } from '~/internals/machines/utils/next';
 
+import { SignInStartMachineId } from '../sign-in';
 import { SignUpContinueMachine } from './continue.machine';
 import type {
   SignUpRouterContext,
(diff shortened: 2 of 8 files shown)


packages/elements/src/internals/machines/sign-up/router.machine.ts
```diff
@@ -343,8 +344,10 @@ export const SignUpRouterMachine = setup({
       },
       on: {
         'RESET.STEP': {
-          target: 'Start',
-          reenter: true,
```
BRKalow, line 347: This was causing the child machine to be re-created, and subsequently it would move right back to `Attempting` because of the presence of the ticket. For `RESET.STEP` specifically, we want to update the `formRef` of the child machine as this even is fired when the form machine is re-created.

packages/elements/src/internals/machines/sign-up/start.machine.ts
```diff
@@ -105,11 +115,20 @@ export const SignUpStartMachine = setup({
...
+  },
   states: {
     Init: {
       description:
         'Handle ticket, if present; Else, default to Pending state. Per tickets, `Attempting` makes a `signUp.create` request allowing for an incomplete sign up to contain progressively filled fields on the Start step.',
       always: [
+        {
+          guard: and(['hasTicket', 'isMissingRequirements']),
+          target: 'Pending',
+        },
```
BRKalow, line 131: When the sign up attempt is missing requirements, users need to be able to submit the form, so we move to `Pending`

packages/elements/src/internals/machines/sign-up/router.machine.ts
BRKalow: There might be a better spot to put this that makes the intent more explicit. The underlying attempt actor was getting canceled, and subsequently re-created, by the reinitialization of the start machine. Changing this to only updating the form ref (which is what triggers `RESET.STEP`) solved the re-initialization issue.

packages/elements/src/internals/machines/form/form.machine.ts
```diff
@@ -169,7 +169,7 @@ export const FormMachine = setup({
 
           if (field) {
             field.checked = event.field.checked;
-            field.disabled = event.field.disabled || false;
+            field.disabled = event.field.disabled ?? field.disabled;
```
BRKalow, line 172: I'm not sure if this matters in practice (field shouldn't get updated if disabled), but if this does happen, we want to retain the existing disabled value.

tmilewski (approved):

dstaley (approved):

PR #4320, opened 2024-10-11
chore(backend): Organization takes precedence over personal account in organization sync options

izaaklauer:
## What changed?

This change makes organization URL patterns take precedence over personal account patterns in the new OrganizationSync handshake options.

## Why now?

Initially, we didn't have a strong argument for which option should have precedence. Now, a use-case has become clear! If a fixed prefix is going to be present in a URL, we think it's more likely that it will be in org-indicating URLs and not personal-account-indicating URLs.

E.g. the following pattern:
`/orgs/myorg/settings` <- Fixed prefix "orgs" implies "mygreatorg" is active
`/settings` <- Absence of prefix implies that the personal account is active

Will occur more often than:
`/me/settings` <- Fixed prefix "me" indicates that the personal account is active
`/mygreatorg/settings` <- Absence of fixed prefix implies an organization is active

This is relevant because it's easiest to have the activatible entity type (org or personal account) with a fixed prefix take presedence, so that the other can wildcard match all routes.

For example, for an app using the first pattern, they could use the following middleware options:
```js
{
  personalAccountPatterns: [
    '/', 
    '/(.*)'
  ],
  organizationPatterns: [
    '/orgs/:slug', 
    '/orgs/:slug/(.*)'
  ]
}
```

E.g. letting the personal account match everything that does not have the org prefix.

## Is this a breaking change?

Technically yes, but this feature was introduced early this week, and has not yet been documented or formally announced, so I don't expect it to be in use yet, much less for anyone to be depending on this nuance yet.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:


LauraBeatris:
@izaaklauer I think this is more like a bug fix than an chore, cause if there was a case like the following: 

```
{
  personalAccountPatterns: [
    '/', 
    '/(.*)'
  ],
  organizationPatterns: [
    '/orgs/:slug', 
    '/orgs/:slug/(.*)'
  ]
}
```

then it wouldn't apply the organization patterns even if the route was matching correctly.

.changeset/stale-pans-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/backend": patch
```
panteliselef, line 2: I'd do minor as the change is significant but necessary.

panteliselef (approved):

izaaklauer:
> @izaaklauer I think this is more like a bug fix than an chore, cause if there was a case like the following:
> 
> ```
> {
>   personalAccountPatterns: [
>     '/', 
>     '/(.*)'
>   ],
>   organizationPatterns: [
>     '/orgs/:slug', 
>     '/orgs/:slug/(.*)'
>   ]
> }
> ```
> 
> then it wouldn't apply the organization patterns even if the route was matching correctly.

You're right that those settings today will always activate the personal workspace, but [today's jsdoc comments](https://github.com/clerk/javascript/pull/4320/files#diff-64e006762089509c48eb751fd14dfaae98bbdcb5a00a40b82ee0760b06f47bddR39) also specify that the personal workspace takes prescience! This change only switches which pattern matcher is applied first. It used to be personal workspace, but now it's organization, because organization is more likely to be prefixed.

PR #4329, opened 2024-10-14
fix(clerk-js): Only retry oauth if captcha failed

nikosdouvlis:
## Description
We only want to reload env and retry if the error thrown was related to captcha falling.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/nasty-melons-cross.md
+++ b/.changeset/nasty-melons-cross.md
@@ -1,0 +1,6 @@
+---
+"@clerk/clerk-js": patch
+"@clerk/shared": patch
+---
+
+Only retry the OAuth flow if the captcha check failed.
--- a/packages/shared/src/error.ts
+++ b/packages/shared/src/error.ts
@@ -6,6 +6,10 @@ export function isUnauthorizedError(e: any): boolean {
   return code === 'authentication_invalid' && status === 401;
 }
 
+export function isCaptchaError(e: ClerkAPIResponseError): boolean {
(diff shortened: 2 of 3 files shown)


BRKalow (approved):

.changeset/nasty-melons-cross.md
```diff
@@ -0,0 +1,6 @@
+---
+"@clerk/clerk-js": patch
+"@clerk/shared": patch
+---
+
+Only retry the OAuth flow if the captcha check failed.
```
anagstef, line 6:
```suggestion
Retry the OAuth flow only if the captcha check failed.
```

packages/shared/src/error.ts
```diff
@@ -6,6 +6,10 @@ export function isUnauthorizedError(e: any): boolean {
   return code === 'authentication_invalid' && status === 401;
 }
 
+export function isCaptchaError(e: ClerkAPIResponseError): boolean {
+  return ['captcha_invalid', 'captcha_not_enabled', 'captcha_missing_token'].includes(e.errors[0].code);
```
anagstef, line 10:
```suggestion
  return ['captcha_invalid', 'captcha_not_enabled', 'captcha_missing_token'].includes(e?.errors?.[0]?.code);
```

anagstef (approved):

packages/shared/src/error.ts
nikosdouvlis: Not needed as `e`is `ClerkAPIResponseError` so these are already defined :)

PR #4340, opened 2024-10-15
feat(express): Export `authenticateRequest`

jescalan:
This can be useful to use directly for express users who want to go down to bare metal

## Description

For some users of the express SDK, they may want to directly use `authenticateRequest`, which is notably different from the backend SDK version since it's adapted to express' request wrapper. This makes that possible.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/gorgeous-teachers-scream.md
+++ b/.changeset/gorgeous-teachers-scream.md
@@ -1,5 +1,11 @@
 ---
-"@clerk/express": patch
+"@clerk/express": minor
 ---
 
-Export authenticateRequest method from express sdk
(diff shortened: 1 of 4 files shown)


wobsoriano (approved):
Yeah, let's do this. This one has custom logic that would be a hassle to implement in userland

LekoArts (approved):

ejarnutowski:
Thanks guys.  This will lead to a significantly more straightforward implementation of auth guards in Nest.js.  Appreciate your work.

PR #4347, opened 2024-10-17
chore(nextjs): Support unstable_rethrow inside clerkMiddleware

panteliselef:
## Description

We changed the errors thrown by `protect()` inside `clerkMiddleware` in order for `unstable_rethrow` to recognise them and rethrow them.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/nextjs/src/server/nextErrors.ts
```diff
@@ -0,0 +1,111 @@
...
+ */
+const NOT_FOUND_ERROR_CODE = 'NEXT_NOT_FOUND';
+
+type NotFoundError = Error & {
+  digest: typeof NOT_FOUND_ERROR_CODE;
+  clerk_digest: typeof CONTROL_FLOW_ERROR.FORCE_NOT_FOUND;
+};
+
+function isNextjsNotFoundError(error: unknown): error is NotFoundError {
+  if (typeof error !== 'object' || error === null || !('digest' in error)) {
```
jacekradko, line 22: I am wondering if we can just check `instanceof` instead here 🤔 

```suggestion
  if (!error instanceof Error || !('digest' in error)) {
```

Or perhaps we are trying to also detect duck-typed custom errors?

jacekradko: I see.. this is just an extension of how Next.js does it.  🤷‍♂️

PR #4362, opened 2024-10-18
feat(clerk-react,nextjs,shared): Introduce experimental `useReverification`

panteliselef:
## Description

Introduce a new experimental hook called `useReverification` that makes it easy to handle reverification errors.
It returns a high order function (HOF) and allows developers to wrap any function that triggers a fetch request which might fail due to a user's session verification status.
- When such error is returned, the recommended UX is to offer a way to the user to recover by re-verifying their credentials.
- This helper will automatically handle this flow in the developer's behalf, by displaying a modal the end-user can interact with.
- Upon completion, the original request that previously failed, will be retried (only once).

### Video of UX flow

https://github.com/user-attachments/assets/e2071442-131e-4ea9-881d-ac4971ffd565

### Nextjs example

A client component
```tsx
import { __experimental_useReverification as useReverification } from "@clerk/nextjs";

export const DeletePatientButton = ({ name }: { name: string }) => {
  const [ deletePatient ] = useReverification(deletePatientAction);

  const [, trigger, isPending] = useActionState(deletePatient, { success: false });

  return (
    <button
      disabled={isPending}
      onClick={async () => {
        startTransition(() => {
          trigger(name);
        });
      }}
    >
      <LoadingChildren pending={isPending} />
    </button>
  );
};

function LoadingChildren({ pending }: { pending: boolean }) {
  if (pending) {
    return (
      <>
        Please wait
      </>
    );
  }

  return <>Delete</>;
}
```

our server actions

```ts
"use server";

import { removePatient } from "@/lib/db/patients";
import { auth } from "@clerk/nextjs/server";
import { reverificationMismatch } from "@clerk/shared/authorization-errors";

export const deletePatientAction = async (
  prevState: { sucess: boolean },
  name: string
) => {
  const { orgId, has } = await auth.protect();
  const shouldUserRevalidate = !has({
    __experimental_reverification: "strict",
  })
  if (shouldUserRevalidate) {
    return reverificationMismatch("strict");
  }

  removePatient(name, orgId);
  return { sucess: true };
};
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


tmilewski (changes requested):

packages/react/src/hooks/useReverification.ts
```diff
@@ -0,0 +1,146 @@
...
+    .then(r => {
+      if (r instanceof Response) {
+        return r.json();
+      }
+      return r;
+    })
+    .catch(e => {
+      // Treat fapi assurance as an assurance hint
+      if (isClerkAPIResponseError(e) && e.errors.find(({ code }) => code == 'session_step_up_verification_required')) {
+        return reverificationMismatch({} as __experimental_ReverificationConfig);
```
nikosdouvlis, line 24: ❓ Why do we need the empty object here?

packages/react/src/hooks/useReverification.ts
```diff
@@ -0,0 +1,146 @@
...
+  return (
+    result &&
+    typeof result === 'object' &&
+    'clerk_error' in result &&
+    result.clerk_error?.type === 'forbidden' &&
+    result.clerk_error?.reason === 'reverification-mismatch'
+  );
+};
+
+type InferParameters<T> = T extends (...args: infer P) => any ? P : never;
```
nikosdouvlis, line 66: Is this the same as the built-in `Parameters<function>`?

packages/react/src/hooks/useReverification.ts
```diff
@@ -0,0 +1,146 @@
...
+type InferReturnType<T> = T extends (...args: any[]) => infer R ? R : never;
```
nikosdouvlis, line 67: Is this the same as the built-in `ReturnType<function>`?

packages/react/src/hooks/useReverification.ts
```diff
@@ -0,0 +1,146 @@
...
+
+function createReverificationHandler(params: { onOpenModal: Clerk['__experimental_openUserVerification'] }) {
+  function assertReverification<H extends (...args: InferParameters<H>) => InferReturnType<H>>(
+    fetcher: H,
+  ): (...args: InferParameters<H>) => Promise<ExcludeClerkError<Awaited<InferReturnType<H>>>> {
```
nikosdouvlis, line 72: ❓  

Is `...args` supposed to be the callback the user passed in?  If so, I think we can probably simplify the TS types here by changing `...args` to `callback` as that'd remove all `array` types from the inferred types.

Also, looking at the example in the PR description I understood that the `handleReverification` just needs to return the same type as the return of the callback it accepts as the argument, is that correct?

.changeset/five-insects-grin.md
```diff
@@ -0,0 +1,34 @@
...
+
+Introduce a new experimental hook called `useReverification` that makes it easy to handle reverification errors.
+It returns a high order function (HOF) and allows developers to wrap any function that triggers a fetch request which might fail due to a user's session verification status.
+When such error is returned, the recommended UX is to offer a way to the user to recover by re-verifying their credentials.
+This helper will automatically handle this flow in the developer's behalf, by displaying a modal the end-user can interact with.
+Upon completion, the original request that previously failed, will be retried (only once).
+
+Example with clerk-js methods.
+```tsx
+function DeleteAccount() {
```
nikosdouvlis, line 14: Here's an alternative:

```ts
function DeleteAccount() {
  const { user } = useUser();
  const deleteUser = useReverification(() => {
    if (!user) return;
    return user.delete();
});

  return <>
      <button
        onClick={deleteUser}>
        Delete account
      </button>
  </>
}
```

Would that work?

packages/react/src/hooks/useReverification.ts
panteliselef: seems so

.changeset/five-insects-grin.md
panteliselef: Probably something like this
```tsx
  const { handleReverification } = useReverification();
  const { user } = useUser();

  const deleteUserAccount = handleReverification(() => {
    if (!user) return;
    return user.delete()
  })

```

packages/react/src/hooks/useReverification.ts
panteliselef: i've updated the signature to allow for undefined

packages/react/src/hooks/useReverification.ts
panteliselef: Seems like for the types to be inferred correctly below `return async (...args)` I need to explicitly type them again as `(...args: Parameters<Fetcher>)`

packages/clerk-js/src/ui/components/UserProfile/DeleteUserForm.tsx
```diff
@@ -1,10 +1,9 @@
-import { useClerk, useUser } from '@clerk/shared/react';
+import { __experimental_useReverification, useClerk, useUser } from '@clerk/shared/react';
```
tmilewski, line 1:
```suggestion
import { __experimental_useReverification as useReverification, useClerk, useUser } from '@clerk/shared/react';
```

nikosdouvlis (approved):

tmilewski (approved):

PR #4380, opened 2024-10-22
feat(upgrade): Add ClerkProvider dynamic prop messaging

jacekradko:
## Description

Adding check for usages of `useAuth` so we can message consumers about `<ClerkProvider dynamic>`

Screenshot 2024-10-22 at 2 26 45 PM

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


dstaley (approved):

.changeset/purple-gifts-beam.md
```diff
@@ -0,0 +1,2 @@
+---
```
dstaley, line 1: just confirming: doesn't this need a changeset to trigger a new version?

dstaley (approved):

.changeset/purple-gifts-beam.md
jacekradko: Yes, but there are already queued up version bumps to `@clerk/upgrade` so this will be part of that set of changes. We can make this an explicit changeset also if we feel that's useful

dstaley: nope, no need, totally cool with this being empty then!

PR #4385, opened 2024-10-23
chore(repo): Test task cache, instance cleanup

BRKalow:
## Description

Fixes caching for integration test tasks, all task dependencies incorrectly contained `^`, which means the task should depend on its package's dependencies running the tasks specified with `^`. This was effectively a no-op when used for a root task with package-specific tasks in its dependencies. This should fix integration test caching 🤞

Also fixes an issue with the integration test cleanup script where not all instances were being cleaned up.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,6 +139,7 @@ jobs:
     timeout-minutes: ${{ vars.TIMEOUT_MINUTES_LONG && fromJSON(vars.TIMEOUT_MINUTES_LONG) || 15 }}
 
     strategy:
+      fail-fast: false
       matrix:
         test-name: ['generic', 'express', 'quickstart', 'ap-flows', 'elements', 'sessions', 'astro', 'expo-web', 'tanstack-start', 'tanstack-router']
         test-project: ['chrome']
--- a/integration/cleanup/cleanup.setup.ts
+++ b/integration/cleanup/cleanup.setup.ts
@@ -5,20 +5,19 @@ import { test as setup } from '@playwright/test';
 import { appConfigs } from '../presets/';
 
 setup('cleanup instances ', async () => {
-  const entries = Object.values(appConfigs.envs)
-    .map(e => e.toJson())
-    .map(json => {
-      const secretKey = json.private['CLERK_SECRET_KEY'];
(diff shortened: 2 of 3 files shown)


panteliselef (approved):
What's the reason for dropping `apiUrl`. Unused property ?

.github/workflows/ci.yml
```diff
@@ -139,6 +139,7 @@ jobs:
     timeout-minutes: ${{ vars.TIMEOUT_MINUTES_LONG && fromJSON(vars.TIMEOUT_MINUTES_LONG) || 15 }}
 
     strategy:
+      fail-fast: false
```
BRKalow, line 142: Given the flakiness of our integration tests, it feels valuable to continue running tests even if others fail. Many times tests are minutes into running when cancelled, which I don't think has the cost savings we expect in practice.

integration/cleanup/cleanup.setup.ts
```diff
@@ -5,20 +5,19 @@ import { test as setup } from '@playwright/test';
 import { appConfigs } from '../presets/';
 
 setup('cleanup instances ', async () => {
-  const entries = Object.values(appConfigs.envs)
```
BRKalow, line 8: `appConfig.envs` didn't contain all the keys in use, so not all instances were getting cleaned up as far as I could tell.

turbo.json
```diff
@@ -136,84 +136,79 @@
       "outputs": []
     },
     "//#test:integration:ap-flows": {
-      "dependsOn": ["^@clerk/clerk-js#build", "^@clerk/backend#build", "^@clerk/nextjs#build"],
+      "dependsOn": ["@clerk/clerk-js#build", "@clerk/backend#build", "@clerk/nextjs#build"],
```
BRKalow, line 139: All of the adjustments here remove the `^` for each dependency. The `^` says to wait on the package's dependencies, which for a root task like this effectively seems like a no-op.

I've validated that this properly caches the integration test run only if the dependent build steps remain unchanged, which is the desired behavior.

panteliselef (approved):

PR #4388, opened 2024-10-23
Fix a `cookies` call that was not await-ed

ChristianIvicevic:
## Description

This PR fixes a dangling `cookies()` call that was not awaited causing errors when using Next.js 15. With `@clerk/nextjs` v6.0.0 I noticed that performing certain actions such as clicking a `<SignOutButton>` would error with that particular callsite.

Since this function was already marked as `async` there shouldn't be any other required changes.

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/nextjs/src/app-router/server-actions.ts
```diff
@@ -2,10 +2,6 @@
 
 import { cookies } from 'next/headers';
 
-// This function needs to be async as we'd like to support next versions in the range of [14.1.2,14.2.0)
-// These versions required 'use server' files to export async methods only. This check was later relaxed
-// and the async is no longer required in newer next versions.
-// ref: https://github.com/vercel/next.js/pull/62821
 export async function invalidateCacheAction() {
-  return cookies().delete(`__clerk_invalidate_cache_cookie_${Date.now()}`);
+  return (await cookies()).delete(`__clerk_invalidate_cache_cookie_${Date.now()}`);
```
BRKalow, line 6:
```suggestion
  void (await cookies()).delete(`__clerk_invalidate_cache_cookie_${Date.now()}`);
```

BRKalow:
Thank you so much! We missed this one.

Closing in favor of https://github.com/clerk/javascript/pull/4396 as we also need to not return the result of the call.

PR #4366, opened 2024-10-18
feat(nextjs): Next.js@15 compatibility

jacekradko:
## Description

Next.js@15 compatibility integration branch

## Checklist

- [ ] `npm test` runs as expected.
- [ ] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #4398, opened 2024-10-23
fix(nextjs): Include protect types in `auth`

panteliselef:
## Description

During #4366 we typed the arguments of `protect()` for page/route/action as `any`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/src/app-router/server/auth.ts
+++ b/packages/nextjs/src/app-router/server/auth.ts
@@ -66,6 +66,6 @@ auth.protect = async (...args) => {
     redirect,
   });
 
-  // @ts-ignore
+  // @ts-expect-error TS flattens all possible combinations of the for AuthProtect signatures in a union.
   return protect(...args);
 };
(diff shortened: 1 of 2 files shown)


packages/nextjs/src/app-router/server/auth.ts
```diff
@@ -59,5 +65,7 @@ auth.protect = async (...args: any[]) => {
     notFound,
     redirect,
   });
+
+  // @ts-ignore
```
alexcarpenter, line 69: Can we include the reasoning for ts-ignore here?

jacekradko (approved):
Good catch! 🙇🏼

packages/nextjs/src/app-router/server/auth.ts
panteliselef: i was meant to, i forgot 😄

PR #4406, opened 2024-10-24
feat(clerk-js,localizations,types): Enable email verification in `UserProfile` via enterprise SSO

NicolasLopes7:
## Description

Resolves ORGS-395

Introduces the ability to verify an email address based on an existing enterprise SSO connection. It supports the following use case:
- User tests an app within their personal account within an organization
- Their organization admin upgrades to the SaaS enterprise plan
- App owner sets up SSO for their customer organization via Clerk Dashboard
- User adds their organization email in their profile, triggering verification with enterprise SSO

https://github.com/user-attachments/assets/c99b672c-fa1e-48d0-b43b-1f3fd0cbe4a0

Disclaimers from the video above: 
- The enterprise account icon is broken since our staging CDN doesn't have that asset, but it works for production
- I was already authenticated on the identity provider side, therefore the callback was super quick

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/components/UserProfile/EmailForm.tsx
```diff
@@ -21,10 +22,9 @@ export const EmailForm = withCardStateProvider((props: EmailFormProps) => {
   const { user } = useUser();
   const environment = useEnvironment();
   const preferEmailLinks = emailLinksEnabledForInstance(environment);
-
   const [createEmailAddress] = useReverification(() => user?.createEmailAddress({ email: emailField.value }));
-
   const emailAddressRef = React.useRef<EmailAddressResource | undefined>(user?.emailAddresses.find(a => a.id === id));
+  const strategy = getVerificationStrategy(emailAddressRef.current, preferEmailLinks);
```
LauraBeatris, line 27: TODO: We could map each component based on the strategy instead of performing multiple conditionals. 

Also should take a look at not using the ref, but actually using the email within user settings.

packages/clerk-js/src/ui/components/UserProfile/VerifyWithEnterpriseConnection.tsx
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * The following workaround is used in order to make magic links work when the
+     * <UserProfile/> is used as a modal. In modals, the routing is virtual. For
+     * magic links the flow needs to end by invoking the /verify path of the <UserProfile/>
+     * that renders the <VerificationSuccessPage/>. So, we use the userProfileUrl that
+     * defaults to Clerk Hosted Pages /user as a fallback.
+     */
+    const { routing } = profileContext;
+    const baseUrl = routing === 'virtual' ? displayConfig.userProfileUrl : '';
+    const redirectUrl = buildEmailLinkRedirectUrl(profileContext, baseUrl);
```
LauraBeatris, line 39: TODO: We should rename `buildEmailLinkRedirectUrl` to `buildVerificationRedirectUrl`

packages/clerk-js/src/ui/components/UserProfile/ConnectedAccountsSection.tsx
```diff
@@ -164,7 +164,7 @@ const ConnectedAccount = ({ account }: { account: ExternalAccountResource }) =>
           <ImageOrInitial />
           <Box sx={{ whiteSpace: 'nowrap', overflow: 'hidden' }}>
             <Flex
-              gap={2}
+              gap={1}
```
LauraBeatris, line 167: This fixes an issue with the spacing between the provider and email address:

With gap 1:

CleanShot 2024-12-06 at 07 38 19

With gap 2:

CleanShot 2024-12-06 at 07 37 55

panteliselef:
@brkalow @NicolasLopes7 @LauraBeatris In the video demo we can see that we are loosing the state of the modal when the user returns back to the app. We do have this mechanism in place for OAuth connections via the UserProfile, do you think we should add it to this new flow as well ?

LauraBeatris:
@panteliselef I introduced changes on https://github.com/clerk/javascript/pull/4406/commits/65231ce62af8fc4943d0afc80b406fb15ffe016e to preserve the `UserProfile` modal state upon SSO redirection

I've updated the PR description with a video - disclaimer: the enterprise account icon is broken since our staging CDN doesn't have that asset, but it works for production

packages/localizations/src/de-DE.ts
```diff
@@ -694,19 +694,23 @@ export const deDE: LocalizationResource = {
...
         formTitle: 'Bestätigungslink',
         resendButton: 'Link erneut senden',
         successMessage: 'Die E-Mail-Adresse {{identifier}} wurde Ihrem Konto hinzugefügt.',
       },
+      enterpriseSsoLink: {
+        formSubtitle: undefined,
+        resendButton: undefined,
+        successMessage: undefined,
+      },
+      formHint: undefined,
```
LekoArts, line 713:
```suggestion
        formSubtitle: 'Ein neuer Tab hat sich geöffnet, schließe die Anmeldung mit {{identifier}} dort ab.',
        resendButton: 'Anmeldung erneut versuchen',
        successMessage: 'Die E-Mail {{identifier}} wurde zu Ihrem Account hinzugefügt.',
      },
      formHint: 'Diese E-Mail-Adresse wird durch Ihre primäre Verifizierungsmethode bestätigt.',
```

packages/clerk-js/src/core/resources/EmailAddress.ts
```diff
@@ -78,6 +81,45 @@ export class EmailAddress extends BaseResource implements EmailAddressResource {
     return { startEmailLinkFlow, cancelEmailLinkFlow: stop };
   };
 
+  createEnterpriseSsoLinkFlow = (): CreateEnterpriseSsoLinkFlowReturn<
```
brkalow, line 84: I believe we use `SSO` (all caps) in names (seeing that for `buildSSOCallbackURL` at least):
```suggestion
  createEnterpriseSSOLinkFlow = (): CreateEnterpriseSSOLinkFlowReturn<
```

packages/clerk-js/src/ui/hooks/useEnterpriseSsoLink.ts
```diff
@@ -0,0 +1,29 @@
...
+  StartEnterpriseSsoLinkFlowParams,
+} from '@clerk/types';
+import React from 'react';
+
+type EnterpriseSsoLinkEmailAddressReturn = CreateEnterpriseSsoLinkFlowReturn<
+  StartEnterpriseSsoLinkFlowParams,
+  EmailAddressResource
+>;
+
+function useEnterpriseSsoLink(resource: EmailAddressResource): EnterpriseSsoLinkEmailAddressReturn {
```
brkalow, line 13:
```suggestion
function useEnterpriseSSOLink(resource: EmailAddressResource): EnterpriseSSOLinkEmailAddressReturn {
```

packages/clerk-js/src/ui/hooks/useEnterpriseSsoLink.ts
```diff
@@ -0,0 +1,29 @@
...
+  const { startEnterpriseSsoLinkFlow, cancelEnterpriseSsoLinkFlow } = React.useMemo(
+    () => resource.createEnterpriseSsoLinkFlow(),
```
brkalow, line 15:
```suggestion
    () => resource.createEnterpriseSSOLinkFlow(),
```

packages/clerk-js/src/ui/hooks/useEnterpriseSsoLink.ts
```diff
@@ -0,0 +1,29 @@
...
+} from '@clerk/types';
+import React from 'react';
+
+type EnterpriseSsoLinkEmailAddressReturn = CreateEnterpriseSsoLinkFlowReturn<
+  StartEnterpriseSsoLinkFlowParams,
+  EmailAddressResource
+>;
+
+function useEnterpriseSsoLink(resource: EmailAddressResource): EnterpriseSsoLinkEmailAddressReturn {
+  const { startEnterpriseSsoLinkFlow, cancelEnterpriseSsoLinkFlow } = React.useMemo(
```
brkalow, line 14:
```suggestion
  const { startEnterpriseSSOLinkFlow, cancelEnterpriseSSOLinkFlow } = React.useMemo(
```

packages/clerk-js/src/ui/hooks/useEnterpriseSsoLink.ts
```diff
@@ -0,0 +1,29 @@
...
+    return cancelEnterpriseSsoLinkFlow;
+  }, []);
+
+  return {
+    startEnterpriseSsoLinkFlow,
+    cancelEnterpriseSsoLinkFlow,
+  };
+}
+
+export { useEnterpriseSsoLink };
```
brkalow, line 29:
```suggestion
  React.useEffect(() => {
    return cancelEnterpriseSSOLinkFlow;
  }, []);

  return {
    startEnterpriseSSOLinkFlow,
    cancelEnterpriseSSOLinkFlow,
  };
}

export { useEnterpriseSSOLink };
```

packages/clerk-js/src/ui/hooks/useEnterpriseSsoLink.ts
```diff
@@ -0,0 +1,29 @@
...
+  CreateEnterpriseSsoLinkFlowReturn,
+  EmailAddressResource,
+  StartEnterpriseSsoLinkFlowParams,
+} from '@clerk/types';
+import React from 'react';
+
+type EnterpriseSsoLinkEmailAddressReturn = CreateEnterpriseSsoLinkFlowReturn<
+  StartEnterpriseSsoLinkFlowParams,
+  EmailAddressResource
+>;
```
brkalow, line 11:
```suggestion
import type {
  CreateEnterpriseSSOLinkFlowReturn,
  EmailAddressResource,
  StartEnterpriseSSOLinkFlowParams,
} from '@clerk/types';
import React from 'react';

type EnterpriseSSOLinkEmailAddressReturn = CreateEnterpriseSSOLinkFlowReturn<
  StartEnterpriseSSOLinkFlowParams,
  EmailAddressResource
>;
```

packages/clerk-js/src/ui/hooks/index.ts
```diff
@@ -18,3 +18,4 @@ export * from './useDebounce';
 export * from './useScrollLock';
 export * from './useClerkModalStateParams';
 export * from './useNavigateToFlowStart';
+export * from './useEnterpriseSsoLink';
```
brkalow, line 21:
```suggestion
export * from './useEnterpriseSSOLink';
```

brkalow (dismissed):
Let's use consistent casing across our SSO APIs: change all `Sso` -> `SSO`

packages/types/src/localization.ts
```diff
@@ -484,20 +484,29 @@ type _LocalizationResource = {
     emailAddressPage: {
       title: LocalizationValue;
       verifyTitle: LocalizationValue;
+      formHint: LocalizationValue;
       emailCode: {
+        /**
+         * @deprecated UserProfile now only uses `emailAddressPage.formHint`.
+         */
```
anagstef, line 491: If I'm not missing something, this is a breaking change. Users will get the latest ClerkJS version but not the latest localization file, if they are using a different localization than `en-US`.

LauraBeatris: I'll push some changes to keep the previous form hints instead of removing them, therefore it won't break for previous clerk-js versions.

packages/clerk-js/src/ui/components/UserProfile/__tests__/EmailsSection.test.tsx
```diff
@@ -48,7 +48,7 @@ describe('EmailSection', () => {
       await waitFor(() => getByRole('heading', { name: /Add email address/i }));
 
       getByLabelText(/email address/i);
-      getByText('An email containing a verification code will be sent to this email address.');
+      getByText('This email address will be verified using your primary authentication method.');
```
panteliselef, line 51: Does that mean it is required to verify ownership of the new email address ? I don't think so, so I must be missing something

packages/types/src/localization.ts
LauraBeatris: For newer clerk-js versions, the new localizations will have to be added for the "generic" form hint, which I think is reasonable considering it won't display the label:

CleanShot 2025-01-21 at 09 01 43

And then it'll keep showing the localized ones per strategy:

CleanShot 2025-01-21 at 09 03 55

packages/types/src/emailAddress.ts
```diff
@@ -11,6 +17,10 @@ export type PrepareEmailAddressVerificationParams =
   | {
       strategy: EmailLinkStrategy;
       redirectUrl: string;
+    }
+  | {
+      strategy: EnterpriseSSOStrategy;
+      redirectUrl: string;
```
panteliselef, line 23: 🙃
```suggestion
  | {
      strategy: EmailLinkStrategy | EnterpriseSSOStrategy
      redirectUrl: string;
    }
```

packages/types/src/signUp.ts
```diff
@@ -133,6 +133,11 @@ export type PrepareVerificationParams =
       strategy: SamlStrategy | EnterpriseSSOStrategy;
       redirectUrl?: string;
       actionCompleteRedirectUrl?: string;
+    }
+  | {
+      strategy: EnterpriseSSOStrategy;
+      redirectUrl?: string;
+      actionCompleteRedirectUrl?: string;
```
panteliselef, line 140: I have to ask now that i see it twice, but why not this ?
```suggestion
      strategy: SamlStrategy | EnterpriseSSOStrategy 
| EnterpriseSSOStrategy
      redirectUrl?: string;
      actionCompleteRedirectUrl?: string;
    }
```

packages/types/src/localization.ts
LauraBeatris: Updated on https://github.com/clerk/javascript/pull/4406/commits/299abde85794cde015942f531bee82fab82dcd1c

packages/types/src/verification.ts
```diff
@@ -43,3 +43,12 @@ export type CreateEmailLinkFlowReturn<Params, Resource> = {
   startEmailLinkFlow: (params: Params) => Promise<Resource>;
   cancelEmailLinkFlow: () => void;
 };
+
+export interface StartEnterpriseSSOLinkFlowParams {
+  redirectUrl: string;
+}
+
+export type CreateEnterpriseSSOLinkFlowReturn<Params, Resource> = {
```
panteliselef, line 51: Is there a pattern for param for  folks or us to respect here ? Could we have `Params extends MyStructure` ?

packages/clerk-js/src/core/resources/EmailAddress.ts
```diff
@@ -78,6 +81,45 @@ export class EmailAddress extends BaseResource implements EmailAddressResource {
...
+    StartEnterpriseSSOLinkFlowParams,
+    EmailAddressResource
+  > => {
+    const { run, stop } = Poller();
+
+    const startEnterpriseSSOLinkFlow = async ({
+      redirectUrl,
+    }: StartEnterpriseSSOLinkFlowParams): Promise<EmailAddressResource> => {
+      if (!this.id) {
+        clerkVerifyEmailAddressCalledBeforeCreate('SignUp');
```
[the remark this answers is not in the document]
LauraBeatris: We're applying the same for email links: 

https://github.com/clerk/javascript/blob/a7c7ea273a7c30b01361c7c0eaff766040fcfea1/packages/clerk-js/src/core/resources/EmailAddress.ts#L57-L60

The above is wrong. The race condition we want to catch here is if `user?.createEmailAddress` isn't called before verifying.

The error message output should be:

`ClerkJS: You need to start an EmailAddress flow by calling EmailAddress.create() first.`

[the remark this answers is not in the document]
LauraBeatris: I think "EmailAddress flow" doesn't make sense 😅 - it should be: "You need to create an EmailAddress by calling `EmailAddress.create()` first."

I'll add a new error message for this case

packages/clerk-js/src/ui/components/UserProfile/__tests__/EmailsSection.test.tsx
panteliselef: Reading "This email address will be verified using your primary authentication method."

I understand that if my primary email is "pantelis@gmail.com" and I want to add "pantelis@clerk.dev", something will be send to "pantelis@gmail.com".

I just found the text a bit confusing.

LauraBeatris: That makes sense, I'll think of a better label here and push some changes

packages/clerk-js/src/core/resources/EmailAddress.ts
```diff
@@ -78,6 +81,45 @@ export class EmailAddress extends BaseResource implements EmailAddressResource {
...
+                resolve(res);
+              }
+            })
+            .catch(err => {
+              stop();
+              reject(err);
+            });
+        });
+      });
+    };
```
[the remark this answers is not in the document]
LauraBeatris: It doesn't work with Safari. I'm currently refactoring to open the `externalVerificationRedirectURL` with the browser navigation API on a button click, as we have for connected accounts. 

It's more reliable than introducing a workaround for `window.open` here specifically for Safari, in which we wouldn't have control if their behavior changes.

LauraBeatris:
Updated UI to rely on button click to kick-off the verification flow, instead of using `window.open`:

packages/types/src/verification.ts
LauraBeatris: `Params` seems to very a lot, so I'm afraid we wouldn't be able to restrict to a specific shape/pattern:

CleanShot 2025-01-21 at 17 18 12

packages/clerk-js/src/ui/components/UserProfile/utils.ts
```diff
@@ -16,12 +10,6 @@ export const currentSessionFirst = (id: string) => (a: IDable) => (a.id === id ?
 
 export const defaultFirst = (a: PhoneNumberResource) => (a.defaultSecondFactor ? -1 : 1);
 
-export function emailLinksEnabledForInstance(env: EnvironmentResource): boolean {
```
LauraBeatris, line 19: I've moved this to `UserProfile` since it was easier to read it being placed in the same context as we changed requirements to include `enterprise_sso` 

Happy to move all of the utils included on this PR to this utils file here as well

panteliselef (approved):

PR #4345, opened 2024-10-17
fix(tanstack-start): Merge router context instead of updating

jeremy-clerk:
## Description

This PR fixes bug where the initialisation of Clerk overwrites the context, by merging the original/existing context with the `clerkInitialState`. 

## Checklist

- [x] `npm test` runs as expected.
- [x] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #4414, opened 2024-10-27
Fix Tanstack Start Context Corruption and stop loaders from re-running.

phibr0:
## Description

This PR fixes two issues in the Tanstack Start Plugin: Currently, on the initial (server-side) rendered load, the default router context is overwritten with Clerk's initial state. This was problematic for our app because we provide the tRPC client via the default router context. Since it was overwritten on the second run, all loaders would fail. It also doesn’t seem necessary to call `router.load()`, since Tanstack Start will do that anyway, having that in here, was causing all loaders to run twice.

To test this, you can use my minimal reproduction: https://github.com/phibr0/clerk-tanstack-start-repro

1. Provide keys in a `.env` file (`VITE_CLERK_PUBLISHABLE_KEY`, `CLERK_SECRET_KEY`).
2. Run `pnpm install`.
3. Run `pnpm dev`.
   - Navigate to http://localhost:3000/.
   - The route context will be logged in the server console.

Screenshot 2024-10-27 at 12 42 38

The context is logged **once**, and the property `THIS_IS_DEFINED_IN_ROOT_ROUTE` exists.

4. Remove the patch by deleting `patchedDependencies` in the `package.json`.
5. Run `pnpm install && pnpm dev`.
   - Navigate to http://localhost:3000/.
   - The route context will be logged twice in the server console, and the second time, the `THIS_IS_DEFINED_IN_ROOT_ROUTE` property is missing.

(This is the current behavior.)

Screenshot 2024-10-27 at 12 49 10

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/cold-tigers-scream.md
+++ b/.changeset/cold-tigers-scream.md
@@ -1,0 +1,8 @@
+---
+"@clerk/tanstack-start": patch
+---
+
+Fix Tanstack Start SSR
+
+- Stop overwriting default router context
+- Stop unnecessarily re-running route loaders
--- a/packages/tanstack-start/src/server/middlewareHandler.ts
+++ b/packages/tanstack-start/src/server/middlewareHandler.ts
@@ -28,7 +28,10 @@ export function createClerkHandler<TRouter extends AnyRouter>(
 
         // Updating the TanStack router context with the Clerk context and loading the router
         router.update({
-          context: clerkInitialState,
+          context: {
+            ...router.options.context,
+            ...clerkInitialState,
+          },
         });
 
         await router.load();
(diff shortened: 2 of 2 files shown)


alexcarpenter:
This is a duplicate of https://github.com/clerk/javascript/pull/4345

alexcarpenter:
Thanks @phibr0 for the PR! Will be closing in favor of https://github.com/clerk/javascript/pull/4345 which was already in progress.

phibr0:
No problem! Not sure how I missed that

kaandok:
@alexcarpenter This is still happening since the fix has been reverted in https://github.com/clerk/javascript/commit/dac21ed7ba585549a1e981902f0fe66d90fb4ba5

PR #4427, opened 2024-10-29
feat(elements,ui,clerk-js): Legal consent elements support and improvements

octoper:
## Description

This PR adds Elements support for Legal consent feature, it also includes implementation of the UI build with Elements and includes a couple improvements on the AIO components
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/ui/src/components/sign-up/steps/start.tsx
+++ b/packages/ui/src/components/sign-up/steps/start.tsx
@@ -146,7 +146,7 @@ export function SignUpStart() {
 
                   {options.socialButtonsPlacement === 'bottom' ? connectionsWithSeperator.reverse() : null}
 
-                  {!(hasConnection && hasIdentifier && legalConsentEnabled) && <LegalAcceptedField />}
+                  {legalConsentEnabled && hasConnection && !hasIdentifier && <LegalAcceptedField />}
 
                   {userSettings.signUp.captcha_enabled ? <SignUp.Captcha className='empty:hidden' /> : null}
                 </Card.Body>


packages/elements/src/internals/machines/sign-up/utils/fields-to-params.ts
```diff
@@ -17,11 +25,17 @@ export function fieldsToSignUpParams<T extends SignUpCreateParams | SignUpUpdate
...
     if (isSignUpParam(key) && value !== undefined) {
       params[key] = value as string;
     }
+
+    if (isSignUpParam(key) && type === 'checkbox') {
+      params[key] = checked as boolean;
+    }
   });
 
+  console.log(params);
```
tmilewski, line 38:
```suggestion
```

tmilewski (approved):
Elements stuff generally looks good to me, I'll leave UI up to @alexcarpenter.

packages/elements/src/internals/machines/sign-up/utils/fields-to-params.ts
octoper: Good catch!

packages/ui/src/common/legal-accepted.tsx
```diff
@@ -0,0 +1,77 @@
...
+            className={cx('flex-shrink-0', className)}
+            checked={checked}
+            {...restProps}
+          >
+            <Field.Checkbox />
+          </Common.Input>
+
+          <Common.Label asChild>
+            <Field.Label>
+              <span>
```
alexcarpenter, line 59: is there a reason for the extra wrapping span here?

packages/ui/src/components/sign-up/steps/start.tsx
```diff
@@ -144,10 +146,13 @@ export function SignUpStart() {
...
                   {options.socialButtonsPlacement === 'bottom' ? connectionsWithSeperator.reverse() : null}
 
+                  {!(hasConnection && hasIdentifier && legalConsentEnabled) && <LegalAcceptedField />}
+
                   {userSettings.signUp.captcha_enabled ? <SignUp.Captcha className='empty:hidden' /> : null}
                 </Card.Body>
-                {hasConnection || hasIdentifier ? (
+                {hasIdentifier ? (
                   <Card.Actions>
+                    {legalConsentEnabled && hasIdentifier && <LegalAcceptedField />}
```
alexcarpenter, line 155: Similar to the continue step, I feel like this should live within the body not the actions.

packages/ui/src/common/legal-accepted.tsx
```diff
@@ -0,0 +1,77 @@
...
+    });
+  }
+
+  return (
+    <Common.Field
+      name='__experimental_legalAccepted'
+      asChild
+    >
+      <Field.Root>
+        <div className='flex justify-center gap-2'>
```
alexcarpenter, line 46: Will want to replace the `gap-2` usage here with padding left on the label.

alexcarpenter (commented):
Couple small notes, but everything else looks good.

packages/ui/src/common/legal-accepted.tsx
octoper: Makes sense!

packages/ui/src/components/sign-up/steps/start.tsx
octoper: This is how we render it on AIO right now, it's closer to the `Continue` button when sign up has identifier, when not having identifier and only having social buttons we put this closer to the social button (within the body), if this feels a bit close to the continue button it's probably because we need to adjust the gap

octoper: If this still does not look right to you, we can bring Alvish to take a look and after that I will make the appropriate changes

alexcarpenter: ok, either way lets make it consistent. Looks like in the continue step its placed in the body.

packages/ui/src/common/legal-accepted.tsx
alexcarpenter: Ok, maybe we need a prop on the label to set the flex direction? Trying to reduce the amount of extra markup added to reduce the need for descriptors.

packages/ui/src/components/sign-up/steps/start.tsx
octoper: Yeah it's different there though as we wanted to have it on the body, because it's one of the missing fields and it didn't looked right when we were looking the designs with Alvish

packages/elements/src/internals/machines/sign-up/continue.machine.ts
```diff
@@ -62,6 +62,21 @@ export const SignUpContinueMachine = setup({
       context.parent.send({ type: 'NEXT', resource: (event as unknown as DoneActorEvent<SignUpResource>).output }),
     sendToLoading,
   },
+  guards: {
+    isMissingRequirements: ({ context }) =>
+      context.parent.getSnapshot().context.clerk?.client?.signUp?.status === 'missing_requirements',
+    requirementsMet: ({ context }) => {
```
tmilewski, line 68: Per a convo with @octoper, we're going to rename this guard.

tmilewski (approved):
Per Elements, 👍. Leaving the rest up to @alexcarpenter.

alexcarpenter (approved):
👍🏼 will tackle any tweaks in a follow up pr. Thanks @octoper!

packages/elements/src/internals/machines/sign-up/continue.machine.ts
octoper: Done!

PR #4434, opened 2024-10-30
fix(nextjs): Nonce not being populated in Clerk provider and browser script

jeremy-clerk:
## Description

Fixes a bug where the nonce was not being applied to the Clerk browser script causing CSP errors and failing to load Clerk. 
Added an else statement to the ClerkProvider otherwise the nonce was resolving to an empty string. Potentially could be moved to line 32 of of ClerkProvider, but it may cause issues in older versions of Next. 

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/nervous-months-happen.md
+++ b/.changeset/nervous-months-happen.md
@@ -1,0 +1,5 @@
+---
+"@clerk/nextjs": patch
+---
+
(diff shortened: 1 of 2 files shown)


jeremy-clerk:
Breaks static rendering - won't work.

PR #4435, opened 2024-10-30
fix(repo): Ensure turbo cache is used properly for e2e tests

BRKalow:
## Description

Currently, args pass to turbo tasks are (understandably) included in the hash, and ultimately the cache key. As it relates to e2e tests, we pass a `--project` argument to the turbo task run, along with `--only`. What this means is turbo doesn't consider the task's dependencies when determining if the task has been cached or not.

This is problematic for us, as only the `integration/` directory is included in the inputs. I've changed the command to not pass the `--project` arg, as it only has the `chrome` value today, and is otherwise unused. I also removed the `--only` flag. This should allow turbo to only re-run the integration tests if dependent packages have changed.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -25,8 +25,7 @@
     "VITE_CLERK_*",
     "EXPO_PUBLIC_CLERK_*",
     "REACT_APP_CLERK_*",
-    "NEXT_PHASE",
-    "E2E_NEXTJS_VERSION"
+    "NEXT_PHASE"
   ],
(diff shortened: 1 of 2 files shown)


turbo.json
```diff
@@ -25,8 +25,7 @@
     "VITE_CLERK_*",
     "EXPO_PUBLIC_CLERK_*",
     "REACT_APP_CLERK_*",
-    "NEXT_PHASE",
-    "E2E_NEXTJS_VERSION"
```
BRKalow, line 29: This would bust the build cache across tests against diff next versions, which we don't want.

panteliselef (approved):

PR #4376, opened 2024-10-22
feat(clerk-js,nextjs,clerk-react,types): Introduce `<Waitlist />` component

nikospapcom:
## Description

Show `Join waitlist` promt from `<SignIn />` component when mode is `waitlist`
Change the text in `RestrictedAccess` component when mode is `waitlist`
Introduce `<Waitlist />` component to allow users to join in the waitlist via email
Introduce `joinWaitlist` method in `clerk` class
Introduce `redirectToWaitlist` function in `clerk` class to allow user to redirect to waitlist page

## Checklist

- [ ] `npm test` runs as expected.
- [ ] `npm run build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

### `<SignIn />` Component with `sign_up.mode = waitlist`
Screenshot 2024-10-23 at 5 07 42 PM

### `<SignUp />` Component with `sign_up.mode = waitlist`
Screenshot 2024-10-23 at 5 07 35 PM

### `<Waitlist />` Component 
Screenshot 2024-10-23 at 5 07 48 PM

### `<Waitlist />` Component success screen
Screenshot 2024-10-23 at 5 08 19 PM

### `<Waitlist />` Component success screen with `redirectUrl` presented
Screenshot 2024-10-23 at 5 08 07 PM

PR #4461, opened 2024-11-02
feat(vue): Introduce Vue SDK

wobsoriano:
## Description

This PR introduces an experimental version of Vue SDK! It ports the original code of [`vue-clerk`](https://github.com/wobsoriano/vue-clerk) inside this monorepo as a new package called `@clerk/vue`.

These are the adjustments that happened while porting the original code:

- Removed [`IsomorphicClerk`](https://github.com/wobsoriano/vue-clerk/blob/main/packages/vue-clerk/src/isomorphicClerk.ts) as it's hard to keep in sync with `@clerk/clerk-react`'s `IsomorphicClerk` and there's really no benefit of it in Vue.
- Custom pages and menu items have been intentionally excluded from this PR to focus on the main components and functions. These features will be implemented in a separate PR to follow.
- Added unit tests to unstyled components using Vitest
- Dropped CJS support (nobody's using cjs in Vue since Vue 3)

Things to keep in mind when reviewing this PR:

- All components (UI, Control, Unstyled) are using [Render Functions](https://vuejs.org/guide/extras/render-function) - this allow us to use pure JavaScript to create Vue components. Another plus is that we dont have to create a separate `.vue` file for each component and keep all related components in the same file.
- When you see the term "composables", it's like "hooks" in React for Vue.
- All components and hooks from `@clerk/clerk-react` have an equivalent in `@clerk/vue`. They all use the same name.

We will introduce E2E tests and quickstart repo in a separate PR.

Resolves ECO-222

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/vue/src/components/controlComponents.ts
wobsoriano, line 1: This file exports all [Clerk control components](https://clerk.com/docs/components/overview#what-are-control-components) that is a 1:1 equivalent of the React components

packages/vue/src/components/uiComponents.ts
wobsoriano, line 1: This file contains all Clerk UI components that is a 1:1 equivalent of the React UI components. They can be used like

```vue
<script setup>
import { UserProfile } from '@clerk/vue'
</script>

<template>
  <UserProfile />
</template>
```

packages/vue/tsup.config.ts
```diff
@@ -0,0 +1,31 @@
...
+    sourcemap: true,
+    minify: false,
+    dts: true,
+    esbuildPlugins: [
+      // Automatically generates runtime props from TypeScript types/interfaces for all
+      // control and UI components, adding them to Vue components during build via
+      // Object.defineProperty
+      autoPropsPlugin({
+        include: ['**/*.ts'],
+      }) as EsbuildPlugin,
```
wobsoriano, line 23: Instead of redeclaring all props for all of available components, this plugin will generate a runtime prop using the type prop.

So this component:

Screenshot 2024-11-04 at 2 15 55 PM

will turn into this when built:

Screenshot 2024-11-04 at 9 59 17 AM

So without this plugin, we will have to manually add runtime props to each component, and make sure they will always match.

packages/vue/src/keys.ts
```diff
@@ -0,0 +1,5 @@
+import type { InjectionKey } from 'vue';
+
+import type { VueClerkInjectionKeyType } from './types';
+
+export const ClerkInjectionKey = Symbol('clerk') as InjectionKey<VueClerkInjectionKeyType>;
```
wobsoriano, line 5: Read more [here](https://vuejs.org/guide/components/provide-inject#working-with-symbol-keys) regarding Symbols as key

octoper (approved):

packages/vue/src/components/uiComponents.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export const OrganizationProfile = defineComponent((props: OrganizationProfileProps) => {
+  const clerk = useClerk();
+  return () =>
+    h(Portal, {
+      mount: clerk.value?.mountOrganizationProfile,
+      unmount: clerk.value?.unmountOrganizationProfile,
+      props,
+    });
+});
```
octoper, line 135: There is `<Waitlist/>` missing, which was added [today](https://github.com/clerk/javascript/pull/4376/files)

wobsoriano: Thanks! Was not aware of this, but I think it makes sense to add it in this PR. Just a few lines of code!

wobsoriano: Updated!

wobsoriano:
!snapshot

packages/vue/src/components/uiComponents.ts
```diff
@@ -0,0 +1,176 @@
...
+
+export const UserProfile = defineComponent((props: UserProfileProps) => {
+  const clerk = useClerk();
+
+  return () =>
+    h(Portal, {
+      mount: clerk.value?.mountUserProfile,
+      unmount: clerk.value?.unmountUserProfile,
+      updateProps: (clerk.value as any)?.__unstable__updateProps,
```
octoper, line 67: ❓ Why are we using `as any` here?

octoper (approved):

packages/vue/src/global.d.ts
```diff
@@ -0,0 +1,6 @@
+export {};
+
+declare global {
+  const PACKAGE_NAME: string;
+  const PACKAGE_VERSION: string;
+}
```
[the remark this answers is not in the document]
wobsoriano: That is correct, thanks!

packages/vue/src/components/controlComponents.ts
```diff
@@ -0,0 +1,205 @@
...
+  return () => null;
+});
+
+export const RedirectToSignUp = defineComponent((props: RedirectOptions) => {
+  const clerk = useClerk();
+
+  watchEffect(() => {
+    if (!clerk.value) {
+      return;
+    }
```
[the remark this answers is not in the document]
wobsoriano: Good catch. That should be valid in Vue land and should be simpler to write and less boilerplate.

I was also thinking of having something like `withClerk`:

```ts
export const RedirectToSignUp = withClerk(
  defineComponent(({ clerk, ...props }: WithClerkProp<SignUpRedirectOptions>) => {
    onMounted(() => {
      void props.clerk.redirectToSignUp(props);
    });

    return () => null;
  }),
)
```

this gives the same mental model as what we have in the React SDK but requires slightly more setup code 🤔

packages/vue/src/composables/useOrganization.ts
```diff
@@ -0,0 +1,98 @@
+import type { OrganizationMembershipResource, OrganizationResource } from '@clerk/types';
+import { computed } from 'vue';
+
+import { useClerkContext } from './useClerkContext';
+import { useSession } from './useSession';
+import type { ToComputedRefs } from './utils';
+import { toComputedRefs } from './utils';
+
+type UseOrganizationReturn =
```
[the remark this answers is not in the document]
wobsoriano: Regarding the type suggestions, I agree and let's do it in a separate ticket. I can help with this after completing the Vue project 🫡. The Astro project will also benefit from these shared types.

packages/vue/src/utils/useClerkLoaded.ts
```diff
@@ -0,0 +1,31 @@
...
+    unwrappedClerk => {
+      if (!unwrappedClerk?.loaded) {
+        return;
+      }
+
+      callback(unwrappedClerk as LoadedClerk);
+    },
+    { immediate: true },
+  );
+};
```
wobsoriano, line 31: This composable is used inside the control components where we want to execute a Clerk method when Clerk is loaded.

We're using a `watch` here instead of `watchEffect` to react to a specific state, instead of watching every reactivate state inside the callback

packages/vue/src/components/controlComponents.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    const hasActiveSessions = clientCtx.value?.activeSessions && clientCtx.value.activeSessions.length > 0;
+
+    if (sessionCtx.value === null && hasActiveSessions) {
+      void clerk.value.redirectToAfterSignOut();
+    } else {
+      void clerk.value.redirectToSignIn(props);
+    }
```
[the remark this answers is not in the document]
wobsoriano: Added a ticket that will address shared types and this 🫡

packages/vue/src/components/controlComponents.ts
```diff
@@ -0,0 +1,171 @@
...
+  RedirectOptions,
+} from '@clerk/types';
+import { defineComponent } from 'vue';
+
+import { useAuth } from '../composables/useAuth';
+import { useClerk } from '../composables/useClerk';
+import { useClerkContext } from '../composables/useClerkContext';
+import { useClerkLoaded } from '../utils/useClerkLoaded';
+
+export const SignedIn = defineComponent((_, { slots }) => {
```
LekoArts, line 15: Would users benefit from us adding JSDoc comments to these exports or wouldn't it be shown in ones IDE?

packages/vue/src/composables/useClerk.ts
```diff
@@ -0,0 +1,26 @@
+import { useClerkContext } from './useClerkContext';
+
+/**
+ * Returns the [`Clerk`](https://clerk.com/docs/references/javascript/clerk/clerk) instance which provides
+ * low-level access to Clerk's functionality, enabling custom authentication UI and flows.
+ *
+ * @example
+ * A simple example:
```
LekoArts, line 8: We can remove the "A simple example" from the `@example` blocks (also in other files of this PR) because the tag will already mark it as such + the word "simple" should probably be avoided as for some folks it might indeed not be simple (and thus discouraging)

packages/vue/src/composables/useClerkContext.ts
```diff
@@ -0,0 +1,15 @@
+import { inject } from 'vue';
+
+import { ClerkInjectionKey } from '../keys';
+
+export function useClerkContext() {
+  const ctx = inject(ClerkInjectionKey);
+
+  if (!ctx) {
+    throw new Error(
+      'This component/composable can only be used when the Vue plugin is installed. Learn more: https://clerk.com/docs/components/clerk-provider',
```
LekoArts, line 10: Shouldn't this link to the installation docs of the Clerk Vue SDK instead?

packages/vue/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilerOptions": {
+    "target": "ESNext",
```
jacekradko, line 3: Are we planning on only supporting latest modern browsers with this SDK?

packages/vue/src/composables/useClerkContext.ts
wobsoriano: Updated to add a quickstart link to Vue SDK (docs to follow)

packages/vue/src/composables/useClerk.ts
wobsoriano: Updated! thank you

packages/vue/src/components/controlComponents.ts
wobsoriano: Good question! So the doc shows up in the import statement:

Screenshot 2024-11-07 at 7 51 59 AM

but not when used inside template:

Screenshot 2024-11-07 at 7 52 10 AM

I skipped it because of that but we can definitely add it 🤔

packages/vue/tsconfig.json
wobsoriano: Good question! I looked up Vue's recommended [tsconfig](https://github.com/vuejs/tsconfig/blob/22e15e3947c8ce1d9c5992229da0351c464fb9e8/tsconfig.json#L56) and saw they use `ESNext`. I updated it to ES2020 to be more explicit 👍🏼

packages/vue/src/components/controlComponents.ts
LekoArts: Why would it only show up on the import? I guess typically you'd expect this to exist during usage

wobsoriano: Okay, looks like it is a limitation when using render functions. When doing normal Vue SFC, it works:

Screenshot 2024-11-08 at 7 33 53 AM

Screenshot 2024-11-08 at 7 34 09 AM

wobsoriano: Let's tackle this in a separate ticket 🫡

PR #4462, opened 2024-11-02
fix(nextjs): Import helper from subpath to avoid static analysis warnings

nikosdouvlis:
## Description

We need to import from the shared subpath, otherwise next prod builds will report the following issue:

```A Node.js API is used (MessageEvent at line: 28) which is not supported in the Edge Runtime.
Learn more: https://nextjs.org/docs/api-reference/edge-runtime

Import trace for requested module:
./node_modules/@clerk/shared/dist/chunk-RSOCGYTF.mjs ./node_modules/@clerk/shared/dist/index.mjs
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


BRKalow (approved):
🙏

packages/shared/src/index.ts
```diff
@@ -1,4 +1,4 @@
-/** The following files are not exported on purpose:
+1; /** The following files are not exported on purpose:
```
jacekradko, line 1:
```suggestion
/** The following files are not exported on purpose:
```

nikosdouvlis: Oh good catch! 
I'll commit the fix when the tests are done so we don't have to wait for them to run again, thanks!

PR #4483, opened 2024-11-04
fix(clerk-js): Directory traversal relative URL detection

BRKalow:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/utils/__tests__/url.test.ts
+++ b/packages/clerk-js/src/utils/__tests__/url.test.ts
@@ -84,8 +84,10 @@ describe('isValidUrl(url,base)', () => {
   });
 });
 
-describe('isRelativeUrl(url,base)', () => {
-  const cases: Array<[string, boolean]> = [
+describe.only('isRelativeUrl(url,base)', () => {
+  const cases: Array<[string, boolean]> = [['/%2e%2e/%2e%2e/evil.com', false]];
+
+  const cases1: Array<[string, boolean]> = [
     ['', true],
     ['/', true],
     ['/test', true],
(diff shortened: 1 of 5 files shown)


packages/clerk-js/src/utils/url.ts
```diff
@@ -352,20 +353,25 @@ export function requiresUserInput(redirectUrl: string): boolean {
   return frontendApiRedirectPathsWithUserInput.includes(url.pathname);
 }
 
-export const isAllowedRedirectOrigin =
-  (allowedRedirectOrigins: Array<string | RegExp> | undefined) => (_url: string) => {
-    if (!allowedRedirectOrigins) {
-      return true;
+export const isAllowedRedirect =
```
BRKalow, line 356: This function now deals with full URLs, instead of potentially relative URLs as well. This makes our checks more consistent, and removes situations where we treat a relative URL as okay, make it absolute, and then it ends up being problematic.

packages/clerk-js/src/utils/__tests__/url.test.ts
```diff
@@ -84,18 +84,38 @@ describe('isValidUrl(url,base)', () => {
   });
 });
 
-describe('isRelativeUrl(url,base)', () => {
+describe('isProblematicUrl(url)', () => {
```
BRKalow, line 87: Could potentially do with more test cases here, but with the shift to checking against full URLs we get more guarantees from the existing origin check.

nikosdouvlis (approved):

PR #4486, opened 2024-11-05
fix(clerk-js): Replace generic ComponentContext with component-specific contexts

dstaley:
## Description

This PR replaces the generic `ComponentContext` provider with component-specific versions.

The biggest benefit to this PR is reducing the amount of "magic" happening, and moving us towards a more typical approach to handling Context within React applications. Instead of a generic Context that can be one of 11 different types, we now have 11 separate Contexts. This adds a bit more safety in that there won't be an instance where the `SignIn` component tries to access the Context of another component (which we currently guard against via a runtime check of `componentName`). Ultimately it's a small change intended to simplify complexity and increase readability.

Progress:
- [x] `SignIn`
- [x] `SignUp`
- [x] `UserProfile`
- [x] `UserVerification`
- [x] `UserButton`
- [x] `OrganizationSwitcher`
- [x] `OrganizationList`
- [x] `OrganizationProfile`
- [x] `CreateOrganization`
- [x] `GoogleOneTap`
- [x] `Waitlist`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
```diff
@@ -74,6 +148,12 @@ export const useSignUpContext = (): SignUpContextType => {
     [],
   );
 
+  if (!context || context.componentName !== 'SignUp') {
```
dstaley, line 151: For simplicity we retain the componentName check. In a future PR we can refactor this out since we always know that we're accessing a value from the expected Context.

packages/clerk-js/src/ui/utils/test/createFixtures.tsx
```diff
@@ -84,22 +89,32 @@ const unboundCreateFixtures = <N extends UnpackContext<typeof ComponentContext>[
...
+        <>{children}</>
+      );
+
       return (
         <CoreClerkContextWrapper
           clerk={clerkMock}
           // Clear swr cache
           swrConfig={{ provider: () => new Map() }}
         >
+          <GoogleOneTapContext.Provider value={{ componentName: 'GoogleOneTap' }} />
```
alexcarpenter, line 111: Whats the reasoning for this addition outside the other providers?

dstaley: oh whoops! Removed

alexcarpenter (approved):

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
```diff
@@ -258,30 +342,38 @@ export const useUserProfileContext = (): UserProfileContextType => {
...
+  if (!context || context.componentName !== 'UserVerification') {
     throw new Error('Clerk: useUserVerificationContext called outside of the mounted UserVerification component.');
   }
 
+  const { componentName, ...ctx } = context;
+
   return {
     ...ctx,
     componentName,
   };
```
jacekradko, line 359: Looks like this could be simplified to just 
```suggestion
  return context;
```

jacekradko (approved):
lgtm

packages/clerk-js/src/ui/contexts/ClerkUIComponentsContext.tsx
dstaley: yeah this is left over from the previous implementation. When I refactor this to split these contexts into separate files and remove the `componentName` check I'll clean this up as well!

PR #4498, opened 2024-11-06
fix(clerk-js): Init options on updateProps

nikosdouvlis:
We need to re-init the options during the updateProps() call  in order to keep the options passed to ClerkProvider in sync with the state of clerk-js. If we don't init the options here again, the following scenario is possible:
1. User renders <ClerkProvider propA={undefined} propB={1} />
2. clerk-js initializes propA with a default value
3. The customer update propB independently of propA and window.Clerk.updateProps is called
4. If we don't merge the new props with the current options, propA will be reset to undefined

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/nine-pets-rhyme.md
+++ b/.changeset/nine-pets-rhyme.md
@@ -1,0 +1,5 @@
+---
+"@clerk/clerk-js": patch
+---
+
+Re-init window.Clerk options when `ClerkProvider` props change in `@clerk/clerk-react`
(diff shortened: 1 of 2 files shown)


panteliselef (approved):
💯

anagstef (approved):

PR #4509, opened 2024-11-06
feat(clerk-js): Add nonce to theme provider

jescalan:
## Description

If using a [strict-dynamic CSP](https://clerk.com/docs/security/clerk-csp#implementing-a-strict-dynamic-csp) with Clerk, you still currently need to add `unsafe-inline` to the `style-src` property since we do not thread the nonce into the emotion cache. This change will make it such that you can run the nonce into `style-src` instead of `unsafe-inline` and it will work, which means there are no unsafe properties required to be passed for strict dynamic policies when using Clerk in production.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


jacekradko (approved):
@jescalan The changes look good. Thanks!

PR #4511, opened 2024-11-07
chore(clerk-js): Use fapi error long message instead of statusText when throwing API errors

panteliselef:
## Description
Use FAPI error long message instead of statusText when throwing API errors
Also a customized error for using production keys in dev is added.

### Before
Screenshot 2024-11-07 at 1 31 44 PM

### After
Screenshot 2024-11-07 at 1 30 57 PM

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -67,8 +68,22 @@ export abstract class BaseResource {
...
+        throw new ClerkAPIResponseError(
+          `Clerk: Production Keys are only allowed for domain "${prodDomain}". \nAPI Error: ${safeErrorMessage}`,
+          {
+            data: errors,
+            status: status,
+          },
+        );
+      }
+
+      throw new ClerkAPIResponseError(safeErrorMessage || statusText, {
```
panteliselef, line 85: 1. `statusText` was empty most of the time (not sure why, `status` was populated)
2. We do have a custom `toString()` for `ClerkAPIResponseError` but that is not guaranteed to be called.

packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -67,8 +68,22 @@ export abstract class BaseResource {
...
+      if (errors?.[0]?.code === 'origin_invalid' && isProductionFromPublishableKey(BaseResource.clerk.publishableKey)) {
+        const prodDomain = BaseResource.clerk.frontendApi.replace('clerk.', '');
+        throw new ClerkAPIResponseError(
+          `Clerk: Production Keys are only allowed for domain "${prodDomain}". \nAPI Error: ${safeErrorMessage}`,
+          {
+            data: errors,
+            status: status,
+          },
+        );
+      }
```
anagstef, line 83: Can we move this in a separate file or function, because I'd like us to keep the `_fetch` function as clean as possible?

LekoArts (approved):

anagstef (approved):

PR #4515, opened 2024-11-07
fix(clerk-js): Prevent memory leak in worker timers setTimeout

myandrienko:
## Description

When creating a timeout using worker timers, timeout callback is saved on the page side in map, and a mapping between page-side id and worker-side id is saved in an object in the worker.

Both of these entries are only cleaned up if the timeout is canceled with `clearTimeout`. Normally this isn't done. So even after timeout has fired, related entries are kept, causing a memory leak. It's especially noticeable when `setTimeout` is called recursively, as is done in `Poller` used by email link flow.

This PR cleans timeout-related entries once the timer has successfully fired.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


BRKalow (approved):
Thanks!

myandrienko:
@BRKalow Anything else I can do here to help this get merged?

PR #4518, opened 2024-11-07
feat(clerk-js,types): Surface enterprise accounts in `UserProfile`

LauraBeatris:
## Description

Resolves ORGS-293, ORGS-294 

Surfaces `user.enterprise_accounts` instead of `user.saml_accounts`. This is going to support multiple enterprise protocols from now on, such as OpenID connect.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -15,58 +42,83 @@ export const EnterpriseAccountsSection = () => {
...
       </ProfileSection.ItemList>
     </ProfileSection.Root>
   );
 };
+
+const EnterpriseAccountProviderIcon = ({ account }: { account: EnterpriseAccountResource }) => {
+  const { provider } = account;
+  const providerName = getEnterpriseAccountProviderName(account);
+
+  if (isOAuthBuiltInProvider(provider) || isSamlProvider(provider)) {
```
LauraBeatris, line 60: We could simplify this by looking at `account.enterpriseConnection.logoPublicUrl`, but I'll wait until FAPI changes get deployed in order to validate this.

packages/clerk-js/src/ui/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -1,12 +1,39 @@
...
+  SamlIdpSlug,
+} from '@clerk/types';
+import { getOAuthProviderData, OAUTH_PROVIDERS, SAML_IDPS } from '@clerk/types';
 
+import { ProviderInitialIcon } from '../../common';
 import { Badge, Box, descriptors, Flex, Image, localizationKeys, Text } from '../../customizables';
 import { ProfileSection } from '../../elements';
-import { useSaml } from '../../hooks';
+
+const isSamlProvider = (provider: EnterpriseProvider): provider is SamlIdpSlug => provider.startsWith('saml');
```
LauraBeatris, line 16:
```suggestion
const isSamlProvider = (provider: EnterpriseProvider): provider is SamlIdpSlug => provider.includes('saml');
```

packages/types/src/appearance.ts
```diff
@@ -1,4 +1,5 @@
 import type * as CSS from 'csstype';
+import type { EnterpriseProvider } from 'enterpriseAccount';
```
LauraBeatris, line 2:
```suggestion
import type { EnterpriseProvider } from './enterpriseAccount';
```

packages/clerk-js/src/ui/components/UserProfile/AccountPage.tsx
```diff
@@ -12,21 +12,22 @@ import { UserProfileSection } from './UserProfileSection';
...
-  const { attributes, saml, social } = useEnvironment().userSettings;
+  const { attributes, social } = useEnvironment().userSettings;
   const card = useCardState();
   const { user } = useUser();
 
   const showUsername = attributes.username.enabled;
   const showEmail = attributes.email_address.enabled;
   const showPhone = attributes.phone_number.enabled;
   const showConnectedAccounts = social && Object.values(social).filter(p => p.enabled).length > 0;
-  const showSamlAccounts = saml && saml.enabled && user && user.samlAccounts.length > 0;
```
LauraBeatris, line 23: @NicolasLopes7 @izaaklauer Do you know what `saml.enabled` means in this case? Is it `true` when you have at least one SAML connection?

LauraBeatris:
@NicolasLopes7 Preserving the `disableAdditionalIdentifications` check:

CleanShot 2024-11-12 at 15 33 01

https://github.com/user-attachments/assets/5d5db062-ed21-44bb-8b48-4f4bda4cea02

packages/clerk-js/src/ui/components/UserProfile/__tests__/EnterpriseAccountsSection.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+          id: 'ver_123',
+          object: 'verification',
+        },
+        id: 'eac_123',
+      },
+    ],
+  });
+});
+
+const withSamlEnterpriseConnection = createFixtures.config(f => {
```
LauraBeatris, line 116: Should also add a test for when `userSettings.saml` is disabled, or when the enterprise connection is inactive.

packages/clerk-js/src/ui/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -15,58 +31,110 @@ export const EnterpriseAccountsSection = () => {
...
+};
+
+const EnterpriseAccountProviderIcon = ({ account }: { account: EnterpriseAccountResource }) => {
+  const { provider } = account;
+  const providerName = getEnterpriseAccountProviderName(account);
+
+  if (isOAuthCustomProvider(provider)) {
+    return (
+      <ProviderInitialIcon
+        id={provider}
```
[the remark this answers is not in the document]
alexcarpenter: Maybe introduce a more generic `providerInitialIcon` descriptor?

```tsx
elementDescriptor={[descriptors.providerIcon, descriptors.providerInitialIcon]}
elementId={descriptors.enterpriseButtonsProviderIcon.setId(account.provider)}
```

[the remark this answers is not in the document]
LauraBeatris: @alexcarpenter I've added a new description that it's generic for SSO connections, and not specific to enterprise or social.

Also in that commit, it includes a fix where we display the logo URL for custom OAuth providers that can be defined from the Clerk Dashboard.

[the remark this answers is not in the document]
alexcarpenter: I think we need to preserve `socialButtonsProviderInitialIcon` descriptors for within social buttons since those could already be in use and would be a breaking change removing them.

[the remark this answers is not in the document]
LauraBeatris: Gotcha, does it make sense to pass it as prop to `ProviderInitialIcon` then? 🤔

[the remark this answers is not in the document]
LauraBeatris: Actually, I think it makes sense to provide as an array as you mentioned - I'll update here later.

packages/types/src/json.ts
```diff
@@ -218,7 +248,11 @@ export interface UserJSON extends ClerkResourceJSON {
   phone_numbers: PhoneNumberJSON[];
   web3_wallets: Web3WalletJSON[];
   external_accounts: ExternalAccountJSON[];
+  enterprise_accounts: EnterpriseAccountJSON[];
   passkeys: PasskeyJSON[];
+  /**
+   * @deprecated use `enterprise_accounts` instead
+   */
   saml_accounts: SamlAccountJSON[];
```
LauraBeatris, line 256: FAPI is going to continue returning `saml_accounts`, to be backwards compatibility, however from now on newer clerk-js versions we should implement against `enterprise_accounts`.

packages/clerk-js/src/ui/components/UserProfile/EnterpriseAccountsSection.tsx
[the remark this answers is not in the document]
LauraBeatris: @alexcarpenter I've added a more generic description here and kept `socialButtonsProviderInitialIcon` [here](https://github.com/clerk/javascript/pull/4518/commits/3466922e99e279a1ee7c5578e0e2f1395d36a328)

packages/clerk-js/src/ui/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -15,58 +32,115 @@ export const EnterpriseAccountsSection = () => {
...
+    ) : (
+      <ProviderInitialIcon
+        id={provider}
+        value={providerName}
+        aria-label={`${providerName}'s icon`}
+      />
+    );
+  }
+
+  const providerWithoutPrefix = provider.replace('oauth_', '').trim() as OAuthProvider;
```
izaaklauer, line 113: We talked about this a bit with @Nikpolik . It's a bit weird that we're stripping the `oauth_` here, but it's certainly not the first time we've done it. Sometimes `provider` has that `oauth_` prefix, and sometimes it doesn't.

It would be good for us to figure out what we want `provider` to mean, and then standardize it across the board. But that can happen in the future, and shouldn't block this PR.

alexcarpenter: It is being used in element descriptors, so we should feel confidant with its usage before this goes out since it'd be a breaking change to remove/modify it.

packages/clerk-js/src/ui/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -15,58 +31,114 @@ export const EnterpriseAccountsSection = () => {
...
+  if (isOAuthCustomProvider(provider)) {
+    return enterpriseConnection?.name;
+  }
+
+  return getOAuthProviderData({ strategy: provider })?.name;
+};
+
+const isSamlProvider = (provider: EnterpriseProvider): provider is SamlIdpSlug => provider.includes('saml');
+
+const isOAuthCustomProvider = (provider: EnterpriseProvider): provider is `oauth_${CustomOauthProvider}` =>
```
LauraBeatris, line 143: Ideally, we shouldn't check this against the provider value. It'd be better to check against a unique discriminator property on `EnterpriseAccount`. 

Since we only have `protocol=oauth`, but it could be either EASIE or Enterprise OAuth, then the SDK needs to check against the provider value.

cc @Nikpolik @izaaklauer

Nikpolik: Agreed if they need different handling they should have a different type.

Any opinions on `oidc` & `easie` VS `oauth_custom` & `oauth_easie`?

Nikpolik: ❓ If this is just for the image, built in OAuth and SAML should have the logo already set. Is there any reason to handle it differently?

packages/clerk-js/src/ui/components/UserProfile/EnterpriseAccountsSection.tsx
LauraBeatris: @alexcarpenter Yeah, we need to be mindful of updating `provider` values in the future, since element descriptors rely on it and it could be a breaking change.

On this PR, I'm keeping the behavior as before, where we have the following variations: 
- `enterpriseButtonsProviderIcon__ saml_okta`
- `enterpriseButtonsProviderIcon__ google`
- `enterpriseButtonsProviderIcon__ custom_roblox` (could be any other custom provider name)

packages/clerk-js/src/ui/components/UserProfile/EnterpriseAccountsSection.tsx
LauraBeatris: I've refactored it here to rely on FAPI as the source of truth for connection names, provider names, and logos, instead of mapping to our objects on the SDK. 

The benefit of this is that we won't have to do those explicit checks now against the protocol/provider enum to them access on our own mappings for labels.

panteliselef (approved):
🛳️

packages/types/src/appearance.ts
```diff
@@ -172,7 +173,10 @@ export type ElementsConfig = {
   socialButtonsProviderIcon: WithOptions<OAuthProvider | Web3Provider, LoadingState>;
   socialButtonsProviderInitialIcon: WithOptions<OAuthProvider | Web3Provider, LoadingState>;
 
-  enterpriseButtonsProviderIcon: WithOptions<SamlIdpSlug, LoadingState>;
+  enterpriseButtonsProviderIcon: WithOptions<EnterpriseProvider, LoadingState>;
+
+  providerIcon: WithOptions<OAuthProvider | Web3Provider | SamlIdpSlug, LoadingState>;
+  providerInitialIcon: WithOptions<OAuthProvider | Web3Provider | SamlIdpSlug, LoadingState>;
```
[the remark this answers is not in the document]
alexcarpenter: only one letter shows

PR #4530, opened 2024-11-08
chore(repo): Increase timeouts and retries in CI

dstaley:
## Description

This PR increases our `retries`, `maxFailures`, and `timeout` in CI in an effort to combat some of our flaky tests.

The hypothesis is that our runners are simply being overwhelmed during more complicated tests like sessions, which runs multiple applications and a proxy server. These increases are intended to suss out whether we might see better pass rates if we throw more hardware at the problem.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/integration/playwright.config.ts
+++ b/integration/playwright.config.ts
@@ -12,9 +12,9 @@ export const common: PlaywrightTestConfig = {
   snapshotDir: './tests/snapshots',
   fullyParallel: true,
   forbidOnly: !!process.env.CI,
-  retries: process.env.CI ? 2 : 0,
+  retries: process.env.CI ? 5 : 0,
   timeout: 90_000,
-  maxFailures: process.env.CI ? 1 : undefined,
+  maxFailures: process.env.CI ? 5 : undefined,
   workers: process.env.CI ? '50%' : '70%',
   reporter: process.env.CI ? 'line' : 'list',
   use: {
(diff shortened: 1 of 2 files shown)


panteliselef (approved):

integration/playwright.config.ts
```diff
@@ -12,9 +12,9 @@ export const common: PlaywrightTestConfig = {
   snapshotDir: './tests/snapshots',
   fullyParallel: true,
   forbidOnly: !!process.env.CI,
-  retries: process.env.CI ? 2 : 0,
+  retries: process.env.CI ? 5 : 0,
   timeout: 90_000,
-  maxFailures: process.env.CI ? 1 : undefined,
+  maxFailures: process.env.CI ? 5 : undefined,
```
anagstef, line 17: ❓  Will this allow tests to pass even if some of them fail? 🤔

LekoArts: As I understand the option it means that now in CI 5 tests can fail and only then the rest of the tests is skipped / the suite is stopped: https://playwright.dev/docs/test-parallel#limit-failures-and-fail-fast

So currently we stop the whole suite after one test fails

nikospapcom (approved):

PR #4537, opened 2024-11-11
feat(@clerk/backend): add createTotp method to call create TOTP in usersAPI

SzymonMatynia:
## Description

This branch introduces a createTotp method in the Users API. The createTotp method is a simple wrapper, calling the necessary API endpoint for TOTP setup.

example of use:
`clerkClient.users.createTOTP(userId)`

ℹ️ Endpoint exists in the API documentation but not in the SDK
https://clerk.com/docs/reference/backend-api/tag/Users#operation/CreateUserTOTP
image

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


panteliselef:
Hey @SzymonMatynia is there a use case for you needing this method ? We are thinking of deprecating the endpoint itself.

SzymonMatynia:
> Hey @SzymonMatynia is there a use case for you needing this method ? We are thinking of deprecating the endpoint itself.

Hey @panteliselef Yes, there is. I need to generate OTP codes on the fly on API Gateway side (backend) when specific case is met (in this case ip address is not in given range).

Since, I can't get `totp_secret` from the user directly, it seems to be the only way to get such approach working. (to regenerate totp to generate otp when necessary, + the endpoint itself already exists)

panteliselef:
> I need to generate OTP codes on the fly on API Gateway side (backend) when specific case is met (in this case ip address is not in given range).

Hey @SzymonMatynia after talking with our Backend Engineering team we concluded that this would not be a valid use-case for that endpoint. We will be deprecating that endpoint from our Backend API shortly, and we want to avoid introducing this on our SDK since the decision for deprecating it has been made.

SzymonMatynia:
@panteliselef Yup, well I got your point and I agree with your decision

Anyway, I have a question then - Do you plan to add an endpoint to e.g. generate OTP based on current user tote secret? Because I'm not able to generate otp code manually :/

The case: I have to trigger OTP validation when: a user logs in outside working hours or is not in specified ip range.

octoper:
Hey @SzymonMatynia!

As @panteliselef mentioned before this endpoint is going to be deprecated so I will close this PR.

Regarding your question about "generate OTP based on the current user totp secret", I will have to check with the team and reach back to you!

PR #4542, opened 2024-11-12
fix(clerk-js): Propagate `redirectUrl` when continuing sign up with missing identifier

LauraBeatris:
## Description

When SSO connections return no email claim, users are correctly prompted for their email but encounter an error due to missing `action_complete_redirect_url` during submission.

We should continue to sign up by calling `completeSignUpFlow` with the `redirectUrl` so that it can call the SSO callback correctly. 

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

Fixes ORGS-328

### DIFF
--- a/packages/clerk-js/src/ui/components/SignUp/SignUpContinue.tsx
+++ b/packages/clerk-js/src/ui/components/SignUp/SignUpContinue.tsx
@@ -1,6 +1,7 @@
 import { useClerk } from '@clerk/shared/react';
 import React, { useMemo } from 'react';
 
+import { buildSSOCallbackURL } from '../../common';
 import { useCoreSignUp, useEnvironment, useSignUpContext } from '../../contexts';
 import { descriptors, Flex, Flow, localizationKeys } from '../../customizables';
 import {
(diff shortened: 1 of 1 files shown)


packages/clerk-js/src/ui/components/SignUp/SignUpContinue.tsx
```diff
@@ -150,6 +152,9 @@ function _SignUpContinue() {
     card.setLoading();
     card.setError(undefined);
 
+    const redirectUrl = buildSSOCallbackURL(ctx, displayConfig.signUpUrl);
+    const redirectUrlComplete = ctx.afterSignUpUrl || '/';
```
panteliselef, line 156: 🙃 I believe we already properly resolve this inside the `RedirectUrls` class
```suggestion
    const redirectUrlComplete = ctx.afterSignUpUrl
```

packages/clerk-js/src/ui/components/SignUp/SignUpContinue.tsx
```diff
@@ -150,6 +152,9 @@ function _SignUpContinue() {
     card.setLoading();
     card.setError(undefined);
 
+    const redirectUrl = buildSSOCallbackURL(ctx, displayConfig.signUpUrl);
```
panteliselef, line 155: I see that we do this constantly, does it make sense to move it inside `useSignUpContext` ?

PR #4552, opened 2024-11-13
fix(clerk-js): Only apply id to FormFeedback when the type is error

alexcarpenter:
## Description

Fixes an issue where multiple elements in the dom include the same id attribute which leads to invalid markup and causes a11y issues.

multiples-errors

Since we only apply `aria-describedby` on the input element when there is an `errorMessageId`, we only apply the id to the `FormFeedback` when its of type `error`

After:

after

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


tmilewski (approved):

BRKalow (approved):
👏

PR #4555, opened 2024-11-14
fix(clerk-js): No TokenUpdate side-effect on `new Session()`

BRKalow:
## Description

Currently, initialization of a `Session` resource has a side-effect of emitting a `TokenUpdate` event. This is problematic for a few reasons:

1. Setting of the session cookie relies on the `clerk.organization` "accessor" being available, which is set after the client object is fetched and initialized
2. Session resources are initialized during the Client object init, so this side-effect triggers before the accessors are available
3. A global side-effect like this in a class constructor is ripe for problematic bugs

I believe the intent here is to ensure the session cookie is fresh immediately on load of Clerk, and so I've moved this behavior to explicitly live in the Clerk loading flow.

fixes SDKI-765

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


panteliselef (approved):
lgtm

alexcarpenter (approved):

nikosdouvlis:
!snapshot

nikosdouvlis:
(don't mind me, just testing the snapshot name change)

nikosdouvlis:
!snapshot

PR #4559, opened 2024-11-14
Protect /tokens with captcha

nikosdouvlis:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/resources/Session.ts
+++ b/packages/clerk-js/src/core/resources/Session.ts
@@ -1,5 +1,5 @@
 import { createCheckAuthorization } from '@clerk/shared/authorization';
-import { is4xxError } from '@clerk/shared/error';
+import { is4xxError, isClerkAPIResponseError } from '@clerk/shared/error';
 import { runWithExponentialBackOff } from '@clerk/shared/utils';
 import type {
   __experimental_SessionVerificationJSON,
@@ -22,6 +22,7 @@ import type {
   UserResource,
 } from '@clerk/types';
 
+import { getCaptchaToken, retrieveCaptchaInfo } from '../../utils/captcha';
 import { unixEpochToDate } from '../../utils/date';
 import { clerkInvalidStrategy } from '../errors';
 import { eventBus, events } from '../events';
@@ -267,9 +268,24 @@ export class Session extends BaseResource implements SessionResource {
       return cachedToken.getRawString() || null;
     }
     const path = template ? `${this.path()}/tokens/${template}` : `${this.path()}/tokens`;
+
     // TODO: update template endpoint to accept organizationId
-    const params = template ? {} : { organizationId };
-    const tokenResolver = Token.create(path, params);
+    const params: Record<string, string | null> = template ? {} : { organizationId };
+
+    const createTokenWithCaptchaProtection = async () => {
+      try {
+        return Token.create(path, params);
+      } catch (e) {
+        if (isClerkAPIResponseError(e) && e.errors[0].code === 'requires_captcha') {
+          const captchaParams = await this.#triggerCaptchaChallenge();
+          return Token.create(path, { ...params, ...captchaParams });
+        }
+        throw e;
+      }
+    };
+
+    const tokenResolver = createTokenWithCaptchaProtection();
+
     SessionTokenCache.set({ tokenId, tokenResolver });
     return tokenResolver.then(token => {
       if (shouldDispatchTokenUpdate) {
@@ -278,5 +294,26 @@ export class Session extends BaseResource implements SessionResource {
       // Return null when raw string is empty to indicate that there it's signed-out
       return token.getRawString() || null;
     });
+  }
+
+  async #triggerCaptchaChallenge() {
+    const { captchaSiteKey, canUseCaptcha, captchaURL, captchaWidgetType, captchaProvider, captchaPublicKeyInvisible } =
+      retrieveCaptchaInfo(Session.clerk);
+
+    if (canUseCaptcha && captchaSiteKey && captchaURL && captchaPublicKeyInvisible) {
+      return getCaptchaToken({
+        siteKey: captchaSiteKey,
+        widgetType: captchaWidgetType,
+        invisibleSiteKey: captchaPublicKeyInvisible,
+        scriptUrl: captchaURL,
+        captchaProvider,
+        modalWrapperQuerySelector: '#cl-modal-captcha-wrapper',
+        modalContainerQuerySelector: '#cl-modal-captcha-container',
+        openModal: () => Session.clerk.__internal_openBlankCaptchaModal(),
+        closeModal: () => Session.clerk.__internal_closeBlankCaptchaModal(),
+      });
+    }
+
+    return {};
   }
 }
--- a/packages/clerk-js/src/utils/captcha/turnstile.ts
+++ b/packages/clerk-js/src/utils/captcha/turnstile.ts
@@ -1,7 +1,8 @@
+import { waitForElement } from '@clerk/shared/dom';
 import { loadScript } from '@clerk/shared/loadScript';
-import type { CaptchaWidgetType } from '@clerk/types';
 
 import { CAPTCHA_ELEMENT_ID, CAPTCHA_INVISIBLE_CLASSNAME } from './constants';
+import type { CaptchaOptions } from './types';
 
 // We use the explicit render mode to be able to control when the widget is rendered.
 // CF docs: https://developers.cloudflare.com/turnstile/get-started/client-side-rendering/#disable-implicit-rendering
@@ -41,6 +42,10 @@ interface RenderOptions {
    */
   'error-callback'?: (errorCode: string) => void;
   /**
+   * A JavaScript callback invoked before the challenge enters interactive mode.
+   */
+  'before-interactive-callback'?: () => void;
+  /**
    * A JavaScript callback invoked when a given client/browser is not supported by the widget.
    */
   'unsupported-callback'?: () => boolean;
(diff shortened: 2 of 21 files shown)


packages/clerk-js/src/utils/captcha/turnstile.ts
```diff
@@ -112,19 +116,19 @@ async function loadCaptchaFromFAPIProxiedURL(fallbackUrl: string) {
...
-export const getTunstileToken = async (captchaOptions: {
-  siteKey: string;
-  scriptUrl: string;
-  widgetType: CaptchaWidgetType;
-  invisibleSiteKey: string;
-}) => {
-  const { siteKey, scriptUrl, widgetType, invisibleSiteKey } = captchaOptions;
-  let captchaToken = '',
-    id = '';
+export const getTurnstileToken = async (opts: CaptchaOptions) => {
```
nikosdouvlis, line 119: I plan to refactor this using a strategy pattern in an upcoming PR

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -267,9 +268,24 @@ export class Session extends BaseResource implements SessionResource {
...
     // TODO: update template endpoint to accept organizationId
-    const params = template ? {} : { organizationId };
-    const tokenResolver = Token.create(path, params);
+    const params: Record<string, string | null> = template ? {} : { organizationId };
+
+    const createTokenWithCaptchaProtection = async () => {
+      try {
+        return Token.create(path, params);
+      } catch (e) {
+        if (isClerkAPIResponseError(e) && e.errors[0].code === 'requires_captcha') {
```
anagstef, line 279: Can we extract the `requires_captcha` error code to a variable. I can see it is used twice!

packages/clerk-js/src/utils/captcha/turnstile.ts
```diff
@@ -112,59 +116,71 @@ async function loadCaptchaFromFAPIProxiedURL(fallbackUrl: string) {
...
+  const { modalContainerQuerySelector, modalWrapperQuerySelector, closeModal, openModal } = opts;
+  const captcha: Turnstile = await loadCaptcha(scriptUrl);
+  const errorCodes: (string | number)[] = [];
+
+  let captchaToken = '';
+  let id = '';
   let turnstileSiteKey = siteKey;
+  let retries = 0;
+  let widgetContainerQuerySelector: string | undefined;
+  const widgetTypeUsed = [widgetType || ''];
```
anagstef, line 130: I don't understand why this is an array and we combine these at the end. I think that the backend expects `smart` or `invisible` here as a value.

packages/clerk-js/src/utils/captcha/turnstile.ts
```diff
@@ -112,59 +116,71 @@ async function loadCaptchaFromFAPIProxiedURL(fallbackUrl: string) {
...
+  }
 
-  let widgetDiv: HTMLElement | null = null;
+  // smart widget with container provided by user
+  if (!widgetContainerQuerySelector && widgetType === 'smart') {
+    widgetContainerQuerySelector = `#${CAPTCHA_ELEMENT_ID}`;
+    const visibleDiv = document.getElementById(CAPTCHA_ELEMENT_ID);
+    if (visibleDiv) {
+      visibleDiv.style.display = 'block';
+    } else {
```
anagstef, line 145: This breaks the fallback mechanism to invisible if the element with the `CAPTCHA_ELEMENT_ID` does not exist.
```suggestion
  if (!widgetContainerQuerySelector && widgetType === 'smart') {
    const visibleDiv = document.getElementById(CAPTCHA_ELEMENT_ID);
    if (visibleDiv) {
      visibleDiv.style.display = 'block';
      widgetContainerQuerySelector = `#${CAPTCHA_ELEMENT_ID}`;
    } else {
```

packages/clerk-js/src/utils/captcha/turnstile.ts
```diff
@@ -112,59 +116,71 @@ async function loadCaptchaFromFAPIProxiedURL(fallbackUrl: string) {
...
+        'Cannot initialize Smart CAPTCHA widget because the `clerk-captcha` DOM element was not found; falling back to Invisible CAPTCHA widget. If you are using a custom flow, visit https://clerk.com/docs/custom-flows/bot-sign-up-protection for instructions',
+      );
+    }
+  }
 
-  const createInvisibleDOMElement = () => {
+  // invisible widget for which we create the container automatically
+  if (!widgetContainerQuerySelector) {
+    turnstileSiteKey = invisibleSiteKey;
+    widgetTypeUsed.push('invisible');
```
anagstef, line 155: Again, here the `widgetTypeUsed` will end up having `['invisible','invisible']` as a value, if the user has selected the `Invisible` option from the Dashboard.

packages/clerk-js/src/utils/captcha/turnstile.ts
```diff
@@ -209,14 +225,18 @@ export const getTunstileToken = async (captchaOptions: {
...
+    const invisibleWidget = document.querySelector(`.${CAPTCHA_INVISIBLE_CLASSNAME}`);
+    if (invisibleWidget) {
+      document.removeChild(invisibleWidget);
+    }
+    const visibleWidget = document.getElementById(CAPTCHA_ELEMENT_ID);
+    if (visibleWidget) {
+      visibleWidget.style.display = 'none';
     }
   }
+  console.log({ captchaToken, captchaWidgetTypeUsed: widgetTypeUsed.join(',') });
```
anagstef, line 239: Do we need this `console.log`?

packages/clerk-js/src/utils/captcha/turnstile.ts
```diff
@@ -209,14 +225,18 @@ export const getTunstileToken = async (captchaOptions: {
...
 
-  return { captchaToken, captchaWidgetTypeUsed: isInvisibleWidget ? 'invisible' : 'smart' };
+  return { captchaToken, captchaWidgetTypeUsed: widgetTypeUsed.join(',') };
```
anagstef, line 241: Echoing here again that I'm pretty sure that backend expects only one of the two values (`invisible/smart`) for the `captchaWidgetTypeUsed` param.

anagstef (changes requested):
Great job! Just some minor fixes and it's good to go! 🚀

packages/clerk-js/src/utils/captcha/turnstile.ts
nikosdouvlis: I think the BE ignores this for now but I'll check again, in any case we need these for analytics so the BE will need to change 👍🏻

packages/clerk-js/src/utils/captcha/turnstile.ts
nikosdouvlis: Good catch! ✔️

packages/clerk-js/src/utils/captcha/turnstile.ts
nikosdouvlis: We want to know if the challenge got escalated from invisible/non-interactive to interactive

packages/clerk-js/src/core/resources/Session.ts
nikosdouvlis: I'm going to refactor this logic out once we've verified that bot protection works as expected

packages/clerk-js/src/utils/captcha/turnstile.ts
anagstef: Until now, we used this parameter to hint backend in order to know which of the two site-keys was actually used, so that it can use the respective one in the backend to verify the token with Cloudflare.

packages/clerk-js/src/utils/captcha/turnstile.ts
anagstef: We can introduce a new parameter if we want analytics on the BE.

packages/clerk-js/src/utils/captcha/turnstile.ts
```diff
@@ -112,59 +117,72 @@ async function loadCaptchaFromFAPIProxiedURL(fallbackUrl: string) {
...
+  let id = '';
   let turnstileSiteKey = siteKey;
+  let retries = 0;
+  let widgetContainerQuerySelector: string | undefined;
+  // The backend uses this to determine which turnstile to used in order to verify the token
+  let captchaWidgetTypeUsed: CaptchaWidgetType = null;
+
+  // modal
+  if (modalContainerQuerySelector && modalWrapperQuerySelector) {
+    captchaWidgetTypeUsed = 'smart';
```
anagstef, line 136: ❗  How are we making sure that this is `smart` here? If the `siteKey` coming from dashboard is `invisible`, then it will use the invisible captcha here!!

packages/clerk-js/src/utils/captcha/turnstile.ts
nikosdouvlis: We say that the BE can get the strategy used from CF while verifying the token so I'm reverting this change

packages/clerk-js/src/utils/captcha/turnstile.ts
```diff
@@ -209,14 +230,17 @@ export const getTunstileToken = async (captchaOptions: {
...
+      document.body.removeChild(invisibleWidget);
+    }
+    const visibleWidget = document.getElementById(CAPTCHA_ELEMENT_ID);
+    if (visibleWidget) {
+      visibleWidget.style.display = 'none';
     }
   }
 
-  return { captchaToken, captchaWidgetTypeUsed: isInvisibleWidget ? 'invisible' : 'smart' };
+  return { captchaToken, captchaWidgetTypeUsed: captchaWidgetTypeUsed };
```
anagstef, line 245: 🤷 
```suggestion
  return { captchaToken, captchaWidgetTypeUsed };
```

anagstef (approved):

PR #4564, opened 2024-11-14
[WIP] chore(shared): Update signature of `useReverification` for error handling configuration

panteliselef:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/src/react/hooks/useReverification.ts
+++ b/packages/shared/src/react/hooks/useReverification.ts
@@ -2,7 +2,7 @@ import type { Clerk } from '@clerk/types';
 import { useMemo, useRef } from 'react';
 
 import { __experimental_isReverificationHint, __experimental_reverificationMismatch } from '../../authorization-errors';
-import { ClerkRuntimeError, isClerkAPIResponseError } from '../../error';
+import { ClerkRuntimeError, isClerkAPIResponseError, isClerkRuntimeError } from '../../error';
 import { createDeferredPromise } from '../../utils/createDeferredPromise';
 import { useClerk } from './useClerk';
 import { useSafeLayoutEffect } from './useSafeLayoutEffect';
@@ -28,14 +28,21 @@ async function resolveResult<T>(
     });
 }
 
+type ExcludeClerkError<T, P> = T extends { clerk_error: any } ? (P extends { throwOnCancel: true } ? never : null) : T;
+
 type UseReverificationOptions = {
-  onCancel?: (() => void) | 'throw';
+  onCancel?: () => void;
+  throwOnCancel?: boolean;
 };
 
 function createReverificationHandler(
   params: { onOpenModal: Clerk['__experimental_openUserVerification'] } & UseReverificationOptions,
 ) {
-  function assertReverification<Fetcher extends () => Promise<any>>(fetcher: Fetcher): Fetcher {
+  function assertReverification<Fetcher extends (...args: any[]) => Promise<any>>(
+    fetcher: Fetcher,
+  ): (
+    ...args: Parameters<Fetcher>
+  ) => Promise<ExcludeClerkError<Awaited<ReturnType<Fetcher>>, Parameters<Fetcher>[1]>> {
     return (async (...args) => {
       let result = await resolveResult(fetcher(...args));
 
@@ -68,12 +75,14 @@ function createReverificationHandler(
            */
           await resolvers.promise;
         } catch (e) {
-          if (params.onCancel === 'throw') {
+          if (params.onCancel) {
+            params.onCancel();
+          }
+
+          if (isClerkRuntimeError(e) && e.code === 'reverification_cancelled' && params.throwOnCancel) {
             throw e;
           }
-          params?.onCancel?.();
 
-          // Is this even right ?
           return null;
         }
 
@@ -90,10 +99,13 @@ function createReverificationHandler(
   return assertReverification;
 }
 
-function __experimental_useReverification<Fetcher extends () => Promise<any>>(
+function __experimental_useReverification<
+  Fetcher extends (...args: any[]) => Promise<any>,
+  O extends UseReverificationOptions,
+>(
   fetcher: Fetcher,
-  options?: UseReverificationOptions,
-): readonly [Fetcher] {
+  options?: O,
+): readonly [(...args: Parameters<Fetcher>) => Promise<ExcludeClerkError<Awaited<ReturnType<Fetcher>>, O>>] {
   const { __experimental_openUserVerification } = useClerk();
   const fetcherRef = useRef(fetcher);
   const optionsRef = useRef(options);
--- a/packages/shared/src/react/hooks/useReverification.ts
+++ b/packages/shared/src/react/hooks/useReverification.ts
@@ -28,7 +28,13 @@ async function resolveResult<T>(
     });
 }
 
-function createReverificationHandler(params: { onOpenModal: Clerk['__experimental_openUserVerification'] }) {
+type UseReverificationOptions = {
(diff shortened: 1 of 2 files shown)


packages/shared/src/react/hooks/useReverification.ts
```diff
@@ -74,16 +103,24 @@ function createReverificationHandler(params: { onOpenModal: Clerk['__experimenta
...
+  const optionsRef = useRef(options);
 
   const handleReverification = useMemo(() => {
     const handler = createReverificationHandler({
       onOpenModal: __experimental_openUserVerification,
+      ...options,
     })(fetcherRef.current);
     return [handler] as const;
-  }, [__experimental_openUserVerification, fetcherRef.current]);
+  }, [__experimental_openUserVerification, fetcherRef.current, optionsRef.current]);
```
BRKalow, line 123: I don't think it's necessary to put ref values in the dependency array, as they are mutable outside of the react lifecycle. When they change they wouldn't trigger a re-execution of the memoized value here.

packages/shared/src/react/hooks/useReverification.ts
```diff
@@ -74,16 +103,24 @@ function createReverificationHandler(params: { onOpenModal: Clerk['__experimenta
...
 }
 
-function __experimental_useReverification<Fetcher extends () => Promise<any>>(fetcher: Fetcher): readonly [Fetcher] {
+function __experimental_useReverification<
+  Fetcher extends (...args: any[]) => Promise<any>,
+  O extends UseReverificationOptions,
+>(
+  fetcher: Fetcher,
+  options?: O,
+): readonly [(...args: Parameters<Fetcher>) => Promise<ExcludeClerkError<Awaited<ReturnType<Fetcher>>, O>>] {
```
BRKalow, line 112: 👀  let's use more readable names:

```suggestion
  Options extends UseReverificationOptions,
>(
  fetcher: Fetcher,
  options?: Options,
): readonly [(...args: Parameters<Fetcher>) => Promise<ExcludeClerkError<Awaited<ReturnType<Fetcher>>, Options>>] {
```

packages/shared/src/react/hooks/useReverification.ts
```diff
@@ -28,8 +29,21 @@ async function resolveResult<T>(
...
-  function assertReverification<Fetcher extends () => Promise<any>>(fetcher: Fetcher): Fetcher {
+type ExcludeClerkError<T, P> = T extends { clerk_error: any } ? (P extends { throwOnCancel: true } ? never : null) : T;
+
+type UseReverificationOptions = {
+  onCancel?: () => void;
+  throwOnCancel?: boolean;
+};
+
+function createReverificationHandler(
+  params: { onOpenModal: Clerk['__experimental_openUserVerification'] } & UseReverificationOptions,
```
BRKalow, line 40: Can we include `onOpenModal` in the above options type?

packages/shared/src/react/hooks/useReverification.ts
```diff
@@ -74,16 +103,24 @@ function createReverificationHandler(params: { onOpenModal: Clerk['__experimenta
   return assertReverification;
 }
 
-function __experimental_useReverification<Fetcher extends () => Promise<any>>(fetcher: Fetcher): readonly [Fetcher] {
+function __experimental_useReverification<
```
BRKalow, line 106: Is this the public export? let's add a JSDoc comment if so.

packages/react/src/hooks/__tests__/useReverification.test.ts
BRKalow, line 1: Should these tests live in shared? We're only testing the exports from the shared package 👀

packages/shared/src/react/hooks/useReverification.ts
panteliselef: this should not be configurable from the developers, maybe if i rename to `openModal` is more clear. Basically we do `onOpenModal: __experimental_openUserVerification,` which is an internal concern.

packages/react/src/hooks/__tests__/useReverification.test.ts
panteliselef: i moved these to shared

packages/shared/src/react/hooks/useReverification.ts
```diff
@@ -68,26 +98,55 @@ function createReverificationHandler(params: { onOpenModal: Clerk['__experimenta
...
 
   const handleReverification = useMemo(() => {
     const handler = createReverificationHandler({
-      onOpenModal: __experimental_openUserVerification,
+      openUIComponent: __experimental_openUserVerification,
+      ...optionsRef.current,
     })(fetcherRef.current);
     return [handler] as const;
-  }, [__experimental_openUserVerification, fetcherRef.current]);
+  }, [__experimental_openUserVerification, fetcherRef.current, optionsRef.current]);
```
panteliselef, line 144: Adding the references in here, will not cause the `useMemo` to re-render explicitly, but if something else causes the Component that the hook is mounted on, this will use the new references if they changed

packages/shared/src/react/hooks/useReverification.ts
```diff
@@ -68,26 +98,55 @@ function createReverificationHandler(params: { onOpenModal: Clerk['__experimenta
...
+ * @example
+ * A simple example:
+ *
+ * function Hello() {
+ *   const [fetchBalance] = useReverification(()=> fetch('/transfer-balance',{method:"POST"}));
+ *   return <button onClick={fetchBalance}>...</button>
+ * }
+ */
+function __experimental_useReverification<
+  Fetcher extends (...args: any[]) => Promise<any> | undefined,
```
panteliselef, line 131: I've added `undefined` here, in order to support the following

```tsx
const {user} = useUser()
const [update] = useReverification(()=>user?.update({userName})
```

packages/shared/src/react/hooks/useReverification.ts
```diff
@@ -1,36 +1,51 @@
...
 import { __experimental_isReverificationHint, __experimental_reverificationError } from '../../authorization-errors';
-import { ClerkRuntimeError, isClerkAPIResponseError } from '../../error';
+import { ClerkRuntimeError, isClerkAPIResponseError, isClerkRuntimeError } from '../../error';
 import { createDeferredPromise } from '../../utils/createDeferredPromise';
 import { useClerk } from './useClerk';
 import { useSafeLayoutEffect } from './useSafeLayoutEffect';
 
 async function resolveResult<T>(
-  result: Promise<T>,
+  result: Promise<T> | T,
```
panteliselef, line 12: Accept the value itself, e.g. undefined instead of `Promise<undefined>`

panteliselef (commented):
For reviewers, do you think this is acceptable ?

```tsx
const [createEmailAddress] = useReverification(() => user?.createEmailAddress({ email: emailField.value }));

...

const onSubmit = async () => {
 await createEmailAddress()
}
```

or we should force this ?

```tsx
const [updatePasswordWithReverification] = useReverification(
    (user: UserResource, opts: Parameters<UserResource['updatePassword']>) => user.updatePassword(...opts),
  );
  
...
  
   
const onSubmit = async () => {
  try {
    const opts = {
      newPassword: passwordField.value,
      signOutOfOtherSessions: sessionsField.checked,
      currentPassword: user.passwordEnabled ? currentPasswordField.value : undefined,
    } satisfies Parameters<typeof user.updatePassword>[0];

    await updatePasswordWithReverification(user, [opts]);
    onSuccess();
  } catch (e) {
    handleError(e);
  }
};
```

BRKalow (approved):

PR #4583, opened 2024-11-15
chore(types,clerk-react,vue): Share hook return types

wobsoriano:
## Description

T

Resolves ECO-231

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/react/src/hooks/useAuth.ts
+++ b/packages/react/src/hooks/useAuth.ts
@@ -1,11 +1,5 @@
 import { createCheckAuthorization } from '@clerk/shared/authorization';
-import type {
-  ActJWTClaim,
-  CheckAuthorizationWithCustomPermissions,
-  GetToken,
-  OrganizationCustomRoleKey,
-  SignOut,
-} from '@clerk/types';
+import type { CheckAuthorizationWithCustomPermissions, GetToken, SignOut, UseAuthReturn } from '@clerk/types';
 import { useCallback, useEffect, useState } from 'react';
 
 import { useAuthContext } from '../contexts/AuthContext';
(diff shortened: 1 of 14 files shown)


panteliselef:
In same cases we might need to only edit only types for react and not vue. For example, adding experimental features, which  usually are only implemented for the react hooks. Do you think this PR will make that process harder ? If not this move is reasonable.

wobsoriano:
> In same cases we might need to only edit only types for react and not vue. For example, adding experimental features, which usually are only implemented for the react hooks. Do you think this PR will make that process harder ? If not this move is reasonable.

Good question! I think we have 2 approaches in case of experimental features:

1. Say an experimental property in the `UseSessionListReturn` hook for example:

```ts
export type UseSessionListReturn =
  | {
      isLoaded: false;
      sessions: undefined;
      setActive: undefined;
      __experimental_test: undefined
    }
  | {
      isLoaded: true;
      sessions: SessionResource[];
      setActive: SetActive;
      __experimental_test: true
    };
```

We can keep this in `@clerk/types`, then in `@clerk/vue` we can use the [utility type](https://github.com/clerk/javascript/blob/dd2191b8e26f09e13bc8c451ca67251fe1836c87/packages/types/src/utils.ts#L96) we have that omits without union flattening

```ts
import type { Without } from '@clerk/types'

type type VueUseSessionListReturn = Without<UseSessionListReturn, '__experimental'>
```

The only problem I can see with this is that `@clerk/vue` will also need to be updated when adding `@clerk/clerk-react` experimental features

2. If it can't be done with number 1, then probably rewrite the hook return type in `@clerk/clerk-react` for example 🤷

What do you think?

panteliselef:
Probably i'd just expands the types for react only

wobsoriano:
> Probably i'd just expands the types for react only

sounds good to me

panteliselef (approved):

PR #4589, opened 2024-11-15
feat(clerk-js): Add sandbox component showcase

dstaley:
## Description

This PR adds a minimal sandbox application that can render each of our top-level components from `clerk-js` inside a framework-less environment. This is done through Rspack's dev server, which supports HMR. You can run the sandbox with `pnpm dev:sandbox` while in the `packages/clerk-js` directory.

The goal is to provide a quick-and-easy way to test the functionality of a core UI component without involving other libraries like `@clerk/clerk-react` or frameworks like Next.js.

To make things as easy as possible to get started, the sandbox uses an instance preconfigured with multiple OAuth providers and Organization support.

Screen Shot 2024-11-15 at 13 45 58 PM

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


LekoArts (approved):

packages/clerk-js/sandbox/app.js
```diff
@@ -0,0 +1,87 @@
...
+  link.classList.remove('text-gray-300', 'hover:bg-gray-700', 'hover:text-white');
+  link.classList.add('bg-gray-900', 'text-white');
+}
+
+(async () => {
+  const route = window.location.pathname;
+  if (route in routes) {
+    const renderCurrentRoute = routes[route];
+    addCurrentRouteIndicator(route);
+    await Clerk.load();
```
alexcarpenter, line 84: I think we want to define the `signInUrl` and `signUpUrl` in the load method since we don't have ENV variable in use right? Seeing navigations to AP when going through the sign-in flows without it.

```suggestion
    await Clerk.load({
      signInUrl: '/sign-in',
      signUpUrl: '/sign-up',
    });
```

PR #4595, opened 2024-11-18
fix(clerk-js): Block /tokens requests until fraud detection completes

nikosdouvlis:
## Description
Block all /token requests until the captcha challenge is solved. 

For getToken():
- this ensures that if a getToken() triggers fraud protection, all subsequent getToken() invocations will wait for the captcha challenge to be complete AND the initial getToken() to resolve
- if multiple getToken() invocation are queued, they will all resolve at the same time using the value of the first getToken. Only a single request will fire because of the in-memory token cache

For getToken({ skipCache: true }):
- similar to above, but multiple requests will be fired once the first getTOken() resolves
- we will optimize this behavior if needed

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/fraudProtection.ts
+++ b/packages/clerk-js/src/core/fraudProtection.ts
@@ -1,0 +1,25 @@
+/**
+ * TODO: @nikos Move captcha and fraud detection logic to this class
+ */
+class FraudProtectionService {
+  private inflightRequest: Promise<unknown> | null = null;
+
+  public async execute<T extends () => Promise<any>>(cb: T): Promise<Awaited<ReturnType<T>>> {
+    if (this.inflightRequest) {
+      await this.inflightRequest;
+    }
+
+    const prom = cb();
+    this.inflightRequest = prom;
+    return prom.then(res => {
+      this.inflightRequest = null;
+      return res;
+    });
+  }
+
+  public blockUntilReady() {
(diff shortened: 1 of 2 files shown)


packages/clerk-js/src/core/fraudProtection.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * TODO: @nikos Move captcha and fraud detection logic to this class
+ */
+class FraudProtectionService {
+  private inflightRequest: Promise<unknown> | null = null;
+
+  public async execute<T extends () => Promise<any>>(cb: T): Promise<Awaited<ReturnType<T>>> {
+    if (this.inflightRequest) {
+      console.error('Clerk: This state should not be possible. Please report this to Clerk support.');
```
panteliselef, line 9: Why is that ? Doesn't it make sense to queue callback here ? 

Will this cause 2 captcha modals back to back ?

nikosdouvlis: This is mostly for debugging, as this state should never be possible based on the changes in Session.ts
 We always  block on  `await fraudProtection.blockUntilReady();` and only after do we call `execute`

anagstef (approved):

PR #4596, opened 2024-11-18
feat(types,clerkjs,backend): Add support for enterprise_sso strategy 

Nikpolik:
## Description

This PR adds support for the new `enterprise_sso` strategy for signing in and signing up. 

This strategy supersedes SAML to provide a single strategy as the entry point for Enterprise Single Sign On regardless of the underlying protocol used to authenticate the user. For now there are two new types of connections that are supported in addition to SAML, Custom OAuth and EASIE (multi-tenant OAuth).

This new strategy will is available from API version `2024-10-01` and after because the way the API responds for Sign In's with an email address that matches a **SAML connection** changes.

From this version and onwards the API will respond with a status of `needs_first_factor`, the only strategy that will be included in supported first factors is `enterprise_sso` and the email address that matched will be returned in the identifier field. This change was made to make a bit clear the next step required.

### Changes

1. Include `enterprise_sso` in responses and request types.
2. Update Sign In response handling to take in to account `enterprise_sso` in supported first factors.
3. Update Sign Up response handling to take in to account `enterprise_sso` in missing fields
4. Update the `SUPPORTED_API_VERSION` constant to `2024-10-01`

### API Version `2024-10-01`

Besides this change updating the version to `2024-10-01` will also enable the email's for Sign Ins from new clients feature.

## Checklist

- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated (TODO)

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/famous-experts-begin.md
+++ b/.changeset/famous-experts-begin.md
@@ -1,5 +1,4 @@
 ---
-'@clerk/clerk-js': patch
 '@clerk/types': patch
 ---
 
@@ -7,3 +6,8 @@ Add support for the new `enterprise_sso` strategy.
 
 This strategy supersedes SAML to provide a single strategy as the entry point for Enterprise Single Sign On regardless of the underlying protocol used to authenticate the user. 
 For now there are two new types of connections that are supported in addition to SAML, Custom OAuth and EASIE (multi-tenant OAuth).
+
+- Add a new user setting `enterpriseSSO`, this gets enabled when there is an active enterprise connection for an instance.
+- Add support for signing in / signing up with the new `enterprise_sso` strategy.
+- Deprecated `userSettings.saml` in favor of `enterprise_sso`.
+- Deprecated `saml` sign in strategy in favor of `enterprise_sso`.
--- a/.changeset/famous-experts-begin.md
+++ b/.changeset/famous-experts-begin.md
@@ -1,0 +1,9 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/types': patch
+---
+
+Add support for the new `enterprise_sso` strategy.
+
+This strategy supersedes SAML to provide a single strategy as the entry point for Enterprise Single Sign On regardless of the underlying protocol used to authenticate the user. 
+For now there are two new types of connections that are supported in addition to SAML, Custom OAuth and EASIE (multi-tenant OAuth).
--- a/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
+++ b/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
@@ -266,7 +266,7 @@ export function _SignInStart(): JSX.Element {
         return signInResource;
       }
       /**
-       * For EnterpriseSSO enabled instances, perform sign in with password only when it is allowed for the identified user.
+       * For instances with Enterprise SSO enabled, perform sign in with password only when it is allowed for the identified user.
        */
       const passwordField = fields.find(f => f.name === 'password')?.value;
       if (!passwordField || signInResource.supportedFirstFactors?.some(ff => ff.strategy === 'saml')) {
--- a/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
+++ b/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
@@ -257,7 +257,7 @@ export function _SignInStart(): JSX.Element {
     } as SignInCreateParams;
   };
 
-  const safePasswordSignInForSamlInstance = (
+  const safePasswordSignInForEnterpriseSSOInstance = (
     signInCreatePromise: Promise<SignInResource>,
     fields: Array<FormControlState<string>>,
   ) => {
@@ -269,7 +269,10 @@ export function _SignInStart(): JSX.Element {
        * For instances with Enterprise SSO enabled, perform sign in with password only when it is allowed for the identified user.
        */
       const passwordField = fields.find(f => f.name === 'password')?.value;
-      if (!passwordField || signInResource.supportedFirstFactors?.some(ff => ff.strategy === 'saml')) {
+      if (
+        !passwordField ||
+        signInResource.supportedFirstFactors?.some(ff => ff.strategy === 'saml' || ff.strategy === 'enterprise_sso')
+      ) {
         return signInResource;
       }
       return signInResource.attemptFirstFactor({ strategy: 'password', password: passwordField });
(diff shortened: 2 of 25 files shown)


LauraBeatris:
@Nikpolik https://github.com/clerk/javascript/blob/a97a830eae82e54d0d4217f74ff046e9e7ad145e/packages/clerk-js/src/ui/components/UserProfile/AccountPage.tsx#L23

Could we update `showEnterpriseAccounts` here to check for `userSettings.enteprise_sso.enabled`

packages/types/src/userSettings.ts
```diff
@@ -104,6 +108,7 @@ export interface UserSettingsJSON extends ClerkResourceJSON {
   social: OAuthProviders;
 
   saml: SamlSettings;
```
LauraBeatris, line 110: 🔧 We should mark `saml` as deprecated in favor of `enterprise_sso`:
```suggestion
  /**
  * @deprecated Use `enterprise_sso` instead
  */
  saml: SamlSettings;
```

packages/types/src/redirects.ts
```diff
@@ -66,17 +66,17 @@ export type AuthenticateWithRedirectParams = {
 
   /**
    * One of the supported OAuth providers you can use to authenticate with, eg 'oauth_google'.
-   * Or alternatively `saml`, to authenticate with SAML.
+   * Alternatively `saml` or `enteprise_sso`, to authenticate with Enterprise Single Sign On.
```
LauraBeatris, line 69:
```suggestion
   * Alternatively `saml` or `enterprise_sso`, to authenticate with Enterprise SSO.
```

packages/types/src/strategies.ts
```diff
@@ -13,6 +13,7 @@ export type BackupCodeStrategy = 'backup_code';
 export type ResetPasswordPhoneCodeStrategy = 'reset_password_phone_code';
 export type ResetPasswordEmailCodeStrategy = 'reset_password_email_code';
 export type CustomOAuthStrategy = `oauth_custom_${string}`;
+export type EnterpriseSSOStrategy = 'enterprise_sso';
```
LauraBeatris, line 16: 🔧 We should mark `SamlStrategy` as deprecated in favor of EnterpriseSSOStrategy:
```ts
/**
 * @deprecated Use `EnterpriseSSOStrategy` instead
 */
export type SamlStrategy = 'saml';
```

packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
```diff
@@ -262,11 +262,11 @@ export function _SignInStart(): JSX.Element {
     fields: Array<FormControlState<string>>,
   ) => {
     return signInCreatePromise.then(signInResource => {
-      if (!userSettings.saml.enabled) {
+      if (!userSettings.enterpriseSSO.enabled) {
         return signInResource;
       }
       /**
-       * For SAML enabled instances, perform sign in with password only when it is allowed for the identified user.
+       * For EnterpriseSSO enabled instances, perform sign in with password only when it is allowed for the identified user.
```
LauraBeatris, line 269:
```suggestion
       * For Enterprise SSO enabled instances, perform sign in with password only when it is allowed for the identified user.
```

packages/clerk-js/src/utils/completeSignUpFlow.ts
```diff
@@ -22,9 +22,9 @@ export const completeSignUpFlow = ({
   if (signUp.status === 'complete') {
     return handleComplete && handleComplete();
   } else if (signUp.status === 'missing_requirements') {
-    if (signUp.missingFields.some(mf => mf === 'saml')) {
+    if (signUp.missingFields.some(mf => mf === 'saml' || mf === 'enterprise_sso')) {
       return signUp.authenticateWithRedirect({
-        strategy: 'saml',
+        strategy: 'enterprise_sso',
```
LauraBeatris, line 27: ❓For backward compatibility, shouldn't we set `strategy` equivalent to the missing field? Essentially, `saml` or `enterprise_sso`
```suggestion
        strategy: mf,
```

.changeset/famous-experts-begin.md
```diff
@@ -0,0 +1,9 @@
+---
+'@clerk/clerk-js': patch
```
LauraBeatris, line 2: 🚫 We shouldn't have different entries for `clerk-js` (minor vs patch) 

Also, all of the changes on this PR should be placed on minor versions. The structure could be the following:
- Changeset for `@clerk/backend` giving more context around the version update.
- Changeset for `@clerk/types` & `@clerk/clerk-js` with adding support for `enterprise_sso` strategy and on the user settings. Give more context on when `enterprise_sso` resolves to `true` in comparison to `saml`.

LauraBeatris (changes requested):

packages/clerk-js/src/utils/completeSignUpFlow.ts
Nikpolik: Hmm, wanted to remove the usage of saml from the code base, a similar change was made to [sign in as well](https://github.com/clerk/javascript/blob/nikpolik%2Fuser-939-add-support-for-enterprise_sso-strategy-in-clerkjs/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx#L285-L288). 

The `enterprise_sso` strategy can be used regardless of the version so it seems ok? 🤔

packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
Nikpolik: Fixed! [f0338ef](https://github.com/clerk/javascript/pull/4596/commits/f0338ef8f04ec5c3a7417f587052c136d0dcabf7)

Nikpolik: Also updated the function name [faa0e3c](https://github.com/clerk/javascript/pull/4596/commits/faa0e3cb76a841e4e93f17d3747ac4c6e1030fd7)

packages/types/src/userSettings.ts
Nikpolik: Fixed [f9c2194](https://github.com/clerk/javascript/pull/4596/commits/f9c2194678113a36fdcbb84bb55042d170b9e230)

packages/types/src/strategies.ts
Nikpolik: Fixed [f9c2194](https://github.com/clerk/javascript/pull/4596/commits/f9c2194678113a36fdcbb84bb55042d170b9e230)

Thanks for catching this, we need to mention it in the API docs as well I'll update them!

LauraBeatris: Well reminded on the docs, we'll need to update it here:

https://github.com/clerk/clerk-docs/blob/3a68d1a415b9928f69b01afbb8936bcc719a877c/docs/custom-flows/saml-connections.mdx?plain=1#L47

.changeset/famous-experts-begin.md
LauraBeatris: Makes sense to have a separate changeset for each package. The versions look correct to me now 👍🏻

LauraBeatris (approved):
Just need to update the changeset but overall looks good to me 👍🏻

panteliselef (approved):

PR #4599, opened 2024-11-18
feat(repo): playground workspace: protocol

jacekradko:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


BRKalow:
@jacekradko does this include all the playgrounds as workspaces? Not sure if we want to do that...we don't want to have to install all of these deps by default.

jacekradko:
@BRKalow Yes, it would. I am just testing this out and seeing what impact if would have on install times. pnpm should mitigate most of the additional cost of additional packages getting install, especially if we lean into `catalog:` protocol to use as few versions of the same package as possible.  

I think there is a case to be made for these packages to be treated as part of the workspace as it makes DX within the monorepo much easier. No need for yalc-ing things around and automatically getting interlinked packages

PR #4600, opened 2024-11-18
test(nuxt, e2e): Create long-running app for Nuxt

wobsoriano:
## Description

This PR adds support for running E2E tests in Nuxt applications.

- Added integration template using default Nuxt project (`/integration/templates/nuxt-node`).
- Added Nuxt integration preset (`/integration/presets/nuxt.ts`).
- E2E tests showcasing usage of `@clerk/nuxt` server middleware and `clerkClient` as well as components and composables which is auto-imported from `@clerk/vue`

> [!IMPORTANT]  
> Nuxt has [auto-imports](https://nuxt.com/docs/guide/concepts/auto-imports) enabled by default, so if you're seeing components and functions without imports in the template, this is expected. The `@clerk/nuxt` module will automatically register its components and composables for auto-imports, making them available throughout your application without explicit import statements.

Resolves ECO-245

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


integration/templates/nuxt-node/middleware/auth.global.js
wobsoriano, line 1: This is a Nuxt [global route middleware](https://nuxt.com/docs/guide/directory-structure/middleware).

Route middlewares are used to protect pages, on initial load and client-side routing.

integration/templates/nuxt-node/nuxt.config.js
```diff
@@ -0,0 +1,15 @@
...
+  runtimeConfig: {
+    public: {
+      clerk: {
+        publishableKey: process.env.NUXT_PUBLIC_CLERK_PUBLISHABLE_KEY,
+      },
+    },
+    clerk: {
+      secretKey: process.env.CLERK_SECRET_KEY,
+    },
+  },
```
wobsoriano, line 14: For some reason, E2E tests are not running without manually declaring the vars here. Will investigate more but this only happens in our test environment

integration/templates/nuxt-node/pages/user.vue
```diff
@@ -0,0 +1,11 @@
+<script setup>
+const { data: user } = await useFetch('/api/me');
```
wobsoriano, line 2: [`useFetch`](https://nuxt.com/docs/api/composables/use-fetch) makes API calls on the server during initial page load, and on the client during subsequent navigation

octoper (approved):

PR #4608, opened 2024-11-19
feat(clerk-js): Emit sandbox to dist on build

dstaley:
## Description

This PR updates the sandbox in `clerk-js` to support publishing the `dist` folder to Vercel previews. This will allow us to have previews of our clerk-js components for every PR that updates clerk-js.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/package.json
+++ b/packages/clerk-js/package.json
@@ -27,7 +27,9 @@
   "types": "dist/types/index.d.ts",
   "files": [
     "dist",
-    "headless"
+    "headless",
+    "!dist/index.html",
+    "!dist/sandbox.js"
   ],
   "scripts": {
     "build": "pnpm build:bundle && pnpm build:declarations",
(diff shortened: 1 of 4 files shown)


packages/clerk-js/package.json
```diff
@@ -27,7 +27,9 @@
   "types": "dist/types/index.d.ts",
   "files": [
     "dist",
-    "headless"
+    "headless",
+    "!dist/index.html",
```
dstaley, line 31: This ensures that we don't publish these files to npm.

BRKalow: can we omit to `/dist/preview/` to ensure we don't accidentally add additional files here and publish them?

```
/dist/preview/index.html
/dist/preview/sandbox.js
```

dstaley: I think that'd be complicated given how Rspack makes decisions based on the emit path. As an alternative, how about we only emit these assets when we run `pnpm build:sandbox`?

dstaley: yeah doing that feels better [17d1d90](https://github.com/clerk/javascript/pull/4608/commits/17d1d906aad4f296e6d7d1805c471d72e64542e4)

packages/clerk-js/turbo.json
```diff
@@ -22,6 +22,10 @@
         "!node_modules/**"
       ]
     },
+    "build:sandbox": {
+      "dependsOn": ["^build"],
+      "inputs": ["sandbox/**"]
```
jacekradko, line 27: We will also want to configure `outputs` to cache assets produced by this new task

PR #4611, opened 2024-11-20
fix(astro): Add components to turborepo cache

jacekradko:
## Description

We weren't caching the `components` folder in our turborepo cache even though it gets created during `build`. Added an override to `astro` turborepo config to address the issue

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.changeset/orange-radios-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/astro': patch
+---
+
+Fixed turborepo cache to correctly store components directory
```
panteliselef, line 5: @jacekradko I believe this is internal, so we don't need to notify developers about this change. We can have an empty changesets file

jacekradko: I figured it would be good to force a release of `@clerk/astro` in case we published a version with missing `@clerk/astro/components`, but I can take it out

jacekradko: Looks like latest has it, so I will empty out the changeset

jacekradko: Screenshot 2024-11-20 at 10 08 50 AM

panteliselef: Hmm, shall we do this then ?
```suggestion
---
'@clerk/astro': patch
---

Fix possible missing exported files from `@clerk/astro/components`
```

tmilewski (approved):

panteliselef (approved):

wobsoriano (approved):
Thanks!

PR #4619, opened 2024-11-21
fix(repo): Use patched yalc with support for catalog:

jacekradko:
## Description

Current version of `yalc` does not support the `catalog:` protocol. Switching to use a patched version of `yalc` in the monorepo which supports this new functionality.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


LekoArts:
To play devil's advocate here: If we want to dogfood and use https://github.com/clerk/javascript/tree/main/packages/dev-cli more we could also just remove yalc altogether

jacekradko:
> To play devil's advocate here: If we want to dogfood and use [main/packages/dev-cli](https://github.com/clerk/javascript/tree/main/packages/dev-cli) more we could also just remove yalc altogether

@LekoArts I am definitely on board with moving away from `yalc`. We can put that on our backlog.  

I am just trying to address the immediate issue with `yalc` now that doesn't work properly

LekoArts (approved):

PR #4620, opened 2024-11-21
fix(clerk-js): Enable captcha for development instances

issuedat:
## Description

Enables captcha for development instances.

Previously captcha was only usable on production instances and with recent changes, we now allow development instances running on localhost or other domains to also use captcha for parity with production.

We now rely on the value in the environment to determine if captcha is enabled or not (`userSettings.signUp.captcha_enabled`). For current development instances, it is disabled by default so this will be a no-op.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/utils/captcha/retrieveCaptchaInfo.ts
+++ b/packages/clerk-js/src/utils/captcha/retrieveCaptchaInfo.ts
@@ -11,11 +11,7 @@ export const retrieveCaptchaInfo = (clerk: Clerk) => {
     captchaWidgetType: _environment ? _environment.displayConfig.captchaWidgetType : null,
     captchaProvider,
     captchaPublicKeyInvisible: _environment ? _environment.displayConfig.captchaPublicKeyInvisible : null,
-    canUseCaptcha: _environment
-      ? _environment.userSettings.signUp.captcha_enabled &&
-        clerk.isStandardBrowser &&
-        clerk.instanceType === 'production'
-      : null,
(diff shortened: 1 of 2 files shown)


packages/clerk-js/src/utils/captcha/retrieveCaptchaInfo.ts
```diff
@@ -11,11 +11,7 @@ export const retrieveCaptchaInfo = (clerk: Clerk) => {
     captchaWidgetType: _environment ? _environment.displayConfig.captchaWidgetType : null,
     captchaProvider,
     captchaPublicKeyInvisible: _environment ? _environment.displayConfig.captchaPublicKeyInvisible : null,
-    canUseCaptcha: _environment
-      ? _environment.userSettings.signUp.captcha_enabled &&
-        clerk.isStandardBrowser &&
-        clerk.instanceType === 'production'
-      : null,
+    canUseCaptcha: _environment ? _environment.userSettings.signUp.captcha_enabled && clerk.isStandardBrowser : null,
```
panteliselef, line 14: @issuedat Can you update the PR description with an explanation on this one ?
I remember that this was done on purpose in the first iteration, can you give context on has changed ?

issuedat: I've updated the description with some more context! Let me know if that helps ☺️

panteliselef (approved):
lgtm

PR #4649, opened 2024-11-25
chore(clerk-js): Bundle Accountless only in dev

panteliselef:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


tmilewski (approved):

tmilewski (changes requested):
@panteliselef Following-up here, per our DM:

This all looks cool, but I'd suggest that we use specific flags vs `__DEV__` directly.

It'll make it easier to follow over time, and if/when we need to remove it.

Seeing as we only need it in the dev environment, the flag could follow a format such as `__BUILD_FLAG_ACCOUNTLESS__`, and rely on the same conditions as `__DEV__`.

/cc @BRKalow

tmilewski (approved):

PR #4691, opened 2024-12-02
chore(clerk-js,types): Add experimental `SignInCombinedProps` option

alexcarpenter:
## Description

Resolves SDKI-751

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/contexts/components/SignIn.ts
+++ b/packages/clerk-js/src/ui/contexts/components/SignIn.ts
@@ -37,7 +37,8 @@ export const useSignInContext = (): SignInContextType => {
     throw new Error(`Clerk: useSignInContext called outside of the mounted SignIn component.`);
   }
 
-  const { componentName, ...ctx } = context;
+  const { componentName, ..._ctx } = context;
+  const ctx = _ctx.__experimental?.combinedProps || _ctx;
 
   const initialValuesFromQueryParams = useMemo(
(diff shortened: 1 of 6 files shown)


packages/clerk-js/src/ui/contexts/components/SignIn.ts
```diff
@@ -37,7 +37,8 @@ export const useSignInContext = (): SignInContextType => {
     throw new Error(`Clerk: useSignInContext called outside of the mounted SignIn component.`);
   }
 
-  const { componentName, ...ctx } = context;
+  const { componentName, ..._ctx } = context;
+  const ctx = _ctx.__experimental?.combinedProps || _ctx;
```
BRKalow, line 41: Should these lines be reversed?
```suggestion
  const ctx = _ctx.__experimental?.combinedProps || _ctx;
  const { componentName, ..._ctx } = context;
```

alexcarpenter: I don't think so, `componentName` only lives on the `SignInCtx` so we pull that off from the original `context` and the use the rest of the props as the context values below based on the usage of `combinedProps`.

BRKalow: Ahh I missed the `_` 😅

BRKalow (approved):

PR #4692, opened 2024-12-02
ci(repo): Use `pull_request_target` event by triggering actor permissions

LauraBeatris:
## Description

https://github.com/clerk/javascript/pull/4702 needs to be merged first

This PR ensures that when a new untrusted PR is submitted, it verifies the triggering actor permissions, to proceed with the next steps that rely on repository secrets. 

#### Context

Currently, CI steps that rely on secrets get skipped once it detects that it's an external fork. It's also using `pull_request` event which doesn't provide access to secrets. This is bringing a ton of friction to merge external contributions.

#### Security considerations with `pull_request_target`

Any automated processing of PRs from an external fork is potentially dangerous and such PRs should be treated like untrusted input. Malicious changes could be such as: 
- make or powershell files or redefine the build script in the package.json file 
- npm packages may have custom preinstall and postinstall scripts, so running npm install would already trigger any malicious code if the attackers added a new package reference

With the `pull_request` event, as per the [documentation](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request), with the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository.

On the other hand, workflows triggered via `pull_request_target` have write permission to the target repository. They also have access to target repository secrets. 

#### `pull_request_target` + checking trigger actor permissions

GitHub Actions provides a triggering actor field that lets you know who ran (or re-ran) a workflow. This allows maintainers to check the changes manually, and proceed by re-running the jobs.

The workflow is:
1. Check permissions
2. Checkout code
3. Run CI with secrets

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.github/workflows/ci.yml
```diff
@@ -2,13 +2,9 @@ name: CI
 
 on:
   workflow_dispatch:
-    inputs:
-      run_integration_tests:
```
LauraBeatris, line 6: This wasn't being used at all, it was introduced in a previous PR with the attempt to continue using `pull_request` event with [workflow_run](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run)

anagstef (approved):

panteliselef (approved):

LauraBeatris:
I'm closing this one as it's impossible to apply a event target show via a PR, it has to be done directly on `main`

We already placed protections for GitHub actors that don't have permissions, so doing the above is safe.

Issue #4667, opened 2024-11-27
"tauri:" is not a valid protocol


ryansteil-pl:
### Preliminary Checks

- [X] I have reviewed the documentation: https://clerk.com/docs

- [X] I have searched for existing issues: https://github.com/clerk/javascript/issues

- [X] I have not already reached out to Clerk support via email or Discord (if you have, no need to open an issue here)

- [X] This issue is not a question, general help request, or anything other than a bug report directly related to Clerk. Please ask questions in our Discord community: https://clerk.com/discord.

### Reproduction

n/a

### Publishable key

n/a

### Description

I'm using Clerk from within a [tauri](https://tauri.app/) app.

When building for production, I'm getting the following warning in the developer console of the webview

`Clerk: "tauri:" is not a valid protocol. Redirecting to "/" instead. If you think this is a mistake, please open an issue.`

due to missing the protocol here

https://github.com/clerk/javascript/blob/b0d165aa940f1f87ca95045188b6627f3519cd7e/packages/clerk-js/src/utils/windowNavigate.ts#L3-L9

### Environment

```Shell
n/a
```

jacekradko:
@ryansteil-pl Thanks for filling this issue. We heard similar feedback from other users and will be releasing a fix in the near future that will allow you specify additional protocols that your app supports. For reference, here is the PR https://github.com/clerk/javascript/pull/4705

louis030195:
i implemented clerk in tauri here if useful

https://github.com/mediar-ai/screenpipe/tree/main/screenpipe-app-tauri

PR #4705, opened 2024-12-03
feat(clerk-js): Add allowedRedirectProtocols

BRKalow:
## Description

Introduce a new Clerk option, `allowedRedirectProtocols` that allows users to specify custom protocols to be allowed as part of user-provided redirect URLs. This is necessary for cases where deep-linking is used, in an Electron app for example.

As part of this, I've removed the protocol check from `windowNavigate`. This is an internal method that is used in only a few places, none of which should be accepting arbitrary, user-provided URLs. I've also added a JSDoc comment indicating `windowNavigate` should never be used when accepting user-provided URLs.

fixes SDKI-782

Closes: https://github.com/clerk/javascript/issues/4667

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/__tests__/clerk.test.ts
+++ b/packages/clerk-js/src/core/__tests__/clerk.test.ts
@@ -48,7 +48,7 @@ describe('Clerk singleton', () => {
   const productionPublishableKey = 'pk_live_Y2xlcmsuYWJjZWYuMTIzNDUucHJvZC5sY2xjbGVyay5jb20k';
 
   const mockNavigate = jest.fn((to: string) => Promise.resolve(to));
-  const mockedLoadOptions = { routerPush: mockNavigate, routerReplace: mockNavigate };
+  const mockedLoadOptions = { routerDebug: true, routerPush: mockNavigate, routerReplace: mockNavigate };
 
   const mockDisplayConfig = {
     signInUrl: 'http://test.host/sign-in',
@@ -686,7 +686,6 @@ describe('Clerk singleton', () => {
       const toUrl = 'https://www.origindifferent.com/';
       await sut.navigate(toUrl);
       expect(mockHref).toHaveBeenCalledWith(toUrl);
-      expect(logSpy).not.toHaveBeenCalled();
     });
 
     it('wraps custom navigate method in a promise if provided and it sync', async () => {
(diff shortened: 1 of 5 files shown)


packages/clerk-js/src/core/__tests__/clerk.test.ts
```diff
@@ -48,7 +48,7 @@ describe('Clerk singleton', () => {
   const productionPublishableKey = 'pk_live_Y2xlcmsuYWJjZWYuMTIzNDUucHJvZC5sY2xjbGVyay5jb20k';
 
   const mockNavigate = jest.fn((to: string) => Promise.resolve(to));
-  const mockedLoadOptions = { routerPush: mockNavigate, routerReplace: mockNavigate };
+  const mockedLoadOptions = { routerDebug: true, routerPush: mockNavigate, routerReplace: mockNavigate };
```
alexcarpenter, line 51: Is this addition intentional or just for local testing purposes?

alexcarpenter (approved):

packages/clerk-js/src/core/__tests__/clerk.test.ts
BRKalow: This is just for the unit tests 👍 The option already exists

packages/clerk-js/src/utils/windowNavigate.ts
```diff
@@ -8,16 +11,13 @@ export const ALLOWED_PROTOCOLS = [
   'chrome-extension:',
 ];
 
+/**
+ * Helper utility to navigate via window.location.href. Also dispatches a clerk:beforeunload custom event.
+ *
+ * Note that this utility should **never** be called with a user-provided URL. We make no specific checks against the contents of the URL here and assume it is safe. Use `Clerk.navigate()` instead for user-provided URLs.
+ */
```
panteliselef, line 18: Should we replace the usage for `windowNavigate` from SignIn.ts and SignUp.ts with `Clerk.navigate()` ?

panteliselef (approved):

packages/clerk-js/src/utils/windowNavigate.ts
BRKalow: I don't think we need to, as the url is provided from our API and it will always be external.

PR #4724, opened 2024-12-06
fix(repo): Fix failing integration tests

BRKalow:
## Description

- Updates our OTP test helper to have a longer delay between keystrokes. In a specific test, the prepare call was being fired after the input was submitted. Increasing the delay causes the prepare call to fire before the OTP is submitted.
- Also a few text selectors that were recently impacted by some localization string updates from an external contributor.
- Undoes a localization update that impacted the Delete account confirmation prompt

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] other: test fix


jacekradko (approved):

packages/localizations/src/en-US.ts
```diff
@@ -40,7 +40,7 @@ export const enUS: LocalizationResource = {
   formFieldHintText__optional: 'Optional',
   formFieldHintText__slug: 'A slug is a human-readable ID that must be unique. It’s often used in URLs.',
   formFieldInputPlaceholder__backupCode: 'Enter backup code',
-  formFieldInputPlaceholder__confirmDeletionUserAccount: 'Confirm account deletion',
+  formFieldInputPlaceholder__confirmDeletionUserAccount: 'Delete account',
```
BRKalow, line 43: This value is used for the confirmation prompt, needs to be "Delete account" 👀

PR #4734, opened 2024-12-09
fix(clerk-js): Always invoke an injected Web3 Wallet provider

chanioxaris:
## Description

If we weren't able to find the exact requested Web3 Wallet provider, then we should fallback and use the injected provider in order for the user to be able to continue the flow

### Before

https://github.com/user-attachments/assets/2951731a-867b-47ad-ae3b-3d18a8e5ed3c

### After

https://github.com/user-attachments/assets/02dd7ece-0bcc-44ab-a695-8e2558bf5635

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/nine-penguins-grab.md
+++ b/.changeset/nine-penguins-grab.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Always invoke an injected Web3 Wallet provider, if any, to allow the user to continue the flow rather than be a no-op
--- a/packages/clerk-js/src/utils/injectedWeb3Providers.ts
+++ b/packages/clerk-js/src/utils/injectedWeb3Providers.ts
@@ -53,13 +53,15 @@ class InjectedWeb3Providers {
   }
 
   get = (provider: InjectedWeb3Provider) => {
-    // In case there is a single injected provider, use that directly.
-    // This is needed in order to support other compatible Web3 Wallets (e.g. Rabby Wallet)
-    // and maintain the previous behavior
-    if (this.#providers.length === 1) {
-      return this.#providers[0].provider;
+    const ethProvider = this.#providers.find(p => p.info.name === this.#providerIdMap[provider])?.provider;
+    if (ethProvider !== undefined) {
+      return ethProvider;
     }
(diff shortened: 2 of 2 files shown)


packages/clerk-js/src/utils/injectedWeb3Providers.ts
```diff
@@ -53,13 +53,15 @@ class InjectedWeb3Providers {
...
+    const ethProvider = this.#providers.find(p => p.info.name === this.#providerIdMap[provider])?.provider;
+    if (ethProvider !== undefined) {
+      return ethProvider;
     }
-    return this.#providers.find(p => p.info.name === this.#providerIdMap[provider])?.provider;
+
+    // In case we weren't able to find the requested provider, fallback to the
+    // global injected provider instead, if any, to allow the user to continue
+    // the flow rather than be a no-op
+    return window.ethereum;
```
panteliselef, line 64: 1. Maybe i'm mistaken, but i think we were doing this and we removed it.
1. Is there a scenario where if want to request with metamask but the browser does not have it i should throw an error instead ? If yes, maybe the fallback needs to happen at the component level.

chanioxaris: 1. That's not what we were doing before. We were explicitly trying to find the requested provider (e.g. MetaMask, OKX Wallet etc) and if we couldn't, we weren't initiated the flow, giving the user the impression that the application is broken
2. Maybe we are looking to throw an error if there isn't a web3 provider available, but we are looking to investigate that in a different effort

.changeset/nine-penguins-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Always invoke an injected Web3 Wallet provider, if any, to allow the user to continue the flow rather than be a no-op
```
panteliselef, line 5:
```suggestion
Bug fix: When the requested web3 provider cannot be found use any other available injected Web3 Wallet provider, instead of blocking the sign-in/sign-up flow.
```

panteliselef (approved):

PR #4745, opened 2024-12-10
fix(astro): Prevent Netlify from doing a handshake redirect loop

wobsoriano:
## Description

Netlify caches the responses from pages. When the middleware initiates a handshake redirect and returns to the original URL, Netlify serves the cached response (along with the cookies), leading to an infinite handshake redirect loop.

I've tried using Netlify's [cache key variations](https://docs.netlify.com/platform/caching/#cache-key-variation) but it doesn't seem to work.

This PR adds a cache bust parameter to the original URL to prevent Netlify from serving the cached response. The parameter is automatically removed from the URL after the page loads.

I don't have e2e tests for it, but you can compare the ff.:

1. https://astro-clerk-template.netlify.app - one without fix, notice the redirect loop after signing in
2. https://deploy-preview-2--astro-clerk-template.netlify.app - the one with fix, should work as expected after signing in and handshake flow

Resolves ECO-202

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/astro/src/server/clerk-middleware.ts
+++ b/packages/astro/src/server/clerk-middleware.ts
@@ -243,9 +243,8 @@ Check if signInUrl is missing from your configuration or if it is not an absolut
  * in Netlify and Clerk's dev instance.
  */
 function handleNetlifyCacheInDevInstance(locationHeader: string, requestState: RequestState) {
-  const isHandshakeUrl = locationHeader.includes('/v1/client/handshake');
   const hasHandshakeQueryParam = locationHeader.includes('__clerk_handshake');
-  if (!isHandshakeUrl && !hasHandshakeQueryParam) {
+  if (!hasHandshakeQueryParam) {
     const url = new URL(locationHeader);
     url.searchParams.append(NETLIFY_CACHE_BUST_PARAM, Date.now().toString());
     requestState.headers.set('Location', url.toString());
--- a/packages/astro/src/server/clerk-middleware.ts
+++ b/packages/astro/src/server/clerk-middleware.ts
@@ -238,8 +238,9 @@ Check if signInUrl is missing from your configuration or if it is not an absolut
    PUBLIC_CLERK_IS_SATELLITE='true'`;
 
 /**
- * Adds a cache bust parameter to non-handshake redirects to prevent infinite redirects
- * in Netlify and Clerk's dev instance.
+ * Prevents infinite redirects in Netlify's functions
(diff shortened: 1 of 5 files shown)


packages/astro/src/server/clerk-middleware.ts
```diff
@@ -234,6 +237,27 @@ Check if signInUrl is missing from your configuration or if it is not an absolut
...
 
+/**
+ * Prevents infinite redirects in Netlify's functions
+ * by adding a cache bust parameter to the original redirect URL. This ensures Netlify
+ * doesn't serve a cached response during the authentication flow.
+ */
+function handleNetlifyCacheInDevInstance(locationHeader: string, requestState: RequestState) {
+  // Only run on Netlify environment
+  // eslint-disable-next-line turbo/no-undeclared-env-vars
+  if (!process.env.NETLIFY) {
```
wobsoriano, line 248: Do we have a way to check if Clerk is running in dev instance? This redirect loop issue happens only in Netlify and Clerk development instance

LekoArts: You can use these: https://github.com/clerk/javascript/blob/52a24eba0d71635cea3bc2f3380aa698907b63a0/packages/shared/src/keys.ts#L97-L111

wobsoriano: Perfect, thanks @LekoArts!

packages/astro/src/server/clerk-middleware.ts
```diff
@@ -234,6 +237,25 @@ Check if signInUrl is missing from your configuration or if it is not an absolut
...
 
+/**
+ * Prevents infinite redirects in Netlify's functions
+ * by adding a cache bust parameter to the original redirect URL. This ensures Netlify
+ * doesn't serve a cached response during the authentication flow.
+ */
+function handleNetlifyCacheInDevInstance(locationHeader: string, requestState: RequestState) {
+  // Only run on Netlify environment and Clerk development instance
+  // eslint-disable-next-line turbo/no-undeclared-env-vars
+  if (import.meta.env.NETLIFY && isDevelopmentFromPublishableKey(requestState.publishableKey)) {
```
wobsoriano, line 248: `import.meta.env.NETLIFY` is an env var provided by NETLIFY. You can test the result [here](https://deploy-preview-2--astro-clerk-template.netlify.app/api/me).

LekoArts (approved):

PR #4771, opened 2024-12-12
fix(clerk-js,localizations,types): Add min and max lengths to username field error

alexcarpenter:
## Description

Adds missing min_length and max_length values within `form_username_invalid_length` error.

Resolves SDK-2006

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/resources/UserSettings.ts
+++ b/packages/clerk-js/src/core/resources/UserSettings.ts
@@ -19,6 +19,9 @@ import { BaseResource } from './internal';
 const defaultMaxPasswordLength = 72;
 const defaultMinPasswordLength = 8;
 
+const defaultMinUsernameLength = 4;
+const defaultMaxUsernameLength = 64;
+
 export type Actions = {
   create_organization: boolean;
   delete_self: boolean;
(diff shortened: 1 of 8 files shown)


packages/clerk-js/src/core/resources/UserSettings.ts
```diff
@@ -84,6 +86,11 @@ export class UserSettings extends BaseResource implements UserSettingsResource {
           ? defaultMaxPasswordLength
           : Math.min(data?.password_settings?.max_length, defaultMaxPasswordLength),
     };
+    this.usernameSettings = {
+      ...data.username_settings,
+      min_length: Math.max(data?.username_settings?.min_length, 4),
+      max_length: Math.min(data?.username_settings?.max_length, 64),
```
panteliselef, line 92: Only thing i'm skeptical of is what happens if we decided to change the limits. For example if FAPI in the future allows for `max_length=65` characters it will require us to update the sdk.

panteliselef (approved):
Apart from the min/max comment, everything looks good

packages/clerk-js/src/core/resources/UserSettings.ts
alexcarpenter: Yeah, not sure the best route there, we have similar issue with the password defaults https://github.com/clerk/javascript/blob/f9059a4ab24b39bc10041d7c37fbc73127e28817/packages/clerk-js/src/core/resources/UserSettings.ts#L19-L20

panteliselef: since this is a pattern let's respect it

PR #4785, opened 2024-12-16
fix(shared): Improve error message when Publishable Key is missing

anagstef:
## Description

The shared function `parsePublishableKey` should throw a different error when the key is missing.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


tmilewski (approved):

LekoArts (approved):

packages/shared/src/keys.ts
```diff
@@ -39,7 +39,10 @@ export function parsePublishableKey(
   key = key || '';
 
   if (!key || !isPublishableKey(key)) {
-    if (options.fatal) {
+    if (options.fatal && !key) {
+      throw new Error('Publishable key is missing.');
```
BRKalow, line 43: Can we be more specific with the error here? What are remediation steps to address?

BRKalow (dismissed):

wobsoriano (approved):

packages/shared/src/keys.ts
```diff
@@ -39,7 +39,12 @@ export function parsePublishableKey(
   key = key || '';
 
   if (!key || !isPublishableKey(key)) {
-    if (options.fatal) {
+    if (options.fatal && !key) {
+      throw new Error(
+        'Publishable key is missing. Ensure that your publishable key is correctly configured. Make sure to double-check your environment configuration (.env file).',
```
octoper, line 44:
```suggestion
        (.env file).',
```

Not sure if we need to specify the file here as it can be a bit different depending on platform/framework etc.
example: Cloudflare Wrangler locally reads from `.dev.vars`

packages/shared/src/__tests__/keys.test.ts
```diff
@@ -58,6 +58,12 @@ describe('parsePublishableKey(key)', () => {
     expect(() => parsePublishableKey('fake_pk', { fatal: true })).toThrowError('Publishable key not valid.');
   });
 
+  it('throws an error if the publishable key is missing, when fatal: true', () => {
+    expect(() => parsePublishableKey(undefined, { fatal: true })).toThrowError(
+      'Publishable key is missing. Ensure that your publishable key is correctly configured. Make sure to double-check your environment configuration (.env file).',
```
BRKalow, line 63:
```suggestion
      'Publishable key is missing. Ensure that your publishable key is correctly configured. Double-check your environment configuration (e.g. your .env file) for your keys, or access them here: https://dashboard.clerk.com/last-active?path=api-keys',
```

anagstef: Thanks! Dropped the `(e.g. your .env file)` part, because it may confuse users on frameworks that have different env files.

packages/shared/src/keys.ts
anagstef: Removed it!

PR #4787, opened 2024-12-16
chore(nextjs): Improve experience when swapping keys on Keyless mode

panteliselef:
## Description

In Keyless mode it is expected that the user will eventually set keys as environment variables. When they do that, the nextjs server will restart, but there seems to be a mismatch because first the rendering lifecycle runs with the new keys and then the keys are updated in the middleware. This causes an error to be thrown to the user which we can actually improve.

**For the happy path**: we avoid displaying that error by retrying to decrypt data with the dummy key.

**For any other case**: when swapping keys or going back to headless you will now see this (see picture). this does not only reply to keyless.

image

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/young-beans-trade.md
+++ b/.changeset/young-beans-trade.md
@@ -1,0 +1,5 @@
+---
+'@clerk/nextjs': patch
+---
+
+Improve experience when swapping keys on Keyless mode.
--- a/.changeset/young-beans-trade.md
+++ b/.changeset/young-beans-trade.md
@@ -2,4 +2,4 @@
 '@clerk/nextjs': patch
 ---
 
-Improve experience when swapping keys on Keyless mode.
+Improves the error message when changing keys in development causes a subsequent request to fail.
--- a/packages/nextjs/src/server/errors.ts
+++ b/packages/nextjs/src/server/errors.ts
@@ -42,4 +42,4 @@ export const authSignatureInvalid = `Clerk: Unable to verify request, this usual
 
 export const encryptionKeyInvalid = `Clerk: Unable to decrypt request data, this usually means the encryption key is invalid. Ensure the encryption key is properly set. For more information, see: https://clerk.com/docs/references/nextjs/clerk-middleware#dynamic-keys. (code=encryption_key_invalid)`;
 
-export const encryptionKeyInvalidDev = `Clerk: Unable to decrypt request data.\n\nRefresh the page if your .env file was just updated. In the issue persists ensure the encryption key is valid and properly set.\n\nFor more information, see: https://clerk.com/docs/references/nextjs/clerk-middleware#dynamic-keys. (code=encryption_key_invalid)`;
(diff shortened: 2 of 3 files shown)


packages/nextjs/src/server/utils.ts
```diff
@@ -280,10 +287,34 @@ export function decryptClerkRequestData(
...
+     *
+     * Attempt one more time with the default dummy value.
+     */
+    if (canUseKeyless__server) {
+      try {
+        return decryptData(encryptedRequestData, KEYLESS_ENCRYPTION_KEY);
+      } catch (e) {
+        throwInvalidEncryptionKey();
+      }
+    }
```
panteliselef, line 304: Simple retrying with the dummy value

LauraBeatris (approved):

.changeset/young-beans-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/nextjs': patch
+---
+
+Improve experience when swapping keys on Keyless mode.
```
LauraBeatris, line 5: 💡 Should we mention here regarding the hot-swapped .env case? Just to clarify the "improve experience" part

packages/nextjs/src/server/errors.ts
```diff
@@ -41,3 +41,5 @@ ${verifyMessage}`;
 export const authSignatureInvalid = `Clerk: Unable to verify request, this usually means the Clerk middleware did not run. Ensure Clerk's middleware is properly integrated and matches the current route. For more information, see: https://clerk.com/docs/references/nextjs/clerk-middleware. (code=auth_signature_invalid)`;
 
 export const encryptionKeyInvalid = `Clerk: Unable to decrypt request data, this usually means the encryption key is invalid. Ensure the encryption key is properly set. For more information, see: https://clerk.com/docs/references/nextjs/clerk-middleware#dynamic-keys. (code=encryption_key_invalid)`;
+
+export const encryptionKeyInvalidDev = `Clerk: Unable to decrypt request data.\n\nRefresh the page if your .env file was just updated. In the issue persists ensure the encryption key is valid and properly set.\n\nFor more information, see: https://clerk.com/docs/references/nextjs/clerk-middleware#dynamic-keys. (code=encryption_key_invalid)`;
```
BRKalow, line 45:
```suggestion
export const encryptionKeyInvalidDev = `Clerk: Unable to decrypt request data.\n\nRefresh the page if your .env file was just updated. If the issue persists, ensure the encryption key is valid and properly set.\n\nFor more information, see: https://clerk.com/docs/references/nextjs/clerk-middleware#dynamic-keys. (code=encryption_key_invalid)`;
```

.changeset/young-beans-trade.md
panteliselef, line 5:
```suggestion
Improve error message when swapping keys defined as environment variables. Affects (code=encryption_key_invalid).
 **On Keyless mode**: Attempt to not throw the error.
```

panteliselef: @BRKalow @LauraBeatris is this better ?

BRKalow, line 5:
```suggestion
Improves the error message when changing keys in development causes a subsequent request to fail.
```

PR #4816, opened 2024-12-19
chore(backend,nextjs,types): Prevent system permissions usage in server-side

LauraBeatris:
## Description

Resolves ORGS-441

#### Context

System permissions (e.g., `org:sys_domains:manage`) are intentionally excluded from session claims to maintain reasonable JWT sizes. While these permissions work in client-side authorization checks (where they're validated against FAPI organization memberships), they cannot be verified server-side.

#### Problem

Despite documentation updates, developers continue to use server-side authorization checks with system permissions, leading to confusion and support tickets.

#### Solution

Add type-level validation to catch misuse of system permissions during development. I've opted not to introduce a runtime warning since developers might ignore it.

CleanShot 2024-12-19 at 14 02 48

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/src/server/__tests__/clerkMiddleware.test.ts
+++ b/packages/nextjs/src/server/__tests__/clerkMiddleware.test.ts
@@ -117,6 +117,10 @@ describe('ClerkMiddleware type tests', () => {
     clerkMiddlewareMock(async (auth, _event, _request) => {
       // @ts-expect-error - system permissions are not allowed
       (await auth()).has({ permission: 'org:sys_foo' });
+      // @ts-expect-error - system permissions are not allowed
+      await auth.protect(has => has({ permission: 'org:sys_foo' }));
+      // @ts-expect-error - system permissions are not allowed
+      await auth.protect({ permission: 'org:sys_foo' });
     });
   });
 
--- a/packages/nextjs/src/server/__tests__/clerkMiddleware.test.ts
+++ b/packages/nextjs/src/server/__tests__/clerkMiddleware.test.ts
@@ -113,6 +113,13 @@ describe('ClerkMiddleware type tests', () => {
     clerkMiddlewareMock();
   });
 
+  it('prevents usage of system permissions with auth.has()', () => {
+    clerkMiddlewareMock(async (auth, _event, _request) => {
+      // @ts-expect-error - system permissions are not allowed
+      (await auth()).has({ permission: 'org:sys_foo' });
+    });
+  });
+
   describe('Multi domain', () => {
     const defaultProps = { publishableKey: '', secretKey: '' };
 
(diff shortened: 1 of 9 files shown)


packages/nextjs/src/server/__tests__/clerkMiddleware.test.ts
```diff
@@ -113,6 +113,13 @@ describe('ClerkMiddleware type tests', () => {
     clerkMiddlewareMock();
   });
 
+  it('prevents usage of system permissions with auth.has()', () => {
+    clerkMiddlewareMock(async (auth, _event, _request) => {
+      // @ts-expect-error - system permissions are not allowed
+      (await auth()).has({ permission: 'org:sys_foo' });
+    });
+  });
+
```
panteliselef, line 122: do you think we should add a test case for `useAuth()` where the type says that sys permissions are allowed ?

In order to not break something in the future

packages/types/src/session.ts
```diff
@@ -25,6 +26,29 @@ import type {
 import type { TokenResource } from './token';
 import type { UserResource } from './user';
 
+type DisallowSystemPermissions<P extends string> = P extends `${OrganizationSystemPermissionPrefix}${string}`
+  ? 'System permissions are not included in session claims and cannot be used on the server-side'
+  : P;
```
BRKalow, line 31: This is really cool!

packages/nextjs/src/server/__tests__/clerkMiddleware.test.ts
LauraBeatris: Applied on https://github.com/clerk/javascript/pull/4816/commits/1da1b64b4db21806401d5d338a746d290acb725c

panteliselef:
This is great initiative. I opened [another PR](https://github.com/clerk/javascript/pull/4843/files#diff-11b10bff17e97625ed7578044dd5d8dfe0d5a57413aae77baff82a781b4473d7R76-R90) pointing to this one that also addresses usage with `protect()`.

I noticed that type checking does not work currently in our unit tests, so I would avoid merging this PR until that is resolved.

panteliselef:
🙃 Avoid merging until [this is merged](https://github.com/clerk/javascript/pull/4844)

jacekradko:
I am curious what our stance is on these type-only (build-time) checks when it comes to users that don't use Typescript is their projects? Is there a downside to also have a run-time check/warning?

panteliselef:
In this case it's mostly improved DX. You don't really care about DX if you are using JS in my opinion. But if you feel strongly a warning makes sense more than an error here. Since the utility will simply return false or redirect you.

LauraBeatris:
Also worth mentioning that this requirement might change in the future, when we deal with the wider problem of session claims bloating the browser cookie size

Most developers might not even notice the console warning when running on the server-side, although I see the point that they won't get the DX benefits of type-checks when using JavaScript only

panteliselef (approved):

PR #4821, opened 2024-12-20
fix(localizations): Fix delete account action for `nb-NO` localization

lasseklovstad:
## Description
Hi! There was no obvious way to delete the account in Norwegian because the translation was wrong compared to the `formFieldInputPlaceholder__confirmDeletionUserAccount`. SO this should fix it.

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


anagstef (approved):
thank you for this fix! 🙏

PR #4837, opened 2025-01-03
fix(vue): Fix `getToken()` hanging when Clerk is already loaded

wobsoriano:
## Description

This PR makes sure `getToken` from `useAuth()` composable returns a token by adding `immediate: true` to the watcher. Without this option, the internal `clerkLoaded` promise would never resolve if Clerk was already loaded.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/vue/src/composables/useAuth.ts
+++ b/packages/vue/src/composables/useAuth.ts
@@ -12,12 +12,15 @@ import { useClerkContext } from './useClerkContext';
  */
 function clerkLoaded(clerk: ShallowRef<Clerk | null>) {
   return new Promise<Clerk>(resolve => {
-    const unwatch = watch(clerk, value => {
-      if (value?.loaded) {
-        unwatch();
-        resolve(clerk.value!);
-      }
-    });
+    watch(
+      clerk,
+      value => {
+        if (value?.loaded) {
+          resolve(clerk.value!);
+        }
+      },
+      { immediate: true },
+    );
   });
 }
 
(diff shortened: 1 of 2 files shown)


packages/vue/src/composables/useAuth.ts
```diff
@@ -73,12 +76,12 @@ type UseAuth = () => ToComputedRefs<UseAuthReturn>;
 export const useAuth: UseAuth = () => {
   const { clerk, authCtx } = useClerkContext();
 
+  const getToken: GetToken = createGetToken(clerk);
+  const signOut: SignOut = createSignOut(clerk);
```
wobsoriano, line 80: Moved this out from the `computed` property since the value doesn't need to be recalculated

packages/vue/src/composables/useAuth.ts
```diff
@@ -12,12 +12,15 @@ import { useClerkContext } from './useClerkContext';
...
-      }
-    });
+    watch(
+      clerk,
+      value => {
+        if (value?.loaded) {
+          resolve(clerk.value!);
+        }
+      },
+      { immediate: true },
```
wobsoriano, line 22: We need to force the callback to be executed immediately to handle the case where clerk is already loaded when the watcher is created

wobsoriano: Also removed `unwatch` here since Vue will clean it up automatically when the component that uses `useAuth()` unmounts

bakatz:
Thanks for fixing this! (I'm the original reporter)

wobsoriano:
!snapshot

clerk-cookie:
Hey @wobsoriano - the snapshot version command generated the following package versions:
| Package | Version |
| --- | --- |
| @clerk/nuxt | 1.0.1-snapshot.v20250105004218 |
| @clerk/vue | 1.0.1-snapshot.v20250105004218 |

`@clerk/nuxt`
```sh
npm i @clerk/nuxt@1.0.1-snapshot.v20250105004218 --save-exact
```
`@clerk/vue`
```sh
npm i @clerk/vue@1.0.1-snapshot.v20250105004218 --save-exact
```

panteliselef (approved):

tmilewski (approved):

PR #4878, opened 2025-01-13
fix(localizations): Fix de-DE key `socialButtonsBlockButtonManyInView`

christiankaindl:
## Description

A previous change changed the label for buttons when multiple SSO providers are active from the provider name to a long string, which breaks the layout. This change brings the label for the key `socialButtonsBlockButtonManyInView` back to the desired one and in line with other languages.

The previous value roughly translates to "Too many buttons are shown. Reduce amount of buttons to continue"—this doesn't make any sense and is unique to the German translations. The broken string was introduced in this commit, together with other changes, so it seems it fell through the cracks: https://github.com/clerk/javascript/commit/b9a5bea404cb09d7e328d48b41bd2669feb4e518#diff-96883e89b065cc0abf16578b55e5a53c76a77b62d3ea953fa4e6064c29da93aaR550

Without the fix, the SignIn component with German translations looks like this:
Screenshot 2025-01-13 at 09 31 43

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/localizations/src/de-DE.ts
+++ b/packages/localizations/src/de-DE.ts
@@ -547,8 +547,7 @@ export const deDE: LocalizationResource = {
     },
   },
   socialButtonsBlockButton: 'Weiter mit {{provider|titleize}}',
-  socialButtonsBlockButtonManyInView:
-    'Zu viele Buttons angezeigt. Reduzieren Sie die Anzahl der Buttons, um fortzufahren.',
+  socialButtonsBlockButtonManyInView: '{{provider|titleize}}',
   unstable__errors: {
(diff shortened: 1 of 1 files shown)


panteliselef:
Hey @christiankaindl thanks for your contribution. This seems to be affecting more package so I opened #4887 to address that. Also the new PR will force the property to match a new pattern I order to avoid any similar issues in the future.

christiankaindl:
@panteliselef Since your PR got merged, I guess this one can be closed?

LekoArts:
Yes :) Thanks again for the PR and bringing up this issue, much appreciated!

PR #4879, opened 2025-01-13
fix(nextjs): Detect middlware usage to avoid infinite loop on Keyless mode

panteliselef:
## Description

When `clerkMiddleware` is not used in an application we know that any attempt to use `await auth()` will fail, so we don't need to force "middleware" to rerun in order for our server side helpers to have the correct keys. Accessing them will certainly fail either way.

In this scenario, we could have simply perform `detectClerkMiddleware()` inside the Server ClerkProvider, but we want to avoid doing that since that will opt in all routes into dynamic rendering implicitly which may annoy many developers.

### Necessary changes
Calling `buildRequestLike` inside a server action in Next.js will cause the application to break due to bundling issues of the framework that cause `fs` to leak into he client. For that reason, I've adjusted the implementation to work with the standard `Request` from WebAPI.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/src/server/headers-utils.ts
+++ b/packages/nextjs/src/server/headers-utils.ts
@@ -1,0 +1,53 @@
+import { constants } from '@clerk/backend/internal';
+import type { NextRequest } from 'next/server';
+
+import type { RequestLike } from './types';
+
+export function getCustomAttributeFromRequest(req: RequestLike, key: string): string | null | undefined {
+  // @ts-expect-error - TS doesn't like indexing into RequestLike
+  return key in req ? req[key] : undefined;
+}
+
+export function getAuthKeyFromRequest(
+  req: RequestLike,
+  key: keyof typeof constants.Attributes,
+): string | null | undefined {
+  return getCustomAttributeFromRequest(req, constants.Attributes[key]) || getHeader(req, constants.Headers[key]);
+}
+
+export function getHeader(req: RequestLike, name: string): string | null | undefined {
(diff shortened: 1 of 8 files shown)


LekoArts (approved):

packages/nextjs/src/server/headers-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export function isRequestWebAPI(val: unknown): val is Request {
+  try {
+    const { headers } = (val || {}) as Request;
+    return typeof headers?.get === 'function';
+  } catch (e) {
+    return false;
+  }
+}
```
panteliselef, line 53: This is the new part.

packages/nextjs/src/server/headers-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export function getAuthKeyFromRequest(
+  req: RequestLike,
+  key: keyof typeof constants.Attributes,
+): string | null | undefined {
+  return getCustomAttributeFromRequest(req, constants.Attributes[key]) || getHeader(req, constants.Headers[key]);
+}
+
+export function getHeader(req: RequestLike, name: string): string | null | undefined {
+  if (isNextRequest(req) || isRequestWebAPI(req)) {
```
panteliselef, line 19: Updated this line. All the rest are simply moved for tree shaking reasons.

PR #4884, opened 2025-01-13
feat(backend): Set cookies from the refresh token flow

brkalow:
## Description

Supports the `format=cookies` parameter for the session refresh endpoint. When this parameter is passed, the API will return an array of `set-cookie` header values to be passed along with the response.

SDKI-832

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/api/endpoints/SessionApi.ts
+++ b/packages/backend/src/api/endpoints/SessionApi.ts
@@ -67,7 +67,7 @@ export class SessionAPI extends AbstractAPI {
     });
   }
 
-  public async refreshSession(sessionId: string, params: RefreshTokenParams & { format: 'token ' }): Promise<Token>;
+  public async refreshSession(sessionId: string, params: RefreshTokenParams & { format: 'token' }): Promise<Token>;
   public async refreshSession(sessionId: string, params: RefreshTokenParams & { format: 'cookie' }): Promise<Cookies>;
   public async refreshSession(sessionId: string, params: RefreshTokenParams): Promise<Token>;
   public async refreshSession(sessionId: string, params: RefreshTokenParams): Promise<Token | Cookies> {
--- a/packages/backend/src/api/endpoints/SessionApi.ts
+++ b/packages/backend/src/api/endpoints/SessionApi.ts
@@ -22,7 +22,7 @@ type RefreshTokenParams = {
   request_originating_ip?: string;
   request_headers?: Record<string, string[]>;
   suffixed_cookies?: boolean;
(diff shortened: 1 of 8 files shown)


packages/backend/src/tokens/__tests__/request.test.ts
```diff
@@ -972,130 +972,138 @@ describe('tokens.authenticateRequest(options)', () => {
...
-    expect(refreshSession).not.toHaveBeenCalled();
-  });
-
-  test('refreshToken: uses suffixed refresh cookie even if un-suffixed is present', async () => {
-    server.use(
-      http.get('https://api.clerk.test/v1/jwks', () => {
-        return HttpResponse.json(mockJwks);
-      }),
-    );
+  describe('refreshToken', async () => {
```
BRKalow, line 975: I moved these tests into a separate describe block, consider looking at the diff without whitespace to see the actual changes: https://github.com/clerk/javascript/pull/4884/files?w=1

packages/backend/src/api/endpoints/SessionApi.ts
```diff
@@ -64,12 +67,17 @@ export class SessionAPI extends AbstractAPI {
     });
   }
 
-  public async refreshSession(sessionId: string, params: RefreshTokenParams) {
+  public async refreshSession(sessionId: string, params: RefreshTokenParams & { format: 'token ' }): Promise<Token>;
```
panteliselef, line 70:
```suggestion
  public async refreshSession(sessionId: string, params: RefreshTokenParams & { format: 'token' }): Promise<Token>;
```

packages/backend/src/tokens/request.ts
```diff
@@ -273,14 +273,16 @@ ${error.getFullMessage()}`,
 
     try {
       // Perform the actual token refresh.
-      const tokenResponse = await options.apiClient.sessions.refreshSession(decodeResult.payload.sid, {
+      const response = await options.apiClient.sessions.refreshSession(decodeResult.payload.sid, {
+        format: 'cookie',
```
panteliselef, line 277: Do we want this to be set for every call ? Could this be configured and passed down from `authenticateRequest` ?

brkalow: I don't see a good reason to make it configurable, not setting the cookie allows requests to fire with the expired session token before clerk-js hydrates.

panteliselef (approved):

anagstef (approved):

PR #4887, opened 2025-01-14
fix(localizations): Force interpolation for `socialButtonsBlockButtonManyInView`

panteliselef:
## Description

This PR expands on the fix and issue mentioned in #4878

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/localizations/src/be-BY.ts
+++ b/packages/localizations/src/be-BY.ts
@@ -547,7 +547,7 @@ export const beBY: LocalizationResource = {
     },
   },
   socialButtonsBlockButton: 'Працягнуць з дапамогай {{provider|titleize}}',
-  socialButtonsBlockButtonManyInView: '{{provider|titleize}}',
+  socialButtonsBlockButtonManyInView: undefined,
 
   unstable__errors: {
     already_a_member_in_organization: 'Вы ўжо з’яўляецеся членам гэтай арганізацыі.',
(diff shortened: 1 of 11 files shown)


packages/localizations/src/be-BY.ts
```diff
@@ -547,7 +547,7 @@ export const beBY: LocalizationResource = {
     },
   },
   socialButtonsBlockButton: 'Працягнуць з дапамогай {{provider|titleize}}',
-  socialButtonsBlockButtonManyInView: 'Увядзіце дадатковыя спосабы ўваходу',
+  socialButtonsBlockButtonManyInView: undefined,
```
anagstef, line 550: @panteliselef  Why don't we do the following for all localizations?
```suggestion
  socialButtonsBlockButtonManyInView: '{{provider|titleize}}',
```

panteliselef: @anagstef if left undefined it will use the one defined in `en-US` right ?

anagstef: Yes, but there is a risk that someone will try to translate this wrong in the future again.

panteliselef: i'm confused, but Typescript will not let them.

anagstef (approved):
💯

LekoArts (approved):

PR #4905, opened 2025-01-16
fix(clerk-expo): Trigger deep linking on SSO callback

LauraBeatris:
## Description

Fixes deep linking behavior on SSO callback from the authorization server

Currently, `WebBrowser.openAuthSessionAsync` doesn't trigger the deep linking once the authorization server responds with a 301 status to the URI scheme. 

[Expo issue](https://github.com/expo/expo/issues/34187)

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/expo/package.json
+++ b/packages/expo/package.json
@@ -79,6 +79,7 @@
     "@clerk/expo-passkeys": "workspace:*",
     "@types/base-64": "^1.0.2",
     "expo-auth-session": "^5.4.0",
+    "expo-linking": "^7.0.4",
     "expo-local-authentication": "^13.8.0",
     "expo-secure-store": "^12.8.1",
     "expo-web-browser": "^12.8.2",
@@ -87,6 +88,7 @@
   "peerDependencies": {
     "@clerk/expo-passkeys": ">=0.0.6",
     "expo-auth-session": ">=5",
+    "expo-linking": ">=7",
     "expo-local-authentication": ">=13.5.0",
     "expo-secure-store": ">=12.4.0",
     "expo-web-browser": ">=12.5.0",
--- a/packages/expo/package.json
+++ b/packages/expo/package.json
@@ -79,7 +79,7 @@
     "@clerk/expo-passkeys": "workspace:*",
     "@types/base-64": "^1.0.2",
     "expo-auth-session": "^5.4.0",
-    "expo-linking": "^7.0.4",
+    "expo-linking": "^6.2.2",
     "expo-local-authentication": "^13.8.0",
     "expo-secure-store": "^12.8.1",
     "expo-web-browser": "^12.8.2",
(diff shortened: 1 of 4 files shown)


packages/expo/package.json
```diff
@@ -87,6 +88,7 @@
   "peerDependencies": {
     "@clerk/expo-passkeys": ">=0.0.6",
     "expo-auth-session": ">=5",
+    "expo-linking": ">=7",
```
panteliselef, line 91: We need to be careful with this, I believe any other `expo-*` package we have as peer is compatible with >= Expo 50 and `expo-linking@7` seems too "new". 

Link.openUrl() existed for [Expo 50](https://docs.expo.dev/versions/v50.0.0/sdk/linking/), we just need to find the correct major. Based on [their changelog](https://github.com/expo/expo/blob/sdk-50/packages/expo-linking/CHANGELOG.md) for expo 50, we can do 
```suggestion
    "expo-linking": ">=6.2.2",
```

packages/expo/src/hooks/useOAuth.ts
```diff
@@ -101,6 +102,8 @@ export function useOAuth(useOAuthParams: UseOAuthFlowParams) {
       createdSessionId = signUp.createdSessionId || '';
     }
 
+    Linking.openURL(url);
```
LauraBeatris, line 105: We shouldn't trigger the deep linking if there are verification errors, missing factors, such as MFA

LauraBeatris: Ideally the `action_complete_redirect_url` would be nice here, cause then this would be the one FAPI would redirect to if the SignIn object is complete

LauraBeatris:
I thought more about the DX here, and triggering a deep link navigation on callback might break progressive flows 

Here's my current inclination: It's up the consumer to verify if the sign in got completed, if there's `createdSessionId` and then perform the redirection with Expo router

packages/expo/package.json
LauraBeatris: Great catch on this, thanks for the heads up!

packages/expo/src/hooks/useOAuth.ts
```diff
@@ -93,6 +97,7 @@ export function useOAuth(useOAuthParams: UseOAuthFlowParams) {
 
     if (status === 'complete') {
       createdSessionId = signIn.createdSessionId!;
+      await Linking.openURL(url);
```
LauraBeatris, line 100: Thinking a bit more, if we trigger the deep linking here, it can cause a race condition with `setActive`, where it'll navigate to a page that expects the user to be authenticated but it might take some milliseconds for it to happen, causing flickering on the UI 

The flow should be like the following:
- Create sign in, open web browser for `externalVerificationRedirectURL`
- Closes web browser once server responds with 301, deep links back to the current page
- Parse the rotating token nonce out of the URL
- Update sign-in, verify status, handle transfer flow if necessary
- Consumer page performs routing according to the sign-in status, this gives more freedom to deal with the progressive flow and it follows our current recommendation here: https://clerk.com/docs/custom-flows/error-handling

PR #4907, opened 2025-01-16
fix(clerk-js,types): Skip fraud protection if client has bypass enabled

nikosdouvlis:
## Description

If a client has captcha bypass enabled, then skip all FE captcha code

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -81,54 +81,25 @@ export class SignUp extends BaseResource implements SignUpResource {
...
-          }
-        }
+  create = async (_params: SignUpCreateParams): Promise<SignUpResource> => {
+    let params: Record<string, unknown> = _params;
+
+    if (!__BUILD_DISABLE_RHC__ && !this.clientBypass() && !this.shouldBypassCaptchaForAttempt(params)) {
+      const captchaChallenge = new CaptchaChallenge(SignUp.clerk);
+      const captchaParams = await captchaChallenge.managed();
+      if (!captchaParams) {
+        throw new ClerkRuntimeError('', { code: 'captcha_unavailable' });
```
anagstef, line 91: 🤔  Is this error getting set to the Card component? If yes, then we show no message now, while before we had the `e.message` here.

nikosdouvlis: Won't this get the message from the localization file instead? 
We don't know what e.message was and we don't want to show CF errors to the end users for sure.

anagstef: Yes, we have this code on the localization file

packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -81,54 +81,25 @@ export class SignUp extends BaseResource implements SignUpResource {
...
-          } else {
-            throw new ClerkRuntimeError(e.message, { code: 'captcha_unavailable' });
-          }
-        }
+  create = async (_params: SignUpCreateParams): Promise<SignUpResource> => {
+    let params: Record<string, unknown> = _params;
+
+    if (!__BUILD_DISABLE_RHC__ && !this.clientBypass() && !this.shouldBypassCaptchaForAttempt(params)) {
+      const captchaChallenge = new CaptchaChallenge(SignUp.clerk);
+      const captchaParams = await captchaChallenge.managed();
```
anagstef, line 89: Do we force managed here? Because it could be invisible, right?

packages/clerk-js/src/core/resources/SignUp.ts
anagstef: Thanks!

packages/clerk-js/src/core/resources/SignUp.ts
nikosdouvlis: `managed` will default to the settings used in env, `invisible` forces invisible

packages/clerk-js/src/utils/captcha/CaptchaChallenge.ts
```diff
@@ -39,7 +40,7 @@ export class CaptchaChallenge {
    * Important: For this to work at the moment, the instance needs to be using SMART protection
    * as we need both keys (visible and invisible) to be present.
    */
```
anagstef, line 42: Is this comment still relevant?

packages/clerk-js/src/core/resources/SignUp.ts
anagstef: The naming is a bit confusing here maybe. The managed somehow implies it uses the managed turnstile challenge.

packages/clerk-js/src/utils/captcha/CaptchaChallenge.ts
anagstef: But this is now used for everyone, both for instances that have invisible and smart selected. This comment indicates that this function will fail if SMART is not selected on the Clerk Dashboard.

packages/clerk-js/src/core/resources/SignUp.ts
anagstef: Correct, but this is the flow for instances that have selected the invisible setting also. I suggest we rename this function to something like `runChallenge()` or something that includes the invisible as a valid option and not only as a fallback.

anagstef (approved):

nikosdouvlis:
Merging as mailsac is failling

PR #4910, opened 2025-01-16
fix(clerk-js): Show error during instant password error

alexcarpenter:
## Description

When an password is incorrectly autofilled during sign-in, render an incorrect password error vs moving to first factor.

https://github.com/user-attachments/assets/48b4bc58-476f-49d7-815c-57e60b32f425

Fixes SDK-2029

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/integration/tests/sign-in-flow.test.ts
+++ b/integration/tests/sign-in-flow.test.ts
@@ -123,6 +123,15 @@ testAgainstRunningApps({ withEnv: [appConfigs.envs.withEmailCodes] })('sign in f
     await u.po.expect.toBeSignedOut();
   });
 
+  test('cannot sign in with email and wrong instant password', async ({ page, context }) => {
+    const u = createTestUtils({ app, page, context });
+    await u.po.signIn.goTo();
+    await u.po.signIn.signInWithEmailAndInstantPassword({ email: fakeUser.email, password: 'wrong-password' });
+    await expect(u.page.getByText(/password you entered is incorrect/i)).toBeVisible();
+
+    await u.po.expect.toBeSignedOut();
+  });
+
   test('cannot sign in with wrong password but can sign in with email', async ({ page, context }) => {
     const u = createTestUtils({ app, page, context });
 
--- a/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
+++ b/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
@@ -308,7 +308,6 @@ export function _SignInStart(): JSX.Element {
             await authenticateWithEnterpriseSSO();
             break;
           }
-          console.log(res);
           return navigate('factor-one');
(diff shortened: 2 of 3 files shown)


packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
```diff
@@ -308,6 +308,7 @@ export function _SignInStart(): JSX.Element {
             await authenticateWithEnterpriseSSO();
             break;
           }
+          console.log(res);
```
LekoArts, line 311: Debugging statement that's left over?

octoper (approved):

jacekradko (approved):

integration/tests/sign-in-flow.test.ts
```diff
@@ -123,6 +123,15 @@ testAgainstRunningApps({ withEnv: [appConfigs.envs.withEmailCodes] })('sign in f
     await u.po.expect.toBeSignedOut();
   });
 
+  test('cannot sign in with email and wrong instant password', async ({ page, context }) => {
```
brkalow, line 126: 👏

panteliselef (approved):

packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
```diff
@@ -355,7 +355,7 @@ export function _SignInStart(): JSX.Element {
     );
 
     if (instantPasswordError) {
-      await signInWithFields(identifierField);
```
brkalow, line 358: Interesting, so this was actually intentional 🤔 I'm wondering if we want to continue to move to the factor-one view but show an error, or if we want to keep them on the start view. What do you think?

brkalow (changes requested):

packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
panteliselef: What was the intention here ? Why do we think navigating them to another screen is better (even with your proposed fix) ?

brkalow: Likely to automatically give them another sign in option if one is available, without requiring additional action. If they have another auth option available they would now have to manually empty the password field and click continue. The start screen isn't setup to give them additional options (e.g. no "forgot password" link). It's possible someone gets stuck in this error case.

panteliselef: Sounds like a state that you can find yourself if your password manager is not up to date and still tries to use a password that you had before. Yeah I get it now. Let's add a comment in that line that explains the intent, i this this will be useful for the future.

PR #4925, opened 2025-01-17
chore(repo): Add Typedoc

LekoArts:
## Description

This PR adds [Typedoc](https://typedoc.org/) to the repository for this project: https://linear.app/clerk/project/auto-generate-api-docs-for-typescript-based-sdks-ddf6285ae589/overview

The PR does more than only adding Typedoc because I initially started this branch with the goal of already changing a bunch of SDKs at once. Since then I've decided to just do one PR per SDK but didn't want to split up the work already -- this PR also mostly contains smaller changes/whitespace changes anyways.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.changeset/shiny-schools-search.md
```diff
@@ -0,0 +1,2 @@
+---
```
LekoArts, line 1: I didn't drastically improve the JSDoc comments so I've opted for no explicit release here. The future PRs will be more substantial for each individual SDK

packages/astro/src/react/hooks.ts
```diff
@@ -121,8 +121,6 @@ type UseAuth = () => UseAuthReturn;
  * For projects using a server, you can have immediate access to this data during SSR.
  *
  * @example
- * A simple example:
```
LekoArts, line 124: The "A simple example" where removed because they'd otherwise would show up in the example code. We can already give context that it's an example without that sentence

packages/backend/src/tokens/authObjects.ts
```diff
@@ -40,7 +40,7 @@ export type SignedInAuthObject = {
    * Each item represents the minutes that have passed since the last time a first or second factor were verified.
    * [fistFactorAge, secondFactorAge]
    */
-  factorVerificationAge: [number, number] | null;
+  factorVerificationAge: [firstFactorAge: number, secondFactorAge: number] | null;
```
LekoArts, line 43: In your IDE you'll get names shown instead of only `number`

packages/expo/src/provider/ClerkProvider.tsx
```diff
@@ -17,8 +17,8 @@ export type ClerkProviderProps = React.ComponentProps<typeof ClerkReactProvider>
   /**
    * Note: Passkey support in Expo is currently in a limited rollout phase.
    * If you're interested in using this feature, please contact us for early access or additional details.
-   *
-   * @experimental This API is experimental and may change at any moment.
+   * This API is experimental and may change at any moment.
+   * @experimental
```
LekoArts, line 21: The tag only "works" if it is on its own, so it shouldn't have any comment after it

LekoArts (commented):
Self-review

package.json
```diff
@@ -103,6 +104,7 @@
     "tsup": "catalog:repo",
     "turbo": "^2.0.14",
     "turbo-ignore": "^2.0.6",
+    "typedoc": "^0.27.6",
```
jacekradko, line 107: [nit] would need to go along with `pnpm install`

```suggestion
    "typedoc": "0.27.6",
```

typedoc.json
```diff
@@ -0,0 +1,34 @@
+{
```
jacekradko, line 1: I think it would be easier to maintain this config if it was in JS: https://typedoc.org/documents/Options.Configuration.html#javascript-files

One benefit would be to automatically generate a list of entry points based on current state of `packages/*` folder

jacekradko:
Definitely enjoyed `typedoc` from other projects I worked on. Maybe we should consider something like https://www.npmjs.com/package/eslint-plugin-tsdoc to statically analyze our TSDocs. Caveat, I have never used that plugin so not sure it's worth it

LekoArts:
Eventually I want to add it (https://linear.app/clerk/issue/ECO-242/add-linting-to-validate-tsdoc-comments) but I think right now it would complain too much and get in the way, so I'll try to add it at the end.

packages/expo/src/provider/ClerkProvider.tsx
panteliselef: TIL

packages/astro/typedoc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "$schema": "https://typedoc-plugin-markdown.org/schema.json",
+  "entryPoints": ["./src/client/index.ts", "./src/types.ts"]
+}
```
panteliselef, line 4: @LekoArts I'm curious, why does this need to only exists for `@clerk/astro` ?

panteliselef (approved):

PR #4969, opened 2025-01-22
fix(clerk-js): Expand Keyless UI if user is signed-in

kaftarmery:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/strange-rings-applaud.md
+++ b/.changeset/strange-rings-applaud.md
@@ -2,4 +2,4 @@
 '@clerk/clerk-js': minor
 ---
 
-Expand Keyless popover on initial render if user signs-in.
+Expand Keyless popover when a user signs-in.
--- a/.changeset/strange-rings-applaud.md
+++ b/.changeset/strange-rings-applaud.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': minor
+---
+
+Expand Keyless popover on initial render if user signs-in.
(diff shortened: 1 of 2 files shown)


packages/clerk-js/src/ui/components/KeylessPrompt/index.tsx
```diff
@@ -22,7 +23,15 @@ const buttonIdentifier = `${buttonIdentifierPrefix}-button`;
...
 
 const _KeylessPrompt = (_props: KeylessPromptProps) => {
+  const { isSignedIn } = useUser();
   const [isExpanded, setIsExpanded] = useState(false);
+
+  useEffect(() => {
+    if (isSignedIn) {
+      setIsExpanded(true);
+    }
+  }, [isSignedIn]);
```
panteliselef, line 33: For other reviewers, I acknowledge this is not recommended, in this case we only want to set expanded to true, every time we detect that the developer signed in with a user.

.changeset/strange-rings-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/clerk-js': minor
+---
+
+Expand Keyless popover on initial render if user signs-in.
```
panteliselef, line 5:
```suggestion
Expand Keyless popover when a user signs-in.
```

PR #4970, opened 2025-01-22
Update nl-NL.ts

FiremanPete:
Update to the dutch localization.
Only string changes of existing code.


packages/localizations/src/nl-NL.ts
```diff
@@ -458,7 +458,7 @@ export const nlNL: LocalizationResource = {
       actionLink__use_phone: 'Gebruik telefoon',
       actionLink__use_username: 'Gebruik gebruikersnaam',
       actionText: 'Geen account?',
-      actionText__join_waitlist: 'Nog geen account? Meld je aan voor de wachtlijst.',
```
tmilewski, line 461: Let's leave the call-out to the waitlist here.

tmilewski (changes requested):
Hey @FiremanPete, thanks for the contribution! I made one small call-out around the waitlist copy. Happy to push this through, once that's up-to-date.

Thanks again!

packages/localizations/src/nl-NL.ts
FiremanPete: Hey @tmilewski, Thanks for the quick feedback!

The callout in the waitlist component clashes with the actual link
So its callout and then same callout but clickable, that's why I edited it out.

Don't suppose there is any way to have the element flex more so the text always fits? 
 
image

tmilewski: @FiremanPete Ah, I see. Happy to merge another PR, with just that change.

PR #4971, opened 2025-01-22
feat(nextjs): Mark keyless onboarding as complete when stored keys match explicit keys

panteliselef:
## Description

Calling `completeAccountlessApplicationOnboarding` will notify the dashboard to automatically proceed from step 2, to step 3 on the claim flow.

| Step 2 | Step 3|
| ------ | ------|
| image | image|

Since this is called inside `ClerkProvider`, we can to avoid making a BAPI request frequently. When the request successfully resolved, we are caching it for 24 hours, otherwise in case of an error it will be cached only for 10 minutes.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/nextjs/src/app-router/server/ClerkProvider.tsx
```diff
@@ -118,6 +119,24 @@ export async function ClerkProvider(
...
+            mod => mod.safeParseClerkFile()?.secretKey,
+          );
+          if (!secretKey) {
+            // we will ignore it later
+            throw new Error(secretKey);
+          }
+          const client = createClerkClientWithOptions({
+            secretKey,
+          });
+          // Add caching here
```
panteliselef, line 134: i think we need a `clerk_unstable_cache()` for this

brkalow (approved):

PR #4974, opened 2025-01-22
fix(clerk-js): Always pass domain to createFapiClient()

brkalow:
## Description

Ensure `createFapiClient()` always receives a domain so we can decide if we should be using it or not.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -306,10 +306,11 @@ export class Clerk implements ClerkInterface {
     this.#instanceType = publishableKey.instanceType;
 
     this.#fapiClient = createFapiClient({
-      domain: (this.instanceType === 'development' && this.isSatellite && this.domain) || undefined,
+      domain: this.domain,
       frontendApi: this.frontendApi,
       // this.instanceType is assigned above
       instanceType: this.instanceType as InstanceType,
+      isSatellite: this.isSatellite,
       getSessionId: () => {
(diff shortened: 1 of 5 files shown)


alexcarpenter (approved):

tmilewski (approved):

packages/clerk-js/src/core/clerk.ts
```diff
@@ -306,10 +306,11 @@ export class Clerk implements ClerkInterface {
     this.#instanceType = publishableKey.instanceType;
 
     this.#fapiClient = createFapiClient({
-      domain: (this.instanceType === 'development' && this.isSatellite && this.domain) || undefined,
+      domain: this.domain,
       frontendApi: this.frontendApi,
       // this.instanceType is assigned above
       instanceType: this.instanceType as InstanceType,
+      isSatellite: this.isSatellite,
```
panteliselef, line 313: Was this missing and that's why we had issues, or also domain being undefined was part of the reason ?

panteliselef (approved):

packages/clerk-js/src/core/clerk.ts
brkalow: `domain` being passed as undefined was the main reason. I missed that it was used unconditionally when initializing the client.

PR #4976, opened 2025-01-22
fix(backend): Add `orderBy` to `getOrganizationMembershipList`

LauraBeatris:
## Description

`orderBy` is a valid query param for retrieving memberships of a given organization, however it wasn't being forward to fetch call.

https://clerk.com/docs/reference/backend-api/tag/Organization-Memberships#operation/ListOrganizationMemberships!in=query&path=order_by&t=request

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/api/endpoints/OrganizationApi.ts
+++ b/packages/backend/src/api/endpoints/OrganizationApi.ts
@@ -206,13 +206,13 @@ export class OrganizationAPI extends AbstractAPI {
   }
 
   public async getOrganizationMembershipList(params: GetOrganizationMembershipListParams) {
-    const { organizationId, limit, offset } = params;
+    const { organizationId, limit, offset, orderBy } = params;
     this.requireId(organizationId);
 
     return this.request<PaginatedResourceResponse<OrganizationMembership[]>>({
       method: 'GET',
       path: joinPaths(basePath, organizationId, 'memberships'),
-      queryParams: { limit, offset },
+      queryParams: { limit, offset, orderBy },
     });
   }
 
--- a/packages/backend/src/api/endpoints/OrganizationApi.ts
+++ b/packages/backend/src/api/endpoints/OrganizationApi.ts
@@ -206,53 +206,45 @@ export class OrganizationAPI extends AbstractAPI {
   }
 
   public async getOrganizationMembershipList(params: GetOrganizationMembershipListParams) {
-    const { organizationId, limit, offset, orderBy } = params;
+    const { organizationId, ...queryParams } = params;
     this.requireId(organizationId);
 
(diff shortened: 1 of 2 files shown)


packages/backend/src/api/endpoints/OrganizationApi.ts
```diff
@@ -206,13 +206,13 @@ export class OrganizationAPI extends AbstractAPI {
   }
 
   public async getOrganizationMembershipList(params: GetOrganizationMembershipListParams) {
-    const { organizationId, limit, offset } = params;
+    const { organizationId, limit, offset, orderBy } = params;
```
jacekradko, line 209: Any thoughts on automatically forwarding any params (outside of `organizationId`) the the query string? 

```suggestion
    const { organizationId, ...queryParams } = params;
```

packages/backend/src/api/endpoints/OrganizationApi.ts
```diff
@@ -206,13 +206,13 @@ export class OrganizationAPI extends AbstractAPI {
...
     this.requireId(organizationId);
 
     return this.request<PaginatedResourceResponse<OrganizationMembership[]>>({
       method: 'GET',
       path: joinPaths(basePath, organizationId, 'memberships'),
-      queryParams: { limit, offset },
+      queryParams: { limit, offset, orderBy },
```
jacekradko, line 215: This goes with the other suggestion
```suggestion
      queryParams,
```

jacekradko (commented):
To avoid these situations in the future, we could just forward all params to the query string. Thoughts?

packages/backend/src/api/endpoints/OrganizationApi.ts
LauraBeatris: This was already being done for another existing method - I agree it's more extendable! 

If folks are using TS, then invalid params are caught in static types. For runtime, then BAPI is going to return a 400 error as well.

packages/backend/src/api/endpoints/OrganizationApi.ts
LauraBeatris: @jacekradko Updated here

LauraBeatris:
I've found two empty modules without any sort of usage, not being imported within `index.ts` - decided to delete it

packages/backend/src/api/endpoints/OrganizationPermissionApi.ts
LauraBeatris, line 1: I've found two empty modules without any sort of usage, not being imported within index.ts - decided to delete it

tmilewski (approved):

jacekradko (approved):
LGTM!

PR #4991, opened 2025-01-24
fix(shared): Catch atob error

Wolfleader101:
## Description

`atob()` returns an exception which currently isn't handled and causes vague exception to be thrown which can be hard to track down.

Example of the exception can be found below.
image

A simple fix for this involves wrapping the contents of `isPublishableKey(key)` in a `try-catch` and returning false on exception, which will then produce the exception `Publishable key not valid.`

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


Wolfleader101:
Not too sure why some of those have failed? Doesnt seem related to anything Ive changed?
image

PR #5012, opened 2025-01-26
Add bun in the crypto section of the package.json imports.

p9f:
## Description

Hi there,

I am trying to add clerk to a tanstack start project running with bun.
The production command is failing with the following error:

> error: Cannot find package '#crypto' from '/home/runner/work/start-clerk-bun-repro/start-clerk-bun-repro/.output/server/node_modules/@clerk/backend/dist/chunk-G4VEKB6A.mjs'

(full minimal repro is there -> https://github.com/p9f/start-clerk-bun-repro)

It appears that adding bun to the package.json imports map (this PR) fixes it.

Even if you don't officially support bun, would you consider merging this?

Thanks

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -17,6 +17,7 @@
       "edge-light": "./dist/runtime/browser/crypto.mjs",
       "worker": "./dist/runtime/browser/crypto.mjs",
       "browser": "./dist/runtime/browser/crypto.mjs",
+      "bun": "./dist/runtime/browser/crypto.mjs",
       "node": {
         "require": "./dist/runtime/node/crypto.js",
         "import": "./dist/runtime/node/crypto.mjs"


LekoArts:
Hey, thanks for your PR!

Unfortunately I won't merge it as I don't think we should cater for non-spec compliant things. The list of exports is here: https://nodejs.org/api/packages.html#conditional-exports
I know, webpack-specific syntax also exists and we use it in some places but only because it's so super common. With bun I'm more inclined to close it because their stance (mentioned here: https://bun.sh/docs/runtime/nodejs-apis) is:

> If a package works in Node.js but doesn't work in Bun, we consider it a bug in Bun. Please [open an issue](https://bun.sh/issues) and we'll fix it.

So if you're up for it, please open an issue with them. Hope that makes sense!

p9f:
makes sense - thanks @LekoArts for the reply.

With more thinking - i guess it might be a `nitro` / `rollup` issue instead. I will investigate more.

PR #5013, opened 2025-01-27
fix(clerk-js): Navigate to specific instance for configuration after keyless

panteliselef:
## Description

Managing urls to dashboard like that is not sustainable, this will be addressed the following PR, once the flow has been finalized

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/KeylessPrompt/index.tsx
+++ b/packages/clerk-js/src/ui/components/KeylessPrompt/index.tsx
@@ -46,6 +46,36 @@ const baseElementStyles = css`
   text-decoration: none;
 `;
 
+/**
+ * If we cannot reconstruct the url properly, then simply fallback to Clerk Dashboard
+ */
+function withLastActiveFallback(cb: () => string): string {
+  try {
+    return cb();
+  } catch {
+    return 'https://dashboard.clerk.com/last-active';
+  }
+}
+
+function handleDashboardUrlParsing(url: string) {
+  // make sure this is a valid url
+  const __url = new URL(url);
+  const regex = /^https?:\/\/(.*?)\/apps\/app_(.+?)\/instances\/ins_(.+?)(?:\/.*)?$/;
+
+  const match = __url.href.match(regex);
+
+  if (!match) {
+    throw new Error('invalid value dashboard url structure');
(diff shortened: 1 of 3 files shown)


packages/clerk-js/src/ui/components/KeylessPrompt/index.tsx
```diff
@@ -50,29 +97,25 @@ const _KeylessPrompt = (_props: KeylessPromptProps) => {
...
-    text-decoration: none;
-  `;
+  const instanceUrlToDashboard = useMemo(() => {
+    try {
+      const url = new URL(
+        `${redirectUrlParts.baseDomain}/apps/${redirectUrlParts.appId}/instances/${redirectUrlParts.instanceId}/user-authentication/email-phone-username`,
+      );
+      return url.href;
+    } catch {
+      return '';
```
LekoArts, line 107: Wouldn't it be better to use the last-active path as fallback instead of an empty string?

LekoArts (commented):
Please add a PR description, thanks!

packages/clerk-js/src/ui/components/KeylessPrompt/index.tsx
LekoArts: I see 👍 I guess my point was that after `copyKeysUrl` is malformed and then also the regex fails, should we have a fallback that "does something" (because empty string won't do anything and the user also doesn't know what's wrong) or should we throw an error? We probably don't want to show an error since it's not actionable (they can't do anything about it) but silently failing is also not nice.

Not familiar with all moving pieces here so I'll trust your judgement :)

anagstef (approved):

packages/clerk-js/src/ui/components/KeylessPrompt/index.tsx
panteliselef: Let's having something more solid and then iterate. I've pushed [this](https://github.com/clerk/javascript/pull/5013/commits/5872d9faf958e83c50ecaee1afad7518c9b7091c)

PR #5020, opened 2025-01-27
feat(clerk-js): Opt out of combined flow when in restricted mode

alexcarpenter:
## Description

When sign up mode is listed as `restricted` we should opt out of the combined flow even when the sign up url is undefined.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


panteliselef (approved):

integration/tests/sign-in-or-up-restricted-mode.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  test.afterAll(async () => {
+    await fakeUser.deleteIfExists();
+    await app.teardown();
+  });
+
+  test('It does not allow sign-in-or-up flow', async ({ page, context }) => {
+    const u = createTestUtils({ app, page, context });
+    await u.po.signIn.goTo();
+    await u.po.signIn.waitForMounted();
+    await u.po.signIn.getIdentifierInput().fill(fakeUser.email);
```
dstaley, line 32: We should assert that we're in the combined flow, just to be sure. If for some reason the combined flow and restricted mode _wasn't_ being used, I believe this test would pass.

dstaley (approved):

PR #5035, opened 2025-01-29
feat(backend): cache by optional kid param in loadClerkJWKFromLocal

itsjxck:
## Description

Updates `loadClerkJWKFromLocal` to accept an optional second parameter `kid` that allows overriding the default `local` cache key.

Original behaviour preserved using default value for new `kid` field.

I also updated the `loads the local key in PEM format` test to make use of the cache key override, because the original test would _always_ be successful since the first `loads the local key` test would create a cache entry, so the logic to parse the key in the second test was skipped.

A secondary note, while adding this and updating tests, I noticed the `throws an error if no key has been provided` test will only ever succeed if there is nothing in the cache for the cache key. I haven't changed this test as part of this PR but thought it was worth raising that the test does not work when there is a cached value, allowing passing `undefined` to be successful. (I actually make use of this in a test to demonstrate `loadClerkJWKFromLocal` using a custom cache key).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


itsjxck:
Can someone check this out...?

clerk-cookie:
Hello 👋

PR #5028, opened 2025-01-28
feat(nextjs): [improved] Hint correct middleware location when missing clerkMiddleware

panteliselef:
## Description

This PR brings back the improvements instoduced in  #4979 , but it also fixes warnings caused from codesplitting issues when importing `clerkMiddelware` from `/server` but `/server` exports other utilities that use node apis.

Since this reverts a reverted commit, the actual issue is resolved by this [new commit](https://github.com/clerk/javascript/pull/5028/commits/648f2e8c1a2f59dc62afe25e66c03a7104d0442d)

Running `next build` will not longer complain about usage of `process.cwd`

Fixes ACTLS-63

|  `next start` with defaut rendering (node)   | `next start` with edge rendering|
| -------- | ----| 
| missing middelware on node rendering start | missing middleware on edge rendering |

## Checklist

- [ ] `pnpm test` runs as expected.
- [ ] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #5037, opened 2025-01-29
test(nextjs): E2E tests for middleware file placement

panteliselef:
## Description

Adds e2e tests for #5028, but also changes `useTemplate()` to completely override the previous selection. Before it would merge them.

Fixes ACTLS-63

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


integration/tests/middleware-placement.test.ts
```diff
@@ -0,0 +1,115 @@
+import { expect, test } from '@playwright/test';
+
+import type { Application } from '../models/application';
+import { appConfigs } from '../presets';
+import { createTestUtils } from '../testUtils';
+
+const commonSetup = appConfigs.next.appRouterQuickstart.clone().removeFile('src/middleware.ts');
+
+test.describe('next start - missing middleware @quickstart', () => {
```
alexcarpenter, line 9: Still learning the testing setup/configs, whats the reasoning your tagging `@quickstart` vs `@nextjs` here?

integration/models/application.ts
```diff
@@ -93,7 +94,11 @@ export const application = (
     get buildOutput() {
       return buildOutput;
     },
+    get serveOutput() {
```
alexcarpenter, line 97: Is this supposed to be `serverOutput`?

integration/tests/middleware-placement.test.ts
```diff
@@ -0,0 +1,115 @@
+import { expect, test } from '@playwright/test';
+
+import type { Application } from '../models/application';
+import { appConfigs } from '../presets';
+import { createTestUtils } from '../testUtils';
+
+const commonSetup = appConfigs.next.appRouterQuickstart.clone().removeFile('src/middleware.ts');
```
alexcarpenter, line 7: Can we add a case for valid middleware usage ensuring no error is thrown?

integration/models/application.ts
panteliselef: Actually it is the output of the `server` command, hence the name

integration/tests/middleware-placement.test.ts
panteliselef: In this case I chose quickstart, because i'm checking against the quickstart template. AFAIK the behaviour I am testing for will not based on the nextjs version, so putting it to quickstart will avoid running the tests against all nextjs versions. Having said that I will double check.

integration/tests/middleware-placement.test.ts
panteliselef: I think this is implicitly covered from the existing long runnning application and the current quickstart test, because if middleware didn't exist those tests would fail.

panteliselef: but i'm ok to go with the more explicit one, if you insist

alexcarpenter: No strong opinion, feel free to ignore if its already covered.

alexcarpenter (approved):

PR #5040, opened 2025-01-29
chore(clerk-js,chrome-extension): Check for RHC

tmilewski:
## Description

This script will do a better job of testing against `chrome-extension` and `clerk-js` to ensure that remotely hosted code isn't present in particular builds.

Fixes ECO-351

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


brkalow (approved):
👏

scripts/search-for-rhc.mjs
```diff
@@ -7,13 +7,32 @@
...
-  throw new Error('Found RHC in build output');
-} else {
-  console.log('✅ No RHC found in build output');
+const targetType = argv._[0]; // file | directory
+const target = argv._[1]; // Target of the resource
+
+async function asyncSearchRHC(name, search) {
+  const cmd = () =>
+    targetType === 'directory'
+      ? $`grep -rFq --include=\\*.js --include=\\*.mjs "${search}" ${target}`
```
LekoArts, line 16: @tmilewski As a note in case we want to change it back:

- I've used the longer forms of the flags because then you don't have to look up what `-rFq` does
- `zx` has a feature where you can pass an array of flags to the command and it escapes everything already, that's why `${flags}` was used

tmilewski: @LekoArts Yeah, we can move to the longer forms. 

My original goal was to have `zx` handle flags for in that manner, but for some reason, it simply wasn’t parsing things correctly. This seems fine for the time being as it’s quite minimal.

PR #5043, opened 2025-01-29
chore(nextjs,astro,shared): Move path matcher logic to shared

wobsoriano:
## Description

This PR moves the core path matcher logic living inside the `createRouteMatcher` of our Astro and Next.js SDKs to shared package to avoid duplication and pave the way for other SDKs like Nuxt. We'd want to make it easy to adopt in other SDKs while keeping framework-specific handling separate.

In Nuxt, it will be implemented like this to be used within pages:

```ts
import type { RouteLocationNormalizedLoadedGeneric } from 'vue-router'

export const createRouteMatcher = (routes: RouteMatcherParam) => {
  const matcher = createPathMatcher(routes);
  return (to: RouteLocationNormalizedLoadedGeneric) => matcher(to.path)
}
```

and in a [global nuxt route middleware](https://nuxt.com/docs/guide/directory-structure/middleware#ordering-global-middleware), they can use it like so:

```ts
export default defineNuxtRouteMiddleware((to) => {
  const { userId } = useAuth()

  if (!userId.value && isProtectedRoute(to)) {
    return navigateTo('/sign-in')
  }
})
```

For API routes, a separate `createRouteMatcher` will be created that accepts an [H3 Event](https://h3.unjs.io/guide/event)

```ts
import { H3Event, getRequestURL } from 'h3'

export const createRouteMatcher = (routes: RouteMatcherParam) => {
  const matcher = createPathMatcher(routes);
  return (event: H3Event) => {
    const url = getRequestURL(event); // returns the full url
    return matcher(new URL(url).pathname)
  }
}
```

Resolves ECO-352

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


tmilewski (approved):

panteliselef (approved):

.changeset/brown-suns-clap.md
```diff
@@ -0,0 +1,7 @@
+---
+"@clerk/astro": patch
+"@clerk/nextjs": patch
+"@clerk/shared": patch
+---
+
+Move path matcher logic to shared package
```
LekoArts, line 7:
````suggestion
Previously the `createPathMatcher()` function was re-implemented both in `@clerk/astro` and `@clerk/nextjs`, this PR moves this logic to `@clerk/shared`.

You can use it like so:

```ts
import { createPathMatcher } from '@clerk/shared/pathMatcher'
```
````

File: packages/shared/src/__tests__/pathMatcher.test.ts [hunk 2]
```diff
@@ -0,0 +1,52 @@
...
+    expect(matcher('/foo/bar/baz')).toBe(true);
+    expect(matcher('/bar')).toBe(false);
+  });
+
+  test('matches array of patterns', () => {
+    const matcher = createPathMatcher(['/foo', '/bar(.*)']);
+    expect(matcher('/foo')).toBe(true);
+    expect(matcher('/bar')).toBe(true);
+    expect(matcher('/bar/baz')).toBe(true);
+    expect(matcher('/baz')).toBe(false);
```
LekoArts, line 27: You could also add this assertion:

```ts
expect(matcher('/foo/bar')).toBe(false);
```

anagstef (approved):

PR #5050, opened 2025-01-30
feat(nuxt): Add `createRouteMatcher()` helper

wobsoriano:
## Description

This PR adds a `createRouteMatcher()` helper for Nuxt that allows you to protect multiple pages or API routes.

For pages, `createRouteMatcher` is auto-imported and can be used like so (inside a global route middleware):

```ts
// middleware/auth.global.ts

export default defineNuxtRouteMiddleware(to => {
  const { userId } = useAuth();

  if (userId.value && isProtectedRoute(to)) {
    // Add custom logic to run before redirecting
    return navigateTo('/sign-in');
  }
});
```

For API routes, `createRouteMatcher` needs to be imported from `@clerk/nuxt/server`:

```ts
// server/middleware/auth.ts
import { clerkMiddleware, createRouteMatcher } from '@clerk/nuxt/server';

const isProtectedRoute = createRouteMatcher(['/api/user(.*)', '/api/projects(.*)']);

export default clerkMiddleware((event) => {
    const { userId } = event.context.auth

    if (!userId && isProtectedRoute(event)) {
        setResponseStatus(event, 401)
        return 'You are not authorized to access this resource.'
    }
});
```

Why 2 `createRouteMatcher`s? Because they serve different purposes here:

- Route middleware runs on initial page load and subsequent client-side route changes. It receives route objects and can control page navigation.
- Server middleware runs on the server for every request (API calls, HTML rendering). It receives H3 event objects and can handle things like headers, cookies, and request context.

Each `createRouteMatcher` is optimized for its specific middleware type.

Resolves ECO-312

## Checklist

- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated - to follow

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.changeset/sour-foxes-sort.md
```diff
@@ -0,0 +1,39 @@
...
+For protecting pages (in a global route middleware):
+
+```ts
+// createRouteMatcher is automatically imported
+const isProtectedRoute = createRouteMatcher(['/dashboard(.*)', '/forum(.*)'])
+
+export default defineNuxtRouteMiddleware(to => {
+  const { userId } = useAuth();
+
+  if (userId.value && isProtectedRoute(to)) {
```
LekoArts, line 16: Shouldn't it be `!userId.value` instead?

packages/nuxt/src/runtime/client/routeMatcher.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type RouteMatcherParam = PathMatcherParam;
+
+/**
+ * Returns a function that accepts a Vue Router route location and returns whether the route matches the list of
+ * predefined routes that can be passed in as the first argument.
+ *
+ * You can use glob patterns to match multiple routes or a function to match against the route object.
+ * Path patterns and regular expressions are supported, for example: `['/foo', '/bar(.*)'] or `[/^\/foo\/.*$/]`
+ * For more information, see: https://clerk.com/docs
```
LekoArts, line 14: Adapted from the existing Next.js docs

```suggestion
 * `createRouteMatcher` is a Clerk helper function that allows you to protect multiple routes. It accepts an array of routes and checks if the route the user is trying to visit matches one of the routes passed to it.
 *
 * The `createRouteMatcher` helper function returns a function that accepts a Vue Router route location and will return `true` if the user is trying to access a route that matches on of the routes passed to `createRouteMatcher`.
 *
 * @example
 * ['/foo', '/bar(.*)']
 * @example
 * [/^\/foo\/.*$/]
```

LekoArts (changes requested):

.changeset/sour-foxes-sort.md
wobsoriano: good catch!

LekoArts (approved):

PR #5051, opened 2025-01-31
fix(tanstack-start): Fix event handler return type mismatch

wobsoriano:
## Description

This PR fixes an issue where the handler ([`createStartHandler`](https://github.com/TanStack/router/blob/c5b3f548df260abb54683042dea901ddb414eac0/packages/start-server/src/createStartHandler.ts#L9,L11)) exposed by `@tanstack/start/server` doesn't match the type expected by our TanStack SDK handler ([`createClerkHandler`](https://github.com/clerk/javascript/blob/d3152be7f01fbb5ca26aeddc2437021f4b7ecc83/packages/tanstack-start/src/server/middlewareHandler.ts#L14)). Right now we're using `EventHandler` from `vinxi/http` which causes type incompatibility with TanStack's handler that returns `ReturnType<typeof eventHandler>` from `h3`.

A key difference is that `eventHandler` from `h3` returns a function with overloaded signatures, while `EventHandler` from `vinxi/http` is a simpler type without overloads. This type mismatch creates errors for users trying to pass the handler from `createStartHandler` directly to `createClerkHandler`.

Screenshot 2025-01-31 at 10 05 44 AM

Resolves ECO-353

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/tanstack-start/package.json
```diff
@@ -69,7 +69,8 @@
   "devDependencies": {
     "@tanstack/react-router": "^1.97.25",
     "@tanstack/start": "^1.97.25",
-    "esbuild-plugin-file-path-extensions": "^2.1.2"
+    "esbuild-plugin-file-path-extensions": "^2.1.2",
+    "h3": "^1.13.0"
```
wobsoriano, line 73: Just adding to dev deps as we don't use it in runtime. Just need the return type

tmilewski (approved):

PR #5053, opened 2025-01-31
chore(clerk-react): Improve JSDoc comments

LekoArts:
## Description

Improve and add JSDoc comments to the public API of `@clerk/clerk-react` to document all client-side helpers (later with Typedoc). Since some of the hooks are in `@clerk/shared` you'll also see changes there.

Some comments are duplicated and _not_ DRY, unfortunately you can't add a comment to a type alias and use that instead. TypeScript doesn't support that 😢 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/src/app-router/server/currentUser.ts
+++ b/packages/nextjs/src/app-router/server/currentUser.ts
@@ -12,9 +12,7 @@ import { auth } from './auth';
  * - counts towards the [Backend API request rate limit](https://clerk.com/docs/backend-requests/resources/rate-limits#rate-limits).
  *
  * @example
- * ```tsx
- * // app/page.tsx
- *
+ * ```tsx {{ filename: 'app/page.tsx' }}
  * import { currentUser } from '@clerk/nextjs/server'
  *
  * export default async function Page() {
--- a/packages/nextjs/src/server/buildClerkProps.ts
+++ b/packages/nextjs/src/server/buildClerkProps.ts
@@ -12,11 +12,9 @@ type BuildClerkProps = (req: RequestLike, authState?: BuildClerkPropsInitState) 
  * Clerk uses `buildClerkProps` to inform the client-side helpers of the authentication state of the user. This function is used SSR in the `getServerSideProps` function of your Next.js application.
  *
  * @example
- * **Basic usage**
+ * ### Basic usage
  *
- * ```tsx
- * // pages/myServerSidePage.tsx
- *
+ * ```tsx {{ filename: 'pages/myServerSidePage.tsx' }}
  * import { getAuth, buildClerkProps } from '@clerk/nextjs/server'
  * import { GetServerSideProps } from 'next'
  *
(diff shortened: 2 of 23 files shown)


packages/nextjs/src/app-router/server/currentUser.ts
```diff
@@ -12,9 +12,7 @@ import { auth } from './auth';
  * - counts towards the [Backend API request rate limit](https://clerk.com/docs/backend-requests/resources/rate-limits#rate-limits).
  *
  * @example
- * ```tsx
- * // app/page.tsx
- *
+ * ```tsx {{ filename: 'app/page.tsx' }}
```
LekoArts, line 15: I've changed my approach to indicate the filename to what we currently use in the docs. This won't show up in your IDE but we then can later use it in the docs.

packages/nextjs/src/server/buildClerkProps.ts
```diff
@@ -12,11 +12,9 @@ type BuildClerkProps = (req: RequestLike, authState?: BuildClerkPropsInitState)
  * Clerk uses `buildClerkProps` to inform the client-side helpers of the authentication state of the user. This function is used SSR in the `getServerSideProps` function of your Next.js application.
  *
  * @example
- * **Basic usage**
- *
- * ```tsx
- * // pages/myServerSidePage.tsx
+ * ### Basic usage
```
LekoArts, line 15: Realized I can use headings, so switched them to what they'll be in the docs

packages/react/src/hooks/__tests__/useRoutingProps.test.tsx
```diff
@@ -16,7 +16,7 @@ describe('useRoutingProps()', () => {
   });
 
   test('defaults to path routing and a path prop is required', () => {
-    const TestingComponent = props => {
+    const TestingComponent = (props: any) => {
```
LekoArts, line 19: Typedoc complained about type errors here, so I fixed those

packages/react/src/hooks/useSignIn.ts
```diff
@@ -5,9 +5,38 @@ import type { UseSignInReturn } from '@clerk/types';
...
+ * }
+ * ```
+ *
+ * @example
+ * ### Create a custom sign-in flow with `useSignIn()`
+ *
+ * The `useSignIn()` hook can also be used to build fully custom sign-in flows, if Clerk's prebuilt components don't meet your specific needs or if you require more control over the authentication flow. Different sign-in flows include email and password, email and phone codes, email links, and multifactor (MFA). To learn more about using the `useSignIn()` hook to create custom flows, see the [custom flow guides](https://clerk.com/docs/custom-flows/overview).
+ *
+ * ```
+ * ```
```
LekoArts, line 37: This empty code block is on purpose as without it the whole `@example` block would be interpreted as a code block. Only when you manually add a codeblock the text above is treated as such

packages/react/src/isomorphicClerk.ts
```diff
@@ -1151,7 +1151,7 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
     }
   };
 
-  authenticateWithOKXWallet = async (params: AuthenticateWithOKXWalletParams): Promise<void> => {
+  authenticateWithOKXWallet = async (params?: AuthenticateWithOKXWalletParams): Promise<void> => {
```
LekoArts, line 1154: Typedoc complained about type errors here, so I fixed those. It's weird that it doesn't happen for other functions above that also have optional `params`

packages/shared/src/react/hooks/useOrganization.tsx
LekoArts, line 1: Had to repeat all comments 😢

LekoArts (commented):
Comments

tmilewski (approved):

PR #5054, opened 2025-01-31
fix(e2e): Always navigate to `/protected` after account creation in Oauth flow tests

panteliselef:
## Description

In same cases, a transfer flow it is triggered from out `oauth-flow.test.ts`. In our tests we expect to always return to `/protected` but since our test template application only defines a forceRedirectUrl in SignInButton only the `sign_in_force_redirect_url` query param will be set. During the transfer from, flow sign-in to sign-up, that query param has no effect so the user will end up stuck in Account Portal.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


anagstef (approved):

dstaley (approved):
LGTM! I'll add a new button for the oauth flows test that disables transfers, that way we can assert that a sign up wasn't transferred to a sign in.

PR #5058, opened 2025-02-01
fix(backend): change export type to export for accessing static methods

LuccaRebelloToledo:
## Description

This PR addresses an issue where static methods could not be accessed due to the use of export type. The error encountered was:

```sh
'User' cannot be used as a value because it was exported using 'export type'.
```

### Changes
- Changed export type to export for the necessary modules to allow access to static methods.

### Testing
 - Verified that the static methods can now be accessed without any errors.

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/blue-geckos-hug.md
+++ b/.changeset/blue-geckos-hug.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': major
+---
+
+Changed export type to export for necessary modules to fix the issue with accessing static methods.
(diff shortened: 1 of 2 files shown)


LekoArts:
Hi, thanks for the PR!

We only export the types on purpose so we can't accept the change you're proposing here. Please open an issue explaining what problem you're running into and I'm sure we can help with it that doesn't require that change.

PR #5068, opened 2025-02-02
Revert "fix(tanstack-start): Fix event handler return type mismatch"

wobsoriano:
Reverts clerk/javascript#5051

This reversion is no longer needed (and didn't actually fix the issue) as the underlying h3 version compatibility issue 
has been resolved upstream in the TanStack Router repository ([TanStack/router#3310](https://github.com/TanStack/router/pull/3310))

### DIFF
--- a/.changeset/quiet-ties-walk.md
+++ b/.changeset/quiet-ties-walk.md
@@ -4,5 +4,4 @@
 
 Reverts [#5051](https://github.com/clerk/javascript/pull/5051)
 
-This reversion is no longer needed as the underlying h3 version compatibility issue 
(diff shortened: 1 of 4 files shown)


panteliselef:
@wobsoriano If we revert would that be a breaking change ? If it is are we consider it a bug and thus justifiable ?

wobsoriano:
> @wobsoriano If we revert would that be a breaking change ? If it is are we consider it a bug and thus justifiable ?

This should not be a breaking change and only an internal type update. I thought I fixed it locally when testing, turns out I have a cached dependency. My previous update didn’t really do anything at all so reverting it should be safe 🫡

LekoArts (approved):

panteliselef (approved):

PR #5073, opened 2025-02-03
fix(nextjs): Fix infinite redirect loop when page does not exist

panteliselef:
## Description

On keyless avoid infinite redirect loop when page does not exist and application is attempting to sync state with middleware.

### Before

https://github.com/user-attachments/assets/ddaf1578-3aa2-47e0-89e4-decf4e251cfe

### After
https://github.com/user-attachments/assets/1f56404f-3b37-486b-8c1e-f04f387ded1e

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/src/app-router/keyless-actions.ts
+++ b/packages/nextjs/src/app-router/keyless-actions.ts
@@ -5,7 +5,7 @@ import { redirect, RedirectType } from 'next/navigation';
 
 import { errorThrower } from '../server/errorThrower';
 import { detectClerkMiddleware } from '../server/headers-utils';
-import { getKeylessCookieName } from '../server/keyless';
+import { getKeylessCookieName, keylessRedirectCountCookieName } from '../server/keyless';
 import { canUseKeyless } from '../utils/feature-flags';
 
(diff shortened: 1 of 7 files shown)


packages/nextjs/src/app-router/keyless-actions.ts
```diff
@@ -20,7 +20,14 @@ export async function syncKeylessConfigAction(args: AccountlessApplication & { r
 
   // We cannot import `NextRequest` due to a bundling issue with server actions in Next.js 13.
   // @ts-expect-error Request will work as well
-  if (detectClerkMiddleware(request)) {
+  const detectedMiddelware = detectClerkMiddleware(request);
+
+  /**
+   * Prevents infinite redirects in when the `returnUrl` will result to a 404 (not-found) page.
```
dstaley, line 26:
```suggestion
   * Prevents infinite redirects when the `returnUrl` will result to a 404 (not-found) page.
```

dstaley (approved):

brkalow (changes requested):
Let's find a different approach that checks the data we're trying to sync

brkalow (approved):
👏

PR #5077, opened 2025-02-04
Elef/sdki 856 userprofile modal backdrop is still mounted after delete

panteliselef:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


panteliselef:
Currently linting is expected to fail, until all `setActive({redirectUrl})` usages are updated within our UI components.

dstaley:
Why was `setActive({ beforeEmit })` deprecated originally? This feels like we're remaking that API, only more convoluted with the introduction of the scoped context. I'd love to understand why it was removed originally because as it's shown here I feel like `setActive({ __internal_beforeEmit })` would be a cleaner alternative API.

Also, I wonder if there's a better expression of this idea that we could safely expose as a public API?

panteliselef:
Alternative solution #5092 got approval, so I will be closing this

PR #5084, opened 2025-02-05
fix(nextjs): Do not run invalidateCacheAction on v15

dstaley:
## Description

This PR updates the `onBeforeSetActive` handler to only run the `invalidateCacheAction` server action on Next.js v14, since the router cache in v15 is less aggressive. This also works around an issue when v15 apps are deployed to Vercel where server actions reject when their response is rewritten by middleware, causing sign out functionality on protected routes to not work correctly.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/nasty-weeks-hear.md
+++ b/.changeset/nasty-weeks-hear.md
@@ -1,0 +1,5 @@
+---
+'@clerk/nextjs': patch
+---
+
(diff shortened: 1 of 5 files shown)


panteliselef:
#5171 includes a similar solution, and also expands on it.

PR #5091, opened 2025-02-06
fix(nextjs): Shallow routing for in-component navigations

panteliselef:
### Before (notice the brief flash of the page)

https://github.com/user-attachments/assets/f79e0336-b85d-4222-9ba6-2d189ef1921a

### After
https://github.com/user-attachments/assets/de4029bc-9657-46d3-9566-b1d4bc08a2fd

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/src/app-router/client/useInternalNavFun.ts
+++ b/packages/nextjs/src/app-router/client/useInternalNavFun.ts
@@ -15,7 +15,7 @@ export const useInternalNavFun = (props: {
   windowNav: typeof window.history.pushState | typeof window.history.replaceState | undefined;
   routerNav: AppRouterInstance['push'] | AppRouterInstance['replace'];
   name: string;
-}) => {
+}): NavigationFunction => {
   const { windowNav, routerNav, name } = props;
   const pathname = usePathname();
   const [isPending, startTransition] = useTransition();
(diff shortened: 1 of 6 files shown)


packages/nextjs/src/app-router/client/useInternalNavFun.ts
```diff
@@ -68,8 +68,8 @@ export const useInternalNavFun = (props: {
     }
   }, [pathname, isPending]);
 
-  return useCallback((to: string) => {
-    return getClerkNavigationObject(name).fun(to);
+  return useCallback<NavigationFunction>((to, metadata) => {
+    return getClerkNavigationObject(name).fun(to, metadata);
```
dstaley, line 72: mind sharing an instance of where we were passing the second argument `metadata` to `push`/`replace`? I want to look into why that wasn't causing a typescript error. `useAwaitableReplace`/`useAwaitablePush` should have only accepted a single argument.

panteliselef: it was not giving any typescript errors because the 2nd argument is optional

dstaley: TypeScript is inferring the type as `const useAwaitablePush: () => (to: string) => unknown` though, so the returned function should be typed as only accepting a single argument. I guess we're losing that type safety somewhere along the way 🤔

dstaley: ahhh okay clerk-js accepts that as `RouterFn` which has the optional `metadata` argument. Since the previous type was `(to: string) => unknown`, that satisfies `(to: string, metadata?: whatever) => void`.

dstaley (approved):

LauraBeatris (approved):

PR #5095, opened 2025-02-06
fix(clerk-js): Avoid triggering email code verification twice on React strict mode

LauraBeatris:
#### Before

https://github.com/user-attachments/assets/346dc624-08f6-49cb-822d-c9309c6d08b9

#### After

https://github.com/user-attachments/assets/f4b6cde4-08db-4485-858c-86dc8575b60e

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/SignIn/SignInFactorOneCodeForm.tsx
+++ b/packages/clerk-js/src/ui/components/SignIn/SignInFactorOneCodeForm.tsx
@@ -54,6 +54,8 @@ export const SignInFactorOneCodeForm = (props: SignInFactorOneCodeFormProps) => 
       .catch(err => handleError(err, [], card.setError));
   };
 
+  console.log(props.factor, shouldAvoidPrepare);
+
   useFetch(
     shouldAvoidPrepare
       ? undefined
@@ -63,8 +65,11 @@ export const SignInFactorOneCodeForm = (props: SignInFactorOneCodeFormProps) => 
             .then(() => props.onFactorPrepare())
             .catch(err => handleError(err, [], card.setError)),
     {
-      name: 'prepare',
-      strategy: props.factor.strategy,
+      name: 'signIn.prepareFirstFactor',
+      factor: props.factor,
+    },
+    {
+      staleTime: 100,
     },
   );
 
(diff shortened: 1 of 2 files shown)


packages/clerk-js/src/ui/components/SignIn/SignInFactorOneCodeForm.tsx
```diff
@@ -37,23 +37,37 @@ export const SignInFactorOneCodeForm = (props: SignInFactorOneCodeFormProps) =>
...
+      ? undefined
+      : () =>
+          signIn
+            ?.prepareFirstFactor(props.factor)
+            .then(() => props.onFactorPrepare())
+            .catch(err => handleError(err, [], card.setError)),
+    {
+      name: 'prepare',
+      strategy: props.factor.strategy,
+    },
```
panteliselef, line 68: 1. Let's use a more specific name for the cache key
2. If you go back to start and use a different identifier (email) it would not auto-prepare on render.
```suggestion
    {
      name: 'signIn.prepareFirstFactor',
      factor: props.factor,
    },
```

panteliselef (changes requested):

packages/clerk-js/src/ui/components/SignIn/SignInFactorOneCodeForm.tsx
LauraBeatris: Good catch!

LauraBeatris: I've updated it here: https://github.com/clerk/javascript/pull/5095/commits/2c0390594be32c7b143abdc1e8ea7ad983dd378c

The behavior now is the following: It'll cache for a given identifier, so if you go back to `/sign-in` and submit again, it won't send another email cause it's cached - I've added a smaller stale time for that reason, unless you click to "Resend"

When refreshing the page, it'll only fetch once instead of twice (previous bug)

IamJoshing:
This still happens on "@clerk/clerk-expo": "^2.10.2",

PR #5106, opened 2025-02-08
fix(clerk-js): Reduce retries for `/tokens` on 5xx errors

anagstef:
## Description

When requesting a token from FAPI on `/tokens`, we retry 4 times if the response is NOT a 4xx error. 

We want to also stop retrying on 5xx errors.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


octoper (approved):

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -85,7 +85,8 @@ export class Session extends BaseResource implements SessionResource {
 
   getToken: GetToken = async (options?: GetTokenOptions): Promise<string | null> => {
     return runWithExponentialBackOff(() => this._getToken(options), {
-      shouldRetry: (error: unknown, currentIteration: number) => !is4xxError(error) && currentIteration < 4,
+      shouldRetry: (error: unknown, currentIteration: number) =>
```
brkalow, line 88: Do we really not want to retry on 5xx at all? Traditionally retrying on 5xx is common practice in the case of intermittent errors, but I do see the risk with potentially making the underlying problem worse.

LauraBeatris (commented):
I see the case of FAPI going down, and the first call to `/tokens` fail with a 5xx, but the subsequential ones might succeed once the server normalizes 

Maybe instead of not retrying at all, we can throttle with a larger delay?

anagstef:
Closing this in favor of another solution we are already working on!

PR #5124, opened 2025-02-10
fix(clerk-js): Fix stale `SignIn` object on `authenticateWithRedirect` for enterprise custom flows

LauraBeatris:
## Description

Resolves ORGS-540

Currently, `authenticateWithRedirect` always uses the existing `SignIn` for `saml` or `enterprise_sso` strategy, even if the identifier gets changed by the user in a custom flow, which leads to an issue where FAPI queries the wrong enterprise connection:

https://github.com/user-attachments/assets/3ba67507-1ec8-4861-9350-fd13f4abc3bc

Our `SignInStart` component always creates a new `SignIn`; therefore, the flow above works smoothly: https://github.com/clerk/javascript/blob/e1820558624c61f5e76e2d8e1c8132a844005a5a/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx#L299-L312 

#### With fix 

This PR introduces a new `continueSignIn` option for `signIn.authenticateWithRedirect`, which defaults to `false`, therefore it'll create a new `SignIn` object on every call. 

Within `SignInStart`, we pass `continueSignIn` as `false` in order to reuse the current `SignIn` created, in order to avoid calling `signIn.create` twice.

https://github.com/user-attachments/assets/bf2ccb59-b165-423f-8cbb-a4a912269d78

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/spicy-toes-behave.md
+++ b/.changeset/spicy-toes-behave.md
@@ -1,0 +1,7 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Fixes stale `SignIn` object on `authenticateWithRedirect` for `saml` and `enterprise_sso` custom flows
+
+Previously, the same connection identifier would be used on every `authenticateWithRedirect` call leading to redirecting to the wrong identity provider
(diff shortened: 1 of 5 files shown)


LauraBeatris:
The issue still remains that this PR is super stale since we merged changes for `authenticateWithRedirect` in the latest months. I'll raise another one today to fix it.

PR #5133, opened 2025-02-11
fix(clerk-js): Broadcast sign out to all tabs on long lived clients

panteliselef:
## Description

Seems like our broadcasting logic inside `setActive` really depended on having the client being deleted and so that `updateClient` would never be called to update Clerk.user and Clerk.session.

Since Long lived clients, we are no longer deleting the client, which means `updateClient` would run before `setActive` causing the broadcasting logic to always be skipped.

**The solution** the PR suggests is to explicitly broadcast when `signOut()` is called. As in improvement, we should broadcast the sign out event when `user.delete()`  gets called.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/resources/Client.ts
```diff
@@ -83,7 +83,16 @@ export class Client extends BaseResource implements ClientResource {
   removeSessions(): Promise<ClientResource> {
     return this._baseDelete({
       path: this.path() + '/sessions',
-    }) as unknown as Promise<ClientResource>;
+      /**
+       * Skipping updating the client matches the behaviour of `client.destroy`, which allows broadcasting a sign-out event,
+       * and delays emitting until `setActive` is called within `Clerk.signOut()`
+       * TODO: Remove this property while persisting the desired outcome
+       */
+      skipUpdateClient: true,
```
panteliselef, line 91: This simply restores the previous behaviour and it does not solve the core problem. The issue is that I am unaware of whether the the previous behaviour was intentional or a byproduct of complex state.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -878,14 +881,6 @@ export class Clerk implements ClerkInterface {
...
     await onBeforeSetActive();
 
-    // If this.session exists, then signOut was triggered by the current tab
-    // and should emit. Other tabs should not emit the same event again
-    const shouldSignOutSession = this.session && newSession === null;
-    if (shouldSignOutSession) {
-      this.#broadcastSignOutEvent();
-      eventBus.dispatch(events.TokenUpdate, { token: null });
-    }
-
```
panteliselef, line 888: Since we are no longer delete the client on sign out, `this.session` would always be null, causing broadcasting to never occur.

Also removing `eventBus.dispatch(events.TokenUpdate, { token: null })` does not cause issues, because the code a few lines below will handle it appropriately.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2100,7 +2099,7 @@ export class Clerk implements ClerkInterface {
     }
   };
 
-  #broadcastSignOutEvent = () => {
+  public __internal_broadcastSignOutEvent = () => {
```
panteliselef, line 2102: Alternatively we could use the event bus

brkalow: 👍 let's use the event bus, that way we can avoid a new Clerk method.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1534,6 +1529,7 @@ export class Clerk implements ClerkInterface {
     });
   };
 
+  // TODO: Deprecate this one, and mark it as internal. Is there actual benefit for external developers to use this ? Should they ever reach for it ?
```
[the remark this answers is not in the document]
brkalow: Most likely not, it should really be something that's handled internally.

packages/clerk-js/src/core/resources/Client.ts
```diff
@@ -83,7 +83,10 @@ export class Client extends BaseResource implements ClientResource {
   removeSessions(): Promise<ClientResource> {
     return this._baseDelete({
       path: this.path() + '/sessions',
-    }) as unknown as Promise<ClientResource>;
+    }).then(e => {
+      SessionTokenCache.clear();
```
brkalow, line 87: I don't think this is the right place for this call. What's the desired behavior? `SessionTokenCache` is cleared on sign out?

packages/clerk-js/src/core/resources/User.ts
```diff
@@ -241,7 +241,10 @@ export class User extends BaseResource implements UserResource {
   };
 
   delete = (): Promise<void> => {
-    return this._baseDelete({ path: '/me' });
+    return this._baseDelete({ path: '/me' }).then(res => {
+      User.clerk.__internal_broadcastSignOutEvent();
```
brkalow, line 245: this would be much better with the event emitter in use, no unnecessary awareness of `Clerk` here.

packages/clerk-js/src/core/resources/Client.ts
panteliselef: Correct, clearing the cache on sign out.

We're calling `SessionTokenCache.clear()` on `Client.destroy()`, on `Session.end()`, and on `Session.remove()`. Giving the fact that we clear the cache on an individual session removal, i think we should do the same when removing all of them.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -374,6 +374,9 @@ export class Clerk implements ClerkInterface {
 
     const handleSetActive = () => {
       const signOutCallback = typeof callbackOrOptions === 'function' ? callbackOrOptions : undefined;
+
+      // Notify other tabs that user is signing out.
+      eventBus.dispatch(events.UserSignOut, null);
```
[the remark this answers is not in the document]
panteliselef: Previously inside setActive it was one of the first things that got called. And since by that point we have already cleared cookies seems appropriate.

packages/clerk-js/src/core/events.ts
```diff
@@ -2,6 +2,7 @@ import type { TokenResource } from '@clerk/types';
 
 export const events = {
   TokenUpdate: 'token:update',
+  UserSignOut: 'user:signOut',
```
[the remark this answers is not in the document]
panteliselef: I was thinking the same, I added the scope to respect the pattern. Since it does not do any harm, I think I'll leave it as is.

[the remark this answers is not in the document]
brkalow: Fair enough, I wouldn't say 1 event indicates a pattern though 😉

brkalow (approved):

PR #5136, opened 2025-02-11
feat(clerk-js): Handle new session `pending` status as authenticated state

LauraBeatris:
### Context

Introducing a new FAPI session status: `pending`. It builds a fundamental layer for the after-auth project and the future concept of tasks, eg: Forcing to select an organization after sign-in. 

Previously, only `active` sessions were considered to be in an authenticated state. Now, we're introducing a new status assigned to the user session after a successful authentication process but when there are pending tasks.

### Next steps  

These changes do not introduce new behavior on helpers / AIO components, neither breaking changes, as it handles `pending` as an authenticated state, pairing the same functionality as `active`

Next PRs will start introducing the concept of pending tasks and enforcing resolution upon after-auth. 

### Developer-facing changes 

Once these clerk-js get served, developers shouldn't have to worry about changing their app logic or breaking changes. However, some interface changes are preparing the DX for the next steps mentioned above:

**Unifying an signed-in state check based on the session status with `Clerk.isSignedIn` or `Clerk.client.isSignedIn`** 

For custom flows, `Clerk.user` shouldn't be used to determine if the user has fully authenticated or not, and also, it shouldn't be necessary for developers to explicitly check against the session as it's prone for breaking changes, therefore we're abstracting this behind a new property.

````diff
- if (Clerk.session.status === 'active') {
+ if (Clerk.isSignedIn) {
  // Mount user button component
  document.getElementById('signed-in').innerHTML = `
    <div id="user-button"></div>
  `

  const userbuttonDiv = document.getElementById('user-button')

  clerk.mountUserButton(userbuttonDiv)
} else {
````

**Deprecating `activeSessions` in favor of `signedInSessions`**

Deprecating explicit checks against "active" sessions in favor of a generic property that expresses the "signed-in" state instead, since in the future, we might add other types of session statuses for different levels of user verification as we're doing now with after-auth.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/proud-cycles-roll.md
+++ b/.changeset/proud-cycles-roll.md
@@ -3,20 +3,4 @@
 ---
 
 - Initialize new `pending` session status as an signed-in state
-- Add new `Clerk.isSignedIn` property to provide a more explicit way to check authentication status, replacing the previous `Clerk.user` check
-
-```diff
-- if (Clerk.user) {
-+ if (Clerk.isSignedIn) {
-  // Mount user button component
-  document.getElementById('signed-in').innerHTML = `
-    <div id="user-button"></div>
-  `
-
-  const userbuttonDiv = document.getElementById('user-button')
-
-  clerk.mountUserButton(userbuttonDiv)
-} else {
- // ...
-}
-```
+- Add new `Clerk.isSignedIn` property to provide a more explicit way to check authentication status
--- a/.changeset/proud-cycles-roll.md
+++ b/.changeset/proud-cycles-roll.md
@@ -1,0 +1,33 @@
+---
+'@clerk/clerk-js': minor
+'@clerk/elements': patch
+'@clerk/shared': patch
+'@clerk/astro': patch
+'@clerk/clerk-react': patch
+'@clerk/types': patch
+'@clerk/clerk-expo': patch
+'@clerk/vue': patch
+---
+
+### @clerk/clerk-js
+
+- Introduce a new client session status: `pending`
+- Handle and initialize a `pending` session as an authenticated state, similarity as `active`
+
+[TODO - Add more details here on DX changes such as `activeSessions` deprecation after tacking code review feedback]
+
+```diff
+- if (Clerk.user) {
++ if (clerk.hasAuthenticatedClient) {
+  // Mount user button component
+  document.getElementById('signed-in').innerHTML = `
+    <div id="user-button"></div>
+  `
+
+  const userbuttonDiv = document.getElementById('user-button')
+
+  clerk.mountUserButton(userbuttonDiv)
+} else {
+ // ...
+}
+```
--- a/.changeset/proud-cycles-roll.md
+++ b/.changeset/proud-cycles-roll.md
@@ -1,20 +1,9 @@
 ---
 '@clerk/clerk-js': minor
-'@clerk/elements': patch
-'@clerk/shared': patch
-'@clerk/astro': patch
-'@clerk/clerk-react': patch
-'@clerk/types': patch
-'@clerk/clerk-expo': patch
-'@clerk/vue': patch
 ---
 
-### @clerk/clerk-js
-
 - Introduce a new client session status: `pending`
-- Handle and initialize a `pending` session as an authenticated state, similarity as `active`
-
-[TODO - Add more details here on DX changes such as `activeSessions` deprecation after tacking code review feedback]
+- Initialize a `pending` session as an authenticated state
 
 ```diff
 - if (Clerk.user) {
--- a/.changeset/proud-cycles-roll.md
+++ b/.changeset/proud-cycles-roll.md
@@ -3,4 +3,4 @@
 ---
 
 - Initialize new `pending` session status as an signed-in state
-- Add new `Clerk.isSignedIn` property to provide a more explicit way to check authentication status
+- Deprecate `Clerk.client.activeSessions` in favor of `Clerk.client.signedInSessions`
--- a/.changeset/proud-cycles-roll.md
+++ b/.changeset/proud-cycles-roll.md
@@ -4,3 +4,17 @@
 
 - Initialize new `pending` session status as an signed-in state
 - Deprecate `Clerk.client.activeSessions` in favor of `Clerk.client.signedInSessions`
+- Introduce `Clerk.isSignedIn` property as an explicit signed-in state check, instead of `!!Clerk.session` or `!!Clerk.user`:
+
+```ts
+- if (Clerk.user) {
++ if (Clerk.isSignedIn) {
+  // Mount user button component
+  document.getElementById('signed-in').innerHTML = `
+    <div id="user-button"></div>
+  `
+
+  const userbuttonDiv = document.getElementById('user-button')
+
+  clerk.mountUserButton(userbuttonDiv)
+}
--- a/.changeset/proud-cycles-roll.md
+++ b/.changeset/proud-cycles-roll.md
@@ -2,12 +2,12 @@
 '@clerk/clerk-js': minor
 ---
 
-- Introduce a new client session status: `pending`
-- Initialize a `pending` session as an authenticated state
+- Initialize new `pending` session status as an signed-in state
+- Add new `Clerk.isSignedIn` property to provide a more explicit way to check authentication status, replacing the previous `Clerk.user` check
 
 ```diff
 - if (Clerk.user) {
-+ if (clerk.hasAuthenticatedClient) {
++ if (Clerk.isSignedIn) {
   // Mount user button component
   document.getElementById('signed-in').innerHTML = `
     <div id="user-button"></div>
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -289,7 +289,7 @@ export class Clerk implements ClerkInterface {
   }
 
   get isSignedIn(): boolean {
-    return !!this.client?.isSignedIn;
+    return !!this.session;
   }
 
   public constructor(key: string, options?: DomainOrProxyUrl) {
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -10,7 +10,7 @@ import { addClerkPrefix, isAbsoluteUrl, stripScheme } from '@clerk/shared/url';
 import { handleValueOrFn, noop } from '@clerk/shared/utils';
 import type {
   __internal_UserVerificationModalProps,
-  ActiveSessionResource,
+  AuthenticatedSessionResource,
   AuthenticateWithCoinbaseWalletParams,
   AuthenticateWithGoogleOneTapParams,
   AuthenticateWithMetamaskParams,
@@ -164,7 +164,7 @@ export class Clerk implements ClerkInterface {
   };
 
   public client: ClientResource | undefined;
-  public session: ActiveSessionResource | null | undefined;
+  public session: AuthenticatedSessionResource | null | undefined;
   public organization: OrganizationResource | null | undefined;
   public user: UserResource | null | undefined;
   public __internal_country?: string | null;
@@ -387,7 +387,7 @@ export class Clerk implements ClerkInterface {
       });
     };
 
-    if (!opts.sessionId || this.client.activeSessions.length === 1) {
+    if (!opts.sessionId || this.client.authenticatedSessions.length === 1) {
       if (this.#options.experimental?.persistClient ?? true) {
         await this.client.removeSessions();
       } else {
@@ -397,7 +397,7 @@ export class Clerk implements ClerkInterface {
       return handleSetActive();
     }
 
-    const session = this.client.activeSessions.find(s => s.id === opts.sessionId);
+    const session = this.client.authenticatedSessions.find(s => s.id === opts.sessionId);
     const shouldSignOutCurrent = session?.id && this.session?.id === session.id;
     await session?.remove();
     if (shouldSignOutCurrent) {
@@ -877,12 +877,12 @@ export class Clerk implements ClerkInterface {
         : noop;
 
     if (typeof session === 'string') {
-      session = (this.client.sessions.find(x => x.id === session) as ActiveSessionResource) || null;
+      session = (this.client.sessions.find(x => x.id === session) as AuthenticatedSessionResource) || null;
     }
 
     let newSession = session === undefined ? this.session : session;
 
-    // At this point, the `session` variable should contain either an `ActiveSessionResource`
+    // At this point, the `session` variable should contain either an `AuthenticatedSessionResource`
     // ,`null` or `undefined`.
     // We now want to set the last active organization id on that session (if it exists).
     // However, if the `organization` parameter is not given (i.e. `undefined`), we want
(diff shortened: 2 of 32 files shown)

### CODE REVIEW

File: packages/clerk-js/src/core/clerk.ts [hunk 1]
```diff
@@ -286,6 +286,10 @@ export class Clerk implements ClerkInterface {
     return this.#options[key];
   }
 
+  get hasAuthenticatedClient(): boolean {
```
LauraBeatris, line 289: Note for reviewer: The goal of this property is to deprecate/do not recommend the use of `Clerk.user` for custom flows to determine the authenticate state 

Developers shouldn't care whether the session is `pending` or `active` as well, and abstracting this away behind a property should make it more scalable if we introduce new statuses for a similar purpose later

Still, open for feedback on the nomenclature here

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2018,7 +2022,7 @@ export class Clerk implements ClerkInterface {
   };
 
   // TODO: Be more conservative about touches. Throttle, don't touch when only one user, etc
-  #touchLastActiveSession = async (session?: ActiveSessionResource | null): Promise<void> => {
+  #touchCurrentSession = async (session?: AuthenticatedSessionResource | null): Promise<void> => {
```
LauraBeatris, line 2025: Renaming away from "last active", as it resembles a lot from the `active` status, and we're trying to `/touch` against the last current/selected session on the client here

packages/types/src/session.ts
```diff
@@ -130,11 +130,29 @@ export interface SessionResource extends ClerkResource {
...
+export interface PendingSessionResource extends SessionResource {
+  status: 'pending';
+  user: UserResource;
+}
+
+/**
+ * Represents session resources for users who have completed
+ * the full authentication flow.
+ */
+export type AuthenticatedSessionResource = ActiveSessionResource | PendingSessionResource;
```
LauraBeatris, line 154: Thought was better to add some JSDocs explaining the actual meaning and difference between those statuses, feel free to leave more feedback on it!

.changeset/proud-cycles-roll.md
```diff
@@ -0,0 +1,14 @@
+---
+'@clerk/clerk-js': minor
```
LauraBeatris, line 2: Also thought of bumping clerk-js as a minor release, but leaving the others as patches, with the reasoning that the main functionally is on clerk-ks and the other packages are just patching on top of it

packages/types/src/session.ts
```diff
@@ -130,11 +130,29 @@ export interface SessionResource extends ClerkResource {
   __internal_toSnapshot: () => SessionJSONSnapshot;
 }
 
+/**
+ * Represents a session resource that has completed all pending tasks
```
izaaklauer, line 134: It might be too soon to document the concept of tasks!

packages/clerk-js/src/core/clerk.ts
izaaklauer: I think that `authenticated` has the right feel - it implies that the user has completed the authentication steps, which covers both active and pending.

packages/elements/src/react/sign-in/choose-session/choose-session.tsx
```diff
@@ -8,8 +8,8 @@ import {
   SignInActiveSessionContext,
   type SignInActiveSessionListItem,
   useSignInActiveSessionContext,
-  useSignInActiveSessionList,
   useSignInChooseSessionIsActive,
+  useSignInSessionList,
```
izaaklauer, line 12: Why not `useSignInAuthenticatedSessionList`? Do we want this to return revoked sessions?

izaaklauer (approved):
Nice work! Naming is the trickiest part, so getting that out of the way early is great. 

I like the split here of the old Active into "Authenticated" (for session statuses that includes `pending`) and "Current" (for the "setActive" meaning of active - making the session current on the browser/client).

packages/clerk-js/src/core/clerk.ts
panteliselef, line 289: I propose `Clerk.isAuthenticated`, i don't like having "client" in the name, since as a developer using Clerk that term is unfamiliar to me (compared to user and session).

packages/clerk-js/src/core/resources/Client.ts
```diff
@@ -53,10 +54,17 @@ export class Client extends BaseResource implements ClientResource {
...
   }
 
+  /**
+   * @deprecated Use `authenticatedSessions` instead
+   */
   get activeSessions(): ActiveSessionResource[] {
     return this.sessions.filter(s => s.status === 'active') as ActiveSessionResource[];
   }
 
+  get authenticatedSessions(): AuthenticatedSessionResource[] {
```
panteliselef, line 64: Curious, would `authenticatedSessions` be in conflict with sessions from an anonymous login ? Are you authenticated when you use anonymous login ? (thinking out loud here, you can ignore)

.changeset/proud-cycles-roll.md
```diff
@@ -0,0 +1,22 @@
...
+---
+
+- Introduce a new client session status: `pending`
+- Initialize a `pending` session as an authenticated state
+
+```diff
+- if (Clerk.user) {
++ if (clerk.hasAuthenticatedClient) {
```
panteliselef, line 10:
```suggestion
+ if (Clerk.hasAuthenticatedClient) {
```

packages/clerk-js/src/core/clerk.ts
panteliselef: > The goal of this property is to deprecate/do not recommend the use of Clerk.user for custom flows to determine the authenticate state

Is this custom flows only or in their codebase overall ?

packages/types/src/session.ts
LauraBeatris: > That might be worth the tradeoff of making sure we don't forget to add this later.

I went with the same mindset when making this change! It's difficult to find a good balance here cause pending is already related with the concept of tasks, so I wanted to clarify what it means and it's difference with active 

I think it's quite rate when developers directly import this type definition, so it's mostly for our own internal DX purposes

packages/clerk-js/src/core/clerk.ts
LauraBeatris: This is for custom flows in general, a lot of apps have `Clerk.user` to determine whether to render `SignIn` or not 

For AIO components, we have redirect guards based on the session itself:

https://github.com/clerk/javascript/blob/c0ecbe41ffde6992e9953fb47f8907ea0f9d2a18/packages/clerk-js/src/utils/componentGuards.ts#L9-L11

And some control components that check for `userId` - but this will be updated on a next PR related to enforcing protection for tasks:

https://github.com/clerk/javascript/blob/c0ecbe41ffde6992e9953fb47f8907ea0f9d2a18/packages/astro/src/react/controlComponents.tsx#L103-L105

LauraBeatris: I agree! I tried that out first but was on the fence about whether to mention "client"

packages/clerk-js/src/core/resources/Client.ts
LauraBeatris: > Are you authenticated when you use anonymous login ?

I'd expect not, cause the user haven't gone through "identity verification" in a way

packages/elements/src/react/sign-in/choose-session/choose-session.tsx
LauraBeatris: My change was mainly around not having the hook name tied to a certain statuses but rely on the return type 

This hook is mainly used internally for `@clerk/elements` components, so I see the reasoning of leaving it more explicit for our management

packages/clerk-js/src/core/clerk.ts
brkalow: We use `isSignedIn` elsewhere, I would say we stick with that unless we feel strongly that `isAuthenticated()` is a better representation.

packages/clerk-js/src/core/resources/Client.ts
```diff
@@ -110,6 +118,10 @@ export class Client extends BaseResource implements ClientResource {
     return this._basePostBypass({ body: params, path: this.path() + '/verify' });
   }
 
+  get hasAuthenticated() {
```
brkalow, line 121: Would recommend present tense instead of past-tense:
```suggestion
  get isAuthenticated() {
```

packages/clerk-js/src/core/clerk.ts
LauraBeatris: I quite like that! Here's a comparison of how it'd play out with the current `SignedIn` control component vs a custom flow

```ts
<SignedIn>
  <p> I have successfully signed up but I still have pending stuff </p>
</SignedIn>
```
```ts
if (!Clerk.isSignedIn){
  mountSignIn() 
} else {
  if (Clerk.session.tasks) { 
    // ...
  }
}
```

LauraBeatris: I think the same nomenclature should be extended to the session arrays, so would be something like: `signedInSessions`

packages/expo/src/provider/singleton/createClerkInstance.ts
```diff
@@ -123,8 +123,8 @@ export function createClerkInstance(ClerkClass: typeof Clerk) {
             if (client) {
               void ClientResourceCache.save(client.__internal_toSnapshot());
               if (client.lastActiveSessionId) {
-                const lastActiveSession = client.activeSessions.find(s => s.id === client.lastActiveSessionId);
-                const token = lastActiveSession?.lastActiveToken?.getRawString();
+                const currentSessionToken = client.authenticatedSessions.find(s => s.id === client.lastActiveSessionId);
+                const token = currentSessionToken?.lastActiveToken?.getRawString();
```
brkalow, line 127:
```suggestion
                const currentSession = client.authenticatedSessions.find(s => s.id === client.lastActiveSessionId);
                const token = currentSession?.lastActiveToken?.getRawString();
```

packages/shared/src/react/contexts.tsx
```diff
@@ -17,7 +17,7 @@ import { createContextAndHook } from './hooks/createContextAndHook';
 const [ClerkInstanceContext, useClerkInstanceContext] = createContextAndHook<LoadedClerk>('ClerkInstanceContext');
 const [UserContext, useUserContext] = createContextAndHook<UserResource | null | undefined>('UserContext');
 const [ClientContext, useClientContext] = createContextAndHook<ClientResource | null | undefined>('ClientContext');
-const [SessionContext, useSessionContext] = createContextAndHook<ActiveSessionResource | null | undefined>(
+const [SessionContext, useSessionContext] = createContextAndHook<AuthenticatedSessionResource | null | undefined>(
```
brkalow, line 20: I think this might be a breaking change, if people are importing `ActiveSessionResource` from `@clerk/types` directly and depending on the return type of `useSession()`.

packages/clerk-js/src/core/resources/Client.ts
octoper: I agree with Bryce here, present tense seems a bit better

packages/clerk-js/src/ui/components/ImpersonationFab/ImpersonationFab.tsx
```diff
@@ -66,7 +66,7 @@ const FabContent = ({ title, signOutText }: FabContentProps) => {
   const { otherSessions } = useMultipleSessions({ user });
   const { navigateAfterSignOut, navigateAfterMultiSessionSingleSignOutUrl } = useSignOutContext();
 
-  const handleSignOutSessionClicked = (session: ActiveSessionResource) => () => {
+  const handleSignOutSessionClicked = (session: SignedInSessionResource) => () => {
```
LauraBeatris, line 69: It's much more idiomatic now! Thanks for the feedback @brkalow 🫡

packages/clerk-js/src/core/clerk.ts
LauraBeatris: Updated on https://github.com/clerk/javascript/pull/5136/commits/add33132e33e5f1e1b4bde60b522c5cbbe571039 

Should be `Clerk.isSignedIn` or `Clerk.session.signedInSessions` now

panteliselef (approved):
Nice work

.changeset/proud-cycles-roll.md
```diff
@@ -0,0 +1,22 @@
+---
+'@clerk/clerk-js': minor
+---
+
+- Initialize new `pending` session status as an signed-in state
+- Add new `Clerk.isSignedIn` property to provide a more explicit way to check authentication status, replacing the previous `Clerk.user` check
+
+```diff
+- if (Clerk.user) {
++ if (Clerk.isSignedIn) {
```
nikosdouvlis, line 10: The `isSignedIn` property makes sense to me! 

In the PR description, you mention that: 
> Clerk.user shouldn't be used to determine if the user has fully authenticated or not

But the logic in the `Client` resource tells me that `isSignedIn` is going to be true if the current session is `active` or `pending`, meaning that currently, `isSignedIn` cannot be used to check if the user is fully authenticated or not. 

Is this intentional? If yes, could you please provide an example where `Clerk.isSignedIn !== !!Clerk.user`?

LauraBeatris: > meaning that currently, isSignedIn cannot be used to check if the user is fully authenticated or not.

The user has fully authenticated even with a `pending` status. Still, they have pending tasks to complete, eg: After sign-in, complete all factors but FAPI returns a pending session for tasks such as having to select an org. 

> Is this intentional? If yes, could you please provide an example where Clerk.isSignedIn !== !!Clerk.user

Currently, they are the same. This PR treats `pending` exactly like `active` to maintain current functionality and to incrementally add protections for it - the feature is gated on FAPI and toggled via Dashboard, therefore it's an opt-in behavior.

This property was added to avoid relying only on the `user` object, or `session.status == 'active'` for our internal checks. 

Once we introduce tasks (https://github.com/clerk/javascript/pull/5170) -> I'd add another property that specifically checks if the user has a session that resolved all pending tasks, something like:
```ts
// Would resolve to `false` for session.status === 'pending' 
// Would resolve to `true` for session.status === 'active'
if (Clerk.hasValidSession) { 
  clerk.mountUserButton(userbuttonDiv)
} else {
  clerk.mountSignIn(signInDiv)
}
```

As a syntax sugar so that developers don't have to manually check for the session statuses on custom flows as well.

144mdgross:
Hello! Sorry if this is the wrong place to ask but I wasn't able to get my discord account working. I begin getting type errors (error TS2322) when updating to version [4.47.0](https://github.com/clerk/javascript/blob/main/packages/types/CHANGELOG.md#4470) of types. Is this expected and is it resolved by an update to the corresponding clerk packages? We are using `"@clerk/remix": "4.1.1",` in conjunction with `@clerk/types": "4.6.0` currently without issue.

```
error TS2322: Type 'SignedInSessionResource | null | undefined' is not assignable to type 'ActiveSessionResource | null | undefined'.
  Type 'PendingSessionResource' is not assignable to type 'ActiveSessionResource'.
    Types of property 'status' are incompatible.
      Type '"pending"' is not assignable to type '"active"'.```

PR #5138, opened 2025-02-12
fix(upgrade): Remove unused getMajorVersion function

LekoArts:
## Description

A user reported this error while running the upgrade tool on their repo:

```shell
Cannot read properties of null (reading '0')

node_modules/@clerk/upgrade/dist/util/get-clerk-version.js:6:43
```

This comes from these lines:

```js
const clerk = pkg.dependencies.clerk;
return clerk ? semverRegex().exec(clerk)[0][0] : false;
```

`clerk` will always be `undefined` because `read-pkg` doesn't normalize namespaced packages in such a way. It e.g. returns back `dependencies: { '@clerk/nextjs': '^5.11.2', 'read-pkg': '^9.0.1' },`.
But for whatever reason for this user the regex is executed.

Anyways, the function is unused anyways and the desired functionality is already fulfilled by the `guessFrameworks` function.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/upgrade/src/app.js
+++ b/packages/upgrade/src/app.js
@@ -6,7 +6,6 @@ import { Header } from './components/Header.js';
 import { Scan } from './components/Scan.js';
 import { SDKWorkflow } from './components/SDKWorkflow.js';
 import SDKS from './constants/sdks.js';
-import { getClerkMajorVersion } from './util/get-clerk-version.js';
 import guessFrameworks from './util/guess-framework.js';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/upgrade/src/app.js
```diff
@@ -34,15 +33,13 @@ export default function App(props) {
...
   const [fromVersion, setFromVersion] = useState(props.fromVersion);
-  const [fromVersionGuessAttempted, setFromVersionGuessAttempted] = useState(false);
 
   const [toVersion, setToVersion] = useState(props.toVersion);
   const [dir, setDir] = useState(props.dir);
   const [ignore, setIgnore] = useState(props.ignore ?? []);
   const [configComplete, setConfigComplete] = useState(false);
   const [configVerified, setConfigVerified] = useState(false);
   const [uuid, setUuid] = useState();
-  let fromVersionGuess = false;
```
LekoArts, line 45: This wasn't used anywhere, so dead code!

nikosdouvlis (approved):

PR #5140, opened 2025-02-12
chore(astro): Remove deprecated `as` prop from button components

wobsoriano:
## Description

It's been 5 months since we began deprecating the `as` prop in favor of `asChild` for customizing our unstyled components (`<SignInButton />`, `<SignUpButton />`, `<SignOutButton />`). As documented in our [component guides](https://clerk.com/docs/components/unstyled/sign-in-button#custom-usage), the `asChild` prop is the recommended approach for passing custom buttons. This PR completes the deprecation by removing the `as` prop entirely.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/smart-rabbits-reflect.md
+++ b/.changeset/smart-rabbits-reflect.md
@@ -1,0 +1,17 @@
+---
+"@clerk/astro": minor
+---
+
+Remove deprecated `as` prop in the `<SignInButton />`, `<SignOutButton />`, and `<SignUpButton>` components. Please use the `asChild` prop if you are passing children to it.
+
+Example:
+
+```astro
+---
+import { SignInButton } from '@clerk/astro/components'
+---
+
+<SignInButton asChild>
+  <button>Custom sign in button</button>
+</SignInButton>
+```
--- a/packages/astro/src/astro-components/unstyled/SignInButton.astro
+++ b/packages/astro/src/astro-components/unstyled/SignInButton.astro
@@ -1,21 +1,14 @@
 ---
-import type { HTMLTag, Polymorphic } from 'astro/types'
-import type { SignInButtonProps } from '@clerk/types'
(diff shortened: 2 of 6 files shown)


.changeset/smart-rabbits-reflect.md
```diff
@@ -0,0 +1,17 @@
+---
+"@clerk/astro": minor
```
anagstef, line 2: 🤔 Isn't this a breaking change?

wobsoriano: Good question! Hmmm originally when the `as` prop was added, we immediately did a follow up and introduced the `asChild` prop and documented that instead.

If we're going major on this one, I think we can skip this PR for now as it will be a waste of a major version bump. We can include this change in our next major release 👍🏼

panteliselef: i think this is a breaking change, we should schedule for the next major instead.

packages/astro/src/astro-components/unstyled/SignInButton.astro
```diff
@@ -44,9 +37,9 @@ if (asChild) {
   asChild ? (
     <Fragment set:html={htmlElement} />
   ) : (
-    <Tag {...props} data-clerk-unstyled-id={safeId}>
+    <button {...props} data-clerk-unstyled-id={safeId}>
```
panteliselef, line 40: with `asChild` aren't we loosing `data-clerk-unstyled-id` ? is that ok ?

.changeset/smart-rabbits-reflect.md
wobsoriano: Yeah, I will close this PR for now and revisit on our next major!

PR #5142, opened 2025-02-12
feat(shared): Cast `EmailLinkErrorCode` as const

alexcarpenter:
## Description

Cast EmailLinkErrorCode as const to help users with types within custom flows.

```tsx
import { EmailLinkErrorCode } from '@clerk/nextjs/errors'

type EmailLinkError = typeof EmailLinkErrorCode[keyof typeof EmailLinkErrorCode];

const getStatus = (err: EmailLinkError): EmailLinkError => {
  if (err === EmailLinkErrorCode.Expired) {
    return 'expired'
  }

  if (err === EmailLinkErrorCode.ClientMismatch) {
    return 'client_mismatch'
  }

  return 'failed'
}
```

Resolves SDK-2055

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/src/error.ts
+++ b/packages/shared/src/error.ts
@@ -214,7 +214,7 @@ export const EmailLinkErrorCode = {
   Expired: 'expired',
   Failed: 'failed',
   ClientMismatch: 'client_mismatch',
-};
+} as const;
 
 const DefaultMessages = Object.freeze({
   InvalidProxyUrlErrorMessage: `The proxyUrl passed to Clerk is invalid. The expected value for proxyUrl is an absolute URL or a relative path with a leading '/'. (key={{url}})`,


dstaley (approved):
lgtm but this is technically a breaking change in the most technical sense given that we're narrowing `EmailLinkErrorCode` from `Record<string, string>` to a more restrictive type. I think this is _probably_ fine, but there's technically user code that could break as a result of this change.

packages/nextjs/src/client-boundary/hooks.ts
```diff
@@ -20,6 +20,7 @@ export {
   isKnownError,
   isMetamaskError,
   EmailLinkErrorCode,
+  EmailLinkErrorCodeStatus,
```
panteliselef, line 23: We need to re-export it from `src/errors.ts`

.changeset/dry-keys-look.md
```diff
@@ -0,0 +1,15 @@
...
+'@clerk/shared': minor
+'@clerk/clerk-react': minor
+'@clerk/remix': minor
+---
+
+Introduce `EmailLinkErrorCodeStatus` to support users in custom flows and mark `EmailLinkErrorCode` as deprecated.
+
+```diff
+- import { EmailLinkErrorCode } from '@clerk/nextjs/errors'
++ import { EmailLinkErrorCodeStatus } from '@clerk/nextjs/errors'
```
panteliselef, line 14: We need to re-export `EmailLinkErrorCodeStatus` from src/errors.ts

panteliselef (approved):
Nice work

PR #5144, opened 2025-02-13
feat(shared): Replace `callWithRetry` and `runWithExponentialBackoff` with `retry`

nikosdouvlis:
## Description

- Replace `callWithRetry` with `runWithExponentialBackOff` so we only have a single retry mechanism in use 
- Rename `runWithExponentialBackOff` to `retry` for clarity 
- Add the following config options to `runWithExponentialBackOff` :
    - `retryImmediately` (boolean): Controls whether the helper should retry the operation immediately. If true, the exponential backoff delay will start only once this retry is over. Defaults to true.
    - `jitter` (boolean): If true, the intervals will be multiplied by a factor of `[1,2]`. Defaults to true.
- Increase unit test coverage (retry.test.ts)

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/shared/src/retry.ts
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Controls whether the helper should retry the operation immediately once before applying exponential backoff.
+   * @default true
+   */
+  retryImmediatelyOnce: boolean;
+  /**
+   * The delay for the immediate retry.
+   * @default 100
+   */
+  retryImmediatelyDelay: Milliseconds;
```
panteliselef, line 37: immediatelyDelay  sounds confusing, is it immediate or delayed ? How is it different from firstDelay ?

packages/shared/src/__tests__/retry.test.ts
```diff
@@ -0,0 +1,209 @@
...
+    expect(attempts).toBe(1);
+    await jest.advanceTimersByTimeAsync(100);
+    expect(attempts).toBe(2);
+    await jest.advanceTimersByTimeAsync(400);
+    expect(attempts).toBe(3);
+    await jest.advanceTimersByTimeAsync(1600);
+    expect(attempts).toBe(4);
+  });
+
+  test('applies jitter when enabled', async () => {
```
LauraBeatris, line 173:
```suggestion
  test('applies jitter when enabled by default', async () => {
```

Suggestion: Indicate that it's enabled by default on the description, since the `// jitter: true` below was a bit confusing to me if I haven't read the PR description / implementation

.changeset/metal-bobcats-wash.md
```diff
@@ -0,0 +1,38 @@
...
+### @clerk/shared
+- Introduced a new `retry` utility function to replace the deprecated `callWithRetry`.
+- Removed the `callWithRetry` function and its associated tests.
+- Renamed `runWithExponentialBackOff` to `retry` for consistency.
+
+### @clerk/clerk-js
+- Updated internal usage of `callWithRetry` to use the new `retry` function.
+
+### @clerk/backend
+- Updated internal usage of `callWithRetry` to use the new `retry` function.
```
anagstef, line 16: It's better to separate this into 3 different changelog entries, because this will mess the final chengelog.

packages/shared/src/__tests__/retry.test.ts
nikosdouvlis: Great suggestion Laura, will change the test accordingly!

packages/shared/src/retry.ts
nikosdouvlis: Does the JSDoc for `retryImmediatelyOnce` help at all? 
```
  /**
   * Controls whether the helper should retry the operation immediately once before applying exponential backoff.
   * @default true
   */
```

packages/shared/src/retry.ts
```diff
@@ -0,0 +1,108 @@
+type Milliseconds = number;
+
+type RetryOptions = Partial<{
+  /**
+   * The initial delay before the first retry.
+   * @default 125
+   */
+  firstDelay: Milliseconds;
```
LekoArts, line 8: How about `initialDelay`? How will it be different from the "immediate" delay?

packages/shared/src/retry.ts
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * The maximum delay between retries.
+   * If set to 0, the delay will increase indefinitely.
+   * @default 0
+   */
+  maxDelay: Milliseconds;
```
LekoArts, line 14: Only knowing the name (and not reading the jsdoc comment) it's not really clear if this means the max. allowed delay between retries or if this is the max. allowed duration of all retries. Also what happens when this maximum value is passed?

packages/shared/src/retry.ts
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * The multiplier for the exponential backoff.
+   * @default 2
+   */
+  factor: number;
```
LekoArts, line 19: How about `backoffFactor` or something similar?

packages/shared/src/retry.ts
LekoArts, line 37: These two options feel _weird_, I can't put a finger on it but this seems like something that should be only one option and be renamed.

- Retrying immediately already implies that it can only happen once, right? Afterwards it's not immediate anymore
- Technically "immediate" would be a delay of 0 so the current default of 100 is already quite arbitrary. Why should we need to change this from a case by case basis?

So I'd just put those two options into one called `retryImmediate: boolean`

packages/shared/src/retry.ts
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * The delay for the immediate retry.
+   * @default 100
+   */
+  retryImmediatelyDelay: Milliseconds;
+  /**
+   * If true, the intervals will be multiplied by a factor in the range of [1,2].
+   * @default true
+   */
+  jitter: boolean;
```
LekoArts, line 42: Seems technical, how about `randomizeBackoff`?

LekoArts (changes requested):

packages/shared/src/retry.ts
nikosdouvlis: Not a fan of `randomizeBackoff` to be honest - `jitter` is the most common term for random intervals between retries in the context of handling transient errors (all AWS/MS guides etc)

packages/shared/src/retry.ts
nikosdouvlis: Tagging @panteliselef as he was also interested about this - Im also open to suggestions to improve this further. 
There are at least two cases where we might want to immediately retry, before applying the backoff strategy: 
- mission-critical operations that need to be completed as fast as possible 
- transient errors on the network level (think about mobile devices)

This property decouples the immediate retry with the rest of the backoff logic, as its very hard to tune the options using a combination of factor + initialDelay otherwise.

packages/shared/src/retry.ts
nikosdouvlis: Sounds good, will update the code accordingly 👍🏻

packages/shared/src/retry.ts
nikosdouvlis: Updated the comment to indicate that `The delay between retries will never exceed this value.` and updated the prop name to `maxDelayBetweenRetries`

packages/shared/src/retry.ts
nikosdouvlis: Similarly to my comment about `jitter`, `factor` is the most commonly used term for this and I think adding `backoff` doesn't add much in this case

packages/shared/src/retry.ts
nikosdouvlis: I'm going to merge these two into a single `retryImmediate: boolean` with a default delay of `100ms`. If we ever need to change this, we can add a new prop then

nikosdouvlis: Done

nikosdouvlis: Resolved [here](https://github.com/clerk/javascript/pull/5144#discussion_r1955760720)

packages/shared/src/retry.ts
```diff
@@ -0,0 +1,111 @@
...
+   * that caused the error and starts at 1 (as in, this is the 1st, 2nd, nth retry).
+   * @default (error, iterations) => iterations < 5
+   */
+  shouldRetry: (error: unknown, iterations: number) => boolean;
+  /**
+   * Controls whether the helper should retry the operation immediately once before applying exponential backoff.
+   * The delay for the immediate retry is 100ms.
+   * @default true
+   */
+  retryImmediately: boolean;
```
panteliselef, line 34: 🙃 the PR description seems to use the old ones

panteliselef (approved):

packages/shared/src/__tests__/retry.test.ts
```diff
@@ -0,0 +1,208 @@
...
+        factor: 1,
+        jitter: false,
+      },
+    );
+    await jest.advanceTimersByTimeAsync(200);
+    expect(await result).toBe('success');
+    expect(attempts).toBe(2);
+  });
+
+  test('maxDelay prevents delays from growing beyond the limit', async () => {
```
LekoArts, line 38:
```suggestion
  test('maxDelayBetweenRetries prevents delays from growing beyond the limit', async () => {
```

packages/shared/src/__tests__/retry.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    // Run all timer advances before testing the promise
+    await jest.advanceTimersByTimeAsync(100);
+    await jest.advanceTimersByTimeAsync(300);
+    await jest.advanceTimersByTimeAsync(300);
+    await jest.advanceTimersByTimeAsync(300);
+    expect(attempts).toBe(1 + 4);
+    jest.useRealTimers();
+  });
+
+  test('respects firstDelay option', async () => {
```
LekoArts, line 67:
```suggestion
  test('respects initialDelay option', async () => {
```

packages/shared/src/__tests__/retry.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    ).catch(() => {});
+
+    expect(attempts).toBe(1);
+    await jest.advanceTimersByTimeAsync(101);
+    expect(attempts).toBe(2);
+    await jest.advanceTimersByTimeAsync(1000);
+    expect(attempts).toBe(3);
+  });
+
+  test('disables immediate retry when retryImmediatelyOnce is false', async () => {
```
LekoArts, line 106:
```suggestion
  test('disables immediate retry when retryImmediately is false', async () => {
```

LekoArts (approved):

anagstef (approved):

packages/backend/src/tokens/keys.ts
```diff
@@ -137,8 +137,8 @@ export async function loadClerkJWKFromRemote({
         reason: TokenVerificationErrorReason.RemoteJWKFailedToLoad,
       });
     }
-    const fetcher = () => fetchJWKSFromBAPI(apiUrl, secretKey, apiVersion);
-    const { keys } = await callWithRetry<{ keys: JsonWebKeyWithKid[] }>(fetcher);
+    const fetcher = () => fetchJWKSFromBAPI(apiUrl, secretKey, apiVersion) as Promise<{ keys: JsonWebKeyWithKid[] }>;
```
brkalow, line 140: ❓ why do we need to cast here?

brkalow (approved):
👏

PR #5170, opened 2025-02-14
chore(clerk-js,types): Initialize `tasks` on `Session` resource

LauraBeatris:
## Description

Adds the `tasks` property on the session client resource to be leveraged/accessed for displaying in AIO components 

```tsx
Object.keys(session.tasks).map(task => (
   <Route path={task}>
       <Task task={task}  />
   </Route>
)
```

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/types/src/session.ts
+++ b/packages/types/src/session.ts
@@ -96,13 +96,14 @@ export interface SessionResource extends ClerkResource {
   /**
    * Factor Verification Age
    * Each item represents the minutes that have passed since the last time a first or second factor were verified.
-   * [fistFactorAge, secondFactorAge]
+   * [firstFactorAge, secondFactorAge]
    */
-  factorVerificationAge: [fistFactorAge: number, secondFactorAge: number] | null;
+  factorVerificationAge: [firstFactorAge: number, secondFactorAge: number] | null;
   lastActiveToken: TokenResource | null;
   lastActiveOrganizationId: string | null;
   lastActiveAt: Date;
   actor: ActJWTClaim | null;
+  tasks: SessionTasks | null;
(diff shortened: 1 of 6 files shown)


packages/types/src/session.ts
```diff
@@ -169,6 +170,10 @@ export interface PublicUserData {
   userId?: string;
 }
 
+export interface SessionTasks {
+  orgs: Record<string, unknown>;
```
LauraBeatris, line 174: For now, FAPI is returning an empty object by purpose but it might include some additional metadata about the task in the future so I'll update it by then 

The purpose of including it now is to unblock next PRs

packages/types/src/session.ts
```diff
@@ -98,11 +98,12 @@ export interface SessionResource extends ClerkResource {
    * Each item represents the minutes that have passed since the last time a first or second factor were verified.
    * [fistFactorAge, secondFactorAge]
```
alexcarpenter, line 99:
```suggestion
   * [firstFactorAge, secondFactorAge]
```

alexcarpenter (approved):

anagstef (approved):

panteliselef (approved):

LauraBeatris:
Currently holding on this one, cause we've been discussing on FAPI returns a list of tasks instead of an object, so we don't have to manually transform it to entries here when serializing the JSON.

LauraBeatris:
I've just updated it to use an array of session tasks instead

PR #5186, opened 2025-02-17
chore(repo): Replace `Clerk.user` with `Clerk.isSignedIn` on integration tests 

LauraBeatris:
## Description

Blocked by https://github.com/clerk/javascript/pull/5136 - waiting to be deployed (currently stacked branch)

Instead of relying on the `Clerk.user` to determine whether to mount elements for a custom sign-in flow, use the newest `Clerk.isSignedIn`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/integration/templates/express-vite/src/views/index.ejs
+++ b/integration/templates/express-vite/src/views/index.ejs
@@ -20,7 +20,7 @@
         const appEl = document.querySelector('#app');
         const controlStateEl = document.querySelector('#user-state');
 
-        if (Clerk.user) {
+        if (Clerk.isSignedIn) {
           Clerk.mountUserButton(appEl);
           controlStateEl.innerHTML = 'SignedIn';
         } else {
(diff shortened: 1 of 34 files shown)


integration/templates/express-vite/src/views/index.ejs
```diff
@@ -20,7 +20,7 @@
         const appEl = document.querySelector('#app');
         const controlStateEl = document.querySelector('#user-state');
 
-        if (Clerk.user) {
+        if (Clerk.isSignedIn) {
```
panteliselef, line 23: For applications that don't have pending sessions, checking Clerk.user will still be the same as Clerk.isSignedIn, correct ?

LauraBeatris: Correct! We'll start replacing `Clerk.user` for custom flow branches of logic to keep it scalable for both active/pending - the following will be applied to our docs especially

LauraBeatris:
Actually closing this one, realized that it's not super necessary as right now `Clerk.user` still gets resolved even when `Clerk.isSignedIn` is true 

It'd just be required to replace it we were explicitly checking for `session.status === 'active'`

PR #5187, opened 2025-02-17
feat(clerk-js,types): Navigate to session task on after sign-in/sign-up

LauraBeatris:
## Description

Resolves ORGS-562

This PR introduces the redirection behavior after sign-in/sign-up if the session has pending tasks to be solved. 

- Displays the current pending session task as `SignIn/SignUp` routes 
- Introduce navigation to tasks after sign-in/sign-up

#### Developer flows 

For AIO components, it'll work out of the box. Just rendering `<SignIn />` and `<SignUp />` will work, since the task gets appended to the routing path and displayed within each component.

For custom flows, no manual changes are needed to trigger navigation, however, the developer has to handle rendering the respective component on the task route, such as `OrganizationSwitcher` 

The next PR will introduce a custom URL for customing the task route path to eject out of AIO components, eg: Instead of `sign-in/select-organization` - `onboarding/select-organization`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/utils/componentGuards.ts
```diff
@@ -7,7 +7,7 @@ export type ComponentGuard = (
 ) => boolean;
 
 export const sessionExistsAndSingleSessionModeEnabled: ComponentGuard = (clerk, environment) => {
-  return !!(clerk.session && environment?.authConfig.singleSessionMode);
+  return !!(clerk.session?.status === 'active' && environment?.authConfig.singleSessionMode);
```
LauraBeatris, line 10:
```suggestion
  return !!(clerk.hasValidSession && environment?.authConfig.singleSessionMode);
```

This will avoid breaking changes for statuses changes

packages/types/src/session.ts
```diff
@@ -197,7 +198,7 @@ export interface PublicUserData {
 }
 
 export interface SessionTask {
-  key: 'orgs';
```
LauraBeatris, line 200: FAPI was previously returning `orgs` but it got updated to `org`

LauraBeatris:
Would be useful if we could render the task component ahead of time to avoid this glitch on the navigation:

packages/clerk-js/src/ui/components/Task/Task.tsx
```diff
@@ -0,0 +1,30 @@
+import { useSessionContext } from '@clerk/shared/react/index';
+import type { SessionTask } from '@clerk/types';
+import type { ComponentType } from 'react';
+
+import { OrganizationListContext } from '../../contexts';
+import { OrganizationList } from '../OrganizationList';
+
+const TaskRegistry: Record<SessionTask['key'], ComponentType> = {
+  org: () => (
+    <OrganizationListContext.Provider value={{ componentName: 'OrganizationList', hidePersonal: true }}>
```
LauraBeatris, line 10: `hidePersonal` is being directly provided here, but this is going to be updated in a next PR.

We're going to include a property on `Enviroment` to indicate whether an after-auth task is enabled for instance. 

For the use case of forcing to select an org, this is going to be checked once the environment loads, and the org components are going to have `hidePersonal: true` by default.

LauraBeatris:
On my local setup in order to test the whole routing flow, I'm triggering navigation on `setActive`, with a similar logic as for `redirectUrl` 

We want to trigger navigation to the task once the client session gets updated with a task on the `setActive` after-auth (sign-in/sign-up)

However, the `setTransitiveState` method cleans all of the accessors and causes the components to unmount, therefore I changed it to call `#emit` right away

packages/clerk-js/src/utils/componentGuards.ts
```diff
@@ -7,7 +7,7 @@ export type ComponentGuard = (
 ) => boolean;
 
 export const sessionExistsAndSingleSessionModeEnabled: ComponentGuard = (clerk, environment) => {
-  return !!(clerk.session && environment?.authConfig.singleSessionMode);
+  return !!(clerk.isSignedIn && !clerk.session?.hasTask && environment?.authConfig.singleSessionMode);
```
LauraBeatris, line 10: We should also add another redirect guard such as: `hasPendingTasksAndSingleSessionModeEnabled` in order to redirect to the task path.

LauraBeatris: Just realized that this is a problem, if the user has a session with a pending task, and tries to go back to sign-up, it'll error out for single session mode.

The right approach here is to navigate to the task when sign-in mounts

packages/types/src/session.ts
```diff
@@ -197,7 +198,7 @@ export interface PublicUserData {
...
+  key: 'org';
```
panteliselef, line 201: is scope a better name ? or does key represent something else ?

integration/tests/session-tasks.test.ts
```diff
@@ -0,0 +1,35 @@
+import { appConfigs } from '../presets';
+import { testAgainstRunningApps } from '../testUtils';
+
+// TODO ORGS-566 - write integration tests for after-auth flow
+testAgainstRunningApps({ withEnv: [appConfigs.envs.withEmailCodes] })('after-auth flows @generic @nextjs', () => {
```
LauraBeatris, line 5: This is just a placeholder for me to come back later which is helpful to visualize the redirection behavior introduced in this PR

The reason integration tests aren't being implemented is that FAPI changes for that are under development and it could break CI

LauraBeatris:
I'm opening another PR for this as it became stale and I want to validate other approaches for in-component navigation

LauraBeatris:
https://github.com/clerk/javascript/pull/5280/files

PR #5194, opened 2025-02-18
fix(shared): Restore scrollbar on re-verification completion or on canceled/closed

octoper:
## Description

This PR fixes an issue where the re-verification modal does not restores the scrollbar behaviour after it's completed or canceled/closed.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/sweet-forks-care.md
+++ b/.changeset/sweet-forks-care.md
@@ -2,4 +2,4 @@
 "@clerk/shared": patch
 ---
 
-fix(shared): Fixes restoretion of the scrollbar behaviour after re-verification it's completed or canceled/closed.
+Fixes restoration of the scrollbar behaviour after re-verification is completed or canceled/closed.
--- a/.changeset/sweet-forks-care.md
+++ b/.changeset/sweet-forks-care.md
@@ -1,0 +1,5 @@
+---
+"@clerk/shared": patch
+---
+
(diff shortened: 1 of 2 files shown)


panteliselef:
This was definitely working before, I'm curious if some of my recent changes in setActive broke this behaviour. If that is the case, then I would dive into that before reaching for an "onClose" parameter.

octoper:
@panteliselef it's closing but it does not restore the scrollbar behaviour

.changeset/sweet-forks-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/shared": patch
+---
+
+fix(shared): Fixes restoretion of the scrollbar behaviour after re-verification it's completed or canceled/closed.
```
anagstef, line 5:
```suggestion
fix(shared): Fixes restoration of the scrollbar behaviour after re-verification is completed or canceled/closed.
```

anagstef (approved):

panteliselef:
I think this stems from the modal on modal interaction and not the `useReverification` itself. So I think we should investigate more here.

octoper:
This does not happens only with when modal on modal this happens every time

PR #5197, opened 2025-02-18
chore(clerk-react): Align types between Clerk and IsomorphicClerk

brkalow:
## Description

Aligns types between `IsomorphicClerk` and `Clerk` to ensure the implementation remains in sync.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/react/src/isomorphicClerk.ts
+++ b/packages/react/src/isomorphicClerk.ts
@@ -1,6 +1,5 @@
 import { inBrowser } from '@clerk/shared/browser';
 import { loadClerkJsScript } from '@clerk/shared/loadClerkJsScript';
-import type { TelemetryCollector } from '@clerk/shared/telemetry';
 import { handleValueOrFn } from '@clerk/shared/utils';
 import type {
   __internal_UserVerificationModalProps,
@@ -19,7 +18,6 @@ import type {
   GoogleOneTapProps,
   HandleEmailLinkVerificationParams,
   HandleOAuthCallbackParams,
-  InstanceType,
   JoinWaitlistParams,
   ListenerCallback,
   LoadedClerk,
@@ -28,22 +26,16 @@ import type {
   OrganizationResource,
   OrganizationSwitcherProps,
   RedirectOptions,
-  SDKMetadata,
   SetActiveParams,
-  SignedInSessionResource,
   SignInProps,
   SignInRedirectOptions,
   SignInResource,
-  SignOut,
-  SignOutCallback,
-  SignOutOptions,
   SignUpProps,
   SignUpRedirectOptions,
   SignUpResource,
   UnsubscribeCallback,
   UserButtonProps,
   UserProfileProps,
-  UserResource,
   WaitlistProps,
   WaitlistResource,
   Without,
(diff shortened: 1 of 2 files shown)


packages/react/src/isomorphicClerk.ts
```diff
@@ -85,95 +77,17 @@ type MethodName<T> = {
 
 type MethodCallback = () => Promise<unknown> | unknown;
 
+type WithVoidReturn<F extends (...args: any) => any> = (
+  ...args: Parameters<F>
+) => ReturnType<F> extends Promise<infer T> ? Promise<T | void> : ReturnType<F> | void;
+type WithVoidReturnFunctions<T> = {
+  [K in keyof T]: T[K] extends (...args: any) => any ? WithVoidReturn<T[K]> : T[K];
+};
```
brkalow, line 85: Most methods were adjust to add `| void` to the return type, so I've gone ahead and created a helper type to do that automatically.

packages/react/src/isomorphicClerk.ts
```diff
@@ -85,95 +77,17 @@ type MethodName<T> = {
...
-  | 'mountSignUp'
-  | 'mountSignIn'
-  | 'mountUserProfile'
-  | 'mountWaitlist'
-  | 'client'
-  | '__internal_getCachedResources'
-  | '__internal_reloadInitialResources'
-  | '__internal_addNavigationListener'
+  WithVoidReturnFunctions<LoadedClerk>,
+  'client' | '__internal_addNavigationListener' | '__internal_getCachedResources' | '__internal_reloadInitialResources'
```
brkalow, line 89: I'm not sure if we need the `__internal_` methods here, they don't exist today.

packages/react/src/isomorphicClerk.ts
```diff
@@ -574,27 +488,27 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
       clerkjs.openWaitlist(this.preOpenWaitlist);
     }
 
-    this.premountSignInNodes.forEach((props: SignInProps, node: HTMLDivElement) => {
+    this.premountSignInNodes.forEach((props, node) => {
```
brkalow, line 491: I've also removed unnecessary type annotations and instead am relying on the underlying `IsomorphicClerk` type definition.

tmilewski (approved):

wobsoriano (approved):

anagstef (approved):
👏

packages/react/src/isomorphicClerk.ts
panteliselef: yeah, some of the internals we need some we don't. I know that `__internal_addNavigationListener` is only needed within clerk-js, which makes me think we can refactor it 🤔.

packages/react/src/isomorphicClerk.ts
```diff
@@ -1113,10 +1026,10 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
     }
   };
 
-  handleGoogleOneTapCallback = (
+  handleGoogleOneTapCallback = async (
```
panteliselef, line 1029: 🚨 Technically this is a breaking change at the type level.

panteliselef:
!snapshot

packages/react/src/isomorphicClerk.ts
brkalow: Here: https://github.com/clerk/javascript/blob/main/packages/types/src/clerk.ts#L540-L544

PR #5198, opened 2025-02-19
chore(repo): Use `--profile` flag for attw

LekoArts:
## Description

Here and there we were fighting with https://github.com/arethetypeswrong/arethetypeswrong.github.io by disabling rules that aren't relevant for our packages at all. By default, it shows errors for Node 10 usage and in ESM-only packages it also shows CJS errors.

In v0.17.0 a `--profile` flag was added: https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/packages/cli/README.md#profiles

We can use `--profile node16` as default and `--profile esm-only` for our ESM-only packages. Using the profile won't remove the need for `--ignore-rules` completely but it cuts it down a bit and sets the correct baseline if we ever want to fix those (given that there's something to fix).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


nikosdouvlis (approved):
👏🏻

PR #5204, opened 2025-02-20
fix(clerk-js): Avoid client updates on sign out

panteliselef:
## Description

Prior to long lived clients, we were calling client.destroy which will not cause `#emit` to fire as part of `Clerk.updateClient`. Once long lived client got introduced, we started using `client.removeSessions()` which do update the client causing an early `#emit` to fire prior to the expected `setActive`.

### The problem
If an `<SignedIn/>` RSC was used to wrap to protect some client side code that was blindly expected `sessionId` from `useAuth` to always be defined, it would break.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


panteliselef:
This will probably be solved by #6905 or another similar mechanism.

PR #5280, opened 2025-03-04
feat(clerk-js,clerk-react,types): Navigate to tasks on after sign-in/sign-up

LauraBeatris:
## Description

Introduces navigation to tasks after sign-in/sign-up for UI components and custom flows:

https://github.com/user-attachments/assets/0fade809-cb54-412e-b3a2-b751b4f8a51f

#### Developer flow

It doesn't require changes on the `setActive` signature, since the navigation is triggered within `Clerk` once the session transitions to `pending` 

#### What's next: 
- Introduce a `Clerk.load` option (`tasksUrl`) for resolving URLs for tasks on custom flows 
- Handle redirection after resolving tasks 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/clerk.ts
```diff
@@ -926,6 +926,18 @@ export class Clerk implements ClerkInterface {
...
 
+    const hasSessionTaskToResolve = !!newSession?.tasks?.length;
+    if (hasSessionTaskToResolve && this.#options.standardBrowser) {
+      const hasEventHandler = eventBus.has(events.InternalComponentNavigate);
+      if (hasEventHandler && newSession) {
+        await new Promise<void>(resolveNavigation =>
+          eventBus.dispatch(events.InternalComponentNavigate, { resolveNavigation, session: newSession }),
+        );
+      } else {
+        await this.navigate(newSession?.currentTask?.__internal_getAbsoluteUrl(this.#options, this.environment));
```
izaaklauer, line 937: Idea: comment explaining why we land in each of these branches (custom flow vs our components)

packages/clerk-js/src/core/resources/SessionTask.ts
```diff
@@ -0,0 +1,61 @@
...
+  SessionTaskResource,
+} from '@clerk/types';
+
+import { buildURL, inBrowser } from '../../utils';
+
+export const SESSION_TASK_PATHS = ['add-organization'] as const;
+type SessionTaskPath = (typeof SESSION_TASK_PATHS)[number];
+
+export const SESSION_TASK_PATH_BY_KEY: Record<SessionTaskKey, SessionTaskPath> = {
+  org: 'add-organization',
```
izaaklauer, line 16: Idea: comment here explaining that when we add a new task type, we also need to add a corresponding resolution url, enforced by the compiler

integration/tests/session-tasks-sign-in.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    await u.po.signIn.setPassword(fakeUser.password);
+    await u.po.signIn.continue();
+    await u.po.expect.toBeSignedIn();
+
+    const addOrganizationTask = u.page.getByText(/create organization/i);
+    await expect(addOrganizationTask).toBeVisible();
+    expect(page.url()).toContain('add-organization');
+  });
+
+  test('redirect to task when accessing root sign in component', async ({ page, context }) => {
```
LauraBeatris, line 43: Currently, integration tests are failing, it seems that we changed some underlying feature flag logic on FAPI 

I'll also be adding another test suit for the after sign-up flow, covering the same cases regarding redirection

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1003,6 +1010,27 @@ export class Clerk implements ClerkInterface {
     await onAfterSetActive();
   };
 
+  #handlePendingSession = async (session: PendingSessionResource) => {
```
LauraBeatris, line 1013: @panteliselef @brkalow I think this captures what we have discussed regarding returning earlier on `setActive` to avoid bypassing a lot of the branches of logic by checking the session status 

This method is performing the following:
- Updating session token 
- Performing navigation for UI components vs custom flows
- Updating accessors and emitting listeners

packages/clerk-js/src/ui/components/SignIn/SignIn.tsx
```diff
@@ -132,12 +136,27 @@ function SignInRoutes(): JSX.Element {
...
+                <LazySignUpSessionTask />
+              </Route>
+            ))}
             <Route index>
               <LazySignUpStart />
             </Route>
           </Route>
         )}
-
+        {SESSION_TASK_PATHS.map(path => (
```
LauraBeatris, line 152: There's an curious behavior happening here, in which might be related with the routing structure, but I wanted to comment here anyway 

When we navigate to next factors, the DOM gets repainted on the exact nodes, but the whole page gets re-rendered for the task navigation (you can see the diffs below) 

Our Dashboard contains the following snippet: 

```
<div className='relative z-20 [--animation-duration:250ms] [&>div]:animate-fade-in'>
                <SignIn
                  routing='path'
                  path='/sign-in'
```

The above animation gets triggered again and it causes this small flickering on navigation: 


https://github.com/user-attachments/assets/04d6fedc-3899-40bd-93bf-167e7ae5d25e

packages/clerk-js/src/ui/components/SignIn/SignIn.tsx
```diff
@@ -132,12 +136,27 @@ function SignInRoutes(): JSX.Element {
                 <LazySignUpContinue />
               </Route>
             </Route>
+            {SESSION_TASK_PATHS.map(path => (
```
LauraBeatris, line 139: I need to be mindful of not including task components on the bundle by default, but only when the session transitions to have a task 

The complexity is on the fallback behavior, such as suspending on a shell once arriving on `/sign-in/add-organization`

Do we have any existing mechanisms similar to this case?

LauraBeatris:
Some unit tests started to fail from our previous permutations to handle `pending` exactly like `active`, but I think this is expected:

CleanShot 2025-03-05 at 23 07 17

I'll be updating this tomorrow morning and validating if there are valid cases broken

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1003,6 +1010,27 @@ export class Clerk implements ClerkInterface {
...
+    if (this.#internalComponentNavigate) {
+      // Handles navigation for UI components
+      await this.#internalComponentNavigate(session.currentTask.__internal_getPath());
+    } else {
+      // Handles navigation for custom flows
+      await this.navigate(session.currentTask.__internal_getUrl(this.#options, this.environment));
+    }
+
+    this.#setAccessors(session);
+    this.#emit();
```
LauraBeatris, line 1031: Our Dashboard contains the following snippet: 

```
<div className='relative z-20 [--animation-duration:250ms] [&>div]:animate-fade-in'>
                <SignIn
                  routing='path'
                  path='/sign-in'
```

The below animation gets triggered again and it causes this small flickering on navigation: 


https://github.com/user-attachments/assets/04d6fedc-3899-40bd-93bf-167e7ae5d25e


`this.#emit` updates `Clerk.session` which causes all React tree under `MultisessionAppSupport` to re-render, leading to a DOM re-painting and re-triggering the animation 

I think this is an issue unrelated to the current PR, and we didn't see it before since the previous sign-in routes didn't rely on `Clerk.session`

LauraBeatris:
The "Invite members" screen is broken, might have something to do with routing:

CleanShot 2025-03-06 at 13 45 26

packages/clerk-js/src/core/clerk.ts
LauraBeatris: I'm auditing this logic now against our unit tests, and it seems we might have to call touch for `pending` still 

On a multi-session application, if you're using an `active` session, but then by the time you switch to another one which is in `pending`, it might have expired and then we need to call `/touch` to handle sign-out

packages/clerk-js/src/core/resources/SessionTask.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    this.key = data.key;
+
+    return this;
+  }
+
+  /**
+   * Resolves path for internal component routing
+   */
+  public __internal_getPath() {
```
LauraBeatris, line 38: I'll move those to a separate module, either a singleton or simple `.ts`, cause it has nothing to do with the actual data resource but rather UI routing

packages/clerk-js/src/ui/common/withRedirect.tsx
```diff
@@ -89,11 +89,3 @@ export const withRedirectToAfterSignUp = <P extends AvailableComponentProps>(Com
 
   return HOC;
 };
-
-export const withRedirectToHomeSingleSessionGuard = <P extends AvailableComponentProps>(Component: ComponentType<P>) =>
```
LauraBeatris, line 93: This is unused, took the liberty to already remove it here

LauraBeatris:
Today I'll take a look at introducing a `experimental` prop to trigger lazy loading for after-auth routes in order to not affect the bundle size until the feature is released 

Furthermore, we're planning on introducing some FAPI changes on the environment resource to indicate whether after-auth tasks are enabled for that instance, so it could be leverage ahead of sign-in time to trigger lazy loading as well

packages/types/src/clerk.ts
```diff
@@ -772,6 +779,7 @@ export type ClerkOptions = ClerkOptionsNavigation &
       {
         persistClient: boolean;
         rethrowOfflineNetworkErrors: boolean;
+        withSessionTasks: boolean;
       },
```
LauraBeatris, line 783: This is only being introduced temporarily to not affect the bundle size once this PR gets merged 

We're working on FAPI changes to indicate that tasks are enabled on a instance by looking at the environment response, and so the SDK will only load those components based on that

LauraBeatris:
💡 Not a blocker for this PR, but definitely possible to address next week: Prefetch the organization memberships data before triggering the client-side navigation to `/add-organization` in order to avoid that extra flickering

packages/clerk-js/src/ui/contexts/components/SignIn.ts
```diff
@@ -112,6 +119,18 @@ export const useSignInContext = (): SignInContextType => {
...
+    ? buildSessionTaskRedirectUrl({ routing: ctx.routing, path: ctx.path }, signInUrl, clerk.session?.currentTask)
+    : null;
+
+  useEffect(() => {
+    clerk.__internal_setComponentNavigate((endpoint: string) =>
+      navigate(
+        buildRedirectUrl({ routing: ctx.routing, path: ctx.path, baseUrl: signInUrl, endpoint, authQueryString }),
+      ),
+    );
+  }, []);
```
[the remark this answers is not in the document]
LauraBeatris: > I'd expect to only pass navigate, was there a reason we need to construct the url in here ?

We need to have access to the `path` property passed on the component via prop: `<SignIn path="/sign-in" />`

> Should we clean this on unmount
From my last tests, the cleanup function here gets called on every in-component navigation, eg: from `_SignInStart` to `_SignInFactorOne`

Because of the above, it gets cleaned and it loses the reference by the time `setActive` - therefore I took the approach of cleaning it in-memory everytime the navigation finishes on `handlePendingSession`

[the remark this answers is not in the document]
panteliselef: I missed the part that we set it as null inside `handlePendingSession`. I think that covers it. What I am worried about is that our intentions are not clear here. I think a comment in both places that x-reference each other why we set and why we unset might be useful for the future if we need to refactor.

💡An alternative idea to explore: set and unset the function inside BaseRouter, if it works I think it is a cleaner approach.

integration/templates/next-app-router/src/app/layout.tsx
```diff
@@ -21,6 +21,7 @@ export default function RootLayout({ children }: { children: React.ReactNode })
         },
       }}
       experimental={{
+        withSessionTasks: true,
```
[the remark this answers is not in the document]
LauraBeatris: This is a temporary experimental prop that I'll plan on removing in the next PR once the FAPI changes on `/environment` get landed 

That's one reason that I thought of not adding the env var, cause otherwise will have to remove it later again

integration/tests/session-tasks-sign-up.test.ts
```diff
@@ -0,0 +1,40 @@
...
+test.describe('session tasks after sign-up flow @nextjs', () => {
+  test.describe.configure({ mode: 'serial' });
+  let app: Application;
+
+  test.beforeAll(async () => {
+    app = await appConfigs.next.appRouter.clone().commit();
+    await app.setup();
+    await app.withEnv(appConfigs.envs.withSessionTasks);
+    await app.dev();
+  });
```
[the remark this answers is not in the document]
LauraBeatris: > can we merge with the signin tests to avoid spinning up unnecessary applications ?

There's a FAPI feature flag targeting that new instance ID, and I didn't want to introduce it for existing sign-in/sign-up ones to avoid breaking the existing flow since it'll redirect to the task URL

packages/clerk-js/src/ui/contexts/components/SignIn.ts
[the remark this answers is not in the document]
LauraBeatris: > What I am worried about is that our intentions are not clear here 

I agree! I also didn't like having to place effects twice on different contexts 

I've refactored a bit, instead of storing only the navigation function, it includes a object with more like the internal context from `BaseRouter`  `https://github.com/clerk/javascript/pull/5280/commits/3696bd0a917d3ae5bff4a1dbb3df02bdf34aceeb` 

Another thing that we have to count for is the base path, that can also change based on the component context, eg: `<SignIn path='/my-sign-in' />`, so this also being passed

packages/clerk-js/src/ui/contexts/components/SignUp.ts
```diff
@@ -107,6 +114,18 @@ export const useSignUpContext = (): SignUpContextType => {
...
+  const sessionTaskUrl = clerk.session?.currentTask
+    ? buildSessionTaskRedirectUrl({ routing: ctx.routing, path: ctx.path }, signUpUrl, clerk.session?.currentTask)
+    : null;
+
+  useEffect(() => {
+    clerk.__internal_setComponentNavigate((endpoint: string) =>
+      navigate(
+        buildRedirectUrl({ routing: ctx.routing, path: ctx.path, baseUrl: signUpUrl, endpoint, authQueryString }),
+      ),
+    );
```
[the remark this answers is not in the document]
LauraBeatris: Replied [here](https://github.com/clerk/javascript/pull/5280#discussion_r1987802723)

packages/clerk-js/bundlewatch.config.json
```diff
@@ -2,8 +2,8 @@
   "files": [
     { "path": "./dist/clerk.js", "maxSize": "560kB" },
     { "path": "./dist/clerk.browser.js", "maxSize": "75kB" },
-    { "path": "./dist/clerk.headless.js", "maxSize": "48.3KB" },
-    { "path": "./dist/ui-common*.js", "maxSize": "89KB" },
+    { "path": "./dist/clerk.headless.js", "maxSize": "48.6KB" },
```
LauraBeatris, line 5: Bundle increased a bit because of all the `Clerk` core changes. Fortunally tho we could apply lazy loading and avoid increasing the bundle size for sign-in/sign-up

LauraBeatris:
Integration tests are failing for generic/sessions: 

CleanShot 2025-03-11 at 11 22 35

Seems unrelated to my PR tho, just wanted to flag in case of further reviews

packages/clerk-js/bundlewatch.config.json
```diff
@@ -2,8 +2,8 @@
   "files": [
     { "path": "./dist/clerk.js", "maxSize": "560kB" },
     { "path": "./dist/clerk.browser.js", "maxSize": "75kB" },
-    { "path": "./dist/clerk.headless.js", "maxSize": "48.3KB" },
-    { "path": "./dist/ui-common*.js", "maxSize": "89KB" },
+    { "path": "./dist/clerk.headless.js", "maxSize": "48.7KB" },
```
izaaklauer, line 5: ❓ How much margin does this leave in the bundle size?

LauraBeatris: I think we should increase the margin here to at least 50KB - the reason why, I've started this branch on top of another to continue implementing the feature, and added a couple of lines on `Clerk` core, which increased to the following:

CleanShot 2025-03-11 at 14 52 06

packages/clerk-js/src/ui/components/SignIn/SignIn.tsx
```diff
@@ -128,6 +129,11 @@ function SignInRoutes(): JSX.Element {
               >
                 <LazySignUpVerifyPhone />
               </Route>
+              {signInContext.withSessionTasks && (
+                <Route path='add-organization'>
```
izaaklauer, line 133: ❓  Do we want to pull this in from `SESSION_TASK_ROUTE_BY_KEY` ?

izaaklauer (approved):
Nice work! Well tested to my eye.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1003,6 +1013,46 @@ export class Clerk implements ClerkInterface {
...
+      await navigateToTask(session.currentTask, {
+        globalNavigate: this.navigate,
+        componentNavigationContext: this.#componentNavigationContext,
+        options: this.#options,
+        environment: this.environment,
+      });
+
+      // Reset component navigation context once navigation finishes
+      // to not conflict with after sign-in / after sign-up
+      this.#componentNavigationContext = null;
```
brkalow, line 1049: I'm surprised to see this value being reset here. What happens if we don't do it?

LauraBeatris: `BaseRouter` only gets unmounted once the application gets closed, or at least if you navigate to page where there's not a single Clerk component 

Since the setter/cleanup logic is happening on `BaseRouter`, there could be a case where we trigger the navigation to tasks once the session updates on the fly (another flow to be handled in a next PR), but `this.#componentNavigationContext` is still defined, therefore it would use it for navigation instead of `Clerk.navigate` that it's important to close opened modals

packages/clerk-js/src/ui/components/SignIn/SignIn.tsx
LauraBeatris: Makes sense! I'll have to refactor a bit of that constant lookups since it's value have a `/` concatenated on the string, and the `Route` path shouldn't include it

packages/clerk-js/src/ui/router/BaseRouter.tsx
```diff
@@ -121,6 +121,10 @@ export const BaseRouter = ({
     return internalNavRes;
   };
 
+  React.useEffect(() => {
+    return __internal_setComponentNavigationContext?.({ basePath, navigate: baseNavigate });
+  }, []);
```
brkalow, line 126: This might be better suited in our actual components, if `BaseRouter` isn't getting unmounted, per your other comment!

packages/clerk-js/src/core/sessionTasks.ts
```diff
@@ -0,0 +1,58 @@
...
+    const taskUrl = buildURL({
+      base: componentNavigationContext.basePath,
+      ...(isHashRouting
+        ? {
+            hashPath: SESSION_TASK_ROUTE_BY_KEY[task.key],
+          }
+        : {
+            pathname: componentNavigationContext.basePath + SESSION_TASK_ROUTE_BY_KEY[task.key],
+          }),
+    }) as URL;
```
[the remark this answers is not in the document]
LauraBeatris: I'm thinking of the case where `<SignIn routing='hash' />` gets used, but then we build a URL that uses path - would it fallback back to hash when it renders `HashRouter`?

packages/types/src/clerk.ts
brkalow: Let's add a comment above the option to explain what it's for, so we know to remove later.

packages/clerk-js/src/core/sessionTasks.ts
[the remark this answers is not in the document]
brkalow: If we're calling navigate we shouldn't need to be aware of the routing strategy, it accepts a path and the underlying router should handle it correctly. Does that make sense?

packages/clerk-js/src/core/clerk.ts
brkalow: Got it, thanks! I think we need to make sure that `componentNavigationContext` is only defined if a component is defined then, which appears to be the underlying case we're handling here.

packages/clerk-js/src/ui/router/BaseRouter.tsx
LauraBeatris: > This might be better suited in our actual components

My first implementation was only calling the following on `SignInContext/SignUpContext` - without a cleanup: 

```
React.useEffect(() => {
     __internal_setComponentNavigationContext?.({ basePath, navigate: baseNavigate });
  }, []);
```

The reason why, is that from `_SignInStart` to `_SignInFactorOne`, the cleanup gets triggered and we lose the reference before executing `setActive` 

Before, I was placing the effect on two different places (the component contexts) and then executing the cleanup after the navigation - which felt disconnected, and then I tried to move the setter at least to a single place

packages/clerk-js/src/core/sessionTasks.ts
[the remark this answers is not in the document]
LauraBeatris: Makes sense, I can perform more tests around hash vs path as well locally. 

We do have to forward the base path tho, besides navigate. In order to proceed with remaining the same base path regardless of additional ones, eg: 
- `/sign-in/factor-one` to `/sign-in/add-organization` instead of calling navigate('../add-organization') in which would lead to `/sign-in/factor-one/add-organization`

packages/clerk-js/src/ui/router/BaseRouter.tsx
brkalow: @LauraBeatris can we put it in the `<SignIn>` component itself then?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1003,6 +1013,46 @@ export class Clerk implements ClerkInterface {
...
   };
 
+  #handlePendingSession = async (session: SignedInSessionResource) => {
+    if (!this.environment) {
+      return;
+    }
+
+    if (session.lastActiveToken) {
+      eventBus.dispatch(events.TokenUpdate, { token: session.lastActiveToken });
+    }
```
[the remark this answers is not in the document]
LauraBeatris: Makes sense, I forgot that we call it on client piggybacking: https://github.com/clerk/javascript/blob/c65db9adfd71bdd668004f6f1d993d31c28aebff/packages/clerk-js/src/core/clerk.ts#L1765-L1766

packages/clerk-js/src/ui/common/redirects.ts
```diff
@@ -25,6 +28,22 @@ export function buildVerificationRedirectUrl({
...
+  const { routing, path } = ctx;
+
+  return buildRedirectUrl({
+    routing,
+    baseUrl,
+    path,
+    endpoint: SESSION_TASK_ROUTE_BY_KEY[task.key],
+    authQueryString: null,
+  });
+}
```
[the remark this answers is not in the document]
LauraBeatris: Makes sense to me! I tried to follow the existing function signature patterns of this module, but the above looks better indeed

packages/clerk-js/src/core/clerk.ts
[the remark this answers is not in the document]
LauraBeatris: Updated on https://github.com/clerk/javascript/pull/5280/commits/9ecac57bb1b585029ab8325d415ea03499c3e035

packages/clerk-js/src/core/clerk.ts
LauraBeatris: > if a component is defined then

For `SignIn/SignUp`, makes sense since those are the specific components handling tasks. 

In case we want to leverage this "navigation context" property for other features outside of after sign-in / sign-up in the future, then it makes sense to place it on `BaseRouter`

I don't see a case for the above now, so will try to place it the setter on SignIn/SignUp in a way that the cleanup only happens if the root component gets unmounted

packages/clerk-js/src/ui/router/BaseRouter.tsx
LauraBeatris: Putting inside each root component works better than my previous approach of relying on their React context! 

Here's the refactor (take a look at the git description for more details on the other changes as well) 🫡 https://github.com/clerk/javascript/pull/5280/commits/57203a1e1ceeb78869f8ceb3c61296bedcef8102

packages/clerk-js/src/ui/common/redirects.ts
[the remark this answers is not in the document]
LauraBeatris: Refactored on https://github.com/clerk/javascript/pull/5280/commits/57203a1e1ceeb78869f8ceb3c61296bedcef8102

packages/clerk-js/src/core/sessionTasks.ts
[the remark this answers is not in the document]
LauraBeatris: Refactored on https://github.com/clerk/javascript/commit/57203a1e1ceeb78869f8ceb3c61296bedcef8102 - and hash routing is working as well: 


https://github.com/user-attachments/assets/eed4ad50-f56f-44ab-a451-a365eb278cfe

packages/clerk-js/src/ui/components/SignUp/SignUp.tsx
```diff
@@ -21,8 +27,16 @@ function RedirectToSignUp() {
 }
 
 function SignUpRoutes(): JSX.Element {
+  const { __internal_setComponentNavigationContext } = useClerk();
+  const { navigate, basePath } = useRouter();
   const signUpContext = useSignUpContext();
 
+  usePreloadSessionTask(signUpContext.withSessionTasks);
+
+  React.useEffect(() => {
```
LauraBeatris, line 36: Seems that now `bundlewatch` started to fail due to those new changes, let me perform a quick bundle analysis here just to make sure and then commit the bump on the limit

LauraBeatris:
Fix coming up for the current integration test errors here: https://github.com/clerk/javascript/pull/5328

packages/clerk-js/src/ui/components/SignIn/SignIn.tsx
```diff
@@ -170,6 +187,13 @@ function SignInRoot() {
    */
   usePreloadSignUp(signInContext.isCombinedFlow);
 
+  // `experimental.withSessionTasks` will be removed soon in favor of checking via environment response
+  usePreloadSessionTask(signInContext.withSessionTasks);
+
+  React.useEffect(() => {
+    return __internal_setComponentNavigationContext?.({ basePath, navigate });
+  }, []);
```
brkalow, line 195: consider adding `basePath` and `navigate` to the dependency array
```suggestion
  React.useEffect(() => {
    return __internal_setComponentNavigationContext?.({ basePath, navigate });
  }, [basePath, navigate]);
```

packages/clerk-js/src/ui/components/SignUp/SignUp.tsx
```diff
@@ -21,8 +27,17 @@ function RedirectToSignUp() {
...
+  const { __internal_setComponentNavigationContext } = useClerk();
+  const { navigate, basePath } = useRouter();
   const signUpContext = useSignUpContext();
 
+  // `experimental.withSessionTasks` will be removed soon in favor of checking via environment response
+  usePreloadSessionTask(signUpContext.withSessionTasks);
+
+  React.useEffect(() => {
+    return __internal_setComponentNavigationContext?.({ basePath, navigate });
+  }, []);
```
brkalow, line 39:
```suggestion
    return __internal_setComponentNavigationContext?.({ basePath, navigate });
  }, [basePath, navigate]);
```

panteliselef (approved):
Looks good to 🚢 Some minor comments are 
- Update changeset to include `@clerk/clerk-react`
- Avoid logging for `<SignIn/>` being unable to render when user can access `'/sign-in/add-organization'`

LauraBeatris:
I'm currently doing more tests on SSO callbacks + combined flows that trigger `setActive` with a pending session. Once @panteliselef gets addressed, then I'll proceed with merging.

LauraBeatris:
Regarding the SSO callback flow -> It navigates to task when FAPI redirects to `/sso-callback` since it calls `setActive` with the pending session

However, it's not working if the sign-in is already created on FAPI, and it's redirecting to the `redirectUrlComplete` which doesn't trigger the navigation

There are two ways to approach (in a next PR):
- We trigger the navigation once the client gets updated -> find some way to react to this instead of relying on `setActive`
- FAPI navigates to `/sso-callback` or a new route such as `/sign-in/task-callback`

PR #5283, opened 2025-03-04
fix(shared): Offline detection with `isBrowserOnline`

panteliselef:
## Description

We should expect both `rtt` and `downlink` to be zero, in order to determine "offline" status

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/fuzzy-rules-perform.md
+++ b/.changeset/fuzzy-rules-perform.md
@@ -1,0 +1,5 @@
+---
+'@clerk/shared': patch
+---
+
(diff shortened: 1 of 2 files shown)


jacekradko:
These APIs are not very widely supported as of right now, so the usefulness of this method somewhat lessened. Have we considered more crude, but more reliable, approaches, like maybe making a HEAD request to a well known and stable URL as a fallback?

PR #5289, opened 2025-03-05
fix(clerk-js): Remove extra legal consent checkbox

alexcarpenter:
## Description

Removes legal consent checkbox from sign-up start card when only social or web3 strategies are enabled.

Fixes SDKI-904

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
+++ b/packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
@@ -8,9 +8,7 @@ import { SignInContext, useCoreSignUp, useEnvironment, useSignUpContext } from '
 import { descriptors, Flex, Flow, localizationKeys, useAppearance, useLocalizations } from '../../customizables';
 import {
   Card,
-  Form,
   Header,
-  LegalCheckbox,
   LoadingCard,
   SocialButtonsReversibleContainerWithDivider,
   withCardStateProvider,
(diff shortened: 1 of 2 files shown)


anagstef:
Does this PR only include the changeset or I'm missing something? 🤔

panteliselef:
@alexcarpenter I think we are missing the actual implementation.

packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
```diff
@@ -320,14 +318,6 @@ function _SignUpStart(): JSX.Element {
...
               )}
             </SocialButtonsReversibleContainerWithDivider>
-            {!shouldShowForm && isLegalConsentEnabled && (
-              <Form.ControlRow elementId='legalAccepted'>
-                <LegalCheckbox
-                  {...formState.legalAccepted.props}
-                  isRequired={fields.legalAccepted?.required}
-                />
-              </Form.ControlRow>
-            )}
```
panteliselef, line 330: Doesn't thins removes it completely ? If we have another checkbox can you point me to it ?

alexcarpenter: No, it gets passed through in fields.

Screenshot 2025-03-06 at 8 53 04 AM

octoper (approved):

packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
octoper: What I remember is that this was on the official iteration and if you checked the ToS and then clicked the `Continue with X` button you will sign up without asking legal consent

tuliren:
~@alexcarpenter, with this change, user can bypass agreeing to the terms when they sign up with social accounts. Is this intended? Isn't this a legal loophole?~

Oh, actually never mind. I realized that after signing-up with social accounts, there is a separate screen asking for user consent. So this PR makes sense. Thanks~~

PR #5299, opened 2025-03-07
feat(clerk-js): Track fapi requests triggered by UI components

panteliselef:
## Description

This PR decorates all requests towards the Frontend API with a new query param `_clerk_ui_triggered=true` which indicates that the request originated from our UI components.

The solution proposed by this PR, uses a proxy to wrap all resources (e.g. user and client) before passing them to their respective react context. The proxy will detect any function usage and will flag it as "used by UI", so that fapiClient can use this to attach the new query param. This solution allows to not make any changes to public API or affect the implementation of our resources.

**Attention:** This solution does not guarantee that the detection will work every time since it uses a stack to determine if the request should be flagged as "used by UI". If a request from a custom flow fires while there is at least one pending request triggered by a UI component, the former will be wrongly flagged as "used by UI".  This is acceptable because we are using the query param for analytics, and such errors will not affect the performance or the behaviour of the application.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/contexts/CoreClerkContextWrapper.tsx
+++ b/packages/clerk-js/src/ui/contexts/CoreClerkContextWrapper.tsx
@@ -8,6 +8,7 @@ import {
 import type { Clerk, LoadedClerk, Resources } from '@clerk/types';
 import React from 'react';
 
+import { makeUICaller } from '../../utils/detect-ui-caller';
 import { assertClerkSingletonExists } from './utils';
 
 type CoreClerkContextWrapperProps = {
@@ -35,10 +36,11 @@ export function CoreClerkContextWrapper(props: CoreClerkContextWrapperProps): JS
   }, []);
 
   const { client, session, user, organization } = state;
-  const clerkCtx = React.useMemo(() => ({ value: clerk }), []);
-  const clientCtx = React.useMemo(() => ({ value: client }), [client]);
-  const sessionCtx = React.useMemo(() => ({ value: session }), [session]);
-  const userCtx = React.useMemo(() => ({ value: user }), [user]);
+  const clerkCtx = React.useMemo(() => ({ value: makeUICaller(clerk) }), []);
+  const clientCtx = React.useMemo(() => ({ value: makeUICaller(client) }), [client]);
+  const sessionCtx = React.useMemo(() => ({ value: makeUICaller(session) }), [session]);
+  const userCtx = React.useMemo(() => ({ value: makeUICaller(user) }), [user]);
+
   const organizationCtx = React.useMemo(
     () => ({
       value: { organization: organization },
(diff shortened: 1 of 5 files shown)


packages/clerk-js/src/utils/detect-ui-caller.ts
```diff
@@ -0,0 +1,49 @@
...
+    },
+    decrease: () => {
+      state = Math.max(state - 1, 0);
+    },
+  };
+}
+
+export const usageByUIComponents = createStore(0);
+
+const isThenable = (value: unknown): value is Promise<unknown> => {
```
LauraBeatris, line 17:
```suggestion
const isPromise = (value: unknown): value is Promise<unknown> => {
```

🙃 I'd suggest updating the name to match the type predicate as well, `thenable` was a bit hard to understand at a first glance

packages/clerk-js/src/utils/detect-ui-caller.ts
```diff
@@ -0,0 +1,49 @@
...
+  return !!value && typeof (value as any).then === 'function';
+};
+
+export const makeUICaller = <T>(resource: T): T => {
```
LauraBeatris, line 21:
```suggestion
/** @internal **/ 
export const makeUICaller = <T>(resource: T): T => {
```

🙃 We could add JSDocs to indicate for external contributors that this is a tooling for internal usage

packages/clerk-js/src/ui/contexts/CoreClerkContextWrapper.tsx
```diff
@@ -35,10 +36,11 @@ export function CoreClerkContextWrapper(props: CoreClerkContextWrapperProps): JS
   }, []);
 
   const { client, session, user, organization } = state;
-  const clerkCtx = React.useMemo(() => ({ value: clerk }), []);
-  const clientCtx = React.useMemo(() => ({ value: client }), [client]);
-  const sessionCtx = React.useMemo(() => ({ value: session }), [session]);
-  const userCtx = React.useMemo(() => ({ value: user }), [user]);
+  const clerkCtx = React.useMemo(() => ({ value: makeUICaller(clerk) }), []);
```
LauraBeatris, line 39: 💡Not against this approach, mostly curious regarding other approaches - have you thought of forwarding a property to `Clerk` if it loads on a React context, and then accessing this on the FAPI client?

panteliselef: Can you share more about this. `Clerk` is already loaded when we mount the providers and the UI components.

LauraBeatris: Once the `CoreClerkContextWrapper` loads, it sets a property in `Clerk` that can be accessed on the `fapiClient`

```
if (clerk.hasUIListeners) {
   searchParams.append('_clerk_ui_triggered', 'true');
 }
```

panteliselef: I think this increases the margin of faulty detections by a lot, since all requests triggered outside of a UI component will be flagged incorrectly while a UI component is mounted.

panteliselef:
This PR has been open for a while, while the implementation works it adds a lot of complexity, we should do another attempt on this.

PR #5324, opened 2025-03-11
fix(vue): add check for clerk.value in UseSessionList composable

Dorilama:
## Description

Fixes UsesessionList bug: when `clientCtx` is defined `clerk` can be `null` and the composable throws when trying to access `clerk.value.setActive`.
The error can be reproduced in [this project on stackblitz](https://stackblitz.com/edit/vitejs-vite-jypanyhv?file=src%2FApp.vue)
Fixed by adding a condition to return the default values also when `clerk.value` is `null`.

## Checklist

- [ x] `pnpm test` runs as expected.
- [ x] `pnpm build` runs as expected.

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/seven-walls-drive.md
+++ b/.changeset/seven-walls-drive.md
@@ -1,0 +1,5 @@
+---
+'@clerk/vue': patch
+---
+
+fix UsesessionList bug. When `clientCtx` is defined `clerk` can be `null` and throw when trying to access `clerk.value.setActive`. Added a condition to return the default values also when `clerk.value` is `null`.
--- a/packages/vue/src/composables/useSessionList.ts
+++ b/packages/vue/src/composables/useSessionList.ts
@@ -35,14 +35,14 @@ export const useSessionList: UseSessionList = () => {
   const { clerk, clientCtx } = useClerkContext();
 
   const result = computed<UseSessionListReturn>(() => {
-    if (!clientCtx.value) {
+    if (!clientCtx.value || !clerk.value) {
       return { isLoaded: false, sessions: undefined, setActive: undefined };
     }
 
     return {
       isLoaded: true,
       sessions: clientCtx.value.sessions,
-      setActive: clerk.value!.setActive,
+      setActive: clerk.value.setActive,
     };
   });
 


.changeset/seven-walls-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/vue': patch
+---
+
+fix UsesessionList bug. When `clientCtx` is defined `clerk` can be `null` and throw when trying to access `clerk.value.setActive`. Added a condition to return the default values also when `clerk.value` is `null`.
```
panteliselef, line 5: @wobsoriano Clerk being `null` while `clientCtx` is defined seems like an invalid state. I'd expect the existing code to work as expected, which makes me that there is another underlying cause for this issue.

wobsoriano:
Hi @Dorilama! Thanks for the PR and reproduction.

As @panteliselef mentioned, having `clientCtx` defined while `clerk` is null is indeed an invalid state.

Our React `useSessionList()` hook [has the same logic](https://github.com/clerk/javascript/blob/8a77e9b4e43fb7f7dd7ed4def786826f8581291a/packages/shared/src/react/hooks/useSessionList.ts#L32) as this PR, but it utilizes an [extended version of the Clerk class](https://github.com/clerk/javascript/blob/main/packages/react/src/isomorphicClerk.ts), which allows methods to be lazy loaded even when Clerk is not yet available.

I'll investigate and provide feedback shortly! Thanks again

wobsoriano:
Hi @Dorilama! Thanks for the PR and reproduction.

As @panteliselef mentioned, having `clientCtx` defined while `clerk` is null is indeed an invalid state.

While our React `useSessionList()` hook [has a different logic](https://github.com/clerk/javascript/blob/8a77e9b4e43fb7f7dd7ed4def786826f8581291a/packages/shared/src/react/hooks/useSessionList.ts#L32), it utilizes an extended Clerk class, specifically the [`IsomorphicClerk`](https://github.com/clerk/javascript/blob/main/packages/react/src/isomorphicClerk.ts), which allows methods to be called lazily even when `clerk` is null.

In the Vue implementation, we do not use the `IsomorphicClerk`, so we should avoid rendering `clerk.value!.setActive` in `useSessionList.ts` if `clerk.value` is null. So this PR makes sense, thanks!

packages/vue/src/composables/useSessionList.ts
```diff
@@ -35,14 +35,14 @@ export const useSessionList: UseSessionList = () => {
   const { clerk, clientCtx } = useClerkContext();
 
   const result = computed<UseSessionListReturn>(() => {
-    if (!clientCtx.value) {
+    if (!clientCtx.value || !clerk.value) {
```
wobsoriano, line 38: Okay, the issue is that we have the `clientCtx` value i[nitially set to an empty object](https://github.com/clerk/javascript/blob/3c225d90227141dc62d955e76c7f8e0202524bc7/packages/vue/src/plugin.ts#L39) that's why it passes the condition even if `clerk` is null.

Would you be able to update the [initial value to undefined](https://github.com/clerk/javascript/blob/3c225d90227141dc62d955e76c7f8e0202524bc7/packages/vue/src/plugin.ts#L39) and assert the type for now? (`client: undefined as unknown as ClientResource`)

This should be a safe update as we check the value of `clerk` in all of our composables except `useSessionList()`.

Thanks!

.changeset/seven-walls-drive.md
wobsoriano: FYI the issue was `client` resource was initially set to an empty object. See my comment [here](https://github.com/clerk/javascript/pull/5324#discussion_r1989445357).

This one now looks good!

Dorilama:
thank you for all the work with the sdk :)

.changeset/fuzzy-pillows-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/vue': patch
+---
+
+updating the initial value of `clientCtx` to `undefined` as requested [here](https://github.com/clerk/javascript/pull/5324#discussion_r1989445357)
```
panteliselef, line 5:
```suggestion
Bug fix: Update the initial value of `clientCtx` to `undefined` to correctly infer that `clerk` is defined.
```

.changeset/seven-walls-drive.md
panteliselef, line 5: @Dorilama Can you please remove this changeset, in favour of the new one ? We should only have one changeset file per package

panteliselef (changes requested):

.changeset/seven-walls-drive.md
Dorilama: yes, sorry about that 😅

panteliselef (approved):

PR #5325, opened 2025-03-11
chore(nextjs): Improve observability on keyless failures

panteliselef:
## Description

This PR abstracts the keyless implementation under the `KeylessProvider` (RSC) and also allows for logs to be displayed when `<ClerkProvide debug/>` for data related to keyless. I remind you that when keyless fails for any reason, we display an error the a PK needs to be provided, since this is the action the developer needs to perform in order to proceed. 

In cases where keyless is expected to work, but fails to do so, it is usefully to have such logs to determine, what makes the feature to fail.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/nextjs/src/types.ts
```diff
@@ -22,4 +22,8 @@ export type NextClerkProviderProps = Without<ClerkProviderProps, 'publishableKey
    * @default false
    */
   dynamic?: boolean;
+  /**
+   *
+   */
+  debug?: boolean;
```
panteliselef, line 28: Maybe a generic `CLERK_DEBUG` env var ?

PR #5328, opened 2025-03-11
fix(repo):  Handle errors in E2E proxy server

jacekradko:
## Description

Long story short... crashing the proxy server is bad. 

Long story long... 

#### Background

Our integration tests run against app servers using Next.js in dev mode, meaning Hot Module Replacement (HMR) is enabled. In some tests, we route traffic through a proxy server, which unexpectedly crashed when handling the `/_next/webpack-hmr` route.

Why? Because our proxy server wasn’t handling errors properly.

#### What Changed?

- Implemented error handling in the proxy server to prevent crashes.
- This issue started suddenly, likely because we always install the latest version of each Next.js major release in our test matrix.
- Suspicious Timing? Next.js v15.2.2 was released today, and it seems like the HMR route behavior changed. Coincidence? I think not.

#### Open Questions

- Should we be running Next.js in dev mode during tests?
- Should we pin Next.js versions instead of always using the latest major release?

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/integration/scripts/proxyServer.ts
+++ b/integration/scripts/proxyServer.ts
@@ -1,3 +1,4 @@
+import type { IncomingMessage, ServerResponse } from 'node:http';
 import http from 'node:http';
 import type { createServer as _createServer, Server, ServerOptions } from 'node:https';
 import https from 'node:https';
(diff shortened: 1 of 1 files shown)


LauraBeatris (approved):

LekoArts:
> Should we be running Next.js in dev mode during tests?

If the test assertions require it, then yes. If it can be either I personally (in my own projects) prefer to use the build + serve option as it's faster, more stable/reliable, and ensures that it works in "production". Sometimes there are differences between dev and build behavior and I rather test against what is shown live.
But sometimes I have to test functionality that is only relevant during dev, so I use dev in those instances.

> Should we pin Next.js versions instead of always using the latest major release?

Keeping us auto-updated is a good thing as we can react to (un)intended changes that might break something on our end. As a user I expect it to always work with the latest version of Next.js so we should test this. And us having to manually updated pinned versions (even with bots) won't work, we'll forget it.

octoper (approved):

anagstef (approved):

panteliselef (approved):

PR #5345, opened 2025-03-13
feat(clerk-js): Allow token refresh when Client failed to resolve

panteliselef:
## Description

Until now we depended on the client to be successfully fetched in order to mark clerk as initialized. This is changing in this PR.

When `/client` fails for any reason, a 4xx error, a 5xx error, a network error (aka blocked), we are creating a "local" only representation of the client. If the user is already signed in, clerk-js leverages the existing JWT of the `__session` cookie to create the client based on the existing information (userId, sessionId, orgId etc.).

With this PR, (only for standard browsers) if `/client` fails, clerk-js always initiates a local-only client to allow for operations like `Clerk.session?.getToken()` to still operate as expected. That means `Clerk.loaded` will be `true` even when we don't have a proper client object in memory.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/types/src/json.ts
```diff
@@ -68,7 +71,6 @@ export interface EnvironmentJSON extends ClerkResourceJSON {
 export interface ClientJSON extends ClerkResourceJSON {
   object: 'client';
   id: string;
-  status: any;
```
panteliselef, line 71: this is safe to remove

packages/types/src/jwt.ts
```diff
@@ -86,6 +86,18 @@ export interface ClerkJWTClaims {
...
+   */
+  org_permissions?: OrganizationCustomPermissionKey[];
+
+  /**
+   * Factor verification age (fva). The tuple represents the minutes that have passed since the last time a first or second factor were verified.
+   * This API is experimental and may change at any moment.
+   * @experimental
+   */
+  fva?: [fistFactorAge: number, secondFactorAge: number];
+
```
panteliselef, line 100: Something that was missing

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2052,10 +2053,26 @@ export class Clerk implements ClerkInterface {
...
+              const localClient = createClientFromJwt(jwtInCookie);
+
+              this.updateClient(localClient);
+
+              // Always grab a fresh token
+              await this.session?.getToken({ skipCache: true });
+
+              // Allows for Clerk to be marked as loaded with the client and session created from the JWT.
+              return null;
+            });
```
panteliselef, line 2075: I wouldn't focus too much on where this code is located, we might refactor once both this and #5287 are merged.

packages/clerk-js/src/core/auth/cookies/session.ts
```diff
@@ -35,8 +36,11 @@ export const createSessionCookie = (cookieSuffix: string): SessionCookieHandler
     sessionCookie.set(token, { expires, sameSite, secure });
   };
 
+  const get = () => suffixedSessionCookie.get() || suffixedSessionCookie.get();
```
LauraBeatris, line 39: Removing the duplication + referencing the function:

```suggestion
  const get = suffixedSessionCookie.get;
```

packages/clerk-js/src/core/jwt-client.ts
```diff
@@ -0,0 +1,96 @@
...
+import { Token } from './resources';
+import { Client } from './resources/Client';
+
+/**
+ * Create a new client instance from a jwt.
+ * The caller is responsible for reading the jwt from the `__session` cookie.
+ */
+export function createClientFromJwt(jwt: string | undefined | null): Client {
+  // Use `Token` class to parse the JWT token
+  let token;
```
LauraBeatris, line 20:
```suggestion
  let token: Token | null;
```

Can we type it earlier rather than leaving the comment about it?

packages/clerk-js/src/core/jwt-client.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  const defaultClient = {
+    object: 'client',
+    last_active_session_id: sid,
+    id: 'client_init',
+    sessions: [
+      {
+        object: 'session',
+        id: sid,
+        status: 'active',
```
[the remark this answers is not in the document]
panteliselef: Doing this in another PR sounds proper. TBH i don't think it is that important, because it might force redirect you to the appropriate page, but clerk will be down so you will not be able to proceed.

packages/clerk-js/src/core/auth/cookies/session.ts
panteliselef: nice catch probably i wanted to grab the suffixed and unsuffixed versions.

jacekradko (approved):
:shipit:

PR #5352, opened 2025-03-14
fix(clerk-js): Handle two factor redirect when authenticate with web3

nikospapcom:
## Description

In this pr we're handling the redirect to `two-factor` when user try to login with web3 and has enabled the `Multi-factor authentication`

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### BEFORE

https://github.com/user-attachments/assets/07f2ab5b-ca62-4da5-a0b1-9a7324aea80d

### AFTER

https://github.com/user-attachments/assets/5b05234c-6499-4a0f-bc09-849d175a8202

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -1901,11 +1901,20 @@ export class Clerk implements ClerkInterface {
       }
     }
 
-    if (signInOrSignUp.createdSessionId) {
-      await this.setActive({
-        session: signInOrSignUp.createdSessionId,
-        redirectUrl,
-      });
+    switch (signInOrSignUp.status) {
+      case 'needs_second_factor':
+        await navigate('factor-two');
+        break;
+      case 'complete':
+        if (signInOrSignUp.createdSessionId) {
+          await this.setActive({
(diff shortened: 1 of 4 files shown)


anagstef (commented):
Is it possible to add an integration test for this? 🤔

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1808,11 +1808,20 @@ export class Clerk implements ClerkInterface {
...
     }
 
-    if (signInOrSignUp.createdSessionId) {
-      await this.setActive({
-        session: signInOrSignUp.createdSessionId,
-        redirectUrl,
-      });
+    switch (signInOrSignUp.status) {
+      case 'needs_second_factor':
+        await navigate('factor-two');
```
panteliselef, line 1813: This assumes usage is always with in our components which is not the case.

Sounds like we need to borrow a few things from handleRedirectCallback.

panteliselef (changes requested):

panteliselef (approved):

PR #5372, opened 2025-03-17
chore(shared,clerk-react,types): Improve Typedoc

LekoArts:
## Description

TL;DR: Improve the JSDoc comments of all React hooks and add additional Next.js examples to some of them. Modify the MDX output so that it can be used for rendering on clerk.com/docs

- Format Typedoc MDX output with Prettier
- Remove headings from certain Typedoc files
- Add support for replacing links in Typedoc output
- Remove type signature titles and type parameter tables from Typedoc output
- Use `{@include}` JSDoc comments to add additional examples

Fixes ECO-508
Fixes ECO-424

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.typedoc/custom-plugin.mjs
```diff
@@ -1,5 +1,43 @@
 // @ts-check
-import { MarkdownRendererEvent } from 'typedoc-plugin-markdown';
+import { MarkdownPageEvent, MarkdownRendererEvent } from 'typedoc-plugin-markdown';
+
+/**
+ * A list of files where we want to remove any headings
+ */
+const FILES_WITHOUT_HEADINGS = [
```
LekoArts, line 7: Some of the files only contain the contents of a single interface.

For example:

```md
# Parameters

Table goes here
```

So since I want to use the table contents as a partial/include somewhere else I need to get rid off the heading. But since I don't want to remove headings everywhere I'm removing them only from the files I want to use in such way.

.typedoc/custom-plugin.mjs
```diff
@@ -1,5 +1,43 @@
...
+  'pages-or-infinite-options.mdx',
+  'paginated-hook-config.mdx',
+  'use-organization-list-return.mdx',
+  'use-organization-list-params.mdx',
+];
+
+/**
+ * An array of tuples where the first element is the file name and the second element is the new path.
+ */
+const LINK_REPLACEMENTS = [
```
LekoArts, line 21: Sometimes there are relative links like `[PaginatedResponse](../types/paginated-response.mdx)` in the files. The link to the also generated file through typedoc. But we already have that page manually created and want to link to it. So this replaces relative links with links to our existing docs.

packages/react/examples/use-auth.md
```diff
@@ -0,0 +1,43 @@
+<!-- #region nextjs-01 -->
```
LekoArts, line 1: In order to use the https://typedoc.org/documents/Tags.__include_.html flag the examples should be in the root of the package, so that relative links work for the files inside `dist`. Because it has to be resolved from `dist` to that folder.

packages/react/src/hooks/useAuth.ts
```diff
@@ -21,6 +21,9 @@ import { createGetToken, createSignOut } from './utils';
  *
  * The following example demonstrates how to use the `useAuth()` hook to access the current auth state, like whether the user is signed in or not. It also includes a basic example for using the `getToken()` method to retrieve a session token for fetching data from an external resource.
  *
+ * <Tabs items='React,Next.js'>
```
LekoArts, line 24: If you use these MDX components (they are not resolved, just passed along) in such way, they won't be rendered in IntelliSense

packages/shared/src/react/hooks/useOrganization.tsx
```diff
@@ -264,7 +269,7 @@ const undefinedPaginatedResource = {
  * }
  * ```
  */
-export const useOrganization: UseOrganization = params => {
+export function useOrganization<T extends UseOrganizationParams>(params?: T): UseOrganizationReturn<T> {
```
LekoArts, line 272: No behavior change, just removing a type alias and explicitly making it a `function` (sometimes typedoc can be tripped up by using a `const` for a function)

packages/shared/src/react/hooks/useOrganization.tsx
```diff
@@ -22,19 +22,24 @@ import {
 import type { PaginatedHookConfig, PaginatedResources, PaginatedResourcesWithDefault } from '../types';
 import { usePagesOrInfinite, useWithSafeValues } from './usePagesOrInfinite';
 
-type UseOrganizationParams = {
+/**
+ * @interface
+ */
+export type UseOrganizationParams = {
```
LekoArts, line 28: Inside `useOrganization` and `useOrganizationList` you'll notice that I added `export` to the params and return types. I need those to be accessible to typedoc, which requires you to export them.

Since we don't re-export them from `src/react/hooks/index.ts` file they won't surface to the users.

packages/shared/typedoc.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://typedoc.org/schema.json",
-  "entryPoints": ["./src/index.ts", "./src/react/index.ts", "./src/react/types.ts"],
+  "entryPoints": ["./src/index.ts", "./src/react/types.ts", "./src/react/hooks/*.{ts,tsx}"],
```
LekoArts, line 3: I needed access to the exported interfaces inside `useOrganization`/`useOrganizationList` but didn't want to change the public API. So we're looking at the hooks individually now.

packages/react/.gitignore
```diff
@@ -1,2 +1,3 @@
 /*/
 !/src/
+!/examples/
```
panteliselef, line 3: Can we rename to `doc-examples`, or simply `docs`  ?

packages/shared/src/react/hooks/useSession.ts
```diff
@@ -12,6 +12,9 @@ type UseSession = () => UseSessionReturn;
  *
  * The following example uses the `useSession()` hook to access the `Session` object, which has the `lastActiveAt` property. The `lastActiveAt` property is a `Date` object used to show the time the session was last active.
  *
+ * <Tabs items='React,Next.js'>
```
panteliselef, line 15: Could this be `items='react,next'` ?

panteliselef (approved):

packages/react/.gitignore
LekoArts: Yeah, we can choose whatever name we want really. I'd probably pick `docs` then

PR #5375, opened 2025-03-17
feat(clerk-expo): WIP cache exports

wobsoriano:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/expo/package.json
```diff
@@ -100,9 +108,6 @@
     },
     "expo-local-authentication": {
       "optional": true
-    },
-    "expo-secure-store": {
-      "optional": true
```
anagstef, line 105: We should remove this in the next major, because if someone is not using this package they will get errors when updating their package.

packages/expo/src/secure-store/index.ts
```diff
@@ -1 +0,0 @@
-export { createSecureStore as secureStore } from './secure-store';
```
panteliselef, line 1: We should not remove this. We are already exporting from `/secure-store`. Let's deprecate instead.

Doing the deprecation from `packages/expo/src/resource-cache/index.ts` is not enough, since the application will still break since nothing is exported from `/secure-store` anymore.

packages/expo/src/token-cache/index.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Create a token cache using Expo's SecureStore
+ */
+const createTokenCache = (): TokenCache => {
+  return {
+    getToken: async (key: string) => {
+      try {
+        const item = await SecureStore.getItemAsync(key);
+        return item;
+      } catch (error) {
+        console.error('Clerk: Failed to retrieve token from SecureStore:', error);
```
panteliselef, line 16: I'd remove that.

panteliselef (changes requested):
Seems like we are not deprecating properly. And we are introducing breaking changes since `secureStore` is no longer exported from the expected path.

packages/expo/package.json
wobsoriano: Good catch and I agree, thanks!

packages/expo/src/secure-store/index.ts
wobsoriano: Oh, awesome catch! I added back the `/secure-store` folder and re-exported `secureStore` there with alias and deprecation. Thanks!

packages/expo/src/token-cache/index.ts
wobsoriano: Removed 👍🏼

panteliselef (approved):
:shipit:

PR #5377, opened 2025-03-17
feat(clerk-js,clerk-react,types): Navigate to next task

LauraBeatris:
## Description

Resolves ORGS-581

Handles navigation to next tasks on an after-auth flow. Once all tasks get completed, it navigates to the `redirectUrlComplete` option, which defaults to after sign-in/sign-up URL based on the `Clerk.client` state. 

### Developer flows 

For AIO components, there's no signature changes and it'll work out of the box 

For custom flows, `Clerk.nextTask` should be called after performing a FAPI mutation that resolves a task. Here's an example for the upcoming force MFA task:

```ts 
await fapiClient.configureTotp(...) 
await Clerk.__experimental_nextTask()
```

Building your on custom React "onboarding" components: 
```tsx
function MyFirstOnboardingPage(){ 
  const clerk = useClerk()

  const handleOrganizationSelection = async (organization: string) => {
   // Select the organization into the current session 
    await setActive({ organization })

   // Proceed with verifying if there's a next task, either navigating to it 
   // or to `redirectUrlComplete` once session transitions to active
    await clerk.__experimental_nextTask()
  }

  return <MyCustomOrganizationList onOrgSelection={handleOrgSelection} />
}

function MySecondOnboardingPage(){ 
  const clerk = useClerk()
  return <ConfigureMFA onSuccess={clerk.__experimental_nextTask} />
}
```
---- 

https://github.com/user-attachments/assets/fe55ffd7-3e33-4bad-9086-d0bfebc5f995

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/bundlewatch.config.json
+++ b/packages/clerk-js/bundlewatch.config.json
@@ -2,7 +2,7 @@
   "files": [
     { "path": "./dist/clerk.js", "maxSize": "577.51kB" },
     { "path": "./dist/clerk.browser.js", "maxSize": "78.5kB" },
-    { "path": "./dist/clerk.headless.js", "maxSize": "51KB" },
+    { "path": "./dist/clerk.headless.js", "maxSize": "55KB" },
     { "path": "./dist/ui-common*.js", "maxSize": "94KB" },
     { "path": "./dist/vendors*.js", "maxSize": "30KB" },
     { "path": "./dist/coinbase*.js", "maxSize": "35.5KB" },
--- a/packages/clerk-js/bundlewatch.config.json
+++ b/packages/clerk-js/bundlewatch.config.json
@@ -1,6 +1,6 @@
 {
   "files": [
-    { "path": "./dist/clerk.js", "maxSize": "577.51kB" },
+    { "path": "./dist/clerk.js", "maxSize": "580kB" },
     { "path": "./dist/clerk.browser.js", "maxSize": "78.5kB" },
     { "path": "./dist/clerk.headless.js", "maxSize": "55KB" },
     { "path": "./dist/ui-common*.js", "maxSize": "94KB" },
--- a/packages/clerk-js/src/core/__tests__/clerk.test.ts
+++ b/packages/clerk-js/src/core/__tests__/clerk.test.ts
@@ -1,5 +1,6 @@
 import type {
   ActiveSessionResource,
+  PendingSessionResource,
   SignedInSessionResource,
   SignInJSON,
   SignUpJSON,
@@ -486,6 +487,15 @@ describe('Clerk singleton', () => {
         lastActiveToken: { getRawString: () => 'mocked-token' },
         tasks: [{ key: 'org' }],
         currentTask: { key: 'org', __internal_getUrl: () => 'https://foocorp.com/add-organization' },
+        reload: jest.fn(() =>
+          Promise.resolve({
+            id: '1',
+            status: 'pending',
+            user: {},
+            tasks: [{ key: 'org' }],
+            currentTask: { key: 'org', __internal_getUrl: () => 'https://foocorp.com/add-organization' },
+          }),
+        ),
       };
       let eventBusSpy;
 
@@ -509,7 +519,7 @@ describe('Clerk singleton', () => {
 
         const sut = new Clerk(productionPublishableKey);
         await sut.load();
-        await sut.setActive({ session: mockSession as any as ActiveSessionResource });
+        await sut.setActive({ session: mockSession as ActiveSessionResource });
         expect(mockSession.touch).toHaveBeenCalled();
       });
 
@@ -520,7 +530,7 @@ describe('Clerk singleton', () => {
 
         const sut = new Clerk(productionPublishableKey);
         await sut.load({ touchSession: false });
-        await sut.setActive({ session: mockSession as any as ActiveSessionResource });
+        await sut.setActive({ session: mockSession as ActiveSessionResource });
         await waitFor(() => {
           expect(mockSession.touch).not.toHaveBeenCalled();
           expect(mockSession.getToken).toHaveBeenCalled();
@@ -534,7 +544,7 @@ describe('Clerk singleton', () => {
 
         const executionOrder: string[] = [];
         mockSession.touch.mockImplementationOnce(() => {
-          sut.session = mockSession as any;
+          sut.session = mockSession;
           executionOrder.push('session.touch');
           return Promise.resolve();
         });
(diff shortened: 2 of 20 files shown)


packages/types/src/clerk.ts
```diff
@@ -1432,6 +1442,9 @@ export type OrganizationListProps = {
    * @default false
    */
   hideSlug?: boolean;
+
+  __internal_onOrganizationSelected?: () => void;
+  __internal_onOrganizationCreated?: () => void;
```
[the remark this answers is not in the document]
LauraBeatris: `afterSelectOrganizationUrl` and `afterCreateOrganizationUrl` are passed to the `navigate` function internally within the component 

https://github.com/clerk/javascript/blob/37278b55c381f26028046f59abf31deda632909e/packages/clerk-js/src/ui/contexts/components/OrganizationList.ts#L51-L53

We want to pass `nextTask` reference in order to handle the underlying session, if it's still pending with tasks, or if it has resolved all of them.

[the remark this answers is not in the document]
LauraBeatris: I also thought of the case, if folks want to "eject" out of the sign-in/sign-up routes in the future for their onboarding flow, once we provide that option 

And if rendering the `OrganizationList` component still, if that property should be provided. Since we're keeping this as internal within `SessionTask`, I'd prefer to set as a internal prop for now.

packages/react/src/isomorphicClerk.ts
```diff
@@ -612,6 +613,22 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
...
+  };
+
+  __internal_setComponentNavigationContext = (params: __internal_ComponentNavigationContext) => {
+    if (this.clerkjs) {
+      return this.clerkjs.__internal_setComponentNavigationContext(params);
+    } else {
+      return undefined;
+    }
+  };
+
```
panteliselef, line 631: I don't think exposing this makes a lot of sense, only our components use it, correct ?

packages/types/src/clerk.ts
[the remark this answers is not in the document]
panteliselef: Maybe we need a `<SessionTaskProvider/>` ?

packages/react/src/isomorphicClerk.ts
LauraBeatris: That's correct, I thought the purpose of this class was to avoid issues in hydration contexts, could you refresh my mind on that? 🤔

packages/types/src/clerk.ts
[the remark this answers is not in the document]
LauraBeatris: I like that in order to not change the type interface for organization list if it's only for internal route usage 

We'll have to do something like it: `useContext(SessionTaskContext)` also in order to not break for standalone usage of organization list 

Going to push some changes soon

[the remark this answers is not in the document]
panteliselef: Regarding the standalone OrgList, why if pending task exists but afterCreateOrgUrl is provided ? I'd feel that org list should still call nextTask()

packages/react/src/isomorphicClerk.ts
panteliselef: I'd argue that here we need to offer internal utilities that need to be publicly available aka `__internal_openReverification` which is used by `@clerk/shared`. But if something doesn't have to be public, we can simply use it internally.

LauraBeatris:
I also remembered that virtual routing is currently broken for `SessionTask` - I'll push a fix for it here!

packages/react/src/isomorphicClerk.ts
LauraBeatris: Makes sense - I've removed on https://github.com/clerk/javascript/pull/5377/commits/5459bb0d41fd9c7a152121e9eed7e867a72e4df3

packages/types/src/clerk.ts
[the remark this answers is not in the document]
LauraBeatris: It's an impossible case for now. Let's say, if the `OrganizationList` component gets rendered standalone in an authenticated route, and the next `/touch` returns a `pending` session - this could be due to the app owner removing the user org membership. 

In the case above, the SDK should immediately detect and either navigate back to the session task sign-in route or display a modal similar to re-verification. This is going to be tackled in the next PR.

packages/clerk-js/bundlewatch.config.json
```diff
@@ -2,7 +2,7 @@
   "files": [
     { "path": "./dist/clerk.js", "maxSize": "577kB" },
     { "path": "./dist/clerk.browser.js", "maxSize": "78kB" },
-    { "path": "./dist/clerk.headless.js", "maxSize": "51KB" },
+    { "path": "./dist/clerk.headless.js", "maxSize": "55KB" },
```
LauraBeatris, line 5: I'm bumping this to +4 to not break too soon on the next PRs - let me know if you have any concerns!

LauraBeatris:
https://github.com/clerk/javascript/blob/13e3ddc3acacfee3e52278e6a14a556a5203cde2/packages/clerk-js/src/core/clerk.ts#L1019-L1023

Regarding the above, I couldn't reproduce this behavior (flickers / automatic reloading) on navigating within `nextTask` to `afterSignInUrl`

`setActive({ organization })` runs, which emits the latest `Clerk.session` to the React context, which re-renders `SignIn -> SessionTask`, but it doesn't cause a flicker

I haven't tried the following tho: 
`<SignedIn><SignIn /></SignedIn>`

LauraBeatris:
Related to the above - I'm trying to get up to speed atm on why the `#emit` from `updateClient` doesn't call the same issue as above (with SignedOut unmounts and causing flash of content), when `signIn.create` returns a client session. 

If reviewers here could provide more context in the meanwhile, that would be helpful!

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1054,15 +1054,35 @@ export class Clerk implements ClerkInterface {
...
+  #handlePendingSession = async (
+    session: PendingSessionResource,
+    organization?: string | OrganizationResource | null | undefined,
+  ) => {
     if (!this.environment) {
       return;
     }
 
+    const shouldSwitchOrganization = organization !== undefined;
+    if (shouldSwitchOrganization) {
```
LauraBeatris, line 1066: Potentially thinking of extracting this bit of organization resolution logic to a separate private method in order to not duplicate on `setActive`

brkalow (commented):
This looks good overall! I like the approach, the API feels pretty intuitive I'd say.

Had a question about consolidating the org logic out of `handlePendingSession()`, let me know what you think 👀

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1088,15 +1101,43 @@ export class Clerk implements ClerkInterface {
...
+        environment: this.environment,
+        globalNavigate: this.navigate,
+        componentNavigationContext: this.#componentNavigationContext,
+      });
+      return;
     }
 
+    this.#setTransitiveState();
+    const defaultRedirectUrlComplete = this.client?.signUp ? this.buildAfterSignUpUrl() : this.buildAfterSignUpUrl();
+    await this.navigate(redirectUrlComplete ?? defaultRedirectUrlComplete);
```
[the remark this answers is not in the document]
LauraBeatris: Updated at https://github.com/clerk/javascript/pull/5377/commits/0429c06c28cae93eac1d2ef504cc2d5b0683b0fa

packages/clerk-js/src/core/__tests__/clerk.test.ts
```diff
@@ -486,6 +487,15 @@ describe('Clerk singleton', () => {
         lastActiveToken: { getRawString: () => 'mocked-token' },
         tasks: [{ key: 'org' }],
         currentTask: { key: 'org', __internal_getUrl: () => 'https://foocorp.com/add-organization' },
+        reload: jest.fn(() =>
+          Promise.resolve({
+            id: '1',
+            status: 'pending',
+            user: {},
+            tasks: [{ key: 'org' }],
+            currentTask: { key: 'org', __internal_getUrl: () => 'https://foocorp.com/add-organization' },
```
izaaklauer, line 496: For posterity: I'm always just a little nervous of including random routable URLs in tests. 

Truly unlikely security threat potential: If we actually GET this URL, and someone knows we're getting it, they could register the domain and have it serve some exploit that gets us to leak our CI secrets.

I'm not even proposing you change this - it's the pattern laid down elsewhere in the test - but when in doubt, I like `example.com` or cnames of it. It's not routable. `https://foocorp.example.com/add-organization` is safer in my book.

LauraBeatris: Makes sense! Also just to note that our tests only run if the triggering actor (owner of the PR) has the GitHub org permissions to access secrets, otherwise it'll fail 

For URLs like those that are used to trigger navigation rather than HTTP requests, then I think it's fine!

packages/clerk-js/src/ui/components/SignIn/SignIn.tsx
```diff
@@ -132,7 +132,10 @@ function SignInRoutes(): JSX.Element {
               </Route>
               {signInContext.withSessionTasks && (
                 <Route path={SESSION_TASK_ROUTE_BY_KEY['org']}>
-                  <SessionTask task='org' />
+                  <SessionTask
+                    task='org'
+                    redirectUrlComplete={signInContext.afterSignUpUrl}
```
izaaklauer, line 137: ❓  Not `afterSignInUrl`

LauraBeatris: This is within the `isCombinedFlow` routes, so those are the sign-up routes rendered within `SignIn` component therefore it uses `signUpContext`

PR #5391, opened 2025-03-18
chore(clerk-js,types): Hide personal workspace options when organization selection is enforced

LauraBeatris:
## Description

Resolves ORGS-537

Introduces `force_organization_selection` on environment settings. When enabled, all organization components should hide the "personal workspace" options. 

This is also important for the after-auth flow with `force_organization_selection`, to avoid displaying the option to select a personal account on `OrganizationList`, and to also not display "Personal Workspace" on `OrganizationSwitcher` 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/contexts/components/OrganizationList.ts
+++ b/packages/clerk-js/src/ui/contexts/components/OrganizationList.ts
@@ -11,7 +11,7 @@ export const OrganizationListContext = createContext<OrganizationListCtx | null>
 export const useOrganizationListContext = () => {
   const context = useContext(OrganizationListContext);
   const { navigate } = useRouter();
-  const { displayConfig } = useEnvironment();
+  const { displayConfig, organizationSettings } = useEnvironment();
 
   if (!context || context.componentName !== 'OrganizationList') {
     throw new Error('Clerk: useOrganizationListContext called outside OrganizationList.');
@@ -80,7 +80,7 @@ export const useOrganizationListContext = () => {
     afterCreateOrganizationUrl,
     skipInvitationScreen: ctx.skipInvitationScreen || false,
     hideSlug: ctx.hideSlug || false,
-    hidePersonal: ctx.hidePersonal || false,
+    hidePersonal: organizationSettings.forceOrganizationSelection || ctx.hidePersonal || false,
     navigateAfterCreateOrganization,
     navigateAfterSelectOrganization,
     navigateAfterSelectPersonal,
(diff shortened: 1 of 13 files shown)


packages/clerk-js/src/ui/contexts/components/OrganizationList.ts
```diff
@@ -80,7 +80,7 @@ export const useOrganizationListContext = () => {
     afterCreateOrganizationUrl,
     skipInvitationScreen: ctx.skipInvitationScreen || false,
     hideSlug: ctx.hideSlug || false,
-    hidePersonal: ctx.hidePersonal || false,
+    hidePersonal: organizationSettings.forceOrganizationSelection || ctx.hidePersonal || false,
```
brkalow, line 83: Should we deprecate this prop once the feature releases?

brkalow (approved):

packages/clerk-js/src/ui/contexts/components/OrganizationList.ts
brkalow: Makes sense! Let's add it to the project as a discussion topic so we don't forget. I agree that there might be cases where someone could take advantage of it, but if we have the opportunity to remove options in favor of our recommended approach that would be good 👍

LauraBeatris: Let's say the developer currently uses `hidePersonal=false` on the application, and then enables force an org in the Dashboard. 

The expected behavior is to hide personal workspace across all org components. Especially for the after-auth flow, because if it gets displayed, then it'd lead to a bug or the user could select a personal workspace and get stuck with a pending session.

packages/clerk-js/src/ui/contexts/components/OrganizationSwitcher.ts
```diff
@@ -94,9 +97,18 @@ export const useOrganizationSwitcherContext = () => {
...
 
+  const hidePersonal = organizationSettings.forceOrganizationSelection || ctx.hidePersonal || false;
+
+  clerk?.telemetry?.record(
+    eventComponentMounted('OrganizationList', {
+      hidePersonal,
+      forceOrganizationSelection: organizationSettings.forceOrganizationSelection,
+    }),
+  );
+
```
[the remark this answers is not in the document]
LauraBeatris: Fixed on https://github.com/clerk/javascript/pull/5391/commits/21a931c17ebe79bd2c09d03e441f30d6a8455b29

PR #5395, opened 2025-03-19
fix(backend): Handle Basic auth gracefully

jacekradko:
## Description

Handle cases where we get Authorization header with basic auth, like `Basic ....` instead of `Bearer ....`

Right now we would throw an error since we assume any authorization header contains a `Bearer` token.

Also switched from `.replace` to `.slice` for perf since we know the string will begin with `Basic ` or `Bearer ` 

Fixes: SDKI-896

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/ninety-bugs-jam.md
+++ b/.changeset/ninety-bugs-jam.md
@@ -1,0 +1,6 @@
+---
+'@clerk/backend': patch
+'@clerk/nextjs': patch
+---
+
+Handle Basic auth gracefully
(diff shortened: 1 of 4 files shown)


brkalow:
@jacekradko is this breaking? It looks like right now we technically accept an authorization header without the Bearer prefix

jacekradko:
@brkalow I guess technically yes. But the Authorization header does require an auth scheme to be set: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Authorization

tmilewski (approved):

PR #5422, opened 2025-03-21
chore(clerk-js,types): Load tasks based on environment settings

LauraBeatris:
## Description

This PR removes the previous `experimental.withSessionTasks` flag in favor of preloading `SessionTasks` component based on the current environment settings, to avoid downloading its chunk of JS for instances that aren't using this feature.

#### Breakdown 

Main changes:
- Introduce a new root path (`/tasks`) so that `SessionTasks` can declare inner routes to be reused across `SignIn/SignUp`
- With the above, `SessionTasks` can now be used as standalone, outside of `SignIn/SignUp` paths, such as rendering it on a catchall (`/onboarding/[[...index]].tsx`) -> coming in a next PR
- Also with the root path, allows to redirect from `@clerk/backend` when the incoming JWT claims contain `sts` pending -> coming in a next PR 

Misc:
- Introduce webpack chunk for `SessionTasks` on bundlewatch to track size across PRs 
- Fix nomenclature to be consistent across `tasks` on plural, since that component can have multiple tasks routes / render multiple tasks 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/lazyModules/components.ts
```diff
@@ -95,9 +95,9 @@ export const PricingTable = lazy(() =>
   componentImportPaths.PricingTable().then(module => ({ default: module.__experimental_PricingTable })),
 );
 
-export const preloadSessionTask = () => import(/* webpackChunkName: "sessionTask" */ '../components/SessionTask');
-export const SessionTask = lazy(() =>
-  componentImportPaths.SessionTask().then(module => ({ default: module.SessionTask })),
+export const preloadSessionTasks = () => import(/* webpackChunkName: "sessionTasks" */ '../components/SessionTasks');
+export const SessionTasks = lazy(() =>
+  componentImportPaths.SessionTasks().then(module => ({ default: module.SessionTask })),
```
LauraBeatris, line 100:
```suggestion
```

LauraBeatris:
Here are a few considerations when it comes to optimizations:

- It's not possible to prefetch FAPI organization memberships ahead of navigating to a task. By the time the `SignIn/SignUp` component loads, the user hasn't had a session yet. 
- Lazy loading is being applied to the entire `SessionTasks` within `SignIn/SignUp` 
    - I've experimented with lazy loading the inner routes to `SessionTasks` based on the environment settings, eg: the upcoming billing task wouldn't get downloaded if there's only a force org task, however, it'd require dealing with `Suspense` and possibility leading to a waterfall of loading -> it's a tradeoff tho, we can revisit this later

packages/clerk-js/src/ui/components/SessionTasks/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const clerk = useClerk();
+  const { navigate } = useRouter();
+  const { redirectUrlComplete } = useSessionTasksContext();
+
+  useEffect(() => {
+    // Simulates additional latency to avoid a abrupt UI transition when navigating to the next task
+    const timeoutId = setTimeout(() => {
+      void clerk.__experimental_nextTask({ redirectUrlComplete });
+    }, 500);
+    return () => clearTimeout(timeoutId);
```
NicolasLopes7, line 26: Just to make sure I'll parse correctly, that will display the loading spinner for 500ms and then redirect to the next task?

How do they land on this page? After a sign-in/sign-up (?)

packages/clerk-js/src/ui/hooks/usePreloadTasks.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Preloads tasks based on the environment settings
+ * @internal
+ */
+export function usePreloadTasks() {
+  const { organizationSettings } = useEnvironment();
+
+  const hasTasks = organizationSettings.forceOrganizationSelection;
+
+  void useFetch(
+    hasTasks ? () => import(/* webpackChunkName: "sessionTasks" */ '../components/SessionTasks') : undefined,
```
NicolasLopes7, line 14: :chef-kiss:

packages/clerk-js/src/ui/components/SessionTasks/index.tsx
LauraBeatris: They land on this PR from a server-side redirection from `await auth()` -> it's implemented here https://github.com/clerk/javascript/pull/5440

LauraBeatris: There's the argument of consistency, and always redirect to `tasks` on first `Clerk.nextTask` execution. 

I'm against that cause I think it's an unnecessary additional latency since we already have the client session loading. The whole point of this root path component is to resolve the client tasks since the server isn't aware of it.

packages/clerk-js/src/ui/components/SessionTasks/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+ * @internal
+ */
+export function SessionTask(): JSX.Element {
+  const clerk = useClerk();
+  const { navigate } = useRouter();
+  const signInContext = useContext(SignInContext);
+  const signUpContext = useContext(SignUpContext);
+
+  const redirectUrlComplete =
+    signInContext?.afterSignInUrl ?? signUpContext?.afterSignUpUrl ?? clerk?.buildAfterSignInUrl();
```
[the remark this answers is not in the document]
LauraBeatris: We have `Clerk.buildAfterSignUpURL()`, but it only takes the ones provided to `Clerk.load`

`SignInContext` and `SignUpContext` resolve those URLs from env vars as well, or even from the middleware options.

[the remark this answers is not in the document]
LauraBeatris: > I think the logic will be clearer if we check the context of sign-in/up and if it's valid, then we use the URL from the right context

I believe we're already doing it here, but let me know if I'm getting it wrong or if the code is not clear. If `SessionTask` is not within `SignIn`, then `signInContext` is the undefined, same thing happens for sign-up.

```clerk?.buildAfterSignInUrl();``` is sort of to satisfy the compiler on that union scenario. There's a case in which we don't support yet where `SessionTask` gets rendered outside of `SignIn/SignUp` and therefore it would need another way to get the redirect URL, either a new option or a prop

File: packages/clerk-js/src/ui/components/SignIn/SignIn.tsx [hunk 5]
```diff
@@ -2,7 +2,7 @@ import { useClerk } from '@clerk/shared/react';
 import type { SignInModalProps, SignInProps } from '@clerk/types';
 import React from 'react';
 
-import { SESSION_TASK_ROUTE_BY_KEY } from '../../../core/sessionTasks';
+import { SessionTasks } from '../../../ui/lazyModules/components';
```
panteliselef, line 5:
```suggestion
import { SessionTasks as LazySessionTasks} from '../../../ui/lazyModules/components';
```

packages/clerk-js/src/ui/components/SignIn/SignIn.tsx
```diff
@@ -130,14 +130,9 @@ function SignInRoutes(): JSX.Element {
...
-              {signInContext.withSessionTasks && (
-                <Route path={SESSION_TASK_ROUTE_BY_KEY['org']}>
-                  <SessionTask
-                    task='org'
-                    redirectUrlComplete={signInContext.afterSignUpUrl}
-                  />
-                </Route>
-              )}
+              <Route path='tasks'>
+                <SessionTasks />
```
panteliselef, line 134:
```suggestion
                <LazySessionTasks />
```

packages/clerk-js/src/ui/components/SignIn/SignIn.tsx
```diff
@@ -147,14 +142,9 @@ function SignInRoutes(): JSX.Element {
...
-        {signInContext.withSessionTasks && (
-          <Route path={SESSION_TASK_ROUTE_BY_KEY['org']}>
-            <SessionTask
-              task='org'
-              redirectUrlComplete={signInContext.afterSignInUrl}
-            />
-          </Route>
-        )}
+        <Route path='tasks'>
+          <SessionTasks />
```
panteliselef, line 146:
```suggestion
          <LazySessionTasks />
```

packages/clerk-js/src/ui/components/SignUp/SignUp.tsx
```diff
@@ -2,22 +2,18 @@ import { useClerk } from '@clerk/shared/react';
 import type { SignUpModalProps, SignUpProps } from '@clerk/types';
 import React from 'react';
 
-import { SESSION_TASK_ROUTE_BY_KEY } from '../../../core/sessionTasks';
+import { SessionTasks } from '../../../ui/lazyModules/components';
```
panteliselef, line 5:
```suggestion
import { SessionTasks as LazySessionTasks } from '../../../ui/lazyModules/components';
```

panteliselef (approved):

PR #5447, opened 2025-03-24
feat(clerk-js,types): Update `Protect` to consider pending session task as unauthorized state

LauraBeatris:
## Checklist

## Type of change

- [X] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


LauraBeatris:
Just managed to replicate this behavior where, if I don't have a sign-in URL set either as a middleware option, or an env var  + Account Portal is not enabled, it fallbacks to `/tasks` on my application


https://github.com/user-attachments/assets/8a313243-47a3-4956-aa46-63e4434980d2

Is there a more assertive way to figure out the sign in URL here?

packages/shared/src/deriveState.ts
```diff
@@ -22,6 +22,7 @@ const deriveFromSsrInitialState = (initialState: InitialState) => {
   const userId = initialState.userId;
   const user = initialState.user as UserResource;
   const sessionId = initialState.sessionId;
+  const sessionStatus = initialState.sessionClaims?.sts;
```
panteliselef, line 25: I cannot recall if we are actually populating `sessionClaims` here. For `factorVerificationAge` we are passing it explicitly instead of inferring it. This is merely an observation, if you've verified this you can safely ignore.

panteliselef (approved):

packages/shared/src/deriveState.ts
LauraBeatris: I actually would prefer to also explicitly pass the `sessionStatus` here to be consistent then! It's not a big effort to change it now, I'll push a commit later

LauraBeatris:
Updated on https://github.com/clerk/javascript/pull/5447/commits/9f4364639fff578c7580d3996f8336e9fe0f4217 to forward `sessionStatus` explicitly from the server auth state 

I want to test a bit more before merging it, I will continue once I'm back later this morning!

PR #5453, opened 2025-03-26
chore(repo): Update Typedoc to 0.28

LekoArts:
## Description

This PR updates our Typedoc version to 0.28, see its changelog: https://typedoc.org/documents/Changelog.html#v0280-2025-03-15

There were a couple of breaking changes in this version that mostly the markdown plugin had to handle, but since we override parts of its theme I had to update our code, too. So that's why there are changes in `custom-theme.mjs`.

The reason why I did the upgrade was that Typedoc 0.28 added a `--router` option which now allows for a better way to modify the output.

Fixes ECO-507

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.typedoc/custom-theme.mjs
LekoArts, line 1: The changes here are accounting for upstream changes in the markdown plugin and the breaking changes in Typedoc itself

packages/react/src/components/controlComponents.tsx
```diff
@@ -169,6 +169,7 @@ export const RedirectToSignUp = withClerk(({ clerk, ...props }: WithClerkProp<Re
 }, 'RedirectToSignUp');
 
 /**
+ * @function
```
LekoArts, line 172: Typedoc is now more strict about `const foo = () => {}` vs `function bar() {}`.

Without the `@function` keyword it marks the former as a variable, not a function. Since I didn't want to change the function signature everywhere (since that should be enforced with ESLint) I just added the annotations

PR #5457, opened 2025-03-26
chore(react,types): Update typedoc comments for react hooks and clerkprovider

alexisintech:
## Description

Related to Docs PR: https://github.com/clerk/clerk-docs/pull/2124

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/types/src/hooks.ts
```diff
@@ -37,7 +37,7 @@ export type UseAuthReturn =
        */
       sessionId: undefined;
       /**
-       * JWT Actor - [RFC8693](https://www.rfc-editor.org/rfc/rfc8693.html#name-act-actor-claim).
+       * The JWT actor for the session. Holds identifier for the user that is impersonating the current user. Read more about [impersonation](/docs/users/user-impersonation).
```
alexcarpenter, line 40: How will this work for folks in their IDEs (I understand this is for docs generation). The link will not resolve to anywhere right? Do we need these to be full docs URLs?

LekoArts: Yes, you're correct @alexcarpenter, the links will need to be absolute. We then later replace those absolute links with relative ones:

https://github.com/clerk/javascript/blob/3c47ce8dc328fc1613e9156ba92c160f89447968/typedoc.config.mjs#L44-L52

alexisintech: great point - I missed this one as I copied and pasted it, thanks for the eagle eyes 👀

LekoArts (approved):

PR #5476, opened 2025-03-28
feat(clerk-react,clerk-js, types,shared): Introduce `Clerk.status`

panteliselef:
## Description

Introduce `useClerk().status`, `Clerk.status` alongside `<ClerkFailed/>` and `<ClerkDegraded/>`.

### `useClerk().status`
Possible values are:
- `loading` (initial state)
- `error` set when hotloading clerk-js failed or `Clerk.load()` failed
- `ready` set when Clerk is fully operational
- `degraded` set when Clerk is partially operational

`useClerk().loaded` is `true` when `useClerk().status` is either `"ready"` or `"degraded"`.
`useClerk().loaded` is `false` when `useClerk().status` is `"loading"` or `"error"`.

# Important
**🚨🚨 `Clerk.status` becomes `loading` upon instantiation not upon calling `.load()`**

# New APIs
```ts
const handler = (status)=> {}
window.Clerk.on('status', handler)
window.Clerk.off('status', handler)
```

### on()

Calls handler on every status event with the new status as parameter
```ts
window.Clerk.on('status',(status)=>{})
```

Calls handler immediately with the last payload of the event.
```ts
window.Clerk.on('status',(status)=>{}, {notify:true})
```

### off()

Accepts a handler to unregister from the specified event.
```ts
window.Clerk.off('status',handler)
```

Removes all handlers for the specified event.
```ts
window.Clerk.off('status')
```

# New Components

### `<ClerkFailed/>`
```tsx
<ClerkLoaded>
  <MyCustomSignInForm/>
</ClerkLoaded>
<ClerkFailed>
  <ContactSupportBanner/>
</ClerkFailed>
```

### `<ClerkDegraded/>`
```tsx
<ClerkLoaded>
  <MyCustomPasskeyRegistration/>
  <ClerkDegraded>
    We are experiencing issues, registering a passkey might fail. 
  </ClerkDegraded>
</ClerkLoaded>
```

### Clerk.status values
image

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/types/src/clerk.ts
```diff
@@ -90,6 +91,11 @@ export interface SignOut {
   (signOutCallback?: SignOutCallback, options?: SignOutOptions): Promise<void>;
 }
 
+/**
+ * @inline
+ */
+export type Status = 'degraded' | 'error' | 'loading' | 'ready';
```
panteliselef, line 97: Maybe `ClerkStatus` ? wdyt ?

.changeset/cool-showers-admire.md
```diff
@@ -0,0 +1,36 @@
+---
+'@clerk/nextjs': minor
+'@clerk/clerk-react': minor
+---
+
+Introduce `useClerk().status` alongside `<ClerkFailed/>` and `<ClerkDegraded/>`.
```
LekoArts, line 6: nit

```suggestion
Introduce `useClerk().status` alongside `<ClerkFailed />` and `<ClerkDegraded />`.
```

.changeset/cool-showers-admire.md
```diff
@@ -0,0 +1,36 @@
...
+
+### `useClerk().status`
+Possible values are:
+- `loading` set on during initialization
+- `error` set when hotloading clerk-js failed or `Clerk.load()` failed
+- `ready` set when Clerk is fully operational
+- `degraded` set when Clerk is partially operational
+
+`useClerk().loaded` is `true` when `useClerk().status` is either `"ready"` or `"degraded"`.
+`useClerk().loaded` is `false` when `useClerk().status` is `"loading"` or `"error"`.
```
LekoArts, line 16:
```suggestion
The computed value of `useClerk().loaded` is:

- `true` when `useClerk().status` is either `"ready"` or `"degraded"`.
- `false` when `useClerk().status` is `"loading"` or `"error"`.
```

.changeset/cool-showers-admire.md
```diff
@@ -0,0 +1,36 @@
...
+
+### `<ClerkFailed/>`
```
LekoArts, line 18: nit

```suggestion
### `<ClerkFailed />`
```

.changeset/cool-showers-admire.md
```diff
@@ -0,0 +1,36 @@
...
+```tsx
+<ClerkLoaded>
+  <MyCustomSignInForm/>
+</ClerkLoaded>
+<ClerkFailed>
+  <ContactSupportBanner/>
+</ClerkFailed>
+```
+
+### `<ClerkDegraded/>`
```
LekoArts, line 28: nit

```suggestion
### `<ClerkDegraded />`
```

.changeset/cool-showers-admire.md
```diff
@@ -0,0 +1,36 @@
...
+---
+
+Introduce `useClerk().status` alongside `<ClerkFailed/>` and `<ClerkDegraded/>`.
+
+### `useClerk().status`
+Possible values are:
+- `loading` set on during initialization
+- `error` set when hotloading clerk-js failed or `Clerk.load()` failed
+- `ready` set when Clerk is fully operational
+- `degraded` set when Clerk is partially operational
```
LekoArts, line 13:
```suggestion
Possible values for `useClerk().status` are:
- `"loading"`: Set during initialization
- `"error"`: Set when hotloading clerk-js failed or `Clerk.load()` failed
- `"ready"`: Set when Clerk is fully operational
- `"degraded"`: Set when Clerk is partially operational
```

.changeset/olive-onions-do.md
```diff
@@ -0,0 +1,15 @@
+---
+'@clerk/clerk-js': minor
+'@clerk/clerk-react': minor
```
LekoArts, line 3: Should we remove clerk-react here?

.changeset/olive-onions-do.md
```diff
@@ -0,0 +1,15 @@
...
+'@clerk/types': minor
+---
+
+Introduce `Clerk.status` for tracking the state of the clerk singleton.
+Possible values are:
+- `loading` set on during initialization
+- `error` set when hotloading clerk-js failed or `Clerk.load()` failed
+- `ready` set when Clerk is fully operational
+- `degraded` set when Clerk is partially operational
```
LekoArts, line 12:
```suggestion
Possible values for `Clerk.status` are:
- `"loading"`: Set during initialization
- `"error"`: Set when hotloading clerk-js failed or `Clerk.load()` failed
- `"ready"`: Set when Clerk is fully operational
- `"degraded"`: Set when Clerk is partially operational
```

.changeset/olive-onions-do.md
```diff
@@ -0,0 +1,15 @@
...
+
+`Clerk.loaded` is `true` when `Clerk.status` is either `ready` or `degraded`.
+`Clerk.loaded` is `false` when `Clerk.status` is `loading` or `error`.
```
LekoArts, line 15:
```suggestion
The computed value of `Clerk.loaded` is:
- `true` when `Clerk.status` is either `"ready"` or `"degraded"`.
- `false` when `Clerk.status` is `"loading"` or `"error"`.
```

integration/tests/resiliency.test.ts
```diff
@@ -4,9 +4,27 @@ import { appConfigs } from '../presets';
...
+      },
+    ],
+    clerk_trace_id: 'some-trace-id',
+  }),
+});
+
 testAgainstRunningApps({ withEnv: [appConfigs.envs.withEmailCodes] })('resiliency @generic', ({ app }) => {
   test.describe.configure({ mode: 'serial' });
 
+  if (app.name.includes('next')) {
```
LekoArts, line 24: Why that skip?

packages/react/src/components/controlComponents.tsx
```diff
@@ -49,7 +49,28 @@ export const ClerkLoading = ({ children }: React.PropsWithChildren<unknown>) =>
...
+    return null;
+  }
+  return children;
+};
+
+export const ClerkDegraded = ({ children }: React.PropsWithChildren<unknown>) => {
+  useAssertWrappedByClerkProvider('ClerkDegraded');
+
+  const isomorphicClerk = useIsomorphicClerkContext();
+  // What if status form  clerk js is not available ?
```
LekoArts, line 72: Is this still a TODO for you?

packages/types/src/clerk.ts
LekoArts: `ClerkStatus` 👍

packages/types/src/clerk.ts
```diff
@@ -110,6 +116,15 @@ export interface Clerk {
    */
   loaded: boolean;
 
+  /**
+   * Describes the state the clerk singleton operates in.
+   * If `error`, it means Clerk failed to initialize.
+   * If `loading`, it means Clerk is still attempting to load.
+   * If `ready`, it means Clerk singleton is fully operational.
+   * If `degraded`, it means Clerk singleton has partially operational.
```
LekoArts, line 124:
```suggestion
   * Describes the state the clerk singleton operates in:
   * - `"error"`: Clerk failed to initialize.
   * - `"loading"`: Clerk is still attempting to load.
   * - `"ready"`: Clerk singleton is fully operational.
   * - `"degraded"`: Clerk singleton is partially operational.
```

LekoArts (dismissed):

integration/tests/resiliency.test.ts
panteliselef: Because, we'd have to implement the same page on the next template, and there is no difference between the 2 implementations.

LekoArts: I guess to better be safe than sorry we could copy the page. Then you'd catch something easy like "it's not re-exported in Next.js anymore". Or in the future an actual regression if there is difference in behavior.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -200,9 +201,9 @@ export class Clerk implements ClerkInterface {
   //@ts-expect-error with being undefined even though it's not possible - related to issue with ts and error thrower
   #fapiClient: FapiClient;
   #instanceType?: InstanceType;
-  #loaded = false;
-
+  #status: ClerkInterface['status'] = 'loading';
   #listeners: Array<(emission: Resources) => void> = [];
+  #statusListeners: Array<(status: Clerk['status']) => void> = [];
```
jacekradko, line 206: I think we need a more general event pattern in the core Clerk class. Adding on additional properties with corresponding getters/setters to the core class is not sustainable in the long term

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1172,6 +1183,26 @@ export class Clerk implements ClerkInterface {
...
+  public __internal_setStatus = (status: Clerk['status'], opts = { notify: true }): void => {
+    this.#status = status;
+    if (!opts?.notify) {
+      return;
+    }
+
+    for (const listener of this.#statusListeners) {
+      listener(status);
+    }
+  };
```
jacekradko, line 1204: We should encapsulate both the state change and the notification logic so that clients of the class aren’t forced to know about or interact with internal listeners. Instead of exposing an “internal” method that both updates state and conditionally notifies listeners via an options flag, this should be private to the Clerk class to keep the abstraction free of internal implementation details

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2118,27 +2149,26 @@ export class Clerk implements ClerkInterface {
     const isInAccountsHostedPages = isDevAccountPortalOrigin(window?.location.hostname);
     const shouldTouchEnv = this.#instanceType === 'development' && !isInAccountsHostedPages;
 
+    let initializationDegradedCounter = 0;
```
jacekradko, line 2152: As a follow up let's design a more robust system of determining when the instance is in a degraded state

packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -101,16 +103,18 @@ const useLoadedIsomorphicClerk = (options: IsomorphicClerkOptions) => {
...
-    isomorphicClerk.addOnLoaded(() => setLoaded(true));
+  useEffect(() => {
+    const unsub = isomorphicClerk.addStatusListener(setStatus);
+    return () => unsub();
   }, []);
 
   React.useEffect(() => {
     return () => {
+      // reset to initial loading state
+      isomorphicClerk.__internal_setStatus('loading');
```
jacekradko, line 114: Shouldn't this be part of `IsomorphicCler.clearInstance`? This is sub-optimal as the internal implementation details are leaking into the outside context

panteliselef: We can if status is static and not per instance.

packages/react/src/hooks/utils.ts
```diff
@@ -5,10 +5,11 @@ import type { IsomorphicClerk } from '../isomorphicClerk';
...
   return new Promise<void>(resolve => {
-    if (isomorphicClerk.loaded) {
-      resolve();
-    }
-    isomorphicClerk.addOnLoaded(resolve);
+    isomorphicClerk.addStatusListener(status => {
+      if (['ready', 'degraded'].includes(status)) {
+        resolve();
+      }
+    });
```
jacekradko, line 12: This goes along with the `isClerkJSOperational` improvement to encapsulate it in the Clerk class

packages/types/src/clerk.ts
```diff
@@ -422,6 +437,20 @@ export interface Clerk {
...
+   * @returns - Unsubscribe callback
+   */
+  addStatusListener: (listener: StatusListenerCallback) => UnsubscribeCallback;
+
+  /**
+   * Updates the status of Clerk singleton based on the provided argument.
+   * By default, it notifies subscribers wit the new status.
+   * @internal
+   */
+  __internal_setStatus: (status: Status, opts?: { notify: boolean }) => void;
```
jacekradko, line 452: I don't think either of these methods should be a part of the public API. We already have a concept of listeners along with `addListener` and setting `status` should not be an external concern

packages/react/src/hooks/utils.ts
panteliselef: Can you give an example of how a developer would consume that ?

jacekradko: Isn't this essentially `clerk.loaded` ?

panteliselef: It is, but i'd like to remove `addOnLoaded`.

packages/types/src/clerk.ts
panteliselef: I completely removed  `__internal_setStatus`, and replaced `addStatusListener` with `on('status')` and `off('status')`

packages/clerk-js/src/core/clerk.ts
panteliselef: Sure thing, I think some kind of buffer in the event bus could potentially address it.

.changeset/olive-onions-do.md
```diff
@@ -0,0 +1,15 @@
+---
+'@clerk/clerk-js': minor
+'@clerk/clerk-react': minor
+'@clerk/types': minor
+---
```
panteliselef, line 5:
```suggestion
'@clerk/clerk-js': minor
'@clerk/types': minor
---
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2090,15 +2113,20 @@ export class Clerk implements ClerkInterface {
     }
   };
 
-  #loadInStandardBrowser = async (): Promise<boolean> => {
+  #loadInStandardBrowser = async (): Promise<void> => {
```
jacekradko, line 2116: We might want to add JSDoc explaining this will now throw instead of always returning a boolean

packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -31,8 +31,14 @@ export function ClerkContextProvider(props: ClerkContextProvider) {
     return clerk.addListener(e => setState({ ...e }));
   }, []);
 
-  const derivedState = deriveState(clerkLoaded, state, initialState);
-  const clerkCtx = React.useMemo(() => ({ value: clerk }), [clerkLoaded]);
+  const derivedState = deriveState(clerk.loaded, state, initialState);
+  const clerkCtx = React.useMemo(
+    () => ({ value: clerk }),
+    [
+      // Only upda the clerk reference on status change
```
jacekradko, line 38:
```suggestion
      // Only update the clerk reference on status change
```

packages/clerk-js/src/core/clerk.ts
panteliselef: `#loadInStandardBrowser` would throw even before this PR. It's not that this PR is changing this behaviour.

jacekradko: Ok, this makes more sense then 👍🏼

integration/tests/resiliency.test.ts
panteliselef: If we had `@react` I'd use that tbh, this is why i wanted to skip for next. I don't there any value to added tests for nextjs, since the implementation only affects react. This is the reason we don't have tests like these for every react meta framework.

packages/react/src/components/controlComponents.tsx
panteliselef: yeah, the current implementation accounts for it, so I simply removed the comment.

packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -1,7 +1,7 @@
 import { deriveState } from '@clerk/shared/deriveState';
 import { ClientContext, OrganizationProvider, SessionContext, UserContext } from '@clerk/shared/react';
 import type { ClientResource, InitialState, Resources } from '@clerk/types';
-import React from 'react';
+import React, { useEffect } from 'react';
```
LekoArts, line 4: Nit: Your editor probably added this, throughout the file we use `React.` and even `React.useEffect`

packages/react/src/isomorphicClerk.ts
```diff
@@ -458,12 +499,22 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
...
   };
 
+  private beforeLoad = (clerkjs: BrowserClerk | HeadlessBrowserClerk | undefined) => {
+    if (!clerkjs) {
+      throw new Error('Failed to hydrate latest Clerk JS');
+    }
+  };
+
   private hydrateClerkJS = (clerkjs: BrowserClerk | HeadlessBrowserClerk | undefined) => {
+    console.log('hydrateClerkJS');
```
LekoArts, line 517: Stray console.log

packages/types/src/clerk.ts
```diff
@@ -90,6 +90,21 @@ export interface SignOut {
...
 }
 
+type ClerkEvent = keyof EventPayload;
+type EventHandler<E extends ClerkEvent> = (payload: EventPayload[E]) => void;
+
+type EventPayload = {
+  status: ClerkStatus;
+};
+
+type OnEventListener = <E extends ClerkEvent>(event: E, handler: EventHandler<E>, opt?: { notify: boolean }) => void;
```
LekoArts, line 100: Isn't the `dispose` option missing?

LekoArts (commented):
Overall LGTM, some smaller comments

packages/types/src/clerk.ts
panteliselef: this has changed. We don't need dispose.

packages/clerk-js/src/core/constants.ts
```diff
@@ -43,11 +43,18 @@ export const SIGN_UP_INITIAL_VALUE_KEYS = ['email_address', 'phone_number', 'use
...
   WAITLIST: 'waitlist',
-};
+} satisfies Record<string, SignUpModes>;
+
+export const CLERK_STATUS = {
+  DEGRADED: 'degraded',
+  READY: 'ready',
+  LOADING: 'loading',
+  ERROR: 'error',
+} satisfies Record<string, ClerkStatus>;
```
LekoArts, line 57: You could use `Record<Uppercase<ClerkStatus>, ClerkStatus>` instead of `string`
https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#uppercasestringtype

panteliselef: I ended up removing this

jacekradko:
!snapshot

packages/shared/src/clerkEventBus.ts
```diff
@@ -0,0 +1,11 @@
+import type { ClerkEventPayload } from '@clerk/types';
+
+import { createEventBus } from './eventBus';
+
+export const clerkEvents = {
+  Status: 'status',
+} satisfies Record<string, keyof ClerkEventPayload>;
+
+export const createClerkEventBus = () => {
```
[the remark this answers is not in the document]
panteliselef: This is shared between clerk-react and clerk-js. I think the benefit is the type safety it provides, and exports the supported events in a separate variable which i a pattern we already follow for the internal bus of clerk-js.

jacekradko (approved):

PR #5481, opened 2025-03-28
feat(clerk-js,types,clerk-react): Add internal open/close checkout methods

alexcarpenter:
## Description

Introduce `clerk.__internal_openCheckout()` and `clerk.__internal_closeCheckout()` methods, and move `<Checkout />` component out of the `<PricingTable />` component.

Resolves COM-257

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/Components.tsx
```diff
@@ -322,6 +344,26 @@ const Components = (props: ComponentsProps) => {
...
+
+    componentsControls.closeDrawer = name => {
+      setState(s => ({
+        ...s,
+        [`${name}Drawer`]: {
+          ...s[`${name}Drawer`],
+          open: false,
+        },
+      }));
+    };
```
alexcarpenter, line 365: Introducing new drawer specific methods as we want to have an exit animation for the drawers, and removing all the props causes issues with portaling into the profile components.

packages/clerk-js/src/ui/components/PricingTable/PricingTable.tsx
```diff
@@ -40,8 +37,13 @@ export const __experimental_PricingTable = (props: __experimental_PricingTablePr
...
       setShowSubscriptionDetailDrawer(true);
     } else {
-      setCheckoutPlan(plan);
-      setShowCheckout(true);
+      clerk.__internal_openCheckout({
+        planId: plan.id,
+        planPeriod,
+        orgId: subscriberType === 'org' ? organization?.id : undefined,
+        onSubscriptionComplete: onSubscriptionChange,
+        portalId: mode === 'modal' ? PROFILE_CARD_SCROLLBOX_ID : undefined,
```
brkalow, line 45: does `mode` indicate the context `PricingTable` is rendered in? I feel like I would expect `mode` to dictate how PricingTable behaves itself, but not necessarily how it behaves depending on where it's rendered.

As a semi-related nit, I might rename `mode` to `context`.

brkalow, line 45: This is much cleaner!

packages/clerk-js/src/ui/Components.tsx
brkalow: It would be great to consolidate the modal/drawer handling at some point in the future. I would expect them to function the same, just be a different flavor of UI. Maybe there's a good reason to keep them separate though? 🤔

brkalow (approved):
👍 works good, I like the idea here. Main comment is I'm not sure the drawer/modal distinction is valuable in the low level component plumbing, but we can consolidate later if we want.

Nice job

packages/clerk-js/src/ui/components/PricingTable/PricingTable.tsx
alexcarpenter: Mode is set on the PricingTable and I think was originally adding following the other components, but there is no modal mode for PricingTable, so yeah, it is confusing, as its only a mounted component. Will refactor naming approach in a follow up PR.

packages/clerk-js/src/ui/Components.tsx
alexcarpenter: If we want modals to have an exit animation, we'd need to follow this updated drawer pattern, currently all modals are removed immediately but wiping out the props.

PR #5492, opened 2025-03-31
feat(clerk-js): Add payment source tab to `UserProfile`

aeliox:
## Description

Adds the payment sources tab to the Billing section of `UserProfile`, allowing for the removing and adding of payment sources.

Also extracts out the Stripe Elements into a reusable `AddPaymentSource` component that is used both here and now within the `Checkout` flow.

Screenshot 2025-03-31 at 8 32 49 AM

https://linear.app/clerk/issue/COM-183/build-out-payment-sources-tab-within-userprofile

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/components/PaymentSources/AddPaymentSource.tsx
```diff
@@ -0,0 +1,258 @@
...
+              type='submit'
+              colorScheme='primary'
+              size='sm'
+              textVariant={'buttonLarge'}
+              block
+              isLoading={isSubmitting}
+            >
+              {submitButtonText ?? 'Add Payment Method'}
+            </Button>
+          </Flex>
```
alexcarpenter, line 253: We should include the heading and button layout/styling that we use for adding an email address. We have a FormButtons component that we could maybe use here?

https://github.com/clerk/javascript/blob/5bd3cbb0a19e0b13b812453014aca9fc3418d98a/packages/clerk-js/src/ui/elements/FormButtons.tsx#L15-L33

Screenshot 2025-03-31 at 2 05 28 PM

packages/clerk-js/src/ui/components/PaymentSources/PaymentSources.tsx
```diff
@@ -0,0 +1,198 @@
...
+                    </Flex>
+                    <PaymentSourceMenu paymentSource={paymentSource} />
+                  </ProfileSection.Item>
+
+                  <Action.Open value={`remove-${paymentSource.id}`}>
+                    <Action.Card variant='destructive'>
+                      <RemoveScreen
+                        paymentSource={paymentSource}
+                        revalidate={revalidate}
+                      />
```
alexcarpenter, line 160: Seeing multiple spinners when deleting a payment source. I think we should only see it on the delete button and not over the content.

Screenshot 2025-03-31 at 2 05 28 PM

packages/clerk-js/src/ui/components/PaymentSources/AddPaymentSource.tsx
```diff
@@ -0,0 +1,258 @@
...
+      stripe={stripe}
+      options={{ clientSecret: externalClientSecret, appearance: elementsAppearance }}
+    >
+      <AddPaymentSourceForm
+        submitButtonText={submitButtonText}
+        onSuccess={onSuccess}
+        onExpand={onExpand}
+        cancelAction={cancelAction}
+        cancelButtonText={cancelButtonText}
+      />
```
alexcarpenter, line 97: Should we be seeing this 1 click checkout during adding a new source within the payment sources tab. Would assume this is only relevant within checkout? Not familiar with what this does tbh.

Screenshot 2025-03-31 at 2 13 00 PM

aeliox: I'm not super familiar with this either – seems to make sense that we figure out how to hide this in both payment sources and checkout, since we're "saving the info" regardless...

packages/types/src/appearance.ts
```diff
@@ -830,4 +831,8 @@ export type Appearance<T = Theme> = T & {
    * Theme overrides that only apply to the `<Checkout />` component
    */
   checkout?: T;
+  /**
+   * Theme overrides that only apply to the `<PaymentSources />` component
+   */
+  paymentSources?: T;
```
alexcarpenter, line 837: hm, I wouldn't expect this to have its own appearance theme. Wouldn't this inherit the appearance provider from the component its placed within?

alexcarpenter (commented):
Update looking good, just a question about the addition of the payment sources theme.

PR #5495, opened 2025-03-31
feat(clerk-js): Introduce legacy browser variant

dstaley:
## Description

This PR introduces a new `legacy` variant of `clerk.browser`, intended to be used for instances that require compatibility with browsers released in mid-2019 or newer. This is achieved through the use of `core-js`, which polyfills modern JavaScript APIs, in addition to the language-level transformations provided by SWC.

We now emit two primary `clerk.browser` bundles:
- `clerk.browser.js`: This is our "modern"/default bundle that adheres to our browser support policy as specified by `browserslist` in `package.json`. This bundle includes language-level transformations for supported browsers, but does _not_ include `core-js` polyfills. This is because `core-js` will polyfill divergent behaviors between browsers, inflating our bundle size.
- `clerk.legacy.browser.js`: This is our legacy bundle that adheres to the browser support policy specified by `browserslistLegacy` in `package.json`, which supports browsers released in mid-2019 and later. It includes language-level transformations from SWC in addition to `core-js` polyfills. It is an additional 35KB gzipped compared to the default `clerk.browser.js` bundle.

In addition to the bundles, we also emit separate chunks for the `legacy` variant. `clerk.legacy.browser.js` will load legacy-compatible chunks, whereas `clerk.browser.js` will load modern-compatible chunks.

As of writing, it's not intended for customers to be able to manually switch to the `legacy` bundle since it's outside our official browser support matrix. We will be enabling this at the Cloudflare proxy level only for specific customers who have an agreement with us and a demonstrated need for legacy browser compatibility.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/package.json
```diff
@@ -94,5 +94,6 @@
   },
   "publishConfig": {
     "access": "public"
-  }
+  },
+  "legacybrowserslist": "Chrome > 73, Firefox > 66, Safari > 12, iOS > 12, Edge > 18, Opera > 58"
```
dstaley, line 98: I was hoping we could keep this next to the `"browserslist"` property, but the plugin we use to sort fields in `package.json` files can't be configured that way without specifying the order of _all_ fields.

packages/clerk-js/rspack.config.js
```diff
@@ -356,54 +387,66 @@ const prodConfig = ({ mode, env, analysis }) => {
...
-      // Include the lazy chunks in the bundle as well
-      // so that the final bundle can be imported and bundled again
-      // by a different bundler, eg the webpack instance used by react-scripts
-      new rspack.optimize.LimitChunkCountPlugin({
-        maxChunks: 1,
-      }),
-    ],
-    optimization: {
-      splitChunks: false,
+  const clerkEsm = merge(
```
dstaley, line 394: none of this was actually changed, it's just a formatting thing

packages/clerk-js/package.json
LekoArts: How about `"browserslistLegacy"` then?

packages/clerk-js/rspack.config.js
LekoArts: Where does the formatting change come from then? Your personal prettier settings? If I open the file in my editor ESLint doesn't want to format the file.

Ideally such unintended formatting changes don't make it into the PR as it's quite possible that another PR will undo them later again.

LekoArts (commented):
I'd argue that with this change we should also change our existing `browserslist` entry and remove the legacy information (`Safari > 12, iOS > 12`), write down the exact browsers we support in our docs, and set the `browserslist` according to that

packages/clerk-js/rspack.config.js
dstaley: Previously Prettier kept everything on one line since it was below our max line length threshold. Since the line is longer (due to the inclusion of the `{ targets: packageJSON.browserslist }` argument), it splits the arguments to `merge` into separate lines. This confuses the GitHub diff formatter into thinking the lines have changed due to how the final argument to `merge` is now indented. So this is Prettier working as expected! I just wanted to make sure folks weren't struggling to find the difference between the two :)

dstaley:
> I'd argue that with this change we should also change our existing browserslist entry and remove the legacy information (Safari > 12, iOS > 12)

I _absolutely_ agree with this, but it might break other customers who were relying on the transformations that SWC was applying to fit those targets. We're going to be discussing this more during the SDK Infra meeting this week to determine if we're okay accepting that. The current plan is to opt instances into the legacy bundle via CF worker, so we might be hesitant to suddenly break previously working browsers during a minor update.

packages/clerk-js/rspack.config.js
```diff
@@ -65,7 +68,7 @@ const common = ({ mode, disableRHC = false }) => {
         }),
     ].filter(Boolean),
     output: {
-      chunkFilename: `[name]_[fullhash:6]_${packageJSON.version}.js`,
+      chunkFilename: `[name]_${variant}_[fullhash:6]_${packageJSON.version}.js`,
```
dstaley, line 71: This is so that we can distinguish which variant a chunk belongs to, mainly so we could (in the future) target specific variant chunks with specific bundlewatch rules.

packages/clerk-js/rspack.config.js
```diff
@@ -154,8 +157,8 @@ const svgLoader = () => {
   };
 };
 
-/** @type { () => (import('@rspack/core').RuleSetRule[]) } */
-const typescriptLoaderProd = () => {
+/** @type { (opts: { targets: string, useCoreJs?: boolean }) => (import('@rspack/core').RuleSetRule[]) } */
+const typescriptLoaderProd = ({ targets, useCoreJs = false }) => {
```
LekoArts, line 161: Nit:

What about using `packageJSON.browserslist` as a default value for `targets`? We probably (with our current knowledge/approach) want to change the targets for more variants, so that could clean up some repetition

LekoArts (commented):
Generally speaking LGTM, left one minor comment.

Did you test a current snapshot of the normal browser and legacy bundle to see that old and new behavior is OK?

It would also be good to start a docs PR (if not already done) to update our https://clerk.com/docs/upgrade-guides/long-term-support#library-compatibility page

packages/clerk-js/rspack.config.js
dstaley: ooo great idea!

packages/clerk-js/rspack.config.js
```diff
@@ -525,31 +584,31 @@ const devConfig = ({ mode, env }) => {
...
       isSandbox ? { entry: { sandbox: './sandbox/app.ts' } } : {},
-      common({ mode }),
+      common({ mode, variant: variants.clerkBrowser }),
       commonForDev(),
     ),
     // prettier-ignore
     [variants.clerkBrowserNoRHC]: merge(
       entryForVariant(variants.clerkBrowserNoRHC),
-      common({ mode, disableRHC: true }),
+      common({ mode, disableRHC: true, variant: variant.clerkBrowserNoRHC }),
```
jacekradko, line 600: Not sure if this is correct, but it seems like it would be undefined without this change
```suggestion
      common({ mode, disableRHC: true, variant: variants.clerkBrowserNoRHC }),
```

jacekradko (approved):
Let's give it a whirl!

PR #5507, opened 2025-04-02
chore(shared,vue,nextjs,astro,clerk-react): Update `useAuth` to handle pending session as signed-out

LauraBeatris:
## Description

Part of ORGS-621 

Allows to pass `treatPendingAsSignedOut` as an option to `useAuth` - refer to the following behavior when a session has a pending status:

```ts
// userId is undefined, since pending is treat as signed-out state (default behavior)
const { userId } = useAuth() 

// userId is defined, since pending is treat as signed-out state (opt-in behavior)
const { userId } = useAuth({ treatPendingAsSignedOut: true })
```

Globally changing the default: 
```tsx
<ClerkProvider treatPendingAsSignedOut={false} />
```

--- 

Example from Nuxt route middleware: 

```ts

export default defineNuxtRouteMiddleware((to) => {
  const { userId } = useAuth()

  // If the user has a pending session, they aren't allowed to access
  // the protected route and are redirected to the sign-in page
  // If the sign-in page renders `SignIn`, then it'll redirect to `/tasks` internal route on component mount
  if (!userId.value && isProtectedRoute(to)) {
    return navigateTo('/sign-in')
  }
})
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/modern-paws-mate.md
+++ b/.changeset/modern-paws-mate.md
@@ -6,4 +6,4 @@
 '@clerk/vue': patch
 ---
 
-Update `useAuth` to handle pending sessions as signed-out by default, with opt-out via `treatPendingAsSignedOut: false`
+Update `useAuth` to handle pending sessions as signed-out by default, with opt-out via `useAuth({ treatPendingAsSignedOut: false })` or `<ClerkProvider treatPendingAsSignedOut={false} />`
--- a/.changeset/modern-paws-mate.md
+++ b/.changeset/modern-paws-mate.md
@@ -1,9 +1,7 @@
 ---
 '@clerk/nextjs': patch
 '@clerk/shared': patch
-'@clerk/astro': patch
 '@clerk/clerk-react': patch
-'@clerk/vue': patch
 ---
 
 Update `useAuth` to handle pending sessions as signed-out by default, with opt-out via `useAuth({ treatPendingAsSignedOut: false })` or `<ClerkProvider treatPendingAsSignedOut={false} />`
--- a/.changeset/modern-paws-mate.md
+++ b/.changeset/modern-paws-mate.md
@@ -1,0 +1,9 @@
+---
+'@clerk/nextjs': patch
+'@clerk/shared': patch
+'@clerk/astro': patch
+'@clerk/clerk-react': patch
+'@clerk/vue': patch
+---
+
+Update `useAuth` to handle pending sessions as signed-out by default, with opt-out via `treatPendingAsSignedOut: false`
--- a/packages/react/src/hooks/useAuth.ts
+++ b/packages/react/src/hooks/useAuth.ts
@@ -173,7 +173,7 @@ export function useDerivedAuth(authObject: any, options: PendingSessionOptions =
       has: derivedHas,
     },
     options: {
-      treatPendingAsSignedOut: treatPendingAsSignedOut ?? options.treatPendingAsSignedOut,
+      treatPendingAsSignedOut: treatPendingAsSignedOut ?? options.treatPendingAsSignedOut ?? true,
     },
   });
 
--- a/packages/react/src/hooks/useAuth.ts
+++ b/packages/react/src/hooks/useAuth.ts
@@ -1,5 +1,4 @@
 import { createCheckAuthorization, resolveAuthState } from '@clerk/shared/authorization';
-import { useOptionsContext } from '@clerk/shared/react';
 import type {
   CheckAuthorizationWithCustomPermissions,
   GetToken,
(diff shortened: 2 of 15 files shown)


packages/react/src/hooks/useAuth.ts
```diff
@@ -158,69 +165,21 @@ export function useDerivedAuth(authObject: any): UseAuthReturn {
...
-      orgRole,
-      orgSlug: orgSlug || null,
       has: derivedHas,
-      signOut,
-      getToken,
-    };
-  }
+    },
+    options: {
+      treatPendingAsSignedOut: treatPendingAsSignedOut ?? clerk?.__internal_getOption('treatPendingAsSignedOut'),
```
panteliselef, line 176: I believe doing `clerk?.__internal_getOption('treatPendingAsSignedOut')` on the server pass will always be undefined, instead we should read directly from the props passed to `ClerkProvider`

LauraBeatris: I've updated it to read from React context instead! 

For the Astro hook, I believe it's fine to read for `Clerk` if it's loaded on the client, but I'm also defaulting to `true` just in case it resolves to `undefined` on SSR - did the same thing for React as well

packages/react/src/hooks/useAuth.ts
```diff
@@ -127,20 +143,11 @@ export const useAuth = (initialAuthState: any = {}): UseAuthReturn => {
...
-    orgSlug,
-    orgRole,
-    has,
-    signOut,
-    getToken,
-    orgPermissions,
-    factorVerificationAge,
-  } = authObject ?? {};
+export function useDerivedAuth(authObject: any, options: PendingSessionOptions = {}): UseAuthReturn {
+  const optionsContext = useOptionsContext();
```
LauraBeatris, line 147: Starting to get this error: 

> <w> [webpack.cache.PackFileCacheStrategy] Serializing big strings (105kiB) impacts deserialization performance (consider using Buffer instead and decode when needed)

I think this is due to `useOptionsContext` being added here, currently investigating

packages/react/src/hooks/useAuth.ts
LauraBeatris: > we should read directly from the props passed to ClerkProvider

@panteliselef Do you know what's the recommended approach to read `ClerkProvider` options here? I've tried to use `useOptionsContext` but it's leading to the following error: 

> [webpack.cache.PackFileCacheStrategy] Serializing big strings (105kiB) impacts deserialization performance (consider using Buffer instead and decode when needed)

LauraBeatris:
A ton of build/tests failing here - my bad @panteliselef, I'll turn this into a draft for now, and fix tomorrow morning!

Thought stuff was passing but since it was stacked on top of another branch before, then CI didn't run.

LauraBeatris:
@panteliselef  + @wobsoriano helped me figure out the best approach to pass options between SSR / CSR contexts!

I've made some updates to `useAuth` across Vue, Astro and React, in which it performs the following checks for `treatPendingAsSignedOut`:
- Checks if it passed as parameter: `useAuth({ treatPendingAsSignedOut: false })`
- Checks if it's passed as clerk-js option: `<ClerkProvider treatPendingAsSignedOut />` - or via Vue plugin, Astro integration, etc 
- Defaults to SSR context - for Astro in this case, would be built as an environment variable accessible via `import.meta.env.PUBLIC_CLERK_TREAT_PENDING_AS_SIGNED_OUT`, and for Next.js, we're defaulting to `IsomorphicClerk` options while clerk-js loads

packages/react/src/hooks/useAuth.ts
LauraBeatris: https://github.com/clerk/javascript/pull/5507#issuecomment-2776497678

wobsoriano (approved):
Looks good on the Vue and Astro side. Thanks for abstracting the auth state logic away 🙇🏼

.changeset/modern-paws-mate.md
```diff
@@ -0,0 +1,9 @@
+---
+'@clerk/nextjs': patch
+'@clerk/shared': patch
+'@clerk/astro': patch
+'@clerk/clerk-react': patch
+'@clerk/vue': patch
+---
+
+Update `useAuth` to handle pending sessions as signed-out by default, with opt-out via `useAuth({ treatPendingAsSignedOut: false })` or `<ClerkProvider treatPendingAsSignedOut={false} />`
```
LauraBeatris, line 9: I'll spin up different changeset per package since `<ClerkProvider treatPendingAsSignedOut={false} />` is specific to React

PR #5509, opened 2025-04-02
chore(shared): Set a timeout when loading Clerk from an inline script.

panteliselef:
## Description

The inline script might fail before the event listeners are attached which causes an infinite pending state. Instead, we would like to retry fetching the script and/or throw an error.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/src/loadClerkJsScript.ts
+++ b/packages/shared/src/loadClerkJsScript.ts
@@ -32,6 +32,7 @@ type LoadClerkJsScriptOptions = Without<ClerkOptions, 'isSatellite'> & {
   proxyUrl?: string;
   domain?: string;
   nonce?: string;
+  existingScriptTimeout?: number;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


packages/shared/src/loadClerkJsScript.ts
```diff
@@ -50,16 +51,49 @@ type LoadClerkJsScriptOptions = Without<ClerkOptions, 'isSatellite'> & {
 const loadClerkJsScript = async (opts?: LoadClerkJsScriptOptions) => {
   const existingScript = document.querySelector<HTMLScriptElement>('script[data-clerk-js-script]');
 
+  const existingScriptTimeout = opts?.existingScriptTimeout ?? 2_000;
+  let timeoutId: ReturnType<typeof setTimeout>;
+
+  function clTimeout() {
```
nikosdouvlis, line 57: What does "cl" mean in this context?

packages/shared/src/loadClerkJsScript.ts
```diff
@@ -50,16 +51,49 @@ type LoadClerkJsScriptOptions = Without<ClerkOptions, 'isSatellite'> & {
 const loadClerkJsScript = async (opts?: LoadClerkJsScriptOptions) => {
   const existingScript = document.querySelector<HTMLScriptElement>('script[data-clerk-js-script]');
 
+  const existingScriptTimeout = opts?.existingScriptTimeout ?? 2_000;
```
nikosdouvlis, line 54: ❓ How did we land on 2000ms as the timeout here?

packages/shared/src/loadClerkJsScript.ts
```diff
@@ -50,16 +51,49 @@ type LoadClerkJsScriptOptions = Without<ClerkOptions, 'isSatellite'> & {
...
+        reject();
+      }, existingScriptTimeout);
     });
+
+    const loaded = await attemptExisting
+      .then(() => true)
+      .catch(msg => {
+        if (msg) {
+          // TODO: Align with `throw new Error(FAILED_TO_LOAD_ERROR)` in the next major
+          throw msg;
```
nikosdouvlis, line 89: Can you briefly explain how consumers (code-wise) handle this error?

packages/shared/src/loadClerkJsScript.ts
panteliselef: From a quick test, anything between 1500ms to 3000ms seemed appropriate. on Fast 4g, clerk-js hotloaded loads on approximately 1s.

The issue with this approach is that you can have false possitives, causing timeouts on slow connections. But we retry below.

An alternative solution here is to add this to our nextjs package.
```tsx
<script>
  {`
  function() {
    const s = document.createElement('script');
    window.__clerk_script_promise = new Promise(function (res,rej){
      script.addEventListener('load', function() {
        resolve(script);
      });

      script.addEventListener('error', function() {
        reject();
      });
    })
    script.setAttribute('crossorigin', 'anonymous');
    script.setAttribute('data-clerk-js-script', 'true');
    script.async = true;
    script.defer = false;

    script.src = "${scriptUrl}";
    script.setAttribute('data-clerk-publishable-key', "${publishableKey}")
    document.head.appendChild(script);
  }`}
</script>
```

But I didn't choose this approach because and inline script like this one, can be hard to maintain.

nikosdouvlis: The script above doesn't look that bad. I'm not a fan of the hardcoded-timeout approach as it might impact users - if I had to choose between c2 UX and maintainability, I'd vote for UX

panteliselef: I will opern a new pr with the script approach

PR #5515, opened 2025-04-02
fix(clerk-js): Use __internal_setActiveInProgress to prevent navigations

dstaley:
## Description

This PR introduces a new flag on the core Clerk class, `__internal_setActiveInProgress`, to indicate that a call to `setActive` is in flight. This can be used to guard against performing navigations from components that are remounted by Next.js during `setActive` as a result of our `invalidateCacheAction`. This fixes issues where routes like `/sso-callback` and `/sign-up/continue` were remounted during `setActive`, which trigger navigations that can sometimes resolve _before_ `setActive` is complete, leading the user to the incorrect location.

Additionally, this PR fixes an issue where the SSO callback URL was incorrectly set for the combined flow, which prevented users from being sent to `/sign-up/continue` when attempting to sign up.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/common/SSOCallback.tsx
```diff
@@ -17,16 +17,18 @@ export const SSOCallback = withCardStateProvider<HandleOAuthCallbackParams | Han
...
   const { navigate } = useRouter();
   const card = useCardState();
 
   React.useEffect(() => {
     let timeoutId: ReturnType<typeof setTimeout>;
-    handleRedirectCallback({ ...props }, navigate).catch(e => {
-      handleError(e, [], card.setError);
-      timeoutId = setTimeout(() => void navigate('../'), 4000);
-    });
+    if (__internal_setActiveInProgress !== true) {
```
brkalow, line 26: ❓ why don't we do this in our router instead?

packages/clerk-js/src/ui/common/redirects.ts
```diff
@@ -54,6 +54,11 @@ export function buildSSOCallbackURL(
   ctx: Partial<SignInContextType | SignUpContextType>,
   baseUrl: string | undefined = '',
 ): string {
+  // If the context contains an SSO callback URL, use it instead of building a new one, as it likely contains the
+  // combined flow path.
+  if ('ssoCallbackUrl' in ctx && ctx.ssoCallbackUrl) {
+    return ctx.ssoCallbackUrl;
+  }
```
brkalow, line 61: 👍 good catch

brkalow (approved):
Approving if we want to get this out to fix the immediate issues, but I would like to see this integrated into our internal router instead. Feels like this is something that will likely crop up again if we don't.

Nice idea.

packages/clerk-js/src/ui/common/SSOCallback.tsx
dstaley: We can definitely explore something like that, which I feel is more intelligent than this blunt approach. It'd likely use the same flag too, so we're good there. The main reason I went with this instead of scoping this purely to preventing navigations is I'm concerned we have more functionality that needs to be prevented from executing a second time. For example, the PR as is prevents the execution of `handleRedirectCallback`, rather than simply preventing a navigation from it. `handleRedirectCallback` has several non-navigation actions that it performs that it would be best if they were prevented completely (stuff like `signIn.reload`, `signIn.create`, `signUp.create`, `setActive`, etc).

PR #5517, opened 2025-04-02
feat(backend, clerk-js,clerk-react,shared,types): Add feature param to has

iagodahlem:
## Description

Add `__experimental_feature` param to `has` function:

```ts
const { has } = await auth()

const hasFeature = has({ __experimental_feature: 'testfeature' })
```

Resolves [COM-203](https://linear.app/clerk/issue/COM-203/sdk-support-features-in-has).

TODO: 
- testing
- fill in features claims in `Session` and `useAuth`
- improving checking roles, permissions and reverifications with feature.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/types/src/jwtv2.ts
+++ b/packages/types/src/jwtv2.ts
@@ -116,15 +116,15 @@ type JWTPayloadBase = {
   sts?: SessionStatusClaim;
 
   /**
-   * Feature claims this represent features, you can have multiple features in the same claim. e.g.: "impersonation,emails"
-   * @experimental
-   */
-  fea?: string | null;
-
-  /**
    * Any other JWT Claim Set member.
    */
   [propName: string]: unknown;
+
+  /**
+   * Feature claims, you can have multiple features in the same claim. e.g.: "impersonation,emails"
+   * @experimental
+   */
+  fea?: string | null;
 };
 
 export type VersionedJwtPayload =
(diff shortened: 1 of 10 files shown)


packages/types/src/jwtv2.ts
```diff
@@ -114,6 +114,12 @@ export interface JwtPayload extends CustomJwtSessionClaims {
    */
   sts?: SessionStatusClaim;
 
+  /**
+   * Feature claims this represent features, you can have multiple features in the same claim. e.g.: "impersonation,emails"
+   * @experimental
+   */
```
panteliselef, line 120: How come we change our approach here and did not stick with `string[]` similar to `org_permissions` ?

iagodahlem: This has been a long discussed topic, but here are some details and questions we came across: https://www.notion.so/clerkdev/Commerce-Features-Permissions-JWT-implications-1a52b9ab44fe80a896ccf9feedd788ed?pvs=4

@izaaklauer and @mzhong9723 had this idea for a new claims format with some inputs from Colin.

packages/types/src/jwtv2.ts
```diff
@@ -114,6 +114,12 @@ export interface JwtPayload extends CustomJwtSessionClaims {
...
+  fea?: string | null;
```
octoper, line 121: I think we will first need to have JWT versioning before proceeding with this addition

iagodahlem: I agree, I'll move this one to draft, and we can update it later with the versioning changes.

PR #5518, opened 2025-04-02
fix(clerk-js): Only refresh when auth with popup is called

dstaley:
## Description

This PR uses `sessionStorage` to set a flag indicating that `authenticateWithPopup` was used to initiate an SSO connection, which likely means that the `signIn` and `signUp` resources need to be reloaded before handling the redirect.

This reduces the instances of a `405 Method Not Allowed` response from FAPI which is returned when we attempt to refresh a resource that we haven't acted upon. For example, if we're attempting to sign in, calling `signUp.reload()` will cause an HTTP 405 error. By only performing this reload when `authenticateWithPopup` has been called, we reduce the instance of 405 errors in our logs and within customer applications.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -112,7 +112,13 @@ import { memoizeListenerCallback } from '../utils/memoizeStateListenerCallback';
 import { RedirectUrls } from '../utils/redirectUrls';
 import { AuthCookieService } from './auth/AuthCookieService';
 import { CaptchaHeartbeat } from './auth/CaptchaHeartbeat';
-import { CLERK_SATELLITE_URL, CLERK_SUFFIXED_COOKIES, CLERK_SYNCED, ERROR_CODES } from './constants';
+import {
+  CLERK_SATELLITE_URL,
+  CLERK_SUFFIXED_COOKIES,
+  CLERK_SYNCED,
+  ERROR_CODES,
+  SESSION_STORAGE_AUTH_WITH_POPUP_KEY,
+} from './constants';
 import {
   clerkErrorInitFailed,
   clerkInvalidSignInUrlFormat,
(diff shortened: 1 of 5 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -1507,11 +1513,23 @@ export class Clerk implements ClerkInterface {
...
+    } catch {
+      // In the event that sessionStorage is disabled, assume the resource needs to be refreshed. Refreshing when not
+      // needed doesn't break anything, but it does cause 405 network errors.
+      shouldRefreshResources = true;
+    }
+
     // If `handleRedirectCallback` is called on a window without an opener property (such as when the OAuth flow popup
     // directs the opening page to navigate to the /sso-callback route), we need to reload the signIn and signUp resources
     // to ensure that we have the latest state. This operation can fail when we try reloading a resource that doesn't
     // exist (such as when reloading a signIn resource during a signUp attempt), but this can be safely ignored.
```
brkalow, line 1531: Could this entire thing be done via a `postMessage()` from the popup? The issue is we don't know which resource we need to fetch so we're just doing both, right?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1507,11 +1513,23 @@ export class Clerk implements ClerkInterface {
...
-    if (!window.opener) {
+    if (!window.opener && shouldRefreshResources) {
       try {
         await signIn.reload();
```
panteliselef, line 1534: Can you remind me why doing `.reload()` is useful, if it always returns a 405 ?

dstaley: It doesn't always return a 405. It only returns a 405 when the underlying resource doesn't exist. So for example if we created a SignIn resource during the popup, `signIn.reload()` will pick that up, but `signUp.reload()` will return a 405.

panteliselef: oh i see, thanks for clarifying.

dstaley:
closing in favor of #5553

PR #5407, opened 2025-03-20
feat: add redirectToSignUp() to ClerkMiddlewareAuthObject

sambarnes:
## Checklist

- [ ] `pnpm test` runs as expected.
- [ ] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #5533, opened 2025-04-04
feat(nextjs): Support `auth().redirectToSignUp()`

panteliselef:
## Description

[Original PR](https://github.com/clerk/javascript/pull/5407). 

This PR simply removes duplicated code and improves the changeset example.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/backend/src/createRedirect.ts
```diff
@@ -97,7 +97,18 @@ export const createRedirect: CreateRedirect = params => {
     }
 
     const accountsSignUpUrl = `${accountsBaseUrl}/sign-up`;
-    const targetUrl = signUpUrl || accountsSignUpUrl;
+
+    // Allows redirection to SignInOrUp path
+    function buildSignUpUrl(signIn: string | URL | undefined) {
```
LauraBeatris, line 102: Just want to make sure that I understand this part here. If there's a sign in URL option (env var or middleware key), we assume that combined flow is enabled - but doesn't it also need be able via the `withSignUp` prop?

panteliselef: yes, but in that case signUpUrl will be defined. `withSignUp` is per component, nothing in can do about it on the server.

anagstef (approved):
LGTM

PR #5553, opened 2025-04-07
fix(clerk-js,types): Use intent parameter to reload resource

dstaley:
## Description

This PR introduces a new URL parameter, `intent`, used when performing SSO login via popup to indicate which resource needs to be refreshed before handling the redirect callback. The `intent` parameter is appended to the `redirectUrl` passed to FAPI such that the popup flow redirects to the `redirectUrl` with the `intent` parameter set to the resource that needs to be reloaded. This is passed to a new `reloadResource` optional parameter on the `handleRedirectCallback` method to indicate which resource to reload.

This reduces the instances of a `405 Method Not Allowed` response from FAPI which is returned when we attempt to refresh a resource that we haven't acted upon. For example, if we're attempting to sign in, calling `signUp.reload()` will cause an HTTP 405 error. By only performing this reload on the `signIn` resource, we reduce the instance of 405 errors in our logs and within customer applications.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/common/SSOCallback.tsx
```diff
@@ -24,7 +24,9 @@ export const SSOCallbackCard = (props: HandleOAuthCallbackParams | HandleSamlCal
   React.useEffect(() => {
     let timeoutId: ReturnType<typeof setTimeout>;
     if (__internal_setActiveInProgress !== true) {
-      handleRedirectCallback({ ...props }, navigate).catch(e => {
+      const intent = new URLSearchParams(window.location.search).get('intent');
```
panteliselef, line 27: Should we maybe use the params from the router instead ?

packages/clerk-js/src/ui/common/SSOCallback.tsx
```diff
@@ -24,7 +24,9 @@ export const SSOCallbackCard = (props: HandleOAuthCallbackParams | HandleSamlCal
...
+      const reloadResource = intent === 'signIn' || intent === 'signUp' ? intent : undefined;
```
panteliselef, line 28: Can we do this ? Or is `reloadResource:false` not proper ?
```suggestion
      const reloadResource = ['signIn', 'signUp'].includes(intent)
```

panteliselef (approved):

packages/clerk-js/src/ui/common/SSOCallback.tsx
dstaley: In the hash and virtual routers, the queryString isn't derived from `window.location.search`, which is where the `intent` param is going to be. For example, this is the SSO callback route when using the hash router:

```
http://localhost:4000/sign-in?intent=sign-in#/sso-callback
```

In this example, `window.location.search` is `?intent=sign-in`, but the value of `queryString` from `useRotuer` is `""`.

Alternatively, we could append the `intent` parameter to different locations based on the router used, but I felt that was a little too much additional complexity in exchange for basically no additional functionality.

packages/clerk-js/src/ui/common/SSOCallback.tsx
dstaley: TypeScript won't narrow from `string` to `'signIn' | 'signUp'` if we use `Array.includes`.

brkalow (approved):

PR #5597, opened 2025-04-11
fix(clerk-js,chrome-extension): Stripe bundles

tmilewski:
## Description

Fixes ECO-595

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/open-cases-happen.md
+++ b/.changeset/open-cases-happen.md
@@ -1,0 +1,5 @@
+---
+'@clerk/chrome-extension': patch
+---
+
+Externalize all Stripe-related modules
--- a/.changeset/slick-taxis-cut.md
+++ b/.changeset/slick-taxis-cut.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Fully strip @stripe/stripe-js from non-RHC builds
--- a/packages/chrome-extension/tsup.config.ts
+++ b/packages/chrome-extension/tsup.config.ts
@@ -17,7 +17,7 @@ export default defineConfig(overrideOptions => {
     legacyOutput: true,
     treeshake: true,
     noExternal: ['@clerk/clerk-react', '@clerk/shared'],
-    external: ['use-sync-external-store'],
+    external: ['use-sync-external-store', '@stripe/stripe-js', '@stripe/react-stripe-js'],
     define: {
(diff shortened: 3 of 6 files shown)


tmilewski:
!snapshot

tmilewski:
!snapshot

alexcarpenter (approved):

wong2:
I'm getting this error with `@clerk/chrome-extension`:

```
Uncaught Error: Cannot find module '@stripe/stripe-js'
```

Any suggestions?

tmilewski:
Hi @wong2, please note that Clerk Billing is not available for Chrome Extensions due to limitations imposed by Chrome. Stripe requires remotely hosted code to function, which, for security purposes, causes Google to deny Chrome Store approvals. 

Could you confirm for me that you’re not attempting to use Billing components?

Thank you!

wong2:
@tmilewski Oh, I'm trying to use Clerk Billing in my extension, I'm using the PricingTable component.

Perhaps you should add a notice in the Chrome Extension docs? The `PricingTable` component is documented and exported from the package, so I thought it's ok to use it.

image

tmilewski:
@wong2 I've added a PR to better catch this case in code here: https://github.com/clerk/javascript/pull/6179 

I'll see what we can do about the documentation on two fronts:

1. Better indicate that the Clerk UI components (and related) sections of the sidebar are global and not framework-specific.
2. Add notices for the unsupported components.

wong2:
@tmilewski Thanks!

Does this mean that the only way to use Billing in a Chrome extension is to enable syncHost and jump from the extension to the website to display the PricingTable? That way, once the payment is completed, the extension should be able to sync to the user's plan status, right?

tmilewski:
@wong2 The answer there is "it depends." If you use a typical popup extension (or other type that you re-open) then, yes, everything should work as expected. If you're using an extension like a sidePanel, and it's persistent, then I would suggest trying out `__experimental_syncHostListener`. I haven't explicitly tested it against billing yet, but I _believe_ it should work.

wong2:
@tmilewski I will give it a try! My use case is the extension opens a persistent page (`chrome-extension://<extension-id>/app.html`).

PR #5615, opened 2025-04-14
chore(e2e): Run existing protect tests with a JWT v2 instance

panteliselef:
## Description

Spins up a new long running application configure to use the new jwt (v2) formatting agains the existing e2e tests around `protect()`

Fixes COM-440

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/testing/src/common/types.ts
```diff
@@ -29,6 +29,13 @@ export type ClerkSetupOptions = {
    */
   secretKey?: string;
 
+  /*
+   * The backend API URL
+   * If not provided, the library will look for the key in the following environment variables:
+   * - CLERK_API_URL
+   */
+  apiUrl?: string;
+
```
anagstef, line 38: Can we not expose this publicly somehow? Because all users of the package should use the `api.clerk.com` 🤷 

Or, at least, mark it as `__internal_XXX`.

anagstef (approved):

PR #5468, opened 2025-03-27
feat(backend,react-router,nuxt,astro,nextjs,tanstack-react-start,express,fastify): Add verify webhook function

wobsoriano:
## Description

Right now, customers have to implement webhook verification when receiving Clerk webhooks to sync data with their apps using [this guide](https://clerk.com/docs/webhooks/sync-data#create-the-endpoint). This PR abstracts away the majority of that logic, and makes it built-in across our backend and full-stack SDKs.

The verification can now be done with a single function call:

```ts
import { verifyWebhook } from '@clerk/backend/webhooks';
// Or from your framework-specific package:
// import { verifyWebhook } from '@clerk/nextjs/webhooks';

try {
  const evt = await verifyWebhook(request);
  if (evt.type === 'user.created') {
    // Handle user creation
    console.log('New user created:', evt.data.id);
  }
} catch (err) {
  // Invalid webhook signature
  return new Response('Webhook verification failed', { status: 400 });
}
```

Most of our SDKs provide a web standard `Request` object that can be passed directly to the `verifyWebhook` function. For Express and Fastify, we've added framework-specific support so that users can pass their route request object directly.

We also marked `svix` as an optional peer dependency for users not using webhooks.

Did some local tests to the ff. SDKs:

- ✅  Express
- ✅ Fastify
- ✅ Nuxt
- ✅ Nextjs

Resolves ECO-533

## Checklist

- [x] `pnpm test` runs as expected.
- [x] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [x] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #5618, opened 2025-04-14
chore(nextjs): Add support for pages router when verifying webhooks

wobsoriano:
## Description

Recently, we added a way to [simplify webhook verification](https://github.com/clerk/javascript/pull/5468) across our SDKs. While this works for all SDKs that provides a standard web `Request`, there are still users who use Next.js Pages Router or frameworks like tRPC that use Node's `IncomingMessage` instead of web `Request`. This PR adds support for webhook verification with these frameworks by accepting a [RequestLike](https://github.com/clerk/javascript/blob/bae9741943248025f5245b7d6a7b842256f2cf4a/packages/nextjs/src/server/types.ts#L9) request, accepting both standard web `Request` and Node Request.

Resolves ECO-571

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/nextjs/src/webhooks.ts
```diff
@@ -1 +1,81 @@
...
+ *     console.error('Webhook verification failed:', err);
+ *     return new Response('Webhook verification failed', { status: 400 });
+ *   }
+ * }
+ * ```
+ */
+export async function verifyWebhook(request: RequestLike, options?: VerifyWebhookOptions) {
+  if (isNextRequest(request) || isRequestWebAPI(request)) {
+    return verifyWebhookBase(request, options);
+  }
```
wobsoriano, line 53: For standard `Request`, use the base `verifyWebhook`.

packages/nextjs/src/webhooks.ts
```diff
@@ -1 +1,81 @@
...
+
+  const webRequest = nextApiRequestToWebRequest(request);
+  return verifyWebhookBase(webRequest, options);
+}
```
wobsoriano, line 57: For standard web `Request`, it uses the default `verifyWebhook` provided by the backend SDK. For `NextApiRequest` type, we're converting it to a `Request` before verifying

tmilewski (approved):
🚀

PR #5645, opened 2025-04-16
fix(clerk-js): Disambiguate redirect url options between flows

brkalow:
## Description

Fixes an issue where `fallbackRedirectUrl` and `forceRedirectUrl` passed to a specific flow would be improperly passed to the other flow in a modal case.

Example:
```
Clerk.openSignUp({ forceRedirectUrl: '/?form=signup' }) // would apply to sign ins if user clicks sign in in the footer
```

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/contexts/components/SignUp.ts
+++ b/packages/clerk-js/src/ui/contexts/components/SignUp.ts
@@ -17,7 +17,7 @@ import { useRouter } from '../../router';
 import type { SignUpCtx } from '../../types';
 import { getInitialValuesFromQueryParams } from '../utils';
 
-export type SignUpContextType = SignUpCtx & {
+export type SignUpContextType = Omit<SignUpCtx, 'fallbackRedirectUrl' | 'forceRedirectUrl'> & {
   navigateAfterSignUp: () => any;
   queryParams: ParsedQueryString;
   signInUrl: string;
@@ -71,6 +71,9 @@ export const useSignUpContext = (): SignUpContextType => {
     queryParams,
     mode,
   );
+
+  delete ctx.fallbackRedirectUrl;
+  delete ctx.forceRedirectUrl;
 
   const afterSignUpUrl = clerk.buildUrlWithAuth(redirectUrls.getAfterSignUpUrl());
   const afterSignInUrl = clerk.buildUrlWithAuth(redirectUrls.getAfterSignInUrl());
--- a/packages/clerk-js/src/ui/contexts/components/SignUp.ts
+++ b/packages/clerk-js/src/ui/contexts/components/SignUp.ts
@@ -65,8 +65,8 @@ export const useSignUpContext = (): SignUpContextType => {
     options,
     {
       ...ctx,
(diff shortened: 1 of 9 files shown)


packages/clerk-js/src/ui/contexts/components/SignUp.ts
```diff
@@ -65,8 +65,8 @@ export const useSignUpContext = (): SignUpContextType => {
     options,
     {
       ...ctx,
```
dstaley, line 67: doesn't this mean that `SignInContext` and `SignUpContext` still contain a `forceRedirectUrl` prop? Can we completely remove any and all usage of `forceRedirectUrl` except at the component boundary?

packages/clerk-js/src/ui/types.ts
```diff
@@ -50,7 +54,8 @@ type ComponentMode = 'modal' | 'mounted';
 export type SignInCtx = SignInProps & {
```
dstaley, line 54: should we update `SignInProps` to accept `SignInForceRedirectUrl` instead of only adding it to the context type?

packages/clerk-js/src/utils/disambiguateRedirectOptions.ts
```diff
@@ -0,0 +1,43 @@
...
+type PrefixedRedirectOptions<F extends Flow> = {
+  [K in keyof RedirectOptions as `${F extends 'signin' ? 'signIn' : 'signUp'}${Capitalize<K>}`]: RedirectOptions[K];
+};
+
+type DisambiguatedRedirectOptions<T extends RedirectOptions, F extends Flow> = {
+  [K in keyof T as K extends 'forceRedirectUrl' | 'fallbackRedirectUrl' ? never : K]: T[K];
+} & PrefixedRedirectOptions<F>;
+
+/**
+ * Ensures redirect props have the appropriate prefix depending on the flow they were originally provided to. Used when passing props from sign up -> sign in, or vice versa
```
dstaley, line 17: should we also note this is only needed for the modals?

packages/clerk-js/src/ui/types.ts
brkalow: My intent was to keep the prop types unchanged and handle it internally, hence only applying it to the context type here.

dstaley:
would you mind adding a test for the "render sign in modal, click sign up, sign up with existing account" flow with the props specified?

packages/clerk-js/src/ui/contexts/components/SignUp.ts
brkalow: I'll remove from the context, as far as I can tell `forceRedirectUrl` isn't actually used outside of this assignment in the components, so we _should_ be safe.

packages/clerk-js/src/ui/types.ts
dstaley: ahhh right right

dstaley (approved):

PR #5659, opened 2025-04-17
chore(e2e): Add e2e tests for has() with JWT v2 format

panteliselef:
## Description

Adds tests for the new JWT v2 format. And tests the output of `has()` for 3 roles across server and client. 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


integration/tests/protect-jwt-v2.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    await u.po.signIn.signInWithEmailAndInstantPassword({ email: fakeAdmin.email, password: fakeAdmin.password });
+    await u.po.expect.toBeSignedIn();
+
+    await u.po.organizationSwitcher.goTo();
+    await u.po.organizationSwitcher.waitForMounted();
+    await u.po.organizationSwitcher.waitForAnOrganizationToSelected();
+
+    await u.page.goToRelative('/jwt-v2-organizations');
+
+    await expect(u.page.getByTestId('client-jwt').getByText(`"per": "create,delete,manage,read"`)).toBeVisible();
```
dstaley, line 61: Instead of checking that a given substring is present on the page, did you consider getting the `textContent` of the selector and running that through `JSON.parse` so you can directly compare values?

integration/tests/protect-jwt-v2.test.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    await expect(
+      u.page.getByTestId('client-jwt').getByText(`"fea": "o:billing,o:impersonation,o:magic_links,o:posts"`),
+    ).toBeVisible();
+    await expect(
+      u.page.getByTestId('server-jwt').getByText(`"fea": "o:billing,o:impersonation,o:magic_links,o:posts"`),
+    ).toBeVisible();
+
+    async function assertPermsRolesFeatures() {
+      await expect(u.page.getByText(`has({ permission: "org:impersonation:read" }) -> true`)).toBeVisible();
```
dstaley, line 75: like the above comment, you could probably emit these as JSON keys (`"hasPermissionImpersonationRead": true`) to make it a bit easier to expect (since you could directly compare an entire object in one call)

dstaley (approved):
lgtm overall! I do think you could probably simplify the actual test by writing JSON to the DOM and then parsing it via `textContent`.

PR #5695, opened 2025-04-22
fix(clerk-react): Add details to `UserButton.*` error message

tmilewski:
## Description

There are error messages which pop-up when using `<UserButton.*>` stating that only specific components may be used as children. This error will also pop-up when the component tree is correct but is contained within a server component.

This PR is a quick-fix to add client component details to the error message in an effort to unblock developers who may find themselves in this state.

Fixes ECO-617

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/react/src/errors/messages.ts
+++ b/packages/react/src/errors/messages.ts
@@ -42,9 +42,6 @@ export const incompatibleRoutingWithPathProvidedError = (componentName: string) 
 
 export const userButtonIgnoredComponent = `<UserButton /> can only accept <UserButton.UserProfilePage />, <UserButton.UserProfileLink /> and <UserButton.MenuItems /> as its children. Any other provided component will be ignored. Additionally, please ensure that the component is rendered in a client component.`;
 
-export const mustBeContainedInClientComponentError = (componentName: string) =>
-  `The <${componentName}/> component must be used in a client component.`;
-
 export const customMenuItemsIgnoredComponent =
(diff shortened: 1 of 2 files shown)


packages/react/src/errors/messages.ts
```diff
@@ -40,10 +40,13 @@ export const noPathProvidedError = (componentName: string) =>
 export const incompatibleRoutingWithPathProvidedError = (componentName: string) =>
   `The \`path\` prop will only be respected when the Clerk component uses path-based routing. To resolve this error, pass \`routing='path'\` to the <${componentName}/> component, or drop the \`path\` prop to switch to hash-based routing. For more details please refer to our docs: https://clerk.com/docs`;
 
-export const userButtonIgnoredComponent = `<UserButton /> can only accept <UserButton.UserProfilePage />, <UserButton.UserProfileLink /> and <UserButton.MenuItems /> as its children. Any other provided component will be ignored.`;
+export const userButtonIgnoredComponent = `<UserButton /> can only accept <UserButton.UserProfilePage />, <UserButton.UserProfileLink /> and <UserButton.MenuItems /> as its children. Any other provided component will be ignored. Additionally, please ensure that the component is rendered in a client component.`;
+
+export const mustBeContainedInClientComponentError = (componentName: string) =>
```
panteliselef, line 45: Is this something new ? Where is it used ?

panteliselef (approved):

packages/react/src/errors/messages.ts
tmilewski: Removed!

PR #5746, opened 2025-04-28
feat(clerk-js): Display proration in `<Checkout />`

panteliselef:
## Description

image

image

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/elements/LineItems.tsx
```diff
@@ -48,17 +48,18 @@ interface GroupProps {
...
+function Group({ children, borderTop = false, variant = 'primary', expand = false }: GroupProps) {
   return (
     <GroupContext.Provider value={{ variant }}>
       <Box
         elementDescriptor={descriptors.lineItemsGroup}
         elementId={descriptors.lineItemsGroup.setId(variant)}
         sx={t => ({
           display: 'grid',
-          gridTemplateColumns: 'repeat(2, minmax(0, 1fr))',
+          gridTemplateColumns: `repeat(${expand ? 1 : 2}, minmax(0, 1fr))`,
```
alexcarpenter, line 62: Not super keen on the prop `expand`. Feels like title and description should have an sx prop that allows you to span multiple columns as needed. This makes it more flexible for responsive situations.

packages/clerk-js/src/ui/components/Checkout/CheckoutForm.tsx
```diff
@@ -40,14 +43,33 @@ export const CheckoutForm = ({
         })}
       >
         <LineItems.Root>
-          <LineItems.Group>
+          {showDowngradeInfo && (
+            <LineItems.Group expand>
+              <LineItems.Title description={'Your features will remain the end of your current subscription.'} />
+            </LineItems.Group>
+          )}
```
alexcarpenter, line 50: This feels more like a notice verse a line item. Especially when there is no accompanying description.

packages/clerk-js/src/ui/elements/LineItems.tsx
```diff
@@ -80,7 +81,7 @@ function Group({ children, borderTop = false, variant = 'primary' }: GroupProps)
  * -----------------------------------------------------------------------------------------------*/
 
 interface TitleProps {
-  title: string | LocalizationKey;
+  title?: string | LocalizationKey;
```
alexcarpenter, line 84: title on the title component should not be optional. This is similar to the not about this update not being a line item, and more of a notice that gets displayed above the line items.

packages/clerk-js/src/ui/elements/LineItems.tsx
```diff
@@ -162,7 +163,6 @@ function Description({ text, prefix, suffix, truncateText = false, copyText = fa
         sx={t => ({
           display: 'inline-flex',
           justifyContent: 'flex-end',
-          alignItems: 'center',
```
alexcarpenter, line 165: why'd we remove this?

panteliselef: to get this 
image
 and not this
image

packages/clerk-js/src/ui/components/Checkout/CheckoutForm.tsx
```diff
@@ -40,14 +43,33 @@ export const CheckoutForm = ({
         })}
       >
         <LineItems.Root>
-          <LineItems.Group>
+          {showDowngradeInfo && (
+            <LineItems.Group expand>
+              <LineItems.Title description={'Your features will remain the end of your current subscription.'} />
```
aeliox, line 48: I wonder if we should fetch the active subscription and grab the plan name and period end date to make this line more informative – eg `Your Silver plan features will remain until April 1st, 2025.`

packages/clerk-js/src/ui/components/Checkout/CheckoutForm.tsx
```diff
@@ -40,14 +43,33 @@ export const CheckoutForm = ({
...
+            </LineItems.Group>
+          )}
+
+          <LineItems.Group borderTop={showDowngradeInfo}>
             <LineItems.Title title={plan.name} />
```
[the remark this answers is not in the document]
aeliox: It'd be great to show this in the Total Due [Starting date] with the amount they _will_ be charged. But again, i think the checkout create payload will need to update to reflect that.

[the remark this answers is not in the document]
aeliox: This makes a good case for the difference between `grandTotal` and `totalDueNow`?

packages/clerk-js/src/ui/components/Checkout/CheckoutForm.tsx
panteliselef: FAPI changes are tracked [here](https://linear.app/clerk/issue/COM-655/add-new-info-to-fapi-responses-for-checkout-component)

aeliox (approved):

PR #5756, opened 2025-04-29
chore(backend,nextjs): Introduce `treatPendingAsSignedOut` option to server-side utilities and control components

LauraBeatris:
## Description

Allow passing `treatPendingAsSignedOut` to `auth`, `getAuth`, and server-side control components. 

If session status is `pending`:
```ts
// Returns signed out object by default
const { userId } = auth()
```
```ts
// Treats as signed-in
const { userId } = auth({ treatPendingAsSignedOut: false })
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/nextjs/src/server/clerkMiddleware.ts
+++ b/packages/nextjs/src/server/clerkMiddleware.ts
@@ -2,6 +2,7 @@ import type { AuthObject, ClerkClient } from '@clerk/backend';
 import type { AuthenticateRequestOptions, ClerkRequest, RedirectFun, RequestState } from '@clerk/backend/internal';
 import { AuthStatus, constants, createClerkRequest, createRedirect } from '@clerk/backend/internal';
 import { parsePublishableKey } from '@clerk/shared/keys';
+import type { PendingSessionOptions } from '@clerk/types';
 import { notFound as nextjsNotFound } from 'next/navigation';
 import type { NextMiddleware, NextRequest } from 'next/server';
 import { NextResponse } from 'next/server';
@@ -41,7 +42,7 @@ export type ClerkMiddlewareAuthObject = AuthObject & {
 };
 
 export interface ClerkMiddlewareAuth {
-  (): Promise<ClerkMiddlewareAuthObject>;
+  (opts?: PendingSessionOptions): Promise<ClerkMiddlewareAuthObject>;
 
   protect: AuthProtect;
 }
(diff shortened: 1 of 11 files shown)


packages/backend/src/tokens/authStatus.ts
```diff
@@ -27,7 +27,11 @@ export type SignedInState = {
   afterSignInUrl: string;
   afterSignUpUrl: string;
   isSignedIn: true;
-  toAuth: () => SignedInAuthObject;
+  toAuth: <TOptions extends PendingSessionOptions>(
+    opts?: TOptions,
+  ) => TOptions extends { treatPendingAsSignedOut: false }
```
LauraBeatris, line 32: This generic makes it so that the return type of standlonone `requestState.toAuth` leads to `SignedInObject` in order to not break other modules. 

The same behavior happens in runtime, since the return type is just a union upon `treatPendingAsSignedOut: false` -> if the session exists, but has a pending status, than we override it as a signed out.

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -182,11 +183,14 @@ export const clerkMiddleware = ((...args: unknown[]): NextMiddleware | NextMiddl
       const redirectToSignUp = createMiddlewareRedirectToSignUp(clerkRequest);
       const protect = await createMiddlewareProtect(clerkRequest, authObject, redirectToSignIn);
 
-      const authObjWithMethods: ClerkMiddlewareAuthObject = Object.assign(authObject, {
-        redirectToSignIn,
-        redirectToSignUp,
-      });
-      const authHandler = () => Promise.resolve(authObjWithMethods);
+      const authHandler = (opts?: PendingSessionOptions) => {
```
LauraBeatris, line 186:
```suggestion
      const authHandler = (options?: PendingSessionOptions) => {
```

packages/nextjs/src/server/data/getAuthDataFromRequest.ts
```diff
@@ -53,5 +58,9 @@ export function getAuthDataFromRequest(
     authObject = signedInAuthObject(options, jwt.raw.text, jwt.payload);
   }
 
+  if (treatPendingAsSignedOut && authObject.sessionStatus === 'pending') {
+    authObject = signedOutAuthObject(options, authObject.sessionStatus);
+  }
```
brkalow, line 63: Same comment as the Astro PR, I would like to see this come from `@clerk/backend`:

```ts
getAuthObjectFromJwt(jwt)
```

brkalow (approved):

packages/nextjs/src/server/data/getAuthDataFromRequest.ts
LauraBeatris: Agreed! I'll raise a follow-up PR for it.

PR #5757, opened 2025-04-29
chore(astro): Introduce `treatPendingAsSignedOut` option to server-side utilities

LauraBeatris:
## Description

Allow passing `treatPendingAsSignedOut` to `auth`, `getAuth`, and server-side control components.

```ts
const { userId } = getAuth(req, { treatPendingAsSignedOut: false })
```

```ts
clerkMiddleware((auth, context) => {
  const { redirectToSignIn, userId } = auth({ treatPendingAsSignedOut: false })

  // Both `active` and `pending` sessions will be treated as authenticated when `treatPendingAsSignedOut` is false
  if (!userId && isProtectedRoute(context.request)) {
    return redirectToSignIn()
  }
})
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/astro/src/server/get-auth.ts
+++ b/packages/astro/src/server/get-auth.ts
@@ -1,6 +1,7 @@
 import type { AuthObject } from '@clerk/backend';
 import { AuthStatus, signedInAuthObject, signedOutAuthObject } from '@clerk/backend/internal';
 import { decodeJwt } from '@clerk/backend/jwt';
+import type { PendingSessionOptions } from '@clerk/types';
 import type { APIContext } from 'astro';
 
 import { getSafeEnv } from './get-safe-env';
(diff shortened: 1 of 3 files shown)


packages/astro/src/server/get-auth.ts
```diff
@@ -37,7 +42,13 @@ export const createGetAuth = ({ noAuthStatusMessage }: { noAuthStatusMessage: st
 
     const jwt = decodeJwt(authToken as string);
     // @ts-expect-error - TODO: Align types
-    return signedInAuthObject(options, jwt.raw.text, jwt.payload);
+    const authObject = signedInAuthObject(options, jwt.raw.text, jwt.payload);
+
+    if (treatPendingAsSignedOut && authObject.sessionStatus === 'pending') {
+      return signedOutAuthObject(options);
+    }
```
brkalow, line 49: This seems like something that _should_ be handled in `@clerk/backend`, but we can do it in a followup. Something like:

```
getAuthObjectFromJwt(jwt)
```

brkalow (approved):

packages/astro/src/server/get-auth.ts
LauraBeatris: Agreed! I'll raise a follow-up PR for it.

PR #5776, opened 2025-04-30
feat(testing): Add checkout and pricingTable page objects

dstaley:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/integration/tests/pricing-table.test.ts
+++ b/integration/tests/pricing-table.test.ts
@@ -23,13 +23,16 @@ testAgainstRunningApps({ withEnv: [appConfigs.envs.withBilling] })('pricing tabl
   test('renders pricing table with plans', async ({ page, context }) => {
     const u = createTestUtils({ app, page, context });
     await u.po.page.goToRelative('/pricing-table');
+
+    await u.po.pricingTable.waitForMounted();
     await expect(u.po.page.getByRole('heading', { name: 'Pro' })).toBeVisible();
   });
 
   test('when signed out, clicking get started button navigates to sign in page', async ({ page, context }) => {
     const u = createTestUtils({ app, page, context });
     await u.po.page.goToRelative('/pricing-table');
-    await u.po.page.getByText('Get started').first().click();
+
+    await u.po.pricingTable.startCheckout({ planSlug: 'plus' });
     await u.po.signIn.waitForMounted();
     await expect(u.po.page.getByText('Checkout')).toBeHidden();
   });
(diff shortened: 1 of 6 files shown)


integration/tests/pricing-table.test.ts
```diff
@@ -48,26 +53,54 @@ testAgainstRunningApps({ withEnv: [appConfigs.envs.withBilling] })('pricing tabl
...
+  });
 
-    await u.po.page.getByRole('button', { name: 'Pay $' }).click();
+  test('can upgrade to a new plan with saved card', async ({ page, context }) => {
+    const u = createTestUtils({ app, page, context });
+    await u.po.signIn.goTo();
+    await u.po.signIn.signInWithEmailAndInstantPassword({ email: fakeUser.email, password: fakeUser.password });
+    await u.po.page.goToRelative('/pricing-table');
+
+    await u.po.pricingTable.startCheckout({ planSlug: 'pro', shouldSwitch: true });
```
alexcarpenter, line 70: Not a super strong opinion, but shouldSwitch is not super obvious. `switchPlans` maybe? Again not super strong opinion here.

packages/testing/src/playwright/unstable/page-objects/checkout.ts
```diff
@@ -0,0 +1,47 @@
...
+      const frame = page.frameLocator('iframe[src*="elements-inner-payment"]');
+      await frame.getByLabel('Card number').fill(card.number);
+      await frame.getByLabel('Expiration date').fill(card.expiration);
+      await frame.getByLabel('Security code').fill(card.cvc);
+      await frame.getByLabel('Country').selectOption(card.country);
+      await frame.getByLabel('ZIP code').fill(card.zip);
+    },
+    continueWithSavedCard: async () => {
+      await page
+        .locator('.cl-disclosureRoot')
```
alexcarpenter, line 30: do we need to select the disclosureRoot here vs just selecting the button trigger?

alexcarpenter (approved):
Looks good! Two small non-blocking questions.

packages/testing/src/playwright/unstable/page-objects/checkout.ts
dstaley: The button to pay with a saved card and the button to pay with a new card use the same labels, so we need to distinguish between them. This selector is "select the subscribe/pay button that's a child of a cl-disclosureRoot containing a button with the name Payment Methods" which distinguishes it from the `Pay $x.xx` button within the Add new payment method form (which is still in the DOM, but not visible). I wanted this method to work even if the `Add new payment method` disclosure was open.

Screen Shot 2025-05-01 at 10 49 26 AM

octoper (approved):

packages/testing/src/playwright/unstable/page-objects/index.ts
```diff
@@ -30,10 +32,12 @@ export const createPageObjects = ({
   return {
     page: app,
     clerk: createClerkPageObject(testArgs),
+    checkout: createCheckoutPageObject(testArgs),
```
panteliselef, line 35: @dstaley since we only expose checkout as an internal component currently, shall we follow the pattern here as well ? We can drop the internal prefix later once, checkout is a standalone component.

dstaley: I think given that Checkout is intended to be a public component in the near future we're fine to leave this as just `checkout`. We're under an `unstable` import so I don't think we really need to double up on the warnings.

PR #5779, opened 2025-04-30
fix(clerk-js,types): Ensure proper checkout plandetails drawer is opened within profile components

alexcarpenter:
## Description

Fixes issue where if two Profile components were rendered at once, the Checkout and PlanDetails drawers would be only portal'd into the first instance on the page.

The PR refactors the drawer functionality to use portalRoot instead of the id to be able to select the closest profile root that we should be rendering within.

I've also updated to use a data attribute vs id to avoid duplicate ids on the page.

BEFORE:

https://github.com/user-attachments/assets/85e85405-c022-4b45-82fd-ba6abd02b70b

AFTER:

https://github.com/user-attachments/assets/18676057-2984-41a3-a4b0-001a3dbe5ae0

Fixes COM-688

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


aeliox (approved):
LGTM

PR #5784, opened 2025-05-01
feat(e2e): Use Turbopack in Next.js app router e2e

dstaley:
## Description

This PR updates our Next.js App Router template to use Turbopack by default. While this doesn't dramatically decrease the execution time, it does make them slightly faster to run locally.

The improved performance of Turbopack versus Webpack surfaced an issue with how we use `signInWithEmailAndInstantPassword` in the billing tests. Since we performed a navigation immediately after calling the signIn method, the page was reloaded before the sign-in could be completed. This is because Turbopack was able to response with the HTML for the document request faster than Webpack. This was fixed by adding a new parameter to the method, `waitForSession` (defaults to `true`) that will wait for `window.Clerk.session` to the truthy before resolving the promise.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


integration/presets/next.ts
```diff
@@ -8,7 +8,7 @@ const appRouter = applicationConfig()
   .useTemplate(templates['next-app-router'])
   .setEnvFormatter('public', key => `NEXT_PUBLIC_${key}`)
   .addScript('setup', constants.E2E_NPM_FORCE ? 'pnpm install --force' : 'pnpm install')
-  .addScript('dev', 'pnpm dev')
+  .addScript('dev', constants.E2E_NEXTJS_VERSION === '13' ? 'pnpm dev:webpack' : 'pnpm dev')
```
dstaley, line 11: Next.js v13 does not support `next dev --turbo`.

integration/templates/next-app-router/next.config.js
```diff
@@ -6,6 +6,7 @@ const nextConfig = {
   experimental: {
     serverActions: true,
   },
+  outputFileTracingRoot: '/',
```
dstaley, line 9: By default, Turbopack doesn't resolve files outside of the project root. We use `pnpm link` to install dependencies, which exist outside of the root. We set the `outputFileTracingRoot` to the root `/` directory to allow Turbopack to resolve dependency files from anywhere on disk.

tmilewski (approved):

PR #5789, opened 2025-05-02
fix(backend): Rename constant to match value

jacekradko:
## Description

Renaming `DEFAULT_CLOCK_SKEW_IN_SECONDS` to `DEFAULT_CLOCK_SKEW_IN_MS` which is the actual value

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/chubby-geese-arrive.md
+++ b/.changeset/chubby-geese-arrive.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
+Renaming DEFAULT_CLOCK_SKEW_IN_SECONDS to DEFAULT_CLOCK_SKEW_IN_MS which is the actual value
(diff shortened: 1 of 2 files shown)


.changeset/chubby-geese-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
+Renaming DEFAULT_CLOCK_SKEW_IN_SECONDS to DEFAULT_CLOCK_SKEW_IN_MS which is the actual value
```
brkalow, line 5: If we want, as this is just an internal change, we can provide an empty changeset:
```suggestion
---
---
```

This can also be done with `px changeset --empty`!

brkalow (approved):

PR #5831, opened 2025-05-05
feat(clerk-js,localizations,types): `Pay with test card` button on `<AddPaymentSource />` component

nikospapcom:
## Description

This pr adds the `Pay with test card` section in the Checkout form
In the add payment method section we still use the test card details instead 

Screenshot 2025-05-05 at 5 10 55 PM

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### User Profile

https://github.com/user-attachments/assets/e5a693ef-385c-44e8-8680-c8ad48d2a534

### Organization Profile

https://github.com/user-attachments/assets/e2b5860b-de67-4b6a-875a-d508d374390c

### DIFF
--- a/.changeset/tiny-wolves-battle.md
+++ b/.changeset/tiny-wolves-battle.md
@@ -1,0 +1,7 @@
+---
+'@clerk/localizations': patch
+'@clerk/clerk-js': patch
+'@clerk/types': patch
+---
+
+Add `Pay with test card` button on `<AddPaymentSource />` component in dev instance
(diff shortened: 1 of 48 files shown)


drewwilson:
I think we can remove the "Pay with test card" after you've already added a card.
So basically IF you have no payment method, show the "Pay with test card" block.. if you DO have a saved payment method, never show it.
Since you can just use the saved payment method.

alexcarpenter (approved):
💯

nikospapcom:
> I think we can remove the "Pay with test card" after you've already added a card. So basically IF you have no payment method, show the "Pay with test card" block.. if you DO have a saved payment method, never show it. Since you can just use the saved payment method.

@drewwilson I'd say that we need to keep the button. The reason behind is that in the next month the test card will be expired so we want to have this button there

octoper (approved):

PR #5846, opened 2025-05-06
chore(clerk-js): Remove Clerk.commerce

brkalow:
## Description

Removes the `Clerk.commerce` namespace in favor of just `Clerk.billing`. The payment source methods that lived under `Clerk.commerce` have been moved to `Clerk.user` and `Clerk.organization`.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


LekoArts (commented):
I think you should also update https://github.com/clerk/javascript/blob/main/.changeset/witty-doors-hear.md because otherwise the changelog is quite confusing. Then the new changeset you can keep minimal as the old one will show the correct one

packages/clerk-js/src/ui/contexts/components/Checkout.ts
```diff
@@ -29,9 +35,22 @@ export const useCheckoutContext = () => {
 
   const { componentName, ...ctx } = context;
 
+  const subscriber = () => {
+    if (ctx.subscriberType === 'org' && clerk.organization) {
+      invariant(clerk.organization, 'Clerk: subscriberType is "org" but no active organization was found');
```
panteliselef, line 40: Should we handle this inside the mount/open singleton methods instead ?

packages/clerk-js/src/ui/contexts/components/Checkout.ts
```diff
@@ -29,9 +35,22 @@ export const useCheckoutContext = () => {
...
+
+      return clerk.organization;
+    }
+
+    invariant(clerk.user, 'Clerk: no active user found');
```
panteliselef, line 45: Do we need this ? `openCheckout` will not mount any component if user is signed out.

panteliselef (approved):
Looks good, left some comments about `invariant`

packages/clerk-js/src/ui/contexts/components/Checkout.ts
brkalow: This is mostly to make types happy without casting, agreed it shouldn't really be necessary in practice

packages/clerk-js/src/ui/contexts/components/Checkout.ts
brkalow: Shouldn't be necessary I agree, it's mostly done to satisfy types in this usage.

PR #5867, opened 2025-05-06
fix(clerk-js): Handle invalid_plan_change error with custom UI

octoper:
## Description

When trying to switch from an annual plan to a monthly plan that has lower price FAPI returns an error, this PR handles that error and displays a custom UI

BEFORE:

CleanShot 2025-05-06 at 22 09 36@2x

AFTER:

image

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/Checkout/CheckoutPage.tsx
+++ b/packages/clerk-js/src/ui/components/Checkout/CheckoutPage.tsx
@@ -153,8 +153,8 @@ export const CheckoutPage = (props: __internal_CheckoutProps) => {
         })}
       >
         <Alert
-          variant='info'
-          colorScheme='info'
+          variant='danger'
+          colorScheme='danger'
         >
           {errors ? translateError(errors[0]) : 'There was a problem, please try again later.'}
         </Alert>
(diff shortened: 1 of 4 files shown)


packages/clerk-js/src/ui/components/Checkout/CheckoutPage.tsx
```diff
@@ -84,6 +96,51 @@ export const CheckoutPage = (props: __internal_CheckoutProps) => {
...
+              borderBottomColor: t.colors.$neutralAlpha100,
+            })}
+          >
+            <LineItems.Root>
+              <LineItems.Group>
+                <LineItems.Title
+                  title={plan.name}
+                  description={planPeriod === 'annual' ? localizationKeys('commerce.billedAnnually') : undefined}
+                />
+                {/* TODO(@Commerce): needs localization */}
```
panteliselef, line 122: Are we still missing localization keys ? If yes this is ok for now, but if these are the only ones, we should localize them.

octoper: These should exists actually, let me remove the comments

PR #5916, opened 2025-05-13
fix(clerk-js): Call setActive after closing checkout drawer

panteliselef:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/Components.tsx
```diff
@@ -372,13 +372,18 @@ const Components = (props: ComponentsProps) => {
...
-      setState(s => ({
-        ...s,
-        [`${name}Drawer`]: {
-          ...s[`${name}Drawer`],
-          open: false,
-        },
-      }));
+      setState(s => {
+        const currentItem = s[`${name}Drawer`];
+        // @ts-expect-error `__internal_PlanDetailsProps` does not accept `onClose`
```
alexcarpenter, line 377: Any strong reason it shouldn't have an onClose prop?

alexcarpenter (approved):

packages/clerk-js/src/ui/Components.tsx
panteliselef: Only reason is that we didn't need it. But when/if we or customers need it we can certainly add it.

octoper (approved):

anagstef (approved):

PR #5927, opened 2025-05-14
chore(repo): Optimize ci.yml

jacekradko:
## Description

Overview of changes:

- separate build step as pre-req for any workflows that assume built packages
- fail fast for unit tests
- run subset of tasks on drafts (skip integration)
- split formatting/changeset/lockfile checks from static analysis, since they don't need built packages
- Continuous Releases with https://pkg.pr.new/

Fixes: SDKI-1007
Fixes: SDKI-1014

Screenshot 2025-05-14 at 8 30 11 PM

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -412,6 +412,4 @@ jobs:
           node-version: 22
 
       - name: Publish with pkg-pr-new
-        env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: pnpm run build && pnpx pkg-pr-new publish --compact --pnpm './packages/*'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -414,4 +414,4 @@ jobs:
       - name: Publish with pkg-pr-new
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-        run: pnpm turbo runpkg-pr-new
+        run: pnpm turbo run pkg-pr-new
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,10 +83,53 @@ jobs:
             pnpm changeset status --since=origin/main;
           fi
 
-  formatting-linting:
+  build-packages:
     needs: [check-permissions]
     if: ${{ github.event_name != 'pull_request' || github.event.pull_request.draft == false }}
-    name: Static analysis & build
+    name: Build Packages
+    runs-on: 'blacksmith-8vcpu-ubuntu-2204'
+    permissions:
+      contents: read
+    defaults:
+      run:
+        shell: bash
+    timeout-minutes: ${{ vars.TIMEOUT_MINUTES_NORMAL && fromJSON(vars.TIMEOUT_MINUTES_NORMAL) || 10 }}
+
+    env:
+      TURBO_SUMMARIZE: false
+
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          show-progress: false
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -292,3 +389,27 @@ jobs:
...
+          show-progress: false
+
+      - name: Setup Node
+        uses: ./.github/actions/init-blacksmith
+        with:
+          turbo-enabled: true
+          node-version: 22
+
+      - name: Publish with pkg-pr-new
+        run: pnpm run build && pnpx pkg-pr-new publish --compact --pnpm './packages/*'
```
nikosdouvlis, line 415: ❓ is there any benefit in using `pnpx` here instead of installing `pkg-pr-new` in the root package.json?

nikosdouvlis (approved):

.github/workflows/ci.yml
nikosdouvlis: BTW, do want this to run for each commit or should we trigger it by commenting `!snapshot` (essentially replacing the prev snapshot impl)

jacekradko: > ❓ is there any benefit in using `pnpx` here instead of installing `pkg-pr-new` in the root package.json?

It allows us not to carry `pkg-pr-new` as a dependency that gets installed all the time and requires upkeep. It's only used in its own CI workflow so the delay introduced by pnpx isn't really an issue

jacekradko: > BTW, do want this to run for each commit or should we trigger it by commenting `!snapshot` (essentially replacing the prev snapshot impl)

@nikosdouvlis That's a good question. Their docs market it as Continuous Releases, so they point you to running it on all PR commits. It could be overkill and a bit wasteful of CI resources if we don't really use these releases a lot. A comment or PR label as opt-in might be better. @clerk/sdk-infra thoughts?

.github/workflows/ci.yml
```diff
@@ -292,3 +389,27 @@ jobs:
...
+      run:
+        shell: bash
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          show-progress: false
+
+      - name: Setup Node
```
dstaley, line 408: If we want turbo caching to work, we'll need these environment variables https://github.com/clerk/javascript/blob/e367ffb03fb03eeaf644b63a51ff9d8642923688/.github/workflows/ci.yml#L113-L116

.github/workflows/ci.yml
```diff
@@ -111,17 +188,29 @@ jobs:
           retention-days: 5
 
   unit-tests:
-    needs: [check-permissions]
+    needs: [check-permissions, build-packages]
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.draft == false }}
     name: Unit Tests
+    permissions:
```
dstaley, line 194: Is this a new requirement? The previous workflow didn't have these permissions, but artifacts were uploaded (as far as I know that is)

.github/workflows/ci.yml
jacekradko: Addressed

.github/workflows/ci.yml
jacekradko: These are more restrictive then default token permissions so they are an additional layer of locking things down. I am fine if we want to skip it though

dstaley: oh no in that case let's leave them! I thought they were adding previously unavailable permissions

PR #5928, opened 2025-05-14
chore(clerk-js): Tidy up layout and logic of PlanDetails drawer

aeliox:
## Description

Update `PlanDetails` to behave similarly to the plan cards used in the PricingTable.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
```diff
@@ -40,18 +40,26 @@ const PlanDetailsInternal = ({
...
+    isDefaultPlanImplicitlyActiveOrUpcoming,
+  } = usePlansContext();
   const subscriberType = useSubscriberTypeContext();
   const canManageBilling = useProtect(
     has => has({ permission: 'org:sys_billing:manage' }) || subscriberType === 'user',
   );
 
+  useEffect(() => {
+    setPlanPeriod(_planPeriod);
+  }, [_planPeriod]);
```
panteliselef, line 56: Let's not sync state like that

packages/clerk-js/src/ui/contexts/components/Plans.tsx
```diff
@@ -329,13 +329,14 @@ export const usePlansContext = () => {
...
     }: HandleSelectPlanProps) => {
-      const subscription = activeOrUpcomingSubscription(plan);
+      const subscription = activeOrUpcomingSubscriptionWithPlanPeriod(plan, planPeriod);
 
       const portalRoot = getClosestProfileScrollBox(mode, event);
 
       if (subscription && subscription.planPeriod === planPeriod && !subscription.canceledAt) {
         clerk.__internal_openPlanDetails({
           plan,
+          planPeriod,
```
panteliselef, line 339: Shouldn't this information be picked up from the plan that we are already passing ?

aeliox: I think was a bug / oversight, because without this, opening the PlanDetails drawer from the SubscriptionsList would default the planPeriod to `month` even when the subscription being referenced had an annual planPeriod.

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
aeliox: Need to think more about this. We want the PlanDetails to manage its own internal planPeriod state, defaulting to the planPeriod that it's passed as a prop, and updating anytime the prop updates.

alexcarpenter (approved):

panteliselef:
I believe the comments are not addressed.

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
aeliox: PlanDetails drawer unmounts now, so useEffect is dropped

panteliselef (approved):
I don't want to delay this any longer, but I think it would be best to add an e2e test that checks for the period toggle behaviour.

PR #5941, opened 2025-05-15
chore(repo): Pin version of pkg-pr-new

jacekradko:
## Description

Pinning version of `pr-pkg-new` in CI

Screenshot 2025-05-15 at 6 56 52 PM

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -420,5 +420,11 @@ jobs:
           turbo-team: ${{ vars.TURBO_TEAM }}
           turbo-token: ${{ secrets.TURBO_TOKEN }}
 
+      - name: Debug Environment
+        run: |
+          echo "PKG_PR_NEW_VERSION: ${{ env.PKG_PR_NEW_VERSION }}"
+          echo "All env vars:"
+          env | sort
+
       - name: Publish with pkg-pr-new
         run: pnpm run build && pnpx pkg-pr-new@${{ env.PKG_PR_NEW_VERSION }} publish --compact --pnpm './packages/*'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -401,6 +401,7 @@ jobs:
 
     env:
       TURBO_SUMMARIZE: false
+      PKG_PR_NEW_VERSION: latest
```
panteliselef, line 404: Isn't this the same as before, since we set it to latest ?

jacekradko: That was my fumbling attempt to use the env var

.github/workflows/ci.yml
```diff
@@ -420,4 +420,4 @@ jobs:
           turbo-token: ${{ secrets.TURBO_TOKEN }}
 
       - name: Publish with pkg-pr-new
-        run: pnpm run build && pnpx pkg-pr-new publish --compact --pnpm './packages/*'
+        run: pnpm run build && pnpx pkg-pr-new@${{ vars.PKG_PR_NEW_VERSION || 'latest' }} publish --compact --pnpm './packages/*'
```
dstaley, line 423: Can we replace `'latest'` with the current version? As it stands now if we didn't have the env var defined, we'd be loading the latest version. Given that only certain folks on the team have env var permission, it's possible that we'd be silently defaulting to `latest` without knowing it. I'm okay with using env var provided there's a fallback to a specific known version.

panteliselef (approved):

dstaley (approved):

PR #5952, opened 2025-05-16
chore(clerk-js): Refactor CheckoutPage into composable stages

panteliselef:
## Description

Refactoring CheckoutPage to make it easy to identify all states of the component.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/components/Checkout/CheckoutComplete.tsx
```diff
@@ -119,14 +117,11 @@ export const CheckoutComplete = ({
...
                 style={{
                   strokeDashoffset: '1',
-                  animationName: 'check',
-                  animationDuration: `${transitionDurationValues.drawer}ms`,
-                  animationTimingFunction: transitionTiming.bezier,
-                  animationFillMode: 'forwards',
-                  animationDelay: `${transitionDurationValues.slow}ms`,
+                  animation: isMotionSafe
+                    ? `check ${transitionDurationValues.drawer}ms ${transitionTiming.bezier} forwards ${transitionDurationValues.slow}ms`
+                    : 'none',
```
panteliselef, line 122: After the changes of this PR the previous implementation started to break and the I was getting this error on the console

> "Removing a style property during rerender (animation) when a conflicting property is set (animationDelay) can lead to styling bugs. To avoid this, don't mix shorthand and non-shorthand properties for the same value; instead, replace the shorthand with separate values."

packages/clerk-js/src/ui/components/Checkout/CheckoutPage.tsx
```diff
@@ -1,157 +1,136 @@
...
-          fontSize: t.fontSizes.$md,
-        })}
-      >
-        <Alert
-          variant='danger'
-          colorScheme='danger'
-        >
-          {errors ? translateError(errors[0]) : t(localizationKeys('unstable__errors.form_param_value_invalid'))}
-        </Alert>
-      </Flex>
```
panteliselef, line 154: Moved to `parts.tsx`

packages/clerk-js/src/ui/components/Checkout/useCheckout.ts
```diff
@@ -1,63 +0,0 @@
...
-  const { planId, planPeriod, subscriberType = 'user' } = props;
-  const clerk = useClerk();
-  const { organization } = useOrganization();
-
-  const { user } = useUser();
-
-  const cacheKey = {
-    key: `commerce-checkout`,
-    userId: user?.id,
-    arguments: {
```
panteliselef, line 17: This is not something that should be used/called multiple times, so it is moved inside `CheckoutContextRoot`

PR #5969, opened 2025-05-20
fix(clerk-js): Maintain identifier selection in transfer flow

jacekradko:
## Description

Fixes: SDKI-1015

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
+++ b/packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
@@ -44,11 +44,12 @@ function SignUpStartInternal(): JSX.Element {
   const isWithinSignInContext = !!React.useContext(SignInContext);
   const { afterSignUpUrl, signInUrl, unsafeMetadata } = ctx;
   const isCombinedFlow = !!(ctx.isCombinedFlow && !!isWithinSignInContext);
+  const initialValues = ctx.initialValues || {};
+
   const [activeCommIdentifierType, setActiveCommIdentifierType] = React.useState<ActiveIdentifier>(
-    getInitialActiveIdentifier(attributes, userSettings.signUp.progressive),
+    getInitialActiveIdentifier(attributes, userSettings.signUp.progressive, initialValues),
   );
   const { t, locale } = useLocalizations();
-  const initialValues = ctx.initialValues || {};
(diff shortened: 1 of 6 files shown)


packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
```diff
@@ -44,11 +44,12 @@ function SignUpStartInternal(): JSX.Element {
   const isWithinSignInContext = !!React.useContext(SignInContext);
   const { afterSignUpUrl, signInUrl, unsafeMetadata } = ctx;
   const isCombinedFlow = !!(ctx.isCombinedFlow && !!isWithinSignInContext);
+  const initialValues = ctx.initialValues || {};
+
   const [activeCommIdentifierType, setActiveCommIdentifierType] = React.useState<ActiveIdentifier>(
-    getInitialActiveIdentifier(attributes, userSettings.signUp.progressive),
+    getInitialActiveIdentifier(attributes, userSettings.signUp.progressive, initialValues),
```
alexcarpenter, line 50: I looked into this briefly the other day, and had a working solution which was simply checking for isCombinedFlow, and whether the signUp resource had a phone number filled. Not sure if it was the right approach but worked from what I had quickly tested. Would be able to avoid needing to pass params around.

```tsx
const [activeCommIdentifierType, setActiveCommIdentifierType] = React.useState<ActiveIdentifier>(
    isCombinedFlow && signUp?.phoneNumber
      ? 'phoneNumber'
      : getInitialActiveIdentifier(attributes, userSettings.signUp.progressive),
  );
```

jacekradko: Hmm.. yeah, we could do that. Definitely much simpler

alexcarpenter: The code above was just a test to see if it'd work, I'd assume `getInitialActiveIdentifier` should handle that internally though. Again, not sure which is the preferred approach. Not sure if I am missing something about the phoneNumber being available on signUp.

jacekradko: https://github.com/clerk/javascript/pull/5972 if we think that is sufficient

jacekradko:
Merged https://github.com/clerk/javascript/pull/5972 instead

PR #5990, opened 2025-05-23
fix(clerk): Do not throttle touch for multisession apps

nikosdouvlis:
## Description

This PR drops throttling for multisession apps to fix edge cases when quickly switching between tabs with different sessions, in apps that have multisession support enabled.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -2434,6 +2434,7 @@ export class Clerk implements ClerkInterface {
         return;
       }
 
+      // In multi-session apps, it's possible that different tabs will have different active sessions. It's critical that the tab's active session is touched in this case so the session is properly updated on the backend, and so we avoid any throttling when multi-session mode is enabled.
       const multisessionMode = this.environment && !this.environment.authConfig.singleSessionMode;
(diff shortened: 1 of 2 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -2434,7 +2434,8 @@ export class Clerk implements ClerkInterface {
         return;
       }
 
-      if (this.#touchThrottledUntil > Date.now()) {
+      const multisessionMode = this.environment && !this.environment.authConfig.singleSessionMode;
+      if (!multisessionMode && this.#touchThrottledUntil > Date.now()) {
```
brkalow, line 2438:
```suggestion
      // In multi-session apps, it's possible that different tabs will have different active sessions. It's critical that the tab's active session is touched in this case so the session is properly updated on the backend, and so we avoid any throttling when multi-session mode is enabled.
      const multisessionMode = this.environment && !this.environment.authConfig.singleSessionMode;
      if (!multisessionMode && this.#touchThrottledUntil > Date.now()) {
```

PR #5992, opened 2025-05-23
fix(clerk-js): Do not treat phone_number as optional if used as identifier

dstaley:
## Description

This PR fixes an issue where the combined flow wouldn't trigger correctly when invoked with a phone number while phone number was an optional field. It resolves the issue by treating phone_number as a non-optional field when a phone number identifier is used. Additionally, it adds some very, _very_ basic unit tests for the `handleCombinedFlowTransfer` function.

For context, we don't automatically convert sign ins to sign ups in the combined flow if there are possible optional fields that the user might want to complete. Before this PR, `phone_number` was an optional field. If the user tried to sign in with a phone number not associated with an account, we would consider `phone_number` an optional field, and redirect to the `create` step. This was unnecessary since the user already provided a phone number.

Fixes SDKI-1020

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/little-wings-bathe.md
+++ b/.changeset/little-wings-bathe.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Fix issue where the combined flow wouldn't trigger if a phone number was used as an identifier while set as an optional field.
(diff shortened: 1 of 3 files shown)


brkalow (approved):
👏

octoper (approved):

PR #6009, opened 2025-05-26
feat(clerk-js): Initiate enterprise SSO from ticket flows

LauraBeatris:
## Description

We've introduced FAPI changes to return `missing_requirements` / `needs_first_factor` when a sign-in/sign-up gets created with an org invite that is related to an active enterprise connection. 

This PR updates the ticket flows to check against those new statuses and initiate enterprise SSO redirection.

https://github.com/user-attachments/assets/460b4862-c9cb-4a12-8cad-0ab67a034b63

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
+++ b/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
@@ -222,7 +222,7 @@ function SignInStartInternal(): JSX.Element {
       .then(res => {
         switch (res.status) {
           case 'needs_first_factor':
-            if (res.supportedFirstFactors?.every(ff => ff.strategy === 'enterprise_sso')) {
+            if (hasOnlyEnterpriseSSOFirstFactors(res)) {
               return authenticateWithEnterpriseSSO();
             }
 
@@ -245,6 +245,9 @@ function SignInStartInternal(): JSX.Element {
         return attemptToRecoverFromSignInError(err);
       })
       .finally(() => {
+        const isRedirectingToSSOProvider = hasOnlyEnterpriseSSOFirstFactors(signIn);
+        if (isRedirectingToSSOProvider) return;
+
         status.setIdle();
         card.setIdle();
       });
@@ -368,7 +371,7 @@ function SignInStartInternal(): JSX.Element {
           }
           break;
         case 'needs_first_factor':
-          if (res.supportedFirstFactors?.every(ff => ff.strategy === 'enterprise_sso')) {
+          if (hasOnlyEnterpriseSSOFirstFactors(res)) {
             await authenticateWithEnterpriseSSO();
             break;
           }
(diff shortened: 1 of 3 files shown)


LauraBeatris:
> Changes looks good to me!
> 
> Question: Is this [OIDC prompt](https://github.com/clerk/javascript/pull/6009/files#diff-9e74583c1fec0058f828916858e72a14b8decf3e98d49db00fd7455d150b71d5R164) change related with this PR?

@NicolasLopes7 That's correct! We need to forward `oidcPrompt` to every single enterprise SSO call now.

packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
```diff
@@ -145,8 +146,13 @@ function SignUpStartInternal(): JSX.Element {
           setMissingRequirementsWithTicket(true);
         }
 
+        const redirectUrl = ctx.ssoCallbackUrl;
```
panteliselef, line 149: Is `ssoCallbackUrl`  always defined (because of a fallback) ? Because that can affect non sso flows.

packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
```diff
@@ -222,6 +222,10 @@ function SignInStartInternal(): JSX.Element {
       .then(res => {
         switch (res.status) {
           case 'needs_first_factor':
+            if (hasOnlyEnterpriseSSOFirstFactors(res)) {
+              return authenticateWithEnterpriseSSO();
+            }
+
```
panteliselef, line 228: Is there a reason we don't introduce a new status and let FAPI to drive this directly ? something like `needs_sso_first_factor` (probably consult with a BE engineer). 

The reason I'm saying is that i think it would be easier for custom flows to identify this case just by looking at the `status`.

packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
```diff
@@ -241,6 +245,12 @@ function SignInStartInternal(): JSX.Element {
         return attemptToRecoverFromSignInError(err);
       })
       .finally(() => {
+        // Keep the card in loading state during SSO redirect to prevent UI flicker
+        // This is necessary because there's a brief delay between initiating the SSO flow
+        // and the actual redirect to the external Identity Provider
+        const isRedirectingToSSOProvider = hasOnlyEnterpriseSSOFirstFactors(signIn);
+        if (isRedirectingToSSOProvider) return;
```
panteliselef, line 252: Is there a case were `isRedirectingToSSOProvider` can be true but  the sign in attempt to had a status different than the expected one ? I believe no, just making sure.

packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
LauraBeatris: We default `ssoCallbackUrl` to `/sso-callback` - this is the first time it's being introduced here, pretty much propagating to the next verification prepare, now that the ticket strategy is returning `enterprise_sso` as `missing_requirements`

`redirectUrl` and `redirectUrlComplete` are only being forwarded to `authenticateWithRedirect` for `enterprise_sso` within `completeSignUpFlow`

packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
LauraBeatris: I believe there's a backwards compatibility argument for it, in which you could have multiple first factors 🤔

panteliselef (approved):

packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx
LauraBeatris: `enterprise_sso` is a factor that only gets returned for sign-in/sign-up resources with either `needs_first_factor` or `missing_requirements` statuses.

PR #6053, opened 2025-06-02
chore(backend): Introduce `getAuthObjectFromJwt` as internal utility 

LauraBeatris:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/backend/src/tokens/authObjects.ts
```diff
@@ -326,3 +328,19 @@ const createGetToken: CreateGetToken = params => {
...
+  { treatPendingAsSignedOut, ...options }: PendingSessionOptions & Partial<AuthenticateContext>,
+) => {
+  const authObject = signedInAuthObject(options, jwt.raw.text, jwt.payload);
+
+  if (treatPendingAsSignedOut && authObject.sessionStatus === 'pending') {
+    return signedOutAuthObject(options);
+  }
+
+  return authObject;
+};
```
panteliselef: @LauraBeatris @wobsoriano should this cover the machine token case as well ? I think `AuthObject` from `@clerk/backend` is now extended with that functionality.

wobsoriano: This looks good to me @panteliselef. It's used for session case only and not machine case. We don't need it in any of the machine token functionality

PR #6057, opened 2025-06-03
fix(tanstack-react-start): Apply response headers

dios-david:
## Description

I'm using `@tanstack/react-start@1.120.15` and `@clerk/tanstack-react-start@0.15.8` in a project and I noticed that the Clerk response headers are not being set on the response, which  is needed for some key functionalities (e.g. for the handshake).

I noticed that the `getResponseClerkState()` function already returns the headers, and they are added to `router.update()`, but this doesn't seem to work.

I applied these changes to my local project and I can confirm that I see the response headers now (`x-clerk-auth-reason` and `x-clerk-auth-status` when not logged in), but I'm not sure what is the best way to fully verify this - I added a test case to the integration tests but I don't have access to run them.

Any feedback or help would be much appreciated!

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


dios-david:
Hi @octoper, tagging you as I saw that you contributed the most to this package. Can you take a look at this please?

wobsoriano (approved):
Awesome catch, thank you for the PR!

dios-david:
@wobsoriano 🙌 
can you help me resolving these integration test issues? not sure but looks like it's something permisson/certificate related

PR #6073, opened 2025-06-05
fix(clerk-js): Do not allow `setActive({ organization: null })` when organization selection is forced

LauraBeatris:
## Checklist

- [ ] `pnpm test` runs as expected.
- [ ] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #6103, opened 2025-06-11
fix(clerk-js): Preserve active organization when attempting to switch to personal workspace

LauraBeatris:
## Description

Related to https://github.com/clerk/javascript/pull/6073 

Instead of throwing an error, this PR maintains the latest active organization ID and does not update the session to the personal workspace. 

Important to note that this only applies to the `Clerk.session` singleton, as FAPI has already been updated on `/tokens` and `/touch` to also handle it in a idempotent way.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of organization selection when force organization selection is enabled, ensuring the active organization is not cleared if set to null.  
- **Tests**
  - Added tests to verify correct behavior when changing the active organization with force organization selection enabled.

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -1071,10 +1071,6 @@ export class Clerk implements ClerkInterface {
         );
       }
 
-      if (organization === null && this.environment?.organizationSettings?.forceOrganizationSelection) {
-        throw new Error('setActive requires an organization parameter when organization selection is forced.');
-      }
-
       const onBeforeSetActive: SetActiveHook =
         typeof window !== 'undefined' && typeof window.__unstable__onBeforeSetActive === 'function'
           ? window.__unstable__onBeforeSetActive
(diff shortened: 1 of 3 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -1071,10 +1071,6 @@ export class Clerk implements ClerkInterface {
         );
       }
 
-      if (organization === null && this.environment?.organizationSettings?.forceOrganizationSelection) {
-        throw new Error('setActive requires an organization parameter when organization selection is forced.');
```
LauraBeatris, line 1075: I believe it's better to not throw an error and just handle it idempotently as we're doing on the backend, cause it might break the entire application.

NicolasLopes7 (approved):

PR #6117, opened 2025-06-13
fix(clerk-js): Do not display organization list after creating organization throw tasks flow

LauraBeatris:
Resolves ORGS-683, ORGS-705

### Issue 

Currently, when creating an organization during the force-an-org step, it re-renders `OrganizationList` to display the list of memberships. This display is shown for a couple of seconds until the navigation to `redirectUrlComplete` is complete. 

### Solution 

This PR introduces a separate chunk of UI for the `force-an-org` step, in which gets rendered only if it's inside a `SessionTasks` flow. With that, it can run a different logic to not display the list of memberships after creation, but only if there are no organization memberships attached to the session. 

https://github.com/user-attachments/assets/0e0214ae-3a6d-4012-b7b3-6a6c0541a660

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- **New Features**
  - Introduced a new flow for organization selection, allowing users to select or create an organization with improved loading states and error handling.
  - Added a dedicated component to manage the forced organization selection process.

- **Bug Fixes**
  - Prevented the organization list from displaying immediately after creating an organization during forced organization selection.

- **Improvements**
  - Enhanced navigation and state management during session tasks, providing clearer feedback and preventing redundant redirects.
  - Updated UI components for more flexible organization creation and selection flows.
  - Improved session navigation hooks to better synchronize client and server authentication states.

- **Chores**
  - Increased JavaScript bundle size limits to accommodate recent updates.


packages/clerk-js/src/ui/components/OrganizationList/OrganizationListPage.tsx
```diff
@@ -227,3 +230,71 @@ const OrganizationListPageList = (props: { onCreateOrganizationClick: () => void
     </>
   );
 };
+
+const ForceOrganizationSelectionFlow = () => {
```
LauraBeatris, line 234: Super WIP, I'll push more commits to extract a couple of pieces here to reusable JSX chunks into the same module.

packages/clerk-js/src/ui/components/OrganizationList/OrganizationListPage.tsx
```diff
@@ -77,46 +78,29 @@ const CreateOrganizationButton = ({
...
-            />
-          </Flex>
-        )}
-
-        {!isLoading && <OrganizationListFlows showListInitially={!(hidePersonal && !hasAnyData)} />}
-      </Card.Content>
-      <Card.Footer />
-    </Card.Root>
+    <FlowCard>
+      {isLoading && <FlowLoadingState />}
```
LauraBeatris, line 94: Could turn this into a ternary

packages/clerk-js/src/ui/components/OrganizationList/OrganizationListPage.tsx
```diff
@@ -77,46 +80,30 @@ const CreateOrganizationButton = ({
...
-
-        {!isLoading && <OrganizationListFlows showListInitially={!(hidePersonal && !hasAnyData)} />}
-      </Card.Content>
-      <Card.Footer />
-    </Card.Root>
+    <FlowCard>
+      {isLoading ? (
+        <FlowLoadingState />
+      ) : (
+        <OrganizationListFlows showListInitially={!hidePersonal && hasOrganizationResources} />
```
LauraBeatris: I also took the liberty of refactoring the `showListInitially` condition, as it was a double negative and therefore difficult to understand at first glance. 

Renamed `hasAnyData` constant to `hasOrganizationResources` as it better indicates entities such as suggestions, memberships, and invitations.

packages/clerk-js/src/ui/components/OrganizationList/OrganizationListPage.tsx
```diff
@@ -227,3 +212,86 @@ const OrganizationListPageList = (props: { onCreateOrganizationClick: () => void
...
+  );
+};
+
+const FlowCard = ({ children }: PropsWithChildren) => {
+  const card = useCardState();
+
+  return (
+    <Card.Root>
+      <Card.Content sx={t => ({ padding: `${t.space.$8} ${t.space.$none} ${t.space.$none}` })}>
+        <Card.Alert sx={t => ({ margin: `${t.space.$none} ${t.space.$5}` })}>{card.error}</Card.Alert>
```
alexcarpenter: Can you share why are we modifying the margin and padding here? Ideally these are consistent across the UI components.

packages/clerk-js/src/ui/components/OrganizationList/OrganizationListPage.tsx
```diff
@@ -227,3 +212,86 @@ const OrganizationListPageList = (props: { onCreateOrganizationClick: () => void
...
+      <Card.Footer />
+    </Card.Root>
+  );
+};
+
+const FlowLoadingState = () => (
+  <Flex
+    direction={'row'}
+    align={'center'}
+    justify={'center'}
```
alexcarpenter:
```suggestion
    center
```

packages/clerk-js/src/ui/components/OrganizationList/OrganizationListPage.tsx
```diff
@@ -227,3 +212,86 @@ const OrganizationListPageList = (props: { onCreateOrganizationClick: () => void
...
+    sx={t => ({
+      height: '100%',
+      minHeight: t.sizes.$60,
+    })}
```
alexcarpenter: Do we need this height/min-height defined? Seeing other usages where that is not the case in loading card states. https://github.com/clerk/javascript/blob/b975d96fa01ad065f98f43246c088866b4c35dec/packages/clerk-js/src/ui/components/SignIn/ResetPasswordSuccess.tsx#L24-L33

LauraBeatris:
Couple of things broke since I was last testing it, so I'm going to re-run the whole flow again and update this PR

packages/clerk-js/src/ui/components/OrganizationList/OrganizationListPage.tsx
LauraBeatris: > these are consistent across the UI components.

Are you referring to the padding and margin consistency across all AIOs instead of using Card defaults? 

My apologies, I just extracted the current implementation in main and moved to different chunks of JSX so that I can reuse it in the other branches of logic now, so I didn't pay enough attention that it was overriding previously. I'm happy to remove it from here if it makes sense!

LauraBeatris:
I'm considering a new approach now, essentially extracting the force-an-org flow to an internal component to be imported within `SessionTasks`, in this way, it won't slightly increase the bundle for `OrganizationsList`

LauraBeatris:
I've extracted the UI to a separate component now, but the UI needs a bit of improvement: 

CleanShot 2025-06-16 at 20 24 27

Tomorrow, I'll introduce UI tests for the `SessionTasks` component as well.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1301,8 +1301,6 @@ export class Clerk implements ClerkInterface {
     const tracker = createBeforeUnloadTracker(this.#options.standardBrowser);
     const defaultRedirectUrlComplete = this.client?.signUp ? this.buildAfterSignUpUrl() : this.buildAfterSignInUrl();
 
-    this.#setTransitiveState();
```
LauraBeatris, line 1304: We do not need the transitive state here. It was causing the `SignIn` component to unmount due to `withCoreSessionGuard` that does not render its children when `session === undefined` 

It makes sense when we're going from an empty session to an actual session, but since in this case we already have one, then I don't see the reason why to set a transitive state here.

LauraBeatris: Here you can see `SignIn` getting unmounted due to `this.#setTransitiveState();`: 


https://github.com/user-attachments/assets/25d79be7-b611-4840-ac25-59fee6dcb217

packages/clerk-js/src/ui/components/SessionTasks/tasks/ForceOrganizationSelection.tsx
```diff
@@ -0,0 +1,140 @@
...
+};
+
+const FlowLoadingState = () => (
+  <Flex
+    direction='row'
+    center
+    sx={t => ({
+      height: '100%',
+      minHeight: t.sizes.$60,
+    })}
```
LauraBeatris, line 108: We need to set a min-height on the loading state while the create organization form isn't rendered. 

Ideally, the loading container should have a similar height as the create organization form / organization list, however it's not possible to know ahead of time since we need to wait for the `useOrganizationList` query to finish.

packages/clerk-js/src/ui/components/SessionTasks/tasks/ForceOrganizationSelection.tsx
```diff
@@ -0,0 +1,140 @@
...
+    </FlowCard>
+  );
+};
+
+const FlowCard = ({ children }: PropsWithChildren) => {
+  const card = useCardState();
+
+  return (
+    <Card.Root>
+      <Card.Content sx={t => ({ padding: `${t.space.$8} ${t.space.$none} ${t.space.$none}` })}>
```
LauraBeatris, line 92: We're resetting the padding on the boundaries in order to have the divider: 

image

Otherwise it'd stay like that:

image

packages/clerk-js/src/ui/components/CreateOrganization/CreateOrganizationForm.tsx
```diff
@@ -87,6 +89,11 @@ export const CreateOrganizationForm = withCardStateProvider((props: CreateOrgani
 
       void userMemberships.revalidate?.();
 
+      if (sessionTasksContext) {
+        await sessionTasksContext.nextTask();
```
LauraBeatris, line 93: This is a pretty similar order as we'd have for custom flows: 

1. Create an organization 
2. Update it to the active session 
3. Verify if there are remaining tasks, and if so, navigate to the next one. 

I've moved it inside of here now instead of passing it within `onComplete` prop so that we can `await` for it. `onComplete` is a sync property and I don't think we should turn into a promise in order to avoid side effects.

LauraBeatris:
Bundle increased quite a lot... let me take a look at it, I think it might have been due to extracting it to a separate component chunk than keeping it within `OrganizationList`. 

CleanShot 2025-06-18 at 12 35 03

packages/clerk-js/src/ui/components/SessionTasks/index.tsx
```diff
@@ -62,34 +51,55 @@ function SessionTaskRoutes(): JSX.Element {
...
 
-  const nextTask = useCallback(
-    () => clerk.__experimental_navigateToTask({ redirectUrlComplete }),
-    [clerk, redirectUrlComplete],
-  );
+  const nextTask = useCallback(() => {
+    setIsNavigatingToTask(true);
+    return clerk.__experimental_navigateToTask({ redirectUrlComplete }).finally(() => setIsNavigatingToTask(false));
+  }, [clerk, redirectUrlComplete]);
+
```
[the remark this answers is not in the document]
LauraBeatris: I don't think this is necessary, as we don't catch errors thrown by `navigate`. The only other async operation that happens within `navigateToTask` is `session.reload` but also doesn't make sense to surface a `GET` request error here.

packages/clerk-js/src/ui/components/SessionTasks/index.tsx
```diff
@@ -62,34 +51,55 @@
...
+    // here to be defensive and ensure proper redirection
+    const task = clerk.session?.currentTask;
+    if (!task || clerk.session?.status === 'active') {
       void navigate(redirectUrlComplete);
       return;
     }
 
     clerk.telemetry?.record(eventComponentMounted('SessionTask', { task: task.key }));
-  }, [clerk, navigate, redirectUrlComplete]);
+  }, [clerk, navigate, isNavigatingToTask, redirectUrlComplete]);
```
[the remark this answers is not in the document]
LauraBeatris: > currentTask becomes undefined) will not re-trigger the effect

That's exactly what we want, to run this effect only on mount in order to handle cases where a user presses a back button the tasks URL.

packages/clerk-js/bundlewatch.config.json
```diff
@@ -1,8 +1,8 @@
 {
   "files": [
-    { "path": "./dist/clerk.js", "maxSize": "610kB" },
-    { "path": "./dist/clerk.browser.js", "maxSize": "70.16KB" },
-    { "path": "./dist/clerk.legacy.browser.js", "maxSize": "113KB" },
+    { "path": "./dist/clerk.js", "maxSize": "612kB" },
```
LauraBeatris, line 3: CleanShot 2025-06-18 at 13 21 54

The bundle increased around 4KB since I extracted the force-an-org flow outside of `OrganizationList`. It increased the general bundle, but we are lazy loading `SessionTasks` within `SignIn`. 

I think the tradeoff is worth considering. I didn't want to introduce another `if (sessionTasksContext)` within `OrganizationList` to hijack the logic in there, but rather keep it separate. 

Ideally, would be nice to use `Suspense` to only download the necessary after-auth steps according to `Enviroment` settings.

packages/clerk-js/src/ui/components/SessionTasks/tasks/ForceOrganizationSelection.tsx
```diff
@@ -0,0 +1,140 @@
...
+        componentName: 'OrganizationList',
+        skipInvitationScreen: true,
+      }}
+    >
+      <FlowCard>
+        <OrganizationListPageList onCreateOrganizationClick={() => setShowCreateOrganizationForm(true)} />
+      </FlowCard>
+    </OrganizationListContext.Provider>
+  );
+};
```
brkalow: This feels odd. `OrganizationSelectionPage` can render `CreateOrganizationPage` conditionally, but also `ForceOrganizationSelectionFlows` handles conditional rendering of the two. Is it feasible to consolidate this conditional logic? Maybe I'm missing something obvious. 👀

LauraBeatris: There's a button within `OrganizationList` to create an organization if users don't want to choose invitations or suggestions. `OrganizationSelectionPage` has an inner state to toggle between those views.

However, there's the flow where a user does not have organization suggestions or invitations, and they must only see the create organization form.

LauraBeatris: The reason I haven't consolidated those conditionals to only `if (hasData) render <OrganizationSelection />`, is because the `memberships` data gets revalidated after the create organization mutation, and then organization list gets rendered right after for a couple of seconds once `__experimental_nextTask` gets triggered.

LauraBeatris: Actually, let me try something here... I believe now that we're conditionally render `SessionTasks` based on `session.currentTask`, it might not render `ForceOrganizationSelection` even after memberships getting invalidated.

LauraBeatris: Alright! I removed a ton of boilerplate [here](https://github.com/clerk/javascript/pull/6117/commits/0b3144ee4ff013eef9cb6f979fa72033c6e4b57f)

packages/clerk-js/src/ui/components/SessionTasks/tasks/ForceOrganizationSelection.tsx
```diff
@@ -0,0 +1,116 @@
...
+      minHeight: t.sizes.$60,
+    })}
+  >
+    <Spinner
+      size='xl'
+      colorScheme='primary'
+      elementDescriptor={descriptors.spinner}
+    />
+  </Flex>
+);
```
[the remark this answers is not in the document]
LauraBeatris: We already have aria attributes within `Spinner` primitive that covers accessibility:  aria-busy, aria-live='polite'

packages/clerk-js/src/ui/components/SessionTasks/tasks/ForceOrganizationSelection.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { organizationListParams } from '../../OrganizationSwitcher/utils';
+
+/**
+ * @internal
+ */
+export const ForceOrganizationSelectionTask = withCardStateProvider(() => {
+  const { userMemberships, userInvitations, userSuggestions } = useOrganizationList(organizationListParams);
+  const isLoading = userMemberships?.isLoading || userInvitations?.isLoading || userSuggestions?.isLoading;
+  const hasData = !!(userMemberships?.count || userInvitations?.count || userSuggestions?.count);
+
```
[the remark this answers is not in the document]
LauraBeatris: I don't think we handle query errors anywhere in AIOs, so I wouldn't add it here - perhaps it's also connected with the `Card` context.

brkalow (approved):

PR #6142, opened 2025-06-17
feat(clerk-js): Add sandbox visual diffing

tmilewski:
## Description

Adds visual diffing to the ClerkJS sandbox to ensure backwards compatibility when CSS vars and `color-mix` are enabled.

Cmd: `(cd packages/clerk-js && pnpm test:sandbox:integration)`

Fixes USER-2200

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- **New Features**
  - Introduced sandbox integration tests for the sign-in page using Playwright, including visual regression via screenshots.
  - Added helper utilities to streamline integration testing and feature toggling within tests.

- **Chores**
  - Added a `.gitignore` to exclude test artifacts and dependencies from version control.
  - Enhanced development and testing scripts to support sandbox integration scenarios.
  - Updated development server configuration to allow dynamic port assignment based on environment variables.
  - Configured Playwright with tailored settings for sandbox integration testing, including server startup and retry logic.
  - Updated test runner and coverage settings to exclude sandbox test files from Vitest execution and coverage reports.


packages/clerk-js/package.json
```diff
@@ -44,6 +44,7 @@
     "dev:headless": "rspack serve --config rspack.config.js --env variant=\"clerk.headless.browser\"",
     "dev:origin": "rspack serve --config rspack.config.js --env devOrigin=http://localhost:4000",
     "dev:sandbox": "rspack serve --config rspack.config.js --env devOrigin=http://localhost:4000 --env sandbox=1",
+    "dev:sandbox:integration": "rspack serve --config rspack.config.js --env devOrigin=http://localhost:4001 --env sandbox=1",
```
dstaley: nit: if possible, I'd rather change the playwright port than introduce another `rspack` task that differs from the existing one only by port.

packages/clerk-js/package.json
```diff
@@ -81,14 +85,17 @@
     "swr": "2.3.3"
   },
   "devDependencies": {
+    "@clerk/testing": "workspace:^",
     "@emotion/jest": "^11.13.0",
+    "@playwright/test": "^1",
```
dstaley: isn't this a root workspace dependency? does it need to be installed in the `clerk-js` package.json?

jacekradko:
@tmilewski Are we going to keep the snapshots checked into the repo?

packages/clerk-js/package.json
tmilewski: Totally, 👍

tmilewski:
> @tmilewski Are we going to keep the snapshots checked into the repo?

@jacekradko There's less of a need _right now_ as our main goal here is to diff between the two versions in the test. However, long-term, I think that would make sense.

PR #6160, opened 2025-06-19
fix(clerk-js,types): Fix stale `SignIn` on `authenticateWithRedirect` for `enterprise_sso`

LauraBeatris:
## Description

This PR introduces a fix specifically for custom flows with `enterprise_sso`

Currently, `authenticateWithRedirect` always uses the existing client `SignIn` for `saml` or `enterprise_sso` strategy, even if the identifier gets changed by the user in a custom flow, which leads to an issue where FAPI queries the wrong enterprise connection:

https://github.com/user-attachments/assets/3ba67507-1ec8-4861-9350-fd13f4abc3bc

Our `SignInStart` component **always** creates a new `SignIn`; therefore, the flow above works smoothly: https://github.com/clerk/javascript/blob/e1820558624c61f5e76e2d8e1c8132a844005a5a/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx#L299-L312 

#### With fix 

This PR introduces a new `continueSignIn` option for `signIn.authenticateWithRedirect`, which defaults to `false`, therefore it'll create a new `SignIn` object on every call. 

Within `SignInStart`, we pass `continueSignIn` as `false` in order to reuse the current `SignIn` created, and call `/prepare-first-factor` after.

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved an issue where users were redirected to the wrong identity provider during SAML and enterprise SSO sign-in flows.
- **New Features**
	- Enhanced authentication flow to properly refresh the connection identifier on each sign-in attempt for SAML and enterprise SSO.
	- Added an option to continue an existing sign-in session during authentication redirects.
- **Tests**
	- Updated tests to verify the improved SSO authentication behavior.

### DIFF
--- a/.changeset/grumpy-lamps-study.md
+++ b/.changeset/grumpy-lamps-study.md
@@ -1,0 +1,8 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/types': patch
+---
+
+Fixes stale `SignIn` object on `authenticateWithRedirect` for `saml` and `enterprise_sso` custom flows
+
+Previously, the same connection identifier would be used on every `authenticateWithRedirect` call leading to redirecting to the wrong identity provider


packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -232,25 +232,27 @@ export class SignIn extends BaseResource implements SignInResource {
...
-            identifier,
-            redirectUrl: SignIn.clerk.buildUrlWithAuth(redirectUrl),
-            actionCompleteRedirectUrl: redirectUrlComplete,
-            oidcPrompt,
-          });
-
-    const { status, externalVerificationRedirectURL } = firstFactorVerification;
+    const { strategy, redirectUrl, redirectUrlComplete, identifier, oidcPrompt, continueSignIn } = params || {};
+
+    if (!this.id || !continueSignIn) {
```
LauraBeatris, line 236: I found a PR comment a while ago (going to track it again to paste here) that the reasoning for not creating within here was to avoid issues if we created sign-in from AIOs. 

My changes tackle exactly for this case, where `SignIn` provides `continueSignIn` as `true`

panteliselef: Is the issue only `saml` and `enterprise_sso` ? 
I think adding a new property just makes things more complicated for the caller. My understanding is that now developers would need to pass both `identifier` and `continueSignIn`.

Can we get around this by checking if identifiers match ?

LauraBeatris: `enterprise_sso` and `saml` since those rely on the `identifier` to query for the SSO connection. 

They wouldn't have to pass `continueSignIn` to `authenticateWithRedirect` since it defaults to `false`, but it's being passed from `SignInStart` since it always creates a fresh sign-in:

https://github.com/clerk/javascript/blob/e1820558624c61f5e76e2d8e1c8132a844005a5a/packages/clerk-js/src/ui/components/SignIn/SignInStart.tsx#L299-L312

LauraBeatris: > Can we get around this by checking if identifiers match ?

I guess this will work 🤔 but we still have the problem where `SignInStart` calls `authenticateWithRedirect`, so we cannot place this logic there otherwise it'd create two sign-ins.

panteliselef: I see, is that because on `needs_identifier` we want to call prepare instead of create ?

LauraBeatris: > I see, is that because on needs_identifier we want to call prepare instead of create ?

Yeah, exactly, so `SignInStart` needs to handle the progressive flow. Custom flows also need to do so, so it's fair to also always create a fresh sign-in.

PR #6167, opened 2025-06-20
fix(clerk-js): Pass metadata as variables for error localization

anagstef:
## Description

When using localization on `unstable_errors` we pass no variables, although in some cases FAPI returns values in the `meta` property. This PR allows to use the following 2 identifiers that are included in the error object coming from FAPI in the localized error message:

- `identifiers`: This is an array of identifiers
- `email_addresses`: This is an array of email addresses

Example error from FAPI:

```
{
  "code": "not_allowed_access",
  "meta": {
    "identifiers": [
      "example@gmail.com"
    ]
  },
  "message": "Access not allowed.",
  "long_message": "example@gmail.com is not allowed to access this application."
}
```

Example localization value provided by the user:
```
localization={{
  unstable__errors: {
    not_allowed_access: '{{identifiers}} is not allowed to access this page!!!!!',
  },
}}
```

Resulting in the following error:
```
example@gmail.com is not allowed to access this page!!!!!
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- **New Features**
  - Enhanced error localization to include additional metadata, allowing error messages to display more detailed and dynamic information such as identifiers and email addresses.

### DIFF
--- a/packages/clerk-js/src/ui/localization/makeLocalizable.tsx
+++ b/packages/clerk-js/src/ui/localization/makeLocalizable.tsx
@@ -95,7 +95,7 @@ export const useLocalizations = () => {
       return '';
     }
 
-    const metaArgsAsVariables = { identifiers: identifiers?.[0], emailAddresses: emailAddresses?.[0] };
+    const metaArgsAsVariables = { identifiers: identifiers?.join(', '), emailAddresses: emailAddresses?.join(', ') };
 
     return (
       t(localizationKeys(`unstable__errors.${code}__${paramName}` as any, metaArgsAsVariables)) ||
--- a/packages/clerk-js/src/ui/localization/makeLocalizable.tsx
+++ b/packages/clerk-js/src/ui/localization/makeLocalizable.tsx
@@ -89,15 +89,17 @@ export const useLocalizations = () => {
     }
 
     const { code, message, longMessage, meta } = (error || {}) as ClerkAPIError;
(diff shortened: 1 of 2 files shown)


packages/clerk-js/src/ui/localization/makeLocalizable.tsx
```diff
@@ -89,15 +89,17 @@ export const useLocalizations = () => {
     }
 
     const { code, message, longMessage, meta } = (error || {}) as ClerkAPIError;
-    const { paramName = '' } = meta || {};
+    const { paramName = '', identifiers, emailAddresses } = meta || {};
```
tmilewski, line 92: Are these concatenated if returned as arrays?

tmilewski, line 92: Per the example, `not_allowed_access: '{{identifiers}} is not allowed to access this page!!!!!'` - would this be confusing there are multiple `identifiers` (or `emailsAddresses`) returned.

anagstef: These are always arrays with 1 or more items.

Yes, the error message needs to be agnostic on the "is"/"are" part.

clerk-cookie:
Hello 👋

PR #6179, opened 2025-06-23
fix(chrome-extension): Enable support for `Metamask` & `OKX`; Mark unsupported components

tmilewski:
## Description

Explicitly mark `GoogleOneTap` UI components as deprecated and unsupported (due to the requirement of Remotely Hosted Code) to help avoid confusion.

While they, technically, never worked. I plan on removing them in a future major rather than outright changing the public interface.

Enable support for `Metamask` and `OKX` while still splitting out Coinbase remotely hosted code.

USER-2235

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- **New Features**
  - Deprecated the GoogleOneTap and SignInWithMetamaskButton UI components, marking them as unsupported due to security restrictions in browser extensions.
  - Re-exported multiple UI components including ClerkProvider and GoogleOneTap for improved clarity.
- **Bug Fixes**
  - Removed environment-based restrictions and warnings from Web3 authentication methods to ensure consistent behavior.
  - Added a warning and disabled Coinbase Wallet provider in unsupported environments to prevent errors.
- **Documentation**
  - Updated documentation to clarify the unsupported status of these components and provide references to relevant security guidelines.

### DIFF
--- a/packages/chrome-extension/src/index.ts
+++ b/packages/chrome-extension/src/index.ts
@@ -3,7 +3,7 @@ export * from '@clerk/clerk-react';
 export type { StorageCache } from './internal/utils/storage';
 
 // The order matters since we want override @clerk/clerk-react components
-export { ClerkProvider, GoogleOneTap, PricingTable, SignInWithMetamaskButton } from './react';
+export { ClerkProvider, GoogleOneTap } from './react';
 
 // Override Clerk React error thrower to show that errors come from @clerk/chrome-extension
 import { setErrorThrowerOptions } from '@clerk/clerk-react/internal';
--- a/packages/chrome-extension/src/index.ts
+++ b/packages/chrome-extension/src/index.ts
@@ -1,8 +1,9 @@
 export * from '@clerk/clerk-react';
+
 export type { StorageCache } from './internal/utils/storage';
 
-// The order matters since we want override @clerk/clerk-react ClerkProvider
-export { ClerkProvider } from './react';
+// The order matters since we want override @clerk/clerk-react components
+export { ClerkProvider, GoogleOneTap, PricingTable, SignInWithMetamaskButton } from './react';
 
 // Override Clerk React error thrower to show that errors come from @clerk/chrome-extension
 import { setErrorThrowerOptions } from '@clerk/clerk-react/internal';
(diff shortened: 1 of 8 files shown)


brkalow (approved):

packages/chrome-extension/src/react/NotSupported.tsx
```diff
@@ -0,0 +1,17 @@
...
+ * @deprecated This component is not supported in Browser Extensions due to Chrome's security restrictions around remotely hosted code.
+ * @see https://clerk.com/docs/references/browser-extensions/browser-security-restrictions
+ */
+export const PricingTable = () => null;
+
+/**
+ * @deprecated This component is not supported in Browser Extensions due to Chrome's security restrictions around remotely hosted code.
+ * @see https://clerk.com/docs/references/browser-extensions/browser-security-restrictions
+ */
+export const SignInWithMetamaskButton = () => null;
```
panteliselef: I believe we can get around this if needed ! It is caused by a single utility functions that loads the coinbase wallet sdk, which is unrelated to metamask.

packages/chrome-extension/src/react/NotSupported.tsx
```diff
@@ -0,0 +1,17 @@
+/**
+ * @deprecated This component is not supported in Browser Extensions due to Chrome's security restrictions around remotely hosted code.
+ * @see https://clerk.com/docs/references/browser-extensions/browser-security-restrictions
+ */
+export const PricingTable = () => null;
```
panteliselef: Let's not include this, as it will be handled in `COM-954`. We are probably looking at a more fine grain solution here. Also we are missing the UserProfile and OrgProfile billing pages which are also  accessing remotely host code, so let's defer all this work to ticket mentioned above.

packages/chrome-extension/src/index.ts
```diff
@@ -1,8 +1,9 @@
 export * from '@clerk/clerk-react';
```
panteliselef, line 1: @tmilewski Is it better to simply control what's exported here instead of breaking without realising it  ?

tmilewski: Yes, though the plan was to deprecate, then remove in a major.

packages/chrome-extension/src/react/NotSupported.tsx
tmilewski: That would be great, and I've removed this from the PR. We just need to make sure we hit this as PR is a direct result of people getting caught up by the RHC requirement.

packages/chrome-extension/src/react/NotSupported.tsx
tmilewski: Originally, we were targeting pop-ups. Of which, none of the Web3 options worked. Now that we support other variants like side panels, this isn't the case anymore. I've updated to PR to reflect that.

PR #6180, opened 2025-06-23
chore(clerk-js, shared): Create experimental PaymentElement

panteliselef:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/bundlewatch.config.json
+++ b/packages/clerk-js/bundlewatch.config.json
@@ -7,6 +7,7 @@
     { "path": "./dist/ui-common*.js", "maxSize": "110KB" },
     { "path": "./dist/vendors*.js", "maxSize": "40.2KB" },
     { "path": "./dist/coinbase*.js", "maxSize": "38KB" },
+    { "path": "./dist/stripe-vendors*.js", "maxSize": "5.54KB" },
     { "path": "./dist/createorganization*.js", "maxSize": "5KB" },
     { "path": "./dist/impersonationfab*.js", "maxSize": "5KB" },
     { "path": "./dist/organizationprofile*.js", "maxSize": "12KB" },
@@ -21,8 +22,7 @@
     { "path": "./dist/waitlist*.js", "maxSize": "1.5KB" },
     { "path": "./dist/keylessPrompt*.js", "maxSize": "6.5KB" },
     { "path": "./dist/pricingTable*.js", "maxSize": "4.02KB" },
-    { "path": "./dist/checkout*.js", "maxSize": "7.25KB" },
-    { "path": "./dist/paymentSources*.js", "maxSize": "9.17KB" },
+    { "path": "./dist/checkout*.js", "maxSize": "8.30KB" },
     { "path": "./dist/up-billing-page*.js", "maxSize": "3.0KB" },
     { "path": "./dist/op-billing-page*.js", "maxSize": "3.0KB" },
     { "path": "./dist/sessionTasks*.js", "maxSize": "1.5KB" }
--- a/packages/clerk-js/rspack.config.js
+++ b/packages/clerk-js/rspack.config.js
@@ -81,6 +81,7 @@ const common = ({ mode, variant, disableRHC = false }) => {
      * Necessary to prevent the Stripe dependencies from being bundled into
      * SDKs such as Browser Extensions.
      */
+    // TODO: @COMMERCE:  Do we still need this?
     externals: disableRHC ? ['@stripe/stripe-js', '@stripe/react-stripe-js'] : undefined,
     optimization: {
       splitChunks: {
@@ -100,6 +101,11 @@ const common = ({ mode, variant, disableRHC = false }) => {
             name: 'coinbase-wallet-sdk',
             chunks: 'all',
           },
+          stripeVendor: {
+            test: /[\\/]node_modules[\\/](@stripe\/stripe-js|@stripe\/react-stripe-js)[\\/]/,
+            name: 'stripe-vendors',
+            chunks: 'all',
+          },
           /**
            * Sign up is shared between the SignUp component and the SignIn component.
            */
@@ -108,17 +114,11 @@ const common = ({ mode, variant, disableRHC = false }) => {
             name: 'signup',
             test: module => !!(module.resource && module.resource.includes('/ui/components/SignUp')),
           },
-          paymentSources: {
-            minChunks: 1,
-            name: 'paymentSources',
-            test: module =>
-              !!(
-                module.resource &&
-                (module.resource.includes('/ui/components/PaymentSources') ||
-                  // Include `@stripe/react-stripe-js` and `@stripe/stripe-js` in the checkout chunk
-                  module.resource.includes('/node_modules/@stripe'))
-              ),
-          },
+          // paymentSources: {
+          //   minChunks: 1,
+          //   name: 'paymentSources',
+          //   test: module => !!(module.resource && module.resource.includes('/ui/components/PaymentSources')),
+          // },
           common: {
             minChunks: 1,
             name: 'ui-common',
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -645,6 +645,19 @@ export class Clerk implements ClerkInterface {
       .then(controls => controls.closeModal('blankCaptcha'));
   };
 
+  public __internal_loadStripeLibs = () => {
+    return {
+      async javascript(): Promise<unknown> {
+        const { loadStripe } = await import('@stripe/stripe-js');
+        return loadStripe;
+      },
+      async react(): Promise<unknown> {
+        const { Elements, PaymentElement, useElements, useStripe } = await import('@stripe/react-stripe-js');
+        return { Elements, PaymentElement, useElements, useStripe };
+      },
+    };
+  };
+
   public openSignUp = (props?: SignUpProps): void => {
     this.assertComponentsReady(this.#componentControls);
     if (sessionExistsAndSingleSessionModeEnabled(this, this.environment)) {
--- a/packages/clerk-js/src/ui/components/PaymentSources/AddPaymentSource.tsx
+++ b/packages/clerk-js/src/ui/components/PaymentSources/AddPaymentSource.tsx
@@ -1,4 +1,9 @@
-import { createContextAndHook, PaymentElementForm, PaymentElementRoot, usePaymentElement } from '@clerk/shared/react';
+import {
+  __experimental_PaymentElementForm as PaymentElementForm,
+  __experimental_PaymentElementRoot as PaymentElementRoot,
+  __experimental_usePaymentElement as usePaymentElement,
+  createContextAndHook,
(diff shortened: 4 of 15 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -645,6 +645,19 @@ export class Clerk implements ClerkInterface {
...
+      async javascript(): Promise<unknown> {
+        const { loadStripe } = await import('@stripe/stripe-js');
+        return loadStripe;
+      },
+      async react(): Promise<unknown> {
+        const { Elements, PaymentElement, useElements, useStripe } = await import('@stripe/react-stripe-js');
+        return { Elements, PaymentElement, useElements, useStripe };
+      },
+    };
+  };
```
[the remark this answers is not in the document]
panteliselef: This has no benefit

packages/types/src/clerk.ts
```diff
@@ -494,6 +494,15 @@ export interface Clerk {
    */
   __internal_unmountOAuthConsent: (targetNode: HTMLDivElement) => void;
 
+  /**
+   * @internal
+   * Loads Stripe libraries for commerce functionality
+   */
+  __internal_loadStripeLibs: () => {
+    javascript(): Promise<any>;
+    react(): Promise<any>;
```
panteliselef: We can't infer returns values from stripe packages because that would require it to be a dependency or peer dependency and we don't want that.

packages/clerk-js/bundlewatch.config.json
```diff
@@ -21,8 +22,7 @@
     { "path": "./dist/waitlist*.js", "maxSize": "1.5KB" },
     { "path": "./dist/keylessPrompt*.js", "maxSize": "6.5KB" },
     { "path": "./dist/pricingTable*.js", "maxSize": "4.02KB" },
-    { "path": "./dist/checkout*.js", "maxSize": "7.25KB" },
-    { "path": "./dist/paymentSources*.js", "maxSize": "9.17KB" },
+    { "path": "./dist/checkout*.js", "maxSize": "8.30KB" },
```
panteliselef: Removing the chunk for `paymentSources`, increases the chunk for checkout as some of the code is now in that chunk.

packages/clerk-js/src/ui/components/PaymentSources/AddPaymentSource.tsx
```diff
@@ -1,91 +1,76 @@
...
 import { FormContainer } from '@/ui/elements/FormContainer';
 
-import { clerkUnsupportedEnvironmentWarning } from '../../../core/errors';
-import { useEnvironment, useSubscriberTypeContext, useSubscriberTypeLocalizationRoot } from '../../contexts';
+import { useSubscriberTypeContext, useSubscriberTypeLocalizationRoot } from '../../contexts';
 import { descriptors, Flex, localizationKeys, Spinner, useAppearance, useLocalizations } from '../../customizables';
 import type { LocalizationKey } from '../../localization';
 import { handleError, normalizeColorString } from '../../utils';
 
+const useStipeAppearance = () => {
```
aeliox: Stipe > Stripe

packages/shared/src/react/commerce.tsx
```diff
@@ -0,0 +1,334 @@
...
+  ReturnType<typeof usePaymentSourceUtils> & {
+    setIsPaymentElementReady: (isPaymentElementReady: boolean) => void;
+    isPaymentElementReady: boolean;
+    checkout?: CommerceCheckoutResource;
+    // TODO(@COMMERCE): What can we do to remove this ?
+    paymentDescription: string;
+  }
+>('PaymentElementContext');
+
+const [StipeUtilsContext, useStipeUtilsContext] = createContextAndHook<{
```
aeliox: Stipe > Stripe

packages/shared/package.json
```diff
@@ -150,6 +150,8 @@
     "swr": "^2.3.3"
   },
   "devDependencies": {
+    "@stripe/react-stripe-js": "3.1.1",
+    "@stripe/stripe-js": "5.6.0",
```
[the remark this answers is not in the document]
panteliselef: They are used only for their types during dev, does this changes things ?

packages/clerk-js/rspack.config.js
```diff
@@ -81,6 +81,7 @@ const common = ({ mode, variant, disableRHC = false }) => {
      * Necessary to prevent the Stripe dependencies from being bundled into
      * SDKs such as Browser Extensions.
      */
+    // TODO: @COMMERCE:  Do we still need this?
```
panteliselef, line 84: @tmilewski  do we need these to be set as external ? We are now stop using `react-stripe-js'` and I think only `stripe-js'` loads remote hosted code.

Also now don't import `@stripe/stripe-js` directly in the AIOs, but instead we dynamically import it from `Clerk.__internal_loadStripeJs()`

packages/clerk-js/src/ui/components/PaymentSources/AddPaymentSource.tsx
```diff
@@ -16,77 +16,62 @@ import { FormContainer } from '@/ui/elements/FormContainer';
...
 import { descriptors, Flex, localizationKeys, Spinner, useAppearance, useLocalizations } from '../../customizables';
 import type { LocalizationKey } from '../../localization';
 
+const useStripeAppearance = () => {
+  const theme = useAppearance().parsedInternalTheme;
+
+  return useMemo(() => {
+    const { colors, fontWeights, fontSizes, radii, space } = theme;
+    return {
+      colorPrimary: normalizeColorString(colors.$primary500),
```
panteliselef, line 29: `normalizeColorString` seems like an expensive operation.

alexcarpenter: We may be able to remove this now. Since this launched it looks like Stripes appearance object, now supports formats with alphas defined + support for CSS variables.

```markdown
Stripe's Appearance API does support CSS variables now, including those with alpha values. This is a relatively recent enhancement to the API.

The variables option in the Appearance object allows you to use CSS-like variables to customize the look of Elements. These variables can include colors with alpha channels, such as rgba() or hsla() values.

For example, you could use:

example.js
const appearance = {
  theme: 'stripe',
  variables: {
    colorPrimary: 'rgba(0, 112, 243, 0.8)',
    colorBackground: 'hsla(0, 0%, 100%, 0.95)',
  }
};
This flexibility allows for more nuanced control over the appearance of Elements, including semi-transparent colors.

It's worth noting that while the Appearance API now supports these formats, it's always a good idea to test your implementation to ensure compatibility across different browsers and devices.

For the most up-to-date and detailed information on customization options, see [Stripe's Elements Appearance API documentation](https://docs.stripe.com/elements/appearance-api).
```

packages/shared/src/react/commerce.tsx
```diff
@@ -0,0 +1,301 @@
...
+    const { setupIntent, error } = await stripe.confirmSetup({
+      elements,
+      confirmParams: {
+        return_url: '', // TODO(@COMMERCE): need to figure this out
+      },
+      redirect: 'if_required',
+    });
+    if (error) {
+      return { data: null, error } as const;
+    }
```
panteliselef: Should we do some kind of provider specific error mapping here ? Otherwise this error is controlled by Stripe.

alexcarpenter (approved):

packages/shared/src/react/commerce.tsx
```diff
@@ -0,0 +1,301 @@
...
+        layout: {
+          type: 'tabs',
+          defaultCollapsed: false,
+        },
+        paymentMethodOrder,
+        applePay: checkout
+          ? {
+              recurringPaymentRequest: {
+                paymentDescription: paymentDescription || '',
+                managementURL: environment?.displayConfig.homeUrl || '', // TODO(@COMMERCE): is this the right URL?
```
brkalow: ❓ do we need a ticket to make sure we find the right value here?

packages/shared/src/react/commerce.tsx
brkalow: 👍 I think that would be good, even if we just wrap the raw stripe error and provide a clerk-specific error code along with it

packages/shared/src/react/use-previous.ts
```diff
@@ -0,0 +1,45 @@
...
+  const ref = useRef(value);
+
+  useEffect(() => {
+    ref.current = value;
+  }, [value]);
+
+  return ref.current;
+};
+
+export const useAttachEvent = <A extends unknown[]>(
```
brkalow: ❓ this should be in its own file right? currently it's in `use-previous.ts`

packages/shared/src/react/commerce.tsx
```diff
@@ -0,0 +1,301 @@
...
+};
+
+const usePaymentElement = () => {
+  const { isPaymentElementReady, initializePaymentSource } = usePaymentElementContext();
+  const { stripe, elements } = useStripeUtilsContext();
+  const { stripe: stripeFromContext, externalClientSecret } = usePaymentElementContext();
+
+  const submit = useCallback(async () => {
+    if (!stripe || !elements) {
+      throw new Error('Stripe and Elements are not yet ready');
```
brkalow:
```suggestion
      throw new Error('Clerk: Unable to submit, Stripe libraries are not yet loaded');
```

packages/shared/src/react/commerce.tsx
```diff
@@ -0,0 +1,301 @@
...
+        value={{
+          value: {
+            ...utils,
+            setIsPaymentElementReady,
+            isPaymentElementReady,
+            checkout: props.checkout,
+            paymentDescription: props.paymentDescription,
+          },
+        }}
+      >
```
brkalow: This `PaymentElementContext` is identical between the two branches of logic, consider consolidating 👀

packages/shared/src/react/commerce.tsx
panteliselef: No need, this is already address from an already merged PR, now that i've merged from main, it is fixed.

We are now doing

```ts
        managementURL:
          subscriberType === 'org'
            ? environment?.displayConfig.organizationProfileUrl || ''
            : environment?.displayConfig.userProfileUrl || '',
```

packages/shared/src/react/commerce.tsx
panteliselef: I went a step further suggesting this

> Clerk: Unable to submit, Stripe libraries are not yet loaded. Be sure to check `isFormReady` before calling `submit`.

packages/shared/src/react/use-previous.ts
panteliselef: yeap, I rearranged those files

packages/shared/src/react/commerce.tsx
panteliselef: Addressed [here](https://github.com/clerk/javascript/pull/6180/commits/8ff115fc087f7b047e8512a7ba038235177d0079)

PR #6183, opened 2025-06-23
chore(clerk-js): Address leftover commerce TODOs

aeliox:
## Description

Address leftover `TODO`s from Commerce launch.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- **Performance Improvements**
  - Converted several profile and billing-related pages to load only when needed, improving initial load times for user and organization profile sections.

- **Chores**
  - Updated bundle size limits for various profile and billing pages.
  - Added a metadata file for tracking a patch update.
  - Removed outdated TODO comments from several components.
  - Updated redirect URL for Stripe setup confirmation to use the current page location.
  - Adjusted Apple Pay management URL based on subscriber type for improved navigation.


packages/clerk-js/src/ui/components/Checkout/parts.tsx
```diff
@@ -101,7 +100,6 @@ export const AddEmailForm = () => {
           padding: t.space.$4,
         })}
       >
-        {/* TODO(@COMMERCE): How does ths operate for orgs ? */}
```
aeliox, line 104: This flow works for orgs by grabbing the current member's email (once added) and associating that with the payer record

packages/clerk-js/src/ui/components/PaymentSources/AddPaymentSource.tsx
```diff
@@ -250,7 +250,7 @@ const AddPaymentSourceForm = ({ children }: PropsWithChildren) => {
     const { setupIntent, error } = await stripe.confirmSetup({
       elements,
       confirmParams: {
-        return_url: '', // TODO(@COMMERCE): need to figure this out
+        return_url: '',
```
aeliox: We prevent payment methods that require a redirect, so this param is never used (though required)

packages/clerk-js/bundlewatch.config.json
```diff
@@ -9,22 +9,26 @@
...
     { "path": "./dist/pricingTable*.js", "maxSize": "4.02KB" },
     { "path": "./dist/checkout*.js", "maxSize": "7.25KB" },
     { "path": "./dist/paymentSources*.js", "maxSize": "9.17KB" },
-    { "path": "./dist/up-billing-page*.js", "maxSize": "3.0KB" },
+    { "path": "./dist/up-billing-page*.js", "maxSize": "3.5KB" },
     { "path": "./dist/op-billing-page*.js", "maxSize": "3.0KB" },
+    { "path": "./dist/up-plans-page*.js", "maxSize": "1.0KB" },
+    { "path": "./dist/op-plans-page*.js", "maxSize": "1.0KB" },
+    { "path": "./dist/statement-page*.js", "maxSize": "1.0KB" },
+    { "path": "./dist/payment-attempt-page*.js", "maxSize": "3.0KB" },
```
panteliselef, line 31: is there a reason we have seperate chunks between UP and OP for plans, but not for statements and PAs ?

packages/clerk-js/src/ui/components/PaymentSources/AddPaymentSource.tsx
panteliselef: Are we plan to support them in the future ? Are we planning to communicate that they will need to bump sdk versions in order to use it at that point ?

Maybe at least we so something like, unless `''` has the same effect.

```suggestion
        return_url: '',
```
```suggestion
        return_url: window.location.href,
```

packages/clerk-js/src/ui/components/PaymentSources/AddPaymentSource.tsx
```diff
@@ -294,7 +294,7 @@ const AddPaymentSourceForm = ({ children }: PropsWithChildren) => {
               ? {
                   recurringPaymentRequest: {
                     paymentDescription: `${t(localizationKeys(checkout.planPeriod === 'month' ? 'commerce.paymentSource.applePayDescription.monthly' : 'commerce.paymentSource.applePayDescription.annual'))}`,
-                    managementURL: displayConfig.homeUrl, // TODO(@COMMERCE): is this the right URL?
+                    managementURL: displayConfig.homeUrl,
```
panteliselef: this will probably land them to the index page of their website or `/` of AP.

Would it be better to do 
`displayConfig.organizationProfileUrl` for orgs and then `displayConfig.userProfileUrl` for users ?

aeliox: I'm not actually sure what Apple uses these for or where they display them, but since it's called a "management url", the UserProfile / OrgProfile urls probably do make more sense

packages/clerk-js/bundlewatch.config.json
aeliox: Only because based on what I observed, webpack seemed to handle pricing table chunks distinctly and appropriately for OP and UP, but wasn't doing the same for statement page or payment attempt page – instead just grabbing a (seemingly) random one. For example, the UserProfile would always grab `up_pricing_table`, but would grab `op_statement_page` for some reason – maybe they aren't different enough and webpack is being smart?

.changeset/silent-zebras-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Cleanup leftover Commerce-launch TODOs
```
panteliselef: This should be customer facing, or it can be empty

panteliselef (approved):
I added a comment about the changeset file. Please address it and then merge.

PR #6201, opened 2025-06-26
fix(shared): Prevent stale closures in useReverification hook

bratsos:
## Description

Addresses a stale closure bug within the `useReverification` hook and improves the stability of our test environment.

The primary issue was that the hook could capture an initial version of the fetcher function and would not use the updated version on subsequent re-renders.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Enhanced Jest testing environment to provide global access to the `Response` API during tests.

* **Bug Fixes**
  * Improved the stability and correctness of the `useReverification` hook by updating its internal logic for handler creation and invocation.

* **Tests**
  * Added new tests to verify the runtime behavior and function identity stability of the `useReverification` hook.

### DIFF
--- a/packages/shared/src/react/hooks/useReverification.ts
+++ b/packages/shared/src/react/hooks/useReverification.ts
@@ -1,5 +1,5 @@
 import type { Clerk, SessionVerificationLevel } from '@clerk/types';
-import { useMemo, useRef } from 'react';
+import { useCallback, useRef } from 'react';
 
 import { validateReverificationConfig } from '../../authorization';
 import { isReverificationHint, reverificationError } from '../../authorization-errors';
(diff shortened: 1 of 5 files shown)


packages/shared/src/react/hooks/useReverification.ts
```diff
@@ -14,7 +14,7 @@ const CLERK_API_REVERIFICATION_ERROR_CODE = 'session_reverification_required';
 async function resolveResult<T>(result: Promise<T> | T): Promise<T | ReturnType<typeof reverificationError>> {
   try {
     const r = await result;
-    if (r instanceof Response) {
+    if (typeof Response !== 'undefined' && r instanceof Response) {
```
nikosdouvlis: ❓ Just curious, did this ever run in an env where Response was not defined as a global identifier?

nikosdouvlis (approved):

panteliselef (approved):
Makes sense, nice find. If we have a ticket related to this could you add it on the description of the PR ?

packages/shared/src/react/hooks/useReverification.ts
bratsos: Great question! My tests were failing without this, and after digging a bit deeper, I figured that even though we're using jsdom `fetch` is not supported and its maintainers have confirmed [they won't be using Node's native fetch](https://github.com/jsdom/jsdom/issues/1724#issuecomment-1233478314). This is because they need their own spec-compliant implementation that integrates with jsdom's internals, so the missing Response global is intentional.

Instead of changing the production code to handle this, the best solution was to [patch our test environment](https://github.com/clerk/javascript/pull/6201/commits/d4495a0bd5ce1877d0570346e8013f165f20fc51) directly with Node's native fetch APIs (which are stable since v18). This makes the tests pass without altering the hook's logic.

Thanks for the prompt, it definitely led to a better solution 🎉

PR #6231, opened 2025-07-01
test(clerk-js): Add dynamic TTL calculation tests for JWT expiration handling

nikosdouvlis:
Improves the token cache by adding dynamic TTL calculation based on JWT `exp` and `iat` claims.

This change:
- Adds tests to verify token expiration handling based on JWT TTL.
- Creates a helper function to generate JWTs with custom TTL values for testing.
- Ensures tokens with TTLs greater than 60 seconds are handled correctly.


tmilewski (approved):

jacekradko:
Looks good. The test cleanup suggestions from Coderabbit would be nice to implement

PR #6238, opened 2025-07-01
feat(backend): Handles cross origin syncing on primary domains

brkalow:
## Description

fixes USER-2298

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/tokens/__tests__/request.test.ts
+++ b/packages/backend/src/tokens/__tests__/request.test.ts
@@ -193,7 +193,8 @@ expect.extend({
       };
     } else {
       return {
-        message: () => `expected to be signed in, but got ${received.status}`,
+        message: () =>
+          `expected to be signed in, but got ${received.status} (${JSON.stringify(received)}, ${JSON.stringify(expected)})`,
         pass: false,
       };
     }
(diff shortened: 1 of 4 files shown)


packages/backend/src/tokens/__tests__/request.test.ts
```diff
@@ -193,7 +193,8 @@ expect.extend({
       };
     } else {
       return {
-        message: () => `expected to be signed in, but got ${received.status}`,
+        message: () =>
+          `expected to be signed in, but got ${received.status} (${JSON.stringify(received)}, ${JSON.stringify(expected)})`,
```
brkalow:
```suggestion
          `expected to be signed in, but got ${received.status}`,
```

packages/backend/src/tokens/authenticateContext.ts
```diff
@@ -167,6 +167,26 @@ class AuthenticateContext implements AuthenticateContext {
     return true;
   }
 
+  /**
+   * Determines if the request came from a different origin based on the referrer header.
+   * Used for cross-origin detection in multi-domain authentication flows.
+   *
+   * @returns {boolean} True if referrer exists and is from a different origin, false otherwise.
+   */
+  public isCrossOriginReferrer(): boolean {
```
jacekradko, line 176: We could also check this modern header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site

jacekradko (approved):
I think this should work. We can probably enhance this check in the future

packages/backend/src/tokens/authenticateContext.ts
brkalow: Nice, done!

PR #6246, opened 2025-07-02
chore(backend): Add optional secret property in api key response

wobsoriano:
## Description

This PR adds a `secret` property to the API key response which is only available in the `create` method:

```ts
const resp = await clerkClient.apiKeys.create({})

resp.secret === 'ak_xxxxxx'
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * API key responses may now include an optional secret property, providing additional information when available.

### DIFF
--- a/packages/backend/src/api/resources/APIKey.ts
+++ b/packages/backend/src/api/resources/APIKey.ts
@@ -17,6 +17,7 @@ export class APIKey {
     readonly lastUsedAt: number | null,
     readonly createdAt: number,
     readonly updatedAt: number,
+    readonly secret?: string,
   ) {}
 
   static fromJSON(data: APIKeyJSON) {
@@ -36,6 +37,7 @@ export class APIKey {
       data.last_used_at,
       data.created_at,
       data.updated_at,
+      data.secret,
     );
   }
 }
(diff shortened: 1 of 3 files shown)


packages/backend/src/api/resources/APIKey.ts
```diff
@@ -17,6 +17,7 @@ export class APIKey {
     readonly lastUsedAt: number | null,
     readonly createdAt: number,
     readonly updatedAt: number,
+    readonly secret?: string,
```
panteliselef, line 20: Why not 
```suggestion
    readonly secret: string | null,
```

panteliselef (approved):

packages/backend/src/api/resources/APIKey.ts
wobsoriano: This was kind of tricky since [only the `create` method](https://github.com/clerk/javascript/blob/c642d427eb689c90ae8fb7f139a01e7539cce437/packages/backend/src/api/endpoints/APIKeysApi.ts#L39,L45) actually returns a `secret` property. The rest does not. So adding `null` I think is not ideal here 👀 

We'll definitely improve this along the way

PR #6247, opened 2025-07-02
feat(clerk-js): Password manager autofill OTP codes

tmilewski:
## Description

Enhanced OTP support allowing for password managers to automatically enter/submit verification codes

Tested Password Managers:
- [x] 1Password
- [ ] LastPass
- [ ] iOS Keychain
- [ ] Chrome Password Manager
- [ ] Firefox Password Manager
- [ ] Edge Password Manager

Fixes USER-1275

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/elements/CodeControl.tsx
```diff
@@ -274,40 +288,92 @@ export const OTPCodeControl = React.forwardRef<{ reset: any }>((_, ref) => {
...
-          isDisabled={isDisabled || isLoading || disabled || feedbackType === 'success'}
-          hasError={feedbackType === 'error'}
-          isSuccessfullyFilled={feedbackType === 'success'}
-          type='text'
-          inputMode='numeric'
-          name={`codeInput-${index}`}
-        />
-      ))}
-    </Flex>
+    <Box sx={{ position: 'relative' }}>
```
alexcarpenter: Should we have a descriptor?

packages/clerk-js/src/ui/elements/CodeControl.tsx
```diff
@@ -274,40 +288,92 @@ export const OTPCodeControl = React.forwardRef<{ reset: any }>((_, ref) => {
...
+          focusInputAt(0);
+        }}
+        sx={theme => ({
+          position: 'absolute',
+          left: '-9999px',
+          width: `calc(1px + ${passwordManagerOffset}px)`,
+          height: '1px',
+          opacity: theme.opacity.$hidden,
+          pointerEvents: 'none',
+          clipPath: `inset(0 ${passwordManagerOffset}px 0 0)`,
```
alexcarpenter: Would the visuallyHidden util work here? or does this need to have the passwordManagerOffset even when its never visible? https://github.com/clerk/javascript/blob/main/packages/clerk-js/src/ui/styledSystem/common.ts#L193-L202

packages/clerk-js/src/ui/foundations/opacity.ts
```diff
@@ -1,4 +1,5 @@
 export const opacity = Object.freeze({
+  hidden: '0%',
```
alexcarpenter: Not a strong opinion held, but it doesn't feel like we need a hidden opacity within the scale, since that'd always resolve 0%.

alexcarpenter (approved):
Nice! Left a couple of non-blocking comments.

packages/clerk-js/src/ui/elements/CodeControl.tsx
tmilewski: Added!

packages/clerk-js/src/ui/elements/CodeControl.tsx
tmilewski: Replaced with the util!

packages/clerk-js/src/ui/foundations/opacity.ts
tmilewski: Makes sense, removed!

PR #6248, opened 2025-07-03
fix(clerk-js,types,localizations): Adjust invalid plan change alert wording

aeliox:
## Description

Adjusts the wording of our invalid plan change alert when the plan being changed to does not have an annual offering (previously the wording suggested selecting the non-available annual option).

*Current message, and will still be used when new plan has available and more expensive annual plan:*
Screenshot 2025-07-08 at 10 28 58 AM

*New message for when new plan does not have available / more expensive annual plan:*
Screenshot 2025-07-02 at 6 52 48 PM

*Checkout works with no error message when new plan + period is more expensive than existing subscription:*
Screenshot 2025-07-08 at 10 29 10 AM

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Improved alert messaging in the checkout experience by conditionally displaying a new message when users cannot subscribe to a plan with an existing annual subscription.
  * Added a new localization string to support this scenario.
  * Enhanced plan change callout to clarify when a plan upgrade is possible.

* **Bug Fixes**
  * Updated error message shown when a subscription plan cannot be changed due to the current subscription being more expensive.


packages/localizations/src/en-US.ts
```diff
@@ -67,6 +67,7 @@ export const enUS: LocalizationResource = {
     cancelSubscriptionTitle: 'Cancel {{plan}} Subscription?',
     cannotSubscribeMonthly:
       'You cannot subscribe to this plan by paying monthly. To subscribe to this plan, you need to choose to pay annually.',
+    cannotSubscribeMonthlyNoAnnual: 'You cannot subscribe to this plan with an existing annual subscription.',
```
aeliox: I think this wording makes sense? I think this error is only possible if we are "upgrading" plans from an annual subscription to a _cheaper_ monthly subscription – and this error will only show if the new, cheaper plan does _not_ offer an annual version.

packages/clerk-js/src/ui/components/Checkout/parts.tsx
```diff
@@ -82,7 +82,11 @@ export const InvalidPlanScreen = () => {
           <Alert
             variant='info'
             colorScheme='info'
-            title={localizationKeys('commerce.cannotSubscribeMonthly')}
+            title={
+              planFromError.annual_monthly_amount_formatted !== '0'
```
aeliox: We probably want to expose a numeric amount to check against here.

packages/localizations/src/en-US.ts
panteliselef: I'm not sure I understand the difference between the new localization key and the old one and their translations. 

The photo in the description is different from what the value above is. And i think the one in the screenshot is more appropriate.

Also since we already have a test for the [existing case](https://github.com/clerk/javascript/blob/ed54bd83f902ec3cf390188d5f820216bfa86321/integration/tests/pricing-table.test.ts#L322), let's add one more for this new case.

packages/clerk-js/src/ui/components/Checkout/parts.tsx
panteliselef: Let's use number here, what happens if string comes back as `0.00` ?
```suggestion
              planFromError.annual_monthly_amount !== 0
```

If the FAPI does not return numbers, we should update it.

panteliselef (changes requested):

packages/localizations/src/en-US.ts
aeliox: The current message –
`You cannot subscribe to this plan by paying monthly. To subscribe to this plan, you need to choose to pay annually.`
– is fine and makes sense so long as the plan that the C2 is trying to subscribe to _supports_ an annual pay period, otherwise the message is instructing them to do something that doesn't exist.

So this PR is trying to check for that case – the case where the C2 cannot subscribe to this new plan _at all_.

aeliox:
I'm unable to add a test for this without access to the integration test instance – we need a plan configured specifically for this case.

.changeset/orange-doors-notice.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
panteliselef:
```suggestion
---
'@clerk/localizations': minor
'@clerk/clerk-js': patch
'@clerk/types': minor
---

Improve invalid plan change callout for monthly-only plans 
```

panteliselef (approved):

PR #6252, opened 2025-07-03
chore(backend): refactor `verifyWebhook()` to use `standardwebhooks` library

brkalow:
## Description

Aligning on the `standardwebhooks` library for webhook verification.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/__tests__/webhooks.test.ts
+++ b/packages/backend/src/__tests__/webhooks.test.ts
@@ -1,39 +1,45 @@
+import { Webhook } from 'standardwebhooks';
 import { beforeEach, describe, expect, it } from 'vitest';
 
 import { verifyWebhook } from '../webhooks';
 
 describe('verifyWebhook', () => {
-  const mockSecret = 'test_signing_secret';
+  const mockSecret = 'whsec_' + Buffer.from('test_signing_secret_32_chars_long').toString('base64');
   const mockBody = JSON.stringify({ type: 'user.created', data: { id: 'user_123' } });
 
   beforeEach(() => {
     process.env.CLERK_WEBHOOK_SIGNING_SECRET = mockSecret;
   });
 
+  // Helper function to create a valid signature with Standard Webhooks
+  const createValidSignature = (id: string, timestamp: string, body: string) => {
+    const webhook = new Webhook(mockSecret);
+    // Create a signature using the Standard Webhooks library
+    return webhook.sign(id, new Date(parseInt(timestamp) * 1000), body);
+  };
+
   it('throws when required headers are missing', async () => {
     const mockRequest = new Request('https://clerk.com/webhooks', {
       method: 'POST',
       body: mockBody,
       headers: new Headers({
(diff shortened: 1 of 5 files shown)


packages/backend/src/webhooks.ts
```diff
@@ -56,39 +86,42 @@ export * from './api/resources/Webhooks';
...
-  if (svixSignature.split(' ').includes(constructedSignature)) {
-    return errorThrower.throw('Incoming webhook does not have a valid signature');
+    return {
+      type: payload.type,
+      object: 'event',
+      data: payload.data,
+    } as WebhookEvent;
+  } catch (e) {
+    return errorThrower.throw(`Unable to verify incoming webhook: ${e instanceof Error ? e.message : 'Unknown error'}`);
   }
```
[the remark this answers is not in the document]
brkalow: Given that this payload comes from our webhook provider and we've verified that the signature is accurate, we have an implicit guarantee of the shape here.

bratsos:
❓ According to [standardwebhooks](https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md#verifying-signatures) (TIL), it's a good practice to verify the webhook timestamp, making sure it's within a reasonable timeframe to prevent replay attacks. Do you think we should implement this here as well, or add this in a follow-up PR?

packages/backend/src/__tests__/webhooks.test.ts
```diff
@@ -1,39 +1,45 @@
...
       headers: new Headers({
-        // Missing svix-signature but with valid format for others
-        'svix-id': 'msg_123',
-        'svix-timestamp': '1614265330',
+        // Missing all required headers
       }),
     });
 
-    await expect(verifyWebhook(mockRequest)).rejects.toThrow('Missing required Svix headers: svix-signature');
+    await expect(verifyWebhook(mockRequest)).rejects.toThrow('Missing required webhook headers');
```
bratsos, line 30: 🙃 Maybe we can add a test for whitespace header values to handle this:

```js
headers: new Headers({
    'svix-id': '',        // Empty - would be caught
    'svix-timestamp': '   ', // Whitespace - would NOT be caught
    'svix-signature': 'v1,signature',
})
```

jacekradko (approved):

brkalow:
> According to [standardwebhooks](https://github.com/standard-webhooks/standard-webhooks/blob/main/spec/standard-webhooks.md#verifying-signatures) (TIL), it's a good practice to verify the webhook timestamp, making sure it's within a reasonable timeframe to prevent replay attacks. Do you think we should implement this here as well, or add this in a follow-up PR?

Good call! The `standardwebhooks` package takes care of this internally already 👍

packages/backend/src/__tests__/webhooks.test.ts
brkalow: Good idea, updated!

PR #6259, opened 2025-07-04
fix(clerk-js): Require email or phone when password field is visible

bratsos:
## Description

When password authentication is enabled alongside email/phone identifiers in progressive sign-up, the field requirement logic was not correctly determining which identifier should be marked as optional. This led to incorrect validation behavior where required fields might be incorrectly treated as optional.

### Solution

The core of the solution is the introduction of a new `determineRequiredCommunicationMethod` helper function in `signUpFormHelpers.ts`. This function centralizes the logic for determining which fields should be required based on the instance's sign-up attribute settings.

- `determineRequiredCommunicationMethod` encapsulates all rules for identifier requirements. It analyzes whether the password, email, phone, and username attributes are enabled, required, or used for the first factor.
- the new logic guarantees that at least one communication method becomes required
  - If only one communication method (e.g., email) is enabled, it is automatically made required
  - If both email and phone are enabled but optional, the logic now correctly enforces that they become required to prevent creating an account with no recovery method.
  - Respects the active identifier when the user can toggle between email and phone
- `getEmailAddressField` and `getPhoneNumberField` functions have been simplified to use the new centralized helper, making their behavior more predictable and easier to maintain

| Before  | After |
| ------------- | ------------- |
| <video src="https://github.com/user-attachments/assets/89d2508b-c774-4832-ba41-17303fde87b8.mp4">  | <video src="https://github.com/user-attachments/assets/edbbfa2b-a9a7-4a04-a901-3fbcea37cc46.mp4">|

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation and display of required and optional fields during sign-up when using password authentication with email or phone number, ensuring accurate feedback to users.

* **Tests**
  * Enhanced and expanded test coverage for sign-up field requirements, verifying correct accessibility attributes and UI labels for optional and required fields.

### DIFF
--- a/packages/clerk-js/src/ui/components/SignUp/signUpFormHelpers.ts
+++ b/packages/clerk-js/src/ui/components/SignUp/signUpFormHelpers.ts
@@ -192,7 +192,7 @@ function getEmailAddressField({
       return;
     }
 
-    const { emailShouldBeRequired } = determineRequiredCommunicationMethod(attributes);
+    const { emailShouldBeRequired } = determineRequiredIdentifier(attributes);
 
     return {
       required: emailShouldBeRequired,
@@ -236,7 +236,7 @@ function getPhoneNumberField({
       return;
     }
 
-    const { phoneShouldBeRequired } = determineRequiredCommunicationMethod(attributes);
+    const { phoneShouldBeRequired } = determineRequiredIdentifier(attributes);
 
     return {
       required: phoneShouldBeRequired,
@@ -324,15 +324,15 @@ const outcomePredicates: Record<Outcome, ((ctx: Context) => boolean)[]> = {
   mirrorServer: [
     // If password is not required, then field requirements are determined by the server.
     ctx => !ctx.passwordRequired,
-    // If any of the communication methods are already required by the server, then we don't need to do anything.
+    // If any of the identifiers are already required by the server, then we don't need to do anything.
     ctx => ctx.email.required || ctx.phone.required || (ctx.username.required && ctx.username.firstFactor),
   ],
   none: [
-    // If none of the communication methods are enabled, then none can be required.
+    // If none of the identifiers are enabled, then none can be required.
     ctx => !ctx.email.enabled && !ctx.phone.enabled && !ctx.username.enabled,
   ],
   email: [
-    // If email is the only enabled communication method, it should be required.
+    // If email is the only enabled identifier, it should be required.
     ctx => ctx.email.enabled && !ctx.phone.enabled && !ctx.username.enabled,
     // If email is enabled but not required, and phone is enabled and not required, then email should be required.
     ctx => ctx.email.enabled && !ctx.email.required && ctx.phone.enabled && !ctx.phone.required,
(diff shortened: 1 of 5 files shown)


panteliselef (approved):

packages/clerk-js/src/ui/components/SignUp/signUpFormHelpers.ts
```diff
@@ -178,6 +178,10 @@ function getEmailAddressField({
   isProgressiveSignUp,
 }: FieldDeterminationProps): Field | undefined {
   if (isProgressiveSignUp) {
+    const isPasswordVisible = Boolean(attributes.password?.enabled && attributes.password.required);
```
nikosdouvlis: super nit-picky comment, feel free to ignore :) 
```suggestion
    const isPasswordRequired = Boolean(attributes.password?.enabled && attributes.password.required);
```

packages/clerk-js/src/ui/components/SignUp/signUpFormHelpers.ts
```diff
@@ -300,3 +312,122 @@ function getGenericField(fieldKey: FieldKey, attributes: Partial<Attributes>): F
...
+      winner: 'mirrorServer',
+    },
+    {
+      when: ctx => !ctx.email.enabled && !ctx.phone.enabled,
+      winner: 'none',
+    },
+    {
+      when: ctx => ctx.emailOrPhoneLayout && ctx.activeField !== null,
+      winner: ctx.activeField !== null ? ctx.activeField : 'email',
+    },
```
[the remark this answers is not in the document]
bratsos: This is for ts compiler, `winner` is not inferring the type of `when` condition return, and we probably shouldn't complicate this function to implement this.

packages/clerk-js/src/ui/components/SignUp/signUpFormHelpers.ts
```diff
@@ -300,3 +304,124 @@ function getGenericField(fieldKey: FieldKey, attributes: Partial<Attributes>): F
...
+    // If username is required but not a first factor, and both email and phone are enabled, it should be required.
+    ctx => ctx.username.required && !ctx.username.firstFactor && ctx.email.enabled && ctx.phone.enabled,
+  ],
+};
+
+/**
+ * When password is required, we need to ensure at least one identifier
+ * (email, phone, or username) is also required
+ */
+export function determineRequiredCommunicationMethod(attributes: Partial<Attributes>): {
```
brkalow: nit: for consistency, I might consider naming this `determineRequiredIdentifier`. This avoids introducing another term "communication method" to describe these things.

bratsos: Makes sense https://github.com/clerk/javascript/pull/6259/commits/5ee7b6ce81bf3e5d9fd10bbd16500f6342a9be08

brkalow (approved):
👏

PR #6260, opened 2025-07-06
Update netlifyCacheHandler.ts

tavoyne:
## Description

Reference to `process` causes an error in environments where `process` is not defined. Even though this file seems to be related to Netlify only, it's causing issues in my Remix + Vite + Cloudflare app ("ReferenceError process is not defined" with the trace pointing to that file).

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/src/netlifyCacheHandler.ts
+++ b/packages/shared/src/netlifyCacheHandler.ts
@@ -33,7 +33,7 @@ export function handleNetlifyCacheInDevInstance({
   publishableKey: string;
 }) {
   const isOnNetlify =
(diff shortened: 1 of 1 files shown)


wobsoriano (changes requested):
Hello, thanks for raising this issue! We have an existing function that handles this, can we use that instead?

https://github.com/clerk/javascript/blob/main/packages/shared/src/getEnvVariable.ts

You can import it from `@clerk/shared/getEnvVariable`

tavoyne:
Added the requested change. I don't have time to add a changeset. This is a critical error on a commercial product I'm paying too much for. I'd appreciate it if you could take it from here!

PR #6273, opened 2025-07-09
fix(clerk-js): Navigate to tasks when switching sessions

LauraBeatris:
### Context

Nowadays, when switching to a pending session with `setActive`, it emits the state update to AIO components, such as `UserButton` and `OrganizationSwitcher`

The components update with the `pending` session, but it doesn't trigger a navigation to tasks for resolution. 

https://github.com/user-attachments/assets/ebaf432f-baac-429b-af7d-320bb4ebdd32

It also doesn't revalidate the server-side state. We currently only do so on `__experimental_navigateToTask`, but forgot to introduce on the internal handling for `pending` on `setActive`.

### Solution 

This PR introduces a check for multi-session to trigger navigation on `setActive`. It's not the best long-term solution for custom flows as navigating within `setActive` could be sudden, but we're going to tackle in a separate project, to solve all the changes when it comes to the UI state updates. 

https://github.com/user-attachments/assets/1368bedd-319e-4b54-a0a2-8d7feb1da6d4

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Automatically navigates to the tasks page when switching user sessions.
  * Added a new test suite to verify multi-session task flows, ensuring correct behavior when switching between user accounts.

* **Bug Fixes**
  * Improved session cleanup and reliability in sign-in test flows.
  * Fixed a variable reference in sign-up tests for more accurate URL checks.

* **Tests**
  * Added a method to facilitate account switching in end-to-end tests.
  * Removed a redundant SSO sign-up test and improved test structure and cleanup routines.

* **Documentation**
  * Updated navigation behavior documentation related to session switching.

* **Chores**
  * Increased allowed bundle size limit for core package.

### DIFF
--- a/integration/tests/session-tasks-sign-up.test.ts
+++ b/integration/tests/session-tasks-sign-up.test.ts
@@ -42,7 +42,9 @@ testAgainstRunningApps({ withEnv: [appConfigs.envs.withSessionTasks] })(
       expect(u.page.url()).toContain('tasks');
 
       // Resolves task
-      const fakeOrganization = u.services.organizations.createFakeOrganization();
+      const fakeOrganization = Object.assign(u.services.organizations.createFakeOrganization(), {
+        slug: u.services.organizations.createFakeOrganization().slug + '-with-sign-up',
+      });
       await u.po.sessionTask.resolveForceOrganizationSelectionTask(fakeOrganization);
       await u.po.expect.toHaveResolvedTask();
 
--- a/integration/tests/session-tasks-sign-up.test.ts
+++ b/integration/tests/session-tasks-sign-up.test.ts
@@ -1,11 +1,8 @@
 import { expect, test } from '@playwright/test';
 
-import { createClerkClient } from '@clerk/backend';
 import { appConfigs } from '../presets';
-import { instanceKeys } from '../presets/envs';
 import type { FakeUser } from '../testUtils';
 import { createTestUtils, testAgainstRunningApps } from '../testUtils';
-import { createUserService } from '../testUtils/usersService';
 
 testAgainstRunningApps({ withEnv: [appConfigs.envs.withSessionTasks] })(
   'session tasks after sign-up flow @nextjs',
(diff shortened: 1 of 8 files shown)


panteliselef (approved):
Can we add an e2e test, since the plan is to refactor/introduce chages in a follow up pr ?

LauraBeatris:
> Can we add an e2e test, since the plan is to refactor/introduce chages in a follow up pr ?

I'll try to modify the current app template to use `UserButton` so we can assert on switching sessions. 

Edit: Oh, it already has! I'll add assertions for it 😄

integration/tests/session-tasks-sign-up.test.ts
```diff
@@ -52,42 +49,5 @@ testAgainstRunningApps({ withEnv: [appConfigs.envs.withSessionTasks] })(
       // Navigates to after sign-up
       await u.page.waitForAppUrl('/');
     });
-
-    test('with sso, navigate to task on after sign-up', async ({ page, context }) => {
```
LauraBeatris: This was duplicated on the sign-in suite, they basically have the same assertions, so I've decided to remove it from here.

LauraBeatris:
I'm facing some challenges to get all of the E2E tests passing. Somehow, there's a correlation between them, in serial mode.

If I change the order, some pass and others not. Considering that we're cleaning users and sessions after each test, I wouldn't expect it to happen. 

Anyway, I'll continue to investigate and try to get the tests in a state where it's not flaky even if it passes.

packages/testing/src/playwright/unstable/page-objects/sessionTask.ts
```diff
@@ -18,6 +19,12 @@ export const createSessionTaskComponentPageObject = (testArgs: { page: EnhancedP
 
       await createOrganizationButton.click();
     },
+    waitForMounted: (task: SessionTask['key']) => {
+      switch (task) {
+        case 'org':
+          return page.waitForSelector('.cl-createOrganization-root', { state: 'attached' });
+      }
+    },
```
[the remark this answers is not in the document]
LauraBeatris: We don't want to do that as it'll break force-an-org tests when new tasks are introduced. I just want an abstraction to map selections per task enum.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1198,16 +1198,17 @@ export class Clerk implements ClerkInterface {
...
-      }
-
       /**
        * Hint to each framework, that the user will be signed out when `{session: null}` is provided.
        */
       await onBeforeSetActive(newSession === null ? 'sign-out' : undefined);
 
+      if (newSession?.status === 'pending') {
+        await this.#handlePendingSession(newSession);
+        await onAfterSetActive();
```
LauraBeatris, line 1219: We weren't revalidating the server state on `setActive` with a pending `session`, fixing it now.

PR #6277, opened 2025-07-09
chore(clerk-js): Extend `buildRedirectToHandshake` to accept search params

jfoshee:
Extend `buildRedirectToHandshake` to accept additional search parameters for improved observability

Add a query parameter for tracking delta between `session.iat` and `client_uat`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] other: observability

### DIFF
--- a/packages/backend/src/tokens/__tests__/handshake.test.ts
+++ b/packages/backend/src/tokens/__tests__/handshake.test.ts
@@ -179,6 +179,18 @@ describe('HandshakeService', () => {
       );
     });
 
+    it('should include additional search params when provided', () => {
+      const headers = handshakeService.buildRedirectToHandshake('test-reason', { iat_uat_delta: '100' });
+      const location = headers.get(constants.Headers.Location);
+      expect(location).toBeDefined();
+      if (!location) {
+        throw new Error('Location header is missing');
+      }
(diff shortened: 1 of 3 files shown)


packages/backend/src/tokens/__tests__/handshake.test.ts
```diff
@@ -179,6 +179,18 @@ describe('HandshakeService', () => {
       );
     });
 
+    it('should include additional search params when provided', () => {
+      const headers = handshakeService.buildRedirectToHandshake('test-reason', { iat_uat_delta: '100' });
+      const location = headers.get(constants.Headers.Location);
+      expect(location).toBeDefined();
+      if (!location) {
+        throw new Error('Location header is missing');
+      }
```
tmilewski:
```suggestion

```
`expect(location).toBeDefined();` **_should_** cut it.

jfoshee: Now I remember why I had that... Following the other tests and we aren't allowed to use `!` to assert non-null. I replaced it with null coalescing.

clerk-cookie:
Hello 👋

PR #6279, opened 2025-07-10
feat(backend): Add additional fields to SamlConnectionListParams type

jaredpiedt:
## Description
 

This adds `query`, `orderBy`, and `organizationId` to the `SamlConnectionListParams` type.

Documentation for these params: https://clerk.com/docs/reference/backend-api/tag/SAML-Connections

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Enhanced SAML connection listing with new options to filter by search query, sort results, and scope results to specific organizations.

* **Tests**
  * Added tests to verify correct handling of the new filtering, sorting, and scoping parameters when listing SAML connections.

### DIFF
--- a/packages/backend/src/api/endpoints/SamlConnectionApi.ts
+++ b/packages/backend/src/api/endpoints/SamlConnectionApi.ts
@@ -3,13 +3,18 @@ import type { SamlIdpSlug } from '@clerk/types';
 import { joinPaths } from '../../util/path';
 import type { SamlConnection } from '../resources';
 import { AbstractAPI } from './AbstractApi';
+import type { WithSign } from './util-types';
 
 const basePath = '/saml_connections';
 
 type SamlConnectionListParams = {
   limit?: number;
   offset?: number;
+  query?: string;
+  orderBy?: WithSign<'phone_number' | 'email_address' | 'created_at' | 'first_name' | 'last_name' | 'username'>;
+  organizationId?: WithSign<string>[];
 };
+
 type CreateSamlConnectionParams = {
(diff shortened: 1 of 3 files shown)


packages/backend/src/api/endpoints/SamlConnectionApi.ts
```diff
@@ -3,13 +3,18 @@ import type { SamlIdpSlug } from '@clerk/types';
...
+import type { WithSign } from './util-types';
 
 const basePath = '/saml_connections';
 
 type SamlConnectionListParams = {
   limit?: number;
   offset?: number;
+  query?: string;
+  orderBy?: WithSign<'phone_number' | 'email_address' | 'created_at' | 'first_name' | 'last_name' | 'username'>;
+  organizationId?: WithSign<string>[];
```
brkalow, line 15: I don't believe we need `WithSign` here:

```suggestion
  organizationId?: string[];
```

jaredpiedt: I'm happy to remove it, the documentation just made it seem like this was supported:

> the `+` and `-` can be prepended to the id, which denote whether the respective organization should be included or excluded from the result set

brkalow: Ahh good call thank you! You are correct 😄

brkalow (approved):

PR #6290, opened 2025-07-11
feat(clerk-js): Add `colorShadow` variable option

alexcarpenter:
## Description

This PR adds support for `colorShadow` variable usage. We've simplified the shadows to be derived from a single value instead of the custom rgb values.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/types/src/appearance.ts
```diff
@@ -731,8 +731,8 @@ export type Variables = {
    */
   colorRing?: CssColor;
   /**
-   * The color of the shadow used in card components.
-   * @default 'rgba(25, 28, 33, 0.20)'
```
panteliselef, line 735: Was this the case for only `shadows.input` and `shadows.focusRing` ? Or for none at all ?

panteliselef (approved):

nikosdouvlis (approved):

packages/types/src/appearance.ts
alexcarpenter: not sure I understand the question.

PR #6306, opened 2025-07-14
chore(e2e): Fix flaky tests for multi-session with tasks

LauraBeatris:
## Description

This PR temporarily skips the multi-session test for Next.js 14. I don't know the root cause yet, due to not having been able to run Cypress locally with debug mode against Next.js 14. 

We made tests on a Next.js 14 app using the latest `@clerk/nextjs` and the feature works correctly. 

For Next.js 15: Tests were flaky before due to a race condition where we were awaiting for the `SignIn` component to be mounted, by the time the client-side navigation to the`/tasks` route was happening. 

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Tests**
  * Improved reliability of session-switching navigation tests by adding explicit waits for UI state and navigation changes.
  * Activated a previously skipped test with adjustments for compatibility across different Next.js versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -301,9 +301,6 @@ jobs:
             next-version: '14'
           - test-name: 'nextjs'
             test-project: 'chrome'
-            next-version: '14.2.25'
-          - test-name: 'nextjs'
-            test-project: 'chrome'
             next-version: '15'
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -301,6 +301,9 @@ jobs:
             next-version: '14'
           - test-name: 'nextjs'
             test-project: 'chrome'
+            next-version: '14.2.25'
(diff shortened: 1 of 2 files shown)


LauraBeatris:
The multi-session test is failing on Next.js 14 inconsistently. I'm not sure if this is related with server-state. 

I'm struggling to get it running locally with the E2E pointing to Next.js 14 so I'll link the app-router playground and try to reproduce any bug related to it, otherwise, might be some race condition on the assertions still.

LauraBeatris:
I'm running after-auth with multi-session on an app running Next.js 14.2.25 and `@clerk/nextjs` 6.24.0, it's working just fine... 

I still couldn't run Cypress with the debug mode to see why the assertions are failing to navigate to tasks on 14.

PR #6319, opened 2025-07-15
chore(clerk-js): Refactor after-auth flows to keep navigation internally

LauraBeatris:
### Context 

This PR updates `experimental_navigateToTask` to `internal`. 

For custom flows using after-auth, it's not a good DX to "enforce" navigating after `setActive({ organization })` or any future frontend API interaction that updates the session, but rather, the UI should verify if there's a current task, and if any, then have control if either navigating or updating UI state in-place.

Our AIOs trigger navigation internally, but there might be custom flows with SPAs, where tasks can be displayed in place, without requiring a new page.

### New flow 

For "Disable Personal Workspaces", as it's the default behavior for the newest applications that enable orgs, the following must be done: 

```
setActive({ session }) 

if (session.currentTask) { 
  // router.push - to a page where it'll complete the task 
  // or update some UI state to display a component that completes the task 
}  else {
  router.push("/")
}
```

The above gives more control, and it's similar to what our progressive flows checks look like for custom flows when verifying statuses of sign-in and sign-up.

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Refactor**
  * Improved internal navigation handling after authentication for smoother user experience.
  * Renamed internal navigation method for clarity; no changes to user-facing behavior.
* **Chores**
  * Added documentation for a patch update.
* **Configuration**
  * Increased maximum allowed bundle size for signin-related files.


packages/clerk-js/src/core/clerk.ts
```diff
@@ -1368,20 +1358,6 @@ export class Clerk implements ClerkInterface {
...
-    /**
-     * Invoke the Next.js middleware to synchronize server and client state after resolving a session task.
-     * This ensures that any server-side logic depending on the session status (like middleware-based
-     * redirects or protected routes) correctly reflects the updated client authentication state.
-     */
-    const onAfterSetActive: SetActiveHook =
-      typeof window !== 'undefined' && typeof window.__unstable__onAfterSetActive === 'function'
-        ? window.__unstable__onAfterSetActive
-        : noop;
-    await onAfterSetActive();
```
LauraBeatris, line 1384: This is not needed anymore, as we already invalidate the server-side state on `setActive` and `__internal_navigateToTaskIfAvailable` would be called by AIOs right after.

panteliselef (approved):

PR #6323, opened 2025-07-15
fix(clerk-js): Do not trigger after-auth navigation from `useMultisessionActions`

LauraBeatris:
## Description

This previous [PR](https://github.com/clerk/javascript/pull/6319/files) introduced calling `__internal_navigateToTaskIfAvailable` after `setActive` when switching sessions via `useMultisessionActions`

This after-auth navigation must only be done if `__internal_navigateToTaskIfAvailable` is called with a `redirectUrlComplete`, which is passed from AIOs: 
https://github.com/clerk/javascript/blob/7f341dc7ed4982324e6dae59e803f42e794f2e2f/packages/clerk-js/src/ui/components/SessionTasks/index.tsx#L84-L87

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented automatic navigation after authentication unless a valid redirect URL is provided, improving navigation control post-login.

### DIFF
--- a/.changeset/deep-poems-sneeze.md
+++ b/.changeset/deep-poems-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Do not trigger after-auth navigation outside of AIOs
--- a/.changeset/deep-poems-sneeze.md
+++ b/.changeset/deep-poems-sneeze.md
@@ -2,4 +2,4 @@
 '@clerk/clerk-js': patch
 ---
 
-Do not trigger after-auth navigation outside of AIOs
+Do not trigger after-auth navigation from `useMultisessionActions`
--- a/.changeset/deep-poems-sneeze.md
+++ b/.changeset/deep-poems-sneeze.md
@@ -1,5 +1,7 @@
 ---
 '@clerk/clerk-js': patch
+'@clerk/clerk-react': patch
+'@clerk/types': minor
 ---
 
 Do not trigger after-auth navigation from `useMultisessionActions`
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -29,9 +29,9 @@ import type {
   AuthenticateWithGoogleOneTapParams,
   AuthenticateWithMetamaskParams,
   AuthenticateWithOKXWalletParams,
-  Clerk as ClerkInterface,
   ClerkAPIError,
   ClerkAuthenticateWithWeb3Params,
+  Clerk as ClerkInterface,
   ClerkOptions,
   ClientJSONSnapshot,
   ClientResource,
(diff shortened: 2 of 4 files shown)


.changeset/deep-poems-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Do not trigger after-auth navigation outside of AIOs
```
LauraBeatris, line 5: Do not trigger after-auth navigation from `useMultisessionActions`

LauraBeatris, line 5:
```suggestion
Do not trigger after-auth navigation from `useMultisessionActions`
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1348,15 +1350,16 @@ export class Clerk implements ClerkInterface {
       return;
     }
 
-    const tracker = createBeforeUnloadTracker(this.#options.standardBrowser);
-    const defaultRedirectUrlComplete = this.client?.signUp ? this.buildAfterSignUpUrl() : this.buildAfterSignInUrl();
```
LauraBeatris, line 1352: This default was previously placed when we thought of exposing this "navigateToTask" as a public API that could be called from custom flows.

packages/types/src/clerk.ts
```diff
@@ -2005,7 +2003,7 @@ export interface AuthenticateWithGoogleOneTapParams {
   legalAccepted?: boolean;
 }
 
-export interface NextTaskParams {
+export interface __internal_NavigateToTaskIfAvailableParams {
```
panteliselef, line 2006: Technically breaking, but you can decide, if it is best to deprecate or swap with the new name.

.changeset/deep-poems-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
```
panteliselef, line 2: Let's include all packages affected
```suggestion
'@clerk/clerk-js': patch
'@clerk/clerk-react': patch
'@clerk/types': minor
```

packages/types/src/clerk.ts
LauraBeatris: We haven't released it publicly previously, even behind `__experimental`, so I expect folks aren't using it. 

Regardless, to be safe, I'll just deprecate the previous interface and create a new one `__internal` one.

LauraBeatris:
Interesting why Typedoc wasn't catching it before...

CleanShot 2025-07-16 at 08 31 51

LauraBeatris:
Well, instead of deprecating `NextTaskParams`, I'll just remove it then. As I mentioned, we haven't offitionally released `navigateToTask` before as a experimental API, therefore I expect 0 clients have been using it...

PR #6340, opened 2025-07-16
fix(types): Ensure `unsafeMetadata` prop is only allowed in valid circumstances

tmilewski:
## Description

`unsafeMetadata` isn't passed along with page redirects and it's causing confusion as the types are incorrect for `<SignUpButton />`:

- If `mode == "modal`: `unsafeMetadata` is allowed.
- If `mode != "modal`: `unsafeMetadata` isn't allowed.

This PR accounts for those scenarios.

USER-2376

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Improved handling of custom metadata during the sign-up process, allowing metadata to be included specifically when using the modal sign-up flow.

* **Refactor**
  * Updated type definitions to enforce that custom metadata is only accepted when using the modal sign-up mode, enhancing type safety without altering the component interface.

### DIFF
--- a/packages/types/src/clerk.ts
+++ b/packages/types/src/clerk.ts
@@ -1900,19 +1900,22 @@ export interface HandleEmailLinkVerificationParams {
   onVerifiedOnOtherDevice?: () => void;
 }
 
-type ButtonPropsModal<T extends SignInProps | SignUpProps> = {
+type SignInButtonPropsModal = {
   mode: 'modal';
-  appearance?: T['appearance'];
-  unsafeMetadata?: T extends SignUpProps ? SignUpUnsafeMetadata : never;
+  appearance?: SignInProps['appearance'];
+};
+
+type SignUpButtonPropsModal = {
+  mode: 'modal';
+  appearance?: SignUpProps['appearance'];
+  unsafeMetadata?: SignUpUnsafeMetadata;
 };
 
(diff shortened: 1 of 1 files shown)


wobsoriano (approved):

packages/types/src/clerk.ts
```diff
@@ -1903,6 +1903,7 @@ export interface HandleEmailLinkVerificationParams {
 type ButtonPropsModal<T extends SignInProps | SignUpProps> = {
   mode: 'modal';
   appearance?: T['appearance'];
+  unsafeMetadata?: T extends SignUpProps ? SignUpUnsafeMetadata : never;
```
panteliselef, line 1906: Doesn't this mean that `SignInButtonProps` will also inherit the `unsafeMetadata` property ?

panteliselef (commented):
I think this affects the sign in button types as a side effect.

tmilewski:
> I don't think this works as intended image

@panteliselef Interesting as I tested the other day and sent out some screenshots of it working:

Screenshot 2025-07-17 at 10 00 50 AM

Regardless, I refactored the types, if you don't mind testing again.

PR #6344, opened 2025-07-17
feat(clerk-js): Add `cssLayerName` option to `experimental_createTheme`

alexcarpenter:
## Description

This enables shipping the shadcn theme without the users needing to specify the `cssLayerName` themselves.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -2727,9 +2727,22 @@ export class Clerk implements ClerkInterface {
   };
 
   #initOptions = (options?: ClerkOptions): ClerkOptions => {
+    // Extract baseTheme from appearance if it exists
+    let processedOptions = options ? { ...options } : {};
+    if (
+      processedOptions.appearance &&
+      typeof processedOptions.appearance === 'object' &&
+      'baseTheme' in processedOptions.appearance
+    ) {
+      processedOptions = {
+        ...processedOptions,
+        appearance: (processedOptions.appearance as any).baseTheme,
+      };
+    }
+
     return {
       ...defaultOptions,
-      ...options,
+      ...processedOptions,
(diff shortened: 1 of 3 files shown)


packages/themes/src/themes/shadcn.ts
```diff
@@ -1,6 +1,7 @@
 import { experimental_createTheme } from '../createTheme';
 
 export const shadcn = experimental_createTheme({
+  cssLayerName: 'components',
```
alexcarpenter, line 4: Since we know shadcn is using TW, we can specify the cssLayerName to be defined within components, relying on TW layer order specified.

panteliselef, line 4:
```suggestion
  cssLayerName: 'clerk-components',
```

panteliselef (approved):

packages/themes/src/themes/shadcn.ts
alexcarpenter: We need this to be `components` as this is the layer name tailwind uses. Which removes the need for them to need to specify the layer order. We can only do this because we know if they are using shadcn, that TW is used and they have a `components` layer already defined.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2727,9 +2727,73 @@ export class Clerk implements ClerkInterface {
...
+          appearance: {
+            ...appearance,
+            baseTheme: processedBaseTheme,
+            ...(finalCssLayerName && { cssLayerName: finalCssLayerName }),
+          },
+        };
+      }
+    }
+
+    console.table({ options, processedOptions });
```
tmilewski:
```suggestion
```

PR #6374, opened 2025-07-22
feat(clerk-js): Use SameSite=none by default for development instances

brkalow:
## Description

In development, we want to make sure that Clerk can work across a variety of environments, one being embedded previews in Gen AI tools (or other similar flavor of builders). When setting cookies in a cross origin iframe, we need to use `"SameSite=none`. Additionally, we want to make sure that the previews retain session state when opened in a new window ("full-screen"), so we can't rely on something like [CHIPS](https://developer.mozilla.org/en-US/docs/Web/Privacy/Guides/Privacy_sandbox/Partitioned_cookies).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/auth/cookies/clientUat.ts
+++ b/packages/clerk-js/src/core/auth/cookies/clientUat.ts
@@ -20,7 +20,13 @@ export type ClientUatCookieHandler = {
  * The cookie is used as hint from the Clerk Backend packages to identify
  * if the user is authenticated or not.
  */
-export const createClientUatCookie = (cookieSuffix: string): ClientUatCookieHandler => {
+export const createClientUatCookie = ({
+  cookieSuffix,
+  isProduction,
+}: {
+  cookieSuffix: string;
+  isProduction: boolean;
+}): ClientUatCookieHandler => {
   const clientUatCookie = createCookieHandler(CLIENT_UAT_COOKIE_NAME);
   const suffixedClientUatCookie = createCookieHandler(getSuffixedCookieName(CLIENT_UAT_COOKIE_NAME, cookieSuffix));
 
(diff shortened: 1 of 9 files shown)


packages/clerk-js/src/core/auth/cookies/session.ts
```diff
@@ -25,25 +29,29 @@ const getCookieAttributes = (): { sameSite: string; secure: boolean; partitioned
...
+    const { sameSite, secure, partitioned } = getCookieAttributes({ isProduction });
 
-    // If setting Partitioned to true, remove the existing session cookies.
-    // This is to avoid conflicts with the same cookie name without Partitioned attribute.
-    if (partitioned) {
-      remove();
-    }
+    // Remove existing cookies to prevent conflicts when SameSite attributes change
+    sessionCookie.remove();
+    suffixedSessionCookie.remove();
```
panteliselef: Don't we prefer `remove()` which also includes `attributes` ?

packages/clerk-js/src/core/auth/cookies/clientUat.ts
```diff
@@ -37,7 +43,13 @@ export const createClientUatCookie = (cookieSuffix: string): ClientUatCookieHand
      * Generally, this is handled by redirectWithAuth() being called and relying on the dev browser ID in the URL,
      * but if that isn't used we rely on this. In production, nothing is cross-domain and Lax is used when client_uat is set from FAPI.
      */
-    const sameSite = __BUILD_VARIANT_CHIPS__ ? 'None' : inCrossOriginIframe() ? 'None' : 'Strict';
+    const sameSite = __BUILD_VARIANT_CHIPS__
+      ? 'None'
+      : inCrossOriginIframe()
```
panteliselef, line 48: What was making `inCrossOriginIframe` to not detect the iframe usage inside the GenAI tools ?

brkalow: This was working, but one issue we ran into with this approach is that the app would set `SameSite=lax` cookies when the preview was opened in a new tab, and so cookies would not persist between the iframe'd preview and the preview opened in its own tab. In this case, it seemed preferable to always set `SameSite=None`.

packages/clerk-js/src/core/auth/cookies/devBrowser.ts
```diff
@@ -13,7 +12,7 @@ export type DevBrowserCookieHandler = {
 };
 
 const getCookieAttributes = (): { sameSite: string; secure: boolean } => {
-  const sameSite = inCrossOriginIframe() ? 'None' : 'Lax';
+  const sameSite = 'None';
   const secure = getSecureAttribute(sameSite);
   return { sameSite, secure };
 };
```
jacekradko, line 18: I wonder if this is even necessary now. When using  `SameSite=None` it's required to set `secure`

brkalow: @jacekradko true, but there are some quirks in Safari and localhost where secure can't be set, even though localhost is considered a secure environment

clerk-cookie:
Hello 👋

PR #6379, opened 2025-07-23
refactor(clerk-js): SocialButtons layout logic

alexcarpenter:
## Description

Replaces row distribution logic with a column count calculation for rendering social buttons. Removes unused Grid component and deletes distributeStrategiesIntoRows utility, simplifying the layout implementation.

Since all of the items are in a single flex container. They can move an item in the second row, into the first row with using CSS order property.

In the following example, Hubspot by default is the last item it the social buttons. Using `order: -1` I can move it to the first item. The top SocialButtons shows the flex implementation, and the bottom is the current where you can see Hubspot only moves to the first item in the second row.

Screenshot 2025-07-23 at 2 43 10 PM
 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Refactor**
  * Redesigned social sign-in buttons to use a flexible, responsive layout for improved consistency across devices.
  * Simplified button rendering by consolidating all buttons into a single flex container with dynamic column sizing.
* **Chores**
  * Removed outdated utility for distributing buttons into multiple rows.


.changeset/salty-jokes-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
```
alexcarpenter, line 2: maybe a minor?

clerk-cookie:
Hello 👋

PR #6384, opened 2025-07-24
fix(clerk-js): Display alert on plan details error

panteliselef:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
+++ b/packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
@@ -1,21 +1,70 @@
 import { useClerk } from '@clerk/shared/react';
-import type { __internal_PlanDetailsProps, CommercePlanResource, CommerceSubscriptionPlanPeriod } from '@clerk/types';
+import type {
+  __internal_PlanDetailsProps,
+  ClerkAPIResponseError,
+  CommercePlanResource,
+  CommerceSubscriptionPlanPeriod,
+} from '@clerk/types';
 import * as React from 'react';
 import { useMemo, useState } from 'react';
 import useSWR from 'swr';
 
+import { Alert } from '@/ui/elements/Alert';
 import { Avatar } from '@/ui/elements/Avatar';
 import { Drawer } from '@/ui/elements/Drawer';
 import { Switch } from '@/ui/elements/Switch';
 
 import { SubscriberTypeContext } from '../../contexts';
-import { Box, Col, descriptors, Flex, Heading, localizationKeys, Span, Spinner, Text } from '../../customizables';
+import {
+  Box,
+  Col,
+  descriptors,
+  Flex,
+  Heading,
+  localizationKeys,
+  Span,
+  Spinner,
+  Text,
+  useLocalizations,
+} from '../../customizables';
 
 export const PlanDetails = (props: __internal_PlanDetailsProps) => {
   return (
     <Drawer.Content>
       <PlanDetailsInternal {...props} />
     </Drawer.Content>
+  );
+};
+
+const BodyFiller = ({ children }: { children: React.ReactNode }) => {
+  return (
+    <Drawer.Body>
+      <Flex
+        align={'center'}
+        justify={'center'}
+        sx={t => ({
+          height: '100%',
+          padding: t.space.$4,
+          fontSize: t.fontSizes.$md,
+        })}
+      >
+        {children}
+      </Flex>
+    </Drawer.Body>
+  );
+};
+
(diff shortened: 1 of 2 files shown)


mauricioabreu (approved):

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
```diff
@@ -19,6 +36,38 @@ export const PlanDetails = (props: __internal_PlanDetailsProps) => {
   );
 };
 
+const BodyFiller = ({ children }: { children: React.ReactNode }) => {
+  return (
+    <Drawer.Body>
```
alexcarpenter: Can we apply the padding to the body? Instead of needing the nested flex container without a descriptor?

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
```diff
@@ -19,6 +36,38 @@ export const PlanDetails = (props: __internal_PlanDetailsProps) => {
...
+      <Flex
+        align={'center'}
+        justify={'center'}
+        sx={t => ({
+          height: '100%',
+          padding: t.space.$4,
+          fontSize: t.fontSizes.$md,
```
alexcarpenter: Why are we needing the fontSize defined here?

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
```diff
@@ -19,6 +36,38 @@ export const PlanDetails = (props: __internal_PlanDetailsProps) => {
...
+      </Flex>
+    </Drawer.Body>
+  );
+};
+
+const PlanDetailsError = ({ error }: { error: ClerkAPIResponseError }) => {
+  const { translateError } = useLocalizations();
+  return (
+    <BodyFiller>
+      <Alert
```
alexcarpenter, line 62: Can this be full width?

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
alexcarpenter: possibly add a new variant prop to body `padded`? to create a consistent padding usage

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
panteliselef: we don't. I removed it.

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
panteliselef: Do you want to force it ? If the error message is large enough it will occupy the necessary space.

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
panteliselef: the variant does not make much sense, I think it is better to use `space.$4` everywhere. There are other places using this long with other settings. I will increase complexity to try to support both.

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
```diff
@@ -38,18 +38,19 @@ export const PlanDetails = (props: __internal_PlanDetailsProps) => {
...
-        {children}
-      </Flex>
+    <Drawer.Body
+      sx={t => ({
+        display: 'flex',
+        flexDirection: 'column',
+        alignItems: 'center',
+        justifyContent: 'center',
+        flex: 1,
+        overflowY: 'auto',
```
alexcarpenter, line 48: body already handles a majority of these.

https://github.com/clerk/javascript/blob/79b4f032505314fce9fe88ed0929ed30742b393a/packages/clerk-js/src/ui/elements/Drawer.tsx#L351-L355

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
alexcarpenter: Got it, just kind of feels weird at the size imo. Not a blocker though.

Here is a another error rendering full width already.

Screenshot 2025-07-24 at 3 29 39 PM

aeliox (approved):

packages/clerk-js/src/ui/components/Plans/PlanDetails.tsx
panteliselef: Those are not inherited tho. If i removed the duplicate ones then this is the UI I end up with.
image

PR #6386, opened 2025-07-24
fix(clerk-js): Fix after-auth navigation when using server cache revalidation

LauraBeatris:
## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


LauraBeatris:
TODO for E2E tests: 
- Render a server component which renders `sessionStatus` from auth and assert that it is `pending` when it lands on the `select-organization` screen
- Not a must, but try to use something like `useId` on the screen to assert that it didn't got re-rendered or unmounted

LauraBeatris:
Opening this PR to generate a snapshot so I can test with https://force-an-org-example.clerkstage.dev as well

PR #6389, opened 2025-07-24
fix(clerk-js): Remove cache revalidation hooks from pending session handling

LauraBeatris:
## Description

This PR removes `onBeforeSetActive` and `onAfterSetActive` from being executed on the pending session handling within `setActive` 

This was leading to multiple issues of `SignIn` and `SignUp` getting unmounted, since the server would refetch/refresh components and leave the `session` with an `undefined` state, causing `withCoreSessionGuard` to unmount AIOs. 

We didn't have these issues outside of after-auth flows (meaning when `pending` status doesn't exist and it goes from an empty session to `active`) since `setActive` would always execute Next.js router navigation to a page outside of AIOs. 

---

**Tradeoffs**

Since those cache revalidation hooks aren't executed, it's possible that RSCs and middleware logic that relies on the pending session, doesn't get trigger once the AIO is on the `select-organization` screen. Examples could be trying to extract the `userId` with `auth({ treatPendingAsSignedOut: false })` 

We think the tradeoff is worth tho, since atm, there's no clear indication that this would make sense while using AIOs in an after-auth flow. 

If the developer wants to rely on the pending session granularly as a signed-in state on server-side logic, they can do so with custom pages by ejecting out of AIOs with `<ClerkProvider taskUrls={{ 'select-organization': '/my-server-rendered-page' />`

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved issues causing component unmounting in SignIn and SignUp All-In-One flows after authentication.
  * Improved session handling to prevent unmounting problems during post-authentication processes.

### DIFF
--- a/.changeset/full-impalas-wonder.md
+++ b/.changeset/full-impalas-wonder.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Remove cache revalidation hooks from pending session handling. This fixes unmounting issues from `SignIn` and `SignUp` AIOs during after-auth flows.
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -1225,16 +1225,15 @@ export class Clerk implements ClerkInterface {
         }
       }
 
+      if (newSession?.status === 'pending') {
+        await this.#handlePendingSession(newSession);
+        return;
+      }
+
       /**
        * Hint to each framework, that the user will be signed out when `{session: null}` is provided.
        */
       await onBeforeSetActive(newSession === null ? 'sign-out' : undefined);
-
-      if (newSession?.status === 'pending') {
-        await this.#handlePendingSession(newSession);
-        await onAfterSetActive();
-        return;
-      }
 
       //1. setLastActiveSession to passed user session (add a param).
       //   Note that this will also update the session's active organization


NicolasLopes7 (approved):
The changes looks straightforward enough 👍

iagodahlem (approved):
Couldn't test it yet, having troubles to setup playgrounds, will try it again later and give this one a try too!

> Examples could be trying to extract the userId with auth({ treatPendingAsSignedOut: false })

Maybe we should mention that in the docs as well.

LauraBeatris:
> Couldn't test it yet, having troubles to setup playgrounds, will try it again later and give this one a try too!
> 
> > Examples could be trying to extract the userId with auth({ treatPendingAsSignedOut: false })
> 
> Maybe we should mention that in the docs as well.

@iagodahlem We already have `treatPendingAsSignedOut` on `auth` docs but haven't introduced it on a specific guide with actual use case... I'd say, I'm curious on how developers would use it as well. 

My idea is that it'd be mostly used for API routes or server actions. The majority of the B2B app logics will require to treat pending as signed-out to make sure orgs are defined so I can imagine that `treatPendingAsSignedOut: false` would be quite rare.

iagodahlem:
> > Couldn't test it yet, having troubles to setup playgrounds, will try it again later and give this one a try too!
> > > Examples could be trying to extract the userId with auth({ treatPendingAsSignedOut: false })
> > 
> > 
> > Maybe we should mention that in the docs as well.
> 
> @iagodahlem We already have `treatPendingAsSignedOut` on `auth` docs but haven't introduced it on a specific guide with actual use case... I'd say, I'm curious on how developers would use it as well.
> 
> My idea is that it'd be mostly used for API routes or server actions. The majority of the B2B app logics will require to treat pending as signed-out to make sure orgs are defined so I can imagine that `treatPendingAsSignedOut: false` would be quite rare.

Oh, I think I get it, makes sense to me, thanks for explaining it @LauraBeatris! 🙌

PR #6430, opened 2025-07-29
fix(clerk-js): Fix SSO callback for after-auth custom flows

LauraBeatris:
### Context

After-auth flows always send the `redirect_url` as `redirect_url_complete` in order to execute `handleRedirectCallback` so that the SDK can control on navigating back to the pending tasks. 

We weren't dealing with the following cases:`
- The navigation to tasks was only included within `SSOCallback` and didn't cover custom flows that execute `handleRedirectCallback` directly 
- If the session is `active`, then `handleRedirectCallback` should redirect to the `redirect_url_complete` from the initial flow 
- For transfer flows, on `setActive`, it wasn't navigating to tasks.

### Proposed fixes 

- Moves task navigation to `handleRedirectCallback`
- Persist `redirect_url_complete` as `?redirect_url` param 

### Flows 

Refer to the navigations below (I haven't implemented a custom spinner component for `handleRedirectCallback`

With `active` session and `__internal_hasAfterAfterFlows`:

https://github.com/user-attachments/assets/e79f05c0-edc7-4d11-af01-d6af272cdaaa

With `pending` session + transfer flow:

https://github.com/user-attachments/assets/f45c9b40-7fc9-4a38-b9ef-51d547e40451

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of single sign-on (SSO) callback flows, ensuring smoother navigation after authentication, especially when custom post-authentication tasks are present.

* **Tests**
  * Added new test cases to verify correct redirection behavior for pending and completed session tasks after SSO authentication.

* **Documentation**
  * Updated session-related documentation to clarify that sessions may be pending, not only active.

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -400,6 +400,7 @@ export class Clerk implements ClerkInterface {
     });
     this.#publicEventBus.emit(clerkEvents.Status, 'loading');
     this.#publicEventBus.prioritizedOn(clerkEvents.Status, s => (this.#status = s));
+
     // This line is used for the piggy-backing mechanism
     BaseResource.clerk = this;
   }
@@ -1878,10 +1879,11 @@ export class Clerk implements ClerkInterface {
     };
 
     if (si.status === 'complete') {
-      return this.setActive({
+      await this.setActive({
         session: si.sessionId,
         redirectUrl: redirectUrls.getAfterSignInUrl(),
       });
+      return this.__internal_navigateToTaskIfAvailable();
     }
 
     const userExistsButNeedsToSignIn =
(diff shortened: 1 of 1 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -2368,8 +2375,8 @@ export class Clerk implements ClerkInterface {
     this.#authService = await AuthCookieService.create(
       this,
       this.#fapiClient,
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      this.#instanceType!,
+
+      this.#instanceType,
```
LauraBeatris: I have to check why my linter is making those changes on editor save... it ends up leading to a lot of build errors.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1953,10 +1954,11 @@ export class Clerk implements ClerkInterface {
     }
 
     if (su.status === 'complete') {
-      return this.setActive({
+      await this.setActive({
         session: su.sessionId,
         redirectUrl: redirectUrls.getAfterSignUpUrl(),
       });
+      return this.__internal_navigateToTaskIfAvailable({ redirectUrlComplete: redirectUrls.getAfterSignUpUrl() });
```
LauraBeatris: To be clear, this doesn't impact custom flows, and developers shouldn't have to do it. 

On their custom pages, `Clerk.handleRedirectCallback` will always navigate to the pending task route.

PR #6452, opened 2025-07-31
feat(clerk-js): Introduce debugLogger

jacekradko:
## Description

Create a debugLogger module that will be lazy instantiated when environment shows we are in a client debug mode. 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -197,12 +197,13 @@ export class Clerk implements ClerkInterface {
 
   public static version: string = __PKG_VERSION__;
   public static sdkMetadata: SDKMetadata = {
-    name: __PKG_NAME__,
+    name: '@clerk/clerk-js',
     version: __PKG_VERSION__,
-    environment: process.env.NODE_ENV || 'production',
   };
+
   private static _billing: CommerceBillingNamespace;
   private static _apiKeys: APIKeysNamespace;
+  private static _debugLoggerInitialized = false;
   private _checkout: ClerkInterface['__experimental_checkout'] | undefined;
 
   public client: ClientResource | undefined;
@@ -211,6 +212,7 @@ export class Clerk implements ClerkInterface {
   public user: UserResource | null | undefined;
   public __internal_country?: string | null;
   public telemetry: TelemetryCollector | undefined;
+  public debugLogger?: any; // Will be typed properly after dynamic import
 
   protected internal_last_error: ClerkAPIError | null = null;
   // converted to protected environment to support `updateEnvironment` type assertion
@@ -1480,6 +1482,7 @@ export class Clerk implements ClerkInterface {
     const customNavigate =
       options?.replace && this.#options.routerReplace ? this.#options.routerReplace : this.#options.routerPush;
 
+    this.debugLogger?.info(`Clerk is navigating to: ${toURL}`);
     if (this.#options.routerDebug) {
       console.log(`Clerk is navigating to: ${toURL}`);
     }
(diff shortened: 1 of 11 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -2233,6 +2261,23 @@ export class Clerk implements ClerkInterface {
 
   public updateEnvironment(environment: EnvironmentResource): asserts this is { environment: EnvironmentResource } {
     this.environment = environment;
+
+    // Initialize debug module if client_debug_mode is enabled
+    if (environment.clientDebugMode) {
+      this.#initializeDebugModule();
+    }
```
brkalow: this should probably be in `load()` after we fetch environment

packages/clerk-js/src/core/clerk.ts
```diff
@@ -212,6 +238,7 @@ export class Clerk implements ClerkInterface {
   public __internal_country?: string | null;
   public telemetry: TelemetryCollector | undefined;
   public readonly __internal_state: State = new State();
+  public debugLogger?: DebugLoggerInterface;
```
brkalow, line 222: I would probably want this to be an importable singleton, instead of on the core Clerk class. If we do want to expose it via `Clerk` for some reason we can add this later, but I like not updating the public interface here if we don't need to.

Preferred:
```
import { debugLogger } from '@/utils/debug';
```

packages/shared/src/telemetry/collector.ts
```diff
@@ -206,11 +297,49 @@ export class TelemetryCollector implements TelemetryCollectorInterface {
...
     } else {
       // This is not an ideal solution, but it at least waits until the next tick
       this.#pendingFlush = setTimeout(() => {
         this.#flush();
+        this.#pendingFlush = null;
+      }, 0);
+    }
+  }
+
+  #scheduleLogFlush(): void {
```
brkalow: per our conversation: consider combining the event buffer here, we can have logic in flush to use a different endpoint depending on what type of event it is (telemetry event or debug log)

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2893,4 +2938,13 @@ export class Clerk implements ClerkInterface {
...
   }
+
+  /**
+   * @internal
+   */
+  public static async __internal_resetDebugLogger(): Promise<void> {
+    // This method is now handled by the debug module itself
+    const { __internal_resetDebugLogger } = await import('./modules/debug');
+    __internal_resetDebugLogger();
+  }
```
brkalow: I think we can remove this and rely on the method from the module?

brkalow (approved):
approving, but would be cool to see it importable 👀

packages/shared/src/telemetry/collector.ts
jacekradko: Done

packages/clerk-js/src/core/modules/debug/logger.ts
```diff
@@ -0,0 +1,185 @@
...
+    if (!this.shouldLogLevel(level)) {
+      return;
+    }
+
+    if (!this.shouldLogFilters(level, message, source)) {
+      return;
+    }
+
+    const entry: DebugLogEntry = {
+      id: generateUuid(),
```
brkalow: consider moving the ID generation to the API, I'm not sure we need it here, and it avoids uuid generation complexity on the client if we drop it

jacekradko: Yeah, I removed it.

PR #6453, opened 2025-08-01
feat(clerk-js): Server-side pagination of API keys component

wobsoriano:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/ApiKeys/ApiKeys.tsx
+++ b/packages/clerk-js/src/ui/components/ApiKeys/ApiKeys.tsx
@@ -59,7 +59,7 @@ export const APIKeysPage = ({ subject, perPage, revokeModalRoot }: APIKeysPagePr
     startingRow,
     endingRow,
     cacheKey,
-  } = useApiKeys({ subject, perPage, enabled: isOrg ? canReadAPIKeys : true });
+  } = useApiKeys({ subject, pageSize: perPage, enabled: isOrg ? canReadAPIKeys : true });
   const card = useCardState();
   const { trigger: createApiKey, isMutating } = useSWRMutation(cacheKey, (_, { arg }: { arg: CreateAPIKeyParams }) =>
     clerk.apiKeys.create(arg),
--- a/packages/clerk-js/src/ui/components/ApiKeys/ApiKeys.tsx
+++ b/packages/clerk-js/src/ui/components/ApiKeys/ApiKeys.tsx
@@ -59,7 +59,7 @@ export const APIKeysPage = ({ subject, perPage, revokeModalRoot }: APIKeysPagePr
     startingRow,
     endingRow,
     cacheKey,
-  } = useApiKeys({ subject, pageSize: perPage, enabled: isOrg ? canReadAPIKeys : true });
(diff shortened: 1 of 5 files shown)


clerk-cookie:
Hello 👋

packages/clerk-js/src/ui/components/ApiKeys/__tests__/ApiKeys.spec.tsx
wobsoriano, line 1: Added unit tests for the simple stuff but most are handled in the e2e test

panteliselef (commented):
I'd avoid doing this inside the component. Chances are that we want to expose `useApiKeys` sooner or later. We should use the strategy similar to other paginated hooks. Let's use `usePagesOrInfinite` and add any missing functionality there.

wobsoriano:
> I'd avoid doing this inside the component. Chances are that we want to expose `useApiKeys` sooner or later. We should use the strategy similar to other paginated hooks. Let's use `usePagesOrInfinite` and add any missing functionality there.

Yeah this is a good suggestion! Thank you, refactoring it 🫡

packages/clerk-js/src/ui/components/ApiKeys/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+  fetchPage: (page: number) => void;
+};
+
+/**
+ * Hook that manages pagination logic for API keys:
+ * - Resets to page 1 when query changes
+ * - Adjusts page when current page exceeds available pages (e.g., after deletion)
+ * - Provides cache invalidation function for mutations
+ */
+export const useAPIKeysPagination = ({
```
tmilewski, line 19: 🙃 Let's look into consistency between `API` & `Api`. It seems they're used interchangeably and it'll be good to settle on one.

packages/clerk-js/src/ui/components/ApiKeys/ApiKeys.tsx
```diff
@@ -183,19 +205,17 @@ export const APIKeysPage = ({ subject, perPage, revokeModalRoot }: APIKeysPagePr
         elementDescriptor={descriptors.apiKeysTable}
         canManageAPIKeys={(isOrg && canManageAPIKeys) || !isOrg}
       />
-      {itemCount > (perPage ?? 5) && (
+      {(itemCount > (perPage ?? 5) || pageCount > 1) && (
```
tmilewski: 🙃 Seems the page size in `useApiKeys` is 10. Maybe we set a const there and reference it here?

packages/clerk-js/src/ui/components/ApiKeys/utils.ts
wobsoriano: we discussed before and went with `API` so I went with that as well! thanks

packages/clerk-js/src/ui/components/ApiKeys/ApiKeys.tsx
wobsoriano: updated to use `10` to have consistency with other hooks!

tmilewski (approved):

PR #6511, opened 2025-08-10
chore(clerk-js): Trigger Next.js hooks on session status transition to `pending`

LauraBeatris:
## Description

This PR introduces a conditional on client update, to detect a session transition from `active` to `pending`, that occurs when a instance task gets "unsatisfied" - this could happen, for the "Allow personal workspace" setting, when an app owner removes a user from an organization, or due to deprovisioning event on an SSO Identity Provider.

Once the transition gets detected, then hooks are executed to revalidate the server auth state. This makes sure that existing page protections on `clerkMiddleware` get triggered, such in the example below: 

https://github.com/user-attachments/assets/4b0ae4b7-fc02-4d6c-b5b5-92c762aa71bf

```ts
export default clerkMiddleware(
  async (auth, req) => {
    if (isProtectedRoute(req)) {
     // Re-run for pending session, redirects to `/sign-in/tasks`
      await auth.protect();
    }

    return NextResponse.next();
  }
);
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Session handling now triggers lifecycle hooks when a session moves from active to pending, improving authentication context updates and Next.js integration reliability.

* **Tests**
  * Added tests ensuring session-status transitions invoke the expected lifecycle hooks once and in the correct order.

### DIFF
--- a/.changeset/mean-jobs-stare.md
+++ b/.changeset/mean-jobs-stare.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Trigger Next.js hooks on session status transition from `active` to `pending` to update authentication context state


nikosdouvlis (approved):

iagodahlem (approved):

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2258,6 +2258,23 @@ export class Clerk implements ClerkInterface {
...
+            ? window.__unstable__onAfterSetActive
+            : noop;
+
+        // Execute hooks to update server authentication context and trigger
+        // page protections in clerkMiddleware or server components
+        void onBeforeSetActive()?.then?.(() => void onAfterSetActive());
+      }
+
       // Note: this might set this.session to null
       this.#setAccessors(session);
```
Ephem, line 2279: @LauraBeatris I found this PR when trying to understand the `updateClient` function and wanted to ask, is there a reason `onAfterSetActive` runs _before_ the `#setAccessors` call? That means the `router.refresh()` runs when `clerk.client` has been updated, but not necessarily `clerk.session` which I'm not sure if it's intentional or not?

Nice catch fixing this btw!

PR #6580, opened 2025-08-20
fix(clerk-js): Pass query params to buildAfterSignInUrl

iagodahlem:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/components/SessionTasks/index.tsx
```diff
@@ -71,9 +71,14 @@ export const SessionTasks = withCardStateProvider(() => {
   const signInContext = useContext(SignInContext);
   const signUpContext = useContext(SignUpContext);
   const currentTaskContainer = useRef<HTMLDivElement>(null);
+  const { queryParams } = useRouter();
 
   const redirectUrlComplete =
-    signInContext?.afterSignInUrl ?? signUpContext?.afterSignUpUrl ?? clerk?.buildAfterSignInUrl();
+    signInContext?.afterSignInUrl ??
```
LauraBeatris: `SignInContext` and `SignUpContext` seems to be returning `afterSignInUrl/SignUpUrl` as undefined in those cases

We should get the redirect URL from either `SignIn` or `SignUp` as a prop, instead of building it here from `clerk`, since their React context already has the logic to extract from query params: 
https://github.com/clerk/javascript/blob/2fa2a4193760dbea6dbd0bbe1315470b0c07488b/packages/clerk-js/src/ui/contexts/components/SignUp.ts#L63-L78

packages/clerk-js/src/ui/components/SessionTasks/index.tsx
```diff
@@ -62,19 +61,18 @@ function SessionTasksRoutes(): JSX.Element {
...
+
 /**
  * @internal
  */
-export const SessionTasks = withCardStateProvider(() => {
+export const SessionTasks = withCardStateProvider(({ redirectUrlComplete }: SessionTasksProps) => {
   const clerk = useClerk();
   const { navigate } = useRouter();
-  const signInContext = useContext(SignInContext);
-  const signUpContext = useContext(SignUpContext);
```
brkalow, line 72: 🤔 so the context value was undefined here? This is surprising, as the only change we've made is to pass the value from the context as a prop instead

LauraBeatris: `SignInContext` doesn't have `afterSignInUrl`, tho it's wrong that it had the type definition on it. 

`useSignInContext` and `useSignUpContext` that have the logic to extract query params, which relies on being called within `SignInContext/SignUpContext`

LauraBeatris: https://github.com/clerk/javascript/blob/2fa2a4193760dbea6dbd0bbe1315470b0c07488b/packages/clerk-js/src/ui/contexts/components/SignUp.ts#L63-L78

LauraBeatris: It's not possible to call `useSignInContext` or `useSignUpContext` within `SessionTasks` directly since it's both rendered within `SignIn/SignUp`

Another option would be to create separate ones, such as `SignInSessionTasks` or `SignUpSessionTasks`, but for me would be the same as passing `redirectUrlComplete` as a prop

packages/clerk-js/src/ui/components/SignIn/index.tsx
```diff
@@ -136,15 +136,15 @@ function SignInRoutes(): JSX.Element {
               </Route>
             </Route>
             <Route path='tasks'>
-              <LazySessionTasks />
+              <LazySessionTasks redirectUrlComplete={signInContext.afterSignUpUrl} />
```
panteliselef, line 139: Potentially what I would do is create a new `<SessionTaskProvider>` and map/pass properties of `signInContext` and `signUpContext` and later simply consume it via a `useSessionTaskContext()`

PR #6592, opened 2025-08-20
fix(clerk-js): Show password in Security tab if password is enabled

dmoerner:
## Description

The Clerk Dashboard now has improved support for enabling an instance configuration with password enabled, but not required. Such a configuration makes passwords optional on sign-up, allows users to add them to accounts, and allows users to use them for sign-in. If the instance is configured with password enabled but not required, we need to still show the password in the user's Security tab so they can set or update their password.

This is implemented by adding a `UserSettings.instanceIsPasswordEnabled()` property which is true if password is enabled on an instance, and using this to determine if we should show the password card on the Security tab.

Fixes USER-3245. Tested with unit testing and with local Clerk.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


dmoerner:
Closing and redoing this on a fresh branch. Since we want to also update instant sign in, we can just modify `instanceIsPasswordBased` and do a patch changeset. That's a simpler change and we don't need to retain this history.

PR #6612, opened 2025-08-22
chore(tanstack-react-start,nuxt): Pass `treatPendingAsSignedOut` to auth functions

wobsoriano:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Auth helpers now accept optional per-call settings (e.g., treatPendingAsSignedOut) for finer pending-session handling.
  * Per-call token acceptance can be specified when calling auth/getAuth.

* **Improvements**
  * Unified options support across server frameworks and examples; middleware/context auth forwards options.

* **Compatibility**
  * Public API shapes remain compatible; existing calls continue to work without providing options.

### DIFF
--- a/.changeset/ninety-wombats-think.md
+++ b/.changeset/ninety-wombats-think.md
@@ -15,6 +15,7 @@ const authStateFn = createServerFn({ method: 'GET' }).handler(async () => {
   return { userId }
 })
 ```
+
 Nuxt
 
 ```ts
--- a/.changeset/ninety-wombats-think.md
+++ b/.changeset/ninety-wombats-think.md
@@ -3,14 +3,14 @@
 "@clerk/tanstack-react-start": patch
 ---
 
-Allows passing of `treatPendingAsSignedOut` to auth functions:
+Allows passing of [`treatPendingAsSignedOut`](https://clerk.com/docs/authentication/configuration/session-tasks#session-handling) to auth functions:
 
 TanStack Start
 
 ```ts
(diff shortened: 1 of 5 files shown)


.changeset/ninety-wombats-think.md
```diff
@@ -0,0 +1,26 @@
+---
+"@clerk/nuxt": patch
+"@clerk/tanstack-react-start": patch
+---
+
+Allows passing of `treatPendingAsSignedOut` to auth functions:
```
LauraBeatris: We can link to our docs that give more information around `treatPendingAsSignedOut`: https://clerk.com/docs/authentication/configuration/session-tasks#session-handling

.changeset/ninety-wombats-think.md
```diff
@@ -0,0 +1,26 @@
...
+
+TanStack Start
+
+```ts
+const authStateFn = createServerFn({ method: 'GET' }).handler(async () => {
+  const request = getWebRequest()
+  const { userId } = await getAuth(request, { treatPendingAsSignedOut: true })
```
LauraBeatris: `treatPendingAsSignedOut: true` is already the default, so we could pass `treatPendingAsSignedOut: false` here 

```ts
// Both `active` and `pending` sessions will be treated as authenticated when `treatPendingAsSignedOut` is `false` - `userId` will be returned
return { userId } 
```

PR #6614, opened 2025-08-22
feat(clerk-js,clerk-react,types): Signal transfer support

dstaley:
## Description

This PR implements sign-in/sign-up transfer support in our Signal implementation, which is not yet intended for public use. It also fixes two smaller issues, one where `useSignUpSignal` was not exported, and another where sign-up errors were sent to the incorrect Signal.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- New Features
  - Sign-in and sign-up now indicate when an account/session is transferable and can expose an existing session ID.
  - Optional transfer flag when initiating sign-in or sign-up.
  - Sign-up adds SSO flow with captcha handling and redirect support.
  - Experimental React hook for sign-up signals added.

- Bug Fixes
  - Sign-up error handling now signals errors more directly.

- Chores
  - Package versions bumped; bundle size threshold adjusted.

### DIFF
--- a/packages/react/src/stateProxy.ts
+++ b/packages/react/src/stateProxy.ts
@@ -43,6 +43,7 @@ export class StateProxy implements State {
       signIn: {
         status: 'needs_identifier' as const,
         availableStrategies: [],
+        isTransferable: false,
 
         create: this.gateMethod(target, 'create'),
         password: this.gateMethod(target, 'password'),
@@ -68,7 +69,10 @@ export class StateProxy implements State {
       signUp: {
         status: 'missing_requirements' as const,
         unverifiedFields: [],
+        isTransferable: false,
 
+        create: this.gateMethod(target, 'create'),
+        sso: this.gateMethod(target, 'sso'),
         password: this.gateMethod(target, 'password'),
         finalize: this.gateMethod(target, 'finalize'),
 


packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -488,6 +488,26 @@ class SignUpFuture implements SignUpFutureResource {
     return this.resource.unverifiedFields;
   }
 
+  get isTransferable() {
+    return (
+      this.resource.verifications.externalAccount.status === 'transferable' &&
+      this.resource.verifications.externalAccount.error?.code === 'external_account_exists'
```
brkalow, line 495: I would expect the `transferable` status is all we need to check for here.

dstaley: happy to take your word for it, but this is a direct copy-paste of the existing logic https://github.com/clerk/javascript/blob/main/packages/clerk-js/src/core/clerk.ts#L1935-L1936

brkalow: makes sense, just thinking from first principles here 😅 I'm not sure why the extra check exists

dstaley: I'll add a TODO note to it that we can likely drop the extra check!

PR #6635, opened 2025-08-25
fix(backend): Prevent excessive handshakes

brkalow:
## Description

- Prevent a handshake from triggering in development after a handshake is resolved with a `__clerk_db_jwt` parameter
- Reduce `max-age` of the redirect counter cookie to `2` to reduce false-positives during redirect detection
- Update the org url sync logic to only consider a redirect loop when 3 or more redirects happen in quick succession, instead of just one

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- Bug Fixes
  - Reduced unintended activation handshake triggers in development by cleaning up redirect URLs during dev flows.
  - Allowed up to three redirects before treating the sequence as a loop, reducing false positives.
  - Shortened redirect-loop cookie lifespan for faster recovery during development.

- Chores
  - Added a changeset to publish a patch release for the backend package.

### DIFF
--- a/packages/backend/src/tokens/handshake.ts
+++ b/packages/backend/src/tokens/handshake.ts
@@ -222,6 +222,7 @@ export class HandshakeService {
       const newUrl = new URL(this.authenticateContext.clerkUrl);
       newUrl.searchParams.delete(constants.QueryParameters.Handshake);
       newUrl.searchParams.delete(constants.QueryParameters.HandshakeHelp);
+      newUrl.searchParams.delete(constants.QueryParameters.DevBrowser);
       headers.append(constants.Headers.Location, newUrl.toString());
       headers.set(constants.Headers.CacheControl, 'no-store');
     }
@@ -323,7 +324,7 @@ ${developmentError.getFullMessage()}`,
 
     const newCounterValue = this.authenticateContext.handshakeRedirectLoopCounter + 1;
     const cookieName = constants.Cookies.RedirectCount;
-    headers.append('Set-Cookie', `${cookieName}=${newCounterValue}; SameSite=Lax; HttpOnly; Max-Age=3`);
+    headers.append('Set-Cookie', `${cookieName}=${newCounterValue}; SameSite=Lax; HttpOnly; Max-Age=2`);
     return false;
   }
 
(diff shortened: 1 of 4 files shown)


packages/backend/src/tokens/handshake.ts
```diff
@@ -323,7 +324,7 @@ ${developmentError.getFullMessage()}`,
 
     const newCounterValue = this.authenticateContext.handshakeRedirectLoopCounter + 1;
     const cookieName = constants.Cookies.RedirectCount;
-    headers.append('Set-Cookie', `${cookieName}=${newCounterValue}; SameSite=Lax; HttpOnly; Max-Age=3`);
+    headers.append('Set-Cookie', `${cookieName}=${newCounterValue}; SameSite=Lax; HttpOnly; Max-Age=2`);
```
brkalow, line 327: 3 was an arbitrary choice here, I'm reducing to 2 to mitigate the case where someone switches between orgs quickly when using org url sync

packages/backend/src/tokens/request.ts
```diff
@@ -386,7 +386,7 @@ export const authenticateRequest: AuthenticateRequest = (async (
     if (!mustActivate) {
       return null;
     }
-    if (authenticateContext.handshakeRedirectLoopCounter > 0) {
+    if (authenticateContext.handshakeRedirectLoopCounter >= 3) {
```
brkalow, line 389: Again, reducing the case where someone might be switching quickly. this should help, in addition to the `max-age` change, reduce false-positives of redirect loop detection

packages/backend/src/tokens/handshake.ts
```diff
@@ -222,6 +222,7 @@ export class HandshakeService {
       const newUrl = new URL(this.authenticateContext.clerkUrl);
       newUrl.searchParams.delete(constants.QueryParameters.Handshake);
       newUrl.searchParams.delete(constants.QueryParameters.HandshakeHelp);
+      newUrl.searchParams.delete(constants.QueryParameters.DevBrowser);
```
nikosdouvlis, line 225: This looks good to me. Was there any reason we used to delete it from the clerk-js side only or just an omission?

PR #6638, opened 2025-08-26
refactor(clerk-js): Using snakeToCamel for signal field errors

nikosdouvlis:
…

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation errors now appear on the relevant form fields instead of as generic/global messages, improving clarity when correcting inputs.

* **Refactor**
  * Streamlined error mapping so field-specific errors are assigned consistently based on incoming parameter names.

* **Chores**
  * Added an empty changeset entry (metadata file) to the repository.


packages/clerk-js/src/core/signals.ts
```diff
@@ -69,44 +70,12 @@ function errorsToParsedErrors(error: unknown): Errors {
...
-          parsedErrors.fields.legalAccepted = error;
-          break;
-        default:
-          parsedErrors.global.push(error);
-          break;
-      }
-    } else {
-      parsedErrors.global.push(error);
+      const name = snakeToCamel(error.meta.paramName);
+      parsedErrors.fields[name as keyof typeof parsedErrors.fields] = error;
```
dstaley, line 74: we should probably guard against setting a new key outside of the explicitly defined ones:

```ts
if (name in parsedErrors.fields) {
  parsedErrors.fields[name] = error;
}
```

nikosdouvlis: I thought about this as well, but I don't think there's a case where FAPI is going to return a new field here (keep in mind we're already in the `if paramName' in error.meta` block). Even if a new field is introduced at some point, we probably want to keep the error (so the users can log it for example) instead of discarding it completely, even if the types are not yet updated.

That was pretty much my reasoning for not including the check here but I don't feel strongly about it!

dstaley (approved):

PR #6644, opened 2025-08-26
fix(clerk-js,clerk-react): Avoid CLS when `fallback` is passed to  PricingTable

panteliselef:
## Description

`<PricingTable/>` is one of the few components that accepts the `fallback` prop and that its initial data are loaded asynchronously after clerk.js has loaded. We need to wait for that plans to be resolved and only then swap the fallback with the content.

⚠️ One disadvantage of this fix is that if your app is on an new `@clerk/clerk-react` version with an older clerk-js version (probably pinned) the component that you pass in the fallback prop will be visible indefinitely.

### Before
https://github.com/user-attachments/assets/f39b481e-7515-4067-adcd-53f929ee6812

### After
https://github.com/user-attachments/assets/dab17ea4-8ada-4e8c-9f53-a96afcef805c

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Pricing Table now reliably waits for pricing data/readiness before hiding its fallback, preventing premature hiding and ensuring correct rendering for both signed-in and signed-out users.
  * Improved component mount readiness handling to reduce intermittent rendering issues.

* **Chores**
  * Prepared patch releases for the JavaScript and React packages.


packages/react/src/components/uiComponents.tsx
```diff
@@ -577,7 +577,11 @@ export const Waitlist = withClerk(
 
 export const PricingTable = withClerk(
   ({ clerk, component, fallback, ...props }: WithClerkProp<PricingTableProps & FallbackProp>) => {
-    const mountingStatus = useWaitForComponentMount(component);
+    const mountingStatus = useWaitForComponentMount(component, {
+      // This attribute is added to the PricingTable root element after we've successfully fetched the plans asynchronously.
+      selector: '[data-component-status="ready"]',
+    });
+    console.log('mountingStatus', mountingStatus);
```
brkalow:
```suggestion
```

packages/react/src/utils/useWaitForComponentMount.ts
```diff
@@ -1,62 +1,77 @@
...
-        reject(new Error(`Timeout waiting for element children`));
-      }, timeout);
-    }
-  });
-}
+const waitForElementChildren = createAwaitableMutationObserver({
+  childList: true,
+  subtree: true,
+  isReady: (el, selector) => !!el?.childElementCount && el?.matches?.(selector) && el.childElementCount > 0,
+});
```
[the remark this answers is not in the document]
brkalow: @panteliselef this comment is worth looking at. I think it's fine as-is because the children will change when the component loads, agree?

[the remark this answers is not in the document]
panteliselef: Correct, the current solution does not require to watch attributes, because events will only trigger for the root, which is `body` and we are not interested in that.

packages/clerk-js/src/ui/elements/InvisibleRootBox.tsx
```diff
@@ -16,15 +16,19 @@ const _InvisibleRootBox = React.memo((props: RootBoxProps) => {
...
 
   return (
     <>
       {props.children}
       {showSpan && (
         <span
-          ref={el => (parentRef.current = el ? el.parentElement : parentRef.current)}
+          ref={el => {
+            parentRef.current = el ? el.parentElement : parentRef.current;
+          }}
```
panteliselef:
```suggestion
          ref={el => (parentRef.current = el ? el.parentElement : parentRef.current)}
```

packages/react/src/utils/useWaitForComponentMount.ts
```diff
@@ -66,15 +81,22 @@ export function useWaitForComponentMount(component?: string) {
...
 
     if (typeof window !== 'undefined' && !watcherRef.current) {
-      watcherRef.current = waitForElementChildren({ selector: `[data-clerk-component="${component}"]` })
+      const defaultSelector = `[data-clerk-component="${component}"]`;
+      const selector = options?.selector;
+      watcherRef.current = waitForElementChildren({
+        selector: selector
+          ? // Allows for `[data-clerk-component="xxxx"][data-some-attribute="123"] .my-class`
+            defaultSelector + selector
+          : defaultSelector,
```
brkalow, line 90: do we need to join with a `' '`? Or intentionally not?

panteliselef: The concatenation happens between `[data-clerk-component="xxxx"]` and `[data-some-attribute="123"] .my-class`, and between those there is no space.

PR #6677, opened 2025-09-01
fix(types, clerk-js): Update Errors interface to allow null for raw and global error arrays

nikosdouvlis:
## Description

`errors.global` can now be used without checking for the `.length` property:

```
if (errors.global) {
   return <div>{errors.global[0].message}</div>
}
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- Bug Fixes
  - Avoids showing empty generic errors on initial load by treating error lists as absent until needed.
  - Ensures non-API errors display consistently and map reliably to field-specific messages.
  - Prevents rare crashes or UI glitches when processing unexpected error shapes.

- Refactor
  - Error lists are now created lazily, reducing unnecessary allocations and improving efficiency.


panteliselef (approved):
Let's verify if an empty changeset is appropriate.

packages/clerk-js/src/core/signals.ts
```diff
@@ -52,30 +52,38 @@ function errorsToParsedErrors(error: unknown): Errors {
       captcha: null,
       legalAccepted: null,
     },
-    raw: [],
-    global: [],
+    raw: null,
```
dstaley, line 55: I hate that this is the case, but we need to update the default errors in the state proxy in clerk-react as well 😭

nikosdouvlis: Good catch! 
Ill try to prioritize some tests to catch these sooner and a (potential) refactor later. Fixing...

PR #6688, opened 2025-09-02
fix(clerk-js): Allow clearing fetch cache on sign out

jacekradko:
## Description

Clear fetchCache after sign out

Fixes: USER-3157

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- Bug Fixes
  - Automatically clears client-side fetch cache after signing out in the browser, preventing stale or previous-user data from reappearing.
  - Improves reliability when switching accounts or returning to the app post sign-out by ensuring a clean state for subsequent requests.
  - Works transparently with existing integrations; no configuration or API changes required.


packages/clerk-js/src/ui/hooks/useFetch.ts
```diff
@@ -31,6 +31,15 @@ export const clearFetchCache = () => {
   requestCache = new Map<string, State>();
 };
 
+/**
+ * Global function to clear the fetch cache.
+ * This is exposed on the window object to allow core modules to clear the cache
+ * without importing from UI modules.
+ */
```
jacekradko, line 38: I don't really like this solution long term, but for now it does the job

panteliselef (changes requested):
As you mentioned this would not be a long term solution and neither #6690 seems to be, since we are focusing solely in the fact that a sign out is performed. The solution is much simpler and I opened a [PR](https://github.com/clerk/javascript/pull/6695) for it.

PR #6702, opened 2025-09-03
fix(clerk-js, shared): Auto revalidate hook on mutations

panteliselef:
### Issue
Currently, when users perform actions like confirming checkouts or canceling subscriptions, the UI doesn't automatically reflect these changes. Users need to manually refresh or navigate away and back to see updated data, leading to a poor user experience.

### Solution
The proposed solution automatically revalidates data when relevant mutation occur, and deduplicates event listeners to avoid multiple parallel revalidation requests.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -239,6 +239,10 @@ export class Clerk implements ClerkInterface {
   #touchThrottledUntil = 0;
   #publicEventBus = createClerkEventBus();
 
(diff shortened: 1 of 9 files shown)


panteliselef:
Closing because we decided to use RQ which solves the same problem, but offers a few benefits as well.

PR #6703, opened 2025-09-03
fix(clerk-js): Do not trigger roles query when the current user's membership lacks permissions

LauraBeatris:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/ui/hooks/useFetchRoles.ts
```diff
@@ -14,9 +15,15 @@ const getRolesParams = {
...
 export const useFetchRoles = (enabled = true) => {
   const { organization } = useOrganization();
+  const canManageMemberships = useProtect({ permission: 'org:sys_memberships:manage' });
+  const canReadMemberships = useProtect({ permission: 'org:sys_memberships:read' });
+
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
   const getRoles = ({ pageSize, initialPage }: GetRolesParams) => organization!.getRoles({ pageSize, initialPage });
-  const { data, isLoading } = useFetch(enabled && !!organization?.id ? getRoles : undefined, {
+
+  const hasSystemPermissions = canManageMemberships || canReadMemberships;
```
LauraBeatris, line 24: We have this hook being called from `OrganizationProfile` pieces and `OrganizationSwitcher`, so instead of applying this condition in all places, I believe it should stay in the query itself.

packages/clerk-js/src/ui/elements/OrganizationPreview.tsx
```diff
@@ -38,10 +36,9 @@ export const OrganizationPreview = (props: OrganizationPreviewProps) => {
   } = props;
 
   const { localizeCustomRole } = useLocalizeCustomRoles();
-  const { options } = useFetchRoles(fetchRoles);
```
LauraBeatris, line 41: I've removed `fetchRoles` entirely from `OrganizationPreview` as it was relying on the array of roles against the membership ID.

We could simplify by just accessing the `roleName` from the membership object; no need to iterate.

packages/clerk-js/src/ui/hooks/useFetchRoles.ts
LauraBeatris: Technically, https://github.com/clerk/javascript/pull/6703/files#r2320397829 would already solve the issue of the 403 on the roles query when rendering `OrganizationSwitcher`, however I think it's worth to place an extra layer of protection here for future usage in other org components.

LauraBeatris:
Some tests are failing, taking a look into it now

iagodahlem (approved):

PR #6709, opened 2025-09-04
feat(shared): Increase sampling rate for component mounted telemetry events for account portal

heatlikeheatwave:
## Description

The changes in component-mounted.ts replaced the hardcoded `EVENT_SAMPLING_RATE = 0.1` with a dynamic `getComponentSamplingRate` function that reads the `NEXT_PUBLIC_CLERK_TELEMETRY_COMPONENT_MOUNTED_SAMPLE_RATE` or `CLERK_TELEMETRY_COMPONENT_MOUNTED_SAMPLE_RATE` environment variable, allowing the telemetry sampling rates for sign-in and sign-up component mounted events to be set by the account portal env var.

The default sampling rate remains 0.1 if the environment variable is not set.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Make component-mounted telemetry sampling configurable via NEXT_PUBLIC_CLERK_TELEMETRY_COMPONENT_MOUNTED_SAMPLE_RATE or CLERK_TELEMETRY_COMPONENT_MOUNTED_SAMPLE_RATE (0–1). Defaults to 10% when unset/invalid. Applies to account-portal component telemetry; other component telemetry keeps the default sampling. No public API changes.

* **Chores**
  * Shared package patch bump and updated release metadata; bundle size threshold increased slightly.

### DIFF
--- a/.changeset/brave-guests-kick.md
+++ b/.changeset/brave-guests-kick.md
@@ -1,0 +1,5 @@
+---
+'@clerk/localizations': patch
+---
+
+Fix formatting of cs-CZ localization strings
--- a/.changeset/fuzzy-rivers-tell.md
+++ b/.changeset/fuzzy-rivers-tell.md
@@ -1,0 +1,5 @@
+---
+"@clerk/vue": patch
+---
+
+Improved error message when Clerk plugin is not installed
--- a/packages/localizations/src/cs-CZ.ts
+++ b/packages/localizations/src/cs-CZ.ts
@@ -840,7 +840,8 @@ export const csCZ: LocalizationResource = {
   },
   unstable__errors: {
     already_a_member_in_organization: '{{email}} je již členem organizace.',
-    captcha_invalid: 'Registrace neúspěšná kvůli neúspěšným bezpečnostním validacím. Prosím, obnovte stránku a zkuste to znovu nebo kontaktujte podporu.',
+    captcha_invalid:
+      'Registrace neúspěšná kvůli neúspěšným bezpečnostním validacím. Prosím, obnovte stránku a zkuste to znovu nebo kontaktujte podporu.',
     captcha_unavailable:
       'Registrace nebyla úspěšná kvůli neúspěšné validaci bota. Prosím obnovte stránku a zkuste to znovu, nebo se obraťte na podporu pro další pomoc.',
     form_code_incorrect: 'Kód je nesprávný.',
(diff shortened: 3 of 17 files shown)


packages/shared/src/telemetry/events/component-mounted.ts
```diff
@@ -2,7 +2,24 @@ import type { TelemetryEventRaw } from '@clerk/types';
...
-const EVENT_SAMPLING_RATE = 0.1;
+const EVENT_SAMPLING_RATE = getComponentSamplingRate();
+
+/**
+ * Get the sampling rate for component mounted events for account portal from NEXT_PUBLIC_CLERK_COMPONENT_ACCOUNT_PORTAL_SAMPLING_RATE environment variable.
+ *
+ * @internal
+ */
+function getComponentSamplingRate(): number {
+  const samplingRate = process.env.NEXT_PUBLIC_CLERK_COMPONENT_ACCOUNT_PORTAL_SAMPLING_RATE;
```
mwickett: Should this be named to describe what it controls, rather than being coupled to where it's used? Like `NEXT_PUBLIC_CLERK_TELEMETRY_COMPONENT_MOUNTED_SAMPLE_RATE` or something?

heatlikeheatwave: sure

packages/shared/src/telemetry/events/component-mounted.ts
```diff
@@ -2,7 +2,24 @@ import type { TelemetryEventRaw } from '@clerk/types';
...
-const EVENT_SAMPLING_RATE = 0.1;
+const EVENT_SAMPLING_RATE = getComponentSamplingRate();
+
+/**
+ * Get the sampling rate for component mounted events for account portal from NEXT_PUBLIC_CLERK_TELEMETRY_COMPONENT_MOUNTED_SAMPLE_RATE environment variable.
+ *
+ * @internal
+ */
+function getComponentSamplingRate(): number {
+  const samplingRate = process.env.NEXT_PUBLIC_CLERK_TELEMETRY_COMPONENT_MOUNTED_SAMPLE_RATE;
```
mwickett: nit: We should support a non-next.js specific version of this env var I think, just to be framework agnostic

mwickett (approved):
Makes sense to me!

packages/shared/src/telemetry/events/component-mounted.ts
heatlikeheatwave: I feel like I'm missing context. The account portal is a Next.js app, and this change is specifically for that application, or do I have that wrong?

mwickett: You're correct - I brought it up for the sake of staying with convention for the monorepo. I think we usually support both prefixed and non-prefixed variants of env vars - I noticed in the [cursor rules](https://github.com/clerk/javascript/blob/b2f8c223e83309af64b3fe79d4af85f3986218fc/.cursor/rules/monorepo.mdc#L69). It's not critical for our specific use case here, but feels like the right thing to do.

heatlikeheatwave: sure, I'll make that change

packages/shared/src/telemetry/events/component-mounted.ts
```diff
@@ -28,7 +27,7 @@ function createPrebuiltComponentEvent(event: typeof EVENT_COMPONENT_MOUNTED | ty
   ): TelemetryEventRaw<EventPrebuiltComponent> {
     return {
       event,
-      eventSamplingRate: EVENT_SAMPLING_RATE,
+      eventSamplingRate: Number(process.env.NEXT_PUBLIC_CLERK_TELEMETRY_COMPONENT_MOUNTED_SAMPLE_RATE) || 0.1,
```
heatlikeheatwave: The bundlewatch check failed in the CI action until I reduced the change to this one line.
See [here](https://github.com/clerk/javascript/actions/workflows/ci.yml) and filter by branch `heat/grow-677-account-portal-telemetry-events-for-sign-in-and-sign-up`

packages/shared/src/telemetry/events/component-mounted.ts
```diff
@@ -28,7 +28,8 @@ function createPrebuiltComponentEvent(event: typeof EVENT_COMPONENT_MOUNTED | ty
   ): TelemetryEventRaw<EventPrebuiltComponent> {
     return {
       event,
-      eventSamplingRate: EVENT_SAMPLING_RATE,
+      eventSamplingRate:
+        Number(process.env.NEXT_PUBLIC_CLERK_TELEMETRY_COMPONENT_MOUNTED_SAMPLE_RATE) || EVENT_SAMPLING_RATE,
```
brkalow: We can't use next-specific env vars here, as it's a shared package used across all of our framework SDKs.

brkalow (changes requested):

packages/shared/src/telemetry/events/component-mounted.ts
heatlikeheatwave: I made this here change [here](https://github.com/clerk/javascript/pull/6709/commits/34124b84dec2cf052e45985514c750f4dc79c921#diff-f92faf6ae3aa69ac325ebe65750a3cc01f7f374017530b81ca12e98622a6b80cR11-R20), but I'm not sure if that's the best way to go about it.

The goal is allow the account portal to override the sampling rate via an environment variable.

brkalow: @heatlikeheatwave Does it just need to be for this event? Or should AP have full sampling? If so, recall that we recently added the `perEventSampling` option to allow a full override of the event sampling rates, so presumably we could leverage this for AP and have AP collect every event.

heatlikeheatwave: It's only for this event. IIRC, there's no telemetry config in AP, it's all coming from clerk/javascript packages. How would it work to leverage `perEventSampling` in AP without duplicating telemetry events?

brkalow: Events are already de-duplicated by the Telemetry collector 👍 The changes here don't feel like a sustainable approach, I would rather we make the decision to stop sampling for all of AP, as opposed to introducing alternative config paths for this specific event.

heatlikeheatwave: Sure. What's the best way to stop sampling for all of AP?

mwickett: Just to close the loop here, @heat is going to expose `perEventSampling` config as a telemetry option at the provider level, and then we can make that change on AP.

PR #6737, opened 2025-09-09
fix(clerk-js): Handle removing origin in non-browser environments

bratsos:
## Description

Users reported a `[TypeError: Cannot read property 'origin' of undefined]` warning/error when calling `signOut()` from the `useAuth` hook in Expo applications. While the `signOut` functionality worked correctly, this error was logged to the console.

This issue started when we introduced the debugLogger in PR #6452, which attempts to strip the origin from the redirect URL before reporting it in debug logs.

## Steps to reproduce
- Clone https://github.com/beerose/clerk-expo-bare/ or use your own Expo + Clerk project
- Make sure you're not using `expo-router` - it imports `@expo/metro-runtime` which sets up polyfills for `window.location.origin` so you won't be able to see the error
- Login with a user and call `useAuth().signOut()`
- You should see the following error in the logs `[TypeError: Cannot read property 'origin' of undefined]`
- Checkout this branch
- On `packages/clerk-js` run `turbo build && yalc publish` (requires `yalc` installed globally)
- On your expo app run `yalc add @clerk/clerk-js`
- Rerun the flow
- You should not see the error anymore 🎉 

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- New Features
  - Added a utility to strip the site origin from same-origin absolute URLs while preserving relative and external URLs.
- Bug Fixes
  - Avoids errors when URL handling runs in non-browser environments (no window/location).
- Tests
  - Added tests covering browser and non-browser URL scenarios.
- Chores
  - Updated bundle size threshold and added a patch release changeset.


packages/clerk-js/src/utils/url.ts
```diff
@@ -174,6 +174,12 @@ export function toURL(url: string | URL): URL {
  * @returns {string} Returns the URL href without the origin
  */
 export function stripOrigin(url: URL | string): string {
+  // In non-browser environments `window.location.origin` might not be available
+  // if not polyfilled, so we can't construct a URL object with the `url` string
+  if (typeof window.location === 'undefined' && typeof url === 'string') {
```
nikosdouvlis, line 180: Sorry if Im missing something obvious, but would this be safer?
```suggestion
  if ((typeof window === 'undefined' || !window.location) && typeof url === 'string') {
```

packages/clerk-js/src/utils/url.ts
```diff
@@ -174,6 +174,12 @@ export function toURL(url: string | URL): URL {
...
+    return url;
+  }
```
brkalow: ❓ should we instead handle this in `toURL()` where the direct `window.location.origin` usage is?

bratsos: That was my initial thought, but I think `stripOrigin` is better suited for making sure `origin` is accessible, and as it accepts both `URL` and `string`, essentially the code looks like this:
- if you somehow managed to create a `URL` object, I'll pass it down to `toURL`
- if you passed me a string, and I cannot safely detect `origin` to remove it, I'll just return the string

Happy to change this if you think moving it to `toURL` is easier to understand though.

packages/clerk-js/src/utils/url.ts
bratsos: This is intentional as we're expecting that at least a `window` global object will be somehow declared. If it's not I prefer to get an error here, plus we'll have many other places that would break as well. WDYT?

_Edit: Happy to make this more defensive though if we think it's valuable_

packages/clerk-js/src/utils/url.ts
brkalow: @bratsos I don't think this covers the case of passing a fully qualified URL as a string, e.g.

```
stripOrigin("https://clerk.com/blog") // -> /blog
```

In this case, I would still expect it to strip the origin. `toURL()` only uses `window.location.origin` as the fallback.

packages/clerk-js/src/utils/url.ts
wobsoriano: `window` is defined in react-native 😬 

See my comment here https://github.com/clerk/javascript/pull/4095/files#r1742755342

packages/clerk-js/src/utils/url.ts
bratsos: In this case, if `location` is available (essentially in all browser environments) this will still work. It won't work if we detect that `location` is no available, so we're in an exotic runtime that isn't polyfilled.

The thing is that `toURL` uses `URL` directly, and even though it's polyfilled in react native ([source](https://github.com/facebook/react-native/blob/8c0c860e38f57e18296f689e47dfb4a54088c260/Libraries/Blob/URL.js#L115-L222)), the implementation is non-standard - and for example `.origin` which is accessed by `stripOrigin` is [not even implemented](https://github.com/facebook/react-native/blob/8c0c860e38f57e18296f689e47dfb4a54088c260/Libraries/Blob/URL.js#L176). So we cannot really assume that if `URL` exists, `new URL('https://clerk.com/blog', undefined)` would work as expected. I think it's too late if we reach to `toURL` when we detect that something is wrong in the runtime APIs.

An even better solution would probably be to try and detect if the required APIs are available, and work as expected, or detect the runtime and decide what to do based on that, but I think that's an overkill as in most cases you're either in a browser/polyfilled* environment (so `window.location` and `URL` exist and we can safely use them), or not, so in that case we cannot do much on a plain string, unless we want to implement a pathname parser in-house.

*: If you polyfilled the environment in a weird way this can also throw us off.

WDYT?

brkalow: Makes sense! Maybe we can adjust the JSDoc comment here to make it explicit that origin won't be stripped in certain non-browser environments.

bratsos: Sure: https://github.com/clerk/javascript/pull/6737/commits/ff7748eb25e460e6e2326857d58e50e5f42b80bb

PR #6747, opened 2025-09-10
fix(clerk-js, shared): Display empty data for authenticated billing hooks after sign out

panteliselef:
## Description

This PR ensures:
1. authed billing hooks clean up data on sign out
1. billing hooks do not error when clerk is not loaded

This happens when the billing hooks use `keepPreviousData: true` (default behaviour of AIOs). We want to keep using that to avoid CLS and content flashing on revalidations, but at the same time we don't want to display the previous "stale" data after the user has signed out.

### Before
https://github.com/user-attachments/assets/bdd891d9-95c9-40a6-9f14-7aca3e9209f6

### After
https://github.com/user-attachments/assets/3cb157d4-9459-46f4-9f9f-b2f1009b669c

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- Bug Fixes
  - Billing and subscription data no longer fetches while signed out; empty data is shown after sign-out.
  - Prevents premature network calls until the app is fully loaded, reducing flicker and errors.
  - Improves pagination and caching behavior by respecting sign-in state.

- Chores
  - Added release metadata for a patch update to @clerk/shared.

### DIFF
--- a/.changeset/major-steaks-hug.md
+++ b/.changeset/major-steaks-hug.md
@@ -1,0 +1,5 @@
+---
+'@clerk/shared': patch
+---
+
+Display empty data for authenticated billing hooks after sign out.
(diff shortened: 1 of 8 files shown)


wobsoriano (approved):
makes sense to me

aeliox (approved):

PR #6761, opened 2025-09-12
feat(clerk-js,shared): Align `useCheckout` with future hooks

panteliselef:
### Vanilla JS
```ts
// Before
const { getState, subscribe, confirm, start, clear, finalize  } = Clerk.checkout({ planId: "xxx", planPeriod: "annual" })
getState().isStarting
getState().isConfirming
getState().error
getState().checkout
getState().fetchStatus
getState().status

// After
const { checkout, errors, fetchStatus } = Clerk.checkout({ planId: "xxx", planPeriod: "annual" })
checkout.plan // null or defined based on `checkout.status`
checkout.status
checkout.start
checkout.confirm
```

### React
```ts
// Before
const { id, plan, status, start, confirm, paymentSource } = useCheckout({ planId: "xxx", planPeriod: "annual" })

// After
const { checkout, errors, fetchStatus } = usecCheckout({ planId: "xxx", planPeriod: "annual" })
checkout.plan // null or defined based on `checkout.status`
checkout.status
checkout.start
checkout.confirm
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Enhanced checkout error reporting with improved status tracking for payment flows.

* **Refactor**
  * Modernized checkout system architecture for more responsive real-time feedback during payment processing.

### DIFF
--- a/packages/clerk-js/src/core/modules/checkout/instance.ts
+++ b/packages/clerk-js/src/core/modules/checkout/instance.ts
@@ -14,7 +14,13 @@ function cacheKey(options: { userId: string; orgId?: string; planId: string; pla
   return `${userId}-${orgId || 'user'}-${planId}-${planPeriod}` as CheckoutKey;
 }
 
-const cache = new Map<CheckoutKey, { resource: CheckoutFuture; signals: ReturnType<typeof createSignals> }>();
+/**
+ * Stores the state of checkout instances in a cached based on their configuration as a cache key.
+ */
+const CheckoutSignalCache = new Map<
+  CheckoutKey,
+  { resource: CheckoutFuture; signals: ReturnType<typeof createSignals> }
+>();
 
 /**
  * Create a checkout instance with the given options
@@ -37,10 +43,9 @@ function createCheckoutInstance(clerk: Clerk, options: __experimental_CheckoutOp
     planPeriod,
   });
 
-  if (cache.has(checkoutKey)) {
-    return (
-      cache.get(checkoutKey) as { resource: CheckoutFuture; signals: ReturnType<typeof createSignals> }
-    ).signals.computedSignal() as CheckoutSignalValue;
+  const checkoutInstance = CheckoutSignalCache.get(checkoutKey);
+  if (checkoutInstance) {
+    return checkoutInstance.signals.computedSignal() as CheckoutSignalValue;
   }
 
   const signals = createSignals();
@@ -50,7 +55,7 @@ function createCheckoutInstance(clerk: Clerk, options: __experimental_CheckoutOp
     planId,
     planPeriod,
   });
-  cache.set(checkoutKey, { resource: checkout, signals });
+  CheckoutSignalCache.set(checkoutKey, { resource: checkout, signals });
   return signals.computedSignal() as CheckoutSignalValue;
 }
 
--- a/packages/clerk-js/src/ui/components/Checkout/parts.tsx
+++ b/packages/clerk-js/src/ui/components/Checkout/parts.tsx
@@ -1,5 +1,4 @@
 import { __experimental_useCheckout as useCheckout } from '@clerk/shared/react';
-import { useMemo } from 'react';
 
 import { Alert } from '@/ui/elements/Alert';
 import { Drawer, useDrawerContext } from '@/ui/elements/Drawer';
@@ -42,21 +41,12 @@ export const InvalidPlanScreen = () => {
   const { planPeriod } = useCheckoutContext();
   const { errors } = useCheckout();
 
-  const planFromError = useMemo(() => {
-    const _error = errors?.global?.find(e => e.code === 'invalid_plan_change');
-    // @ts-expect-error - meta is not yet defined
-    return _error?.meta?.plan;
-  }, [errors.global]);
-
-  const isPlanUpgradePossible = useMemo(() => {
-    const _error = errors?.global?.find(e => e.code === 'invalid_plan_change');
-    // @ts-expect-error - meta is not yet defined
-    return _error?.meta?.isPlanUpgradePossible || false;
-  }, [errors.global]);
-
-  if (!planFromError) {
+  const InvalidPlanError = errors?.global?.find(e => e.code === 'invalid_plan_change');
+  if (!InvalidPlanError) {
     return null;
   }
+
+  const { plan: planFromError, isPlanUpgradePossible } = InvalidPlanError?.meta || {};
 
   return (
     <Drawer.Body>


packages/clerk-js/src/core/resources/CommerceCheckout.ts
```diff
@@ -86,3 +91,171 @@ export class CommerceCheckout extends BaseResource implements CommerceCheckoutRe
...
+
+    if (parsedErrors.global) {
+      parsedErrors.global.push(error);
+    } else {
+      parsedErrors.global = [error];
+    }
+  });
+
+  return parsedErrors;
+}
```
panteliselef, line 126: Error structure is WIP

packages/clerk-js/src/core/clerk.ts
```diff
@@ -351,9 +351,9 @@ export class Clerk implements ClerkInterface {
     return Clerk._apiKeys;
   }
 
-  __experimental_checkout(options: __experimental_CheckoutOptions): __experimental_CheckoutInstance {
+  __experimental_checkout(options: __experimental_CheckoutOptions): CheckoutSignalValue {
     if (!this._checkout) {
-      this._checkout = params => createCheckoutInstance(this, params);
+      this._checkout = (params: any) => createCheckoutInstance(this, params);
```
dstaley, line 356: Is this params possible to be typed?

packages/clerk-js/src/core/modules/checkout/instance.ts
```diff
@@ -16,13 +14,12 @@ function cacheKey(options: { userId: string; orgId?: string; planId: string; pla
   return `${userId}-${orgId || 'user'}-${planId}-${planPeriod}` as CheckoutKey;
 }
 
+const cache = new Map<CheckoutKey, { resource: CheckoutFuture; signals: ReturnType<typeof createSignals> }>();
```
dstaley: Let's add some comments about how the cache is used.

packages/clerk-js/src/core/modules/checkout/instance.ts
```diff
@@ -40,48 +37,21 @@ function createCheckoutInstance(
...
-    });
-  };
-
-  const finalize = (params?: { navigate?: SetActiveNavigate }) => {
-    const { navigate } = params || {};
-    return clerk.setActive({ session: clerk.session?.id, navigate });
-  };
-
-  const clear = () => manager.clearCheckout();
+  if (cache.has(checkoutKey)) {
```
dstaley: If you do this instead, you have a better TypeScript DX:

```ts
const value = cache.get(checkoutKey);
if (value) {
  return value.signals.computedSignal();
}
```

this way you don't need to cast to narrow the type.

packages/clerk-js/src/core/modules/checkout/instance.ts
```diff
@@ -40,48 +37,21 @@ function createCheckoutInstance(
...
 
-  return {
-    start,
-    confirm,
-    finalize,
-    clear,
-    subscribe,
-    getState: manager.getCacheState,
-  };
+  const checkout = new CheckoutFuture(signals, {
```
dstaley, line 53: not saying you should do it this way, but in our SignInFuture class we don't directly access signals. The resources emit events and our state layer listens for those to propagate to the signals.

packages/clerk-js/src/core/resources/CommerceCheckout.ts
```diff
@@ -86,3 +91,171 @@ export class CommerceCheckout extends BaseResource implements CommerceCheckoutRe
     );
   };
 }
+
+function errorsToParsedErrors(error: unknown): CheckoutSignalValue['errors'] {
```
dstaley, line 95: just calling out we don't store this implementation alongside the resource, but that's because we have a dedicated signals file. not saying you need to move it, just wanted to point that out in case you can think of a better place that more closely aligns with signIn/signUp.

packages/clerk-js/src/core/resources/CommerceCheckout.ts
```diff
@@ -86,3 +91,171 @@ export class CommerceCheckout extends BaseResource implements CommerceCheckoutRe
...
+    raw: null,
+    global: null,
+  };
+
+  if (!error) {
+    return parsedErrors;
+  }
+
+  if (!isClerkAPIResponseError(error)) {
+    parsedErrors.raw = [];
```
dstaley, line 106: shouldn't you be putting the error into the arrays? this is just dropping the error completely I think

packages/clerk-js/src/core/resources/CommerceCheckout.ts
```diff
@@ -86,3 +91,171 @@ export class CommerceCheckout extends BaseResource implements CommerceCheckoutRe
...
+      parsedErrors.global.push(error);
+    } else {
+      parsedErrors.global = [error];
+    }
+  });
+
+  return parsedErrors;
+}
+
+export const createSignals = () => {
```
dstaley, line 128: In signIn/signUp we use exported signals so that we can access them in multiple places. By creating signals, you have to pass the created signal to each location where you want to access it. This might be preferred if multiple checkouts can exist at once (as opposed to signIn/signUp where only one can exist at a time).

packages/clerk-js/src/core/resources/CommerceCheckout.ts
```diff
@@ -86,3 +91,171 @@ export class CommerceCheckout extends BaseResource implements CommerceCheckoutRe
...
+
+      const errors = errorsToParsedErrors(error);
+      return { errors: errors, fetchStatus, checkout: resource };
+    },
+  );
+
+  return { resourceSignal, errorSignal, fetchSignal, computedSignal };
+};
+
+export class CheckoutFuture implements CheckoutFutureResourceLax {
```
dstaley, line 146: what's the reasoning behind `Lax`?

packages/clerk-js/src/core/resources/CommerceCheckout.ts
```diff
@@ -86,3 +91,171 @@ export class CommerceCheckout extends BaseResource implements CommerceCheckoutRe
...
+  get paymentSource() {
+    return this.resource.paymentSource ?? null;
+  }
+
+  get planPeriodStart() {
+    return this.resource.planPeriodStart;
+  }
+
+  async start(): Promise<{ error: unknown }> {
+    return this.runAsyncResourceTask(
```
dstaley, line 198: not loving that this is a slightly different implementation of `runAsyncResourceTask`. Can we come up with a single implantation that works for both use cases?

packages/clerk-js/src/ui/components/Checkout/CheckoutForm.tsx
```diff
@@ -138,22 +143,22 @@ export const CheckoutForm = withCardStateProvider(() => {
...
 
   const confirmCheckout = async (params: ConfirmCheckoutParams) => {
     card.setLoading();
     card.setError(undefined);
 
-    const { data, error } = await confirm(params);
+    const { error } = await checkout.confirm(params);
 
     if (error) {
+      // @ts-expect-error - error is not an Error
```
dstaley, line 159: what TS rule was this breaking?

packages/clerk-js/src/ui/components/Checkout/CheckoutPage.tsx
```diff
@@ -56,14 +58,13 @@ const FetchStatus = ({
   status,
 }: {
   children: React.ReactNode;
-  status: 'idle' | 'fetching' | 'error' | 'invalid_plan_change' | 'missing_payer_email';
+  status: 'idle' | 'fetching' | 'generic_error' | 'invalid_plan_change' | 'missing_payer_email';
 }) => {
-  const { checkout } = useCheckout();
-  const { fetchStatus, error } = checkout;
+  const { errors, fetchStatus } = useCheckout();
```
dstaley, line 63: 😍 love that you're using this in the AIO components!

packages/react/src/stateProxy.ts
```diff
@@ -103,6 +121,58 @@ export class StateProxy implements State {
...
+      fetchStatus: 'idle' as const,
+      checkout: {
+        get status() {
+          return gateProperty(target, 'status', 'needs_initialization') as 'needs_initialization';
+        },
+        get externalClientSecret() {
+          return gateProperty(target, 'externalClientSecret', null) as null;
+        },
+        get externalGatewayId() {
+          return gateProperty(target, 'externalGatewayId', null) as null;
```
dstaley, line 143: are these casts necessary? the types should be pulling through `gateProperty` correctly.

packages/types/src/clerk.ts
```diff
@@ -54,50 +52,53 @@ import type { SessionVerificationLevel } from './sessionVerification';
...
-import type { State } from './state';
+import type { FieldError, State } from './state';
 import type { Web3Strategy } from './strategies';
 import type { TelemetryCollector } from './telemetry';
 import type { UserResource } from './user';
 import type { Autocomplete, DeepPartial, DeepSnakeToCamel } from './utils';
 import type { WaitlistResource } from './waitlist';
 
-type __experimental_CheckoutStatus = 'needs_initialization' | 'needs_confirmation' | 'completed';
+interface GlobalErrors {
```
dstaley, line 62: what are your thoughts on putting this into the `./state` file?

packages/clerk-js/src/core/resources/CommerceCheckout.ts
panteliselef: It allows for properties to be `null` but to developers we want to expose a more strict type that uses discriminated unions and seems a class cannot implement a type like that.

packages/clerk-js/src/ui/components/Checkout/CheckoutForm.tsx
panteliselef: Because Error is unknown

panteliselef:
Closing as this was completed in #7232

PR #6766, opened 2025-09-12
feat(clerk-js,clerk-react,types): Signal email link sign in support

dstaley:
## Description

This PR adds support for signing in with an email link to our Signal implementation.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Experimental email-link sign-in: send a sign-in link, wait for verification, and auto-detect completion via URL.
  * Expanded verification payload: status (verified, expired, failed, client_mismatch), created session ID, and whether verification came from the same client.
  * React state: signIn.firstFactorVerification and signIn.emailLink (sendLink, waitForVerification, verification).

* **Chores**
  * Minor package version bumps to publish the experimental feature.
  * Increased bundle size threshold for legacy browser build.


packages/types/src/signInFuture.ts
```diff
@@ -35,6 +36,11 @@ export interface SignInFutureEmailCodeSendParams {
   email: string;
 }
 
+export interface SignInFutureEmailLinkSendParams {
+  email: string;
+  redirectUrl: string;
```
nikosdouvlis: I think we should update this to `verifyUrl`/`verificationUrl` instead - `redirectUrl` is overloaded and usually indicates the URL to redirect to after a complete si/su attempt

packages/types/src/signInFuture.ts
```diff
@@ -131,6 +139,41 @@ export interface SignInFutureResource {
...
+  emailLink: {
+    /**
+     * Used to send an email link to sign-in
+     */
+    sendLink: (params: SignInFutureEmailLinkSendParams) => Promise<{ error: unknown }>;
+
+    /**
+     * Will wait for verification to complete or expire
+     */
+    waitForVerification: () => Promise<{ error: unknown }>;
```
nikosdouvlis, line 160: The DX guide original stated `.verifyLink` but I do think `waitForVerification` is better 👍🏻

packages/react/src/stateProxy.ts
```diff
@@ -44,13 +45,40 @@ export class StateProxy implements State {
...
+          });
+        },
 
         create: this.gateMethod(target, 'create'),
         password: this.gateMethod(target, 'password'),
         sso: this.gateMethod(target, 'sso'),
         finalize: this.gateMethod(target, 'finalize'),
 
         emailCode: this.wrapMethods(() => target().emailCode, ['sendCode', 'verifyCode'] as const),
+        emailLink: this.wrapStruct(
```
nikosdouvlis, line 76: Do we need `wrapStruct` because of the getter (`verification`) that was introduced?

nikosdouvlis (approved):

packages/react/src/stateProxy.ts
dstaley: yeah, `emailLink` has both methods and properties. We might be able to just use `wrapStruct` in place of `wrapMethods` to be consistent though. Not sure if it's worth refactoring since this entire construct might be going away (hopefully [please])

PR #6776, opened 2025-09-15
feat(clerk-js): Filter undefined values from request body

nikosdouvlis:
## Description
WIP

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/utils/filterUndefinedValues.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Filters out undefined values from the first level of an object.
+ * Preserves all other falsy values (null, false, 0, empty string).
+ *
+ * @param obj - The object to filter, or any other value
+ */
+export function filterUndefinedValues<T>(obj: T): T {
+  // Return non-objects as-is (including FormData, arrays, primitives, etc.)
+  if (!obj || typeof obj !== 'object' || Array.isArray(obj) || obj instanceof FormData) {
```
panteliselef: If we want to handle instances of different classes not just FormData, we can do
```suggestion
  if (!obj || typeof obj !== 'object' || Object.getPrototypeOf(obj) !== Object.prototype) {
```

nikosdouvlis: I think thats better, let me change it

PR #6777, opened 2025-09-15
feat(clerk-js,shared): Add onBeforeRetry to FAPI client

jacekradko:
## Description

Adding `x-clerk-retry-attempt` to GET fetch requests initiated as part of retry logic.

Fixes: USER-3480

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * GET retries now add a _clerk_retry_attempt query marker on retried requests and invoke a pre-retry hook before each retry (not added on the initial request).

* **Documentation**
  * Clarified retry option defaults and documented the pre-retry hook behavior and timing.

* **Tests**
  * Added tests covering retry behavior and updated fetch-related assertions and typings for robustness.

* **Chores**
  * Minor package version bumps.

### DIFF
--- a/packages/clerk-js/src/core/fapiClient.ts
+++ b/packages/clerk-js/src/core/fapiClient.ts
@@ -227,7 +227,7 @@ export function createFapiClient(options: FapiClientOptions): FapiClient {
     // avoid triggering a CORS OPTION request as it currently breaks cookie dropping in Safari.
     const overwrittenRequestMethod = method === 'GET' ? 'GET' : 'POST';
     let response: Response;
-    const urlStr = requestInit.url.toString();
+    const url = requestInit.url;
     const fetchOpts: FapiRequestInit = {
       ...requestInit,
       method: overwrittenRequestMethod,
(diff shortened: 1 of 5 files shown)


tmilewski (approved):

packages/clerk-js/src/core/fapiClient.ts
```diff
@@ -250,6 +250,11 @@ export function createFapiClient(options: FapiClientOptions): FapiClient {
             // We want to retry only GET requests, as other methods are not idempotent.
             return overwrittenRequestMethod === 'GET' && iterations < maxTries;
           },
+          onBeforeRetry: (iteration: number): void => {
+            if (fetchOpts.headers instanceof Headers) {
+              fetchOpts.headers.set('x-clerk-retry-attempt', iteration.toString());
+            }
+          },
```
brkalow: This will potentially have CORS implications, as this header is not a "simple request" header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#simple_requests

I think this will need to be a query parameter... 🤔

jacekradko: Yeah. That is a good point. This will make it a non-simple request.

packages/clerk-js/src/core/fapiClient.ts
```diff
@@ -227,7 +227,7 @@ export function createFapiClient(options: FapiClientOptions): FapiClient {
     // avoid triggering a CORS OPTION request as it currently breaks cookie dropping in Safari.
     const overwrittenRequestMethod = method === 'GET' ? 'GET' : 'POST';
     let response: Response;
-    const urlStr = requestInit.url.toString();
+    const url = requestInit.url;
```
jacekradko, line 230: We were converting the url to string here, then converting it back to a url for modification and then converting it a string.  Now we are just converting it to string once and passing the URL object to fetch

packages/clerk-js/src/core/fapiClient.ts
jacekradko: @brkalow Updated!

PR #6815, opened 2025-09-19
Feature/role change session revocation

YektaRoustaei:
Description

This PR ensures that when a user’s role is updated in the dashboard, all active sessions for that user are automatically revoked.

Previously, role changes did not revoke live sessions, which caused stale permission data to remain cached in active sessions. This led to UI inconsistencies and, more critically, potential security issues — for example, a downgraded user could still retain elevated access until their session expired.

By revoking all sessions immediately upon role change:

🚫 Eliminates cache bugs caused by outdated role/permission data.

🔒 Ensures that security updates take effect instantly across all devices.

🎛️ Preserves existing error handling and UI feedback patterns, with added visual indication during the revocation process.

How to Test

Assign a role to a user.

Log in with that user on multiple devices/tabs.

Change the user’s role in the dashboard.

Verify that all existing sessions are revoked and the user is prompted to log in again.

Confirm that the new role/permissions are applied correctly after reauthentication.

### DIFF
--- a/packages/clerk-js/src/ui/components/OrganizationProfile/ActiveMembersList.tsx
+++ b/packages/clerk-js/src/ui/components/OrganizationProfile/ActiveMembersList.tsx
@@ -1,5 +1,6 @@
 import { useOrganization, useUser } from '@clerk/shared/react';
 import type { OrganizationMembershipResource } from '@clerk/types';
+import { useState } from 'react';
 
 import { useCardState } from '@/ui/elements/contexts';
 import { ThreeDotsMenu } from '@/ui/elements/ThreeDotsMenu';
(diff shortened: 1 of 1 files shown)


YektaRoustaei:
@tmilewski

YektaRoustaei:
@tmilewski

tmilewski:
@YektaRoustaei Thank you! I'm going to ping @panteliselef as it'll be better that he reviews this.

panteliselef:
@YektaRoustaei  there are no `Clerk.client.sessions.getUserSessions()` and `Clerk.client.sessions.revokeSession` methods.

panteliselef:
This PR seems to be using methods that are non existent in our JS SDK, also the logic seems to be flawed and attempting to fix a problem that is not well defined. Feel free to open a new PR in the future.

PR #6839, opened 2025-09-22
feat: Make SubscriptionItem's plan and planId optional and nullable

paddycarver:
## Description

We're introducing changes in future API versions that will either not send plan/plan_id at all on some subscription items, or may send null instead. Which hasn't been determined yet. Updating the client and webhooks to accept either, though.

C1s must take an explicit action before any subscription item will return a null or omitted plan/plan_id, so we'll have the opportunity to warn them.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/api/resources/CommerceSubscriptionItem.ts
+++ b/packages/backend/src/api/resources/CommerceSubscriptionItem.ts
@@ -46,11 +46,11 @@ export class CommerceSubscriptionItem {
     /**
      * The plan associated with this subscription item.
      */
-    readonly plan: CommercePlan,
+    readonly plan: CommercePlan | null | undefined,
     /**
      * The plan ID.
      */
-    readonly planId: string,
+    readonly planId: string | null | undefined,
     /**
      * Unix timestamp (milliseconds) of when the subscription item was created.
      */
--- a/packages/backend/src/api/resources/JSON.ts
+++ b/packages/backend/src/api/resources/JSON.ts
@@ -908,24 +908,27 @@ export interface CommerceSubscriptionItemWebhookEventJSON extends ClerkResourceJ
   next_payment_amount: number;
   next_payment_date: number;
   amount: CommerceMoneyAmountJSON;
-  plan: {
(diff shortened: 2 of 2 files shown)


.changeset/silly-rooms-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/backend': minor
+---
+
+Allow subscription item plan and planId to be null or undefined
```
panteliselef: We use `[Billing Beta]` to track breaking changes while in Beta.
```suggestion
[Billing Beta] Update subscription item's `plan` and `planId` properties to be `null` or `undefined`.
```

packages/backend/src/api/resources/CommerceSubscriptionItem.ts
```diff
@@ -105,15 +105,24 @@ export class BillingSubscriptionItem {
...
 
+    let plan = null;
+    if (data.plan !== undefined && data.plan !== null) {
+      plan = BillingPlan.fromJSON(data.plan);
+    }
+    let planId = null;
+    if (data.plan_id !== undefined) {
+      planId = data.plan_id;
+    }
+
```
panteliselef:
```suggestion
```

packages/backend/src/api/resources/CommerceSubscriptionItem.ts
```diff
@@ -105,15 +105,24 @@ export class BillingSubscriptionItem {
...
       data.id,
       data.status,
       data.plan_period,
       data.period_start,
       data.next_payment,
       formatAmountJSON(data.amount),
-      BillingPlan.fromJSON(data.plan),
-      data.plan_id,
+      plan,
+      planId,
```
panteliselef:
```suggestion
      data.plan ? BillingPlan.fromJSON(data.plan) : null,
      data.plan_id ?? null,
```

PR #6852, opened 2025-09-24
feat(nextjs): Add CI environment detection header for Next.js keyless app creation

heatlikeheatwave:
## Description

To help us debug why keyless applications are failing to make certain API requests, this PR forwards a new header, `clerk-is-ci`) with requests to `accountless_applications`.
Adds a new function, `detectCIEnvironment()`, that checks for 25+ common CI environment variables (GitHub Actions, GitLab CI, CircleCI, Jenkins, Vercel, Netlify, etc.)

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/nextjs/src/server/keyless-custom-headers.ts
```diff
@@ -92,5 +140,7 @@ export function formatMetadataHeaders(metadata: MetadataHeaders): Headers {
     headers.set('Clerk-Auth-Status', metadata.xClerkAuthStatus);
   }
 
+  headers.set('Clerk-Is-CI', metadata.isCI.toString());
```
tmilewski: 💭 Should we set this on every request or only when true?

packages/nextjs/src/server/keyless-custom-headers.ts
```diff
@@ -32,9 +33,56 @@ export async function collectKeylessMetadata(): Promise<MetadataHeaders> {
...
+  return ciIndicators.some(indicator => {
+    const value = process.env[indicator];
+    if (value === undefined) {
+      return false;
+    }
+
+    const normalizedValue = value.trim().toLowerCase();
+    const falsyValues = ['', 'false', '0', 'no'];
+
+    return !falsyValues.includes(normalizedValue);
```
tmilewski: 💭 A quick perf win:

```suggestion

  const falsyValues = new Set<string>(['', 'false', '0', 'no']);

  return ciIndicators.some(indicator => {
    const value = process.env[indicator];
    if (value === undefined) {
      return false;
    }

    const normalizedValue = value.trim().toLowerCase();
    return !falsyValues.has(normalizedValue);
```

packages/nextjs/src/server/keyless-custom-headers.ts
heatlikeheatwave: Good call out, this should only be added when it's true. I made the change here https://github.com/clerk/javascript/pull/6852/commits/62c46973516a23439e407cf9c0157c78ed18a0c6

PR #6856, opened 2025-09-25
feat(shared): Improve error handling for clerk-js loading

brkalow:
## Description

This improves error handling during clerk-js loading. Specifically, it uses our `ClerkRuntimeError` class and ensures we are propagating underlying errors up so we can see the cause. Previously, we swallowed errors and only threw a generic error with a generic message ("Clerk: Failed to load Clerk").

Additionally, I've added an optional `cause` option to `ClerkRuntimeError` that allows us to pass along the original, unwrapped error for additional introspect (h/t @jacekradko)

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/src/loadClerkJsScript.ts
+++ b/packages/shared/src/loadClerkJsScript.ts
@@ -1,13 +1,15 @@
 import type { ClerkOptions, SDKMetadata, Without } from '@clerk/types';
 
-import { buildErrorThrower } from './error';
+import { buildErrorThrower, ClerkRuntimeError } from './error';
 import { createDevOrStagingUrlCache, parsePublishableKey } from './keys';
 import { loadScript } from './loadScript';
 import { isValidProxyUrl, proxyUrlToAbsoluteURL } from './proxy';
 import { addClerkPrefix } from './url';
 import { versionSelector } from './versionSelector';
 
-const FAILED_TO_LOAD_ERROR = 'Clerk: Failed to load Clerk';
+const ERROR_CODE = 'failed_to_load_clerk_js';
+const ERROR_CODE_TIMEOUT = 'failed_to_load_clerk_js_timeout';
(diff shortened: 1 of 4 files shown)


packages/shared/src/loadScript.ts
```diff
@@ -11,6 +11,9 @@ type LoadScriptOptions = {
   beforeLoad?: (script: HTMLScriptElement) => void;
 };
 
+/**
+ *
+ */
```
brkalow: Not sure how this ended up here 🙃

brkalow:
```suggestion
```

packages/shared/src/loadClerkJsScript.ts
```diff
@@ -1,13 +1,15 @@
...
+import { buildErrorThrower, ClerkRuntimeError } from './error';
 import { createDevOrStagingUrlCache, parsePublishableKey } from './keys';
 import { loadScript } from './loadScript';
 import { isValidProxyUrl, proxyUrlToAbsoluteURL } from './proxy';
 import { addClerkPrefix } from './url';
 import { versionSelector } from './versionSelector';
 
-const FAILED_TO_LOAD_ERROR = 'Clerk: Failed to load Clerk';
+const ERROR_CODE = 'failed_to_load_clerk_js';
+const ERROR_CODE_TIMEOUT = 'failed_to_load_clerk_js_timeout';
```
panteliselef, line 11: Is this helping because we want to better track why an error is thrown ?

panteliselef (approved):

packages/shared/src/loadClerkJsScript.ts
brkalow: Correct 👍

PR #6858, opened 2025-09-26
feat(shared): Capture auth component mounted for all SDK types

heatlikeheatwave:
## Description

Enables 100% sampling rate for `SignIn` and `SignUp` components for all SDKs. This will enable a more holistic comparison of Auth Component Mounted data between keyless and other SDK types. 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- **New Features**
  - Telemetry: per-event sampling added; SignIn and SignUp mount events now captured at full sampling, other component mounts and open events remain at lower sampling.

- **Chores**
  - Prepared a patch release for @clerk/shared.

### DIFF
--- a/.changeset/pretty-rings-compare.md
+++ b/.changeset/pretty-rings-compare.md
@@ -1,0 +1,5 @@
+---
+"@clerk/shared": patch
+---
+
+Increase sampling for high-signal auth components on mount.
--- a/packages/shared/src/telemetry/events/component-mounted.ts
+++ b/packages/shared/src/telemetry/events/component-mounted.ts
@@ -4,6 +4,19 @@ const EVENT_COMPONENT_MOUNTED = 'COMPONENT_MOUNTED';
 const EVENT_COMPONENT_OPENED = 'COMPONENT_OPENED';
 const EVENT_SAMPLING_RATE = 0.1;
 
+/** Increase sampling for high-signal auth components on mount. */
+const AUTH_COMPONENTS = new Set<string>(['SignIn', 'SignUp']);
+
+/**
+ * Returns the per-event sampling rate for component-mounted telemetry events.
+ * Uses a higher rate for SignIn/SignUp to improve signal quality.
+ *
+ *  @internal
+ */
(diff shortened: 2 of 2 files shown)


.changeset/pretty-rings-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/shared": patch
+---
+
+feat(shared): Capture auth component (SignIn, SignUp) mounted telemetry events for all SDK types.
```
dstaley:
```suggestion
Increase sampling for high-signal auth components on mount.
```

I think this more accurately describes what the change is to an external customer.

packages/shared/src/telemetry/events/component-mounted.ts
```diff
@@ -4,6 +4,19 @@ const EVENT_COMPONENT_MOUNTED = 'COMPONENT_MOUNTED';
...
+const AUTH_COMPONENTS = new Set<string>(['SignIn', 'SignUp']);
+
+/**
+ * Returns the per-event sampling rate for component-mounted telemetry events.
+ * Uses a higher rate for SignIn/SignUp to improve signal quality.
+ *
+ *  @internal
+ */
+function getComponentMountedSamplingRate(component: string): number {
+  return AUTH_COMPONENTS.has(component) ? 1 : EVENT_SAMPLING_RATE;
```
dstaley, line 17: Do we really need to go from 10% to 100%? Not objecting to it, just want to call out that we can _probably_ get a representative sample at a lower sampling rate.

heatlikeheatwave: I think if this was telemetry event data we used in isolation we could lower the sampling rate, but we're trying to compare the telemetry event for this between keyless (which has 100% sampling for this event) and all SDKs. I might have that wrong though, and we could lower it. What do you think?

PR #6860, opened 2025-09-26
feat(shared): Intelligent retries for existing clerk-js script

brkalow:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- Bug Fixes
  - Resolves intermittent failures when loading the Clerk script by detecting failed requests, removing problematic scripts, and retrying automatically.
  - Prevents stalls during initialization by cleaning up stale scripts on timeout and attempting a clean reload.
- Refactor
  - Streamlines script loading and retry logic for more robust startup behavior.
- Documentation
  - Clarifies comments to reflect improved error detection and automatic retry behavior, aiding maintainability.

### DIFF
--- a/packages/shared/src/loadClerkJsScript.ts
+++ b/packages/shared/src/loadClerkJsScript.ts
@@ -88,6 +88,10 @@ function hasScriptRequestError(scriptUrl: string): boolean {
     if (scriptEntry.responseEnd > 0 && scriptEntry.responseStart > 0) {
       return true;
     }
+
+    if (scriptEntry.responseStatus === 0) {
+      return true;
+    }
   }
 
   return false;
--- a/packages/shared/src/loadClerkJsScript.ts
+++ b/packages/shared/src/loadClerkJsScript.ts
@@ -60,6 +60,40 @@ function isClerkProperlyLoaded(): boolean {
 }
 
 /**
+ * Checks if an existing script has a request error using Performance API.
+ *
+ * @param scriptUrl - The URL of the script to check.
+ * @returns True if the script has failed to load due to a network/HTTP error.
+ */
+function hasScriptRequestError(scriptUrl: string): boolean {
+  if (typeof window === 'undefined' || !window.performance) {
+    return false;
+  }
+
+  const entries = performance.getEntries() as PerformanceResourceTiming[];
+  const scriptEntry = entries.find(entry => entry.name === scriptUrl);
+
+  if (!scriptEntry) {
+    return false;
+  }
+
+  // transferSize === 0 with responseEnd === 0 indicates network failure
+  // transferSize === 0 with responseEnd > 0 might be a 4xx/5xx error or blocked request
+  if (scriptEntry.transferSize === 0 && scriptEntry.decodedBodySize === 0) {
+    // If there was no response at all, it's definitely an error
+    if (scriptEntry.responseEnd === 0) {
+      return true;
+    }
+    // If we got a response but no content, likely an HTTP error (4xx/5xx)
+    if (scriptEntry.responseEnd > 0 && scriptEntry.responseStart > 0) {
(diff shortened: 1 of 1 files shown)


packages/shared/src/loadClerkJsScript.ts
```diff
@@ -59,6 +59,40 @@ function isClerkProperlyLoaded(): boolean {
   return typeof clerk === 'object' && typeof clerk.load === 'function';
 }
 
+/**
+ * Checks if an existing script has a request error using Performance API.
+ *
+ * @param scriptUrl - The URL of the script to check.
+ * @returns True if the script has failed to load due to a network/HTTP error.
+ */
+function hasScriptRequestError(scriptUrl: string): boolean {
```
brkalow, line 69: This is a slightly cursed way trying to check for an existing script load error without the ability to add an `onError` callback to the script. `window.performance.getEntries()` should contain the network request to fetch clerk-js, and so we can better infer what the status is without an arbitrary timeout.

nikosdouvlis: I like this!

packages/shared/src/loadClerkJsScript.ts
```diff
@@ -144,20 +184,29 @@ const loadClerkJsScript = async (opts?: LoadClerkJsScriptOptions): Promise<HTMLS
...
 
+  const scriptUrl = clerkJsScriptUrl(opts);
+  const existingScript = document.querySelector<HTMLScriptElement>('script[data-clerk-js-script]');
+
+  if (existingScript) {
+    if (hasScriptRequestError(scriptUrl)) {
+      existingScript.remove();
+    } else {
+      try {
+        return await waitForClerkWithTimeout(timeout);
```
nikosdouvlis: I'm having seconds thoughts about this particular `return` here. I checked the code and it seems that there's no retry logic in IsomorphicClerk (the calling site) or in `loadClerkJsScript` - the retry logic is in `loadScript` which will not be called if we return here.

So, if I understand correctly, if the `<script>` fails to load and we're going to remove it and implicitly retry via `loadScript`. If, however, the script hasn't errored at the point of the check, we're going to wait for a timeout and then completely fail. This is good because the `status` will change, but it feels like we  should at least trigger a few retries first before failing completely. Failing to load the script cannot be recovered from the perspective of the user/dev once `status` switches so retrying makes sense

brkalow: @nikosdouvlis I believe this still works because of `return await`. The function will not return until the promise from `waitForClerkWithTimeout` is resolved, which means it should throw if we can't detect that Clerk loaded in time.

I did spend some time trying to refactor this part on Friday though, I think it could be clearer what's going on.  I will revisit it again today!

nikosdouvlis: I haven't tested it locally yet, but what you're saying makes sense! The current implementation just bails and does not retry if the existing script fails to load, your fix is a nice improvement! 

The `await` part is tricky though, if someone accidentally removes the keyword, retrying for this case will break and TS won't catch it as the return signature will not change (still a promise). Switching to `then().catch` might be a good first step towards making this a bit harder to break?

Another idea would be to move the retry logic from `loadScript` on level up (inside `loadClerkJsScript`) so its a little easier to understand who's responsible to handle the error thrown, but Im not 100% sold on that idea yet

clerk-cookie:
Hello 👋

integration/tests/resiliency.test.ts
brkalow, line 1: Recommend viewing without whitespace changes: https://github.com/clerk/javascript/pull/6860/files?w=1

PR #6861, opened 2025-09-26
fix(clerk-js): Improve appearance selectors for tasks

iagodahlem:
## Description

The `TaskChooseOrganization` component was missing the root selector we infer from the flow routes.

Before:

CleanShot 2025-09-26 at 11 00 15@2x

After:

CleanShot 2025-09-26 at 13 03 32@2x

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed task appearance/selectors so organization options display correctly during onboarding/tasks (visible in Sign In/Sign Up flows).

* **Refactor**
  * Consolidated session task screens under a unified "tasks" flow and added dedicated task parts for organization selection to improve consistency and customization. No behavioral changes beyond the bug fix.

* **Chores**
  * Bumped dependency metadata for a patch release.

### DIFF
--- a/.changeset/shy-times-wash.md
+++ b/.changeset/shy-times-wash.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Fix TaskChooseOrganization component selector
--- a/.changeset/shy-times-wash.md
+++ b/.changeset/shy-times-wash.md
@@ -2,4 +2,4 @@
 '@clerk/clerk-js': patch
 ---
 
-Fix TaskChooseOrganization component selector
+Improve appearance selectors for tasks, such as including it within `SignIn/SignUp` components
--- a/packages/clerk-js/src/ui/elements/contexts/index.tsx
+++ b/packages/clerk-js/src/ui/elements/contexts/index.tsx
@@ -123,8 +123,7 @@ export type FlowMetadata = {
     | 'popover'
     | 'complete'
     | 'accountSwitcher'
-    | 'chooseOrganization'
-    | 'createOrganization';
+    | 'chooseOrganization';
 };
 
 const [FlowMetadataCtx, useFlowMetadata] = createContextAndHook<FlowMetadata>('FlowMetadata');
--- a/packages/clerk-js/src/ui/elements/contexts/index.tsx
+++ b/packages/clerk-js/src/ui/elements/contexts/index.tsx
@@ -101,6 +101,7 @@ export type FlowMetadata = {
     | 'oauthConsent'
     | 'subscriptionDetails'
     | 'subscriptionDetails'
+    | 'tasks'
     | 'taskChooseOrganization';
   part?:
     | 'start'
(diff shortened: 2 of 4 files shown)


packages/clerk-js/src/ui/elements/contexts/index.tsx
```diff
@@ -100,8 +100,7 @@ export type FlowMetadata = {
     | 'apiKeys'
     | 'oauthConsent'
     | 'subscriptionDetails'
-    | 'subscriptionDetails'
-    | 'taskChooseOrganization';
+    | 'sessionTasks';
```
panteliselef: @iagodahlem 
What does the html of a standalone `<TaskChooseOrganization/>` component look like now ?

iagodahlem: Good question @panteliselef, I was thinking we were actually exposing the standalone `<SessionTasks />` component, but we actually expose the single task `<TaskChooseOrganization />` component. 

Previously, the `TaskChooseOrganization` was already setting the root selector correctly to `cl-taskChooseOrganization-root`, so I updated it not to change that and just added the flow parts:

Here's the standalone `<TaskChooseOrganization />` before/after:

| Before | After |
| --- | --- |
| CleanShot 2025-09-29 at 10 14 34@2x | CleanShot 2025-09-29 at 10 13 48@2x |

And within the `<SignIn />` component:

| Before | After |
| --- | --- |
| CleanShot 2025-09-29 at 10 16 05@2x | CleanShot 2025-09-29 at 10 16 42@2x |

panteliselef: Why is the bottom right (after): different from the one that exists in the description ?
image

iagodahlem: The ones mentioned in the description are now outdated. I initially had the changed the <TaskChooseOrganization> root to be `sessionTasks`, but that would break the existing root selector when using the standalone version of that component.

nikosdouvlis (approved):

.changeset/shy-times-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Fix TaskChooseOrganization component selector
```
LauraBeatris:
```suggestion
Improve appearance selectors for tasks, such as including it within `SignIn/SignUp` components
```

packages/clerk-js/src/ui/elements/contexts/index.tsx
LauraBeatris: >  was thinking we were actually exposing the standalone <SessionTasks /> component

We discussed about it before, it might happen in the future when we introduce N+1 tasks, so this fix here to introduce the `Flow` component came in the right time.

LauraBeatris (approved):

PR #6913, opened 2025-10-03
feat(shared): React Query variant for useSubscription

panteliselef:
## Description

Since the introduction of the paginated hooks and useSWR our SDKs have been suffering from a cache synchnonization issue, resulting in bad DX and stale UIs.

To help you visualize the issue, imagine a checkout flow, where the developer is using the prebuilt AIO Checkout component and then uses the useSubscription() hook to display the plan name in their custom implemented User button.

If a C2 were to checkout a paid plan and looked at their User button, the plan would still mention “Free” instead of the checkout they just paid for. 

### Our Plan
Per our recent discussions, we decided to move away from SWR and use React Query instead, since it allows us to correctly cache, invalidate and subscribe to changes across the host application and AIOs.

**Migration path**
In order to slowly migrate every hook, we are using **build-time** feature flag in the `@clerk/shared` package that controls which hook implementation is used based on the file extension. For example for the `useSubscription()` hooks we now have 2 files, `useSubscription.rq.ts` and `useSubscription.swr.ts`.

Once all hooks are implemented with React Query, we will start building the `@clerk/shared` with `CLERK_USE_RQ=true` in order surface the React Query implementation in the SDKs.

**clerk-js** 
It now lazy loads `QueryClient` from `@tanstack/query-core`. That means it will only be loaded if requested by the _new_ hooks. 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Release Notes

* **New Features**
  * Added React Query integration for improved data fetching and caching
  * Introduced new subscription hook for accessing billing information
  * Added query client infrastructure for better data synchronization across components and hooks

* **Chores**
  * Added React Query core dependencies to support new query infrastructure


packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -93,7 +93,10 @@ export function ClerkContextProvider(props: ClerkContextProvider) {
     <IsomorphicClerkContext.Provider value={clerkCtx}>
       <ClientContext.Provider value={clientCtx}>
         <SessionContext.Provider value={sessionCtx}>
-          <OrganizationProvider {...organizationCtx.value}>
+          <OrganizationProvider
+            // key={clerkStatus + queryStatus}
```
Ephem: NIT: ✂️

packages/clerk-js/src/core/clerk.ts
```diff
@@ -240,6 +242,28 @@ export class Clerk implements ClerkInterface {
...
 
+  get __internal_queryClient(): { __tag: 'clerk-rq-client'; client: QueryClient } | undefined {
+    if (!this.#queryClient) {
+      void import('./query-core')
+        .then(module => module.QueryClient)
+        .then(QueryClient => {
+          if (this.#queryClient) {
+            return;
+          }
+          this.#queryClient = new QueryClient();
```
Ephem, line 250: I think there are options we might want to consider here, like retry logic. Do we want to always retry all errors, or ignore retrying 4xx kind of errors for example? How many retries should we do by default? Is the default retry delay good? Do we want to match these closely to what we have in SWR today, or change them as part of this?

To be clear, the default RQ ones does the job so I don't think we _need_ to, especially at this early stage, just wanted to mention it.

Ephem, line 250: _If_ we want revalidation on focus and the network status things to work, you will also need to call `queryClient.mount()`. This is normally done in `QueryClientProvider` but since we are skipping that here we'd need to do it manually if we need it.

This is safe but a noop for native environments.

Ephem (commented):
Great work on this, these are just a few random smaller comments I saw when skimming through it.

packages/clerk-js/src/core/clerk.ts
panteliselef: I'd probably turn off retrying completely and adjust later if needed. There's already retrying logic baked into `clerk-js` methods.

panteliselef: Good to know. I think we can circle back to this.

packages/clerk-js/package.json
```diff
@@ -71,6 +71,7 @@
     "@formkit/auto-animate": "^0.8.2",
     "@stripe/stripe-js": "5.6.0",
     "@swc/helpers": "^0.5.17",
+    "@tanstack/query-core": "^5.87.4",
```
nikosdouvlis:
```suggestion
    "@tanstack/query-core": "5.87.4",
```

nikosdouvlis (approved):

PR #6915, opened 2025-10-03
feat(clerk-js, types): Add optional `locale` from browser to signup flow

guilherme6191:
## Description

This PR adds `locale` to signup flow. If it's not passed by users, we retrieve it from the browser. If the value we get from the browser can be determined (trying to narrow it down to BPC 147 spec), we don't infer and set the default to `null`.

Related PRs for extra context, if needed:
- https://github.com/clerk/clerk_go/pull/13957
- https://github.com/clerk/clerk_go/pull/14078

Note: this shouldn't have issues in Expo (the inferring) , but we might want to explore `expo-localization` in the future or as a follow-up for a expo specific solution.

https://linear.app/clerk/issue/USER-3606/retrieve-and-send-locale-information-with-sign-ups
## Checklist

- [x] generate changeset

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- **New Features**
  - Sign-up now accepts, persists and exposes a locale value; browser locale is used when not provided.

- **Tests**
  - Added unit tests for browser-locale detection across multiple scenarios.

- **Documentation**
  - Playground README updated to use localhost:4011.

- **Chores**
  - Added browser-locale utility and re-export; updated fixtures, proxies and public types to include locale; broadened .env ignore rules; release metadata updated.

### DIFF
--- a/packages/clerk-js/src/core/resources/SignUp.ts
+++ b/packages/clerk-js/src/core/resources/SignUp.ts
@@ -686,6 +686,7 @@ class SignUpFuture implements SignUpFutureResource {
       captchaError,
       ...params,
       unsafeMetadata: params.unsafeMetadata ? normalizeUnsafeMetadata(params.unsafeMetadata) : undefined,
+      locale: params.locale ?? getBrowserLocale(),
     };
 
     await this.resource.__internal_basePost({ path: this.resource.pathRoot, body });
@@ -693,15 +694,7 @@ class SignUpFuture implements SignUpFutureResource {
 
   async create(params: SignUpFutureCreateParams): Promise<{ error: unknown }> {
     return runAsyncResourceTask(this.resource, async () => {
-      // Inject browser locale if not already provided
-      const createParams = { ...params };
-      if (!createParams.locale) {
-        const browserLocale = getBrowserLocale();
-        if (browserLocale) {
-          createParams.locale = browserLocale;
-        }
-      }
-      await this._create(createParams);
+      await this._create(params);
     });
   }
 


packages/clerk-js/src/utils/locale.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @returns The browser's reported locale string (typically BCP 47 format like 'en-US', 'es-ES') or null if locale cannot be determined.
+ */
+export function getBrowserLocale(): string | null {
+  if (!inBrowser()) {
+    return DEFAULT_LOCALE;
+  }
+
+  // Get locale from the browser
+  const locale = navigator?.language;
```
tmilewski:
```suggestion
  try {
    const locale = navigator?.language;
  } catch {
    return DEFAULT_LOCALE;
  }
```

We might as well play it safe. Not too sure about Browser Extensions, etc here.

packages/clerk-js/src/test/core-fixtures.ts
```diff
@@ -231,6 +231,8 @@ export const createSignUp = (signUpParams: Partial<SignUpJSON> = {}) => {
     first_name: signUpParams.first_name,
     has_password: signUpParams.has_password,
     last_name: signUpParams.last_name,
+    legal_accepted_at: signUpParams.legal_accepted_at,
```
tmilewski, line 234: good catch!

packages/clerk-js/src/utils/locale.ts
guilherme6191: Done

packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -670,6 +686,7 @@ class SignUpFuture implements SignUpFutureResource {
       captchaError,
       ...params,
       unsafeMetadata: params.unsafeMetadata ? normalizeUnsafeMetadata(params.unsafeMetadata) : undefined,
+      locale: params.locale ?? getBrowserLocale(),
```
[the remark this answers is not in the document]
guilherme6191: I don't think we need that now especially because that's for the FutureSignUp logic with signal. That's experimental and I'm unsure we need to add more surface and update it now.

PR #6939, opened 2025-10-07
feat(upgrade): Add support for @clerk/react

dstaley:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/upgrade/src/codemods/__tests__/__fixtures__/transform-clerk-react-v6.fixtures.js
+++ b/packages/upgrade/src/codemods/__tests__/__fixtures__/transform-clerk-react-v6.fixtures.js
@@ -1,0 +1,66 @@
+export const fixtures = [
+  {
+    name: 'Basic import transform',
+    source: `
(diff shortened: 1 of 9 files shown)


brkalow (approved):

packages/upgrade/src/codemods/__tests__/__fixtures__/transform-clerk-react-v6.fixtures.js
```diff
@@ -0,0 +1,77 @@
+export const fixtures = [
```
jacekradko, line 1: I think the answer is no here, but do we care about codemoding CJS code with require? 

```
const { useSignIn } = require('@clerk/clerk-react');
```

dstaley: ah 💩 you're totally right. I'll update the codemod!

PR #6993, opened 2025-10-15
fix(backend): Support storing multiple local keys in JWK cache

nikosdouvlis:
## Description
Fixed JWT public key caching in `verifyToken()` to support multi-instance scenarios. Public keys are now correctly cached per `kid` from the token header instead of using a single shared cache key.

- Changed `loadClerkJWKFromLocal` to `loadClerkJwkFromPem` for loading keys in PEM format.
- Updated key loading logic in `verifyHandshakeToken` and `verifyToken` functions.
- Enhanced caching mechanism to avoid collisions with remote keys by prefixing local keys with "local-".

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

- Bug Fixes
  - Fixed intermittent token verification failures in multi-instance deployments by isolating JWT key caching per key identifier.
  - Improved reliability when different public keys are used across instances.

- Refactor
  - Consolidated key loading to consistently accept PEM-formatted keys.
  - Improved TypeScript typings for clearer verification option shapes.

- Tests
  - Added tests covering cache isolation, repeated lookups, and mixed-key scenarios.

### DIFF
--- a/packages/backend/src/tokens/handshake.ts
+++ b/packages/backend/src/tokens/handshake.ts
@@ -78,7 +78,7 @@ export async function verifyHandshakeToken(
     });
   }
 
-  return await verifyHandshakeJwt(token, { key });
+  return verifyHandshakeJwt(token, { key });
 }
 
 export class HandshakeService {


packages/backend/src/tokens/handshake.ts
```diff
@@ -78,9 +78,7 @@ export async function verifyHandshakeToken(
     });
   }
 
-  return await verifyHandshakeJwt(token, {
-    key,
-  });
+  return await verifyHandshakeJwt(token, { key });
```
panteliselef: 🙃 
```suggestion
  return verifyHandshakeJwt(token, { key });
```

packages/shared/src/types/utils.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Useful to flatten the type output to improve type hints shown in editors. And also to transform an interface into a type to aide with assignability.
+ * https://github.com/sindresorhus/type-fest/blob/main/source/simplify.d.ts
+ */
+export type Simplify<T> = {
```
panteliselef, line 5: 🙃 I think most devs know it as prettify
```suggestion
export type Prettify<T> = {
```

nikosdouvlis: I used the OG `type-fest` name :)

bratsos (approved):

PR #6997, opened 2025-10-15
feat(clerk-js,types): Add signals support for passkeys

dstaley:
## Description

This PR adds support for sign-in with passkeys to our experimental Signals APIs.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Experimental support for signing in with passkeys (WebAuthn) in the new sign-in flow.
  * Supports autofill/discoverable and standard passkey flows with runtime browser capability checks, credential creation, and submission handling.
  * Adds a user-facing passkey sign-in method across JS and React SDK surfaces.

* **Chores**
  * Bumped package versions via changeset for a minor release.

### DIFF
--- a/.changeset/honest-insects-deny.md
+++ b/.changeset/honest-insects-deny.md
@@ -1,5 +1,6 @@
 ---
 '@clerk/clerk-js': minor
+'@clerk/clerk-react': minor
 '@clerk/types': minor
 ---
 
--- a/packages/react/src/stateProxy.ts
+++ b/packages/react/src/stateProxy.ts
@@ -131,6 +131,7 @@ export class StateProxy implements State {
           'verifyBackupCode',
         ] as const),
         ticket: this.gateMethod(target, 'ticket'),
+        passkey: this.gateMethod(target, 'passkey'),
         web3: this.gateMethod(target, 'web3'),
       },
     };


packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -979,6 +980,77 @@ class SignInFuture implements SignInFutureResource {
...
+      SignIn.clerk.__internal_isWebAuthnAutofillSupported || isWebAuthnAutofillSupportedOnWindow;
+
+    if (!isWebAuthnSupported()) {
+      throw new ClerkWebAuthnError('Passkeys are not supported', {
+        code: 'passkey_not_supported',
+      });
+    }
+
+    return runAsyncResourceTask(this.resource, async () => {
+      if (flow === 'autofill' || flow === 'discoverable') {
```
panteliselef, line 1003: Could we have called `this.resource.authenticateWithPasskey()` instead of duplicating ?

Is the only difference that we don't return the result ?

dstaley: This is following the same pattern we've used in other methods, which is not to create strong interdependencies between `SignIn` and `SignInFuture`, since the goal is to reverse the structure in the next `clerk-js` major. It also allows us to implement divergent behavior between the new APIs and the legacy APIs.

In this case the difference is that `SignInFuture.passkey()` returns a `{ error }` object rather than the `SignIn` resource.

PR #7006, opened 2025-10-16
add signOutCallback to UserButton

jescalan:
## Description

Clerk's `signOut` method offers [a `signOutCallback` param](https://clerk.com/docs/reference/javascript/clerk#sign-out) which can be used for redirects, but can also be used for anything else a developer needs to do or clean up when signing a user out, such as cleaning up local storage, running analytics events, etc.

Previously, we had a prop on `UserButton` that allowed passing in a custom function in this way, but it was [removed in Core 2](https://github.com/clerk/javascript/blob/d8147fb58bfd6caf9a4f0a36fdc48c630d00387f/packages/react/CHANGELOG.md?plain=1#L2033) under the premise that it was only really used for redirecting after sign out and providing an after sign out URL prop would add simplicity and consistency with other component APIs.

I think this was a poorly considered change, as there are a variety of other things that developers could want to do on sign out and this is no longer possible unless you build your own `UserButton` component. We have received customer feedback confirming this need.

This PR re-introduces the `signOutCallback` parameter to `UserButton`, but without changing anything else. If both `signOutCallback` and `afterSignOutUrl` are passed, `signOutCallback` will override the `afterSignOutUrl` redirect behavior.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * UserButton now supports a `signOutCallback` prop to customize post-sign-out behavior, allowing developers to override default navigation with custom handlers for both single and multi-session scenarios.

### DIFF
--- a/packages/clerk-js/src/ui/components/UserButton/__tests__/UserButton.test.tsx
+++ b/packages/clerk-js/src/ui/components/UserButton/__tests__/UserButton.test.tsx
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 
 import { bindCreateFixtures } from '@/test/create-fixtures';
 import { render, waitFor } from '@/test/utils';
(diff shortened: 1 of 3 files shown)


panteliselef:
@jescalan 

UserButton's `afterSignOutUrl` has been marked for deprecation, indicating this should be handled globally. Could the new function live in `ClerkProvider` ?
```ts
  /**
   * Full URL or path to navigate to after sign out is complete
   * @deprecated Configure `afterSignOutUrl` as a global configuration, either in `<ClerkProvider/>` or in `await Clerk.load()`.
   */ 
```

I'd strongly argue that the `signOutCallback` on `signOut` should only be used for navigation purposes. If a developer is using `Clerk.signOut()` they should navigate in callback, and then await for `Clerk.signOut()` to resolve and run any code they want afterwards.


Something to consider is that by adding a function as a prop, the developer would need to refactor to use our component/provider inside a client component, since functions as props are not allowed in server components.

Ephem:
I stumbled across this and have some questions.

@jescalan What's the usecases that has come up? I can imagine things like firing off analytics events, but curious if we've heard others?

> UserButton's afterSignOutUrl has been marked for deprecation, indicating this should be handled globally. Could the new function live in ClerkProvider ?
>
> Something to consider is that by adding a function as a prop, the developer would need to refactor to use our component/provider inside a client component, since functions as props are not allowed in server components.

@panteliselef Aren't these statements incompatible in the case of Next? The provider is and needs to be a server component, so we can't pass in a function there? I might be misunderstanding or missing something which is why I wanted to ask.

jescalan:
Analytics events and "cleanup of local data" from the user that submitted the request, which is reasonable. If they have stuff they put in local storage for example, that they'd want to flush out on sign out, this would make sense

PR #7011, opened 2025-10-16
feat(clerk-js): Detect `locale` from browser and send it to FAPI if exists during sign-in (i18n)

guilherme6191:
## Description

This PR adds auto detection of browser `locale` (navigator.language) and sends that to sign-in flow. This is similar to the behavior implemented for the sign-up (https://github.com/clerk/javascript/pull/6915), but does not expose `locale` in the API to reduce noise/complexity at sign-in for now. 

For extra context: this is a change where we'll be storing users preferred language from the browser, and C1s can update via BAPI (later on via dashboard and AIOs) so we can send localized templates. The whole feature is behind a FF and we're rolling out only for Spark (new customer implementing Clerk).

https://linear.app/clerk/issue/USER-3712/detect-and-send-locale-in-sign-ins-js-lib

**Note: needs https://github.com/clerk/clerk_go/pull/14182 released**

## Checklist

- [x] **needs https://github.com/clerk/clerk_go/pull/14182  released**

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change

## Summary by CodeRabbit

* **New Features**
  * Sign-in flow now detects and automatically sends the browser locale to improve localization during authentication.

* **Tests**
  * Added tests verifying locale is included when present and omitted when absent.
  * Test setup updated for a predictable navigator.language and centralized mock/cleanup to stabilize sign-in test suites.

### DIFF
--- a/packages/clerk-js/src/core/resources/SignIn.ts
+++ b/packages/clerk-js/src/core/resources/SignIn.ts
@@ -168,7 +168,7 @@ export class SignIn extends BaseResource implements SignInResource {
     const locale = getBrowserLocale();
     return this._basePost({
       path: this.pathRoot,
-      body: { ...(locale ? { locale } : {}), ...params },
+      body: locale ? { locale, ...params } : params,
     });
   };
 
@@ -713,7 +713,7 @@ class SignInFuture implements SignInFutureResource {
     const locale = getBrowserLocale();
     await this.resource.__internal_basePost({
       path: this.resource.pathRoot,
-      body: { ...(locale ? { locale } : {}), ...params },
+      body: locale ? { locale, ...params } : params,
     });
   }
 


packages/clerk-js/src/core/resources/__tests__/SignIn.test.ts
```diff
@@ -1,39 +1,94 @@
...
-        if (method === 'POST' && path === '/client/sign_ins/test_id/prepare_first_factor') {
-          return Promise.resolve({
-            client: null,
-            response: firstFactorPreparedJSON,
-          });
-        }
-
-        throw new Error('Unexpected call to BaseResource._fetch');
+      beforeAll(() => {
+        const signInCreatedJSON = {
```
guilherme6191, line 64: with new tests and mocks, this was source of failing tests. So I just added moved the into the beforeAll function so we can have more control over order and scope

.changeset/quick-kings-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
```
guilherme6191, line 2: since there is no change (even backwards compatible) to the sdk api for users, I think we have a `patch` here -- open to make it a `minor` as well

guilherme6191:
Note: I've labeled to `do not merge` for now as we need FAPI changes released - integration tests should fail because of that too.

packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -164,9 +165,10 @@ export class SignIn extends BaseResource implements SignInResource {
 
   create = (params: SignInCreateParams): Promise<SignInResource> => {
     debugLogger.debug('SignIn.create', { id: this.id, strategy: 'strategy' in params ? params.strategy : undefined });
+    const locale = getBrowserLocale();
     return this._basePost({
       path: this.pathRoot,
-      body: params,
+      body: { ...params, ...(locale ? { locale } : {}) },
```
dstaley: I believe this order should be reversed, as we want to prefer the provided locale from `params` over the auto-detected one.

.changeset/quick-kings-rescue.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add browser locale to sign-in flow
```
dstaley:
```suggestion
Add support for automatically sending the browser locale during the sign-in flow
```

packages/clerk-js/src/core/resources/SignIn.ts
guilherme6191: @dstaley, good call. I'm updating the code as it's applicable in case we expose locale as a param as well, which is not the case now.

For now we're not exposing the `locale` to the sign-in API. We initially think that could add complexity for users and we'll revisit the DX if we expose it. It's available in sign-up as that's more appropriate there. Let me know if that makes sense.

FYI the functionality is going to be behind a FF too.

packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -164,9 +165,10 @@ export class SignIn extends BaseResource implements SignInResource {
 
   create = (params: SignInCreateParams): Promise<SignInResource> => {
     debugLogger.debug('SignIn.create', { id: this.id, strategy: 'strategy' in params ? params.strategy : undefined });
+    const locale = getBrowserLocale();
     return this._basePost({
       path: this.pathRoot,
-      body: params,
+      body: { ...(locale ? { locale } : {}), ...params },
```
tmilewski:
```suggestion
      body: locale ? { locale, ...params } : params
```

packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -708,9 +710,10 @@ class SignInFuture implements SignInFutureResource {
   }
 
   private async _create(params: SignInFutureCreateParams): Promise<void> {
+    const locale = getBrowserLocale();
     await this.resource.__internal_basePost({
       path: this.resource.pathRoot,
-      body: params,
+      body: { ...(locale ? { locale } : {}), ...params },
```
tmilewski:
```suggestion
      body: locale ? { locale, ...params } : params
```

packages/clerk-js/src/core/resources/SignIn.ts
tmilewski: 🙃 We can avoid the multiple spreads here and in the other section.

tmilewski (changes requested):

packages/clerk-js/src/core/resources/SignIn.ts
guilherme6191: yeah, I like it 🚀

PR #7015, opened 2025-10-16
fix(shared): Refactor isClerkAPIResponseError to return true only for ClerkApiResponseErrors

nikosdouvlis:
…

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fix**
  * Corrected error-type detection so the helper only identifies the intended API error type.

* **Refactor**
  * Consolidated internal validation logic for error handling to simplify and reuse code.

### DIFF
--- a/.changeset/metal-plants-live.md
+++ b/.changeset/metal-plants-live.md
@@ -1,0 +1,5 @@
+---
+"@clerk/shared": patch
+---
+
+Fix isClerkAPIResponseError so that it returns true only for ClerkApiResponseErrors


panteliselef:
I think this can cause breaking changes, between newer sdk versions and old clerk-js versions. Take [reverification](https://github.com/clerk/javascript/blob/160dc93488ec1c4acb5f544b7c8dd443781f1ee5/packages/shared/src/react/hooks/useReverification.ts#L23) for example. I think that would stop working if there's a mismatch between how we create the error and how we check for it.

PR #7038, opened 2025-10-20
feat(clerk-js,types): Make payment method optional for free trials

mauricioabreu:
## Description

It's flag-based

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Configurable setting to require (or not) a payment method for free trial subscriptions.
  * Checkout UI updated to conditionally show payment-method sections and an alternate "Start free trial" path when payment info isn't required.
  * Checkout completion text now reflects whether a payment method is involved when a free trial ends.

* **Tests**
  * Expanded coverage for free-trial flows across multiple billing configurations.

* **Chores**
  * Added changelog/changeset entry and package version bumps.

### DIFF
--- a/packages/clerk-js/bundlewatch.config.json
+++ b/packages/clerk-js/bundlewatch.config.json
@@ -24,7 +24,7 @@
     { "path": "./dist/waitlist*.js", "maxSize": "1.5KB" },
     { "path": "./dist/keylessPrompt*.js", "maxSize": "6.5KB" },
     { "path": "./dist/pricingTable*.js", "maxSize": "4.02KB" },
-    { "path": "./dist/checkout*.js", "maxSize": "8.77KB" },
+    { "path": "./dist/checkout*.js", "maxSize": "10KB" },
     { "path": "./dist/up-billing-page*.js", "maxSize": "3.0KB" },
     { "path": "./dist/op-billing-page*.js", "maxSize": "3.0KB" },
     { "path": "./dist/up-plans-page*.js", "maxSize": "1.0KB" },


packages/clerk-js/src/core/resources/CommerceSettings.ts
```diff
@@ -29,6 +30,7 @@ export class CommerceSettings extends BaseResource implements CommerceSettingsRe
     }
 
     this.billing.stripePublishableKey = data.billing.stripe_publishable_key || '';
+    this.billing.freeTrialRequiresPaymentMethod = data.billing.free_trial_requires_payment_method ?? true;
```
LauraBeatris, line 33: Is `true` the default value from the API side? I'd let it come from the API instead, and fallback to `false` here

Otherwise, I believe the existing clients with billing would always fallback to `true` here even outside of free trials context 

I might be totally wrong tho, just a disclaimer that I don't have a good understanding of billing overall 😓

packages/clerk-js/src/ui/components/Checkout/CheckoutForm.tsx
```diff
@@ -180,10 +180,13 @@ const useCheckoutMutations = () => {
       useTestCard: true,
     });
 
+  const subscribeWithoutPaymentMethod = () => confirmCheckout({});
```
LauraBeatris: Nit: I can see that `ConfirmCheckoutParams` type is a union with inner sets of optional properties, therefore passing an empty object works here 

I'd suggest allowing passing `undefined` here such as `confirmCheckout()` instead

packages/clerk-js/src/ui/components/Checkout/CheckoutForm.tsx
```diff
@@ -395,7 +411,10 @@ const ExistingPaymentMethodForm = withCardStateProvider(
       });
     }, [paymentMethods]);
 
-    const showPaymentMethods = isImmediatePlanChange && (totalDueNow.amount > 0 || !!freeTrialEndsAt);
+    const showPaymentMethods =
+      isImmediatePlanChange &&
+      (totalDueNow.amount > 0 ||
+        (!!freeTrialEndsAt && environment.commerceSettings.billing.freeTrialRequiresPaymentMethod));
```
LauraBeatris, line 456: Idea: Extract this UI logic to a utility function on this same module, since it gets repeated between two components

packages/clerk-js/src/ui/components/Checkout/CheckoutForm.tsx
```diff
@@ -462,3 +481,21 @@ const ExistingPaymentMethodForm = withCardStateProvider(
...
+    <Button
+      elementDescriptor={descriptors.formButtonPrimary}
+      onClick={subscribeWithoutPaymentMethod}
+      sx={{
+        width: '100%',
+      }}
+      isLoading={card.isLoading}
+      localizationKey={submitLabel}
+    />
+  );
```
LauraBeatris: I believe we might need to add a unique `elementDescriptor` for this button in case developers want to apply a specific styling to it

packages/clerk-js/src/core/resources/CommerceSettings.ts
mauricioabreu: Yes, the default value is true. It's because many users abuse free trials that don't require a payment method.

packages/types/src/commerceSettings.ts
```diff
@@ -5,6 +5,7 @@ import type { CommerceSettingsJSONSnapshot } from './snapshots';
 export interface CommerceSettingsJSON extends ClerkResourceJSON {
   billing: {
     stripe_publishable_key: string;
+    free_trial_requires_payment_method?: boolean;
```
panteliselef: Is there a specific reason, we want this to be optional ? Can't fapi always return a boolean ?

panteliselef:
@mauricioabreu take a look at my changes in #7065, one test was failing, plus I did some minor improvements.

packages/types/src/commerceSettings.ts
mauricioabreu: It does not need to be optional. I'll fix it.

mauricioabreu: Solved via a29eff7e2c6bd78f4682af6abbdd85235026c152

packages/clerk-js/src/ui/components/Checkout/CheckoutForm.tsx
mauricioabreu: Good idea! 
Pantelis did it 👏🏼

packages/clerk-js/src/ui/components/Checkout/CheckoutForm.tsx
mauricioabreu: Pantelis fixed it too 🙏🏼

PR #7088, opened 2025-10-29
feat: Remove deprecated SDK props

jacekradko:
## Description

Removing deprecated props

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/shared/src/types/clerk.ts
```diff
@@ -970,28 +968,12 @@ type ClerkOptionsNavigation =
...
-  afterSignInUrl?: string | null;
-  /**
-   * @deprecated Use `signUpFallbackRedirectUrl` or `signUpForceRedirectUrl` instead.
-   */
-  afterSignUpUrl?: string | null;
-  /**
-   * @deprecated Use `signInFallbackRedirectUrl`, `signInForceRedirectUrl`, `signUpFallbackRedirectUrl`, or `signUpForceRedirectUrl` instead.
-   */
-  redirectUrl?: string | null;
-};
```
panteliselef, line 1031: Doesn't this require a clerk-js major ? Are we doing that ?

jacekradko: Yep. I can split the clerk-js specific changes to a separate PR in case we stay on v5.

PR #7101, opened 2025-10-30
fix(clerk-js): Navigate to `/sign-up/continue` with ticket strategy

LauraBeatris:
## Description

We caught a bug on IdP-Initiated flows, with legal consent enabled, where FAPI would redirect to `/sign-up` but the form still gets shown with password even tho it's not required for enterprise SSO.

For those cases, where sign-up with ticket has `missing_requirements` status, then it should navigate to `/sign-up/continue` route to show the missing fields.

https://github.com/user-attachments/assets/e1b6f8e8-3b43-4688-b0bc-61bd605b84f8

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Sign-up flows that previously stalled when additional steps (e.g., accepting legal consent) were required now redirect users to a dedicated "continue" sign-up step so they can complete registration.
  * This fix also resolves stalled identity-provider (IdP) initiated sign-ups by routing users to the same continue step when extra requirements are detected.

### DIFF
--- a/packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
+++ b/packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
@@ -1,7 +1,7 @@
 import { getAlternativePhoneCodeProviderData } from '@clerk/shared/alternativePhoneCode';
 import { useClerk } from '@clerk/shared/react';
 import type { PhoneCodeChannel, PhoneCodeChannelData, SignUpResource } from '@clerk/shared/types';
-import React, { useMemo } from 'react';
+import React from 'react';
 
 import { isClerkAPIResponseError } from '@/index.headless';
 import { Card } from '@/ui/elements/Card';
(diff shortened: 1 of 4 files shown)


packages/clerk-js/src/ui/components/SignUp/SignUpStart.tsx
```diff
@@ -130,6 +130,15 @@ function SignUpStartInternal(): JSX.Element {
...
 
+  const onlyLegalAcceptedMissing = useMemo(
+    () =>
+      signUp.missingFields &&
+      signUp.missingFields.length === 1 &&
+      signUp.missingFields[0] === 'legal_accepted' &&
+      signUp.unverifiedFields &&
+      signUp.unverifiedFields.length === 0,
+    [signUp.missingFields, signUp.unverifiedFields],
+  );
```
LauraBeatris: This is already being done on `SignUpContinue`, but wasn't handled on the root sign-up

LauraBeatris:
We've done some testing and the password field is being sent as empty even tho is not being displayed on the sign-up form. 

I'll update the changes here soon to address that.

PR #7118, opened 2025-10-31
fix(clerk-js,types): Add hasBeenFinalized to prevent null updates

dstaley:
## Description

This PR fixes an issue where `SignIn(null)` was propagated to our signals when the `signIn` was completed. This is due to client piggybacking. Now, the `SignIn(null)` will only be propagated if `hasBeenFinalized` is true, which indicates that `signIn.finalize` was called on the instance (and that it can be safely overwritten since there's nothing left to do to it).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -900,6 +906,7 @@ class SignUpFuture implements SignUpFutureResource {
         throw new Error('Cannot finalize sign-up without a created session.');
       }
 
+      this.#hasBeenFinalized = true;
```
panteliselef, line 909: Shouldn't we reset this back to `false` at some point ?

dstaley: A new `SignIn` instance is created every time we get a new response from the API via piggybacking, which will get the default value of `false`. Setting this value to `true` allows the sign-in signal to be overwritten with the next incoming update (which will have `hasBeenFinalized` set to `false`).

It was actually surprising to me to learn that we don't maintain a consistent reference to a `SignIn` instance. Anytime we get a new client response (so, basically every API request), we call `new SignIn(response.client.signIn)`, creating a new `SignIn` instance.

brkalow (approved):

PR #6905, opened 2025-10-02
spike: batch state updates in setActive

panteliselef:
## Checklist

- [ ] `pnpm test` runs as expected.
- [ ] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Summary by CodeRabbit

- Refactor
  - Transition-aware navigation and auth flows to reduce UI jank during sign-in/out and redirects.
  - Deferred state handling in context to improve responsiveness and timing of state updates after navigation.

- Chores
  - Minor version bump for @clerk/clerk-js, @clerk/nextjs, @clerk/clerk-react, and @clerk/types.

PR #7157, opened 2025-11-05
fix(clerk-react): Correct useAuth().isLoaded in transitive state

jacekradko:
## Description

When isomorphicClerk is loaded, we should return `isLoaded: false` from `useAuth()` when transitive state is set. 

The check ensures we only override auth state with initial values when Clerk SDK hasn't loaded yet. Once loaded, we respect the actual auth state, even if it's transitioning, so components know to wait until session data is fully resolved.

Screenshot 2025-11-05 at 3 09 16 PM

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication hook now reports isLoaded = false when the auth system is in a transitive/loading state to avoid showing stale auth status.

* **Tests**
  * Added a test verifying the hook returns isLoaded = false and no session/user IDs while auth is transitive.

* **Chores**
  * Added release metadata for a patch version bump.


jacekradko:
!snapshot

clerk-cookie:
Hey @jacekradko - the snapshot version command generated the following package versions:
| Package | Version |
| --- | --- |
| @clerk/agent-toolkit | 0.1.47-snapshot.v20251105162613 |
| @clerk/astro | 2.14.5-snapshot.v20251105162613 |
| @clerk/backend | 2.19.4-snapshot.v20251105162613 |
| @clerk/chrome-extension | 2.7.13-snapshot.v20251105162613 |
| @clerk/clerk-js | 5.104.1-snapshot.v20251105162613 |
| @clerk/elements | 0.23.78-snapshot.v20251105162613 |
| @clerk/clerk-expo | 2.18.2-snapshot.v20251105162613 |
| @clerk/expo-passkeys | 0.4.15-snapshot.v20251105162613 |
| @clerk/express | 1.7.46-snapshot.v20251105162613 |
| @clerk/fastify | 2.5.3-snapshot.v20251105162613 |
| @clerk/localizations | 3.26.7-snapshot.v20251105162613 |
| @clerk/nextjs | 6.34.4-snapshot.v20251105162613 |
| @clerk/nuxt | 1.11.5-snapshot.v20251105162613 |
| @clerk/clerk-react | 5.53.7-snapshot.v20251105162613 |
| @clerk/react-router | 2.1.9-snapshot.v20251105162613 |
| @clerk/remix | 4.13.12-snapshot.v20251105162613 |
| @clerk/shared | 3.31.1-snapshot.v20251105162613 |
| @clerk/tanstack-react-start | 0.26.8-snapshot.v20251105162613 |
| @clerk/testing | 1.13.12-snapshot.v20251105162613 |
| @clerk/themes | 2.4.33-snapshot.v20251105162613 |
| @clerk/types | 4.97.2-snapshot.v20251105162613 |
| @clerk/vue | 1.14.12-snapshot.v20251105162613 |

`@clerk/agent-toolkit`
```sh
npm i @clerk/agent-toolkit@0.1.47-snapshot.v20251105162613 --save-exact
```
`@clerk/astro`
```sh
npm i @clerk/astro@2.14.5-snapshot.v20251105162613 --save-exact
```
`@clerk/backend`
```sh
npm i @clerk/backend@2.19.4-snapshot.v20251105162613 --save-exact
```
`@clerk/chrome-extension`
```sh
npm i @clerk/chrome-extension@2.7.13-snapshot.v20251105162613 --save-exact
```
`@clerk/clerk-js`
```sh
npm i @clerk/clerk-js@5.104.1-snapshot.v20251105162613 --save-exact
```
`@clerk/elements`
```sh
npm i @clerk/elements@0.23.78-snapshot.v20251105162613 --save-exact
```
`@clerk/clerk-expo`
```sh
npm i @clerk/clerk-expo@2.18.2-snapshot.v20251105162613 --save-exact
```
`@clerk/expo-passkeys`
```sh
npm i @clerk/expo-passkeys@0.4.15-snapshot.v20251105162613 --save-exact
```
`@clerk/express`
```sh
npm i @clerk/express@1.7.46-snapshot.v20251105162613 --save-exact
```
`@clerk/fastify`
```sh
npm i @clerk/fastify@2.5.3-snapshot.v20251105162613 --save-exact
```
`@clerk/localizations`
```sh
npm i @clerk/localizations@3.26.7-snapshot.v20251105162613 --save-exact
```
`@clerk/nextjs`
```sh
npm i @clerk/nextjs@6.34.4-snapshot.v20251105162613 --save-exact
```
`@clerk/nuxt`
```sh
npm i @clerk/nuxt@1.11.5-snapshot.v20251105162613 --save-exact
```
`@clerk/clerk-react`
```sh
npm i @clerk/clerk-react@5.53.7-snapshot.v20251105162613 --save-exact
```
`@clerk/react-router`
```sh
npm i @clerk/react-router@2.1.9-snapshot.v20251105162613 --save-exact
```
`@clerk/remix`
```sh
npm i @clerk/remix@4.13.12-snapshot.v20251105162613 --save-exact
```
`@clerk/shared`
```sh
npm i @clerk/shared@3.31.1-snapshot.v20251105162613 --save-exact
```
`@clerk/tanstack-react-start`
```sh
npm i @clerk/tanstack-react-start@0.26.8-snapshot.v20251105162613 --save-exact
```
`@clerk/testing`
```sh
npm i @clerk/testing@1.13.12-snapshot.v20251105162613 --save-exact
```
`@clerk/themes`
```sh
npm i @clerk/themes@2.4.33-snapshot.v20251105162613 --save-exact
```
`@clerk/types`
```sh
npm i @clerk/types@4.97.2-snapshot.v20251105162613 --save-exact
```
`@clerk/vue`
```sh
npm i @clerk/vue@1.14.12-snapshot.v20251105162613 --save-exact
```

packages/react/src/hooks/__tests__/useAuth.test.tsx
```diff
@@ -317,4 +317,30 @@ describe('useDerivedAuth', () => {
...
+      userId: undefined,
+      signOut: vi.fn(),
+      getToken: vi.fn(),
+    };
+
+    const {
+      result: { current },
+    } = renderHook(() => useDerivedAuth(authObject));
+
+    expect(current.isLoaded).toBe(false);
```
panteliselef: was that `true` before ?

jacekradko: Yeah. 

```
 FAIL  src/hooks/__tests__/useAuth.test.tsx > useDerivedAuth > returns not loaded state during transitive state
AssertionError: expected true to be false // Object.is equality

- Expected
+ Received

- false
+ true

 ❯ src/hooks/__tests__/useAuth.test.tsx:341:30
    339|     } = renderHook(() => useDerivedAuth(authObject));
    340| 
    341|     expect(current.isLoaded).toBe(false);
       |                              ^
    342|     expect(current.isSignedIn).toBe(false);
    343|     expect(current.sessionId).toBe(null);
```

packages/react/src/hooks/useAuth.ts
```diff
@@ -99,13 +99,13 @@ export const useAuth = (initialAuthStateOrOptions: UseAuthOptions = {}): UseAuth
   const initialAuthState = rest as any;
 
   const authContextFromHook = useAuthContext();
+  const isomorphicClerk = useIsomorphicClerkContext();
   let authContext = authContextFromHook;
 
-  if (authContext.sessionId === undefined && authContext.userId === undefined) {
+  if (!isomorphicClerk.loaded && authContext.sessionId === undefined && authContext.userId === undefined) {
     authContext = initialAuthState != null ? initialAuthState : {};
   }
```
panteliselef: Should we flip this to 
```suggestion
  if (!isomorphicClerk.loaded && initialAuthState) {
    authContext = initialAuthState
  }
```

But also i'd expect that `authContextFromHook` has the correct data inside during SSR, since it get populated corrected in [here](https://github.com/clerk/javascript/blob/83eeb56de0d0e903dcb1d406b7668a6d11371f1e/packages/react/src/contexts/ClerkContextProvider.tsx#L40)

panteliselef: Is there a benefit when setting authContext with `{}` ?

jacekradko: @panteliselef I think that makes sense: [`044d9ca` (#7157)](https://github.com/clerk/javascript/pull/7157/commits/044d9ca23d08086351d292f2d4fe79ca35e6a8aa)

packages/react/src/hooks/useAuth.ts
```diff
@@ -99,13 +99,9 @@ export const useAuth = (initialAuthStateOrOptions: UseAuthOptions = {}): UseAuth
   const initialAuthState = rest as any;
 
   const authContextFromHook = useAuthContext();
-  let authContext = authContextFromHook;
-
-  if (authContext.sessionId === undefined && authContext.userId === undefined) {
-    authContext = initialAuthState != null ? initialAuthState : {};
-  }
```
Ephem, line 106: Oh, so previously it was falling back to the initial auth state when `sessionId` and `userId` was `undefined`? This was likely fine in CSR apps as the initial state there would be `isLoading: true`, but not so fine in SSRd apps.

I think you might have stumbled on the root cause of the sign in bug here too! When signing in, the initial state would be `sessionId: null`, user is signed out. When sign in transitions to the transitive state, `sessionId` and `userId` becomes `undefined`, so this would fall back to the _initial_ state, which is `null`.

packages/react/src/hooks/useAuth.ts
```diff
@@ -99,13 +99,9 @@ export const useAuth = (initialAuthStateOrOptions: UseAuthOptions = {}): UseAuth
...
-
   const isomorphicClerk = useIsomorphicClerkContext();
+  const authContext = !isomorphicClerk.loaded && initialAuthState ? initialAuthState : authContextFromHook;
```
Ephem, line 103: This is a separate existing issue, but since you are touching the code I thought I might comment on it in context, not suggesting we tackle it here since it's a bigger one.

With SSR, `initialAuthState` is always what will have been used. With this code, if Clerk has loaded _before_ this part of the application has hydrated (selective hydration can defer hydrating Suspense boundaries), this would use the hook state, which might differ from the `initialAuthState`, causing hydration mismatches. This is probably most likely with static SSR builds.

The canonical fix would be to use `useSyncExternalStore` and pass the initial state to the `getServerSnapshot` function. Since we are using context here there's nothing to subscribe to though, so not sure how that plays out.

We also can't use `useSyncExternalStore` in the ClerkProvider (where we _do_ listen to the Clerk store), since that wouldn't work with selective hydration, it would only use `getServerSnapshot` when _that_ part of the application was hydrating.

packages/react/src/hooks/useAuth.ts
Ephem: Yep, this PR fixes the sign in bug independently of the other one:

CleanShot 2025-11-06 at 10 53 06

packages/react/src/hooks/__tests__/useAuth.test.tsx
```diff
@@ -77,6 +77,39 @@ describe('useAuth', () => {
...
+      orgPermissions: undefined,
+      orgRole: undefined,
+      orgSlug: undefined,
+      sessionClaims: null,
+      sessionId: undefined,
+      sessionStatus: undefined,
+      userId: undefined,
+    };
+
+    const { result } = renderHook(() => useAuth(), {
```
Ephem, line 100: This test is passing even with the existing logic for me.

I think the prerequisite for the bug to happen is that we explicitly pass in `initialAuthState` (like we do from `usePromisifiedAuth`, which is the reason this is failing in Next).

For a test that fails with the old code and pass with the new (might want to update the test description too):

```suggestion
    const mockInitialAuthState = {
      sessionId: null,
      userId: null,
    };

    // Test that we don't fall back to the initial state when
    // isomorphicClerk is loaded but in transitive state
    const { result } = renderHook(() => useAuth(mockInitialAuthState), {
```

We could also keep this test as is since it covers a new case we didn't have coverage for, and add a new test with the above.

Overall, I think several tests here might benefit from the test approach you are using here btw, `useAuth` is the main public API so we should be testing that directly instead of `useDerivedAuth`.

(One could also argue we should move/duplicate this test to `usePromisifiedAuth` since that's really the broken public API here, but I don't think we should do that right now)

packages/react/src/hooks/useAuth.ts
Ephem, line 103: I just want to verify my understanding here. The reason this is safe is because:

* `isomorphicClerk` changes reference when `status` changes, so it's reactive specifically to that value
* `IsomorphicClerk.hydrateClerkJS` _first_ emits the state update, and _then_ emits the status update
  * Meaning `authContextFromHook` is guaranteed to already be set when `loaded` changes from `false`->`true`?
* `isomorphicClerk.loaded` never goes `true`->`false`

Correct? (These are a lot of hidden assumptions/guarantees btw)

What happens if we want to start making the auth state updates in transitions as per #6905? I think it might be possible we get a race condition where `loaded: true` but we don't have the `authContextFromHook` yet? I don't think we need or should figure that out now, just wanted to point it out as some complexity to watch for. Solution might be to do the entire hydration as a transition.

@panteliselef FYI ☝️

panteliselef: You are spot on regarding the assumptions. We can of course change how things work if we want to support transitions.

panteliselef (approved):
Stamping it, but I still think the solution of 
```tsx
const authContext = !isomorphicClerk.loaded && initialAuthState ? initialAuthState : authContextFromHook;
```

is a noop and we can already trust the value of `authContextFromHook`

Ephem (approved):
I'm not sure the test is testing the correct thing, but I think the fix itself is good to go! 👍

PR #7170, opened 2025-11-06
poc: useAuth with uSES

jacekradko:
## Description

POC of useSyncExternalStore in useAuth to get around hydration mismatches on initial SSR -> CSR transfer

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Extended authentication context with organization custom permissions support
  * Added factor verification age tracking to enhance security state management

* **Refactor**
  * Modernized authentication state management with improved server-side rendering support
  * Enhanced real-time authentication synchronization for better state consistency across the application


packages/react/src/hooks/useAuth.ts
```diff
@@ -95,17 +95,16 @@ type UseAuthOptions = Record<string, any> | PendingSessionOptions | undefined |
...
-  const initialAuthState = rest as any;
+  const { treatPendingAsSignedOut } = initialAuthStateOrOptions ?? {};
 
-  const authContextFromHook = useAuthContext();
-  let authContext = authContextFromHook;
+  const isomorphicClerk = useIsomorphicClerkContext();
 
-  if (authContext.sessionId === undefined && authContext.userId === undefined) {
-    authContext = initialAuthState != null ? initialAuthState : {};
-  }
```
jacekradko, line 106: No need to see if we can trust the authContext. We can just the the authStore directly here

packages/react/src/stores/authStore.ts
Ephem, line 1: Isn't `clerk` already our auth store? Introducing a new intermediary one we need to sync to seems unnecessary?

packages/react/src/contexts/ClerkContextProvider.tsx
```diff
@@ -35,58 +36,47 @@ export function ClerkContextProvider(props: ClerkContextProvider) {
...
-  const userCtx = React.useMemo(() => ({ value: user }), [userId, user]);
-  const organizationCtx = React.useMemo(() => {
-    const value = {
-      organization: organization,
+  const { session, user, organization } = derivedState;
+
+  // Set server snapshot for SSR/hydration and connect to Clerk for live updates
+  React.useLayoutEffect(() => {
+    if (initialState) {
+      authStore.setServerSnapshot({
```
Ephem, line 48: This wont work since like `useEffect`, `useLayoutEffect` doesn't run on the server.

The only way to make SSR work with something like this currently is to call this function _in render_. This is technically fine as long as the `serverSnapshot` is not already set, since that counts as seeding a cache, but I'm not sure it's the best approach.

Consider also the case of:

```tsx
<ClerkProvider>
  <ClerkProvider dynamic>
```

in Next. With a singleton `authStore`, we can only have a single server snapshot, when what we want is one _per subtree_. I think this means we want to leverage React context. Somewhat like [this PR](https://github.com/clerk/javascript/pull/7194) does for `initialState`.

I envision it like: `const serverSnapshot = deriveServerSnapshot(initialState)`.

PR #7194, opened 2025-11-10
feat(react): Remove `initialAuthState` option from `useAuth`

Ephem:
## Description

> [!note]
> Keeping this with the `do not merge` label until I've done some more testing and verifying.

This PR refactors the React `ClerkProvider` by introducing a new `InitialAuthStateProvider`, and moves resolving the correct `initialState | state` down into the `useAuth` hook. This is a first step towards moving more logic down into the hooks.

This PR also refactors the `ClerkProvider` in `@clerk/nextjs` to not rely on `PromisifiedAuthContext` and removes the `usePromisifiedAuth`. When nesting two providers like this:

```tsx
<ClerkProvider>
  <ClerkProvider dynamic>
```

The inner provider does not render the _React_ `ClerkProvider`. For this specific case, the `PromisifiedAuthContext` was fulfilling a role as a _separate_ initial state context. This PR instead uses the new `InitialAuthStateProvider` directly for this use case.

The `usePromisifiedAuth` hook was the only place `useAuth(initialAuthState)` was being used, an option specifically  option, we now rely on the internal `<InitialAuthStateProvider>` instead.

This is technically a breaking change, but I expect few if any users to be relying on this option, the encouraged way to populate auth state ahead of Clerk loading on the client is to do it at the `<ClerkProvider>` level.

We are planning improvements around how hydration of the auth state works, and keeping the API surface small and implementation complexity low will help with that.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:


packages/nextjs/src/app-router/server/ClerkProvider.tsx
```diff
@@ -28,38 +27,41 @@ const getNonceHeaders = React.cache(async function getNonceHeaders() {
...
+  if (!dynamic) {
+    return Promise.resolve('');
+  }
+  if (isNext13) {
+    /**
+     * For some reason, Next 13 requires that functions which call `headers()` are awaited where they are invoked.
+     * Without the await here, Next will throw a DynamicServerError during build.
+     */
+    return Promise.resolve(await getNonceHeaders());
+  }
```
panteliselef: We can drop those in theory drop those, since as part of the release we will be dropping support for v13 and v14. 

I recognise that we may want to make minimal changes as part of this PR and address this after the realise. In that case I'd make sure to document this a ticket and come back to it later.

Ephem: That's good context, I think it makes sense to drop those in this PR. 👍 

I'm also considering changing `return Promise.resolve(null);` to `return null;` and some other small refactors.

Ephem: Both of these are done.

PR #7197, opened 2025-11-10
feat(nextjs): Drop support for next v13 and v14

panteliselef:
## Description

This PR raises the minimum supported Next.js version to 15.2.3, removing most legacy compatibility code for Next.js 13 and 14.

- Removed the `startTransition` wrapper around `router.refresh()` that was only needed for Next.js 13. Dropped the `__clerk_internal_invalidateCachePromise` mechanism that coordinated promise resolution with transition state. Removed the `isPending` state tracking from `useTransition` since we no longer wrap operations in transitions. The __unstable__onBeforeSetActive hook now directly calls invalidateCacheAction() (or skips it for Next.js 15+ sign-outs).
- Cleaned up the deprecation warning that alerted users about Next.js 13.x and 14.0.x support ending
- Cleaned up version detection helpers (isNext13, isNextWithUnstableServerActions) that gated legacy behavior
- Simplified keyless feature flag: No longer needs to check for unstable server action versions
- Removed version checks in `useInternalNavFun` for handling shallow routing state
- New peerDependencies: **"next": "^15.2.3 || ^16"**

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -298,9 +298,6 @@ jobs:
         include:
           - test-name: "nextjs"
             test-project: "chrome"
-            next-version: "14"
-          - test-name: "nextjs"
-            test-project: "chrome"
             next-version: "15"
           - test-name: "nextjs"
             test-project: "chrome"


panteliselef:
!allow-major

PR #7254, opened 2025-11-18
fix(clerk-js): Re-throw network errors in FraudProtection

chriscanin:
## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network error handling in fraud protection so transient network failures now fall back to cached responses instead of triggering unnecessary captcha challenges.

* **Chores**
  * Applied a small patch release to ensure the fix is delivered consistently to users.

### DIFF
--- a/.changeset/fix-network-error-cache-fallback.md
+++ b/.changeset/fix-network-error-cache-fallback.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Fix issue where network errors were being masked by fraud protection logic, preventing cache fallback from triggering properly.


tmilewski (approved):

chriscanin:
The app crashed when starting offline with __experimental_resourceCache enabled due to three interconnected issues:
* Missing protect_config field from FAPI: The Frontend API doesn't return protect_config in the environment response, causing it to be undefined.
* Undefined values in cache serialization: When protectConfig.loaders was undefined, the __internal_toSnapshot() method would serialize it as undefined, which can't be properly deserialized.
* Narrow error handling: The cache fallback logic only triggered for ClerkRuntimeError with code 'network_error', but offline startup threw a generic TypeError, preventing cache fallback.

How It Manifested:
* Online: App loads successfully, FAPI returns environment without protect_config.
* Offline on next launch: App tries to load from cache, encounters TypeError: Cannot convert undefined value to object. The error isn't caught as a network error, so cache fallback doesn't trigger → CRASH.

dstaley:
If `protect_config` can be undefined, we should also update the `EnvironmentJSON` type in `shared` to reflect that. I'm pinging some folks on the backend side to make sure this was intentional as no other value on the environment response is undefinable.

packages/clerk-js/src/core/resources/Environment.ts
```diff
@@ -54,9 +53,9 @@ export class Environment extends BaseResource implements EnvironmentResource {
     this.clientDebugMode = this.withDefault(data.client_debug_mode, this.clientDebugMode);
     this.organizationSettings = new OrganizationSettings(data.organization_settings);
     this.userSettings = new UserSettings(data.user_settings);
-    this.commerceSettings = new CommerceSettings(data.commerce_settings);
-    this.apiKeysSettings = new APIKeySettings(data.api_keys_settings);
-    this.protectConfig = new ProtectConfig(data.protect_config);
+    this.commerceSettings = new CommerceSettings(data.commerce_settings ?? null);
+    this.apiKeysSettings = new APIKeySettings(data.api_keys_settings ?? null);
```
dstaley: Looking at the backend code for these fields, I don't think they're potentially undefined. Did you run into that scenario when testing?

chriscanin: Ahrggg. I thought I had taken this out... maybe not pushed from git issues? one moment. 

The changes here should not be needed.

chriscanin: I guess this was just left over from testing. Good catch here. 

I have pushed a fix.

packages/clerk-js/src/core/resources/Environment.ts
```diff
@@ -56,7 +55,7 @@ export class Environment extends BaseResource implements EnvironmentResource {
     this.userSettings = new UserSettings(data.user_settings);
     this.commerceSettings = new CommerceSettings(data.commerce_settings);
     this.apiKeysSettings = new APIKeySettings(data.api_keys_settings);
-    this.protectConfig = new ProtectConfig(data.protect_config);
+    this.protectConfig = new ProtectConfig(data.protect_config ?? null);
```
chriscanin: for this one the ProtectConfig constructor expects: ProtectConfigJSON | ProtectConfigJSONSnapshot | null, not
  undefined, or we could add | null | undefined to the ProtectConfig constructor, but it will receive undefined if we are not connected 


  When Online (Normal Case):
  1. FAPI returns environment data
  2. protect_config field is missing from FAPI response (not in
  the Go struct) (can we alter this? should we? it didn't seem like this was the right approach). 
  3. data.protect_config = undefined
  4. new ProtectConfig(undefined) is called
  5. Constructor accepts | undefined, so no type error
  6. fromJSON checks if (!data) which catches undefined
  7. Returns early with default empty ProtectConfig (id: '',
  loaders: undefined)
  8. When serialized to cache: loaders ?? [] ensures it becomes []
   not undefined

  When Offline (Cache Case):
  1. Load from cache (DUMMY data or previously cached data)
  2. Cached data has protect_config: { object: 'protect_config', 
  id: '', loaders: [] }
  3. data.protect_config = { object: 'protect_config', id: '', 
  loaders: [] }
  4. new ProtectConfig({ ... }) is called with actual data
  5. fromJSON processes it normally
  6. Everything works

chriscanin:
> If `protect_config` can be undefined, we should also update the `EnvironmentJSON` type in `shared` to reflect that. I'm pinging some folks on the backend side to make sure this was intentional as no other value on the environment response is undefinable.

Looking into this.

chriscanin:
Before:
https://github.com/user-attachments/assets/c26e4407-a01e-475c-bd1d-2297258381e5



And after:
https://github.com/user-attachments/assets/b66beaa4-5fd1-482d-91a5-e0991fe050f3

chriscanin:
When the app starts offline, Clerk attempts to fetch environment and client data from FAPI, but the network requests fail, throwing a ClerkRuntimeError with code: network_error. This error gets caught by FraudProtection.execute(), which wraps all FAPI requests to handle captcha challenges. The fraud protection code checks if it's a ClerkAPIResponseError (which network errors are classified as) and then tries to access e.errors[0]?.code to determine if it's a captcha challenge.
However, network errors do not have an errors array that only exists on actual API validation errors. Accessing this undefined property causes a new TypeError: Cannot convert undefined value to object (this led me down the wrong path), which masks the original network error. Because the error type changed, Clerk's cache fallback logic in clerk.ts no longer recognizes it as a network error, so it never attempts to load from cache, causing the app to crash instead of working offline.

 We added a simple check in FraudProtection.execute() at:
 /Users/chris/workspace/javascript/packages/clerk-js/src/core/fraudProtection.ts:48-52
This check detects network errors before trying to access the errors array. When a ClerkRuntimeError with code network_error is caught, it is immediately re-thrown, bypassing the captcha logic entirely.
This allows the network error to propagate correctly back to clerk.ts, where it is recognized as a real network error and triggers the cache fallback mechanism. As a result, the app correctly loads the user's authentication state from cache and works offline without crashing.

Mobile is now working with this change.

PR #7264, opened 2025-11-20
feat(ui): Switch to ESM only builds

nikosdouvlis:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/rspack.config.js
```diff
@@ -1,10 +1,14 @@
...
+import rspack from '@rspack/core';
+import packageJSON from './package.json' with { type: 'json' };
+import path from 'path';
+import { fileURLToPath } from 'url';
+import { merge } from 'webpack-merge';
+import ReactRefreshPlugin from '@rspack/plugin-react-refresh';
+import { svgLoader, typescriptLoaderProd, typescriptLoaderDev } from '../../scripts/rspack-common.js';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
panteliselef, line 12: @nikosdouvlis do we still need this file ?

packages/ui/package.json
```diff
@@ -37,14 +35,14 @@
...
-    "build:bundle": "rspack build --config rspack.config.js --env production",
+    "build:analyze": "rspack build --config rspack.config.mjs --env production --analyze",
+    "build:bundle": "rspack build --config rspack.config.mjs --env production",
     "build:declarations": "tsc -p tsconfig.declarations.json",
     "bundlewatch": "FORCE_COLOR=1 bundlewatch --config bundlewatch.config.json",
     "bundlewatch:fix": "node bundlewatch-fix.mjs",
     "clean": "rimraf ./dist",
-    "dev": "rspack serve --config rspack.config.js",
-    "dev:origin": "rspack serve --config rspack.config.js --env devOrigin=http://localhost:${PORT:-4001}",
+    "dev": "rspack serve --config rspack.config.mjs",
```
dstaley, line 44: `"type": "module"` will treat all files in the current directory as ESM by default, so shouldn't this be `rspack.config.js`? I don't think we need the `.mjs` extension.

packages/ui/rspack.config.js
nikosdouvlis: Tagging @dstaley to answer both questions at the same time :) 
 rspack.config.mjs should be rspack.config.js - I think I managed to stage only some of the changes and the delta here is weird. Pushing a fix in a bit

PR #7267, opened 2025-11-20
poc: Remove contexts and subscribe to clerk changes directly in hooks

Ephem:
## Description

This is a proof of concept that reimagines a few things. It builds on #7194 but takes things quite a bit further.

> [!NOTE]
> I'm targeting this directly at `vincent-and-the-doctor` and not the PR above. That means this PR contains some of the same changes as that one, but since it also reworks things quite a bit I choose to target it this way so this can be reviewed entirely stand alone.

There are a lot of details in here, but in essence, what this PR does is:

* Merge the `ContextProviders` from `ui` and `react` into `shared/react`
* Removes the `clerk.addListener()` in that provider, also removes most contexts
  * Breaking change: Several context providers like `UserContext` are no longer exported from `shared/react`
* Adds new hooks that subscribes directly to `clerk` via `useSyncExternalStore` 
  * These "base" hooks, like `useUserBase` are also reexported as `useUserContext` to avoid breaking change for these
  * They have the exact same signature, leaving all other hooks untouched
* `initialState` can now be a promise
  * `<ClerkProvider dynamic>` in the Next package no longer does `await initialState`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/clerk.ts
```diff
@@ -1476,11 +1477,11 @@ export class Clerk implements ClerkInterface {
     }
   };
 
-  public addListener = (listener: ListenerCallback): UnsubscribeCallback => {
+  public addListener = (listener: ListenerCallback, options?: ListenerOptions): UnsubscribeCallback => {
     listener = memoizeListenerCallback(listener);
     this.#listeners.push(listener);
     // emit right away
-    if (this.client) {
+    if (this.client && !options?.skipInitialEmit) {
```
Ephem, line 1565: `uSES` reads state directly from `clerk`, so this was introduced to avoid extra emits every time `addListener` was called in a hook.

packages/nextjs/src/app-router/server/ClerkProvider.tsx
```diff
@@ -32,28 +30,17 @@ export async function ClerkProvider(
...
-    }
-    return getNonceHeaders();
-  }
+  const statePromiseOrValue = dynamic ? getDynamicClerkState() : undefined;
+  const noncePromiseOrValue = dynamic ? getNonceHeaders() : '';
 
   const propsWithEnvs = mergeNextClerkPropsWithEnv({
     ...rest,
+    initialState: statePromiseOrValue as InitialState | Promise<InitialState> | undefined,
+    nonce: await noncePromiseOrValue,
```
Ephem, line 40: Awaiting this is something we might also want to move further down.

packages/react/src/contexts/AuthContext.tsx
Ephem, line 1: This should likely be renamed now that it doesn't actually contain a context.

packages/react/src/contexts/ClerkProvider.tsx
```diff
@@ -34,3 +39,28 @@ const ClerkProvider = withMaxAllowedInstancesGuard(ClerkProviderBase, 'ClerkProv
 ClerkProvider.displayName = 'ClerkProvider';
 
 export { ClerkProvider };
+
+const useLoadedIsomorphicClerk = (options: IsomorphicClerkOptions) => {
```
Ephem, line 35: This was moved up from the now deleted `ClerkContextProvider`, but code is unchanged.

packages/react/src/hooks/__tests__/useAuth.test.tsx
Ephem, line 1: Diff from the old PR. Ignore.

packages/react/src/hooks/__tests__/useAuth.type.test.ts
Ephem, line 1: Diff from the old PR, ignore. I haven't at all touched the tests for this PoC.

packages/shared/src/react/contexts.tsx
```diff
@@ -6,22 +6,47 @@ import React from 'react';
...
+
+  if (!initialState) {
+    return undefined;
+  }
+
+  if (initialState && 'then' in initialState) {
+    // TODO: If we want to preserve backwards compatibility, we'd need to throw here instead
+    // @ts-expect-error See above
+    return React.use(initialState);
+  }
```
Ephem, line 47: If `initialState` is passed in as a Promise, this is going to make sure any hook that relies on it suspends _at the location of the hook_. Before, you would always have to await at the provider level.

This should not be a breaking change.

packages/nextjs/src/app-router/server/ClerkProvider.tsx
```diff
@@ -32,28 +30,17 @@ export async function ClerkProvider(
...
-      return Promise.resolve('');
-    }
-    return getNonceHeaders();
-  }
+  const statePromiseOrValue = dynamic ? getDynamicClerkState() : undefined;
+  const noncePromiseOrValue = dynamic ? getNonceHeaders() : '';
 
   const propsWithEnvs = mergeNextClerkPropsWithEnv({
     ...rest,
+    initialState: statePromiseOrValue as InitialState | Promise<InitialState> | undefined,
```
Ephem, line 39: Not awaiting this at the top level might be considered a breaking change. I'm not sure it _is_ technically, but it might change the loading experience of apps, so we should be careful.

We could also do this incrementally and optionally by introducing something like `dynamic="stream"`.

packages/react/src/contexts/AuthContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+    useCallback(callback => clerk.addListener(callback, { skipInitialEmit: true }), [clerk]),
+    useCallback(() => snapshot, [snapshot]),
+    useCallback(() => initialSnapshot, [initialSnapshot]),
+  );
+
+  // If an updates comes in during a transition, uSES usually deopts that transition to be synchronous,
+  // which for example means that already mounted <Suspense> boundaries might suddenly show their fallback.
+  // This makes all auth state changes into transitions, but does not deopt to be synchronous. If it's
+  // called during a transition, it immediately uses the new value without deferring.
+  return useDeferredValue(authState);
```
Ephem: This is a common pattern for all the hooks and is a breaking change. Doing this by default in the `react` package might introduce transitions into apps that don't have them currently, which might cause problems, so we'll likely want to find a way to introduce this incrementally/optionally and let the framework level decide.

This does not by itself get rid of the transitive state. To do that safely we probably have to start the transition ourselves from inside `clerk-js` so we can make sure it wraps both the navigation and the state emit.

packages/shared/src/react/contexts.tsx
jacekradko: Does this mean we are requiring React@19? I believe `use()` is not available in 18

Ephem: The comment mentions this but is not super clear, but taking this all the way would mean we refactor this to throw the promise instead for backwards compatibility. That's a bit more work to implement though so went with `use` just for the poc. 😄

If we want to use `use` and avoid introducing a legacy React way to do this, maybe we could instead error if passing in `initialState` as a Promise on React < 19?

Edit: Since we own the Next package that will likely be the first and for a while the only user of this, we could either `await initialState` or pass down the promise directly depending on the Next version. Pre 16 we await it like today, >16 we pass down the promise.

PR #7305, opened 2025-11-25
feat: retry on 503, add Retry-After date (RFC1123) support

hhsnopek:
## Description

This pull-request introduces `with503Retry`, a fetch wrapper that retries on HTTP 503 only and honors the `Retry-After` header in both delta-seconds and HTTP-date (RFC1123) forms. Non-503 responses pass through untouched. Retries cap at 5 attempts, the per-attempt wait clamps at 60s, and waits abort immediately on an `AbortSignal`.

When `Retry-After` is present the wrapper utilizes it (clamped to 60s; past dates resolve to 0). When absent, the delay falls back to exponential backoff with full jitter (`Random(0, 2^attempt * 500ms)`) so concurrent clients don't thunder against a recovering origin.

`with503Retry` returns a new fetch-shaped function rather than mutating globals, so it composes per-client and is safe to use in tests.

NOTE: integration with clerk-js and backend SDK clients is staged behind this PR; follow-ups will plumb it through.

## Checklist

- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports.
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated.

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/backend/src/api/__tests__/factory.test.ts
+++ b/packages/backend/src/api/__tests__/factory.test.ts
@@ -169,6 +169,25 @@ describe('api.client', () => {
     expect(errResponse.retryAfter).toBe(123);
   });
 
+  it('executes a failed backend API request and includes Retry-After header RFC1123 date', async () => {
+    server.use(
+      http.get(
+        `https://api.clerk.test/v1/users/user_deadbeef`,
+        validateHeaders(() => {
+          return HttpResponse.json(
+            { errors: [] },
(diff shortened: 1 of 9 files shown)


Ephem:
This looks like a good change, but looks abandoned. Closing for now but feel free to re-open. 😄

PR #7325, opened 2025-11-27
feat(repo): Make gettoken callable outside of react

bratsos:
## Description

Preview of a hono app: We're loading clerk through a script tag in the HTML, and at the same time we're using `hono/jsx` to run a `getToken` call from the `@clerk/shared` package before `script.onload` calls `clerk.load()`. The whole promise dance works as expected and we clerk is ready, we can see the token value.

https://github.com/user-attachments/assets/1f29a0bd-e24d-4844-9a99-cedd763d64ec

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Added a browser-only getToken() API exposed from multiple SDKs; waits for Clerk initialization, supports ready/degraded states, handles timeouts, and returns null when no session.

* **Runtime**
  * SDKs now coordinate initialization readiness so standalone token calls reliably wait for Clerk lifecycle events.

* **Tests**
  * Added comprehensive tests for ready/loading/absent/error states, timeouts, concurrency, legacy behavior, and error propagation.

* **Chores**
  * Bumped minor versions across affected packages.


panteliselef:
@bratsos fyi people wanted to use this in expo (web/mobile) as well :)

bratsos:
> @bratsos fyi people wanted to use this in expo (web/mobile) as well :)

Hm, the issue here is that our solution depends on `window.Clerk` which is not available in react native. I think users in expo/react-native can use `getClerkInstance().session?.getToken()` instead.

packages/shared/src/getToken.ts
```diff
@@ -0,0 +1,160 @@
...
+    this.name = 'ClerkNotLoadedError';
+  }
+}
+
+class ClerkNotAvailableError extends Error {
+  constructor() {
+    super('Clerk: getToken can only be used in browser environments.');
+    this.name = 'ClerkNotAvailableError';
+  }
+}
```
brkalow: We have a shared error type for runtime errors we should leverage here: https://github.com/clerk/javascript/blob/main/packages/shared/src/errors/clerkRuntimeError.ts

Each error should have a unique code

packages/shared/src/getToken.ts
```diff
@@ -0,0 +1,160 @@
...
+ * });
+ * ```
+ */
+export async function getToken(options?: GetTokenOptions): Promise<string | null> {
+  try {
+    const clerk = await waitForClerk();
+
+    if (!clerk.session) {
+      return null;
+    }
```
brkalow: I think `null` is appropriate here, but wanted to flag just in case: Do you think we should throw an error in this case so the developer knows that a user is signed out? I'm debating both approaches and I'm not totally positive which one might be better. Leaning towards `null`...

packages/shared/src/getToken.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    return await clerk.session.getToken(options);
+  } catch (error) {
+    if (process.env.NODE_ENV === 'development') {
+      console.warn('[Clerk] getToken failed:', error);
+    }
+    return null;
```
brkalow: Is there a reason we want to swallow the error here? This makes me nervous as I expect it could lead to difficult to debug issues where users are sometimes treated as "signed out" when they actually aren't

packages/shared/src/getToken.ts
```diff
@@ -0,0 +1,160 @@
+import { inBrowser } from './browser';
+import type { ClerkStatus, GetTokenOptions, LoadedClerk } from './types';
+
+const POLL_INTERVAL_MS = 50;
+const MAX_POLL_RETRIES = 100; // 5 seconds of polling
+const TIMEOUT_MS = 10000; // 10 second absolute timeout
```
brkalow: Consider using the existing `retry()` helper: https://github.com/clerk/javascript/blob/main/packages/shared/src/retry.ts

packages/shared/src/getToken.ts
```diff
@@ -0,0 +1,160 @@
...
+
+type WindowClerk = LoadedClerk & {
+  status?: ClerkStatus;
+  loaded?: boolean;
+  on?: (event: 'status', handler: (status: ClerkStatus) => void, opts?: { notify?: boolean }) => void;
+  off?: (event: 'status', handler: (status: ClerkStatus) => void) => void;
+};
```
brkalow: [nit] Are all of these properties missing from `LoadedClerk`? If they are public we should consider updating the type if so

packages/shared/src/getToken.ts
Ephem, line 1: I was one of the people asking for this, as it would make things nicer in our own dashboard, but thinking about this again now, I think it might be problematic to introduce this API and I'm sorry I didn't realize it earlier. 😞

The reason is that I think this is _fundamentally_ incompatible with any concurrent features in frameworks, like transitions in React. If we let people access this directly, they kind of bypass any SDK integrations we design to work well with transitions.

If we'd for example leverage transitions to get rid of our transitive state in React, if users were using this we would not be able to control the token they see based on the transition. We want the current page to keep the old token, and the new page to use the new token, but this helper would always return whatever the current value is.

We are still some ways away from being able to fully support concurrent features, but I think this PR would introduce an API that could lead users away from the happy path and make that harder for us to do in the future. At least I think there might be enough uncertainty whether this is a good idea long term that we should hold it until we know for sure?

Another way to say this is, in the framework SDKs we usually lead users away from relying on the `clerk` singleton directly, as that bypasses the SDK layer we've designed for this framework, which can be unsafe. I think this is another version of that?

@brkalow

Ephem: I've been thinking about this some more and I'm not sure we can ever make auth state fork since cookies are involved. If we decide that auth state ultimately needs to be synchronous state, this PR still makes sense. I'm not sure we're in a place to decide either way just yet though, but wanted to update on my thinking. 😄

bratsos: Thanks for the thoughtful feedback @Ephem!

So if I understand this correctly, this is an existing issue as:
- users can access `window.Clerk.session.getToken` directly
- users can access `_session` cookie directly
- `createGetToken` in our react hooks is accessing `isomorphicClerk.session.getToken(options)` which is also not concurrency-safe

We could ship this helper for the intended use cases (axios interceptors, react-query etc) that run outside React's render phase, with clear documentation that framework-specific hooks should be preferred for use within components.

Re:concurrent features and transitions, even though it's not part of this implementation you raised a great point. Just to test my understanding, wouldn't the following be sufficient to handle this (as a future iteration, of course):
- use `useSyncExternalStore` for our react hook to safely get the current session
- introduce a way to maintain "versioned state generations" to handle transitions where different renders need different token values

packages/shared/src/getToken.ts
bratsos: I think the question we need to answer is what is the responsibility of this function, to report the auth state, or to provide a token? If the answer is to report the auth state (which is what I think the intention was by using `getX`), then returning `null` seems the better option.

Plus we don't want to penalize the happy path by forcing devs to always have to wrap this with a `try/catch`.

packages/shared/src/getToken.ts
bratsos: Thanks! Added here https://github.com/clerk/javascript/pull/7325/commits/d8ab553a66ce4205d5129dadc24465fec1aa3365

packages/shared/src/getToken.ts
bratsos: Nah you're right, I was just trying to create a shim `Clerk` type until we get `LoadedClerk` but I get that this might seem confusing. Fixed it here https://github.com/clerk/javascript/pull/7325/commits/d8ab553a66ce4205d5129dadc24465fec1aa3365

packages/shared/src/getToken.ts
bratsos: Fair point https://github.com/clerk/javascript/pull/7325/commits/d8ab553a66ce4205d5129dadc24465fec1aa3365

packages/shared/src/getToken.ts
bratsos: I was debating whether or not it would actually simplify things, as we're only using polling until clerk is loaded and then switch to listen to events, but after implementing it, it looks cleaner indeed. Thanks for the suggestion! https://github.com/clerk/javascript/pull/7325/commits/a3303fc7b0f4438e235cbe79eaccfdc0a471d0a7

packages/shared/src/getToken.ts
Ephem: So, this is tricky, but I'll try to provide some more details around my current thinking. Happy to hop on a call as well.

> users can access window.Clerk.session.getToken directly

Yes, but this is something we explicitly advice against. Same with `useClerk`. When you use those, you step outside of approved path and there be dragons.

> users can access _session cookie directly

Accessing this directly _in code_ is also advised against (I think? should be), but it's accessed implicitly when doing same-domain API calls, which is the main problem for supporting transitions, see later comment.

> `createGetToken` in our react hooks is accessing `isomorphicClerk.session.getToken(options)` which is also not concurrency-safe

Yes, today this is the case. Because it's behind our React hook, we can transparently change this implementation if we want though, which is my point. The hooks are a React-level facade that guarantees we can use React-level features in the future, even if we are not today.

> We could ship this helper for the intended use cases (axios interceptors, react-query etc) that run outside React's render phase, with clear documentation that framework-specific hooks should be preferred for use within components.

This wont help unfortunately. Since `react-query` etc can be triggered from within the render phase anyway (when using suspense), this `getToken` implementation would still be _tied to_ the render phase and the transition that triggered it. If people start using this for example with RQ, and we later want to update `getToken` to be transition-compatible via some React-level API, we couldn't do it since they are relying on this outside function.

---

All this said, the tricky thing here is the cookie. Even if we _could_ fork the state itself, the cookie is not forkable. The transition that uses the "new" state might do suspenseful data fetching _during the render phase_, so it would need the cookie to already be set to do so. If we set the cookie when triggering the transition, we run into the problem that any _synchronous interrupt_ that might happen that triggers new data fetching in the _current state fork of the world_ now has the wrong cookie value. This is all from the point of view of our current model of dealing with all this on the client.

If we consider the canonical way to do this in a RSC framework, that's likely by moving the auth state change to a server function. These are not quite _synchronous_, as the request itself is async and the app can still update while they are in progress (with existing state), but they happen _in serial_ and they _apply_ kinda synchronously in that auth state and cookie and UI would update together. More research is needed though and I'm not sure how this would affect this PR and new API. 🤔

bratsos: Thanks for the detailed response! I want to make sure we're aligned on what this is trying to solve.

I think we might be approaching the "users can access window.Clerk directly" point from different angles. My thinking is that these users already exist - people using vanilla JS, or non-concurrent frameworks - and they're already accessing `window.Clerk` directly or reading cookies because there's no other option for them. They'll likely continue to exist even with better RSC auth support. Right now the experience involves implementing their own timing/retry logic to wait for Clerk to load.
This PR is really about providing a shared SDK utility for those users, rather than replacing framework-specific hooks as the recommended approach.

On the React hooks being a facade - I think we're actually in agreement here! Having a framework-agnostic `getToken()` wouldn't mean our React SDK has to use it internally. The React hooks could continue to have their own implementation that evolves toward transition support, using `useSyncExternalStore` or whatever approach makes sense. The standalone function would serve non-React use cases, while `useAuth().getToken()` remains the recommended path for React users.

Regarding the concern about React Query users - I wonder if those users would have used our hooks anyway, or if they're the same folks already using `window.Clerk` today.

It also feels like requiring a component tree structure to access something that's globally available might be more friction than necessary for cases like configuring an Axios interceptor at app initialization.

I guess my main question is: is the concern primarily about encouraging React users away from hooks (which could potentially be addressed through documentation), or is there a deeper architectural issue with having a non-concurrent-aware `getToken` exist alongside the React implementation?

Happy to dig into either - I just want to make sure we're solving for the right thing.

bratsos:
> **🔴 Critical Issue 1: Race condition between status emission and promise creation**
> 
> **Location:** `packages/shared/src/getToken.ts:48-64` and `packages/clerk-js/src/core/clerk.ts:441-454`
> 
> There's a subtle race condition. Consider this timeline:
> 
> 1. `getToken()` is called
> 2. `getWindowClerk()` returns `undefined` because status is still `'loading'`
> 3. clerk-js status changes to `'ready'` and emits the event
> 4. `getToken()` creates the `__clerk_internal_ready` promise
> 5. The event was already emitted → promise never resolves → **timeout after 10s**
> 
> The window is small but real, especially on fast connections.
> 
> **Suggested fix:** In `clerk.ts`, after setting up the status listener, immediately check if already ready:
> 
> ```ts
> this.#publicEventBus.on(clerkEvents.Status, status => { ... });
> 
> // Immediate check in case status already changed before listener was added
> if ((this.#status === 'ready' || this.#status === 'degraded') && this.#isLoaded()) {
>   if (window.__clerk_internal_ready?.__resolve) {
>     window.__clerk_internal_ready.__resolve(this);
>   }
> }
> ```
> 
> This ensures that even if the status event fired before the promise was created, clerk-js will still resolve it.
> 
> **🔴 Critical Issue 2: Timeout timer not cleaned up**
> 
> **Location:** `packages/shared/src/getToken.ts:66-78`
> 
> ```ts
> return Promise.race([
>   readyPromise,
>   new Promise<never>((_, reject) =>
>     setTimeout(() => reject(...), TIMEOUT_MS),
>   ),
> ]);
> ```
> 
> When `readyPromise` resolves, the timeout timer continues running and eventually fires, rejecting an already-settled promise. While harmless functionally, it:
> 
> * Wastes resources (timer keeps running for 10s)
> * Could cause confusion when debugging
> * Multiple concurrent `getToken()` calls create multiple orphaned timers
> 
> **Suggested fix:**
> 
> ```ts
> let timeoutId: ReturnType<typeof setTimeout>;
> 
> const timeoutPromise = new Promise<never>((_, reject) => {
>   timeoutId = setTimeout(() => reject(
>     new ClerkRuntimeError('Timeout waiting for Clerk to load.', {
>       code: 'clerk_runtime_load_timeout',
>     }),
>   ), TIMEOUT_MS);
> });
> 
> return Promise.race([readyPromise, timeoutPromise]).finally(() => {
>   clearTimeout(timeoutId);
> });
> ```

Thanks! Valid points https://github.com/clerk/javascript/pull/7325/changes/f52cd3e412c5b40b1420bc539f26c0694849e44c

PR #7331, opened 2025-11-28
feat(clerk-js): Untrusted password screen for sign-in

octoper:
## Description

This PR introduces a new alternative method screen variant untrusted password, this screen will be used when the user tries to sign-in with a untrusted/compromised password to show the alternative methods the user can use to sign-in

https://github.com/user-attachments/assets/ebc79e10-f554-4bd4-b8aa-d71c94bfaf50

## Checklist

- [ ] `pnpm test` runs as expected.
- [ ] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Sign-in now detects "untrusted" passwords and shows a dedicated alternative-methods screen with tailored messaging and recovery options (email code, third‑party).

* **Localizations**
  * Added localization keys and error text placeholders for untrusted-password messaging across many locales.

* **Tests**
  * Added tests covering untrusted-password scenarios and UI transitions.

* **Chores**
  * Bumped packages to expose the new variant and messages.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #7356, opened 2025-12-03
chore(repo): Cherry pick compromised password screen changes

octoper:
## Description

Port #7331, #7345, #7352

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/localizations/src/en-US.ts
+++ b/packages/localizations/src/en-US.ts
@@ -903,6 +903,8 @@ export const enUS: LocalizationResource = {
     form_password_pwned__sign_in:
       'This password has been found as part of a breach and can not be used, please reset your password.',
     form_password_size_in_bytes_exceeded: undefined,
+    form_password_compromised__sign_in:
+      "Your password appears to have been compromised or it's no longer trusted and cannot be used. Please use another method to continue.",
     form_password_untrusted__sign_in:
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/SignIn/SignInFactorOnePasswordCard.tsx
```diff
@@ -64,20 +66,20 @@ export const SignInFactorOnePasswordCard = (props: SignInFactorOnePasswordProps)
...
       .attemptFirstFactor({ strategy: 'password', password: passwordControl.value })
       .then(res => {
         switch (res.status) {
           case 'complete':
             return setActive({
               session: res.createdSessionId,
-              navigate: async ({ session }) => {
-                await navigateOnSetActive({ session, redirectUrl: afterSignInUrl });
+              navigate: ({ session }) => {
+                void navigateOnSetActive({ session, redirectUrl: afterSignInUrl });
```
Ephem, line 82: @octoper Why was this changed? It seems to be breaking the redirect after the password flow for me (I'm trying the password flow in the Tanstack test template manually). After this gets called, I see a quick navigation to `/`, and then back to `/sign-in` again. I think because this happens because the navigation is no longer awaitable (which setActive does).

PR #7360, opened 2025-12-03
chore(nextjs)!: Throw when missing encryption key

brkalow:
## Description

When passing a secret key at runtime to `clerkMiddleware()`, we need a stable encryption key to safely pass the value from the middleware to the application server. Previously, we avoided throwing an error to prevent a breaking change. Now that we're releasing a major, we are changing this to throw an error.

fixes USER-3057

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/moody-peaches-stare.md
+++ b/.changeset/moody-peaches-stare.md
@@ -1,0 +1,5 @@
+---
+'@clerk/nextjs': major
+---
+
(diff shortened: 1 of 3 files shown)


brkalow:
!allow-major

nikosdouvlis (approved):

PR #7411, opened 2025-12-09
feat(*): Migrate to useSyncExternalStore

Ephem:
## Description

This is the full implementation of PoCs #7194 and #7267.

The main goal of the PR is to refactor the SDKs away from using a top level `addListener+state+context` combo for auth-state, to using `useSyncExternalStore`.

* Merge the `ContextProviders` from `ui` and `react` into `shared/react`
* Removes the `clerk.addListener()` in that provider, also removes most contexts
  * Breaking change: Several context providers like `UserContext` are no longer exported from `shared/react`
* Adds new hooks that subscribes directly to `clerk` via `useSyncExternalStore` 
  * These "base" hooks, like `useUserBase` are also reexported as `useUserContext` to avoid breaking change for these
  * They have the exact same signature, leaving all other hooks untouched
  * We should discuss this, both naming and if we want to do more breaking changes here instead

### More implementation notes

* Fixes bug in `useClearQueriesOnSignOut` as a conditional hook was causing the new tests to fail
* Removing `initialAuthState` from `useAuth` also fixes a bug where the _initial auth state_ was being used during the transitive state
	* This bug would return `null` instead of `undefined` on the page right after sign-in, which put the app in an akward state where it momentarily believed you were still signed out - Could lead to redirects back to `/sign-in`
* Renamed `this.#setAccessors();` to `this.#updateAccessors();` which now calls `this.#emit()` by default
  * Can pass `skipInitialEmit` to skip emitting
  * This was done because it's important to not let the `clerk` state and the app state tear
* `this.#emit()`
  * Now saves `this.__internal_lastEmittedResources`, so `useSES` can read the last resources emitted in `getSnapshot`
  * Supports a new `skipInitialEmit` to avoid emitting and re-rendering on `useSES` -> `subscribe` 
* Minor refactors for readability in adjacent code in `clerk-js`
* Some type improvements

With all this done, I noticed two timing issues in the bridge between the host React app and the Components app, which is a separate app. Both happened because components that should have unmounted had time to re-render and fire effects, because `useSES` emitted slightly earlier than `setState`.

* In `ui/src/Components.tsx`, the nodes (current components and which node they should be portaled into) was being set via `setState`
  * The `SignInRoutes` has a fallback, which is to redirect back to `/sign-in`
  * After sign in from password in `TanStack Router` specifically, `SignIn` stayed mounted and re-rendered because of `useSES`, even though the node was gone and the next page had already been rendered
  * This hit the fallback `<RedirectToSignIn />` route
  * This was fixed by:
    * Making the `nodes` a ref instead
    * Adding a `useSyncExternalStore` where we don't use the return value, to make sure `<Components />` re-renders before the child and unmounts it
* In `BaseRouter.tsx`, the async `baseNavigate` was doing `setRouteParts` as part of the navigation
  * This ensured the state was _set_ before returning to the caller
  * This did not ensure the component had _re-rendered_ as a response to the route change
  * On sign in, `useSES` in a child ran first and similar to the above case, triggered a redirect back to `/sign-in`
  * This was fixed by a `flushSync` to ensure the components had properly re-rendered, guaranteeing `setActive` does not leave the transitive state before the host app AND the Clerk components are ready
  
Both of these fixes might be considered controversial, happy for feedback!

Note that the unmount flow has multiple `.then()` which schedules unmounting into several microtasks even when `UI` is available. We might want to make this synchronous for even tighter control, but this was not necessary to fix these bugs.

Note also that these bugs result from having multiple React apps. The normal React lifecycle _should_ guarantee they don't happen in host apps.

### Background and motivation

Feel free to skip this if you already have familiarity, but as this is a fairly large change, I wanted to write down the full reasoning for this change for posterity.

As I see it, Clerk has different kinds of state. JWT state, piggybacking state and fetched state. Worth noting is that the fetched state has previously used SWR under the hood, but is about to change over to a custom React Query implementation, but both of these already uses `useSyncExternalStore` and are untouched by this PR.

The JWT state is fairly self-explanatory, and the client piggybacking state is asynchronous state that is included as part of other API-calls as a self-refreshing mechanism. The `clerk-js` base abstraction updates this for example when it polls for the `/tokens`, and _pushes_ the changes to the SDKs.

The JWT state and the piggybacking state together represents the foundational "auth state". This represents the user, the active organization etc.

#### Current solution

`clerk-js` has an `addListener` method. The React SDKs subscribes to this in an effect in their `<ClerkProvider>` and sets an internal state on every update. This state is then massaged a bit and placed on a few different contexts, and the different hooks, like `useAuth`, `useUser` etc read from these contexts.

`<ClerkProvider>` also supports passing in an `initialState`. If this is passed in, you can access it during SSR, but because there is no good way to prefetch the piggybacking or fetched state currently, only the `useAuth` state is really feasible to server render.

In Next, doing `<ClerkProvider dynamic>` automatically provides the `initialState` for `useAuth` behind the scenes. The reason we don't always do this is that it opts out of static generation and caching.

#### Current challenges and issues

Subscribing at the top has a few downsides however. It's currently hard to make things more streamy by allowing `initialState` to be represented as a promise, as we currently need to await it at the top. This is something we want to do to support `cacheComponents` and Partial prerendering better, allowing for finer-grained caching. This also opens up for making things suspenseful.

It also doesn't play well with concurrent rendering. A concrete example of this is an active bug that can happen when _one subtree_ suspends long enough during page load that `clerk-js` has time to load. When it unsuspends, it reads the _loaded_ auth state and not the initial one, causing hydration mismatches.

Another downside is how context is hard to use for fine grained updates, making it hard to implement things like selectors on top of it.

[`useSyncExternalStore`](https://react.dev/reference/react/useSyncExternalStore) is the canonical way to safely subscribe to an external store, and has none of these downsides.

Testing PR: https://github.com/clerk/dashboard/pull/7994

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Breaking Changes**
  * useAuth no longer accepts an initialAuthState parameter — provide initialState via ClerkProvider.
  * Removed public context exports: ClientContext, SessionContext, UserContext, OrganizationProvider.

* **New Features**
  * Listener API accepts skipInitialEmit option to opt out of initial events.
  * InitialStateProvider/initial-state hooks and a consolidated ClerkContextProvider added.
  * Transitions demo page added.

* **Bug Fixes**
  * Fixed hydration mismatches and brief user:null flash after sign-in.


packages/shared/src/react/contexts.tsx
```diff
@@ -1,28 +1,54 @@
...
-import { SWRConfigCompat } from './providers/SWRConfigCompat';
 
 const [ClerkInstanceContext, useClerkInstanceContext] = createContextAndHook<LoadedClerk>('ClerkInstanceContext');
-const [UserContext, useUserContext] = createContextAndHook<UserResource | null | undefined>('UserContext');
-const [ClientContext, useClientContext] = createContextAndHook<ClientResource | null | undefined>('ClientContext');
-const [SessionContext, useSessionContext] = createContextAndHook<SignedInSessionResource | null | undefined>(
-  'SessionContext',
+export { useUserBase as useUserContext } from './hooks/base/useUserBase';
+export { useClientBase as useClientContext } from './hooks/base/useClientBase';
+export { useSessionBase as useSessionContext } from './hooks/base/useSessionBase';
```
Ephem, line 20: The new base hooks return the same thing as the contexts hook did, so I reexported to keep changes minimal and this part non-breaking for a first pass.

Do we want to instead remove the useContext hooks? If so, we should also decide if we keep these new hooks internal or export them.

Also, what do we think about naming, `useUserBase` etc? Not sure what I think, so I just went with something.

packages/shared/src/react/contexts.tsx
```diff
@@ -119,17 +120,10 @@ Learn more: https://clerk.com/docs/components/clerk-provider`.trim(),
...
   OptionsContext,
-  OrganizationProvider,
-  SessionContext,
   useAssertWrappedByClerkProvider,
   useCheckoutContext,
   useClerkInstanceContext,
-  useClientContext,
   useOptionsContext,
   useOrganizationContext,
-  UserContext,
```
Ephem, line 120: The contexts being removed is a breaking change, but we don't think anyone is using them, and if they do it's kind of unavoidable.

packages/shared/src/react/contexts.tsx
```diff
@@ -1,28 +1,54 @@
...
-  'SessionContext',
+export { useUserBase as useUserContext } from './hooks/base/useUserBase';
+export { useClientBase as useClientContext } from './hooks/base/useClientBase';
+export { useSessionBase as useSessionContext } from './hooks/base/useSessionBase';
+
+const [InitialStateContext, _useInitialStateContext] = createContextAndHook<InitialState | undefined>(
+  'InitialStateContext',
 );
 
+export function InitialStateProvider({
```
Ephem, line 35: I just noticed I had forgotten to write a JSDoc comment for this. This is a new public API community SDK packages could use.

packages/react/src/hooks/useAuth.ts
```diff
@@ -35,7 +35,7 @@ type UseAuthOptions = Record<string, any> | PendingSessionOptions | undefined |
  * @unionReturnHeadings
  * ["Initialization", "Signed out", "Signed in (no active organization)", "Signed in (with active organization)"]
  *
- * @param [initialAuthStateOrOptions] - An object containing the initial authentication state or options for the `useAuth()` hook. If not provided, the hook will attempt to derive the state from the context. `treatPendingAsSignedOut` is a boolean that indicates whether pending sessions are considered as signed out or not. Defaults to `true`.
+ * @param [options] - An object containing options for the `useAuth()` hook. `treatPendingAsSignedOut` is a boolean that indicates whether pending sessions are considered as signed out or not. Defaults to `true`.
```
Ephem, line 38: Removing `initialAuthState` here is technically a breaking change, but we don't think anyone is, or should be using this as it was added for internal use.

packages/react/src/hooks/useAuth.ts
```diff
@@ -92,18 +92,11 @@ type UseAuthOptions = Record<string, any> | PendingSessionOptions | undefined |
...
 
-  const { treatPendingAsSignedOut, ...rest } = initialAuthStateOrOptions ?? {};
-  const initialAuthState = rest as any;
-
-  const authContextFromHook = useAuthContext();
-  let authContext = authContextFromHook;
-
-  if (authContext.sessionId === undefined && authContext.userId === undefined) {
-    authContext = initialAuthState != null ? initialAuthState : {};
-  }
```
Ephem, line 106: These lines have a bug that wrongly returns the _initial_ auth state during the "transitive state". This can cause you to appear signed out right after sign in for example.

Removing this also fixes the bug.

packages/react/src/hooks/useAuthBase.tsx
Ephem, line 1: Note for self: This is no longer a context, rename this file and hook to match the other hooks.

Update: ✅

packages/react/src/hooks/useAuthBase.tsx
```diff
@@ -0,0 +1,86 @@
...
+    useCallback(callback => clerk.addListener(callback, { skipInitialEmit: true }), [clerk]),
+    useCallback(() => {
+      if (!clerk.loaded || !clerk.__internal_lastEmittedResources) {
+        return getInitialState();
+      }
+
+      return clerk.__internal_lastEmittedResources;
+    }, [clerk, getInitialState]),
+    getInitialState,
+  );
```
Ephem, line 56: Using `useSyncExternalStore` in all hooks should hopefully fix the partial hydration mismatches we've seen at times.

packages/shared/src/react/contexts.tsx
```diff
@@ -1,28 +1,54 @@
...
+  const initialStateCtx = React.useMemo(() => ({ value: initialStateSnapshot }), [initialStateSnapshot]);
+  return <InitialStateContext.Provider value={initialStateCtx}>{children}</InitialStateContext.Provider>;
+}
+
+export function useInitialStateContext(): InitialState | undefined {
+  const initialState = _useInitialStateContext();
+
+  // If we want to support passing initialState as a promise, this is where we would handle that.
+  // Note that we can not use(promise) as long as we support React 18, so we'll need to have some extra handling
+  // and throw the promise instead.
```
Ephem, line 56: The idea is to implement this as a quick follow up. The reason we want this is to stop awaiting the `initialState` at the top level provider as one thing we should do to enable better PPR support.

I tried this with `use()` just to verify it worked, but an implementation with throwing is slightly trickier so wanted to do that as a separate PR.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2925,15 +2928,18 @@ export class Clerk implements ClerkInterface {
...
+  public __internal_lastEmittedResources: Resources | undefined;
   #emit = (): void => {
     if (this.client) {
+      const resources = {
+        client: this.client,
+        session: this.session,
+        user: this.user,
+        organization: this.organization,
+      };
+      this.__internal_lastEmittedResources = resources;
```
brkalow, line 3076: clever!

packages/react/src/hooks/useAuthBase.tsx
Ephem, line 1: Some of this is identical to the deleted `AuthContext.ts`, but git couldn't track it.

jacekradko:
!snapshot

clerk-cookie:
Hey @jacekradko - the snapshot version command generated the following package versions:
| Package | Version |
| --- | --- |
| @clerk/agent-toolkit | 0.2.9-snapshot.v20260113135251 |
| @clerk/astro | 3.0.0-snapshot.v20260113135251 |
| @clerk/backend | 3.0.0-snapshot.v20260113135251 |
| @clerk/chrome-extension | 3.0.0-snapshot.v20260113135251 |
| @clerk/clerk-js | 6.0.0-snapshot.v20260113135251 |
| @clerk/dev-cli | 1.0.0-snapshot.v20260113135251 |
| @clerk/expo | 3.0.0-snapshot.v20260113135251 |
| @clerk/expo-passkeys | 1.0.0-snapshot.v20260113135251 |
| @clerk/express | 2.0.0-snapshot.v20260113135251 |
| @clerk/fastify | 2.6.9-snapshot.v20260113135251 |
| @clerk/localizations | 4.0.0-snapshot.v20260113135251 |
| @clerk/msw | 0.0.1-snapshot.v20260113135251 |
| @clerk/nextjs | 7.0.0-snapshot.v20260113135251 |
| @clerk/nuxt | 2.0.0-snapshot.v20260113135251 |
| @clerk/react | 6.0.0-snapshot.v20260113135251 |
| @clerk/react-router | 3.0.0-snapshot.v20260113135251 |
| @clerk/shared | 4.0.0-snapshot.v20260113135251 |
| @clerk/tanstack-react-start | 1.0.0-snapshot.v20260113135251 |
| @clerk/testing | 2.0.0-snapshot.v20260113135251 |
| @clerk/ui | 1.0.0-snapshot.v20260113135251 |
| @clerk/upgrade | 2.0.0-snapshot.v20260113135251 |
| @clerk/vue | 2.0.0-snapshot.v20260113135251 |

`@clerk/agent-toolkit`
```sh
npm i @clerk/agent-toolkit@0.2.9-snapshot.v20260113135251 --save-exact
```
`@clerk/astro`
```sh
npm i @clerk/astro@3.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/backend`
```sh
npm i @clerk/backend@3.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/chrome-extension`
```sh
npm i @clerk/chrome-extension@3.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/clerk-js`
```sh
npm i @clerk/clerk-js@6.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/dev-cli`
```sh
npm i @clerk/dev-cli@1.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/expo`
```sh
npm i @clerk/expo@3.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/expo-passkeys`
```sh
npm i @clerk/expo-passkeys@1.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/express`
```sh
npm i @clerk/express@2.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/fastify`
```sh
npm i @clerk/fastify@2.6.9-snapshot.v20260113135251 --save-exact
```
`@clerk/localizations`
```sh
npm i @clerk/localizations@4.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/msw`
```sh
npm i @clerk/msw@0.0.1-snapshot.v20260113135251 --save-exact
```
`@clerk/nextjs`
```sh
npm i @clerk/nextjs@7.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/nuxt`
```sh
npm i @clerk/nuxt@2.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/react`
```sh
npm i @clerk/react@6.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/react-router`
```sh
npm i @clerk/react-router@3.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/shared`
```sh
npm i @clerk/shared@4.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/tanstack-react-start`
```sh
npm i @clerk/tanstack-react-start@1.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/testing`
```sh
npm i @clerk/testing@2.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/ui`
```sh
npm i @clerk/ui@1.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/upgrade`
```sh
npm i @clerk/upgrade@2.0.0-snapshot.v20260113135251 --save-exact
```
`@clerk/vue`
```sh
npm i @clerk/vue@2.0.0-snapshot.v20260113135251 --save-exact
```

Ephem:
!snapshot

clerk-cookie:
Hey @Ephem - the snapshot version command generated the following package versions:
| Package | Version |
| --- | --- |
| @clerk/agent-toolkit | 0.2.9-snapshot.v20260119140711 |
| @clerk/astro | 3.0.0-snapshot.v20260119140711 |
| @clerk/backend | 3.0.0-snapshot.v20260119140711 |
| @clerk/chrome-extension | 3.0.0-snapshot.v20260119140711 |
| @clerk/clerk-js | 6.0.0-snapshot.v20260119140711 |
| @clerk/dev-cli | 1.0.0-snapshot.v20260119140711 |
| @clerk/expo | 3.0.0-snapshot.v20260119140711 |
| @clerk/expo-passkeys | 1.0.0-snapshot.v20260119140711 |
| @clerk/express | 2.0.0-snapshot.v20260119140711 |
| @clerk/fastify | 2.6.9-snapshot.v20260119140711 |
| @clerk/localizations | 4.0.0-snapshot.v20260119140711 |
| @clerk/msw | 0.0.1-snapshot.v20260119140711 |
| @clerk/nextjs | 7.0.0-snapshot.v20260119140711 |
| @clerk/nuxt | 2.0.0-snapshot.v20260119140711 |
| @clerk/react | 6.0.0-snapshot.v20260119140711 |
| @clerk/react-router | 3.0.0-snapshot.v20260119140711 |
| @clerk/shared | 4.0.0-snapshot.v20260119140711 |
| @clerk/tanstack-react-start | 1.0.0-snapshot.v20260119140711 |
| @clerk/testing | 2.0.0-snapshot.v20260119140711 |
| @clerk/ui | 1.0.0-snapshot.v20260119140711 |
| @clerk/upgrade | 2.0.0-snapshot.v20260119140711 |
| @clerk/vue | 2.0.0-snapshot.v20260119140711 |

`@clerk/agent-toolkit`
```sh
npm i @clerk/agent-toolkit@0.2.9-snapshot.v20260119140711 --save-exact
```
`@clerk/astro`
```sh
npm i @clerk/astro@3.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/backend`
```sh
npm i @clerk/backend@3.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/chrome-extension`
```sh
npm i @clerk/chrome-extension@3.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/clerk-js`
```sh
npm i @clerk/clerk-js@6.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/dev-cli`
```sh
npm i @clerk/dev-cli@1.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/expo`
```sh
npm i @clerk/expo@3.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/expo-passkeys`
```sh
npm i @clerk/expo-passkeys@1.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/express`
```sh
npm i @clerk/express@2.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/fastify`
```sh
npm i @clerk/fastify@2.6.9-snapshot.v20260119140711 --save-exact
```
`@clerk/localizations`
```sh
npm i @clerk/localizations@4.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/msw`
```sh
npm i @clerk/msw@0.0.1-snapshot.v20260119140711 --save-exact
```
`@clerk/nextjs`
```sh
npm i @clerk/nextjs@7.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/nuxt`
```sh
npm i @clerk/nuxt@2.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/react`
```sh
npm i @clerk/react@6.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/react-router`
```sh
npm i @clerk/react-router@3.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/shared`
```sh
npm i @clerk/shared@4.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/tanstack-react-start`
```sh
npm i @clerk/tanstack-react-start@1.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/testing`
```sh
npm i @clerk/testing@2.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/ui`
```sh
npm i @clerk/ui@1.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/upgrade`
```sh
npm i @clerk/upgrade@2.0.0-snapshot.v20260119140711 --save-exact
```
`@clerk/vue`
```sh
npm i @clerk/vue@2.0.0-snapshot.v20260119140711 --save-exact
```

bratsos:
Solid work! I'll probably add more comments as I go through everything in detail, but the first thing I noticed is that we're mixing two different notification mechanisms for signaling changes in the Clerk singleton (notify+pull vs push).
Currently it works like this:

<details>
<summary>Current architecture</summary>

```

                           ┌─────────────────────┐
                           │   #emit()           │
                           │                     │
                           │   1. Stores state   │
                           │   2. Pushes data    │
                           │      to listeners   │
                           └──────────┬──────────┘
                                      │
                                      │ listener(resources)  ← data pushed
                                      ▼
                    ┌─────────────────────────────────────┐
                    │           #listeners                 │
                    │                                      │
                    │   (emission: Resources) => void      │
                    │                                      │
                    └─────────────────┬────────────────────┘
                                      │
                    ┌─────────────────┴─────────────────┐
                    │                                   │
                    ▼                                   ▼
         ┌──────────────────────┐           ┌──────────────────────┐
         │  Pull-based usage    │           │  Push-based usage    │
         │  (React hooks)       │           │  (Imperative code)   │
         │                      │           │                      │
         │  - Ignores pushed    │           │  - Uses pushed data  │
         │    data              │           │    directly          │
         │  - Reads from        │           │                      │
         │    __internal_*      │           │                      │
         │  - Needs             │           │                      │
         │    skipInitialEmit   │           │                      │
         └──────────────────────┘           └──────────────────────┘
```

</details>

The React hooks ignore the data passed to the callback and pull state separately, while imperative consumers use the pushed data directly. This works, but feels like we're "holding it wrong" - passing data that gets thrown away, and needing skipInitialEmit as a workaround.
With a minor refactor to how `#emit` works, we could make the pull-based mechanism the primitive and build push-based on top:

<details>
<summary>
New architecture
</summary>

```ts
class Clerk {
  #snapshot: Resources | undefined;
  #subscribers = new Set<() => void>();

  // ═══════════════════════════════════════════════════════════════════════
  // CORE: Pull-based (notify only, consumer reads state themselves)
  // ═══════════════════════════════════════════════════════════════════════

  /**
   * Subscribe to state changes. Callback receives no arguments.
   * Use getSnapshot() to read current state.
   */
  subscribe = (onStoreChange: () => void): (() => void) => {
    this.#subscribers.add(onStoreChange);
    return () => this.#subscribers.delete(onStoreChange);
  };

  /**
   * Get current state snapshot.
   */
  getSnapshot = (): Resources | undefined => this.#snapshot;

  // ═══════════════════════════════════════════════════════════════════════
  // CONVENIENCE: Push-based (built on top of pull-based)
  // ═══════════════════════════════════════════════════════════════════════

  /**
   * Subscribe with data pushed to callback.
   * Convenience wrapper for imperative consumers.
   */
  addListener = (
    listener: ListenerCallback,
    options?: ListenerOptions
  ): UnsubscribeCallback => {
    // Wrap: on notify → pull snapshot → push to listener
    const wrapper = () => {
      if (this.#snapshot) listener(this.#snapshot);
    };

    const unsubscribe = this.subscribe(wrapper);

    // Handle "emit immediately" behavior
    if (!options?.skipInitialEmit && this.#snapshot) {
      listener(this.#snapshot);
    }

    return unsubscribe;
  };

  // ═══════════════════════════════════════════════════════════════════════
  // INTERNAL
  // ═══════════════════════════════════════════════════════════════════════

  #notify = () => {
    if (this.client) {
      this.#snapshot = {
        client: this.client,
        session: this.session,
        user: this.user,
        organization: this.organization,
      };
      this.#subscribers.forEach(cb => cb()); // Just notify, no data
    }
  };
}
```

```
                           ┌─────────────────────┐
                           │      #notify()      │
                           │                     │
                           │   1. Updates        │
                           │      #snapshot      │
                           │   2. Calls all      │
                           │      subscribers    │
                           │      (no data)      │
                           └──────────┬──────────┘
                                      │
                                      │ callback()  ← just a ping
                                      ▼
                           ┌─────────────────────┐
                           │    #subscribers     │
                           │                     │
                           │   Set<() => void>   │
                           │                     │
                           └──────────┬──────────┘
                                      │
                    ┌─────────────────┴─────────────────┐
                    │                                   │
                    ▼                                   ▼
         ┌──────────────────────┐           ┌──────────────────────┐
         │   subscribe()        │           │   addListener()      │
         │   (Pull-based)       │           │   (Push-based)       │
         │                      │           │                      │
         │   - Adds bare cb     │           │   - Wraps listener   │
         │   - Consumer calls   │           │   - Wrapper pulls    │
         │     getSnapshot()    │           │     snapshot & passes│
         │                      │           │   - Initial emit opt │
         └──────────────────────┘           └──────────────────────┘
```

</details>

It's not _that important_ as there are no major differences in behavior, I just think it's cleaner in terms of intention and implementation.

As a small bonus, `subscribe`/`getSnapshot` would be stable method references so hooks wouldn't need `useCallback` wrappers. Also, `skipInitialEmit` would no longer be needed for pull-based consumers since subscribe naturally doesn't emit on registration.
WDYT?

Ephem:
> WDYT?

I'll think think more deeply on it tomorrow, but at a glance I really like it, feels cleaner/more intentional! 🙏

packages/ui/src/contexts/CoreClientContext.tsx
```diff
@@ -1,14 +1,22 @@
...
-  const ctx = useClientContext();
-  assertContextExists(ctx, ClientContext);
-  return ctx.signUp;
+  const client = useClientContext();
+  if (!client) {
+    throw new Error(
+      'Client must be defined before calling useCoreSignUp. This is a bug in Clerk, please report it to support@clerk.com',
+    );
+  }
+  return client.signUp;
```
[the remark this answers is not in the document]
Ephem: This preserves the original behavior, the caller is expected to not render until after Clerk has loaded as this is only used internally (hence "This is a bug in Clerk"). This might not be great but is a pattern in other parts of the codebase too and out of scope to refactor here.

packages/nextjs/src/app-router/client/ClerkProvider.tsx
```diff
@@ -107,6 +102,16 @@ export const ClientClerkProvider = <TUi extends Ui = Ui>(
   const { children, disableKeyless = false, ...rest } = props;
   const safePublishableKey = mergeNextClerkPropsWithEnv(rest).publishableKey;
 
+  // Avoid rendering nested ClerkProviders by checking for the existence of the ClerkNextOptions context provider
+  const isNested = Boolean(useClerkNextOptions());
+  if (isNested) {
+    if (rest.initialState) {
```
nikosdouvlis, line 108: Just double checking - is there any difference here between `initialState` being `null` vs `undefined`? The condition `if (rest.initialState)` would be false for both, but wondering if we need to handle them differently.

packages/shared/src/react/hooks/base/useSessionBase.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import type { SignedInSessionResource } from '@/types';
+
+import { useClerkInstanceContext, useInitialStateContext } from '../../contexts';
+
+export function useSessionBase(): SignedInSessionResource | null | undefined {
+  const clerk = useClerkInstanceContext();
+  const initialState = useInitialStateContext();
+  const getInitialState = useCallback(() => {
+    return initialState?.session as SignedInSessionResource | undefined;
```
nikosdouvlis: The cast `initialState?.session as SignedInSessionResource` is a bit unsafe since `InitialState.session` is typed as `SessionResource | undefined`. Probably fine in practice since session is always signed-in when present, but a type guard would be cleaner if we want to be strict about it.

packages/ui/src/Components.tsx
```diff
@@ -317,11 +315,27 @@ const Components = (props: ComponentsProps) => {
...
+  // re-render and even run effects when they should have already been unmounted.
+  // Forcing this to re-render first to remove the children is a workaround for this issue.
+  // Note that this does not fix the issue at its root, which is that it's possible for Clerk components
+  // to stay mounted even after their node has been removed.
+  useSyncExternalStore(
+    useCallback(callback => clerk.addListener(callback, { skipInitialEmit: true }), [clerk]),
+    useCallback(() => {
+      return clerk.__internal_lastEmittedResources;
+    }, [clerk]),
+  );
```
nikosdouvlis, line 341: The `useSyncExternalStore` here doesn't provide a `getServerSnapshot`, which means it will throw during SSR. I assume the Components app is never server-rendered, but should we add a comment explaining that this cannot be server rendered at the moment? Would help future us if someone tries to SSR this later.

packages/shared/src/react/hooks/useClearQueriesOnSignOut.ts
```diff
@@ -40,7 +40,7 @@ export function useClearQueriesOnSignOut(options: ClearQueriesOnSignOutOptions)
 
     const isNowSignedOut = isSignedOut === true;
 
-    if (previousIsSignedIn && isNowSignedOut) {
+    if (previousIsSignedIn && isNowSignedOut && authenticated === true) {
```
nikosdouvlis: I think the `&& authenticated === true` here is redundant since we already check `if (authenticated !== true) { return; }` at line 37. Did you mean to check something else here, or can we remove it?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -3092,21 +3098,37 @@ export class Clerk implements ClerkInterface {
...
 
-  #setAccessors = (session?: SignedInSessionResource | null) => {
-    this.session = session || null;
-    this.organization = this.#getLastActiveOrganizationFromSession();
-    this.user = this.session ? this.session.user : null;
+  /**
+   * Updates the accessors for the Clerk singleton and emits.
+   * If dangerouslySkipEmit is true, the emit will be skipped and you are responsible for calling #emit() yourself. This is dangerous because if there is a gap between setting these and emitting, library consumers that both read state directly and set up listeners could end up in a inconsistent state.
+   */
+  #updateAccessors = (session?: SignedInSessionResource | null, options?: { dangerouslySkipEmit?: boolean }) => {
```
nikosdouvlis, line 3120: The `dangerouslySkipEmit` flag feels a bit scary. The JSDoc warning is good, but I wonder if it makes sense to add something in our CLAUDE.md about this pattern so future contributors (and Claude) know when it's safe to use?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -3092,21 +3098,37 @@ export class Clerk implements ClerkInterface {
...
+    };
   };
 
-  #setAccessors = (session?: SignedInSessionResource | null) => {
-    this.session = session || null;
-    this.organization = this.#getLastActiveOrganizationFromSession();
-    this.user = this.session ? this.session.user : null;
+  /**
+   * Updates the accessors for the Clerk singleton and emits.
+   * If dangerouslySkipEmit is true, the emit will be skipped and you are responsible for calling #emit() yourself. This is dangerous because if there is a gap between setting these and emitting, library consumers that both read state directly and set up listeners could end up in a inconsistent state.
```
nikosdouvlis, line 3116: 🤔 We now have `skipInitialEmit` (on `addListener`) and `dangerouslySkipEmit` (on `#updateAccessors`). The difference is subtle: one controls whether a new subscriber gets an immediate callback, the other controls whether setting accessors triggers an emit. I wonder if there's a way to consolidate or simplify this flow in the future? Not blocking, but might be worth thinking about as a follow-up to reduce the number of "skip emit" concepts floating around.

packages/clerk-js/src/core/clerk.ts
Ephem: Good idea!

Note that the semantics for existing code is unchanged, so hopefully refactoring the emit into here and adding this "dangerous" option will make it clearer that gaps in timing between update and emit are dangerous. I agree we should do everything we can to get people to not use this for anything new. 😅

packages/clerk-js/src/core/clerk.ts
Ephem: Agreed! I plan to implement @bratsos suggestion in a follow up, which should let me remove `skipInitialEmit` entirely. If you want that behavior, you'll be able to `subscribe` instead of `addListener`.

packages/shared/src/react/hooks/useClearQueriesOnSignOut.ts
Ephem: Nice catch! Looks like @jacekradko added that other check last week and I missed that when resolving merge conflicts. Will remove.

packages/shared/src/react/hooks/base/useSessionBase.ts
Ephem: I'm pretty sure I just moved this cast from elsewhere, there were a bunch of those. I'll double check and will dig a bit deeper to see if I can clean it up a bit.

I've seen quite a few of these that doesn't feel great. 🤔

packages/nextjs/src/app-router/client/ClerkProvider.tsx
Ephem: Always good to double check! I don't think `initialState: null` is valid according to the types, and I do think the correct way to handle that case is to ignore it.

One might argue we'd want a way to _reset_ the `initialState` for a subtree if you do something like:

```tsx
<ClerkProvider dynamic>
  <ClerkProvider>
```

but I don't really see a case for that.

Ephem:
!allow-major

packages/clerk-js/src/core/clerk.ts
Ephem: This felt a bit specific/niche to litter CLAUDE.md etc with, but I expanded the comment a bit to advice against using the option for new code and listed an alternative approach.

packages/shared/src/react/hooks/useClearQueriesOnSignOut.ts
Ephem: ✅

packages/ui/src/Components.tsx
Ephem: I added this as the `getServerSnapshot`, together with a comment:

```tsx
useCallback(() => {
  return clerk.__internal_lastEmittedResources;
}, [clerk]),
```

This will prevent it from throwing when we want to play around with SSR, and should actually be fine here since we don't use the returned state anyway.

packages/shared/src/react/hooks/base/useSessionBase.ts
Ephem: This was from `deriveFromSsrInitialState`, which has a bunch of these and we should probably clean it up.

I refactored `useSessionBase` to call `deriveFromSsrInitialState` instead of using `.session` and casting directly. There's no real overhead to that and that way we at least keep the casts in one place.

PR #7440, opened 2025-12-11
test(e2e): Improve tanstack-router-start test template

Ephem:
## Description

This PR improves two things in the TanStack Router test template.

* Moves `<ClerkProvider>` down inside `<html>`
  * This fixes an issue where we were injecting a script before the framework scripts had run, resulting in trying to call `$_TSR.C()` when it was still undefined
* Makes the sign-in route a catch-all
  * Signing in when running the app manually was not working, but tests were passing, turns out the tests were working because they were using the hidden password field, so they were never updating the url

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Chores**
  * Reorganized authentication provider scope within the application's document structure
  * Updated sign-in route path configuration

### DIFF
--- a/integration/templates/tanstack-react-start/src/routes/__root.tsx
+++ b/integration/templates/tanstack-react-start/src/routes/__root.tsx
@@ -22,25 +22,25 @@ function RootComponent() {
 
 function RootDocument({ children }: { children: React.ReactNode }) {
   return (
-    <ClerkProvider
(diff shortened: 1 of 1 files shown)


wobsoriano (approved):
good catch!

PR #7486, opened 2025-12-16
feat(ui,shared,localizations): Show message in `TaskChooseOrganization` when org creation disabled

LauraBeatris:
## Description

Update `TaskChooseOrganization` screen to show a message when `user.createOrganizationEnabled` is disabled. 

This fixes an UX issue where personal accounts are disabled for an instance, and users are not allowed to create orgs, therefore getting stuck in the create org form.

CleanShot 2025-12-16 at 18 49 11

https://github.com/user-attachments/assets/4c30e616-f551-4150-bcbe-0a68f5240fe3

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Displays a dedicated screen when a user is not allowed to create organizations, with guidance and a sign-out option.
* **Localization**
  * Added localized messages for the organization-creation-disabled state across languages.
* **Tests**
  * Added test coverage verifying the restricted-creation UI and messaging.
* **Chores**
  * Recorded minor version bumps for related packages.

### DIFF
--- a/packages/localizations/src/ar-SA.ts
+++ b/packages/localizations/src/ar-SA.ts
@@ -867,6 +867,10 @@ export const arSA: LocalizationResource = {
       subtitle: 'أدخل تفاصيل منظمتك للمتابعة',
       title: 'إعداد منظمتك',
     },
+    organizationCreationDisabled: {
+      title: 'يجب أن تنتمي إلى منظمة',
+      subtitle: 'تواصل مع مسؤول منظمتك للحصول على دعوة.',
+    },
     signOut: {
       actionLink: 'تسجيل الخروج',
       actionText: 'تم تسجيل الدخول كـ {{identifier}}',
--- a/packages/localizations/src/be-BY.ts
+++ b/packages/localizations/src/be-BY.ts
@@ -875,6 +875,10 @@ export const beBY: LocalizationResource = {
       subtitle: 'Увядзіце даныя вашай арганізацыі для працягу',
       title: 'Наладзьце вашу арганізацыю',
     },
+    organizationCreationDisabled: {
+      title: 'Вы павінны належаць да арганізацыі',
+      subtitle: 'Звярніцеся да адміністратара вашай арганізацыі для атрымання запрашэння.',
+    },
     signOut: {
       actionLink: 'Выйсці',
       actionText: 'Увайшлі як {{identifier}}',
--- a/packages/localizations/src/bg-BG.ts
+++ b/packages/localizations/src/bg-BG.ts
@@ -871,6 +871,10 @@ export const bgBG: LocalizationResource = {
       subtitle: 'Въведете данните за вашата организация, за да продължите',
       title: 'Настройте вашата организация',
     },
+    organizationCreationDisabled: {
+      title: 'Трябва да принадлежите към организация',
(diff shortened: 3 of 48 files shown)


packages/localizations/src/en-US.ts
```diff
@@ -869,6 +869,10 @@ export const enUS: LocalizationResource = {
       subtitle: 'Enter your organization details to continue',
       title: 'Setup your organization',
     },
+    organizationCreationDisabled: {
```
tmilewski, line 872: 🙃  Would it be possible to use cursor to fill out the remaining locales?

LauraBeatris: I was waiting for an review on the actual localization key - let's say there's some feedback to update to a different name, then would be painful to revert the generation of remaining locales 😓 (unless cursor replaces all at once)

LauraBeatris: but will do that before merging this PR!

packages/shared/src/types/localization.ts
```diff
@@ -1303,6 +1303,10 @@ export type __internal_LocalizationResource = {
       action__createOrganization: LocalizationValue;
       action__invitationAccept: LocalizationValue;
     };
+    organizationCreationDisabled: {
+      title: LocalizationValue;
+      subtitle: LocalizationValue;
+    };
```
alexcarpenter, line 1309: will need to run generate to populate the rest of the locales with the new keys

LauraBeatris: https://github.com/clerk/javascript/pull/7486#discussion_r2628709851 related - sorry I should've added a PR comment about it

packages/ui/src/elements/contexts/index.tsx
```diff
@@ -128,8 +128,9 @@ export type FlowMetadata = {
     | 'complete'
     | 'accountSwitcher'
     | 'chooseOrganization'
+    | 'chooseWallet'
     | 'enterpriseConnections'
-    | 'chooseWallet';
+    | 'organizationCreationDisabled';
```
alexcarpenter, line 133: should we keep the keys similar for a group of flows e.g. `chooseOrganizationDisabled` vs `organizationCreationDisabled`?

LauraBeatris: I think `choose` could be confusing here, as this is based on the `creation` action - eg: if the user has existing memberships to choose, then they wouldn't see that flow step

packages/shared/src/types/localization.ts
LauraBeatris: Applied https://github.com/clerk/javascript/pull/7486/commits/01eedf6d65a64bf4c51ad462d1ccead63c09e128

packages/localizations/src/en-US.ts
LauraBeatris: Applied on https://github.com/clerk/javascript/pull/7486/commits/01eedf6d65a64bf4c51ad462d1ccead63c09e128

tmilewski (approved):

Issue #7456, opened 2025-12-14
<SignUp /> redirects the user to clerk when using <SignIn /> in a different part of the page.


ernestoalejo:
### Preliminary Checks

- [x] I have reviewed the documentation: https://clerk.com/docs

- [x] I have searched for existing issues: https://github.com/clerk/javascript/issues

- [x] I have not already reached out to Clerk support via email or Discord (if you have, no need to open an issue here)

- [x] This issue is not a question, general help request, or anything other than a bug report directly related to Clerk. Please ask questions in our Discord community: https://clerk.com/discord.

### Reproduction

https://github.com/ernestoalejo/clerk-signup-reproduction

### Publishable key

pk_test_ZmFuY3ktYmFybmFjbGUtOTIuY2xlcmsuYWNjb3VudHMuZGV2JA

### Description

I've found a pretty interesting bug to debug.

We noticed it when trying to implement inline `<SignIn />` / `<SignUp />` tabs in Vue inside our site. When attempting to sign in with an email code, the page unexpectedly redirects to the clerk domain.

We reproduced the issue in React as well, using only the `<SignUp />` component for simplicity, but it affects all frameworks and completely prevents implementing inline login or sign up workflows. Once the `SignUp` component has been loaded at least once, the sign in process becomes broken.

**Steps to reproduce:**

Easy method:

1. Use `<SignUp />` anywhere in the page to show an inline form.
2. Navigating to http://localhost:5173/#/factor-one redirects to the clerk domain.

Realistic method:

1. Use both `<SignIn />` and `<SignUp />` in the same page (for example as visual tabs inside a popup).
2. Try to sign in with an email code.
3. The form briefly flashes and sends the code, but then the page unexpectedly redirects to the clerk domain, losing the sign in process.

Debugging a redirect in compiled react code is quite cumbersome but I traced the issue to the `useEffect` in this component: https://github.com/clerk/javascript/blob/main/packages/ui/src/components/SignUp/index.tsx#L19

**Expected behavior:**

You should be able to implement an inline UI within a multi-step process without being redirected to the Clerk domain.

Also: signing in with an email code should work even if a sign-up form has been loaded elsewhere on the page.

**Actual behavior:**

The 6-digit form briefly appears and sends the login code, but then an invisible sign-up form triggers a redirect to the Clerk domain, causing loss of context for both the purchase flow and the login process.

I've used a test key in the reproduction, but this is directly afecting two Pro plan apps.

### Environment

```Shell
System:
[...]

PR #7529, opened 2025-12-26
fix(ui): add clerk dependency to redirect effect hooks

Ayush2k02:
## Summary

Fixes redirect conflicts when SignIn and SignUp components are used together on the same page.

## Problem

When both `<SignIn />` and `<SignUp />` components are rendered on the same page, they interfere with each other's navigation flows, causing unexpected redirects to the Clerk domain.

### Root Cause

The `RedirectToSignIn` and `RedirectToSignUp` components had incomplete dependency arrays in their useEffect hooks:

```tsx
React.useEffect(() => {
  void clerk.redirectToSignUp();
}, []); // ❌ Missing 'clerk' dependency
```

This violates React's exhaustive-deps rule and can cause unpredictable behavior.

## Solution

Added `clerk` to the dependency arrays:

```tsx
React.useEffect(() => {
  void clerk.redirectToSignUp();
}, [clerk]); // ✅ Proper dependency tracking
```

## Changes

- Updated `RedirectToSignIn` in [packages/ui/src/components/SignIn/index.tsx](packages/ui/src/components/SignIn/index.tsx)
- Updated `RedirectToSignUp` in [packages/ui/src/components/SignUp/index.tsx](packages/ui/src/components/SignUp/index.tsx)
- Added changeset for patch release

## Notes

Related to issue #7456

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed redirect conflicts when SignIn and SignUp components are rendered on the same page, preventing unintended redirects during authentication flows.

### DIFF
--- a/.changeset/fix-signin-signup-redirect-conflict.md
+++ b/.changeset/fix-signin-signup-redirect-conflict.md
@@ -1,0 +1,5 @@
+---
+"@clerk/ui": patch
+---
+
+Fix redirect conflicts when SignIn and SignUp components are used together on the same page. Added missing dependency arrays to useEffect hooks in redirect functions to prevent unwanted redirects during other component flows.
--- a/packages/ui/src/components/SignIn/index.tsx
+++ b/packages/ui/src/components/SignIn/index.tsx
@@ -41,7 +41,7 @@ function RedirectToSignIn() {
   const clerk = useClerk();
   React.useEffect(() => {
     void clerk.redirectToSignIn();
-  }, []);
+  }, [clerk]);
   return null;
 }
 
--- a/packages/ui/src/components/SignUp/index.tsx
+++ b/packages/ui/src/components/SignUp/index.tsx
@@ -22,7 +22,7 @@ function RedirectToSignUp() {
   const clerk = useClerk();
   React.useEffect(() => {
     void clerk.redirectToSignUp();
-  }, []);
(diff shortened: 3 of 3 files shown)


Ayush2k02:
## Testing Limitation

I've verified that the fix addresses the exact issue identified in #7456 (specifically the `useEffect` hook at [packages/ui/src/components/SignUp/index.tsx#L19](https://github.com/clerk/javascript/blob/main/packages/ui/src/components/SignUp/index.tsx#L19)).

However, I'm unable to fully test this fix locally because:

1. The published `@clerk/clerk-react` package loads UI components dynamically from Clerk's CDN at runtime
2. The fix is in the `@clerk/ui` package source code, which needs to be deployed to Clerk's CDN to be effective
3. Testing with the reproduction repository (https://github.com/ernestoalejo/clerk-signup-reproduction) still shows the redirect issue because it uses the published npm package that doesn't include this fix

### What I've Verified

✅ The fix is applied to the exact location identified by the issue reporter  
✅ Added `clerk` to the dependency arrays in both RedirectToSignIn and RedirectToSignUp components  
✅ This prevents the useEffect from running on every render, which was causing the redirect conflicts  
✅ Created proper changeset documentation

The fix will be testable once merged and deployed to Clerk's CDN.

Ayush2k02:
YAY

PR #7553, opened 2026-01-07
fix(ui): Hide "must belong to org" screen when org resources exist

LauraBeatris:
## Description

This PR fixes an issue where the "You must belong to an organization" was still being shown, even when the user had existing org memberships or suggestions to be chosen.

The previous implementation added a guard to check for the environment setting and display this screen, in order to avoid the loading status of the memberships/suggestions query, however this introduced the above bug.

#### 🐛 

https://github.com/user-attachments/assets/925895ee-19f5-40be-b5a2-92e318615800

#### With fix 

CleanShot 2026-01-07 at 18 21 52

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the "You must belong to an organization" screen appeared for users who already had memberships or suggestions.
  * Added a UI gate that displays an "organization creation disabled" screen when creation is not allowed.

* **Tests**
  * Added coverage for scenarios with existing memberships or suggestions to ensure correct screens are shown.

### DIFF
--- a/packages/ui/src/components/SessionTasks/tasks/TaskChooseOrganization/index.tsx
+++ b/packages/ui/src/components/SessionTasks/tasks/TaskChooseOrganization/index.tsx
@@ -21,6 +21,10 @@ const TaskChooseOrganizationInternal = () => {
   const hasExistingResources = !!(userMemberships?.count || userInvitations?.count || userSuggestions?.count);
   const isOrganizationCreationDisabled = !isLoading && !user?.createOrganizationEnabled && !hasExistingResources;
 
+  if (isOrganizationCreationDisabled) {
+    return <OrganizationCreationDisabledScreen />;
+  }
+
   return (
     <Flow.Root flow='taskChooseOrganization'>
       <Flow.Part part='chooseOrganization'>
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/SessionTasks/tasks/TaskChooseOrganization/index.tsx
```diff
@@ -14,24 +14,16 @@ import { ChooseOrganizationScreen } from './ChooseOrganizationScreen';
...
-  const { otherSessions } = useMultipleSessions({ user });
-  const { navigateAfterSignOut, navigateAfterMultiSessionSingleSignOutUrl } = useSignOutContext();
-
-  const handleSignOut = () => {
-    if (otherSessions.length === 0) {
-      return signOut(navigateAfterSignOut);
-    }
-
-    return signOut(navigateAfterMultiSessionSingleSignOutUrl, { sessionId: session?.id });
-  };
```
LauraBeatris, line 30: This was unused, somehow linter didn't caught it as well

PR #7567, opened 2026-01-08
fix(ui): Keep role select enabled with default org role

LauraBeatris:
#### Fix

Now only sets the field value on mount, if the default role matches an item on `/roles`

https://github.com/user-attachments/assets/65652908-705b-4ee9-8c59-011d6d4904d5

#### Previously

When `organizationSettings.defaultRole` does not exist on the options returned by `/roles`, the field gets initialized with `roleField.setValue`, but the option cannot be found, therefore it disables the select input:

image

image

image

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Role selector on the invite members page is no longer disabled when the organization's default role is missing; users can choose alternative roles and send invitations.

* **Tests**
  * Added test coverage verifying alternate roles can be selected and invitations sent when the default role isn't available.

* **Chores**
  * Added a release changeset entry; no public API changes.

### DIFF
--- a/packages/ui/src/components/OrganizationProfile/InviteMembersForm.tsx
+++ b/packages/ui/src/components/OrganizationProfile/InviteMembersForm.tsx
@@ -199,8 +199,8 @@ const AsyncRoleSelect = (field: ReturnType<typeof useFormControl<'role'>>) => {
       return;
     }
 
-    // Skip if the default role from org settings is not in the current role set.
-    // This will eventually be returned by the roles endpoint
+    // Skip if the default role from org settings is not in the current role set
+    // This will eventually be returned by the roles endpoint, and `organizationSettings.domains.defaultRole` will be deprecated
     const defaultRoleExists = options?.some(option => option.value === defaultRole);
     if (!defaultRoleExists) {
       return;


packages/ui/src/components/OrganizationProfile/InviteMembersForm.tsx
```diff
@@ -200,8 +191,21 @@ export const InviteMembersForm = (props: InviteMembersFormProps) => {
...
-
   const { t } = useLocalizations();
+  const defaultRole = useDefaultRole();
+
+  useEffect(() => {
+    if (field.value || !defaultRole) {
+      return;
+    }
+
+    const defaultRoleExists = options?.some(option => option.value === defaultRole);
```
LauraBeatris, line 204: Thinking of adding a comment claryfing this behavior with the current FAPI bug, just need to confirm with the team tomorrow first

.changeset/social-items-strive.md
```diff
@@ -0,0 +1,6 @@
+---
+'@clerk/clerk-js': patch
```
LauraBeatris:
```suggestion
```

NicolasLopes7 (approved):

packages/ui/src/components/OrganizationProfile/InviteMembersForm.tsx
```diff
@@ -200,8 +191,23 @@ export const InviteMembersForm = (props: InviteMembersFormProps) => {
...
   const { t } = useLocalizations();
+  const defaultRole = useDefaultRole();
+
+  useEffect(() => {
+    if (field.value || !defaultRole) {
+      return;
+    }
+
+    // Skip if the default role from org settings is not in the current role set.
+    // This will eventually be returned by the roles endpoint
```
iagodahlem: I would just mention that we want to deprecate the default role from the organization settings as well:
```suggestion
    // This will eventually be returned by the roles endpoint, and `organizationSettings.domains.defaultRole` will be deprecated.
```

PR #7569, opened 2026-01-08
fix(clerk-js): Keep role select enabled with default org role

LauraBeatris:
## Description

Backport from https://github.com/clerk/javascript/pull/7567

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/ui/components/OrganizationProfile/__tests__/InviteMembersPage.test.tsx
+++ b/packages/clerk-js/src/ui/components/OrganizationProfile/__tests__/InviteMembersPage.test.tsx
@@ -285,7 +285,7 @@ describe('InviteMembersPage', () => {
       });
 
       fixtures.clerk.organization?.inviteMembers.mockResolvedValueOnce([{}] as OrganizationInvitationResource[]);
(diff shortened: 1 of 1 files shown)


packages/clerk-js/src/ui/components/OrganizationProfile/InviteMembersForm.tsx
```diff
@@ -200,8 +191,21 @@ export const InviteMembersForm = (props: InviteMembersFormProps) => {
...
+
+  useEffect(() => {
+    if (field.value || !defaultRole) {
+      return;
+    }
+
+    const defaultRoleExists = options?.some(option => option.value === defaultRole);
+    if (!defaultRoleExists) {
+      return;
+    }
```
LauraBeatris, line 207:
```suggestion
    // Skip if the default role from org settings is not in the current role set.
    // This will eventually be returned by the roles endpoint
    const defaultRoleExists = options?.some(option => option.value === defaultRole);
    if (!defaultRoleExists) {
      return;
    }
```

PR #7586, opened 2026-01-13
feat(ui): replace sync update with deferred promise pattern

jacekradko:
## Summary

Replaces the `flushSync` call in `BaseRouter`'s `baseNavigate` with a deferred promise pattern.

## Problem

The current implementation uses `flushSync` to guarantee the re-render happens before returning control to the caller. While this works, `flushSync` has performance implications:
- Forces React to synchronously flush the entire update queue
- Bypasses React's batching and scheduling optimizations
- Can cause layout thrashing in complex UIs

## Solution

Use a deferred promise pattern instead:

1. Store the resolve function in state alongside the new route parts
2. An effect resolves the promise after React commits the state update
3. This achieves the same guarantee without forcing synchronous updates

```typescript
// Before (flushSync)
flushSync(() => {
  setRouteParts({ path: toURL.pathname, queryString: toURL.search });
});
return internalNavRes;

// After (deferred promise)
return new Promise(resolve => {
  setRouteParts({ path: toURL.pathname, queryString: toURL.search });
  setPendingNavigation({
    routeParts: { path: toURL.pathname, queryString: toURL.search },
    result: internalNavRes,
    resolve,
  });
});
```

## Benefits

- Avoids `flushSync` performance penalty
- Works better with React concurrent features
- Maintains the same behavioral guarantee (re-render completes before promise resolves)

### DIFF
--- a/packages/ui/src/router/__tests__/BaseRouter.test.tsx
+++ b/packages/ui/src/router/__tests__/BaseRouter.test.tsx
@@ -38,7 +38,7 @@ describe('BaseRouter', () => {
   describe('concurrent navigations', () => {
     beforeEach(() => {
       // @ts-ignore
-      window.location = new URL('https://www.example.com/virtual');
+      window.location = new URL('https://www.example.test/virtual');
     });
 
     it('resolves all promises when multiple navigations occur before effect runs', async () => {
--- a/packages/ui/src/router/__tests__/HashRouter.test.tsx
+++ b/packages/ui/src/router/__tests__/HashRouter.test.tsx
@@ -69,7 +69,7 @@ describe('HashRouter', () => {
   describe('when hash has a path included in it', () => {
     beforeEach(() => {
       // @ts-ignore
-      window.location = new URL('https://www.example.com/hash#/foo');
+      window.location = new URL('https://www.example.test/hash#/foo');
     });
 
     it('loads that path', () => {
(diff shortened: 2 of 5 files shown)


packages/ui/src/router/BaseRouter.tsx
```diff
@@ -119,14 +141,19 @@ export const BaseRouter = ({
...
+    // happens before handing things back to the caller. This avoids the flushSync
+    // performance penalty while still ensuring children re-render with the new
+    // navigation state before setActive might emit.
+    return new Promise(resolve => {
       setRouteParts({ path: toURL.pathname, queryString: toURL.search });
+      setPendingNavigation({
+        routeParts: { path: toURL.pathname, queryString: toURL.search },
+        result: internalNavRes,
+        resolve,
+      });
```
Ephem: What happens if several `baseNavigate` get's called before `useEffect` has a chance to run? I think only the last one would resolve and the others would hang?

Naively, I would append all pending to an array instead and resolve them all in the effect, but I'm not sure that's correct either. 🤔

jacekradko: @Ephem Updated in [`1de7582` (this PR)](https://github.com/clerk/javascript/pull/7586/commits/1de758247a2a22eae233cfdf6ff075483bc32964)

PR #7595, opened 2026-01-14
feat(nextjs): Add support for Next.js 16 cache components

jacekradko:
## Summary

Adds support for Next.js 16 cache components by improving error detection and providing helpful error messages when `auth()` or `currentUser()` are called inside a `"use cache"` function.

### Changes

- Add `ClerkUseCacheError` class with symbol marker to prevent double-wrapping of error messages
- Add `isNextjsUseCacheError()` helper to detect "use cache" context errors with tightened regex patterns to reduce false positives
- Add `isClerkUseCacheError()` helper to identify already-formatted Clerk errors
- Update `buildRequestLike()` to throw `ClerkUseCacheError` with helpful message
- Update `auth()` and `currentUser()` to check for already-formatted errors before wrapping
- Update `clerkClient()` to re-throw `ClerkUseCacheError`
- Add unit tests for error detection and handling

## Problem

When a developer calls `auth()` or `currentUser()` inside a `"use cache"` function, they receive a cryptic error about `headers()` not being allowed in cache contexts.

## Solution

Detect these errors and provide a helpful message with a code example showing the correct pattern:

```
Clerk: auth() and currentUser() cannot be called inside a "use cache" function.
These functions access `headers()` internally, which is a dynamic API not allowed in cached contexts.

To fix this, call auth() outside the cached function and pass the values you need as arguments:

  async function getCachedUser(userId: string) {
    "use cache";
    const client = await clerkClient();
    return client.users.getUser(userId);
  }

  // In your component/page:
  const { userId } = await auth();
  if (userId) {
    const user = await getCachedUser(userId);
  }
```

### Implementation details

- Uses a custom `ClerkUseCacheError` class with a symbol marker to prevent double-wrapping when errors bubble through multiple catch handlers (`buildRequestLike` → `auth` → `currentUser`)
- Tightened regex patterns in `isNextjsUseCacheError()` to reduce false positives - now requires specific Next.js error patterns like `"use cache"` combined with `headers`/`cookies` context

## Test plan

- [x] Unit tests for `ClerkUseCacheError`, `isClerkUseCacheError()`, `isNextjsUseCacheError()`, and `isPrerenderingBailout()`
- [x] Integration tests for cache components
- [x] Build succeeds
- [x] All tests pass

### Setup
1. Create or use a Next.js 16 app with cache components enabled:
   ```ts
   // next.config.ts
   import type { NextConfig } from 'next'

   const nextConfig: NextConfig = {
     cacheComponents: true, // No longer under experimental in Next.js 16
   }

   export default nextConfig
   ```

2. Install the PR preview package:
   ```bash
   npm i https://pkg.pr.new/@clerk/nextjs@7595
   ```

### Test the error detection

Create a page that calls `auth()` inside a `"use cache"` function:

```tsx
// app/test-cache/page.tsx
import { auth } from '@clerk/nextjs/server';

async function getCachedData() {
  "use cache";
  const { userId } = await auth(); // This should trigger our helpful error
  return { userId };
}

export default async function TestPage() {
  const data = await getCachedData();
  return <div>{JSON.stringify(data)}</div>;
}
```

### Expected behavior

**Before this PR:** Cryptic Next.js error about `headers()` not being allowed in cache contexts

**After this PR:** Clear error message explaining the issue and showing the correct pattern

## Summary by CodeRabbit

* **New Features**
  * Added support for Next.js 16 cache components.

* **Bug Fixes**
  * Improved detection and clearer, actionable error messages when auth() or currentUser() are used inside a "use cache" context.
  * Standardized handling to surfacing original error details for easier debugging.

* **Documentation**
  * Added interactive example pages demonstrating correct cache usage patterns and an error-triggering demo.

* **Tests**
  * Expanded end-to-end and unit tests covering cache-component scenarios and error messaging.


packages/nextjs/src/app-router/server/utils.ts
```diff
@@ -11,14 +15,59 @@ export const isPrerenderingBailout = (e: unknown) => {
...
   const bailOutPrerendering = lowerCaseInput.includes('this page needs to bail out of prerendering');
 
+  // Next.js 16+ with cacheComponents: headers() rejects during prerendering
+  // Error: "During prerendering, `headers()` rejects when the prerender is complete"
+  const headersRejectsDuringPrerendering = lowerCaseInput.includes('during prerendering');
+
   // note: new error message syntax introduced in next@14.1.1-canary.21
   // but we still want to support older versions.
   // https://github.com/vercel/next.js/pull/61332 (dynamic-rendering.ts:153)
   const routeRegex = /Route .*? needs to bail out of prerendering at this point because it used .*?./;
```
brkalow: nit: consider hoisting this pattern out of the function like the others in this file

jacekradko: Addressed

packages/nextjs/src/app-router/server/utils.ts
```diff
@@ -11,14 +19,56 @@ export const isPrerenderingBailout = (e: unknown) => {
...
+};
+
+/**
+ * Error message for when auth()/currentUser() is called inside a "use cache" function.
+ * Exported so it can be reused in auth.ts and currentUser.ts for consistent messaging.
+ */
+export const USE_CACHE_ERROR_MESSAGE =
+  `Clerk: auth() and currentUser() cannot be called inside a "use cache" function. ` +
+  `These functions access \`headers()\` internally, which is a dynamic API not allowed in cached contexts.\n\n` +
+  `To fix this, call auth() outside the cached function and pass the userId as an argument:\n\n` +
```
Ephem: NIT: `userId` is by far the most likely here, but not the only possible thing one might want to pass in.

Perhaps something like this?

```suggestion
  `To fix this, call auth() outside the cached function and pass the values you need as arguments:\n\n` +
```

packages/nextjs/src/app-router/server/utils.ts
```diff
@@ -1,38 +1,105 @@
...
+    this.name = 'ClerkUseCacheError';
+  }
+}
+
+export const isClerkUseCacheError = (e: unknown): e is ClerkUseCacheError => {
+  return e instanceof Error && CLERK_USE_CACHE_MARKER in e;
+};
+
+// Patterns for Next.js "use cache" errors - tightened to reduce false positives
+const USE_CACHE_WITH_DYNAMIC_API_PATTERN =
```
bratsos, line 26: ❓ I _think_ this regex will miss `'use cache'` - shouldn't we catch this as well?

jacekradko: Nice catch.. updated to handle both single and double quotes

PR #7317, opened 2025-11-26
feat(clerk-js): Proactive session token refresh

jacekradko:
## Description

This PR implements **proactive token refresh** for session tokens. Instead of waiting for `getToken()` calls to trigger refresh when a token is close to expiration, we schedule a timer when tokens are cached that fires proactively before the leeway period begins.

Fixes: USER-4087

Dashboard testing PR: https://github.com/clerk/dashboard/pull/7990

Screenshot 2026-01-15 at 9 38 16 AM

## How It Works

```
Token TTL Timeline (60s token, 15s leeway)
─────────────────────────────────────────────────────────────────────►
T=0                    T=43s              T=55s                    T=60s
(issued)           (timer fires)       (leeway)                 (expires)

│←─────── Fresh zone ───────→│←── Refresh ──→│←─ Danger ─→│
         (return cached)         (proactive      (force sync
                                  refresh)        refresh)

         No action needed        Timer triggers   Token deleted
                                 background       from cache
                                 refresh
```

### Key Behaviors

1. **Fresh zone** (TTL > leeway + 2s): Return cached token, no action needed
2. **Refresh timer fires** (at TTL - leeway - 2s = 43s for 60s tokens): Proactively trigger background refresh
3. **Danger zone** (TTL < 5s): Token deleted from cache, forces synchronous fetch

### Benefits

- **More consistent refresh**: Tokens are refreshed proactively regardless of `getToken()` call frequency
- **Zero risk**: If the timer doesn't fire for any reason, the poller catches it as a fallback
- **Cross-tab sync**: New tokens are broadcast to other tabs via `BroadcastChannel`

### Token Cache (`tokenCache.ts`)

- `onRefresh` callback on cache entries for proactive refresh
- Timer scheduled at `(TTL - leeway - 2s)` when tokens are cached
- Tokens < 5s from expiration are deleted from cache (forces sync fetch)
- Cross-tab synchronization via `BroadcastChannel`

### Session (`Session.ts`)

- `#refreshTokenInBackground()`: Fires token refresh without blocking
- `#backgroundRefreshInProgress` Set prevents duplicate concurrent refreshes
- `onRefresh` callback passed when caching tokens
- `resolvedToken` on cache entries enables synchronous reads (avoids microtask overhead)

### Constants

- **Default leeway**: 15 seconds
- **Minimum leeway**: 5 seconds (poller interval)
- **Refresh lead time**: 2 seconds before leeway starts

## Breaking Changes

[...]

PR #7598, opened 2026-01-14
feat(clerk-js): Improve `getToken` offline handling

bratsos:
## Description

https://github.com/user-attachments/assets/602c3bbd-44be-4cfb-8db7-4871bfeac651

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Breaking Changes**
  * Major-version upgrades to Clerk packages; expect breaking changes.
  * getToken() now throws ClerkOfflineError when offline instead of returning null — wrap calls in try/catch.

* **New Features**
  * New ClerkOfflineError class and type-guard (ClerkOfflineError.is) for detecting offline network failures.

* **Documentation**
  * Migration guide and examples for handling getToken() offline behavior and retry/offline UX.

### DIFF
--- a/.changeset/fast-paws-promise.md
+++ b/.changeset/fast-paws-promise.md
@@ -1,0 +1,5 @@
+---
+"@clerk/tanstack-react-start": minor
+---
+
+Bumped minimum required TanStack Start dependencies to 1.157.0+ and removed usage of deprecated `json()` in favor of standard Web APIs.
--- a/.changeset/great-bats-attack.md
+++ b/.changeset/great-bats-attack.md
@@ -1,0 +1,12 @@
+---
+'@clerk/tanstack-react-start': minor
+'@clerk/localizations': minor
+'@clerk/react-router': minor
+'@clerk/clerk-js': minor
+'@clerk/nextjs': minor
+'@clerk/shared': minor
+'@clerk/react': minor
+'@clerk/ui': minor
+---
+
+Introducing `setup_mfa` session task
--- a/integration/presets/envs.ts
+++ b/integration/presets/envs.ts
@@ -158,6 +158,13 @@ const withSessionTasksResetPassword = base
   .setEnvVariable('private', 'CLERK_SECRET_KEY', instanceKeys.get('with-session-tasks-reset-password').sk)
   .setEnvVariable('public', 'CLERK_PUBLISHABLE_KEY', instanceKeys.get('with-session-tasks-reset-password').pk);
 
+const withSessionTasksSetupMfa = base
+  .clone()
+  .setId('withSessionTasksSetupMfa')
+  .setEnvVariable('private', 'CLERK_SECRET_KEY', instanceKeys.get('with-session-tasks-setup-mfa').sk)
+  .setEnvVariable('public', 'CLERK_PUBLISHABLE_KEY', instanceKeys.get('with-session-tasks-setup-mfa').pk)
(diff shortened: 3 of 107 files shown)


packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -93,22 +91,23 @@ export abstract class BaseResource {
...
     } catch (e) {
+      if (ClerkOfflineError.is(e)) {
+        throw e;
+      }
+
+      if (isNetworkError(e) || !isValidBrowserOnline()) {
+        throw new ClerkOfflineError('Network request failed', {
+          cause: e instanceof Error ? e : undefined,
+        });
+      }
```
jacekradko: Is this something we want to happen for all resources, or just the session resource?

jacekradko:
@bratsos I am concerned that this PR is going to conflict heavily with the stale-while-revalidate change: https://github.com/clerk/javascript/pull/7317 

I think we will want to base this PR off feat/stale-while-revalidate-token and build on top of it. Some of the offline heuristics work well with SWR since we can return a cached token whenever possible and error only when we truly cannot fetch a new token

.changeset/sparkly-aliens-see.md
```diff
@@ -0,0 +1,30 @@
+---
+'@clerk/tanstack-react-start': major
+'@clerk/react-router': major
+'@clerk/clerk-js': major
+'@clerk/upgrade': major
+'@clerk/nextjs': major
+'@clerk/shared': major
+'@clerk/react': major
+'@clerk/nuxt': major
+'@clerk/vue': major
```
bratsos, line 10: I'm not sure if this should be a major change in all sdks 🤔

packages/clerk-js/src/core/resources/Base.ts
bratsos: I've change this since the last refactor, I've only kept the `ClerkOfflineError` exception only on session resource

packages/clerk-js/src/core/resources/__tests__/Session.test.ts
```diff
@@ -260,24 +260,24 @@ describe('Session', () => {
...
         Object.defineProperty(window.navigator, 'onLine', {
           writable: true,
           value: true,
         });
-        warnSpy.mockRestore();
+        vi.useFakeTimers();
       });
 
-      it('returns null', async () => {
+      it('throws ClerkOfflineError when offline', { timeout: 20000 }, async () => {
```
jacekradko: Is this test intended to run long? Not sure how feasible it would be, but perhaps an E2E is a better fit

.changeset/sparkly-aliens-see.md
jacekradko: I don't think so, but it doesn't really matter in practice as all of them are going to get a major anyways

bratsos: Should I leave it like this then?

packages/clerk-js/src/core/resources/__tests__/Session.test.ts
bratsos: Good catch, I don't think its needed anymore. Removed here https://github.com/clerk/javascript/pull/7598/changes/a36112f149e6a21a49045c78d6ae7180ddc9b1ad

bratsos:
!allow-major

PR #7600, opened 2026-01-14
feat(clerk-js): Include current search params in `handleRedirectCallback`

jfoshee:
so that component `AuthenticateWithRedirectCallback` automatically incorporates any redirects in the current query to match the [documented usage](https://clerk.com/docs/guides/development/custom-flows/authentication/oauth-connections)

Part of [USER-4308](https://linear.app/clerk/issue/USER-4308/oauth-integration-flow-fails-for-new-user-w-customized-sign-in-and-sso)

Without a fix like this the developer must manually snag and specify all the possible redirects, which I think is not the intent of the component.

    const redirectUrl = searchParams.get('redirect_url') ?? undefined
    const signInFallbackRedirectUrl = searchParams.get('sign_in_fallback_redirect_url') ?? undefined
    const signInForceRedirectUrl = searchParams.get('sign_in_force_redirect_url') ?? undefined
    const signUpFallbackRedirectUrl = searchParams.get('sign_up_fallback_redirect_url') ?? undefined
    const signUpForceRedirectUrl = searchParams.get('sign_up_force_redirect_url') ?? undefined

```
       <AuthenticateWithRedirectCallback
              continueSignUpUrl={buildContinueSignUpUrl()}
              redirectUrl={redirectUrl}
              signInFallbackRedirectUrl={signInFallbackRedirectUrl}
              signInForceRedirectUrl={signInForceRedirectUrl}
              signUpFallbackRedirectUrl={signUpFallbackRedirectUrl}
              signUpForceRedirectUrl={signUpForceRedirectUrl}
            />
```

### Note to Reviewers

❓ Is this the best place to insert the URL's query params? It could be done at a higher level, e.g. adding an argument to `handleRedirectCallback`. Are there scenarios where it is undesirable to automatically use the window's current query?

❓ What are some scenarios I should manually test?

It would be helpful to consider this change in conjunction with [this docs change](https://github.com/clerk/clerk-docs/pull/2934). I acknowledge that the change to the docs is not ideal; it makes the code example much longer. But, my thinking is to first: make it correct, second: make SDK improvements that could reduce the burden on customers.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/clerk.ts
```diff
@@ -2084,7 +2084,7 @@ export class Clerk implements ClerkInterface {
         buildURL({ base: displayConfig.signInUrl, hashPath: '/reset-password' }, { stringify: true }),
     );
 
-    const redirectUrls = new RedirectUrls(this.#options, params);
+    const redirectUrls = new RedirectUrls(this.#options, params, new URLSearchParams(window.location.search));
```
nikosdouvlis: Could this fix introduce a breaking change? This change is in `_handleRedirectCallback` which is shared by both `handleRedirectCallback` and `handleGoogleOneTapCallback`. For the OAuth redirect callback this makes total sense, that's the whole point of the fix. But for Google One Tap, the dialog fires on whatever page the user is currently browsing, not on a dedicated callback URL. Since search params have highest priority in `#getRedirectUrl` (search params > props > options), any `redirect_url` that happens to be in the page's query string would silently override the redirect params passed via props.

Should we scope this to just the `handleRedirectCallback` path? Something like passing the search params from the caller:

```ts
// In handleRedirectCallback - pass search params
return this._handleRedirectCallback(params, { signUp, signIn, navigate }, new URLSearchParams(window.location.search));

// In handleGoogleOneTapCallback - don't pass search params (keeps current behavior)
return this._handleRedirectCallback(params, { signUp, signIn, navigate });
```

Then `_handleRedirectCallback` would forward whatever it receives to `RedirectUrls` instead of reading `window.location.search` directly.

jfoshee: That's a good point. Thanks, Nikos. I've moved search params up a level and added an optional parameter to `_handleRedirectCallback`.  I manually tested the OAuth app + sso sign-up flow (bug scenario) and it works fine with `<AuthenticateWithRedirectCallback />`. I tested basic sso sign-up w/ customized sign-in page as well. Are there any other scenarios I should test?

packages/clerk-js/src/core/__tests__/clerk.test.ts
```diff
@@ -2075,6 +2075,66 @@ describe('Clerk singleton', () => {
...
+      const sut = new Clerk(productionPublishableKey);
+      await sut.load(mockedLoadOptions);
+      sut.setActive = mockSetActive;
+
+      await sut.handleRedirectCallback();
+
+      await waitFor(() => {
+        expect(mockSetActive).toHaveBeenCalled();
+        // The redirect should include the custom redirect URL from URL search params
+        expect(mockNavigate.mock.calls[0][0]).toContain(customRedirectUrl);
```
nikosdouvlis, line 2135: Would it be worth adding a test where both props and search params are present? Just to lock the precedence behavior (search params > props) and prevent future regressions

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2064,6 +2064,7 @@ export class Clerk implements ClerkInterface {
       signUp: SignUpResource;
       navigate: (to: string) => Promise<unknown>;
     },
+    searchParams?: URLSearchParams,
```
nikosdouvlis, line 2067: Should we add a quick test verifying that `handleGoogleOneTapCallback` ignores `redirect_url` from the URL? Would help protect the intentional scoping here

packages/clerk-js/src/core/__tests__/clerk.test.ts
jfoshee: Yes, good idea. Done.

packages/clerk-js/src/core/clerk.ts
jfoshee: While I agree with not changing the GoogleOneTap behavior, I'm a little hesitant to add a test because I don't know what the behavior should be... 

I did do some manual tests and the behavior seemed unchanged regardless of the search params. Probably in part because with GoogleOneTap, my sample app never redirects and just stays on the custom sign in page.

I suppose someone implementing both custom sign-in and GoogleOneTap and supporting OAuth integrations would figure out they need to handle that...

If you think it's merited I can easily add the test. But, that's my thinking. Let me know your perspective. :-)

PR #7602, opened 2026-01-15
feat(backend): Add Frontend API proxy helpers

brkalow:
## Description

Adds Frontend API proxy support to `clerkMiddleware` for both Next.js and Express. This enables scenarios where direct communication with Clerk's API is blocked or needs to go through the application server.

### Next.js

```typescript
import { clerkMiddleware } from '@clerk/nextjs/server';

export default clerkMiddleware({
  // Enable proxy with defaults (path: '/__clerk')
  frontendApiProxy: {
    enabled: true,
  },
});

// Or with multi-domain support using a function
export default clerkMiddleware({
  frontendApiProxy: {
    enabled: (url) => url.hostname === 'app.example.com',
  },
});

// Custom proxy path
export default clerkMiddleware({
  frontendApiProxy: {
    enabled: true,
    path: '/custom-clerk-proxy',
  },
});
```

### Express

```typescript
import express from 'express';
import { clerkMiddleware } from '@clerk/express';

const app = express();

// Enable proxy with defaults (path: '/__clerk')
app.use(clerkMiddleware({ frontendApiProxy: { enabled: true } }));

// Custom proxy path
app.use(clerkMiddleware({ frontendApiProxy: { enabled: true, path: '/my-proxy' } }));

// Disable proxy
app.use(clerkMiddleware({ frontendApiProxy: { enabled: false } }));
```

## Key Features

- **Embedded in clerkMiddleware**: No separate middleware needed - proxy handling is built into `clerkMiddleware`
- **Auto-derived proxyUrl**: The `proxyUrl` for handshake redirects is automatically derived from the `frontendApiProxy` config
- **Multi-domain support (Next.js)**: Use a function for `enabled` to conditionally enable proxy based on request URL
- **Streaming body support**: Request and response bodies are streamed for efficient handling
- **Header rewriting**: Redirect responses are automatically rewritten to use the proxy URL

## Type of change

- [x] 🌟 New feature

## Testing

- Unit tests added for proxy functionality in both Next.js and Express
- Tests cover: proxy path matching, proxyUrl derivation, multi-domain support, header handling

## Summary by CodeRabbit

* **New Features**
  * Frontend API proxy added across backend, Next.js, and Express with configurable proxy path, automatic proxy URL derivation, and App Router handler helpers.

* **Behavior**
  * Proxy streams requests/responses, preserves queries/headers, manages X-Forwarded headers, strips hop-by-hop headers, and rewrites internal redirects.

* **Tests**
  * Extensive tests for routing, header propagation, error handling, redirects, and multi-domain scenarios.

* **Documentation**
  * Changeset documenting the Frontend API proxy feature.


packages/backend/src/proxy.ts
```diff
@@ -0,0 +1,276 @@
...
+/**
+ * Checks if a request path matches the proxy path.
+ * @param request - The incoming request
+ * @param options - Proxy options including the proxy path
+ * @returns True if the request matches the proxy path
+ */
+export function matchProxyPath(request: Request, options?: Pick<FrontendApiProxyOptions, 'proxyPath'>): boolean {
+  const proxyPath = options?.proxyPath || DEFAULT_PROXY_PATH;
+  const url = new URL(request.url);
+  return url.pathname.startsWith(proxyPath);
```
nikosdouvlis: Should we check for an exact match or a `/` boundary here? `startsWith` alone means something like `/__clerk-admin` would incorrectly match the default `/__clerk` prefix.

```typescript
return url.pathname === proxyPath || url.pathname.startsWith(proxyPath + '/');
```

packages/express/src/authenticateRequest.ts
```diff
@@ -101,17 +104,79 @@ export const authenticateAndDecorateRequest = (options: ClerkMiddlewareOptions =
...
+        }
+        return;
+      }
+    }
+
+    // Auto-derive proxyUrl from frontendApiProxy config if not explicitly set
+    let resolvedOptions = options;
+    if (frontendApiProxy && proxyEnabled && !options.proxyUrl) {
+      const protocol = request.secure || request.headers['x-forwarded-proto'] === 'https' ? 'https' : 'http';
+      const host = request.headers['x-forwarded-host'] || request.headers.host || 'localhost';
```
nikosdouvlis: Should we handle the case where `x-forwarded-host` is an array? Express types it as `string | string[] | undefined`, so if it's an array this would produce something like `"https://host1,host2/__clerk"`.

```typescript
const forwardedHost = request.headers['x-forwarded-host'];
const host = (Array.isArray(forwardedHost) ? forwardedHost[0] : forwardedHost) || request.headers.host || 'localhost';
```

packages/nextjs/package.json
```diff
@@ -59,6 +59,11 @@
       "types": "./dist/types/legacy.d.ts",
       "import": "./dist/esm/legacy.js",
       "require": "./dist/cjs/legacy.js"
+    },
+    "./proxy": {
```
nikosdouvlis: I like the separate subpath, but what's the reasoning over re-exporting from `@clerk/nextjs/server`? Tree-shaking, keeping the proxy opt-in at the import level, something else?

brkalow: Good question, I don't have a great reason. I agree, we can export from `/server`. It's fairly lightweight!

packages/express/src/authenticateRequest.ts
brkalow: yes, good call! I was trying to find explicit docs on x-forwarded-host being an array, seems like it can come in as a comma-separated list and we should use the left-most value, like you've suggested

packages/nextjs/src/server/index.ts
```diff
@@ -74,3 +74,14 @@ export type {
...
+/**
+ * Frontend API proxy exports
+ */
+export {
+  clerkFrontendApiProxy,
+  createFrontendApiProxyHandlers,
+  type FrontendApiProxyHandlers,
+  type FrontendApiProxyOptions,
+  type NextFrontendApiProxyOptions,
+} from './proxy';
```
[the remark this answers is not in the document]
brkalow: This is a sub-path import already, not the root index file. See existing re-exports in this file as reference

packages/express/src/authenticateRequest.ts
```diff
@@ -97,21 +99,101 @@ const absoluteProxyUrl = (relativeOrAbsoluteUrl: string, baseUrl: string): strin
   return new URL(relativeOrAbsoluteUrl, baseUrl).toString();
 };
 
+/**
+ * Removes trailing slashes from a string without using regex
+ * to avoid potential ReDoS concerns flagged by security scanners.
+ */
+function stripTrailingSlashes(str: string): string {
```
jacekradko: [nit] I think this might be duped

packages/express/src/types.ts
```diff
@@ -7,6 +7,23 @@ export type ExpressRequestWithAuth = ExpressRequest & {
...
 
+/**
+ * Options for configuring Frontend API proxy in clerkMiddleware
+ */
+export interface FrontendApiProxyOptions {
+  /**
+   * Enable Frontend API proxy handling. When true, requests to the proxy path
+   * will be proxied to Clerk's Frontend API and the proxyUrl will be auto-derived.
+   */
+  enabled?: boolean;
```
jacekradko: [nit] I think it would make sense to make this a required configuration value so it's easy to grok if the feature is on or off

.changeset/frontend-api-proxy.md
```diff
@@ -0,0 +1,7 @@
+---
+'@clerk/backend': minor
+'@clerk/nextjs': minor
+'@clerk/express': minor
+---
```
jacekradko, line 6:
```suggestion
'@clerk/shared': minor
---
```

packages/express/src/types.ts
brkalow: @jacekradko required if `frontendApiProxy` is present right?

jacekradko: Yep

PR #7608, opened 2026-01-15
feat(repo): add local Verdaccio workflow for testing packages

nikosdouvlis:
## Summary

Testing package changes locally has been painful:
- npm/pnpm linking doesn't always work and doesn't mimic real published package behavior (symlinks behave differently than installed packages)
- YALC is an older solution with the same limitation - it doesn't map 1:1 with how packages behave when published to npm
- The common workaround was pushing to a branch and waiting for snapshot releases, which takes 5+ minutes per iteration from push to install

Verdaccio solves this by running a local npm registry. Packages are built and published exactly as they would be to npm, then installed normally in test apps. This makes iteration cycles much shorter while testing real package behavior.

## What changed

- Added `local:registry:up` to start Verdaccio with npm proxy support
- Added `local:registry:down` to stop the registry
- Added `local:registry:pub` to build and publish all packages
- Script displays install instructions when registry starts
- Git changes are auto-reset after publishing

## Usage

In the monorepo:

```bash
# Terminal 1 - Start the local registry
pnpm local:registry:up

# Terminal 2 - Build and publish all packages
pnpm local:registry:pub

# Or publish a single package
pnpm local:registry:pub:pkg @clerk/backend

# When done
pnpm local:registry:down
```

In your test app:

```bash
# Update package.json catalog to use "local" tag:
"@clerk/backend": "local",

# Install from local Verdaccio
bun install --registry http://localhost:4873
```

## Test plan

- [x] Start registry with `pnpm local:registry:up`
- [x] Publish packages with `pnpm local:registry:pub`
- [x] Install in test app with `bun install --registry http://localhost:4873`
- [x] Verify installed versions match local build

## Summary by CodeRabbit

* **New Features**
  * Added npm scripts to manage a local development registry: `local:registry:up` (start), `local:registry:down` (stop), and `local:registry:pub` (publish packages)
  * Increased upload size limit for the local registry to 200MB to support larger packages


bratsos (approved):
🎖️

octoper (approved):
🔥

PR #7653, opened 2026-01-22
feat(upgrade): add satellite auto-sync codemod for core 3 migration

nikosdouvlis:
## Why

Core 3 introduced `satelliteAutoSync` (defaults to `false`), changing satellite behavior. In Core 2, satellite apps always auto-synced on first visit. In Core 3, they don't unless `satelliteAutoSync: true` is explicitly set. Users upgrading need this prop added automatically to preserve existing behavior.

## What changed

- Add `transform-satellite-auto-sync.cjs` codemod that finds `isSatellite` usage (both JSX props and object properties) and adds `satelliteAutoSync: true` as a sibling
- Skip if `satelliteAutoSync` is already present
- Register the codemod for nextjs, react, expo, astro, and tanstack-react-start packages

## Packages affected

- `@clerk/upgrade`: New codemod + tests

## Summary by CodeRabbit

* **New Features**
  * Automatic injection of satelliteAutoSync during Core 3 upgrades when satellite mode is detected for nextjs, react, expo, astro, and tanstack-react-start.

* **Tests**
  * Added comprehensive fixtures and tests validating the satellite auto-sync transformation across JSX and object configurations.

* **Documentation**
  * Added a changeset documenting the new codemod and its behavior.


packages/upgrade/src/codemods/transform-satellite-auto-sync.cjs
Ephem, line 1: Is adding `satelliteAutoSync` to every component and object that has a `isSatellite` overly broad, given there might be unrelated components/objects that also has it?

I get this is tricky to narrow down given props and objects can be passed around, is the idea that we'd rather have false positives than miss some?

Seems like an edge case, should still be safe and most people review the output after running a codemod anyway so I'm very fine with this, just wanted to double check this is the intent?

packages/upgrade/src/codemods/transform-satellite-auto-sync.cjs
```diff
@@ -0,0 +1,74 @@
...
+
+    const hasAutoSync = properties.some(prop => isObjectPropertyNamed(prop, 'satelliteAutoSync'));
+    if (hasAutoSync) {
+      return;
+    }
+
+    const isSatelliteIndex = properties.findIndex(prop => isObjectPropertyNamed(prop, 'isSatellite'));
+    const autoSyncProp = j.objectProperty(j.identifier('satelliteAutoSync'), j.booleanLiteral(true));
+    properties.splice(isSatelliteIndex + 1, 0, autoSyncProp);
+    dirty = true;
```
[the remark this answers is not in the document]
Ephem: @nikosdouvlis This is an interesting point but in reverse. We likely _want_ the spread to win so any explicit setting is respected. Safest would be to place the new prop/property before any spreads, but bar that maybe we at least want the component and object transform to behave the same?

Very much a NIT though.

packages/upgrade/src/codemods/transform-satellite-auto-sync.cjs
nikosdouvlis: Thanks for the review Fredrik :) I thought about this as well and I decided that I could live with false positives given how impactful the change in behavior could be for apps migrating from Core 2 to Core 3. Happy to change this of course

nikosdouvlis: Thinking more about this and your comment below, I decided to align the behavior and always add the prop at the end - I do think its important for as to override always, and let the end user decide in the very rare event that they have already added satelliteAutoSync manually (very unlikely as this prop will be introduced with core3)

Ephem: Very true, sounds good to me! 🙏

PR #7667, opened 2026-01-23
feat(ui): enforce minimum clerk-js version compatibility

bratsos:
Add runtime version check in ClerkUi constructor to detect incompatible @clerk/clerk-js versions. In development instances, logs a warning; in production, throws ClerkRuntimeError with actionable upgrade guidance.

Also moves shared version utilities to @clerk/shared/utils.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Runtime version compatibility check for Clerk.js during UI initialization (throws in development, warns in production).
  * Added a convenience API to check if a version meets a minimum requirement.
  * Introduced a new minimum-supported Clerk.js version constant.

* **Tests**
  * Added comprehensive tests covering version checks, UI init behaviors (dev vs prod), pre-releases, and edge cases.

* **Chores**
  * Added changeset and updated bundle size threshold.

### DIFF
--- a/packages/ui/bundlewatch.config.json
+++ b/packages/ui/bundlewatch.config.json
@@ -1,7 +1,7 @@
 {
   "files": [
     { "path": "./dist/ui.browser.js", "maxSize": "36KB" },
-    { "path": "./dist/ui.legacy.browser.js", "maxSize": "74KB" },
+    { "path": "./dist/ui.legacy.browser.js", "maxSize": "76KB" },
     { "path": "./dist/framework*.js", "maxSize": "44KB" },
     { "path": "./dist/vendors*.js", "maxSize": "73KB" },
     { "path": "./dist/ui-common*.js", "maxSize": "130KB" },


packages/ui/src/constants.ts
```diff
@@ -1,3 +1,5 @@
+export const MIN_CLERK_JS_VERSION = '5.112.0';
```
nikosdouvlis: Is 5.112.0 the correct version here or should this be pinned to 6.0.0 since thats the major for core 3? The tests seem to use 6.0.0 as the expected minimum which made me wonder if this was intentional or if the constant just needs updating.

packages/ui/src/ClerkUi.ts
```diff
@@ -16,6 +20,33 @@ export class ClerkUi implements ClerkUiInstance {
...
+        `but found an incompatible version${clerkVersion ? ` (${clerkVersion})` : ''}. ` +
+        `Please upgrade @clerk/clerk-js (or your framework SDK) to a compatible version.`;
+    }
+
+    if (incompatibilityMessage) {
+      if (isDevelopmentInstance) {
+        logger.warnOnce(incompatibilityMessage);
+      } else {
+        throw new ClerkRuntimeError(incompatibilityMessage, { code: 'clerk_ui_version_mismatch' });
+      }
```
nikosdouvlis, line 47: Should this be the other way around? Typically we want to fail fast in development (so devs catch issues early) but be more lenient in production (to avoid breaking user-facing apps). The current behavior warns in dev (easy to miss) but throws in prod (could cause outages).

packages/ui/src/ClerkUi.ts
```diff
@@ -16,6 +20,33 @@ export class ClerkUi implements ClerkUiInstance {
...
+      incompatibilityMessage =
+        `@clerk/ui@${ClerkUi.version} requires @clerk/clerk-js@>=${MIN_CLERK_JS_VERSION}, ` +
+        `but found @clerk/clerk-js@${clerkVersion}. ` +
+        `Please upgrade @clerk/clerk-js (or your framework SDK) to a compatible version.`;
+    } else if (!parsedVersion && !moduleManager) {
+      incompatibilityMessage =
+        `@clerk/ui@${ClerkUi.version} requires @clerk/clerk-js@>=${MIN_CLERK_JS_VERSION}, ` +
+        `but found an incompatible version${clerkVersion ? ` (${clerkVersion})` : ''}. ` +
+        `Please upgrade @clerk/clerk-js (or your framework SDK) to a compatible version.`;
+    }
```
nikosdouvlis, line 40: Looking at clerk-js, it always instantiates ClerkUi with `new ModuleManager()`, and `clerkVersion` comes from the Clerk instance which should always have its version set. Is there a scenario where ClerkUi would be instantiated without a moduleManager or without knowing the clerkVersion? If this is just defensive coding thats fine, but if theres an actual use case Im curious what it is :)

nikosdouvlis:
~~this PR adds a new feature to @clerk/ui and @clerk/shared but I dont see a changeset for it. Should we add one describing the new minimum version check behavior?~~

This is fine

packages/ui/src/ClerkUi.ts
bratsos: It's just defensive to future-proof us from a mistake. I thought about all different scenarios with pinned/hotloaded clerk-js and before/after we introduced `moduleManager`/`ClerkUi` and I don't think you land in this case right now, but you never know in the future.

packages/ui/src/ClerkUi.ts
bratsos: Yeah that makes more sense. I went with "You don't want a mismatch on production" as you may end up with weird-looking UI, but I agree. https://github.com/clerk/javascript/pull/7667/changes/4e1751d0a2ad05cc163809e6aabbfe10b14bb630

packages/ui/src/constants.ts
bratsos: I think the version when we added ModuleManager/ClerkUi was `5.112.0` so in theory this could be the minimum clerk js version. But we can also arbitrarily set this to `6.0.0` to make it available only on Core3. WDYT?

packages/ui/src/__tests__/ClerkUi.test.ts
```diff
@@ -0,0 +1,216 @@
...
+  const mockOptions: ClerkOptions = {};
+  const getEnvironment = () => null;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  describe('compatible versions', () => {
+    test('accepts exact minimum version (6.0.0)', () => {
+      const getClerk = () => ({ version: '6.0.0', instanceType: 'production' }) as Clerk;
```
jacekradko: For maintainability sake, we should always accept the exact minimum

```suggestion
      const getClerk = () => ({ version: MIN_CLERK_JS_VERSION, instanceType: 'production' }) as Clerk;
```

bratsos: Sure, I updated it here https://github.com/clerk/javascript/pull/7667/changes/73ebcb5f3c686a274f7dea2d597ba1f6c5c2f694

packages/ui/src/constants.ts
bratsos: I updated this to `6.0.0` https://github.com/clerk/javascript/pull/7667/changes/73ebcb5f3c686a274f7dea2d597ba1f6c5c2f694

PR #7678, opened 2026-01-26
feat(nextjs,react): Add HandleSSOCallback component

dstaley:
## Description

This PR adds a new `HandleSSOCallback` component which wraps our custom ui flow for handling transferrable sign-ins and sign-ups. This component packages the specific logic that we use to handle that flow in our own components; customers will still be able to implement their own specific custom flow, but this component is being made available to make that a little easier.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Adds a HandleSSOCallback component to manage SSO callback flows: finalizes sign-in/sign-up, supports transfer between sign-up/sign-in, activates existing sessions, handles MFA/password cases, and navigates to app/sign-in/sign-up; renders a default container that can be customized.

* **Tests**
  * Adds comprehensive tests covering callback flows, transfer logic, session activation, navigation outcomes, and one-time execution behavior.

* **Chores**
  * Bumps minor package versions.

### DIFF
--- a/packages/react/src/components/__tests__/HandleSSOCallback.test.tsx
+++ b/packages/react/src/components/__tests__/HandleSSOCallback.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen, waitFor } from '@testing-library/react';
+import { render, waitFor } from '@testing-library/react';
 import React from 'react';
 import { afterAll, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
 


packages/react/src/components/HandleSSOCallback.tsx
```diff
@@ -0,0 +1,177 @@
...
+  }, [clerk, signIn, signUp]);
+
+  if (render) {
+    return render();
+  }
+
+  return (
+    <div>
+      {/* Because a sign-in transferred to a sign-up might require captcha verification, make sure to render the
+  captcha element. */}
```
alexcarpenter, line 166: should we allow children for them to render a spinner here possibly?

dstaley: I went with the `render` prop for that, but I'll defer to what you suggest here since I know the thinking has evolved recently!

alexcarpenter: to date we've used children, so I might lean towards children for now. We've yet to introduce a render prop but possibly in the future.

dstaley: stepping back do we even need either? This renders a div and the captcha (possibly). wouldn't folks probably write code that looks like this?

```
<Card>
  <CardBody>
    <Spinner />
    <HandleSSOCallback />
  </CardBody>
</Card>
```

In this situation I don't think they'd want to put their Spinner as a child of `HandleSSOCallback`. If it's a render prop it would make more sense for it to be:
```
<HandleSSOCallback render={() => {
  return (
    <Card><CardBody><Spinner /></CardBody></Card>  
  )
}} />

File: packages/react/src/components/HandleSSOCallback.tsx [hunk 2]
```diff
@@ -0,0 +1,178 @@
...
+              navigateToApp(...params);
+            },
+          });
+          return;
+        }
+        // The sign-in requires additional verification, so we need to navigate to the sign-in page.
+        return navigateToSignIn();
+      }
+
+      if (
```
nikosdouvlis, line 108: Just double checking - this navigates away unless ALL factors are `enterprise_sso`. Is that the intent?

packages/react/src/components/HandleSSOCallback.tsx
```diff
@@ -0,0 +1,178 @@
...
+    (async () => {
+      if (!clerk.loaded || hasRun.current) {
+        return;
+      }
+      // Prevent re-running this effect if the page is re-rendered during session activation (such as on Next.js).
+      hasRun.current = true;
+
+      // If this was a sign-in, and it's complete, there's nothing else to do.
+      // Note: We perform a cast
+      if ((signIn.status as string) === 'complete') {
```
nikosdouvlis, line 84: `// Note: We perform a cast` - comment is cut off (i think?), should it explain why?

packages/react/src/components/HandleSSOCallback.tsx
```diff
@@ -0,0 +1,178 @@
...
+            session: sessionId,
+            navigate: async (...params) => {
+              return navigateToApp(...params);
+            },
+          });
+          return;
+        }
+      }
+    })();
+  }, [clerk, signIn, signUp]);
```
nikosdouvlis: Navigate callbacks missing from deps array - intentional?

packages/react/src/components/HandleSSOCallback.tsx
```diff
@@ -0,0 +1,178 @@
...
+    return render();
+  }
+
+  return (
+    <div>
+      {/* Because a sign-in transferred to a sign-up might require captcha verification, make sure to render the
+  captcha element. */}
+      <div id='clerk-captcha' />
+    </div>
+  );
```
nikosdouvlis, line 169: CodeRabbit flagged this too - no test coverage for this component

nikosdouvlis: Is this something e2e tests are better suited for?

packages/react/src/components/HandleSSOCallback.tsx
dstaley: This is directly copy and pasted from our existing `handleRedirectCallback` implementation. It only navigates to sign-in if the status is `needs_first_factor` and there is at least one supported first-factor that's NOT `enterprise_sso`.

packages/react/src/components/HandleSSOCallback.tsx
dstaley: Yes they're omitted since we're trying to avoid re-running this effect. We want to run this effect _once_ once Clerk is loaded.

packages/react/src/components/HandleSSOCallback.tsx
dstaley: whoops 😬

packages/react/src/components/HandleSSOCallback.tsx
dstaley: given the number of permutations I actually think a unit test with mocks might be better. I'll add that!

packages/react/src/components/HandleSSOCallback.tsx
alexcarpenter: I think this could be ok

```tsx
<Card>
  <CardBody>
    <Spinner />
    <HandleSSOCallback />
  </CardBody>
</Card>
```

because hypothetically the captcha could render right, and I want to be able to place it in a specific spot within the card. Just feels weird due to the naming of the component.

packages/react/src/components/HandleSSOCallback.tsx
```diff
@@ -0,0 +1,179 @@
...
+            session: sessionId,
+            navigate: async (...params) => {
+              return navigateToApp(...params);
+            },
+          });
+          return;
+        }
+      }
+    })();
+  }, [clerk, signIn, signUp]);
```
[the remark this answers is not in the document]
jacekradko: @dstaley Thoughts on this one?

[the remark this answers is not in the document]
dstaley: we have some other hooks that just depend on `clerk`, but I don't see any harm in also depending on `clerk.loaded` for robustness

PR #7687, opened 2026-01-27
refactor(shared): rename billing hook and update semantics

jacekradko:
## Summary
- Renames `useBillingHookEnabled` to `useBillingIsEnabled`
- Updates semantics: calling without `for` param now returns `true` if **either** user or organization billing is enabled (previously defaulted to user billing only)

## New API
```typescript
// returns true if either user or org billing is enabled
useBillingIsEnabled()

// check specific billing type
useBillingIsEnabled({ for: 'user' })
useBillingIsEnabled({ for: 'organization' })
```

## Test plan
- [ ] Verify type checking passes
- [ ] Manual testing with billing features

## Summary by CodeRabbit

* **Refactor**
  * Improved billing eligibility logic to distinguish organization vs. user contexts; internal API renamed with no change to user-facing behavior.
* **Chores**
  * Added a changeset entry for the update.
* **Tests**
  * Updated tests to cover scenarios where billing is disabled for user and/or organization.

### DIFF
--- a/.changeset/fiery-carrots-call.md
+++ b/.changeset/fiery-carrots-call.md
@@ -1,2 +1,5 @@
 ---
+'@clerk/shared': major
 ---
+
+Rename internal `useBillingHookEnabled` to `useBillingIsEnabled` with improved semantics for authentication and organization context checks.


.changeset/fiery-carrots-call.md
```diff
@@ -0,0 +1,2 @@
+---
```
dstaley, line 1: probably needs to be a major changes since we're essentially removing a public API and replacing it with a new one

dstaley (approved):

.changeset/fiery-carrots-call.md
jacekradko: Done

jacekradko:
!allow-major

PR #7730, opened 2026-02-02
feat(*): Make useAuth().getToken() throw during SSR

Ephem:
## Description

The [uSES PR](https://github.com/clerk/javascript/pull/7411) removed the `useAuth` wrapper from the `nextjs` package. That wrapper made `getToken` be `undefined` in server environments, and removing that caused calling `useAuth().getToken()` to stall forever during server-side rendering.

This PR tackles this at the `react` package level instead, since it's a common concern. During server-side rendering, `getToken` is now a function that throws a `clerk_runtime_not_browser` error.

The error aligns with this PR: [Make gettoken callable outside of react](https://github.com/clerk/javascript/pull/7325)

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Breaking Changes**
  * `useAuth().getToken` now throws an error in non-browser environments instead of returning undefined. Wrap calls in try/catch or use only within browser contexts (useEffect, event handlers, or non-suspenseful data fetching). Server-side token access requires using the Auth object instead.

* **Documentation**
  * Updated guidance and error messages for server-side rendering token retrieval scenarios.


Ephem:
!allow-major

PR #7749, opened 2026-02-03
feat(clerk-js): Add signUpIfMissing for SignIn

dmoerner:
## Description

We are building a new sign-in-or-sign-up flow compatible with strict enumeration protection. It is meant to complement our existing sign-in-or-sign-up flow.

Our current sign-in-or-sign-up flow is managed by the SDKs: We start with a sign in attempt, and on a 422 (when an account does not exist), we redirect to a sign up. The flow is thus: Sign In (422) -> Sign Up (200) -> Attempt Verifications for VerifiedAtSignUp identifiers (200). This is vulnerable to user enumeration attacks because the attacker sees the sign in to sign up redirect before they prove their identity by completing a verification.

When `sign_up_if_missing` is passed as a param when POSTing a sign in, instead we do the following: Sign In (200) -> Attempt Verification for Identifier (200) -> Return transferable identification -> Transfer to Sign Up. This is enumeration safe, because you only see if an account already existed or was created after you verify your identity.

This PR is the first step in SDK support for this new flow. We add support for the optional `signUpIfMissing` param on sign in. We also add captcha support for `SignIn`. This is all optional and currently in testing with custom components. Support in Clerk components will be in future PRs.

Implementation Notes:

- To avoid adding dependencies between legacy and future code, we duplicate some logic between legacy and future classes. This matches existing patterns on SignUp.
- Now that we support captcha on sign in, we need to make sure we don't show a second captcha after transfer. We modify the sign up captcha handling to skip captchas on such transfers. There's no corresponding logic in the sign in captcha handling because we don't do sign up to sign in transfers outside of the OAuth flow.
- I deleted a pointless update of the strategy in transfer sign ups; the backend ignores strategies passed on transfers already so we do not need to set it in the SDK before making a request.

Fixes USER-4596

## Checklist

- [ ] 
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * SignIn.create now supports the `sign_up_if_missing` parameter, enabling seamless user sign-up fallback during sign-in.
  * CAPTCHA protection is automatically integrated into sign-in processes when required.
  * Browser locale is automatically detected and included in sign-in requests for improved localization.

### DIFF
--- a/packages/clerk-js/bundlewatch.config.json
+++ b/packages/clerk-js/bundlewatch.config.json
@@ -5,7 +5,7 @@
     { "path": "./dist/clerk.chips.browser.js", "maxSize": "66KB" },
     { "path": "./dist/clerk.legacy.browser.js", "maxSize": "108KB" },
     { "path": "./dist/clerk.no-rhc.js", "maxSize": "307KB" },
-    { "path": "./dist/clerk.native.js", "maxSize": "65KB" },
+    { "path": "./dist/clerk.native.js", "maxSize": "66KB" },
     { "path": "./dist/vendors*.js", "maxSize": "7KB" },
     { "path": "./dist/coinbase*.js", "maxSize": "36KB" },
     { "path": "./dist/base-account-sdk*.js", "maxSize": "203KB" },


packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -162,12 +164,43 @@ export class SignIn extends BaseResource implements SignInResource {
...
+    if (requiresCaptcha) {
+      const captchaChallenge = new CaptchaChallenge(SignIn.clerk);
+      const captchaParams = await captchaChallenge.managedOrInvisible({ action: 'signin' });
+      if (!captchaParams) {
+        throw new ClerkRuntimeError('', { code: 'captcha_unavailable' });
+      }
+      finalParams = { ...finalParams, ...captchaParams };
+    }
+
+    if (params.transfer && this.shouldBypassCaptchaForAttempt(params)) {
```
dmoerner: This check guards against an attacker who tries to make a fake `transfer` and then changes `strategy` at the same time, which would be an attempt to get around WAF rules and Captcha rules that exclude transfers. We already guard against this on the backend on both sign in and sign up, but in the SDK we were only checking it on sign up. So I took this opportunity to also add the check on sign in.

packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -162,12 +163,37 @@ export class SignIn extends BaseResource implements SignInResource {
...
+      const strategy = SignIn.clerk.client?.signUp.verifications.externalAccount.strategy;
+      if (strategy) {
+        // When transfer is true, we're in the OAuth/Enterprise SSO transfer case
+        type TransferParams = Extract<SignInCreateParams, { transfer: true }>;
+        (finalParams as TransferParams).strategy = strategy as TransferParams['strategy'];
+      }
+    }
```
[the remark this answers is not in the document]
dmoerner: If it's a sign in and a transfer and there's no signup or verification or externalaccount, that would be an error.

packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -574,6 +600,76 @@ export class SignIn extends BaseResource implements SignInResource {
...
+
+    if (
+      params.transfer &&
+      captchaOauthBypass.some(
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        strategy => strategy === SignIn.clerk.client!.signUp.verifications.externalAccount.strategy,
+      )
+    ) {
+      return true;
+    }
```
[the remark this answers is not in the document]
dmoerner: As in the other case, a transfer with a sign up that's incomplete in the sense of missing an external account is an error.

packages/shared/src/types/signInCommon.ts
```diff
@@ -162,8 +162,18 @@ export type SignInCreateParams = (
...
+      strategy?: OAuthStrategy | EnterpriseSSOStrategy;
+    }
+) & {
+  transfer?: boolean;
+  locale?: string;
+  sign_up_if_missing?: boolean;
+  captcha_token?: string;
+  captcha_error?: unknown;
+  captcha_widget_type?: string | null;
+};
```
dmoerner, line 171: My intention with these types was for `SignInCreateParams` to match what we send to FAPI, which is why I added to the discriminated union. Another approach would be to reserve it for what is passed into the SDK method, and use implicit typing or just `Record<string,unknown>` for what we sent to FAPI. The latter approach actually fits better what we already do on sign up but it felt haphazard to me.

packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -173,13 +173,6 @@ export class SignUp extends BaseResource implements SignUpResource {
       finalParams = { ...finalParams, ...captchaParams };
     }
 
-    if (finalParams.transfer && this.shouldBypassCaptchaForAttempt(finalParams)) {
```
dmoerner, line 176: I cleaned up this code path, which is pointless: On a transfer, the backend already pulls the strategy from the transfer and ignores any strategy included in the query params. We don't need the SDK to also pull the strategy from the transfer and include it in the query params.

packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -814,11 +875,85 @@ class SignInFuture implements SignInFutureResource {
...
+      body.locale = locale;
+    }
+
+    const { captchaToken, captchaWidgetType, captchaError } = await this.getCaptchaToken(params);
+
+    if (captchaToken !== undefined) {
+      body.captchaToken = captchaToken;
+      body.captchaWidgetType = captchaWidgetType;
+      body.captchaError = captchaError;
+    }
```
wobsoriano: [`SignUpFuture._create`](https://github.com/clerk/javascript/blob/6363edcaa356d0bcf19bfbe876524030579c6fa8/packages/clerk-js/src/core/resources/SignUp.ts#L836,L842) sends all three captcha fields to FAPI regardless of whether `captchaToken` is set, was the guard here intentional or should it match?

wobsoriano (approved):
This looks good, just left a minor question!

packages/clerk-js/src/core/resources/SignIn.ts
dmoerner: Good catch. We definitely don't want to send these to the backend if empty, but the guard here is unnecessary, I now see that we have a [filterUndefinedValues](https://github.com/clerk/javascript/blob/main/packages/clerk-js/src/utils/filterUndefinedValues.ts) helper that we run which will filter these out anyway. I'll remove it.

PR #7751, opened 2026-02-03
fix(ui): Prevent infinite loading spinner on factor-two without active 2FA

alexcarpenter:
## Description

Fixes an issue where users landing on the factor-two sign-in route without an active 2FA session would see an infinite loading spinner. This can occur when navigating directly to the URL with a force redirect parameter. The fix adds a status check to redirect users back to the sign-in start page if the sign-in status is not 'needs_second_factor', matching the existing pattern in SignInFactorOne.

## Test Plan

- ✅ Added test case for navigation when sign-in status is invalid
- ✅ All 29 existing SignInFactorTwo tests pass
- ✅ Test fails without the fix (confirms bug), passes with the fix (confirms fix works)

## Type of change

- [x] 🐛 Bug fix

## Bug Fixes
* Fixed an infinite loading spinner that appeared when accessing the two-factor sign-in page without an active 2FA session. Users are now properly redirected to the main sign-in page in this scenario.

### DIFF
--- a/.changeset/fix-signin-factor-two-spinner.md
+++ b/.changeset/fix-signin-factor-two-spinner.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Fix infinite loading spinner when navigating to factor-two sign-in route without an active 2FA session


packages/ui/src/components/SignIn/SignInFactorTwo.tsx
```diff
@@ -22,6 +28,18 @@ function SignInFactorTwoInternal(): JSX.Element {
...
+  React.useEffect(() => {
+    if (__internal_setActiveInProgress) {
+      return;
+    }
+
+    // If the sign-in doesn't require second factor verification,
+    // redirect back to the start of the sign-in flow
+    if (signIn.status !== 'needs_second_factor') {
+      void router.navigate('../');
+    }
```
Ephem, line 40: Could we detect if the user is signed in and redirect to any present redirect url in that case? 🤔

alexcarpenter:
closing in favor of https://github.com/clerk/javascript/pull/7774

PR #7753, opened 2026-02-03
test(clerk-js): add tests for null resource update behavior

jacekradko:
## Summary

Adds tests for the `shouldIgnoreNullUpdate` mechanism in the `State` class to verify the behavior described in [USER-4372](https://linear.app/clerk/issue/USER-4372).

### Investigation Findings

The `shouldIgnoreNullUpdate` function at `packages/clerk-js/src/core/state.ts:117-119` is designed to prevent null resource updates from propagating when the previous resource has not been finalized:

```typescript
function shouldIgnoreNullUpdate(previousResource, newResource) {
  return !newResource?.id && previousResource && previousResource.__internal_future?.canBeDiscarded === false;
}
```

**Tests verify that the mechanism works correctly:**

1. ✅ Allows first resource update when previous resource is `null`
2. ✅ Ignores null update when previous resource exists and `canBeDiscarded === false`
3. ✅ Allows null update after `finalize()` sets `canBeDiscarded = true`
4. ✅ Allows updates when new resource has an id (not a null update)
5. ✅ Handles rapid successive updates correctly
6. ✅ Handles updates with same instance correctly

### Key Insight: Timing Issue in `finalize()`

The reported flash of the default sign-up form likely occurs due to the order of operations in `finalize()`:

```typescript
async finalize(...) {
  this.#canBeDiscarded = true;  // ← Set BEFORE setActive
  await SignUp.clerk.setActive({ session: this.#resource.createdSessionId, navigate });
}
```

**The sequence:**
1. User verifies email → sign up is complete
2. User/code calls `finalize()`
3. `canBeDiscarded` is set to `true` **immediately**
4. `setActive()` is called
5. In Next.js, `setActive()` triggers `router.refresh()` or navigation
6. During refresh, client is re-fetched with `sign_up: null`
7. Because `canBeDiscarded` is already `true`, the null update is **NOT ignored**
8. Signal updates with empty SignUp → **FLASH** of default form
9. Navigation completes

### Recommended Follow-up

If the flash occurs **before** `finalize()` is called (as the issue title suggests), then either:
1. There's a race condition where `previousResource` is unexpectedly `null`
2. Something in Next.js is causing a full page reload that resets module-level signals

Consider adding debug logging to capture:
- When `shouldIgnoreNullUpdate` is called
- What `previousResource` and `canBeDiscarded` values are at that moment
- The exact timing of the flash in relation to these events

## Test plan

- [x] `pnpm test` passes for new tests
- [x] Tests cover the scenarios described in USER-4372

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for sign-up and sign-in state flows, covering null vs non-null updates, behavior after finalize-like flows and resets, resource replacement, rapid successive updates, same-instance updates, event signaling, and test isolation/restoration.
* **Chores**
  * Added a small changeset file for release bookkeeping.


packages/clerk-js/src/core/__tests__/state.test.ts
```diff
@@ -0,0 +1,266 @@
...
+        expect(mockClient.resetSignUp).toHaveBeenCalled();
+
+        // Assert: Verify canBeDiscarded was set to true on the original SignUp
+        expect(existingSignUp.__internal_future.canBeDiscarded).toBe(true);
+
+        // Assert: Verify the signal was updated with a new SignUp that has no id
+        // The previous id 'signup_123' should be gone
+        expect(signUpResourceSignal().resource).toBe(newSignUpFromReset);
+        expect(signUpResourceSignal().resource?.id).toBeUndefined();
+        expect(signUpResourceSignal().resource).not.toBe(existingSignUp);
```
dstaley: don't think we need this since the assertion on 124 would pass

packages/clerk-js/src/core/__tests__/state.test.ts
```diff
@@ -0,0 +1,266 @@
...
+
+        // Assert: Signal should NOT be updated - should still have the existing SignIn
+        expect(signInResourceSignal().resource).toBe(existingSignIn);
+        expect(signInResourceSignal().resource?.id).toBe('signin_123');
+      });
+    });
+  });
+
+  describe('Simulated client refresh scenarios', () => {
+    it('should protect against null update during email verification (before finalize)', () => {
```
dstaley: isn't this the exact same test as the one on line 48?

packages/clerk-js/src/core/__tests__/state.test.ts
jacekradko: Agreed

PR #7754, opened 2026-02-03
fix(shared): Throw error on unknown scopes

dstaley:
## Description

This PR updates our parsing of the features claim and related functions to throw an error when provided with an invalid scope.

Previously, the following check would return `true` for the `fea` claim of `u:featurename`:
```
has({ feature: 'unknownscope:featurename' })
```

This is because the previous implementation would simply drop any unknown scope. This became confusing because the following would also return `true`:
```
has({ feature: 'organization:featurename' })
```

This is because `organization` is not the scope for an organization feature; it's `org`.

With the new parser, you're able to use any of the following scopes: `u, user, o, org, organization`.

This PR also optimizes the `splitScope` function to be a bit more performant since it was doing lots of redundant maps and filters.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Breaking Changes**
  * Major version bump for the shared library.
  * Feature and plan checks now strictly validate explicit scopes and will throw on invalid/unknown scopes.

* **New Features**
  * Explicit org/user scope support for feature resolution and clearer parsing behavior.

* **Tests**
  * Added authorization tests covering scope parsing, error cases, and feature resolution.

* **Chores**
  * Minor bundle size threshold updated.

### DIFF
--- a/packages/shared/src/__tests__/authorization.spec.ts
+++ b/packages/shared/src/__tests__/authorization.spec.ts
@@ -1,4 +1,5 @@
-import { describe, it, expect } from 'vitest';
+import { describe, expect, it } from 'vitest';
+
 import { createCheckAuthorization, splitByScope } from '../authorization';
 
 describe('createCheckAuthorization', () => {


dstaley:
!allow-major

packages/shared/src/authorization.ts
```diff
@@ -127,13 +139,29 @@ const checkBillingAuthorization: CheckBillingAuthorization = (params, options) =
 };
 
 const splitByScope = (fea: string | null | undefined) => {
```
jacekradko, line 141: This got flagged by Codex

splitByScope edge case: missing :

In the new loop, it does:
	•	const colonIndex = part.indexOf(':')
	•	then slices using that index.  ￼

If a segment ever lacks : (or is malformed), colonIndex becomes -1, and the slicing behavior gets weird (scope becomes part.slice(0, -1) and value becomes the entire string). Today the features claim may always be well-formed, but this helper is the kind of thing that gets reused and then surprises someone later.

Suggestion: guard early:
	•	if colonIndex <= 0 (or === -1) → continue (or treat as invalid and throw, depending on how strict you want claims parsing to be).
	•	also consider trimming out empty value.

jacekradko (approved):

packages/shared/src/authorization.ts
dstaley: Added a check! These are from FAPI so they're likely always well-formed, but it doesn't hurt to at least throw an error.

PR #7759, opened 2026-02-04
feat(nextjs,react,shared): Support suspenseful initialState

Ephem:
## Description

This PR adds internal support for `initialState: Promise<InitialState>` and uses that in the Next SDK when `<ClerkProvider dynamic>`.

This avoids having to await it in the top level provider, instead moving that down to where the `initialState` is actually being consumed, the hooks. This is done via the `React.use` hook, which is only supported in React 19, which all Next versions core3 supports already relies on.

Given that `initialState` currently only comes from a header, this might not seem like a consequential change, but it helps with two things:

* For improved `cacheComponents` support, we want to move everything dynamic out of the `ClerkProvider` down to where things are actually used. This would also avoid users having to wrap the `<ClerkProvider>` in a `<Suspense>` boundary, instead being able to place those further down.
  * This PR does not enable that, it is only a step on the way
* In the future, we might want to support other types of `initialState`, and not blocking at the top level will be beneficial then

## Summary by CodeRabbit

* **New Features**
  * Provider now accepts promise-based initial state, allowing asynchronous state initialization in apps running on React 19+.

* **Improvements**
  * Added runtime validation to block promise-based initial state on unsupported React versions and surface clearer error messages.

* **Chores**
  * Version bumps for related packages.

### DIFF
--- a/.changeset/silly-peaches-march.md
+++ b/.changeset/silly-peaches-march.md
@@ -1,0 +1,5 @@
+---
+'@clerk/nextjs': patch
+'@clerk/shared': patch
+'@clerk/react': patch
+---


.changeset/silly-peaches-march.md
Ephem, line 1: This PR in itself should not change the public behavior since we are still awaiting another header (nonce) in the provider. When we fix that, it will sometimes move where things suspend which is worthy of a proper changeset.

Ephem:
@jacekradko Thanks! I chose not to include that here since the public behavior didn't change as part of this PR. I could have included some unit tests but not sure those are very valuable here so kind of want to figure out the whole thing first.

I went ahead and created a ticket to not forget. 👍

PR #7772, opened 2026-02-05
feat(shared,react,ui): Remove useXContext hooks

Ephem:
## Description

The changeset describes it well:

```
Remove `useUserContext`, `useOrganizationContext`, `useSessionContext` and `useClientContext` from the `shared/react` package.

These hooks have never been meant for public use and have been replaced with internal hooks that do not rely on context.

If you need access to these resources, use the `useUser`, `useOrganization` and `useSession` hooks instead.

If you are building a React SDK and need direct access to the `client`, get in touch with us to discuss!
```

These are now `useUserBase`, `useOrganizationBase`, `useSessionBase` and `useClientBase` and are exported from `shared/react` with the `__internal` prefix.

Except for `client`, you can get the same resources from the already public hooks so I saw no use in having these exposed as part of the public API surface even for the shared package.

I don't _think_ people should be using `client` directly, so I choose not to expose that, but happy for feedback on that.

Using `useClerk().client` is available, but that one is not reactive and also bypasses `__internal_lastEmittedResources` so it's not a good choice, so if people do need it I don't mind exposing the base hook since it's just the shared package anyway.

### DIFF
--- a/.changeset/forty-months-appear.md
+++ b/.changeset/forty-months-appear.md
@@ -1,0 +1,13 @@
+---
+'@clerk/shared': major
+'@clerk/react': patch
+'@clerk/ui': patch
+---
+
+Remove `useUserContext`, `useOrganizationContext`, `useSessionContext` and `useClientContext` from the `shared/react` package.
+
+These hooks have never been meant for public use and have been replaced with internal hooks that do not rely on context.
+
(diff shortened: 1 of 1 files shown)


Ephem:
!allow-major

bratsos (approved):

PR #7794, opened 2026-02-06
feat(react-router): Add support for keyless mode

wobsoriano:
## Description

Brings keyless mode support to `@clerk/react-router`, removing the need for API keys during development setup.

## Changes

- Added a shared keyless resolution logic from framework packages (`resolveKeysWithKeylessFallback.ts`). SDKs will now write a 10-line wrapper to resolve Clerk keys, falling back to keyless mode in development if configured keys are missing.
- Extracted reusable test utilities for keyless testing (`keylessHelpers.ts`)

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Keyless quickstart for React Router: dev-only flow to claim apps and obtain keys, with improved server-side support and environment-aware toggle to enable keyless mode.
  * File-based key storage and runtime checks for non-Node environments; keyless URLs now flow through client/server state for UI and redirects.

* **Tests**
  * Added and consolidated Playwright integration tests and shared helpers covering keyless UI, claiming flow, key copying, and dev-server behaviors.

### DIFF
--- a/packages/shared/src/keyless/index.ts
+++ b/packages/shared/src/keyless/index.ts
@@ -10,8 +10,9 @@ export { createNodeFileStorage } from './nodeFileStorage';
 export type { FileSystemAdapter, NodeFileStorageOptions, PathAdapter } from './nodeFileStorage';
 
 export { createKeylessService } from './service';
-export type { KeylessAPI, KeylessResult, KeylessService, KeylessServiceOptions, KeylessStorage } from './service';
+export type { KeylessAPI, KeylessService, KeylessServiceOptions, KeylessStorage } from './service';
 
 export { resolveKeysWithKeylessFallback } from './resolveKeysWithKeylessFallback';
+export type { KeylessResult } from './resolveKeysWithKeylessFallback';
 
 export type { AccountlessApplication, PublicKeylessApplication } from './types';


packages/react-router/src/server/keyless/fileStorage.ts
```diff
@@ -0,0 +1,29 @@
...
+      cwd,
+      frameworkPackageName: '@clerk/react-router',
+    });
+  } catch {
+    throw new Error(
+      'Keyless mode requires a Node.js runtime with file system access. ' +
+        'Set VITE_CLERK_KEYLESS_DISABLED=1 to disable keyless mode.',
+    );
+  }
+}
```
wobsoriano, line 29: This can be extracted as well, but will plan to do so while implementing keyless for other SDKs. 

Having these node imports in the shared `@clerk/shared/keyless` barrel export would break Next.js at edge runtime, since `packages/nextjs/src/server/keyless-node.ts` imports from that barrel (correct me if Im wrong!)

packages/shared/src/keyless/resolveKeysWithKeylessFallback.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Resolves Clerk keys, falling back to keyless mode in development if configured keys are missing.
+ *
+ * @param configuredPublishableKey - The publishable key from options or environment
+ * @param configuredSecretKey - The secret key from options or environment
+ * @param keylessService - The keyless service instance (or null if unavailable)
+ * @param canUseKeyless - Whether keyless mode is enabled in the current environment
+ * @returns The resolved keys (either configured or from keyless mode)
+ */
+export async function resolveKeysWithKeylessFallback(
```
brkalow, line 21: Consider making this a method on the keyless service itself, instead of passing it in as an argument

wobsoriano: Updated it to be a method on the keyless service instead of a standalone function 👍🏼 

Also, we want keyless to work on Node.js runtime only, right?

brkalow: @wobsoriano eventually I'd like to see keyless work in the browser for SPAs, so I don't want to couple this to Node.

wobsoriano: gotcha, the current implementation should be agnostic then.

PR #7815, opened 2026-02-10
fix(clerk-js): Suppress intermediate emissions from updateClient during setActive

jacekradko:
## Summary

- Suppress intermediate state emissions from `updateClient` when called during `setActive` (e.g. from `touch()`'s piggybacked client response)
- Prevents React components from seeing the new orgId before transitive state (`undefined`) is set during org switches
- Uses the existing `__internal_setActiveInProgress` flag to gate `#emit()` in `updateClient`

## Root Cause

During `setActive`, `touch()` triggers `updateClient()` via `BaseResource._baseFetch` client piggybacking, which calls `#emit()` before `#setTransitiveState()` runs. With the migration from `useState`+`addListener` to `useSyncExternalStore`, each emission causes a synchronous re-render (sync-priority lane updates to prevent tearing). This exposed the new orgId to components before the transitive state was set, causing flickering during org switches.

Emission sequence **before fix**: `updateClient#emit(newOrg)` → `setTransitiveState#emit(undefined)` → `#updateAccessors#emit(newOrg)`
Emission sequence **after fix**: `setTransitiveState#emit(undefined)` → `#updateAccessors#emit(newOrg)`

## Test plan

- [x] Added test: verifies no intermediate orgId emission before transitive state when navigating
- [x] Added test: verifies single emission (no duplicate) when switching orgs without navigation
- [x] Both tests fail without the fix, pass with it
- [x] All existing clerk-js tests pass (99 tests)
- [x] `pnpm build` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented premature/undefined organization state emissions during active organization switches, avoiding transient incorrect state being observed by listeners and preventing unnecessary intermediate updates.

* **Tests**
  * Added tests verifying organization-switching no longer emits intermediate undefined states and that listeners only receive the final, stable organization state.

PR #7823, opened 2026-02-11
fix(clerk-js): Fix broken transitive state for org switch and sign out

Ephem:
## Description

This PR fixes a bug that was introduced when we migrated to `useSyncExternalStore`. The bug is that when switching orgs and navigating to a new page, the _current_ page re-rendered with the _new_ organization before the navigation, which can cause problems.

Alternative solution: https://github.com/clerk/javascript/pull/7815

Compared to that solution, this touches more code, but is scoped to only block `updateClient` from emitting for the specific `setActive` call that triggered it, it does not block all `updateClient` calls to keep the splash radius as small as possible.

We want to refactor this more heavily to have better guarantees, but now is not the time.

I went with a version where a `__internal_touch()` does not call `updateClient` and instead returns a `client` resource, expecting the caller to then call `updateClient` with the result. `updateClient` now takes an option to skip emitting.

If we want to refactor further in the future, this approach let's us move the `updateClient` call around, and also do things with the intermediary `client` if necessary.

Another version would be to thread a `skipUiEmit` option all the way through the base resource calls, but in total that's 6 stacks deep and couples UI concerns to resources in a way I did not like.

UPDATE

I found the same bug applied for signing out. I added two tests for this, single session and multi session.

The issue for signOut was that the session removal ran before setTransitiveState, which like the above bug triggered an updateClient which emitted.

Since this flow was simpler I opted for a simpler fix, to move the setTransitiveState to before the session removal. When updateClient is called, the if (this.session) guard detects there is no current session and skips the updateAccessors call. It still emits, but with the transitive state. Later when updateAccessors is called during the signOut, this converts the undefined to null and we are now in a correctly signed out case.

I have verified these fixesthis fix works in the dashboard.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Added organization switcher, user switcher, and sign-out pages/layouts including an "Emission log" view to observe navigation/state transitions.

* **Bug Fixes**
  * Improved auth state handling to defer or skip premature emissions during sign-in/out, user/org switches, and navigation for more consistent transitive state.

* **Tests**
  * Added integration tests for transitive state emissions and unit tests for internal touch/update flows.

* **Chores**
  * Added a changeset for a patch release and updated bundle size threshold.


integration/tests/transitive-state.test.ts
```diff
@@ -0,0 +1,262 @@
...
+
+import { appConfigs } from '../presets';
+import type { FakeOrganization, FakeUser } from '../testUtils';
+import { createTestUtils, testAgainstRunningApps } from '../testUtils';
+
+/*
+  These tests verify that useAuth emits the correct transitive state sequence when switching
+  auth context (org or user) with navigation. The expected pattern is:
+  Path A - Value A, Path A - undefined, Path B - undefined, Path B - Value B
+*/
```
brkalow, line 13: 👏

packages/clerk-js/src/core/resources/Base.ts
```diff
@@ -162,16 +164,20 @@ export abstract class BaseResource {
     return null;
   }
 
-  protected static _updateClient<J>(responseJSON: FapiResponseJSON<J> | null): void {
+  protected static _getClientResourceFromPayload<J>(
```
brkalow: nit: Consider moving this to `Client`, or making it a standalone helper.

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -104,6 +105,32 @@ export class Session extends BaseResource implements SessionResource {
...
+        body: { active_organization_id: this.lastActiveOrganizationId } as any,
+      },
+      { skipUpdateClient: true },
+    );
+
+    // Update session in-place from response (same as _baseMutate)
+    this.fromJSON((json?.response || json) as SessionJSON);
+
+    return BaseResource._getClientResourceFromPayload(json);
+  };
```
brkalow, line 140: nit: Consider consolidating the fetch across `touch()` and `__internal_touch`

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1550,13 +1554,53 @@ export class Clerk implements ClerkInterface {
...
+
+        Because we want to minimize behavioral changes until we can tackle this properly,
+        this was refactored so that updateClient does not emit under these circumstances.
+      */
       if (inActiveBrowserTab() || !this.#options.standardBrowser) {
-        await this.#touchCurrentSession(newSession);
-        // reload session from updated client
-        newSession = this.#getSessionFromClient(newSession?.id);
+        let updatedClient: ClientResource | undefined;
+        if (shouldNavigate && newSession && '__internal_touch' in newSession) {
```
jacekradko: [nit] Curious why the duck-type check and not just a `newSession typeof Session` here

Ephem: Oh right, I forgot to go back on this. Tests were failing because they mocked Session without the `__internal_touch` function and I just ran with it while testing things out. I don't like tests affecting real code though so meant to go back and implement that mock instead but never did. Will fix!

jacekradko (approved):
Nice work. I don't think there is anything blocking here, just some nits

packages/clerk-js/src/core/resources/Base.ts
Ephem: Good idea, done!

packages/clerk-js/src/core/resources/Session.ts
Ephem: I tried this at first and didn't like it. Then I tried again, and didn't like it. Third time was the charm though, I think the version I committed turned out alright.

packages/clerk-js/src/core/clerk.ts
Ephem: Done! Thanks for catching this. 🙏

PR #7840, opened 2026-02-13
fix(ui): sync BaseRouter state on pushState/replaceState

brkalow:
## Summary
- **Problem**: After popup OAuth completes, the parent page's URL changes via `clerk.navigate()`, which ultimately calls the host application's router (e.g., to `/sso-callback` then the redirect URL), but `BaseRouter` only listened for `popstate` events — which only fire on back/forward navigation, not programmatic `pushState`/`replaceState`. This left the router's internal state out of sync with the actual URL.
- **Fix**: Replace `useWindowEventListener` with a new `useHistoryChangeObserver` hook that uses the Navigation API (`currententrychange`) when available, falling back to monkey-patching `history.pushState`/`replaceState`. `PathRouter` now listens for `pushstate`, `replacestate`, and `popstate` events.
- **Safety**: Uses global subscription sets so multiple `BaseRouter` instances don't conflict when patching history methods. Navigation API detection is evaluated at effect-time (not module-load) for SSR safety. Notifications are deferred via microtask to prevent re-entrancy.
- Removes unused `useWindowEventListener` hook.
- Adds integration test for popup OAuth with path-based routing.

## Test plan
- [ ] Existing integration tests pass (OAuth redirect flows, hash-based routing)
- [ ] New `oauth popup with path-based routing` integration test validates popup OAuth completes and parent navigates to redirect URL
- [ ] Manual: open `<SignIn oauthFlow="popup" />` with path-based routing, complete OAuth in popup, verify parent page lands on the redirect URL

## Summary by CodeRabbit

* **New Features**
  * Added a popup-based sign-in page and an in-page "Sign in" button that opens the OAuth popup.

* **Bug Fixes**
  * Improved router history handling so post-popup OAuth redirects reliably sync with the app.

* **Tests**
  * Added end-to-end tests covering the OAuth popup sign-in flow for React/Vite.

* **Chores**
  * Removed an unused UI hook export and updated router history observation to support the popup flow.

### DIFF
--- a/.changeset/warm-doors-type.md
+++ b/.changeset/warm-doors-type.md
@@ -1,0 +1,5 @@
+---
+"@clerk/shared": patch
+---
+
+Add missing `username` property to `PublicUserData` interface
--- a/packages/shared/src/types/session.ts
+++ b/packages/shared/src/types/session.ts
@@ -326,6 +326,7 @@ export interface PublicUserData {
   hasImage: boolean;
   identifier: string;
   userId?: string;
+  username?: string;
 }
 
 /**


packages/ui/src/router/BaseRouter.tsx
```diff
@@ -5,19 +5,135 @@ import type { NavigateOptions } from '@clerk/shared/types';
...
+
+    if (hasNavigationAPI) {
+      const nav = (window as any).navigation;
+      const allowedTypes = new Set(navigationTypes);
+      const handler = (e: { navigationType: NavigationType }) => {
+        if (allowedTypes.has(e.navigationType)) {
+          Promise.resolve().then(notify);
+        }
+      };
+      nav.addEventListener('currententrychange', handler);
```
brkalow, line 96: Uses the new baseline `currententrychange` listener if available

packages/ui/src/router/BaseRouter.tsx
```diff
@@ -88,7 +204,23 @@ export const BaseRouter = ({
     }
   }, [currentPath, currentQueryString, getPath, getQueryString]);
 
-  useWindowEventListener(refreshEvents, refresh);
+  // Suppresses the history observer during baseNavigate's internal navigation.
+  // Without this, the observer's microtask triggers a render before setActive's
+  // #updateAccessors sets clerk.session, causing task guards to see stale state.
```
brkalow, line 209: `setActive` strikes again 😈

packages/ui/src/router/PathRouter.tsx
```diff
@@ -59,7 +59,7 @@ export const PathRouter = ({ basePath, preservedParams, children }: PathRouterPr
       getPath={getPath}
       getQueryString={getQueryString}
       internalNavigate={internalNavigate}
-      refreshEvents={['popstate']}
+      refreshEvents={['pushstate', 'replacestate', 'popstate']}
```
brkalow, line 62: We already had the refresh mechanism included in the router, this expands it to support `pushstate` and `repalcestate`

integration/tests/oauth-flows.test.ts
```diff
@@ -181,6 +181,64 @@ testAgainstRunningApps({ withEnv: [appConfigs.envs.withEmailCodes] })('oauth flo
...
+
+      // Complete OAuth in the popup
+      await popupUtils.po.signIn.setIdentifier(fakeUser.email);
+      await popupUtils.po.signIn.continue();
+      await popupUtils.po.signIn.enterTestOtpCode();
+
+      // Because the user is new to the app and legal consent is required,
+      // the sign-up can't complete in the popup. The popup sends return_url
+      // back to the parent, which navigates to /sso-callback via pushState.
+      // This exercises BaseRouter's history observation with path-based routing.
```
brkalow:
```suggestion
```

packages/ui/src/router/BaseRouter.tsx
```diff
@@ -118,6 +250,7 @@ export const BaseRouter = ({
 
       toURL.search = stringifyQueryParams(toQueryParams);
     }
+    isNavigatingRef.current = true;
```
nikosdouvlis, line 253: If `internalNavigate` or `flushSync` throws, `isNavigatingRef` stays `true` forever and the router permanently stops observing history changes. Should we wrap this in try/finally?

```suggestion
    isNavigatingRef.current = true;
    try {
      const internalNavRes = await internalNavigate(toURL, { metadata: { navigationType: 'internal' } });
      // We need to flushSync to guarantee the re-render happens before handing things back to the caller,
      // otherwise setActive might emit, and children re-render with the old navigation state.
      // An alternative solution here could be to return a deferred promise, set that to state together
      // with the routeParts and resolve it in an effect. That way we could avoid the flushSync performance penalty.
      flushSync(() => {
        setRouteParts({ path: toURL.pathname, queryString: toURL.search });
      });
      return internalNavRes;
    } finally {
      isNavigatingRef.current = false;
    }
```

brkalow: Addressed!

brkalow:
@jacekradko fair point! I didn't feel like unit tests would provide much value here, as we'd end up with a lot of mocking and mostly ceremonious assertions. Existing integration tests and the new one cover what we need.

PR #7848, opened 2026-02-14
feat(ui): lazy-load phone country code data

jacekradko:
## Description

Split the 12KB `countryCodeData.ts` module into its own async chunk, loaded only when PhoneInput is rendered. This removes static data (~245 countries) from the `ui-common` chunk, reducing its size and improving initial load performance for applications that don't use phone functionality.

### Changes
- Created `countryCodeDataLoader.ts` with a dynamic import and sync cache pattern
- Updated `phoneUtils.ts`, `useFormattedPhoneNumber.ts`, and `PhoneInput` to use lazy getters with US fallback
- Added rspack cache group configuration to isolate the phone country data chunk
- Updated 8 test files to preload data in `beforeAll`

### Verification
- `pnpm build` passes; `phone-country-data` chunk created (~8.7KB raw, 3.8KB gzip, within 10KB budget)
- `pnpm test` passes (1548 tests across 109 files)
- Country data completely removed from `ui-common`

## Checklist

- [x] (If applicable) JSDoc comments have been added or updated for any package exports

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Release Notes

* **Chores**
  * Optimized bundle size by implementing lazy loading for country code data used in phone input features.
  * Updated test infrastructure to ensure country code data is properly initialized before running tests.
  * Improved resource management through separate bundling and caching of phone-country-data.


packages/ui/src/utils/phoneUtils.ts
```diff
@@ -19,6 +25,7 @@ export function getFlagEmojiFromCountryIso(iso: CountryIso, fallbackIso = 'us'):
 }
 
 export function getCountryIsoFromFormattedNumber(formattedNumber: string, fallbackIso = 'us'): string {
+  void loadCountryCodeData();
```
nikosdouvlis, line 28: I think the `void loadCountryCodeData()` here is going to silently break non-US phone numbers for callers outside of `PhoneInput`. Components like `FormattedPhoneNumber`, `IdentityPreview`, and `PhoneSection` call these functions directly, so on a cold call the maps are still `undefined` and everything falls back to US. Nothing triggers a re-render after the data loads either since its just module-level variables, not state.

`IdentityPreview` is especially tricky because it caches the result in a `useRef`, so it never corrects itself.

Should these functions await the load, or should there be a top-level preload somewhere early?

packages/ui/src/elements/PhoneInput/countryCodeDataLoader.ts
```diff
@@ -0,0 +1,51 @@
...
+
+// Module-level cache
+let isoToCountryMap: IsoToCountryMapType | undefined;
+let codeToCountriesMap: CodeToCountryIsoMapType | undefined;
+let subAreaCodeSets: { us: ReadonlySet<string>; ca: ReadonlySet<string> } | undefined;
+let loadPromise: Promise<void> | undefined;
+
+export function loadCountryCodeData(): Promise<void> {
+  if (!loadPromise) {
+    loadPromise = import(/* webpackChunkName: "phone-country-data" */ './countryCodeData').then(mod => {
```
nikosdouvlis, line 25: If the dynamic import fails, `loadPromise` caches the rejected promise permanently and PhoneInput renders null forever with no retry. Should we clear it on error?

```suggestion
export function loadCountryCodeData(): Promise<void> {
  if (!loadPromise) {
    loadPromise = import(/* webpackChunkName: "phone-country-data" */ './countryCodeData').then(mod => {
      isoToCountryMap = mod.IsoToCountryMap;
      codeToCountriesMap = mod.CodeToCountriesMap;
      subAreaCodeSets = mod.SubAreaCodeSets;
    }).catch(err => {
      loadPromise = undefined;
      throw err;
    });
  }
  return loadPromise;
}
```

PR #7857, opened 2026-02-16
fix(nextjs): prevent infinite handshake redirect loop on Netlify

mauricioabreu:
## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Release Notes

* **Bug Fixes**
  * Fixed infinite handshake redirect loop when deploying Next.js apps with Clerk development instances to Netlify
  * Improved redirect handling with fallback mechanism for Netlify deployments in development environments


packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -222,7 +223,13 @@ export const clerkMiddleware = ((...args: unknown[]): NextMiddleware | NextMiddl
...
       const locationHeader = requestState.headers.get(constants.Headers.Location);
       if (locationHeader) {
-        const res = NextResponse.redirect(locationHeader);
+        handleNetlifyCacheInDevInstance({
+          locationHeader,
+          requestStateHeaders: requestState.headers,
+          publishableKey: requestState.publishableKey,
+        });
+
+        const res = NextResponse.redirect(requestState.headers.get(constants.Headers.Location) || locationHeader);
```
wobsoriano, line 232: We can just use locationHeader here 👀

wobsoriano (changes requested):
This is looking good. Just a minor change! Thankss

packages/nextjs/src/server/clerkMiddleware.ts
mauricioabreu: I don't think so? this code looks a bit different from the other frameworks

I read the other framework's implementation, and the response is built with R`esponse(null, { status: 307, headers: requestState.headers })`

In Next.js, we use `NextResponse.redirect(url)` and explicitly skip copying the Location header.

I mean, if we reused the locationHeader value captured before the mutation, the param added by `handleNetlifyCacheInDevInstance` is lost

wobsoriano: Copy. Looks like I have to revisit the helper. Thanks!

wobsoriano: thanks for correcting me here! I just checked and saw we're updating the Location header so your implementation is perfect!

wobsoriano (approved):
Thank you! Let's also cherry-pick the changes here to "release/core-2" branch 🫡

PR #7858, opened 2026-02-17
chore(repo): replace verdaccio scripts with pkglab

nikosdouvlis:
CI integration tests relied on a custom Verdaccio composite action that required manually managing dist-tags, snapshot versions, and npm config workarounds. pkglab wraps all of this into simple `pkglab pub` and `pkglab add` commands.

## What changed

CI workflows now use `pkglab pub --force` and `pkglab add` instead of the Verdaccio action. The `linkPackage()` helper and `E2E_CLERK_JS_VERSION` / `E2E_CLERK_UI_VERSION` env var overrides (verdaccio-era workarounds) are replaced with an explicit `PKGLAB` marker constant. Dependencies using `PKGLAB` as their version are resolved from the local registry via `pkglab add`, while dependencies with real version strings (like `'6'` for `@clerk/nextjs` v6 in ap-flows tests) go through npm as usual.

pkglab is installed as a devDependency with `node_modules/.bin` added to `$GITHUB_PATH` so `pkglab pub` / `pkglab add` are available in all CI shell steps.

## Performance

Integration tests are ~18% faster on average (single sample, so some runner variance):

```
Test                                        Verdaccio   pkglab    Diff
ap-flows                                        138s     129s      -9s
astro                                           168s     150s     -18s
billing                                         325s     310s     -15s
cache-components (16)                            23s      33s     +10s
custom                                          170s     120s     -50s
express                                         127s     101s     -26s
generic                                         326s     284s     -42s
handshake                                       138s     117s     -21s
handshake:staging                               126s      98s     -28s
localhost                                       182s     143s     -39s
machine                                         241s     278s     +37s
nextjs (15)                                     419s     373s     -46s
nextjs (16)                                     569s     288s    -281s
nuxt                                            139s     124s     -15s
quickstart (15)                                 192s     147s     -45s
quickstart (16)                                 168s     137s     -31s
react-router                                    152s     124s     -28s
sessions                                        200s     157s     -43s
sessions:staging                                190s     156s     -34s
tanstack-react-start                            125s     115s     -10s
vue                                             126s      99s     -27s
────────────────────────────────────────────────────────────────────
Total                                          4244s    3483s    -761s
Average per job                                 202s     166s     -36s
```

## Removed

- `.github/actions/verdaccio/action.yml` (128-line composite action)
- `.github/workflows/preview.yml` (unused comment-triggered workflow)
- `scripts/install-site-in-isolation.mjs` (secco-based script, only consumer of preview workflow)
- `scripts/local-registry.sh` (thin wrapper around verdaccio)
- `verdaccio.install.yaml` and `verdaccio.publish.yaml`
- `verdaccio` devDependency
- `release:verdaccio` and `local:registry:*` root scripts

## Summary by CodeRabbit

* **Refactor**
  * Replaced Verdaccio with pkglab for local package registry workflows.

* **Documentation**
  * Updated integration setup and test instructions to use pkglab for publishing and installing local packages.

* **Chores**
  * Removed legacy local registry scripts/configs and preview deployment workflow.
  * Added pkglab installation steps to CI/init flows and bumped pkglab dependency while removing Verdaccio.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -383,7 +383,7 @@ importers:
         specifier: workspace:^
         version: link:../ui
       astro:
-        specifier: ^5.15.9
+        specifier: ^5.17.1
         version: 5.17.1(@types/node@22.19.0)(db0@0.3.4)(idb-keyval@6.2.1)(ioredis@5.8.2)(jiti@2.6.1)(lightningcss@1.30.2)(rollup@4.53.1)(terser@5.44.1)(tsx@4.20.6)(typescript@5.8.3)(yaml@2.8.1)
 
   packages/backend:


pnpm-workspace.yaml
```diff
@@ -35,3 +35,4 @@ minimumReleaseAge: 2880
 minimumReleaseAgeExclude:
   - '@clerk/*'
   - 'pkglab'
+  - 'pkglab-*'
```
jacekradko, line 38: @nikosdouvlis Are you sure we can trust the owner of these packages?

packages/astro/package.json
```diff
@@ -97,7 +97,7 @@
   },
   "devDependencies": {
     "@clerk/ui": "workspace:^",
-    "astro": "^5.17.1"
+    "astro": "^5.15.9"
```
jacekradko: Is this intentional?

pnpm-workspace.yaml
nikosdouvlis: Definitely looks shady, we'll need to pin exact versions to avoid issues in the future

integration/README.md
```diff
@@ -19,6 +19,14 @@ Before you begin writing tests, you should already have:
 - Access to the **Integration testing** organization on Clerk (it owns the Clerk instances used in the tests)
 - Chromium installed. Run `pnpm playwright install chromium`.
 - WebKit installed. Run `pnpm playwright install webkit`.
+- pkglab installed. Run `npm install -g pkglab`.
+
+Integration tests install Clerk packages from a local registry managed by pkglab. Before running tests:
+
+1. Publish packages: `pkglab pub` (builds and publishes to local registry)
```
Ephem: Is this correct? My understanding is that it doesn't build things automatically, just publish what is there, so we'd need a: 

1. `pnpm build`

?

integration/models/application.ts
```diff
@@ -47,7 +47,13 @@ export const application = (
...
       if (force || !nodeModulesExist) {
         const log = logger.child({ prefix: 'setup' }).info;
-        await run(scripts.setup, { cwd: appDirPath, log });
+        // Use pkglab add to install packages from the local registry
+        const pkglabDeps = config.pkglabDependencies;
+        if (pkglabDeps.length > 0) {
+          await run(`pkglab add ${pkglabDeps.join(' ')}`, { cwd: appDirPath, log });
+        } else {
+          await run(scripts.setup, { cwd: appDirPath, log });
+        }
```
Ephem, line 56: Just to verify, I guess the thinking here is that these are temporary apps so they don't need a cleanup step?

I think there's a problem though. Since these apps do not get cleaned up, pkglab maintains a reference to them and try to update them on the next publish. That's bad in itself as it's only going to grow, but it also seems like it can lead to errors:

Image

I think we need to think about how to clean up after. This is probably pretty simple in the standard case, more of a challenge when aborting tests with Ctrl+C, or when `E2E_CLEANUP=0`. 🤔

Ephem (commented):
I like it! I tried it out locally and it seems to work well except the cleanup thing I left a comment about.

integration/models/application.ts
nikosdouvlis: Hey Fredrik, thanks for taking a look at this! 
We definitely need to revisit cleanup on both the e2e and the pkglab side.

On CICD we don't care about cleanup that much as the workers are ephemeral. When running the tests locally thought, we need to improve how the e2e infra handles cleaning up

With regards to `pkglab` now, the cli is going to ask which repos to activate (aka, trigger updates for) after every `pkglab down` / `pkglab up` invocation - the intention was for the cli to keep track of the repos but only update the ones you're actively working on. You can deactivate repos manually using `pkglab repo`, but this is still annoying when actively testing locally on multi apps (which is usually the case for e2e tests). I wonder if a flag that allows you to set the maximum number of repos to update for each `pub`, starting from the most recently added repo makes sense for cases like these (newer test apps would be updated, older test apps would automatically be ignored for that pub cycle)

One obvious improvement is to get the cli to auto prune any linked repo that doesn't exist instead of throwing, so that if you delete a tmp app, pkglab will automatically remote it from its list on the next up/pub :) Ill do that now

package.json
```diff
@@ -134,7 +130,8 @@
     "json5": "2.2.3",
     "jsonwebtoken": "9.0.2",
     "lint-staged": "^14.0.1",
-    "pkglab": "^0.11.1",
+    "pkglab": "^0.15.0",
+    "pkglab-linux-x64": "^0.15.0",
```
jacekradko: I am not sure I understand why we are adding a architecture-specific dependency here. Usually this is handled in the main package (`pkglab`) with `optionalDependencies` so each platform only installs the compatible architecture package

PR #7865, opened 2026-02-17
fix(react): Fix stale SSR state during cross-tab sign-out

bratsos:
## Description

When a user signs out in one browser tab, other tabs viewing protected pages hang instead of redirecting to the sign-in page. The page remains stuck showing stale authentication data even though the session has been terminated.

**Reproduction:**
- Open two tabs, both authenticated and viewing a protected page
- Sign out in Tab 2 using UserButton
- Tab 1 should detect the sign-out and redirect to sign-in
- Bug: Tab 1 hangs showing stale session data instead of redirecting

**Root Cause**

The issue was in `@clerk/nextjs` App Router awaitable navigation:

- `RedirectToSignIn` can trigger duplicate router.push calls to the same destination while the first transition is still in flight (commonly observed in dev/StrictMode-like remount/effect replay behavior).
- The awaitable nav buffer accepted both calls and started another transition for the same URL.
- In this duplicate in-flight scenario, Next.js may dedupe/cancel the redundant action, leaving `isPending` stuck and preventing promise buffer flush.

Result: redirect flow appears hung even though auth state is already signed out.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an App Router navigation edge case: coalesces duplicate in-flight redirects to the same destination and resets pending navigation state after flush to prevent hangs.

* **Tests**
  * Added unit tests for in-flight navigation deduplication, differing destinations, and re-navigation after flush.

* **Chores**
  * Added a changeset with a patch version bump for the Next.js package.

### DIFF
--- a/.changeset/cozy-lemons-crash.md
+++ b/.changeset/cozy-lemons-crash.md
@@ -2,4 +2,4 @@
 '@clerk/nextjs': patch
 ---
 
-Fix an App Router navigation edge case where duplicate in-flight redirects to the same destination could leave Clerk's awaitable navigation pending indefinitely. We now coalesce duplicate same-destination pushes into a single in-flight transition, reset pending destination state after flush, and add regression tests for duplicate and post-flush navigation behavior.
+Fix an App Router navigation edge case where duplicate in-flight redirects to the same destination could leave Clerk's awaitable navigation pending indefinitely.


jacekradko (commented):
@bratsos I am wondering if other hooks like `useUserBase` for example have the same issue of returning stale `initialState` ?

packages/react/src/hooks/useAuthBase.tsx
```diff
@@ -46,11 +46,20 @@ export function useAuthBase(): AuthStateValue {
   const state = useSyncExternalStore(
     useCallback(callback => clerk.addListener(callback, { skipInitialEmit: true }), [clerk]),
     useCallback(() => {
-      if (!clerk.loaded || !clerk.__internal_lastEmittedResources) {
+      const loaded = clerk.loaded;
+      const hasResources = !!clerk.__internal_lastEmittedResources;
+
+      if (!loaded && !hasResources) {
```
Ephem: What happens if we have loaded, but not yet emitted? Wouldn't this lead to a brief `undefined` even when we have `initialState` to show? 🤔

What I want to know is, how does `clerk.loaded` or `__internal_lastEmittedResources` ever go back to being falsy? The `clerk.loaded` is based on `clerk.#status` and should only change during the initial load of Clerk right?

`__internal_lastEmittedResources` is only written to in `clerk.#emit`, and always as an object, so even if the individual states are `undefined`, the empty object should still evaluate to truthy, so I see no way for that to go back to being falsy either?

bratsos: I completely changed the directory, this was indeed not the issue. Please check again! 🙏

bratsos:
> @bratsos I am wondering if other hooks like `useUserBase` for example have the same issue of returning stale `initialState` ?

I completely changed the directory, this was indeed not the issue. Please check again! 🙏

.changeset/cozy-lemons-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/nextjs': patch
+---
+
+Fix an App Router navigation edge case where duplicate in-flight redirects to the same destination could leave Clerk's awaitable navigation pending indefinitely. We now coalesce duplicate same-destination pushes into a single in-flight transition, reset pending destination state after flush, and add regression tests for duplicate and post-flush navigation behavior.
```
dstaley:
```suggestion
Fix an App Router navigation edge case where duplicate in-flight redirects to the same destination could leave Clerk's awaitable navigation pending indefinitely. We now coalesce duplicate same-destination pushes into a single in-flight transition and reset pending destination state after flush.
```

.changeset/cozy-lemons-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/nextjs': patch
+---
+
+Fix an App Router navigation edge case where duplicate in-flight redirects to the same destination could leave Clerk's awaitable navigation pending indefinitely. We now coalesce duplicate same-destination pushes into a single in-flight transition and reset pending destination state after flush.
```
Ephem: NIT & subjective: Are the details of _how_ we fixed it necessary here?

```suggestion
Fix an App Router navigation edge case where duplicate in-flight redirects to the same destination could leave Clerk's awaitable navigation pending indefinitely.
```

.changeset/cozy-lemons-crash.md
bratsos: I thought I committed this change before marking this as resolved 🤔 Anyway this is removed as per @Ephem 's suggestion! https://github.com/clerk/javascript/pull/7865/changes/a850c4cb69557cee25df752a52398723e62c4744

.changeset/cozy-lemons-crash.md
bratsos: Makes sense, handled here https://github.com/clerk/javascript/pull/7865/changes/a850c4cb69557cee25df752a52398723e62c4744

PR #7875, opened 2026-02-18
fix(clerk-js): Prevent duplicate __client_uat cookies in iframe contexts

brkalow:
## Summary
- When an app runs in both an iframe and a standalone tab, `getCookieDomain()` returns `undefined` in the iframe because the eTLD+1 test cookie probe fails (third-party cookie restrictions prevent domain cookies from sticking)
- This causes `clientUat.ts` to set **host-only** cookies (no `Domain` attribute) in the iframe, while the non-iframe context sets **domain-scoped** cookies (with `Domain` attribute)
- The browser treats these as two separate cookies, creating duplicate `__client_uat` entries with conflicting values
- Fix: `getCookieDomain()` now falls back to `hostname` instead of `undefined`, so the cookie `set()` either matches the non-iframe's domain-scoped cookie (common on platforms like Replit where hostname == eTLD+1) or silently fails — both preferable to creating a conflicting host-only cookie

## Test plan
- [x] Updated existing unit test for `getCookieDomain` to reflect new fallback behavior
- [x] Manual testing: load app in iframe on Replit, verify only domain-scoped `__client_uat` cookies are created (no host-only duplicates)
- [x] Manual testing: load app in standalone tab, verify cookie behavior unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed __client_uat cookie domain handling so it behaves correctly when the app loads in iframe and non-iframe contexts.
  * Added a deterministic fallback for cookie domain resolution (returns the hostname when a broader domain can't be derived) to avoid host-only cookie conflicts in restricted environments.
  * Ensured cookie SameSite and Secure attributes are consistently applied during domain probing.

* **Tests**
  * Added/updated tests to verify attribute forwarding to the domain probe and the hostname fallback.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2475,7 +2475,7 @@ packages:
 
   '@expo/bunyan@4.0.1':
     resolution: {integrity: sha512-+Lla7nYSiHZirgK+U/uYzsLv/X+HaJienbD5AKX1UQZHYfWaP+9uuQluRB4GrEVWF0GZ7vEVp/jzaOT9k/SQlg==}
-    engines: {'0': node >=0.10.0}
+    engines: {node: '>=0.10.0'}
 
   '@expo/cli@0.22.26':
     resolution: {integrity: sha512-I689wc8Fn/AX7aUGiwrh3HnssiORMJtR2fpksX+JIe8Cj/EDleblYMSwRPd0025wrwOV9UN1KM/RuEt/QjCS3Q==}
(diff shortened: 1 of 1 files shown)


.changeset/fix-iframe-client-uat-cookie-domain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Fix `__client_uat` cookie being set on two different domain scopes when app is loaded in both iframe and non-iframe contexts. `getCookieDomain()` now falls back to `hostname` instead of `undefined` when the eTLD+1 probe fails, preventing host-only cookies that conflict with domain-scoped cookies.
```
brkalow:
```suggestion
Fix `__client_uat` cookie being set on two different domain scopes when app is loaded in both iframe and non-iframe contexts.
```

packages/clerk-js/src/core/auth/getCookieDomain.ts
```diff
@@ -40,5 +40,10 @@ export function getCookieDomain(hostname = window.location.hostname, cookieHandl
     cookieHandler.remove({ domain: eTLD });
   }
 
-  return;
+  // Fallback to hostname to ensure domain-scoped cookie rather than host-only.
+  // In restricted contexts (e.g. cross-origin iframes), the set() will silently
+  // fail — which is preferable to creating a host-only cookie that conflicts
+  // with domain-scoped cookies set by non-iframe contexts.
+  cachedETLDPlusOne = hostname;
```
jacekradko, line 59: Hmm. I think it makes sense, but I wonder if this could potentially set a cookie on a different subdomain that would still create 2 separate cookies in some cases. 🤔

jacekradko (approved):

Issue #7841, opened 2026-02-13
Astro: PUBLIC_CLERK_PUBLISHABLE_KEY cannot be changed after build


Gr3q:
### Preliminary Checks

- [x] I have reviewed the documentation: https://clerk.com/docs

- [x] I have searched for existing issues: https://github.com/clerk/javascript/issues

- [x] I have not already reached out to Clerk support via email or Discord (if you have, no need to open an issue here)

- [x] This issue is not a question, general help request, or anything other than a bug report directly related to Clerk. Please ask questions in our Discord community: https://clerk.com/discord.

### Reproduction

http://github.com/clerk/clerk-astro-quickstart

### Publishable key

pk_test_cHJvLWNvd2JpcmQtNTcuY2xlcmsuYWNjb3VudHMuZGV2JA

### Description

`PUBLIC_CLERK_PUBLISHABLE_KEY` can't be read from the runtime environment, it is baked in when the astro app is built (tbh all of the variables starting with `PUBLIC_` but I'm going to ignore those here)

**Steps to reproduce:**

1. Set up your `.env` with values `placeholder` for each variable - Optional, you can leave them empty too
2. Run `pnpm build`
3. Run `PUBLIC_CLERK_PUBLISHABLE_KEY=pk_test_cHJvLWNvd2JpcmQtNTcuY2xlcmsuYWNjb3VudHMuZGV2JA node dist/server/entry.mjs`
4. Open the page (http://localhost:4321)

**Expected behavior:**

The server doesn't throw publishable key errors.

**Actual behavior:**

It does.

### Environment

```Shell
System:
    OS: macOS 26.2
    CPU: (12) arm64 Apple M4 Pro
    Memory: 62.94 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.13.0 - /opt/homebrew/opt/node@24/bin/node
    npm: 11.6.2 - /opt/homebrew/opt/node@24/bin/npm
    pnpm: 10.28.1 - /opt/homebrew/opt/node@24/bin/pnpm
    bun: 1.3.3 - /Users/attilagreguss/.bun/bin/bun
  Browsers:
    Chrome: 144.0.7559.133
    Firefox: 147.0.2
    Safari: 26.2
  npmPackages:
    @astrojs/check: ^0.9.6 => 0.9.6 
    @astrojs/node: ^9.5.2 => 9.5.2 
    @clerk/astro: ^2.16.17 => 2.16.17 
[...]

PR #7889, opened 2026-02-19
fix(astro): Fall back to process.env for runtime environment variables

alexcarpenter:
Add process.env as a third fallback in getContextEnvVar() so that PUBLIC_CLERK_PUBLISHABLE_KEY and other env vars set at runtime (e.g., via the Astro Node adapter) are properly read when import.meta.env.PUBLIC_* is statically replaced at build time by Vite.

## Description

fixes https://github.com/clerk/javascript/issues/7841

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixes runtime reading of the publishable key when build-time env replacements aren’t available by adding a fallback to process.env.

* **Tests**
  * Adds comprehensive tests for environment variable precedence and fallback behavior across different runtime environments.

### DIFF
--- a/packages/astro/src/server/get-safe-env.ts
+++ b/packages/astro/src/server/get-safe-env.ts
@@ -14,18 +14,14 @@ function getContextEnvVar(envVarName: keyof InternalEnv, contextOrLocals: Contex
     return locals.runtime.env[envVarName];
   }
 
-  const envValue = import.meta.env[envVarName];
-  if (envValue) {
-    return envValue;
-  }
-
-  // Fallback to process.env for runtime environments (e.g., Node.js adapter)
-  // where import.meta.env.PUBLIC_* is statically replaced at build time by Vite
-  if (typeof process !== 'undefined' && process.env) {
+  // Prefer process.env for runtime environments (e.g., Node.js adapter)
+  // where import.meta.env.PUBLIC_* is statically replaced at build time by Vite.
+  // Runtime values should take precedence over build-time values.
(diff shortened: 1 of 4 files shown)


wobsoriano (approved):
thanks!

packages/astro/src/server/get-safe-env.ts
```diff
@@ -14,7 +14,18 @@ function getContextEnvVar(envVarName: keyof InternalEnv, contextOrLocals: Contex
     return locals.runtime.env[envVarName];
   }
 
-  return import.meta.env[envVarName];
+  const envValue = import.meta.env[envVarName];
+  if (envValue) {
+    return envValue;
+  }
+
+  // Fallback to process.env for runtime environments (e.g., Node.js adapter)
```
asdfjkalsdfla: What do you think of having process.env have a higher precedence than import.meta.env? 

My gut is that if someone defines this in their running environment that is what they intend vs. what they had it set to when they built an app. It would also make the runtime variables have higher priority like when deployed in Cloudflare Workers.

asdfjkalsdfla (commented):
General comments.

packages/astro/src/server/get-safe-env.ts
jacekradko: Thanks @asdfjkalsdfla . I agree that preferring runtime over build-time values is the correct approach. Updated the PR to switch the priority and marking the change as breaking.

Issue #7849, opened 2026-02-15
@clerk/astro doesn't support Astro v6 in Cloudflare - Error: Astro.locals.runtime.env has been removed


asdfjkalsdfla:
### Description
When running the @clerk/astro middleware with cloudlfare, it produces the following error -
`Error: Astro.locals.runtime.env has been removed in Astro v6. Use 'import { env } from "cloudflare:workers"' instead.`

Looks like https://github.com/clerk/javascript/blob/2627442c1c38a5211bdcd707ca19b156c6a63e11/packages/astro/src/server/get-safe-env.ts need to be updated. The calls were removed in the Cloudflare adapter (https://github.com/withastro/astro/pull/14306) and now produces an error (https://github.com/withastro/astro/pull/15075).

### Environment

```Shell
System:
    OS: macOS 26.3
    CPU: (8) arm64 Apple M3
    Memory: 84.53 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.13.1 - /Users/richardwright/.nvm/versions/node/v24.13.1/bin/node
    npm: 11.8.0 - /Users/richardwright/.nvm/versions/node/v24.13.1/bin/npm
    pnpm: 10.24.0 - /Users/richardwright/Library/pnpm/pnpm
  Browsers:
    Chrome: 144.0.7559.133
    Firefox: 147.0.3
    Safari: 26.3
  npmPackages:
    @astrojs/cloudflare: 13.0.0-beta.7 => 13.0.0-beta.7 
    @astrojs/sitemap: ^3.7.0 => 3.7.0 
    @clerk/astro: 3.0.0-snapshot.v20260213202659 => 3.0.0-snapshot.v20260213202659 
    astro: 6.0.0-beta.11 => 6.0.0-beta.11 
    wrangler: ^4.65.0 => 4.65.0
```

asdfjkalsdfla:
Potential patch. This support simple node runtime variables as well.

```
// At the top of the file, attempt to import cloudflare:workers env
// This will only succeed in Cloudflare Workers runtime with Astro v6+
let cloudflareEnv: any = undefined;

// Use a self-executing function to avoid top-level await
(async () => {
  try {
    const cf = await import('cloudflare:workers');
    cloudflareEnv = cf.env;
  } catch {
    // Not in Cloudflare Workers environment or module not available
  }
})();

/**
 * @internal
 * Isomorphic handler for reading environment variables
 * Supports:
[...]

PR #7890, opened 2026-02-19
fix(astro): Fix compatibility with Astro v6 Cloudflare adapter

alexcarpenter:
Astro v6's Cloudflare adapter removed `locals.runtime.env`, which causes `@clerk/astro` to crash. This adds a fallback that imports env from `cloudflare:workers` when `locals.runtime.env` is unavailable, while maintaining backwards compatibility with Astro v4/v5.

## Description

fixes https://github.com/clerk/javascript/issues/7849

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Adds compatibility for Astro v6 Cloudflare deployments.

* **Bug Fixes**
  * Improved environment detection and safe fallbacks to prevent runtime errors on Astro v6 + Cloudflare.

* **Tests**
  * Expanded test coverage for environment initialization, caching/memoization, and fallback precedence across Astro versions.

* **Chores**
  * Updated package compatibility to officially allow Astro v6 and added a patch release changeset.

### DIFF
--- a/packages/astro/src/server/get-safe-env.ts
+++ b/packages/astro/src/server/get-safe-env.ts
@@ -38,6 +38,13 @@ async function initCloudflareEnv(): Promise<void> {
 function getContextEnvVar(envVarName: keyof InternalEnv, contextOrLocals: ContextOrLocals): string | undefined {
   const locals = 'locals' in contextOrLocals ? contextOrLocals.locals : contextOrLocals;
 
+  // Astro v6+ Cloudflare adapter: env from cloudflare:workers
+  // Checked first to avoid the expensive try/catch on locals.runtime.env,
+  // which throws on every call in Astro v6 Cloudflare environments.
+  if (cloudflareEnv) {
+    return cloudflareEnv[envVarName];
+  }
+
   // Astro v4/v5 Cloudflare adapter: env is on locals.runtime.env
   try {
     if (locals?.runtime?.env) {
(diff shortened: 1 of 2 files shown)


wobsoriano (approved):
Thanks for catching this!

packages/astro/src/server/get-safe-env.ts
```diff
@@ -3,15 +3,53 @@ import type { APIContext } from 'astro';
...
+  try {
+    if (locals?.runtime?.env) {
+      return locals.runtime.env[envVarName];
+    }
+  } catch {
+    // Astro v6 Cloudflare adapter throws when accessing locals.runtime.env
+  }
+
+  // Astro v6 Cloudflare adapter: env from cloudflare:workers
+  if (cloudflareEnv) {
```
asdfjkalsdfla: What do you think about having this check prior to the astro v4/v5 check? Throwing and then catching an error in javascript is pretty slow. With the current path, we'd hit that on every API call. If you switch the order, v6 will avoid the slow try catch of an exception. Astro v4/v5 would would have one more check on cloudflareEnv being set, but that should be very quick.

jacekradko: Thanks for the suggestions @asdfjkalsdfla . Implemented in [e2bba51](https://github.com/clerk/javascript/pull/7890/commits/e2bba51e58518e6f434da1c2f9861421e34c16a8)

PR #7899, opened 2026-02-20
fix(js): Turnstile retry logic

tmilewski:
## Description

- Fix a crash in the Turnstile CAPTCHA retry logic where captcha.reset() was called after the widget's DOM container had already been removed, causing an unhandled error.
- Before retrying on a retriable Turnstile error code, the error-callback now checks whether the widget container still exists in the DOM. If the container has been removed (e.g., due to a component unmount or navigation), the promise is rejected immediately with the accumulated error codes instead of attempting a reset on a stale widget ID.
- Add comprehensive unit tests covering the new container guard: verifying immediate rejection when the container is removed before retry, confirming normal retry behavior when the container is still present, and validating that accumulated error codes are correctly reported.

## Test plan

- Container removed before retry fires — rejects immediately, captcha.reset is not called
- Container still exists — captcha.reset proceeds normally
- Multiple errors accumulated before container removal — all error codes included in rejection

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash that could occur with Turnstile CAPTCHA when the widget was removed from the page during a retry attempt.

### DIFF
--- a/packages/clerk-js/src/utils/captcha/turnstile.ts
+++ b/packages/clerk-js/src/utils/captcha/turnstile.ts
@@ -188,6 +188,10 @@ export const getTurnstileToken = async (opts: CaptchaOptions) => {
              */
             if (retries < 2 && shouldRetryTurnstileErrorCode(errorCode.toString())) {
               setTimeout(() => {
+                if (widgetContainerQuerySelector && !document.querySelector(widgetContainerQuerySelector)) {
+                  reject([errorCodes.join(','), id]);
+                  return;
+                }
                 captcha.reset(id as string);
                 retries++;
               }, 250);
--- a/.changeset/strict-worms-allow.md
+++ b/.changeset/strict-worms-allow.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/clerk-js/src/utils/captcha/turnstile.ts
```diff
@@ -188,6 +188,10 @@ export const getTurnstileToken = async (opts: CaptchaOptions) => {
              */
             if (retries < 2 && shouldRetryTurnstileErrorCode(errorCode.toString())) {
               setTimeout(() => {
+                if (widgetContainerQuerySelector && !document.querySelector(widgetContainerQuerySelector)) {
```
nikosdouvlis, line 191: The `widgetContainerQuerySelector &&` guard is technically redundant here since it's always assigned before `handleCaptchaTokenGeneration` is called (lines 98, 124, or 142 guarantee it)

jacekradko:
@tmilewski Should this be part of core 2?

tmilewski:
> @tmilewski Should this be part of core 2?

Yeah, I'll backport

PR #7912, opened 2026-02-23
fix(clerk-js): Prevent session cookie removal during offline token refresh

bratsos:
## Description
When offline, two code paths in `Session.ts` emit `token:update` events with empty tokens. `AuthCookieService` interprets empty tokens as signed-out and removes the `__session cookie` — even though the session is still valid server-side. On the next page load or visibility change, the missing cookie makes the user appear signed out.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unexpected sign-outs by avoiding caching or dispatching empty tokens when offline, preserving stale tokens and adding offline-aware retry/error signaling.

* **Tests**
  * Added and updated tests for offline retries, background refresh timing, deduplication, and recovery-from-offline to ensure no incorrect token events.

* **Chores**
  * Added a changeset entry for the patch.

### DIFF
--- a/packages/clerk-js/vitest.setup.mts
+++ b/packages/clerk-js/vitest.setup.mts
@@ -269,6 +269,12 @@ vi.mock('@formkit/auto-animate/react', () => ({
   useAutoAnimate: () => [null],
 }));
 
+// Also mock the base module to prevent its side effects (setInterval/setTimeout
+// that call requestAnimationFrame) from firing after jsdom environment teardown.
+vi.mock('@formkit/auto-animate', () => ({
+  default: () => ({ enable: () => {}, disable: () => {}, destroy: () => {} }),
+}));
+
 // Mock browser-tabs-lock to prevent window access errors in tests
 vi.mock('browser-tabs-lock', () => {
   return {


packages/clerk-js/src/core/resources/__tests__/Session.test.ts
```diff
@@ -261,8 +261,6 @@ describe('Session', () => {
 
     describe('with offline browser and network failure', () => {
       beforeEach(() => {
-        // Use real timers for offline tests to avoid unhandled rejection issues with retry logic
```
jacekradko, line 264: 🎉

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -435,18 +441,28 @@ export class Session extends BaseResource implements SessionResource {
...
-      return cachedToken.getRawString() || null;
+      result = cachedToken.getRawString() || null;
+    } else {
+      result = await this.#fetchToken(template, organizationId, tokenId, shouldDispatchTokenUpdate, skipCache);
+    }
+
+    // Throw when offline and no token so retry() in getToken() can fire.
+    // Without this, _getToken returns null (success) and retry() never calls shouldRetry.
+    if (result === null && !isValidBrowserOnline()) {
+      throw new ClerkRuntimeError('Network request failed while offline', { code: 'network_error' });
```
brkalow, line 470: Recommend that we re-use `ClerkOfflineError` here

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -435,18 +441,28 @@ export class Session extends BaseResource implements SessionResource {
...
-      if (shouldDispatchTokenUpdate) {
+      // Don't emit empty token update when offline — preserve session cookie
+      if (shouldDispatchTokenUpdate && (cachedToken.getRawString() || isValidBrowserOnline())) {
         eventBus.emit(events.TokenUpdate, { token: cachedToken });
       }
-      // Return null when raw string is empty to indicate signed-out state
-      return cachedToken.getRawString() || null;
+      result = cachedToken.getRawString() || null;
+    } else {
+      result = await this.#fetchToken(template, organizationId, tokenId, shouldDispatchTokenUpdate, skipCache);
```
brkalow, line 464: does this throw if offline? should we avoid fetching if we're offline in the first place?

packages/clerk-js/src/core/resources/Session.ts
bratsos: I'm not sure about this. At this point throwing this error doesn't mean we're 100% offline, we just throw this so that `retry`'s `shouldRetry` mechanism catches the error and retries 3 times (if the browser is offline during this). After the 3rd retry, if the browser is still offline, then [we throw](https://github.com/clerk/javascript/blob/899fb968e2af1f8dbeb2cc2dfdced0983521c9e4/packages/clerk-js/src/core/resources/Session.ts#L180-L182) a `ClerkOfflineError`.

The way I see this is that when we throw this `ClerkRuntimeError` we're in a state where something went wrong with the network, but will give it some time, and we only throw `ClerkOfflineError` when there is no more attempts.

brkalow: Got it, so this is not an error that the user would see?

packages/clerk-js/src/core/resources/Session.ts
bratsos: Great question! So that sent me through a rabbit hole and tl;dr: we can do it and if we throw here when we're offline, we'll trigger the retrying mechanism without firing network requests.

Additionally because of this exploration I figured a race condition that we had where the browser could go back online while mid-request, so I added a test for it and a fix [here](https://github.com/clerk/javascript/pull/7912/changes/fd118d9b00e483d0042b9cd5e44a9eee032a629c#diff-4121946947785ef6da215af50dd60589e9cde4c1a555e3d91b91e8b98a06788bR353)

packages/clerk-js/src/core/resources/Session.ts
bratsos: Yes exactly - if it surfaces to the user it's gonna be a `ClerkOfflineError`. Unless I messed something up that is

packages/shared/src/browser.ts
```diff
@@ -73,13 +73,12 @@ export function isBrowserOnline(): boolean {
...
-  // https://developer.mozilla.org/en-US/docs/Web/API/Navigator/connection#browser_compatibility
-  // @ts-ignore
-  const isExperimentalConnectionOnline = navigator?.connection?.rtt !== 0 && navigator?.connection?.downlink !== 0;
-  return isExperimentalConnectionOnline && isNavigatorOnline;
+  // navigator.onLine is the standard API and is reliable for detecting
+  // complete disconnection (airplane mode, WiFi off, etc.).
+  // The experimental navigator.connection API (rtt/downlink) was previously
+  // used as a secondary signal, but it reports zero values in headless browsers
+  // and CI environments even when connected, causing false offline detection.
+  return !!navigator.onLine;
```
brkalow, line 81: hmm, I'm not sure we want to change this. can we look back in the history? I feel like I recall there being issues with only relying on `navigator.onLine` in the past

bratsos: I couldn't find something concrete - the main difference is that `navigator.onLine` can return `true` even when there's no actual internet connectivity (it only detects whether the device is connected to a network, for example a wifi network with no internet access).

`navigator.connection` seems like a better signal but it's more unreliable as it's not supported on Firefox or Safari at all, in headless browsers it reports `navigator.connection.rtt == 0` and I found other issues around the web as well.

This seems the better fix and we can always iterate. WDYT?

PR #7920, opened 2026-02-24
feat(clerk-js,localizations,msw,shared,ui): Add display for member limits

dstaley:
## Description

This PR adds a footer to the OrganizationProfile members page to render seat usage.

CleanShot 2026-02-24 at 13 21 58@2x

## Testing Instructions

- Checkout locally, run `pnpm i && turbo build`, then `pnpm dev:sandbox` to start the sandbox and go to it at `http://localhost:4000`
- Open the development tools console
- run `scenario.setScenario(AVAILABLE_SCENARIOS.OrgProfileSeatLimit)`
- navigate to the OrganizationProfile page
- click "Members" in the sidebar
- confirm rendered footer matches designs

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/ui/src/components/OrganizationProfile/OrganizationMembers.tsx
+++ b/packages/ui/src/components/OrganizationProfile/OrganizationMembers.tsx
@@ -2,6 +2,7 @@ import { useOrganization } from '@clerk/shared/react';
 import { useState } from 'react';
 
 import { useFetchRoles } from '@/hooks/useFetchRoles';
+import { Users } from '@/icons';
 import { Alert } from '@/ui/elements/Alert';
 import { Animated } from '@/ui/elements/Animated';
 import { Card } from '@/ui/elements/Card';
@@ -19,7 +20,6 @@ import { MembersActionsRow } from './MembersActions';
 import { MembersSearch } from './MembersSearch';
 import { OrganizationMembersTabInvitations } from './OrganizationMembersTabInvitations';
 import { OrganizationMembersTabRequests } from './OrganizationMembersTabRequests';
-import { Users } from '@/icons';
 
 export const ACTIVE_MEMBERS_PAGE_SIZE = 10;
 


packages/localizations/src/en-US.ts
```diff
@@ -454,6 +454,7 @@ export const enUS: LocalizationResource = {
     start: {
       headerTitle__general: 'General',
       headerTitle__members: 'Members',
+      membershipSeatUsageLabel: '{{count}} of {{limit}} seats taken',
```
mzhong9723: ❓ How about `seats used` to match figma?

packages/ui/src/components/OrganizationProfile/OrganizationMembers.tsx
```diff
@@ -173,6 +175,46 @@ export const OrganizationMembers = withCardStateProvider(() => {
...
+              icon={Users}
+              size='md'
+              colorScheme='neutral'
+            />
+            <Text
+              as='span'
+              colorScheme='inherit'
+              localizationKey={localizationKeys('organizationProfile.start.membershipSeatUsageLabel', {
+                count: memberships.count + (invitations?.count ?? 0),
+                limit: organizationSettings.maxAllowedMemberships,
```
mzhong9723: ❓ Is this the specific organization member limit or the instance level default max member limit for new orgs? I think we want the former.

After phase 1 work is complete, that value should be populated from the entitlements table, so should reflect what this specific org's member limit is, whether the entitlement is managed by organizations or by billing

packages/localizations/src/en-US.ts
dstaley: :nod:

packages/ui/src/components/OrganizationProfile/OrganizationMembers.tsx
dstaley: this is a really good catch! thank you :) What I had before was indeed the _environment_ `maxAllowedMemberships`. Updated to use the `organization.maxAllowedMemberships` instead.

PR #7923, opened 2026-02-24
fix(backend): auth _next/data requests when __client_uat is missing

### LINKED ISSUE #7178: authenticateRequest and organizationSyncOptions
### Preliminary Checks

### Reproduction

https://github.com/wuzzeb/clerk-auth-org

### Publishable key

pk_test_ZmFzdC1iZWV0bGUtMjQuY2xlcmsuYWNjb3VudHMuZGV2JA

### Description

High Level Picture: I have users in orgs and a user might be in more than one org.  While most users will be in a single org, a consultant user will be a member of multiple orgs and will wish to have multiple orgs open in different tabs (so not a single active org, but different orgs they are consulting with).

I want to provide pages for each org like explained in https://clerk.com/docs/guides/organizations/org-slugs-in-urls .  So I have a react single page application that on the client has a url such as `https://domain/{orgId}/projects` for example.  The client extracts the `{orgId}` from the route and then makes a websocket connection to `wss://domain/api/websocket/{orgId}`.  I now want to use authenticateRequest in the websocket handler on the server (cloudflare worker) to check if the user can access the org.  Because it is websockets, the auth has to go through the cookies since can't set a bearer token.

### Full Example Repository

I made an example github repository https://github.com/wuzzeb/clerk-auth-org I am using to test by starting with a fresh repository and adding only a little bit of code.  See https://github.com/wuzzeb/clerk-auth-org/commit/69e1582ef7673078000a0bb0ae0e0097b2ee024e for the code above the default scaffold.  But below I post some snippets and my investigation of what the backend code is doing:

### Server Code

On the server, the code looks like

```typescript
const client = createClerkClient(...);

export default {
  async fetch(req) {
    if (req.method !== "GET") {
      return new Response("Method Not Allowed", { status: 405 });
    }

    var url = new URL(req.url);

    // path must be /api/websocket/{clerkOrgId}
    if (!url.pathname.startsWith("/api/websocket/")) {
      return new Response("Not Found", { status: 404 });
    }
    if (req.headers.get("Upgrade")?.toLowerCase() !== "websocket") {
      return new Response("Expected WebSocket Upgrade", { status: 400 });
    }

    const parts = url.pathname.split("/");
    if (parts.length !== 4 || !parts[3]) {
      return new Response("Not Found", { status: 404 });
    }
    const clerkOrgId = parts[3];

    const token = await client.authenticateRequest(req, {
      organizationSyncOptions: {
        organizationPatterns: ["/api/websocket/:id"],
      },
    });

    if (!token.isAuthenticated) {
      return new Response("Unauthorized", { status: 401 });
    }

    const auth = await token.toAuth();

    if (auth.orgId !== clerkOrgId) {
      return new Response("Forbidden", { status: 403 });
    }

   // more auth.has checks, etc.
  },
}
```

### Results

* If the currently active org matches the org id in the websocket URL, the call to `authenticateRequest` works, everything matches.
* If the currently active org is a different org, I put a breakpoint there and single stepped through
    * We go to `authenticateReqeustWithTokenInCookie` https://github.com/clerk/javascript/blob/539fad7b80ed284a7add6cf8c4c45cf4c6a0a8b2/packages/backend/src/tokens/request.ts#L434
    * It gets down to calling `handleMaybeOrganizationSyncHandshake` at https://github.com/clerk/javascript/blob/539fad7b80ed284a7add6cf8c4c45cf4c6a0a8b2/packages/backend/src/tokens/request.ts#L596
    * `handleMaybeOrganizationSyncHandshake` correctly detects that a org switch is needed at line 378 https://github.com/clerk/javascript/blob/539fad7b80ed284a7add6cf8c4c45cf4c6a0a8b2/packages/backend/src/tokens/request.ts#L378
    * `handleMaybeHandshakeStatus` is called https://github.com/clerk/javascript/blob/539fad7b80ed284a7add6cf8c4c45cf4c6a0a8b2/packages/backend/src/tokens/request.ts#L399
    * The first thing `handleMaybeHandshakeStatus` does is call `isRequestEligibleForHandshake` https://github.com/clerk/javascript/blob/539fad7b80ed284a7add6cf8c4c45cf4c6a0a8b2/packages/backend/src/tokens/request.ts#L317
    * `isRequestEligibleForHandshake` checks for the Sec-Fetch-Mode header https://github.com/clerk/javascript/blob/539fad7b80ed284a7add6cf8c4c45cf4c6a0a8b2/packages/backend/src/tokens/handshake.ts#L107
    * `Sec-Fetch-Mode` is empty because this is a websocket request
    * `isRequestEligibleForHandshake` returns false, causing `handleMaybeHandshakeStatus` to return `signedOut`
    * `handleMaybeOrganizationSyncHandshake` returns null, https://github.com/clerk/javascript/blob/539fad7b80ed284a7add6cf8c4c45cf4c6a0a8b2/packages/backend/src/tokens/request.ts#L405
    * I think that comment is wrong, the comment says that returning null is only possible if we are in a redirect loop, but there have been no redirect loops yet.

Why is that sec-fetch-req check there in the first place?  I don't understand `isRequestEligibleForHandshake`.

### Workaround?

On the client, right before making the websocket connection, I can call

```typescript
await clerk.setActive({organization: orgIdFromURL });
await clerk.session?.getToken({ skipCache: true });
const ws = new WebSocket(`wss://${window.location.host}/api/websocket/`${orgIdFromURL}`);
```

This works except if you have multiple tabs open there is a small timing window where things go wrong.  I use ReconnectingWebsocket in that case and it seems to work.  If two tabs are trying to connect at the same time, one will fail because it gets the wrong org, but then it will attempt a reconnect a second later.  The constant switching of active orgs combined with reconnect seems to work.

### Environment

```Shell
System:
    OS: Linux 6.17 Arch Linux
    CPU: (16) x64 AMD Ryzen 7 7700X 8-Core Processor
    Memory: 20.08 GB / 30.47 GB
    Container: Yes
    Shell: 4.1.2 - /usr/bin/fish
  Binaries:
    Node: 25.1.0 - /usr/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 11.6.2 - /usr/bin/npm
    pnpm: 10.20.0 - /usr/bin/pnpm
  Browsers:
    Chromium: 142.0.7444.134
    Firefox: 144.0.2
    Firefox Developer Edition: 144.0.2
  npmPackages:
    @clerk/backend: ^2.20.0 => 2.20.0 
    @clerk/clerk-react: ^5.53.8 => 5.53.8 
    @cloudflare/vite-plugin: ^1.14.0 => 1.14.0 
    @eslint/js: ^9.33.0 => 9.39.1 
    @types/react: ^19.1.10 => 19.2.2 
    @types/react-dom: ^19.1.7 => 19.2.2 
    @vitejs/plugin-react: ^5.0.0 => 5.1.0 
    eslint: ^9.33.0 => 9.39.1 
    eslint-plugin-react-hooks: ^5.2.0 => 5.2.0 
    eslint-plugin-react-refresh: ^0.4.20 => 0.4.24 
    globals: ^16.3.0 => 16.5.0 
    react: ^19.1.1 => 19.2.0 
    react-dom: ^19.1.1 => 19.2.0 
    typescript: ~5.8.3 => 5.8.3 
    typescript-eslint: ^8.39.1 => 8.46.3 
    vite: ^7.1.2 => 7.2.2 
    wrangler: ^4.46.0 => 4.46.0
```

arthurgeron:
### Problem
`authenticateRequest` returns `signedOut` for `_next/data` requests (Next.js Pages Router client-side navigations) when the `__client_uat` cookie is missing but a valid `__session` token exists. The `__client_uat` cookie is not always sent with fetch requests because it is set with `SameSite=Strict` (e.g., during development). As a result, `getAuth()` returns a fully null auth object during client-side navigations, even though the user remains authenticated, and no errors are thrown.

### Fix
- In `backend/src/tokens/request.ts` on the `!hasActiveClient && hasSessionToken` assertion, `handleMaybeHandshakeStatus` is now only called when `isRequestEligibleForHandshake()` returns `true` (document/iframe navigations). For non-handshake-eligible requests (fetch, `_next/data`), the code falls through to the full token verification pipeline instead of short-circuiting to `signedOut`.

- The `iat < clientUat` freshness comparison is now guarded with `authenticateContext.clientUat &&`, so it is skipped when `clientUat` is missing, avoiding a incorrect comparison.

- Handshake-eligible requests (document navigations) are unchanged. This change only affects requests where handshake was already impossible.

### Observations

- Before this, non-handshake-eligible requests with a missing `__client_uat` were returned as signedOut without verifying `__session`. Now, those requests go through the same token verification pipeline used by all other authenticated requests. No verification steps are skipped or weakened.
> `__client_uat` cookie is a client-side freshness signal used to detect cross-tab session changes. It is not involved in token validation.
- Handshake-eligible requests (e.g., document navigations) are unaffected.
 

### Related issues

- No issue for this exact Next.js bug
- Should also fix #7178 (goes through same code path, `handleMaybeHandshakeStatus` fails to `signedOut`)

- Create a Next.js (tested on v14) Pages Router app with `@clerk/nextjs`
- Add a page with getServerSideProps that logs getAuth(req):
```JavaScript
   export const getServerSideProps = (ctx) => {
     const auth = getAuth(ctx.req);
     console.log('auth.userId:', auth.userId);
     return { props: { userId: auth.userId } };
   };
```
- Sign in via Clerk
- Navigate to the page via full page load (i.e., URL bar), `userId` should be present
- Navigate to the page via client-side `<Link>`
  - before: `userId` is null
  - after: `userId` is present

## Checklist

- [ ] ~~(If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports~~
- [ ] ~~(If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated~~

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * getAuth() no longer reports unauthenticated during Next.js Pages Router client-side navigations when a valid session exists but certain client metadata is missing.
  * Prevented unnecessary background handshake attempts when no active client is present.
  * Avoided incorrect token time comparisons when optional client metadata is absent.

* **Tests**
  * Added tests covering non-document requests to ensure stable authentication behavior.


arthurgeron:
There's already no JSDocs for these methods, only a few in that module have it. Should I create docs for it anyway?

cc @nikosdouvlis

Ephem:
Thank you for this contribution! 🙏

Soon after you opened this PR, we released [Core 3](https://clerk.com/changelog/2026-03-03-core-3), our latest major for the SDKs. It would be helpful if you could check and see if this error still reproduces with the latest versions? Note that we have dropped support for Next 14 in this latest version, I hope that doesn't make it too cumbersome.

Not saying we wont take a look at this either way of course, I've shared it with the team!

Just a headsup, we still backport critical fixes like security fixes to Core 2, but I'm not sure this one falls within those boundaries.

arthurgeron:
Hey @Ephem, thanks for getting back to me on this. 

>  It would be helpful if you could check and see if this error still reproduces with the latest versions? Note that we have dropped support for Next 14 in this latest version,

Coincidentally, I'm also upgrading the project that uses Clerk to Next.js 16. Once I finish, I'll update you on whether the issue persists.


> we still backport critical fixes like security fixes to Core 2, but I'm not sure this one falls within those boundaries.

If possible, I suggest doing it for this one. Not only because the issue is intrusive and hard to debug, but also because it's a small and precise change. Additionally, it might currently have a significant impact on your user base, considering that weekly install stats for Next.js 14 on npm.js are still much higher than those for v15 or v16.

Ephem:
> Coincidentally, I'm also upgrading the project that uses Clerk to Next.js 16. Once I finish, I'll update you on whether the issue persists.

Perfect, that will be very helpful, thank you!

> If possible, I suggest doing it for this one. Not only because the issue is intrusive and hard to debug, but also because it's a small and precise change. Additionally, it might currently have a significant impact on your user base, considering that weekly install stats for Next.js 14 on npm.js are still much higher than those for v15 or v16.

You make good points! They also apply to a lot of other bug fixes we make though and it's resource intensive to backport everything. I can see this being a core fix though, we'll see where we land on it!

arthurgeron:
This issue is fixed on the current version with Next.js 16

PR #7928, opened 2026-02-25
feat(ui): Support signUpIfMissing with Clerk <SignIn> component

dmoerner:
## Description

The `<SignIn>` component can already be used in a sign-in-or-sign-up flow (`CombinedFlow`) under certain conditions. When strict enumeration protection is enabled, make that combined flow pass the `signUpIfMissing` parameter to the backend to allow an enumeration-safe combined flow. Previously, attempting to use a combined flow with strict enumeration protection enabled was silently broken.

Under the hood, the backend treats sign up if missing as an account transfer. We therefore add support for this account transfer logic when handling first factor verification in the combined sign in flow when strict enumeration protection is enabled.

Fixes USER-4769.

We should merge this in coordination with documentation and a Changelog release, since it will change existing behavior (although with a fix for something that is currently broken)

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -308,7 +308,7 @@ export class SignIn extends BaseResource implements SignInResource {
           return this.reload()
             .then(res => {
               const status = res[verificationKey].status;
-              if (status === 'verified' || status === 'expired') {
+              if (status === 'verified' || status === 'expired' || status === 'transferable') {
                 stop();
```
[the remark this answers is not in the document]
dmoerner: This is something we missed on an earlier PR https://github.com/clerk/javascript/pull/7749 and is not related to clerk SignIn components. However, we can fix it here with a unit test.

packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -1070,7 +1070,7 @@ class SignInFuture implements SignInFutureResource {
           try {
             const res = await this.#resource.__internal_baseGet();
             const status = res.firstFactorVerification.status;
-            if (status === 'verified' || status === 'expired') {
+            if (status === 'verified' || status === 'expired' || status === 'transferable') {
```
dmoerner: This change is not necessary for Clerk builtin components but was missed in https://github.com/clerk/javascript/pull/7749. As suggested by CodeRabbit, let's just fix it now.

dmoerner:
~~Depends on re-revert of https://github.com/clerk/javascript/pull/7962 with Enterprise SSO captcha transfer fix.~~ That was now fixed.

dmoerner:
Working scenarios I tested locally:
- Password (falls back to normal strict enumeration protection sign in)
- Restricted mode (falls back to normal strict enumeration protection sign in)
- Email code sign in and sign up with and without additional requirements
- Email code sign up with immutable attribute
- Email link sign in and sign up with and without additional requirements
- Email code sign in with 2FA enabled
- Phone code sign in and sign up
- Email link sign up transfer with require same device and browser.

Scenarios I didn't test locally:
- Username (should fall back to normal strict enumeration protection sign in): Can you even sign in with a username with the Clerk components when you also have email or phone enabled (since you need those for enumeration protection)?

Under investigation:
- Some sort of a race condition with email links. The polling tab sometimes does not update.

packages/shared/src/internal/clerk-js/navigateToNextStepSignUp.ts
wobsoriano, line 1: Appreciate you moving this to a helper 🙏🏼

packages/ui/src/contexts/components/SignIn.ts
```diff
@@ -127,6 +128,15 @@ export const useSignInContext = (): SignInContextType => {
     );
   }
 
+  // Static preconditions of the sign-up-if-missing flow, shared by SignInStart (which requests
+  // `signUpIfMissing` on sign-in create) and the factor-one cards (which handle the resulting
+  // `transferable` verification status). Per-attempt conditions (identifier type, password use)
+  // stay at the call sites.
+  const signUpIfMissingEnabled =
+    isCombinedFlow &&
+    userSettings.attackProtection.enumeration_protection.enabled &&
```
wobsoriano: `attackProtection` is added to the resource in this PR, so older clerk-js runtimes don't populate it and this throws under ui/clerk-js version skew. Other `userSettings` reads are safe to deref because their fields exist in every runtime in the wild, this one isn't yet.

It's a small safe change

```suggestion
    Boolean(userSettings.attackProtection?.enumeration_protection?.enabled) &&
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2288,6 +2287,14 @@ export class Clerk implements ClerkInterface {
       throw new EmailLinkError(EmailLinkErrorCodeStatus.Expired);
     } else if (verificationStatus === 'client_mismatch') {
       throw new EmailLinkError(EmailLinkErrorCodeStatus.ClientMismatch);
+    } else if (verificationStatus === 'transferable') {
```
wobsoriano, line 2290: Since no session exists here yet, if the verification link replaces the polling tab or the original tab is gone, nothing runs the transfer and the user is told to return to a tab that no longer exists, correct? Should this branch complete the transfer on the current client and mirror the verified handling further down?

packages/ui/src/components/SignIn/SignInFactorOneCodeForm.tsx
```diff
@@ -141,6 +143,24 @@ export const SignInFactorOneCodeForm = (props: SignInFactorOneCodeFormProps) =>
           return clerk.__internal_navigateWithError('..', err.errors[0]);
         }
 
+        if (signUpIfMissingEnabled && signIn.firstFactorVerification.status === 'transferable') {
```
wobsoriano, line 146: The tests covering this branch mock the rejection as `form_identifier_not_found`, but FAPI returns `sign_up_if_missing_transfer` for this case and marks `firstFactorVerification.status` as `transferable`, correct?

wobsoriano (changes requested):
Left some minor comments.

Overall the approach makes sense to me, and to be clear on rollout, the current behavior is already broken for this population. A non-existent account hitting the combined flow with enumeration protection enabled dead-ends today.

We can ship as minor, and maybe with a changelog/docs entry.

packages/ui/src/contexts/components/SignIn.ts
dmoerner: Thank you, implemented!

packages/ui/src/components/SignIn/SignInFactorOneCodeForm.tsx
dmoerner: Great point, fixed.

packages/clerk-js/src/core/clerk.ts
dmoerner: That's a good point. I went back and forth here on how to handle the polling tab or the new tab. I'm going to switch to using the other tab. Currently re-running my E2E tests with this change.

dmoerner: Fixed in https://github.com/clerk/javascript/pull/7928/changes/ec83f7cfdcc459246de799c519e3a0ed1226f1d5

dmoerner:
@wobsoriano Thanks for the review. I made the changes you suggested. I also re-ran my E2E tests against staging covering the following flows:

1. Email link: same client, new user -> sign up successful
2. Email link: Cross-client, new user -> sign up successful
3. Email link: Same client, existing user -> sign in successful
4. Email link: link click without a transfer -> error as expected
5. Email code: New user, no requirements -> sign up successful
6. Email code: New user, name required -> sign up successful after continue
7. Email link: New user, name required -> sign up successful after continue
8. Phone code: New user -> sign up successful
9. Existing user: Sign in successful
10. Restricted mode: Blocked
11. Waitlist mode: Blocked
12. Password supplied: Falls back to existing behavior

wobsoriano (approved):
Thank you so much for resolving the comments 🙏🏼

dmoerner:
Thanks for the review. I also added a dev-only logger warning if the combined flow is enabled with password enabled and strict enumeration protection enabled, since that's the path that's still not covered by this fix.

Going to work on docs and changelog now so we can get this all out together.

packages/clerk-js/src/core/clerk.ts
dmoerner: That fix introduce the risk of races between the two tabs that I found when testing everything end to end one more time in staging. Reverted in https://github.com/clerk/javascript/pull/9328. I think for now we can just accept this risk that if the original tab is gone, you'd have to start the flow again. Email links are little enough used as it is.

PR #7954, opened 2026-02-27
fix(nextjs): make @clerk/nextjs ESM-safe for non-Node.js runtimes

nikosdouvlis:
## Why

`@clerk/nextjs` crashes on import in pure ESM runtimes like Cloudflare Workers (via [vinext](https://github.com/cloudflare/vinext)). Three separate `require()` patterns fail because `require` is not defined in ESM-only environments.

Ref: https://github.com/cloudflare/vinext/issues/73

## What changed

- `auth()`, `auth.protect()`, and `currentUser()` used `require('server-only')` which crashes in ESM runtimes. Replaced with `assertServerOnly()` that checks `typeof require` before calling it. The `server-only` package uses the `react-server` export condition, which vinext/Workers already enforce at the bundler level, so the guard is redundant there.

- `safe-node-apis.js` (both node and browser variants) used `require('node:fs')` / `module.exports`, which crashes in Vite's ESM module runner. Converted to ESM `import`/`export default`.

- `usePathnameWithoutCatchAll` used `require('next/navigation')` to lazily load hooks. Since `@clerk/nextjs` only supports Next.js 15.2.8+, `next/navigation` is always available as a static import.

## Testing

Verified with a vinext smoke test app deployed to Cloudflare Workers (vinext 0.0.18 + Next.js 16.1.6):

- `/` (home with `auth()`) - 200, renders auth state
- `/api/me` (API route with `auth()`) - 200, returns userId/sessionId
- `/sign-in` (`<SignIn />` component) - 200, renders sign-in UI
- `/protected` (`auth.protect()`) - correctly redirects unauthenticated users

Also added 12 Playwright e2e tests across 3 suites (quickstart, protect, auth-state) running against a local vinext dev server in CI.

## Note on vinext CJS support

vinext 0.0.18 added `vite-plugin-commonjs` ([cloudflare/vinext#198](https://github.com/cloudflare/vinext/pull/198)) to handle `require()` at build time. We tested: the build passes with canary `@clerk/nextjs` (without our fixes), but Cloudflare Workers deploy still fails because Workers statically rejects `require()` in the output bundle. The plugin also doesn't handle mixed ESM+CJS files (like `usePathnameWithoutCatchAll.tsx` which has both `import` and `require()`). Our ESM fixes remain necessary.


wobsoriano (approved):

nbardy:
@nikosdouvlis good work, would like to get this merged to deploy on vinext if I can.

Wondering if this merits changes in:
https://github.com/cloudflare/vinext/pull/126

TBH mostly going blind here and letting codex guide me as I'm not an expert on nodejs internals

nbardy:
@nikosdouvlis Is this ready to merge?

ghostyfreak:
Hey guys I feel like this is the only blocker for vinext migration for us

PR #8001, opened 2026-03-05
fix(backend,clerk-js): treat undefined satelliteAutoSync as false

nikosdouvlis:
## Why

Users upgrading to Core 3 expect `satelliteAutoSync` to default to `false` as documented in the JSDoc (`@default false`) and as the Core 3 upgrade codemod implies (it adds `satelliteAutoSync={true}` to existing satellite configs to preserve Core 2 behavior).

However, the runtime check used `=== false` (strict equality), so `undefined` (not passing the prop) behaved identically to `true`, preserving Core 2 auto-sync-on-every-page-load behavior. A customer reported that their satellite domain was redirecting on every page load despite not opting into auto-sync.

## What changed

- Changed the satellite auto-sync gate from `=== false` to `!== true` in both `@clerk/backend` (SSR middleware path in `request.ts`) and `@clerk/clerk-js` (CSR initialization in `clerk.ts`)
- `undefined` and `false` now both skip automatic satellite handshake, matching the documented Core 3 default
- `true` must be explicitly passed to enable auto-sync (opt-in)
- `__clerk_synced=false` (post sign-in redirect) still triggers handshake regardless of the setting

## Test coverage

- Updated existing unit tests that relied on implicit auto-sync to explicitly pass `satelliteAutoSync: true`
- Added new unit tests for `undefined` default behavior (prod + dev, with and without `__clerk_synced=false`)
- Updated integration test that expected 307 redirect with unset `satelliteAutoSync` to expect 200
- Added integration test verifying `__clerk_synced=false` still triggers handshake when `satelliteAutoSync` is unset

## Packages affected

- `@clerk/backend`: satellite handshake decision logic in `request.ts`
- `@clerk/clerk-js`: `#shouldSyncWithPrimary` in `clerk.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * satelliteAutoSync now defaults to false (opt-in); automatic cross-domain satellite sync runs only when explicitly enabled.
  * Sign-out detection corrected so signed-out state is reported accurately.

* **Tests**
  * Updated and added tests to cover auto-sync opt-in behavior and handshake/redirect expectations.

* **Documentation**
  * Added changelog entry documenting the satelliteAutoSync default fix.


packages/clerk-js/src/core/auth/AuthCookieService.ts
```diff
@@ -111,7 +111,7 @@ export class AuthCookieService {
     if (!this.clerk.loaded) {
       return this.clientUat.get() <= 0;
     }
-    return !!this.clerk.user;
+    return !this.clerk.user;
```
jacekradko, line 114: This seems to be outside of the scope of this PR. Should we split this out into its own change so it has a changelog entry?

nikosdouvlis: I think the change is small enough to not warrant a new PR, I also dont think a changeset is needed as its internal behavior

PR #8004, opened 2026-03-06
fix(clerk-js): Do not treat 429 responses as authentication failures

brkalow:
## Summary

429 (Too Many Requests) responses were being caught by `is4xxError` and triggering `handleUnauthenticated`, which signs the user out. This caused a cascading failure pattern observed in production telemetry for a customer.

1. Token fetch fails with a legitimate 4xx (e.g. `authentication_invalid`)
2. `handleUnauthenticated` fires and calls `/touch` to recover
3. `/touch` returns **429** (rate limited)
4. 429 triggers `handleUnauthenticated` **again** (treated as auth failure)
5. Each retry hits 429 → **infinite loop** (30+ retries in 11 seconds observed)
6. User gets forcibly redirected to sign-in

### Changes

- **`@clerk/shared`**: Add `is429Error` helper alongside existing `is4xxError`
- **`AuthCookieService.handleGetTokenError`**: Exclude 429 from triggering `handleUnauthenticated` — fall through to `degraded` status instead
- **`Clerk.setActive`**: Exclude 429 from triggering `handleUnauthenticated`, and also don't re-throw (swallow like other transient errors)
- **`Clerk.#touchCurrentSession`**: Exclude 429 from triggering `handleUnauthenticated`
- **`Session.getToken` retry logic**: Allow 429 to be retried with exponential backoff instead of immediately giving up (was blocked by `is4xxError` in `shouldRetry`)

### Telemetry evidence (7-day window)

| Event | Count |
|---|---|
| `handleUnauthenticated triggered` | 650 (285 unique sessions) |
| `/touch` returning 429 | Majority of `request failed` errors during cascades |
| Top offending session | 59 `handleUnauthenticated` events from a single session |

## Test plan

- [x] `is429Error` unit tests pass (true for 429, false for all other statuses including null/undefined)
- [x] `is4xxError` unit tests pass (still returns true for 429 — no breaking change)
- [x] All 59 Session tests pass
- [x] All 488 clerk-js core tests pass (36 test files)
- [ ] Verify in staging that 429 from `/touch` no longer triggers sign-out
- [ ] Monitor telemetry for reduction in `handleUnauthenticated triggered` events

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of rate-limit (HTTP 429) responses so they no longer trigger sign-out; clients now enter a degraded state and retry, preserving signed-in sessions and avoiding recursive unauthenticated loops.

* **Tests**
  * Added resiliency tests covering 429 rate-limit scenarios and unauthenticated-edge cases to ensure stable runtime behavior under high traffic.

### DIFF
--- a/packages/clerk-js/src/core/resources/Session.ts
+++ b/packages/clerk-js/src/core/resources/Session.ts
@@ -3,8 +3,8 @@ import { isValidBrowserOnline } from '@clerk/shared/browser';
 import {
   ClerkOfflineError,
   ClerkWebAuthnError,
-  is429Error,
   is4xxError,
+  is429Error,
   MissingExpiredTokenError,
 } from '@clerk/shared/error';
 import {


packages/shared/src/errors/helpers.ts
```diff
@@ -37,6 +37,36 @@ export function is4xxError(e: any): boolean {
...
+ * because it's also returned for unrelated resources (org not found, JWT template
+ * not found) and shares the same `resource_not_found` error code, making it
+ * impossible to distinguish. Session-not-found 401s are already handled directly
+ * by Base._fetch.
+ *
+ * @internal
+ */
+export function isUnauthenticatedError(e: any): boolean {
+  const status = e?.status;
+  return status === 401 || status === 422;
```
jacekradko, line 67: Nice!

PR #8020, opened 2026-03-10
chore(repo): Various release workflow fixes

dominic-clerk:
## Description

Following #7915 there were many small pre-existing things to fix. Now that we went through a whole release loop with the new workflow and everything works fine I packaged the fixes in here.

1. Use the built-in `GITHUB_TOKEN` everywhere possible
2. Fix the bug in the PR freshness check
3. Use `pnpm` instead of `npm` / `npx` where relevant
4. Await workflow dispatches instead of fire-and-forget

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] other: CI

## Release Notes

* **Chores**
  * Optimized GitHub Actions workflows with consolidated dispatch calls for improved CI/CD efficiency.
  * Updated build tooling in release workflows for consistency and reduced command overhead.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -84,28 +84,29 @@ jobs:
               const clerkUiVersion = require('./packages/ui/package.json').version;
               const nextjsVersion = require('./packages/nextjs/package.json').version;
 
-              github.rest.actions.createWorkflowDispatch({
-                owner: 'clerk',
-                repo: 'sdk-infra-workers',
-                workflow_id: 'update-pkg-versions.yml',
-                ref: 'main',
-                inputs: { clerkjsVersion, clerkUiVersion }
-              })
-
-              github.rest.actions.createWorkflowDispatch({
-                owner: 'clerk',
-                repo: 'dashboard',
-                workflow_id: 'prepare-nextjs-sdk-update.yml',
-                ref: 'main',
-                inputs: { version: nextjsVersion }
-              })
-
-              github.rest.actions.createWorkflowDispatch({
-                owner: 'clerk',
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -213,6 +220,7 @@ jobs:
     permissions:
       contents: read
       id-token: write
+      pull-requests: write
```
dominic-clerk, line 223: Needed so that `GITHUB_TOKEN` can react to `!snapshot` comments and write the response when the release finishes.

.github/workflows/release.yml
```diff
@@ -249,16 +257,14 @@ jobs:
           result-encoding: string
           retries: 3
           retry-exempt-status-codes: 400,401
-          github-token: ${{ secrets.CLERK_COOKIE_PAT }}
```
dominic-clerk, line 252: `GITHUB_TOKEN` is the default value, no need to be explicit, see https://github.com/actions/github-script/tree/v7

.github/workflows/release.yml
```diff
@@ -335,15 +341,13 @@ jobs:
         if: steps.version-packages.outputs.success == '1'
         uses: peter-evans/create-or-update-comment@v3.0.0
         with:
-          token: ${{ secrets.CLERK_COOKIE_PAT }}
```
dominic-clerk, line 338: GITHUB_TOKEN is the default value, no need to be explicit, see https://github.com/peter-evans/create-or-update-comment/tree/v3.0.0

PR #8027, opened 2026-03-10
fix(repo): add recovery for release downstream notifications

jacekradko:
## Summary

- Adds a recovery step to the release workflow that catches the case where `changeset publish` succeeds (packages land on npm) but the changesets action step fails afterward (e.g. during `git push --follow-tags`)
- When the changesets step fails, the `published` output is never set to `true`, so downstream repo notifications are silently skipped — and on retry, nothing is left to publish

**Root cause investigation:** The `6.1.0` stable release on March 9 published to npm but the changesets action failed on attempt 1. Attempt 2 found "No unpublished projects" and also skipped the dispatch. As a result, `sdk-infra-workers` was stuck on `6.0.0` until manually dispatched today.

### How the recovery step works

1. Only runs when the changesets step **fails** (`if: always() && steps.changesets.conclusion == 'failure'`)
2. Reads the local `package.json` versions for `@clerk/clerk-js` and `@clerk/ui`
3. Skips if it's a pre-release (canary/snapshot)
4. Checks npm to see if **either** package was actually published
5. If at least one was published, dispatches to all downstream repos (same targets as the normal trigger)

The dispatches are idempotent — `sdk-infra-workers` just pins the version, `dashboard` prepares an update, `clerk-docs` rebuilds typedoc.

## Test plan

- [ ] Verify workflow YAML is valid
- [ ] Recovery step is a no-op in normal version-mode runs (changesets succeeds → condition not met)
- [ ] Recovery step is a no-op when changesets fails before publishing (npm check returns not found for both)
- [ ] Recovery step dispatches when changesets fails after publishing either package

## Summary by CodeRabbit

* **Chores**
  * Improved release workflow reliability with an automatic recovery mechanism that dispatches updates to downstream workflows when packages are confirmed published, while avoiding pre-release disruptions.
  * Added a minimal placeholder changeset file to track release metadata.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,7 +73,6 @@ jobs:
           NPM_CONFIG_PROVENANCE: true
 
       - name: Trigger workflows on related repos
-        id: trigger
         if: steps.changesets.outputs.published == 'true'
         uses: actions/github-script@v7
         with:
@@ -144,6 +143,12 @@ jobs:
               return;
             }
 
+            const preMode = require("fs").existsSync("./.changeset/pre.json");
+            if (preMode) {
+              core.warning("Changeset in pre-mode, skipping recovery dispatch");
+              return;
+            }
+
             // Check if either version was actually published to npm
             function isPublished(name, version) {
               try {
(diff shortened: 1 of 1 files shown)


jacekradko:
Superseded by #8102 which combines this recovery step with the `Promise.allSettled` fix for dispatch resilience.

PR #8028, opened 2026-03-10
fix(shared): Add Netlify-Vary header to prevent CDN caching of auth responses

wobsoriano:
Consolidate Netlify cache handling into a single `handleNetlifyCacheHeaders` function that sets `Netlify-Vary: cookie=__client_uat,cookie=__session` on all auth responses when running on Netlify. This prevents the CDN from serving stale session state across different users/sessions.

The function is called once per request in each framework SDK middleware, right after `authenticateRequest` returns, replacing the previous dev-only cache-bust approach with proper CDN cache isolation for both dev and prod.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


wobsoriano:
!snapshot

wobsoriano:
!snapshot

PR #8030, opened 2026-03-10
feat(clerk-js): Sign up if missing

dmoerner:
## Description

This PR re-reverts https://github.com/clerk/javascript/pull/7749, which adds support for sign up if missing for custom flows. We do this by adding the optional `signUpIfMissing` param on creating a sign in, and adding support for captchas. This requires both supporting captchas on sign in if `signUpIfMissing` is `true`, and making sure we don't show a double captcha on sign up after transfer from a sign up if missing sign in.

The original PR had to be reverted because it did not properly render captchas on Enterprise SSO transfers to sign up. The root of this mistake was my confusion over a bug in the existing code: We bailed out early in `shouldBypassCaptchaForAttempt` if there was no strategy. This had the effect that we unconditionally showed captchas on all transfers. This was correct behavior for EnterpriseSSO, which does not bypass captcha on transfer. This was a bug for OAuth, since we never checked if we should bypass captchas for OAuth.

This PR is identical to the previous one, except the logic in `shouldBypassCaptchaForAttempt` is fixed: We do show captchas on EnterpriseSSO transfers, and we properly check the bypass list for OAuth transfers. I also added further unit tests to verify these fixes.

This was tested in staging with SCIM to confirm that we did not have the EnterpriseSSO Captcha regression.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Added `sign_up_if_missing` parameter to sign-in requests, enabling automatic account creation for users who don't yet have an account
  * Integrated CAPTCHA bot protection challenges into the sign-in flow when your security configuration requires it
  * Improved CAPTCHA bypass logic for passwordless authentication strategies, including email code, email link, and phone code verification


packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -561,22 +554,30 @@ export class SignUp extends BaseResource implements SignUpResource {
    * We delegate bot detection to the following providers, instead of relying on turnstile exclusively
    */
   protected shouldBypassCaptchaForAttempt(params: SignUpCreateParams) {
-    if (!params.strategy) {
```
dmoerner, line 564: This is the previous logic, which unconditionally showed captchas on EnterpriseSSO and OAuth transfers, since transfers do not have a strategy. The former behavior is correct, the latter was a bug.

packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -561,22 +554,30 @@ export class SignUp extends BaseResource implements SignUpResource {
...
-      return false;
-    }
-
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     const captchaOauthBypass = SignUp.clerk.__internal_environment!.displayConfig.captchaOauthBypass;
 
-    if (captchaOauthBypass.some(strategy => strategy === params.strategy)) {
-      return true;
+    // Check for transfer captcha bypass.
+    if (params.transfer) {
```
dmoerner, line 561: The new logic explicitly checks transfer first and checks both paths that can lead to a bypass: OAuth strategy, and sign up if missing. We fall through on EnterpriseSSO transfers, which then ends up returning `false`, i.e., we do show captchas on Enterprise SSO.

packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -787,22 +804,30 @@ class SignUpFuture implements SignUpFutureResource {
...
```
dmoerner, line 811: It's exactly the same change for sign up future.

packages/clerk-js/src/core/resources/__tests__/SignUp.test.ts
```diff
@@ -35,6 +35,114 @@ describe('SignUp', () => {
...
+        },
+      } as any;
+
+      const signUp = new SignUp();
+      await signUp.create({ transfer: true });
+
+      expect(CaptchaChallenge).not.toHaveBeenCalled();
+    });
+
+    it('does not skip captcha challenge for enterprise_sso transfer', async () => {
```
dmoerner, line 117: New unit test to rule out regressions with `enterprise_sso` transfer.

packages/clerk-js/src/core/resources/__tests__/SignUp.test.ts
```diff
@@ -152,6 +266,102 @@ describe('SignUp', () => {
...
+        const signUp = new SignUp();
+        await signUp.__internal_future.create({ transfer: true });
+
+        expect(CaptchaChallenge).not.toHaveBeenCalled();
+        expect(mockFetch.mock.calls[0]?.[0].body).toHaveProperty('captchaToken', undefined);
+        expect(mockFetch.mock.calls[0]?.[0].body).toHaveProperty('captchaWidgetType', undefined);
+        expect(mockFetch.mock.calls[0]?.[0].body).toHaveProperty('captchaError', undefined);
+      });
+
+      it('does not skip captcha challenge for enterprise_sso transfer', async () => {
```
dmoerner, line 338: Sign up future regression test for `enterprise_sso` transfer case.

dmoerner:
CI seems broken, I'm going to try converting to draft and then switching back to see if I can get it to re-run.

PR #8032, opened 2026-03-10
feat(expo): add two-way JS/native session sync for expo native components

chriscanin:
## Summary

- **ClerkProvider**: Skip native `configure()` when no bearer token exists, preventing creation of anonymous native clients that conflict with later JS→native token sync
- **ClerkProvider**: Add `NativeSessionSync` component that automatically pushes JS SDK bearer token to native when user signs in via JS custom forms
- **ClerkViewFactory (iOS)**: Clear stale cached client/environment from keychain when device token changes, preventing 400 API errors from mismatched client IDs during `refreshClient()`
- **UserButton / useUserProfileModal**: Sync JS bearer token to native before presenting the profile modal

### Root cause

When users sign in via JS SDK (custom forms, `useSignIn`, etc.) rather than native `<AuthView />`, the native Clerk SDK has no session. The native SDK's `CacheManager` may load a stale anonymous client from keychain (created by a previous launch), and when `refreshClient()` runs, it sends both the new device token (`Authorization` header) and the stale client ID (`x-clerk-client-id` header) — causing a 400 API error.

## Test plan

- [ ] Sign in via JS custom form → tap UserButton → native profile modal shows user profile
- [ ] Sign in via native AuthView → tap UserButton → native profile modal shows user profile  
- [ ] Sign out from native profile modal → JS SDK `useAuth()` updates reactively
- [ ] App restart after JS sign-in → native components still work
- [ ] Fresh install (no cached data) → no crash on first configure

## Summary by CodeRabbit

* **Bug Fixes**
  * Clears cached auth data when device tokens change to prevent stale client IDs.
  * Improved sign-out coordination so JS and native sessions stay in sync and avoid unnecessary sign-outs.
  * getSession and signOut now return null (instead of error) when the native client is not initialized.

* **New Features**
  * Automatic synchronization of JS and native authentication on sign-in and before showing native profile UI.
  * Faster client refresh when a new bearer token is provided.
  * Public API to read the native client token.

* **Chores**
  * Improved native initialization and token handling flows.


packages/expo/android/src/main/java/expo/modules/clerk/ClerkExpoModule.kt
```diff
@@ -108,6 +127,55 @@ class ClerkExpoModule(reactContext: ReactApplicationContext) :
...
     }
 
+    /**
+     * Forces the Clerk SDK to re-fetch client/environment data from the API.
+     *
+     * This is needed when a new device token has been written to SharedPreferences
+     * but the SDK was already initialized (so Clerk.initialize() is a no-op).
+     *
+     * Uses reflection to find the ConfigurationManager instance by type (field name
+     * may vary across SDK versions), then sets _isInitialized to false so
```
swolfand: I'm curious about the use of reflection here. Generally considered an anti pattern in Java world

chriscanin: Clerk.initialize() is a no-op if the SDK is already initialized, and the Clerk Android SDK doesn't expose any public API to force a re-initialization or swap the device token post-init. We need to trigger a fresh client/environment fetch after writing a new bearer token to SharedPreferences, and there's no way to do that (as far as I can tell) from the android SDK.

The reflection approach finds ConfigurationManager by type (to be resilient to field name changes across SDK versions), then finds the _isInitialized MutableStateFlow<Boolean> and flips it to false so reinitialize() will proceed.

Let me know if you can think of a cleaner way to do it, and I'll put it in!

packages/expo/ios/ClerkViewFactory.swift
mikepitre, line 50: If this function doesn't take in the keychain service as a parameter, I'm not sure how it would ever match the service used by the iOS SDK if they wanted to use a custom service. It looks like the Clerk.configure is always just called with a pub key, and never keychain config.

mikepitre: I think for the v1 of this its fine to not support custom keychain services, but you might get asked to eventually implement many of the [configuration options](https://clerk.com/docs/ios/reference/native-mobile/configuration#configuration-options) offered by the ios sdk.

mikepitre:
@codex review

packages/expo/src/hooks/useUserProfileModal.ts
```diff
@@ -66,17 +69,36 @@ export function useUserProfileModal(): UseUserProfileModalReturn {
...
+        if (!hadNativeSessionBefore) {
+          const bearerToken = (await tokenCache?.getToken(CLERK_CLIENT_JWT_KEY)) ?? null;
+          if (bearerToken) {
+            await ClerkExpo.configure(clerk.publishableKey, bearerToken);
+
+            // Re-check if configure produced a session
+            const postConfigure = (await ClerkExpo.getSession()) as NativeSessionResult | null;
+            hadNativeSessionBefore = !!(postConfigure?.sessionId || postConfigure?.session?.id);
+          }
+        }
```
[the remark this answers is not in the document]
chriscanin: This is already handled. The native `configure()` on iOS (`ClerkViewFactory.swift:71-77`) polls for up to 3s waiting for `Clerk.shared.isLoaded && Clerk.shared.session != nil` before resolving. On subsequent calls, it `await`s `Clerk.shared.refreshClient()` which also completes before returning.

By the time the JS `await ClerkExpo.configure(...)` promise resolves, the native session is already available — so the single post-configure `getSession()` re-check is sufficient. There's no race to poll for on the JS side.

packages/expo/android/src/main/java/expo/modules/clerk/ClerkExpoModule.kt
swolfand: We can add those hooks for you in the Android SDK to avoid it if you'd like?

chriscanin: I've pushed a commit that drops the entire reflection hack and uses `updateDeviceToken()` instead. It saves the token via `StorageHelper` and triggers `refreshClientAndEnvironment()` internally.

chriscanin: Things are working well. I am going to approve the PR here: https://github.com/clerk/clerk-android/pull/564#pullrequestreview-3932635905
If you can publish to maven, that would be amazing!

packages/expo/ios/ClerkViewFactory.swift
chriscanin: `ClerkViewFactory` now resolves the keychain service internally via a computed property that checks `Info.plist` for a `ClerkKeychainService` key first, then falls back to `Bundle.main.bundleIdentifier`. That resolved service gets passed to `Clerk.configure(publishableKey:options:)` as `Clerk.Options(keychainConfig: .init(service: service))`. So it matches whatever the iOS SDK uses without needing to pipe it through the bridge method signature.

I can start working towards including configuration options as a followup after the theming (and possibly testing, but can discuss the testing later in the week / monday). It will be important to include these.

packages/expo/src/provider/ClerkProvider.tsx
```diff
@@ -100,15 +171,25 @@ export function ClerkProvider<TUi extends Ui = Ui>(props: ClerkProviderProps<TUi
...
               if (__DEV__) {
                 console.warn('[ClerkProvider] Token cache read failed:', e);
               }
             }
+
+            // Only configure native SDK if we have a bearer token.
+            // Configuring without a token creates an anonymous native client,
+            // which conflicts when we later try to inject the JS SDK's token.
+            if (!bearerToken) {
+              return;
```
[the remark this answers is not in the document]
chriscanin: 👍

A fix has been provided for this. The current code in `ClerkProvider.tsx` calls `ClerkExpo.configure(pk, bearerToken)` unconditionally on mount regardless of whether `bearerToken` is null. There's no early return based on token presence.

packages/expo/ios/ClerkExpoModule.swift
```diff
@@ -170,37 +170,6 @@ class ClerkExpoModule: RCTEventEmitter {
...
-    if status == errSecSuccess, let data = result as? Data {
-      resolve(String(data: data, encoding: .utf8))
-    } else {
-      resolve(nil)
-    }
-  }
-
   // MARK: - signOut
 
   @objc func signOut(_ resolve: @escaping RCTPromiseResolveBlock,
```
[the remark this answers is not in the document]
chriscanin: 👍 

A fix has been provided for this now as well. `getClientToken` exists in both `ClerkExpoModule.swift` (as `@objc func getClientToken`) and the bridge file `ClerkExpoModule.m` (as `RCT_EXTERN_METHOD`). It's also defined in `ClerkViewFactoryProtocol` and implemented in `ClerkViewFactory` where it delegates to `readNativeDeviceToken()`.

chriscanin:
!snapshot

packages/expo/src/provider/ClerkProvider.tsx
```diff
@@ -52,6 +53,76 @@ const SDK_METADATA = {
...
+  publishableKey: string;
+  tokenCache: TokenCache | undefined;
+}) {
+  const { isSignedIn } = useAuth();
+  const hasSyncedRef = useRef(false);
+  // Use the provided tokenCache, falling back to the default SecureStore cache
+  const effectiveTokenCache = tokenCache ?? defaultTokenCache;
+
+  useEffect(() => {
+    if (!isSignedIn) {
```
mikepitre, line 79: [P1] Clear the native session when JS auth signs out

When `isSignedIn` becomes false this effect only resets `hasSyncedRef` and returns, so the native Clerk SDK keeps the previously mirrored session/device token. Any JS-side sign-out path (`clerk.signOut()`, session expiry, custom auth UI) will therefore leave `ClerkExpo.getSession()` non-null; components like `UserButton` immediately read that native session back and continue showing/opening the signed-in profile even though the JS SDK is signed out.

chriscanin: This has been fixed. The `NativeSessionSync` effect now calls `ClerkExpo.signOut()` when `isSignedIn` becomes false (lines 84-91 in ClerkProvider.tsx). This clears the native SDK's session so `getSession()` returns null and native components like `UserButton` won't show stale signed-in state.

Note: in the typical quickstart pattern where `UserButton` is only rendered inside an `if (isSignedIn)` guard, the stale native session was invisible since the component unmounts. But this fix makes it correct regardless of how the consumer structures their component tree.

chriscanin:
!snapshot

PR #8035, opened 2026-03-11
feat(*): auto-proxy for eligible hosts

brkalow:
## Description

This PR implements automatic FAPI proxying for Vercel preview deployments. When an app is deployed to a `.vercel.app` subdomain without explicit proxy or domain configuration, the SDK automatically routes FAPI requests through the app's own domain via the `/__clerk` proxy path.

The auto-detection is centralized across all SDKs:
- **Shared utilities** (`@clerk/shared/proxy`): New `isVercelPreviewDeploy()` helper detects `.vercel.app` hostnames
- **Client-side** (`@clerk/clerk-js`): Auto-detect in `proxyUrl` getter when `window.location.hostname` is `.vercel.app`
- **Server-side** (`@clerk/backend`): Auto-detect in `authenticateContext` constructor when request hostname is `.vercel.app`
- **Next.js middleware** (`@clerk/nextjs`): Auto-enable proxy interception for `/__clerk/*` requests on `.vercel.app` hostnames

Guard conditions prevent auto-detection when explicit configuration is provided: `proxyUrl`, `domain`, or environment variables `NEXT_PUBLIC_CLERK_PROXY_URL` / `NEXT_PUBLIC_CLERK_DOMAIN`.

## Test Results

- `@clerk/shared`: 981/981 tests passed
- `@clerk/backend`: 1114/1114 tests passed  
- `@clerk/clerk-js`: 624/624 tests passed
- `@clerk/nextjs`: 362 tests passed (49 pre-existing failures unrelated to this change)

Build verification: All packages build cleanly with no type errors.

## Checklist

- [x] `pnpm test` runs as expected
- [x] `pnpm build` runs as expected
- [x] Tests added for new functionality

## Type of change

- [x] 🌟 New feature

## Summary by CodeRabbit

* **New Features**
  * Automatic proxy detection for Vercel-style hosts: when no explicit proxy or domain is provided, the frontend API routing can auto-configure a relative proxy for eligible deployments; explicit proxy or domain settings take precedence.

* **Bug Fixes**
  * Improved handling of relative proxy URLs in non-browser/server contexts and script-host resolution fallback.

* **Tests**
  * Added comprehensive tests validating auto-detection, precedence rules, environment-based behavior, and various host scenarios.


brkalow:
!snapshot

brkalow:
!snapshot

brkalow:
!snapshot

brkalow:
!snapshot

brkalow:
!snapshot

brkalow:
!snapshot

dominic-clerk:
See https://clerkinc.slack.com/archives/C0AJBKMHB51/p1774458716918769 before merging 🙏

packages/clerk-js/src/core/__tests__/clerk.test.ts
Ephem, line 1: NIT: A possible extra test here is to check that it doesn't break when `window` is not defined (for native envs).

The `if (inBrowser()) {` already takes care of it, but tests are always nice.

packages/nextjs/src/server/clerkMiddleware.ts
```diff
@@ -159,12 +160,16 @@ export const clerkMiddleware = ((...args: unknown[]): NextMiddleware | NextMiddl
       );
 
       // Handle Frontend API proxy requests early, before authentication
-      const frontendApiProxyConfig = resolvedParams.frontendApiProxy;
+      const requestUrl = new URL(request.nextUrl.href);
+      const frontendApiProxyConfig =
+        resolvedParams.frontendApiProxy ??
+        (resolvedParams.proxyUrl || PROXY_URL || resolvedParams.domain || DOMAIN
+          ? undefined
+          : getAutoDetectedProxyConfig(requestUrl));
```
Ephem: NIT: Works as is but is not the easiest to follow imo. I'd suggest either moving everything to an external `resolveFrontendApiProxyConfig(resolvedParams)` function, or inlining it all something like this:

```suggestion
      // Handle Frontend API proxy requests early, before authentication
      let frontendApiProxyConfig = resolvedParams.frontendApiProxy;

      // Auto-detect when no explicit proxy or domain is configured
      const hasExplicitProxyOrDomain = resolvedParams.proxyUrl || PROXY_URL || resolvedParams.domain || DOMAIN;
      if (!frontendApiProxyConfig && !hasExplicitProxyOrDomain) {
        const requestUrl = new URL(request.nextUrl.href);
        if (shouldAutoProxy(requestUrl.hostname)) {
          frontendApiProxyConfig = { enabled: true };
        }
      }
```

packages/shared/src/proxy.ts
```diff
@@ -30,7 +32,61 @@ export function proxyUrlToAbsoluteURL(url: string | undefined): string {
...
+
+  if (typeof window === 'undefined' || !window.location?.origin) {
+    return url;
+  }
+
+  return new URL(url, window.location.origin).toString();
+}
+
+const AUTO_PROXY_HOST_SUFFIXES = ['.vercel.app'];
+const AUTO_PROXY_PATH = '/__clerk';
```
Ephem: I think we should export this and use it in any place we call `shouldAutoProxy` and currently hardcode `/__clerk`.

packages/shared/src/proxy.ts
```diff
@@ -30,7 +32,61 @@ export function proxyUrlToAbsoluteURL(url: string | undefined): string {
...
+
+export function shouldAutoProxy(hostname: string): boolean {
+  return AUTO_PROXY_HOST_SUFFIXES.some(hostSuffix => hostname?.endsWith(hostSuffix)) ?? false;
+}
```
Ephem, line 52: The biggest unknown for me with this PR is _when_ should we be auto-proxying exactly? If I understand the PR and the context I've gathered elsewhere, what we are trying to achieve is being able to use production Clerk instances on `.vercel.app` domains without setting up a custom domain for it right?

People might have already have working _development_ instances set up on `.vercel.app` domains today (I think?), and it feels like we wouldn't want to proxy those?

One reason is, if we deploy `clerk-js` with auto-proxying when people have not updated their other SDK packages will likely break things? Since dev instances already work, changing from not proxying to proxying in a minor/patch also just seems off to me.

However, this function is _not_ guarding for only auto-proxying production Clerk instances, and it is used in a few places that adds proxy config when this returns `true`. I'm having a hard time judging if that's intentional or not?

If it is intentional, I'm also not sure if the possible build time (`getAutoProxyUrlFromEnvironment`) / run time (`shouldAutoProxy`) mismatch can turn out to be a problem? 🤔 (The `proxyUrl` that could get passed to `ClerkProvider` at build time might mismatch with what `clerk-js` decides at run time as conditions differ.)

packages/clerk-js/src/core/__tests__/clerk.test.ts
Railly: Addressed in [`0859708`](https://github.com/clerk/javascript/commit/08597080b), also updated existing tests to use production key since auto-proxy is now production-only

packages/nextjs/src/server/clerkMiddleware.ts
Railly: Adopted your suggestion, also removed the now-unused `getAutoDetectedProxyConfig` function. Addressed in [635c9b7](https://github.com/clerk/javascript/commit/635c9b7a1).

packages/shared/src/proxy.ts
Railly: Exported and replaced the two hardcoded occurrences. Addressed in [`96842db`](https://github.com/clerk/javascript/commit/96842db05)

packages/shared/src/proxy.ts
Railly: Great catch 😄, we should auto-proxy for production-only apps that have the `*.vercel.app` associated.

Fixed by guarding with `instanceType === 'production'` in clerk-js and `pk_live_` prefix check in the backend authenticateContext. This aligns with `getAutoProxyUrlFromEnvironment` which already had the production guard.

Here's the full integration overview for context: [v0 + Clerk Integration Status](https://www.notion.so/clerkdev/v0-Clerk-Integration-Status-3422b9ab44fe8070bb92e952752cbe33)

packages/backend/src/tokens/__tests__/authenticateContext.test.ts
```diff
@@ -258,6 +258,46 @@ describe('AuthenticateContext', () => {
     });
   });
 
+  describe('auto-proxy for eligible hosts', () => {
+    it('auto-derives proxyUrl for eligible hostnames', async () => {
+      const clerkRequest = createClerkRequest(new Request('https://myapp-abc123.vercel.app/dashboard'));
+      const context = await createAuthenticateContext(clerkRequest, {
+        publishableKey: pkTest,
```
nikosdouvlis: The auto-proxy tests here still use `pkTest` but the guard on line 75 checks for `pk_live_`. This positive case cant pass at runtime, it only shows green because vitest runs in typecheck mode. Should swap to `pkLive` and add a negative case for dev keys.

packages/backend/src/tokens/authenticateContext.ts
```diff
@@ -70,6 +71,14 @@ class AuthenticateContext implements AuthenticateContext {
     private clerkRequest: ClerkRequest,
     options: AuthenticateRequestOptions,
   ) {
+    // Auto-detect proxy for supported platform deployments (production only)
+    if (!options.proxyUrl && !options.domain && options.publishableKey?.startsWith('pk_live_')) {
+      const hostname = clerkRequest.clerkUrl.hostname;
```
nikosdouvlis: Just double checking - the hostname here comes from `clerkUrl` which trusts `x-forwarded-host`. On Vercel thats fine since the edge sets it, but outside Vercel a misconfigured reverse proxy could let someone spoof `X-Forwarded-Host: anything.vercel.app` and flip auto-proxy on. Im just being paranoid, but could this be exploitable outside Vercel? I don't think it matters, just raising for the sake of completeness :)

packages/shared/src/proxy.ts
```diff
@@ -30,7 +32,61 @@ export function proxyUrlToAbsoluteURL(url: string | undefined): string {
...
+const AUTO_PROXY_HOST_SUFFIXES = ['.vercel.app'];
+export const AUTO_PROXY_PATH = '/__clerk';
+
+export function shouldAutoProxy(hostname: string): boolean {
+  return AUTO_PROXY_HOST_SUFFIXES.some(hostSuffix => hostname?.endsWith(hostSuffix)) ?? false;
+}
+
+function normalizeHostname(hostnameOrUrl: string): string {
+  if (hostnameOrUrl.startsWith('http://') || hostnameOrUrl.startsWith('https://')) {
+    return new URL(hostnameOrUrl).hostname;
```
nikosdouvlis: `normalizeHostname` can throw on malformed `VERCEL_PROJECT_PRODUCTION_URL`. To be extra safe, wrap in try/catch and return `''`

packages/backend/src/tokens/authenticateContext.ts
```diff
@@ -70,6 +71,14 @@ class AuthenticateContext implements AuthenticateContext {
     private clerkRequest: ClerkRequest,
     options: AuthenticateRequestOptions,
   ) {
+    // Auto-detect proxy for supported platform deployments (production only)
+    if (!options.proxyUrl && !options.domain && options.publishableKey?.startsWith('pk_live_')) {
```
nikosdouvlis: Could use `isProductionFromPublishableKey` from `@clerk/shared/keys` here instead of raw `startsWith('pk_live_')` for consistency

packages/backend/src/tokens/__tests__/authenticateContext.test.ts
Railly: Addressed in [`66a8ad1`](https://github.com/clerk/javascript/commit/66a8ad1a6)

packages/backend/src/tokens/authenticateContext.ts
Railly: Good callout. The hostname is derived from `X-Forwarded-Host` which is authoritative on Vercel's edge but spoofable behind misconfigured proxies. In practice, auto-proxy also requires a `pk_live_*` key and no explicit proxy/domain config. The spoofed value only affects `frontendApi` derivation (cookie suffix, cross-origin checks), JWKS fetches use `apiUrl` (BAPI) not `frontendApi`, so no outbound requests to the spoofed host. Worst case is a self-DoS where cookie suffix mismatches. Added a comment documenting the trust assumption.

Railly: Addressed in [`cba7dfd`](https://github.com/clerk/javascript/commit/cba7dfd29)

packages/shared/src/proxy.ts
Railly: Addressed in [`c9c28b0`](https://github.com/clerk/javascript/commit/c9c28b0ef)

packages/backend/src/tokens/authenticateContext.ts
Railly: Addressed in [`2371ab2`](https://github.com/clerk/javascript/commit/2371ab22f)

Railly:
!snapshot

Railly:
@Ephem Tested with a real production instance on vercel.app, auth works end-to-end. 

**One finding**: the default middleware matcher excludes `.js` files, which prevents `/__clerk/npm/@clerk/clerk-js@6/dist/clerk.browser.js` from reaching the middleware. Had to add `"/__clerk/(.*)"` to the matcher manually. Since auto-proxy is meant to work without config, I think `clerkMiddleware` should handle this automatically or the default matcher should always include `__clerk`.

Test app: https://v0-auto-proxy-test.vercel.app

packages/shared/src/proxy.ts
```diff
@@ -30,7 +32,65 @@ export function proxyUrlToAbsoluteURL(url: string | undefined): string {
...
+  }
+
+  const vercelProductionHostname = environment.VERCEL_PROJECT_PRODUCTION_URL;
+
+  if (!vercelProductionHostname || !shouldAutoProxy(normalizeHostname(vercelProductionHostname))) {
+    return '';
+  }
+
+  return AUTO_PROXY_PATH;
 }
```
Ephem, line 100: This function is a bit tricky to reason about. It runs in Next.js only from `mergeNextClerkPropsWithEnv`. That function in turn runs from the providers, and a key thing to understand is that it runs both during _build time_ and during _run time_.

The return here ends up as the `proxyUrl`. During build time, this can make its way to the `<script src="/__clerk/npm/@clerk/clerk-js@…">` for statically generated pages.

In other places, we use the hostname, or window location to determine if we should proxy, which means it can mismatch when:

* Vercel preview deployment+Clerk production instance
  * We do _not_ proxy the script, but we do proxy in the other places

This is likely fine, as that case doesn't work today either, but I wonder if we could widen the `VERCEL_TARGET_ENV` check to include `'preview'` here and make preview deploys work too? `VERCEL_PROJECT_PRODUCTION_URL` does seem to be set correctly for the preview target as well as per [these docs](https://vercel.com/docs/environment-variables/system-environment-variables#VERCEL_PROJECT_PRODUCTION_URL).

We might also want to add a comment explaining this will also resolve at build time since that's not easy to understand from reading the code.

packages/backend/src/tokens/authenticateContext.ts
Ephem: I'm not sure a comment is enough here (it also doesn't explain _why_ this is currently safe). It's very hard to fully understand the implications of something like this, especially as code changes over time, so I think it doesn't hurt to be defensive.

I wonder if we could straight up reuse `getAutoProxyUrlFromEnvironment` here (with minor tweaks)? That way, if the backend is deployed to Vercel auto-proxying will work correctly, which should be fine for the use case we are targeting?

If the backend is hosted separately, I think requiring you to configure a `proxyUrl` for it yourself is fair instead of picking it up from the header?

Ephem (commented):
I think this is looking really good, I only worry slightly about the backend case, made a suggestion.

We still need to figure out the `/__clerk/(.*)` not being in the matcher by default too. Not sure that's going to affect this PR though, but we should probably figure it out before merging.

packages/shared/src/proxy.ts
Railly: For the v0 flow, preview deploys get `pk_test_` by default (the integration sets dev keys as the default, production keys only as environment override). Dev instances connect to FAPI directly without proxy, so preview already works today.

The theoretical case of `pk_live_` on preview would likely fail anyway because the `Clerk-Proxy-Url` header would contain the preview hostname, which isn't registered as a domain in Clerk. Domain sync webhooks only fire for `target: 'production'` events, so Clerk never learns about preview hostnames.

I think keeping `production` only is the right call.

[Added a comment explaining the build-time/runtime dual execution context.](https://github.com/clerk/javascript/blob/e0a275dc9/packages/shared/src/proxy.ts#L73-L99)

Railly:
> I'm not sure a comment is enough here (it also doesn't explain why this is currently safe). It's very hard to fully understand the implications of something like this, especially as code changes over time, so I think it doesn't hurt to be defensive.
> I wonder if we could straight up reuse getAutoProxyUrlFromEnvironment here (with minor tweaks)? That way, if the backend is deployed to Vercel auto-proxying will work correctly, which should be fine for the use case we are targeting?
> If the backend is hosted separately, I think requiring you to configure a proxyUrl for it yourself is fair instead of picking it up from the header?

@Ephem addressed in [`43be86f`](https://github.com/clerk/javascript/commit/43be86fb2), replaced header-based detection with `getAutoProxyUrlFromEnvironment` in the backend. If deployed outside Vercel, auto-proxy simply doesn't activate and you configure proxyUrl manually.

packages/shared/src/proxy.ts
Ephem: > The theoretical case of `pk_live_` on preview would likely fail anyway because the `Clerk-Proxy-Url` header would contain the preview hostname, which isn't registered as a domain in Clerk. Domain sync webhooks only fire for `target: 'production'` events, so Clerk never learns about preview hostnames.

Ah of course, that's a very good point (I hate that saying this nowadays makes me sound like an agent 😂), I was focused on the auto-proxy logic mismatch between the different places in the codebase so didn't think that far.

I guess it feels equally weird to knowingly have different logic for when to auto-proxy in different places, as it feels to proxy for preview envs when we know it wont work further down the line. Not a big deal either way though.

Great comment, thanks!

Ephem (approved):
> @Ephem addressed in [43be86f](https://github.com/clerk/javascript/commit/43be86fb2), replaced header-based detection with `getAutoProxyUrlFromEnvironment` in the backend. If deployed outside Vercel, auto-proxy simply doesn't activate and you configure proxyUrl manually.

Awesome, that sidesteps the concern entirely, thanks!

I think this is looking good to go now, great work Bryce on the original implementation and Railly on picking it up!

@nikosdouvlis Do you have any final concerns?

nikosdouvlis (approved):
LGTM :)

Railly:
hey @nikosdouvlis, could you merge this when you get a chance? I’m not the PR author so I don’t have permissions

PR #8042, opened 2026-03-11
fix(ui): UserProfile should show attributes enabled for sign in

dmoerner:
## Description

Our Dashboard supports instance configurations with email, phone, or username used for sign in, but not used for sign up. This can be useful in cases where accounts are pre-provisioned by the backend, or where users can only add identifiers after sign up. Internally, we represent this as `{"enabled": false, "used_for_first_factor": true}`.

The UserProfile currently guards the display of these identifiers behind `enabled`, which actually maps on to "enabled for sign up". Correct these guards to check for whether the attribute is used for first or second factor sign in as well.

This allows users to both add and also verify these identifications. Email link verification has an additional guard which also requires adjustment.

Fixes USER-4904.

This is an instance with username, email, and phone enabled only for sign in. The user can still add and change them.

CleanShot 2026-06-10 at 17 09 06@2x

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed UserProfile to display sign-in enabled attributes (email, phone, username) correctly, regardless of sign-up settings


packages/ui/src/components/UserProfile/AccountPage.tsx
```diff
@@ -20,9 +21,9 @@ export const AccountPage = withCardStateProvider(() => {
   const { user } = useUser();
   const { shouldAllowIdentificationCreation, immutableAttributes } = useUserProfileContext();
 
-  const showUsername = attributes.username?.enabled;
-  const showEmail = attributes.email_address?.enabled;
-  const showPhone = attributes.phone_number?.enabled;
+  const showUsername = isAttributeAvailable(attributes.username);
```
jacekradko, line 25: What do you think about moving this availability logic to a computed property on AttributeData objects rather than a standalone utility function. `UserSettings` already has getters for pre-computed properties so it's an established pattern and it keeps the responsibility close to the data object

dmoerner: Reasonable suggestion. I did this in [8f1e53b](https://github.com/clerk/javascript/pull/8042/commits/8f1e53b8c771e2b26336a2b90aff44f5897c6ab5) :). Thanks!

packages/ui/src/components/UserProfile/AccountPage.tsx
```diff
@@ -16,14 +16,13 @@ import { UserProfileSection } from './UserProfileSection';
 import { Web3Section } from './Web3Section';
 
 export const AccountPage = withCardStateProvider(() => {
-  const { attributes, social, enterpriseSSO } = useEnvironment().userSettings;
+  const { attributes, availableAttributes, social, enterpriseSSO } = useEnvironment().userSettings;
```
jacekradko: I am looking over this again and I think this actually needs a guard since `availableAttributes` is not going to exist in all v6.x.x versions of clerk-js prior to the one being bumped in this PR. There are no guarantees that this version of `ui` isn't going to run with an older version of `clerk-js` as they are versioned separately. 

I created SDK-115 to track this so we address this soon

packages/ui/src/components/UserProfile/AccountPage.tsx
jacekradko: Coming back to this after a closer look, I have to walk back my suggestion: your original version was right, and moving it onto `UserSettings` creates a cross-bundle hazard I missed. `ui.browser.js` and `clerk.browser.js` load independently (both float on their major tags, and apps can pin `clerkJSVersion`), so a new ui can run against a clerk-js that doesn't have the getter yet. In that pairing `userSettings.availableAttributes` is `undefined`, the `.includes()` calls throw on first render, and since all components share one React root, the whole Clerk UI unmounts. The attribute *data* is safe to read across bundles (it comes from FAPI and older clerk-js copies it as-is); prototype getters are not.

To keep this moving I pushed af31b7f786, which reverts 8f1e53b and the prettier follow-up, restoring your `isAttributeAvailable` helper. All tests pass unchanged and the changeset is right as-is (so disregard my comment there too). Your JSDoc on the helper also answered "available for what?" better than a getter on the public `UserSettings` contract could, and it keeps `password` off that surface, which USER-4904 explicitly warns about.

I filed [SDK-115](https://linear.app/clerk/issue/SDK-115/version-skew-between-clerkui-and-clerk-js-hot-loaded-bundles-have-no) to track the skew problem generally, since this pattern turned out to be very easy to fall into. Sorry for the round trip!

jacekradko (approved):

dmoerner:
Before merging this I want to release `immutable` attribute support in the Dashboard for this sign in configuration. That will allow easy migration for any customers who were using this existing unintended behavior as a kind of gate on users self-serving their own changes.

dmoerner:
We have rolled out improved Dashboard UX to set attributes immutable if enabled for sign in, so this is ready for merge.

PR #8045, opened 2026-03-11
fix(backend): Exclude non-GET requests from handshake eligibility

jacekradko:
## Summary

Fixes [USER-2330](https://linear.app/clerk/issue/USER-2330): POST requests with `sec-fetch-dest: document` (e.g. native HTML form submissions) were incorrectly marked as eligible for handshake, causing a 307 redirect to FAPI's `/v1/client/handshake` endpoint which only accepts GET, resulting in a 405 Method Not Allowed error.

The reported customer issue was triggered via the multi-domain sync flow (satellite app domain sync), not session expiration.

### Changes

- **`authenticateContext.ts`** — Added `method` field populated from the request method
- **`handshake.ts`** — Added early return in `isRequestEligibleForHandshake()` for non-GET requests
- **`request.ts`** — Added `method === 'GET'` guard to three handshake paths:
  1. `isRequestEligibleForMultiDomainSync` (satellite sync)
  2. `shouldForceHandshakeForCrossDomain` (primary cross-origin sync)
- **Tests** — Added unit tests (POST/PUT returning false) and integration tests covering POST requests across all affected flows

### Root cause

`isRequestEligibleForHandshake()` only checked `sec-fetch-dest` and `accept` headers without checking the HTTP method. Native `<form method="POST">` submissions set `sec-fetch-dest: document` just like GET navigations, so they were incorrectly treated as handshake-eligible.

The same gap existed in the multi-domain sync eligibility check and the primary domain cross-origin sync check.

## Test plan

- [x] Unit tests: POST, PUT, and POST-with-accept-header all return `false` from `isRequestEligibleForHandshake()`
- [x] Integration tests: POST requests return signed-out (not handshake) for missing session cookie, satellite domain sync, and stale session token flows
- [x] Integration test: cross-origin POST from satellite to primary returns signed-in (not handshake)
- [x] Existing tests pass (no regressions in GET-based handshake behavior)
- [ ] Manual: verify form submission on a satellite app no longer triggers a 405

## Bug Fixes

* Fixed POST requests with `sec-fetch-dest: document` headers triggering incorrect handshake redirects and 405 errors from FAPI.
* Non-GET requests (including form submissions and PUT requests) are now properly excluded from handshake eligibility.
* Improved multi-domain synchronization logic to only apply to GET requests, ensuring better handling of native form submissions and cross-origin navigation scenarios.


packages/backend/src/tokens/request.ts
```diff
@@ -650,6 +652,7 @@ export const authenticateRequest: AuthenticateRequest = (async (
       // Check for cross-origin requests from satellite domains to primary domain
       const shouldForceHandshakeForCrossDomain =
         !authenticateContext.isSatellite && // We're on primary
+        authenticateContext.method === 'GET' && // Only GET navigations (POST form submissions set sec-fetch-dest: document too)
```
Ephem, line 655: I want to make sure I understand this correctly. My understanding is that in this scenario we _should_ make a handshake, but we _can't_. According to https://github.com/clerk/javascript/pull/6238, we do the `shouldForceHandshakeForCrossDomain` because a satellite might have changed the auth state, so we need to handshake to make sure we have the latest state?

This check bypasses that for POSTs, essentially saying we don't need a handshake for POSTs, which means we might return a different auth state here than the satellite? I'd imagine having the correct auth state would be _more_ important for POSTs, not less though?

Removing this check will make it hit the `isRequestEligibleForHandshake` instead which would fail, essentially saying we needed a handshake, but couldn't do it.

For the specific case we've discussed, the `authenticateContext.method === 'GET'` check would clearly be fine since satellites are not involved, see other comment, what I worry about is the general case.

packages/backend/src/tokens/request.ts
```diff
@@ -650,6 +652,7 @@ export const authenticateRequest: AuthenticateRequest = (async (
       // Check for cross-origin requests from satellite domains to primary domain
       const shouldForceHandshakeForCrossDomain =
```
Ephem, line 653: Not related to the PR, but this check looks off. It looks like it will trigger for _any_ cross origin referrer, regardless of if satellites are involved or not, which does not seem like the intent from the original PR or the comments etc. I wonder if we could tighten that up somehow? 🤔

brkalow: @Ephem this is correct, we currently have no way to know if an app is using satellites here when on a primary (non-satellite) domain, so this is a best effort check.

packages/backend/src/tokens/request.ts
jacekradko: Right. This is a case where a handshake would be appropriate if it was a GET navigation request, but it's a POST instead.  

So yeah, this essentially makes the POST request get handled by the app server (regardless of auth state) instead of forcing a redirect to a handshake endpoint from the Clerk middleware.

PR #8053, opened 2026-03-12
test(e2e): add test for `getToken` with custom JWT templates

tmilewski:
## Description

- Adds a Playwright integration test verifying `getToken({ template })` returns valid custom JWT tokens via a Next.js App Router API route
- Creates a temporary JWT template with a custom claim, signs in a user, and asserts `getToken` return tokens with the expected `test_claim` payload

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] other: Test

## Summary by CodeRabbit

* **Tests**
  * Added integration test for custom JWT template validation.

* **Chores**
  * Added changeset metadata file.


integration/tests/debug-and-custom-jwt.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    const res2 = await fetchToken(true);
+    expect(res2.status()).toBe(200);
+    const body2 = await res2.json();
+    expect(body2.userId).toBeTruthy();
+    expect(body2.sessionId).toBeTruthy();
+    expect(body2.customToken).toBeTruthy();
+
+    const payload2 = JSON.parse(atob(body2.customToken.split('.')[1]));
+    expect(payload2.test_claim).toBe('hello_from_e2e');
+    expect(payload2.iat).toBeGreaterThanOrEqual(payload1.iat);
```
jacekradko: Is this a strong enough assertion to know that we got a freshly minted token with skipCache?

wobsoriano (approved):

integration/tests/debug-and-custom-jwt.test.ts
wobsoriano: approved, but defer to jacek

jacekradko (approved):
Overall looks good, even though I think we can iterate on this a little bit. For one, we should prove caching works — call fetchToken(false) twice, assert tokens are equal, then fetchToken(true) and assert it differs. The `iat` assertion can be stronger to not give false positives.

tmilewski:
> Overall looks good, even though I think we can iterate on this a little bit. For one, we should prove caching works — call fetchToken(false) twice, assert tokens are equal, then fetchToken(true) and assert it differs. The `iat` assertion can be stronger to not give false positives.

@jacekradko I think I accidentally removed my initial call. You're absolutely right.

tmilewski:
> Overall looks good, even though I think we can iterate on this a little bit. For one, we should prove caching works — call fetchToken(false) twice, assert tokens are equal, then fetchToken(true) and assert it differs. The `iat` assertion can be stronger to not give false positives.

@jacekradko So, `getToken` with a custom template never caches. As such, I updated the test.

PR #8066, opened 2026-03-13
fix(expo): prevent crash when importing @clerk/expo in web environments

chriscanin:
## Summary
- Fixes `@clerk/expo` crashing with `Cannot read properties of undefined (reading 'get')` when used in Expo web environments
- The crash occurs because `TurboModuleRegistry` from `react-native` is `undefined` on web (react-native-web), and `NativeClerkModule.ts` calls `TurboModuleRegistry.get('ClerkExpo')` at module evaluation time — before any runtime guards can protect against it
- Adds `.web.ts` variants for `NativeClerkModule` (exports `null`) and polyfills (empty — browsers have native URL and base64), following the existing pattern used by `singleton.web.ts`

**Error:** `Uncaught Error: Cannot read properties of undefined (reading 'get')` at `NativeClerkModule.js`

## Test plan
- [ ] Verify `@clerk/expo` no longer crashes when bundled for web (Expo web / react-native-web)
- [ ] Verify native (iOS/Android) functionality is unaffected — `TurboModuleRegistry.get()` still runs on native via the original `.ts` file
- [ ] Verify hooks (`useSignIn`, `useSSO`, `useAuth`) work correctly on web

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed web crashes in the Expo package by improving native module import handling and providing proper fallback behavior when modules are unavailable on web platforms.

### DIFF
--- a/packages/expo/src/polyfills/index.web.ts
+++ b/packages/expo/src/polyfills/index.web.ts
@@ -1,1 +1,0 @@
-// No polyfills needed on web - browsers have native URL and base64 support


chriscanin:
Screenshot 2026-03-13 at 9 07 32 AM
Screenshot 2026-03-13 at 9 10 34 AM

I was able to reproduce Yu's error, and this is a screenshot of before and after from this fix.

PR #8084, opened 2026-03-15
fix(shared): avoid false offline detection in React Native

### LINKED ISSUE #8083: @clerk/expo: getToken({ skipCache: true }) fails on Expo while cached token fetch succeeds
## Summary

In an Expo repro app using `@clerk/expo` with Convex, cached token fetch works, but fresh token fetch fails.

Specifically:

- cached token fetch succeeds
- fresh token fetch fails with:
  - `Clerk: Network request failed while offline. The browser appears to be disconnected.`
  - `(code="clerk_offline")`

This leaves the app signed in at the Clerk level, but downstream Convex auth never completes.

## Repro repo

https://github.com/eliotgevers/convex-clerk-expo-token-refresh-repro

## Environment

- `@clerk/expo`: `3.1.2`
- `convex`: `1.33.1`
- Expo SDK: `55`
- React Native: `0.83.2`
- Reproduced in Expo iOS simulator

## Reproduction steps

1. Clone the repro repo.
2. Fill in `.env` with a valid Clerk publishable key and Convex URL.
3. Start the app.
4. Sign in with an existing user.
5. Tap `Probe cached token`.
6. Tap `Probe fresh token`.

## Expected

Both cached and fresh token fetches succeed.

## Actual

Cached token succeeds, but fresh token fetch fails with `clerk_offline`.

## Example result

```text
Clerk signed in: true
Convex loading: false
Convex authenticated: false
WebSocket connected: true
Last cached token ok: 2:51:44 PM
Last fresh token failed: 2:42:35 PM

Recent events:
- Fresh token failed: Clerk: Network request failed while offline. The browser appears to be disconnected.
  (code="clerk_offline")
```

## Notes

- The repro intentionally strips away app-specific backend bootstrapping so the failure stays focused on Clerk token fetch behavior.
- If Clerk returns `needs_client_trust`, I disable Client Trust for the test app or use a trusted device. That is separate from the bug being reported here.

eliotgevers:
## Summary

This fixes false offline detection in React Native / Expo when `navigator.onLine` is not implemented.

`session.getToken({ skipCache: true })` goes through the forced-refresh path, which checks the shared online helper before fetching a fresh token. In React Native, `window` and `navigator` exist, but `navigator.onLine` does not, so the current logic treats the runtime as offline and throws `ClerkOfflineError`.

This change treats React Native, and runtimes without a boolean `navigator.onLine`, as online/unknown rather than offline. Actual network failures still surface through the normal fetch path.

## Changes

- return `true` from `isBrowserOnline()` for React Native
- return `true` when `navigator.onLine` is not implemented as a boolean
- add a focused regression test in `@clerk/shared`

## Validation

- `pnpm --filter @clerk/shared test -- src/__tests__/browser.spec.ts`
- manually patched the equivalent native bundle in an Expo repro app and verified that cached token fetch still succeeds and fresh token fetch succeeds after the fix

Fixes #8083

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved online/offline detection for React Native and other environments lacking standard status indicators to reduce false "offline" reports.

* **Tests**
  * Added test coverage for online status detection in React Native-like scenarios to prevent regressions.


packages/shared/src/browser.ts
```diff
@@ -73,6 +73,16 @@ export function isBrowserOnline(): boolean {
...
 
+  // React Native / Expo defines a Navigator object but does not implement
+  // the browser-only navigator.onLine API.
+  if (navigator.product === 'ReactNative') {
+    return true;
+  }
+
+  if (typeof navigator.onLine !== 'boolean') {
+    return true;
+  }
```
brkalow: I think we can simplify this change to default to `true` on L73 above. This gives us a safe default and avoids explicit knowledge of the platform / runtime here.

chriscanin: Yes, that will work! I have updated this.

PR #8085, opened 2026-03-16
fix(expo): pass native proxy config to clerk

brkalow:
## Summary
- pass `proxyUrl` and `domain` through Expo's native Clerk singleton
- recreate the native singleton when that config changes
- add regression coverage for native singleton proxy/domain wiring

## Test plan
- `cd packages/expo && pnpm test -- src/provider/singleton/__tests__/createClerkInstance.test.ts`

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * ClerkProvider now accepts optional proxyUrl and domain and forwards them to the Expo Clerk runtime to support custom endpoints.

* **Tests**
  * Added unit tests verifying proxyUrl, domain, and publishableKey are forwarded correctly and that instances are reused or recreated when these values change.

* **Chores**
  * Added a changeset entry for a patch release documenting this behavioral update.

### DIFF
--- a/packages/expo/src/provider/singleton/createClerkInstance.ts
+++ b/packages/expo/src/provider/singleton/createClerkInstance.ts
@@ -42,6 +42,13 @@ type ResolvedClerkRuntimeOptions = Omit<ClerkRuntimeOptions, 'publishableKey'> &
   publishableKey: string;
 };
 
+function hasOwnOption<Key extends keyof ClerkRuntimeOptions>(
+  options: ClerkRuntimeOptions | undefined,
+  key: Key,
+): options is ClerkRuntimeOptions & Required<Pick<ClerkRuntimeOptions, Key>> {
+  return !!options && Object.prototype.hasOwnProperty.call(options, key);
+}
+
 let __internal_clerk: HeadlessBrowserClerk | BrowserClerk | undefined;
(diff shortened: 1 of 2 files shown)


packages/expo/src/provider/singleton/createClerkInstance.ts
```diff
@@ -37,31 +37,77 @@ type FapiResponse = Response & {
...
+  options: ResolvedClerkRuntimeOptions;
+} {
+  const hasNextProxyUrl = hasOwnOption(nextOptions, 'proxyUrl');
+  const hasNextDomain = hasOwnOption(nextOptions, 'domain');
+  const hasKeyChanged =
+    !!currentOptions &&
+    typeof nextOptions?.publishableKey !== 'undefined' &&
+    nextOptions.publishableKey !== currentOptions.publishableKey;
+  const hasProxyChanged = !!currentOptions && hasNextProxyUrl && nextOptions.proxyUrl !== currentOptions.proxyUrl;
+  const hasDomainChanged = !!currentOptions && hasNextDomain && nextOptions.domain !== currentOptions.domain;
```
brkalow, line 74: I can't really think of a scenario in which these would change at runtime, but good to support I suppose along with `publishableKey` as they all determine which instance we're interacting with.

PR #8086, opened 2026-03-16
fix(shared): correct isValidBrowser/isBrowserOnline for React Native

chriscanin:
## Summary

- `isValidBrowser()` and `isBrowserOnline()` in `@clerk/shared` were returning `false` when `window`/`navigator` are undefined (React Native environments)
- Core-3's token fetching checks these functions before making FAPI calls — returning `false` caused `getToken({ template: '...' })` to silently fail on React Native while `getToken()` (which reads from local cache) continued to work
- Fix: return `true` in non-browser environments since React Native has its own networking layer and should not be treated as a bot or offline

## Reported by
Bledar via Discord — `getToken({ template: 'one_hour_token' })` returns nothing on Expo after upgrading to latest Clerk

## Test plan
- [x] Updated existing `browser.spec.ts` test to verify `isValidBrowser()` returns `true` when window is undefined
- [x] All 983 tests in `@clerk/shared` pass
- [ ] Verify `getToken({ template })` works on React Native after fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed token retrieval with templates in React Native and other non-browser environments. Non-browser runtimes are now correctly recognized as valid execution contexts, ensuring template-based token operations function properly across React Native, server-side rendering, and similar platforms. This resolves compatibility issues that previously prevented proper authentication token management in these environments.

### DIFF
--- a/.changeset/fix-gettoken-template-rn.md
+++ b/.changeset/fix-gettoken-template-rn.md
@@ -1,0 +1,5 @@
+---
+"@clerk/shared": patch
+---
+
+Fix `getToken({ template })` not working in React Native by correcting `isValidBrowser()` and `isBrowserOnline()` to return `true` in non-browser environments instead of `false`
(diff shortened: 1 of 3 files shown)


chriscanin:
This PR is being closed in favor of https://github.com/clerk/javascript/pull/8084.

PR #8087, opened 2026-03-16
fix(expo): synchronous QueryClient for React Native

chriscanin:
## Summary

- `clerk-js` lazily loads `QueryClient` from `@tanstack/query-core` via rspack's dynamic `import('./query-core')` code-splitting
- On React Native, Metro bundles everything into a single file — rspack's chunk loading (`l.e("675")`) never resolves, so `clerk.__internal_queryClient` stays `undefined`
- The `useClerkQueryClient` hook falls back to a no-op mock proxy when undefined, causing `useOrganizationList`, `useSessionList`, and other query-based hooks to silently return empty data
- Fix: `Object.defineProperty` override in `createClerkInstance.ts` that synchronously creates a `QueryClient` via `require('@tanstack/query-core')` on first access, returning the expected `{ __tag: 'clerk-rq-client', client }` shape

## Reported by
Donovan (internal) — `useOrganizationList` returns empty data on React Native

## Test plan
- [x] `@clerk/expo` builds successfully with the change
- [x] TypeScript declarations pass
- [ ] Verify `useOrganizationList` returns data on React Native after fix
- [ ] Verify other query-based hooks (`useSessionList`, etc.) work correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `useOrganizationList` and other query-based hooks in React Native environments that were incorrectly returning empty data.

* **Chores**
  * Updated build configuration for React Native applications.


jacekradko:
### Code review

Found 1 issue:

1. `@tanstack/query-core` is not declared as a dependency in `packages/expo/package.json`. The `require('@tanstack/query-core')` call at line 122 will fail at runtime if the package isn't resolvable from the `@clerk/expo` package context. While it is a transitive dependency via `@clerk/clerk-js`, strict package managers (pnpm) do not hoist transitive dependencies, so this `require()` will throw `Cannot find module '@tanstack/query-core'` for consumers. It needs to be added as a `dependency` or `peerDependency` in `packages/expo/package.json`.

https://github.com/clerk/javascript/blob/944baf9345b3409a1d566f51c2e90ce1d89b5114/packages/expo/src/provider/singleton/createClerkInstance.ts#L121-L123

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>


@chriscanin I think it makes sense to add `@tanstack/query-core` as a `peerDependency` of `@clerk/expo`. Otherwise we could expose it as a pass-through in `@clerk/clerk-js` so it's part of the explicit contract

chriscanin:
> ### Code review
> Found 1 issue:
> 
> 1. `@tanstack/query-core` is not declared as a dependency in `packages/expo/package.json`. The `require('@tanstack/query-core')` call at line 122 will fail at runtime if the package isn't resolvable from the `@clerk/expo` package context. While it is a transitive dependency via `@clerk/clerk-js`, strict package managers (pnpm) do not hoist transitive dependencies, so this `require()` will throw `Cannot find module '@tanstack/query-core'` for consumers. It needs to be added as a `dependency` or `peerDependency` in `packages/expo/package.json`.
> 
> https://github.com/clerk/javascript/blob/944baf9345b3409a1d566f51c2e90ce1d89b5114/packages/expo/src/provider/singleton/createClerkInstance.ts#L121-L123
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.
> 
> @chriscanin I think it makes sense to add `@tanstack/query-core` as a `peerDependency` of `@clerk/expo`. Otherwise we could expose it as a pass-through in `@clerk/clerk-js` so it's part of the explicit contract


@jacekradko I have added @tanstack/query-core as a direct dependency in @clerk/expo's package.json and pinned to the same version @clerk/clerk-js uses (5.90.16). This ensures the require() call resolves correctly with strict package managers like pnpm.

chriscanin:
!snapshot

chriscanin:
@jacekradko , this has been tested locally, and is confirmed to fix.
Before snapshot test:
image

After snapshot test:
image

brkalow:
@chriscanin I think we should address this when bundling clerk-js, instead of creating a separate dependency on the package in `@clerk/expo`.

@jacekradko what do you think?

chriscanin:
> @chriscanin I think we should address this when bundling clerk-js, instead of creating a separate dependency on the package in `@clerk/expo`.
> 
> @jacekradko what do you think?

Hi @brkalow @jacekradko , I spent some time yesterday looking into this and claude-ing, as this issue is effecting users who are attempting to use organizations, and are currently blocked. The fix described below would resolve this at a bundler level instead of creating a separate dependency on the package in @clerk/expo. This should bake @tanstack/query-core into the bundle, rather than trying to resolve it during async chunk loading. 

In the `clerkNative` config block in `rspack.config.js`, we can add `dynamicImportMode: 'eager'` to the existing override object:

```javascript
const clerkNative = merge(
  entryForVariant(variants.clerkNative),
  common({ mode, variant: variants.clerkNative }),
  commonForProd(),
  commonForProdChunked(),
  {
    output: {
      publicPath: '',
    },
    module: {
      parser: {
        javascript: {
          dynamicImportMode: 'eager',
        },
      },
    },
    optimization: {
      splitChunks: false,
    },
  },
);
```

This makes ALL dynamic `import()` calls in the native build resolve eagerly where modules get inlined into the bundle and the returned Promise resolves immediately. Combined with the existing `splitChunks: false`, the native build becomes fully self-contained.

`module.parser.javascript.dynamicImportMode: 'eager'` is the config-level equivalent of putting `/* webpackMode: "eager" */` on every dynamic import in the codebase.

The webpack docs page for [module.parser.javascript.dynamicImportMode](https://webpack.js.org/configuration/module/#moduleparserjavascriptdynamicimportmode) only lists it as an option without explaining what it does. The actual explanation lives on the **Module Methods** page under the `import()` magic comments:

- https://webpack.js.org/api/module-methods/#magic-comments explains that `eager` "generates no extra chunk, all modules are included in the current chunk and no additional network requests are made. A Promise is still returned but is already resolved."

Built `clerk-js` with this change, published to verdaccio, installed into the NativeComponentQuickstart app, and confirmed the native `<AuthView>` renders correctly on iOS simulator. (as well as the JS only flows).

NOTE: This applies to ALL dynamic imports in the native build, not just `query-core`. Other lazy-loaded chunks (`zxcvbn` for password strength, `stripe`, etc.) would also get inlined. That's probably fine for the native build since `splitChunks: false` already implies everything should be in one bundle but worth flagging.

Thoughts on this?

jacekradko:
@chriscanin @brkalow What do you think of this approach: https://github.com/clerk/javascript/pull/8170  ?

Essentially curating the native build entry point to statically import the query client

jacekradko (approved):
Looks good! Thanks for chatting through this

chriscanin:
@jacekradko Doing some quick testing, putting do not merge flag for the moment.

chriscanin:
So this is all working on ios and android but one caveat @jacekradko is that our bundle size has gone from 216kb to 1.3mb. Not sure if that is acceptable. Let me know.

wobsoriano:
Hi all 👋🏼 

Tested this with latest `@clerk/expo` version and can see `useOrganizationList()` rendering properly:

simulator_screenshot_E1A9D6EF-7E5F-4CDF-9F8B-FAFD88C855F1

Closing for now

PR #8089, opened 2026-03-16
feat(ui): Disable invite button when over maxAllowedMemberships

dstaley:
## Description

This PR updates the rendering logic for the `Invite` button in OrganizationProfile to respond to organization memberships and the `maxAllowedMemberships` field. Now, the invite button will be disabled (with a tooltip) if the sum of the organization's members and pending invitations is equal to or above the organization's `maxAllowedMemberships`. This PR also introduces a localizable value for the error code that is returned from the API when a user attempts to invite members that would exceed their limit. This error message is also used for the tooltip when the invite button is disabled.

CleanShot 2026-03-16 at 11 38 00@2x

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/ui/src/components/OrganizationProfile/MembersActions.tsx
+++ b/packages/ui/src/components/OrganizationProfile/MembersActions.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useMemo, type ReactNode } from 'react';
 import { useOrganization } from '@clerk/shared/react';
 import { Animated } from '@/ui/elements/Animated';
 import { Tooltip } from '@/ui/elements/Tooltip';
@@ -21,6 +21,7 @@ export const MembersActionsRow = ({ actionSlot }: MembersActionsRowProps) => {
       return false;
     }
 
+    // A value of 0 means unlimited memberships, thus the organization is always below the limit
     if (organization.maxAllowedMemberships === 0) {
       return true;
     }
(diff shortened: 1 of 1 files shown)


packages/localizations/src/en-US.ts
```diff
@@ -1037,7 +1037,8 @@ export const enUS: LocalizationResource = {
     organization_domain_blocked: undefined,
     organization_domain_common: undefined,
     organization_domain_exists_for_enterprise_connection: undefined,
-    organization_membership_quota_exceeded: undefined,
+    organization_membership_quota_exceeded:
+      'You have reached your limit of organisation memberships, including outstanding invitations.',
```
l-armstrong: maybe we should use organization here instead of organisation for consistency.

packages/ui/src/components/OrganizationProfile/MembersActions.tsx
```diff
@@ -11,6 +14,47 @@ type MembersActionsRowProps = {
...
 export const MembersActionsRow = ({ actionSlot }: MembersActionsRowProps) => {
   const canManageMemberships = useProtect({ permission: 'org:sys_memberships:manage' });
+  const { organization } = useOrganization();
+
+  const isBelowLimit = useMemo(() => {
+    if (!organization) {
+      return false;
+    }
+
+    if (organization.maxAllowedMemberships === 0) {
```
l-armstrong, line 25: maybe add a comment here explaining that 0 means unlimited. for future readers of the code.

packages/localizations/src/en-US.ts
dstaley: we absolutely should! I didn't even realize British English had a different spelling. Good catch!

packages/ui/src/components/OrganizationProfile/MembersActions.tsx
dstaley: 🫡

PR #8091, opened 2026-03-16
feat(ui,clerk-js,shared): Allow to link external accounts to enterprise accounts

NicolasLopes7:
## Description

This PR adds a new feature for the enterprise accounts section in `UserProfile` - it allows to connect to enterprise accounts based on existing enterprise connections for the current user's active organization. 

This is only possible if the enterprise connection has an `allowOrganizationAccountLinking` flag.

https://github.com/user-attachments/assets/4c9dc467-0412-4491-9a27-b4d122bfcb38

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Release Notes

* **New Features**
  * Users can now link external accounts to enterprise accounts through UserProfile
  * Enterprise Accounts section includes a new "Connect account" button for managing account linkage

* **Localization**
  * Added translations for the Enterprise Accounts interface across 40+ languages

### DIFF
--- a/packages/shared/src/react/hooks/useUserEnterpriseConnections.shared.ts
+++ b/packages/shared/src/react/hooks/useUserEnterpriseConnections.shared.ts
@@ -6,18 +6,22 @@ import { createCacheKeys } from './createCacheKeys';
 /**
  * @internal
  */
-export function useUserEnterpriseConnectionsCacheKeys(params: { userId: string | null }) {
-  const { userId } = params;
+export function useUserEnterpriseConnectionsCacheKeys(params: {
+  userId: string | null;
+  withOrganizationAccountLinking?: boolean;
+}) {
+  const { userId, withOrganizationAccountLinking = false } = params;
   return useMemo(() => {
     return createCacheKeys({
       stablePrefix: INTERNAL_STABLE_KEYS.USER_ENTERPRISE_CONNECTIONS_KEY,
       authenticated: Boolean(userId),
       tracked: {
         userId: userId ?? null,
+        withOrganizationAccountLinking,
       },
       untracked: {
         args: {},
       },
     });
-  }, [userId]);
+  }, [userId, withOrganizationAccountLinking]);
 }
--- a/packages/shared/src/react/hooks/useUserEnterpriseConnections.tsx
+++ b/packages/shared/src/react/hooks/useUserEnterpriseConnections.tsx
@@ -33,6 +33,7 @@ function useUserEnterpriseConnections(
 
   const { queryKey, stableKey, authenticated } = useUserEnterpriseConnectionsCacheKeys({
     userId: user?.id ?? null,
+    withOrganizationAccountLinking,
   });
 
   const queryEnabled = enabled && clerk.loaded && Boolean(user);


packages/localizations/src/en-US.ts
```diff
@@ -1188,6 +1188,15 @@ export const enUS: LocalizationResource = {
       successMessage: 'The provider has been added to your account',
       title: 'Add connected account',
     },
+    enterpriseAccountPage: {
```
LauraBeatris: Can we generate the translations on all localizations file via LLM prompt?

packages/ui/src/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -1,19 +1,159 @@
...
+    <ProfileSection.ActionMenuItem
+      key={connection.id}
+      id={connection.id}
+      onClick={connect}
+      isDisabled={card.isLoading}
+      variant='ghost'
+      isLoading={card.loadingMetadata === loadingKey}
+      focusRing={false}
+      closeAfterClick={false}
+      localizationKey={localizationKeys('userProfile.connectedAccountPage.socialButtonsBlockButton', {
```
LauraBeatris, line 75:
```suggestion
      localizationKey={localizationKeys('userProfile.enterpriseAccountPage.socialButtonsBlockButton', {
```

Let's use an specific translation key for the enterprise accounts section

packages/ui/src/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -1,19 +1,159 @@
...
+          alt={`Connect ${connection.name} account`}
+          elementDescriptor={descriptors.providerIcon}
+          elementId={descriptors.providerIcon.setId(providerWithoutPrefix)}
+        />
+      }
+    />
+  );
+};
+
+const AddEnterpriseAccount = ({
```
LauraBeatris, line 98:
```suggestion
const ConnectEnterpriseAccounts = ({
```

Since this component receives a list of enterprise accounts + it has a inner menu to connect to it

packages/ui/src/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -1,19 +1,159 @@
...
+export const EnterpriseAccountsSection = withCardStateProvider(() => {
+  const { user } = useUser();
+  const card = useCardState();
+  const [enterpriseConnections, setEnterpriseConnections] = useState<EnterpriseAccountConnectionResource[]>([]);
+  const [actionValue, setActionValue] = useState<string | null>(null);
+
+  useEffect(() => {
+    user?.getEnterpriseConnections?.().then(connections => {
+      setEnterpriseConnections(connections.filter(c => c.allowAccountLinking));
+    });
```
LauraBeatris: We should use our data fetching hooks for it

packages/ui/src/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -1,19 +1,159 @@
...
+  }, [user]);
 
   const activeEnterpriseAccounts = user?.enterpriseAccounts.filter(
     ({ enterpriseConnection }) => enterpriseConnection?.active,
   );
 
-  if (!activeEnterpriseAccounts?.length) {
+  const hasActiveAccounts = Boolean(activeEnterpriseAccounts?.length);
```
LauraBeatris: We can make this simpler by using `some` above on `activeEnterpriseAccounts`

packages/ui/src/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -1,19 +1,159 @@
...
+  const hasAvailableConnections = enterpriseConnections.length > 0;
```
LauraBeatris:
```suggestion
  const hasAvailableConnectionsForAccountLinking = enterpriseConnectionsForAccountLinking.length > 0;
```

NicolasLopes7:
Todo: filter enterprise connections that the user already have an account with.

Also only display the "connect enterprise account" button if there's at least one available connection to connect

packages/ui/src/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -1,19 +1,140 @@
...
+const EnterpriseConnectMenuButton = (props: { connection: EnterpriseAccountConnectionResource }) => {
+  const { connection } = props;
+  const card = useCardState();
   const { user } = useUser();
+  const { componentName, mode } = useUserProfileContext();
+  const isModal = mode === 'modal';
+  const loadingKey = `enterprise_${connection.id}`;
+
+  const createExternalAccount = useReverification(() => {
+    const redirectUrl = isModal ? appendModalState({ url: window.location.href, componentName }) : window.location.href;
```
iagodahlem, line 32: I wonder if we shouldn't abstract this logic somehow into a hook, I would personally forget all the time to check for the modal mode to append the modal state 😅

packages/ui/src/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -1,19 +1,140 @@
...
+
+    return user?.createExternalAccount({
+      enterpriseConnectionId: connection.id,
+      redirectUrl,
+    });
```
iagodahlem, line 37: ❓ A few questions here around the data fetching:
- When we use the `user.createExternalAccount` is that using the User resource directly?
- Does that handles invalidating the `/me/enterprise_connections` query? 
- I think in this case it doesn't matter because we always redirect to the IdP to sign in and connect the account right?
- In this case, would be worth making this call a hook in the react package? I wonder where do we draw the line between create a custom hook or the resource directly.

packages/ui/src/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -1,19 +1,140 @@
...
+    if (!user) {
+      return;
+    }
+
+    card.setLoading(loadingKey);
+
+    return createExternalAccount()
+      .then(res => {
+        if (res?.verification?.externalVerificationRedirectURL) {
+          void sleep(2000).then(() => card.setIdle(loadingKey));
```
iagodahlem, line 50: Is the sleep just cosmetic to make the windowNavigate smoother?

packages/ui/src/components/UserProfile/EnterpriseAccountsSection.tsx
```diff
@@ -1,19 +1,140 @@
...
     ({ enterpriseConnection }) => enterpriseConnection?.active,
   );
 
-  if (!activeEnterpriseAccounts?.length) {
+  const { data: userEnterpriseConnections = [] } = __internal_useUserEnterpriseConnections({
+    withOrganizationAccountLinking: true,
+  });
+  const linkableEnterpriseConnections = userEnterpriseConnections.filter(
+    c => c.allowOrganizationAccountLinking && !activeEnterpriseAccounts?.some(a => a.enterpriseConnectionId === c.id),
+  );
```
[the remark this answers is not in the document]
iagodahlem: This makes sense. I think it's worth fixing.

packages/ui/src/components/UserProfile/EnterpriseAccountsSection.tsx
LauraBeatris: > I think in this case it doesn't matter because we always redirect to the IdP to sign in and connect the account right?

Correct, so the IdP would redirect back to the app, with the `UserProfile` opened, where it'll trigger `/me/enterprise_connections` again

LauraBeatris: > In this case, would be worth making this call a hook in the react package? I wonder where do we draw the line between create a custom hook or the resource directly.

Since this logic is specific for the enterprise account section, so makes sense to have it in there, and only rely on the mutation + query

We could extract this to a hook within the same module at least for semantics

packages/ui/src/components/UserProfile/EnterpriseAccountsSection.tsx
LauraBeatris: Mostly presentational, I've kept it consistent with the same logic we have to connect to external accounts 

If navigation is slow, loading eventually clears after 2s instead of staying stuck forever

PR #8095, opened 2026-03-17
fix(clerk-js): Respect proxyUrl and domain in non-browser environments

brkalow:
## Description

It's possible for non-browser environments to provide a `domain` or `proxyUrl`, but `clerk-js` currently ignores the provided values when not in a browser. This PR updates `clerk-js` so the provided value is now respected in non-browser environments.

The only caveat is that we cannot accept a function for the value in expo, as we are unable to provide the current base URL (`window.location.href`).

fixes AIE-644

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Support for providing proxy URLs and custom domains in non-browser environments.

* **Bug Fixes**
  * Native/runtime validation now enforces absolute HTTP(S) proxy URLs and rejects unsupported function-valued configurations, emitting warnings and graceful fallbacks.

* **Tests**
  * Added and expanded tests covering non-browser/native behavior for proxy URL and domain resolution.

* **Chores**
  * Bumped package patch versions via a changeset.

### DIFF
--- a/integration/templates/next-app-router/next.config.js
+++ b/integration/templates/next-app-router/next.config.js
@@ -1,6 +1,4 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {
-  outputFileTracingRoot: '/',
-};
+const nextConfig = {};
 
 module.exports = nextConfig;
--- a/integration/templates/next-cache-components/next.config.js
+++ b/integration/templates/next-cache-components/next.config.js
@@ -1,6 +1,5 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
-  outputFileTracingRoot: '/',
   cacheComponents: true,
 };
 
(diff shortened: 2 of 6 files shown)


packages/expo/src/utils/errors.ts
```diff
@@ -1,3 +1,23 @@
...
+  }
+
+  if (isNative()) {
+    if (typeof proxyUrl !== 'string') {
+      return errorThrower.throw('`proxyUrl` must be a string in non-browser environments.');
+    }
+
+    if (!isHttpOrHttps(proxyUrl)) {
+      errorThrower.throw('`proxyUrl` must be an absolute URL in non-browser environments.');
+    }
```
jacekradko: [nit] Just from a quick look at this, it feels weird to return the errorThrower.throw even though it presumably throws in one branch, and not in the other

brkalow: true, this was to appease TS on L19 to narrow the type. Let me see if I can adjust

jacekradko: I figured it's static analysis related

PR #8096, opened 2026-03-17
feat(clerk-js): Send previous token and force_origin on /tokens requests

nikosdouvlis:
## Why

Session Minter needs the previous session JWT to clone claims at the edge without hitting the DB. It also needs a way for `skipCache` callers to bypass the edge minter and go straight to origin.

## What

- Add `oiat` (original_issued_at) to `JwtHeader` type, marked `@internal`. Session Minter uses this to track when claims were last assembled from the DB. Lives in the JWT header, not payload.
- Send `token` (previous session JWT) in the POST body on non-template `/tokens` requests. Uses conditional spread so the key is absent (not `token=`) when there's no previous token.
- Send `force_origin=true` in the query string when `skipCache: true`. Tells FAPI Proxy to route to origin instead of Session Minter.

Both params are currently ignored by the backend and will be wired up in PLAT-2471 and PLAT-2472.

## Test plan

- [ ] Verify `token` appears in POST body when a previous token exists
- [ ] Verify `token` key is absent on first mint (no previous token)
- [ ] Verify `token` is not sent for template token requests
- [ ] Verify `force_origin=true` in query string when `skipCache: true`
- [ ] Verify `force_origin` absent when `skipCache: false`
- [ ] Existing token refresh flows still work (no behavioral change yet)

### DIFF
--- a/.changeset/session-minter-sdk-params.md
+++ b/.changeset/session-minter-sdk-params.md
@@ -1,0 +1,6 @@
+---
+'@clerk/shared': patch
+'@clerk/clerk-js': patch
+---
+
+Add `oiat` field to `JwtHeader` type. Send previous session token and `force_origin` param on `/tokens` requests to support Session Minter edge token minting.


nikosdouvlis:
Split into #8105 (send token) and #8106 (force_origin).

PR #8097, opened 2026-03-17
feat(clerk-js): Monotonic token replacement based on oiat

nikosdouvlis:
## Why

With Session Minter, edge-minted tokens can have fresh `iat` (just minted) but stale claims (copied from an old parent). In multi-tab BroadcastChannel scenarios, a background tab's stale edge token can clobber a fresher DB-minted token in another tab's cache because the old broadcast guard compared `iat` only, which doesn't reflect claim freshness for edge-minted tokens.

## What

Introduces `oiat` (origin-issued-at, JWT header) as the claim freshness metric for the cross-tab broadcast comparison:
- Token with `oiat`: oiat = when claims were last read from DB
- Token without `oiat`: pre-feature legacy token, by definition staler than any oiat-bearing token

### Decision table

`pickFreshestJwt(existing, incoming)` returns whichever side wins. The broadcast handler uses `=== existing` to detect "skip this broadcast".

| # | ex.oiat | in.oiat | comparison | winner | why |
|---|---------|---------|------------|--------|-----|
| A1 | 100 | 90 | ex.oiat > in.oiat | existing | incoming claims older |
| A2 | 90 | 100 | ex.oiat < in.oiat | incoming | incoming claims newer |
| A3 | 100 | 100 | equal, ex.iat > in.iat | existing | same claims, existing minted later |
| A4 | 100 | 100 | equal, ex.iat < in.iat | incoming | same claims, incoming minted later |
| A5 | 100 | 100 | equal, ex.iat == in.iat | incoming | full tie, let incoming through (other claims may differ) |
| B1 | 100 | null | incoming has no oiat | existing | legacy = pre-feature stale |
| B2 | null | 100 | existing has no oiat | incoming | legacy = pre-feature stale |
| B3 | null | null | both have no oiat | incoming | cannot rank, safe default |

A1-A5 is the main path under universal oiat. B1-B3 is the legacy safety net for pre-rollout tokens. On a full tie (A5), incoming wins rather than existing: two tokens with identical oiat+iat may still differ in other claims (azp, org_id, etc.) added in a token-format rollout, so we only suppress when existing is strictly fresher.

### Guard site

The shared comparator (`tokenFreshness.ts`) is invoked at one site:

- `tokenCache.ts handleBroadcastMessage` - replaces the old iat comparison on cross-tab broadcasts

The Session resource and AuthCookieService originally also added guards but those caused regressions on the auth flow (suppressing legitimate token-update events that AuthCookieService needs to write the cookie) and have been removed. The broadcast handler is the only site where a stale edge-minted token could realistically race with a fresher one without other in-flight protections.

## Test plan

- [ ] Broadcast: edge token with fresh iat but stale oiat is rejected
- [ ] Broadcast: oiat-newer token replaces oiat-older token in the same tab
- [ ] First mint with no existing token passes unconditionally
- [ ] Comparator: all 8 decision table rows


packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -518,6 +523,10 @@ export class Session extends BaseResource implements SessionResource {
       return;
     }
 
+    if (this.lastActiveToken && pickFreshestJwt(this.lastActiveToken, token) === this.lastActiveToken) {
```
bratsos: The suppression test at Session.test.ts:103 only proves the guard can suppress, not that it does so for the right reason.

And every test that asserts `dispatch=2` (lines 186, 242, 60) constructs the session without last_active_token, so the `this.lastActiveToken && …` short-circuit means the comparator is never actually called.

We should probably add a test where `lastActiveToken` is set with stale oiat, a fresher-`oiat` token arrives via cache (broadcast from another tab) or fetch, and assert `token:update` fires with the fresher token

bratsos (approved):
Minor comment re:tests, LGTM!

PR #8098, opened 2026-03-17
fix(clerk-js): Fix token cache refresh timer leak

jacekradko:
## Summary

- Fix orphaned refresh timer accumulation in `SessionTokenCache` that caused token refresh requests to fire much earlier than expected
- When `set()` is called multiple times for the same cache key (from `_updateClient` hydration AND `#refreshTokenInBackground`), old timers were never cleared — each leaked timer independently fired `onRefresh`, accelerating the effective polling rate
- The fix clears existing timers at the start of `setInternal` before creating the new cache entry
- Particularly visible after `session.touch()` or any API response that triggers `_updateClient`

## Test plan

- [x] 8 new unit tests covering timer cleanup on overwrite scenarios
- [x] All 45 unit tests pass
- [x] Integration test verifying stable refresh rate after `session.touch()`
- [ ] Manual verification: sign in, trigger touch, observe network tab for ~2 minutes — token requests should maintain ~43s intervals (not accelerating)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a proactive token refresh timer leak that could cause growing token refresh requests after session activity or switching, preventing orphaned/stale timers and reducing redundant network calls.

* **Tests**
  * Added unit and integration tests validating timers are cleaned up across repeated session updates so token refreshes remain stable and non-duplicative.

### DIFF
--- a/.typedoc/custom-plugin.mjs
+++ b/.typedoc/custom-plugin.mjs
@@ -53,10 +53,10 @@ const LINK_REPLACEMENTS = [
   ['signed-in-session-resource', '/docs/reference/objects/session'],
   ['sign-in-resource', '/docs/reference/objects/sign-in'],
   ['sign-in-future-resource', '/docs/reference/objects/sign-in-future'],
-  ['sign-in-errors', '/docs/reference/javascript/types/errors'],
+  ['sign-in-errors', '/docs/reference/types/errors'],
   ['sign-up-resource', '/docs/reference/objects/sign-up'],
   ['sign-up-future-resource', '/docs/reference/objects/sign-up-future'],
-  ['sign-up-errors', '/docs/reference/javascript/types/errors'],
+  ['sign-up-errors', '/docs/reference/types/errors'],
   ['user-resource', '/docs/reference/objects/user'],
   ['session-status-claim', '/docs/reference/types/session-status'],
   ['user-organization-invitation-resource', '/docs/reference/types/user-organization-invitation'],
@@ -164,7 +164,7 @@ function getCatchAllReplacements() {
     {
       pattern: /(?<![\[\w`])`?((?:SignIn|SignUp)Errors)`?(?![\]\w`])/g,
       replace: (/** @type {string} */ _match, /** @type {string} */ type) =>
-        `[\`${type}\`](/docs/reference/javascript/types/errors)`,
+        `[\`${type}\`](/docs/reference/types/errors)`,
     },
     {
       pattern: /(?<![\[\w`])`?SignInFutureResource`?(?![\]\w`])/g,
--- a/integration/tests/session-token-cache/multi-session.test.ts
+++ b/integration/tests/session-token-cache/multi-session.test.ts
@@ -226,5 +226,106 @@ testAgainstRunningApps({ withEnv: [appConfigs.envs.withSessionTasks] })(
       expect(tab1FinalInfo.userId).toBe(user1SessionInfo.userId);
       expect(tab1FinalInfo.activeSessionId).toBe(user1SessionInfo.sessionId);
     });
+
+    /**
+     * Test Flow:
+     * 1. Tab1: Sign in as user1
+     * 2. Tab2: Inherits user1's session, then signs in as user2 (multi-session)
+     * 3. Tab1 has user1's active session; tab2 has user2's active session
+     * 4. Each tab's active session independently hydrates its token cache
+     * 5. Start counting /tokens requests, wait for both refresh timers to fire
+     * 6. Assert exactly 2 /tokens requests (one per session), with each session
+     *    represented exactly once
+     *
+     * Expected Behavior:
+     * - Two different sessions produce two independent refresh requests
+     * - BroadcastChannel does NOT deduplicate across sessions (different tokenIds)
+     * - Each session refreshes exactly once
+     *
+     * Note that this test does not currently assert in which tab the updates happen,
+     * this might be something we want to add in the future, but currently it is not
+     * deterministic.
+     */
+    test('multi-session scheduled refreshes produce one request per session', async ({ context }) => {
+      test.setTimeout(90_000);
+
+      const page1 = await context.newPage();
+      await page1.goto(app.serverUrl);
+      await page1.waitForFunction(() => (window as any).Clerk?.loaded);
+
+      const u1 = createTestUtils({ app, page: page1 });
+      await u1.po.signIn.goTo();
+      await u1.po.signIn.setIdentifier(fakeUser1.email);
+      await u1.po.signIn.continue();
+      await u1.po.signIn.setPassword(fakeUser1.password);
+      await u1.po.signIn.continue();
+      await u1.po.expect.toBeSignedIn();
+
+      const user1SessionId = await page1.evaluate(() => (window as any).Clerk?.session?.id);
+      expect(user1SessionId).toBeDefined();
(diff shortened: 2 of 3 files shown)


packages/clerk-js/src/core/tokenCache.ts
```diff
@@ -350,6 +350,19 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
...
+    // #hydrateCache during _updateClient AND #refreshTokenInBackground).
+    const existing = cache.get(key);
+    if (existing) {
+      if (existing.timeoutId !== undefined) {
+        clearTimeout(existing.timeoutId);
+      }
+      if (existing.refreshTimeoutId !== undefined) {
+        clearTimeout(existing.refreshTimeoutId);
+      }
+    }
```
Ephem: NIT: I don't mind the verbosity, just noting that since `clearTimeout(undefined)` is a noop, this could be shortened for readability:


```suggestion
    clearTimeout(existing?.timeoutId);
    clearTimeout(existing?.refreshTimeoutId);
```

packages/clerk-js/src/core/tokenCache.ts
```diff
@@ -369,6 +382,12 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
 
     entry.tokenResolver
       .then(newToken => {
+        // If this entry was overwritten by a newer set() call while our promise
+        // was pending, bail out to avoid installing orphaned timers.
+        if (cache.get(key) !== value) {
+          return;
+        }
```
Ephem, line 385: This makes sense and as long as `tokenResolver` doesn't have side effects (which it doesn't since that fetch has no client piggybacking), short circuiting here seems safe.

NIT: Moving

```tsx
cache.set(key, value);
```

to above

```tsx
entry.tokenResolver
  .then(newToken => {
```

should be semantically the same, and could make this part a tiny bit clearer to read.

packages/clerk-js/src/core/tokenCache.ts
```diff
@@ -350,6 +350,19 @@ const MemoryTokenCache = (prefix = KEY_PREFIX): TokenCache => {
 
     const key = cacheKey.toKey();
 
+    // Clear timers from any existing entry for this key to prevent orphaned
+    // refresh timers from accumulating across set() calls (e.g., from
+    // #hydrateCache during _updateClient AND #refreshTokenInBackground).
+    const existing = cache.get(key);
+    if (existing) {
+      if (existing.timeoutId !== undefined) {
+        clearTimeout(existing.timeoutId);
```
Ephem: If we only want to have one timer pending at any single time, would it make sense to move the `timeoutId` and `refreshTimeoutId` out into the parent scope and not saving them on the entries to have stricter guarantees?

packages/clerk-js/src/core/__tests__/tokenCache.test.ts
```diff
@@ -1189,6 +1189,284 @@ describe('SessionTokenCache', () => {
...
+      await Promise.resolve();
+
+      // After 30s the old timer would have deleted the entry, but it should still exist
+      vi.advanceTimersByTime(31 * 1000);
+      const result = SessionTokenCache.get(key);
+      expect(result).toBeDefined();
+      expect(result?.entry.tokenId).toBe('exp-overwrite');
+    });
+
+    it('does not accumulate refresh timers across multiple set() calls', async () => {
```
Ephem: NIT: Is this test necessary? It seems implied by the `cancels old refresh timer when set() is called again for the same key` test.

packages/clerk-js/src/core/__tests__/tokenCache.test.ts
```diff
@@ -1189,6 +1189,284 @@ describe('SessionTokenCache', () => {
...
+      }
+
+      // Advance past refresh fire time
+      vi.advanceTimersByTime(44 * 1000);
+
+      // Should fire exactly ONCE, not 10 times
+      expect(onRefresh).toHaveBeenCalledTimes(1);
+    });
+
+    it('simulates the hydration + background refresh double-set scenario', async () => {
```
Ephem: Is this not the same test as `cancels old refresh timer when set() is called again for the same key`, with different names for the `onRefresh` callbacks?

I like that these names reflect a real scenario, so maybe merge, but use these names and point out it's _one_ scenario that could happen?

packages/clerk-js/src/core/__tests__/tokenCache.test.ts
```diff
@@ -1189,6 +1189,284 @@ describe('SessionTokenCache', () => {
...
+
+      // Advance past refresh fire time
+      vi.advanceTimersByTime(44 * 1000);
+
+      // Only the second (background refresh) callback should fire
+      expect(hydrateRefresh).not.toHaveBeenCalled();
+      expect(backgroundRefresh).toHaveBeenCalledTimes(1);
+    });
+
+    it('simulates multiple refresh cycles without timer accumulation', async () => {
```
Ephem, line 1264: This looks like `cancels old refresh timer when set() is called again for the same key` but 5 times in a row.

I think this is a keeper though since it might test for _other_ scenarios of accumulation which is good.

packages/clerk-js/src/core/__tests__/tokenCache.test.ts
```diff
@@ -1189,6 +1189,284 @@ describe('SessionTokenCache', () => {
...
+
+      // Each cycle should show: 0 calls before timer, 1 call after timer
+      // If timers were accumulating, later cycles would show more than 1 call
+      for (let i = 0; i < refreshCounts.length; i += 2) {
+        expect(refreshCounts[i]).toBe(0); // before timer (42s)
+        expect(refreshCounts[i + 1]).toBe(1); // after timer (44s)
+      }
+    });
+
+    it('set() with different key does not affect existing timers', async () => {
```
Ephem, line 1309: Hmm, what's the scenario when we _should_ be refreshing multiple tokens in parallel in a single tab?

Update: I don't _think_ there is one? If so, let's not _fix_ that in this PR to keep it focused, but let's also not include this test that asserts the opposite behavior to what we actually want. 😄

integration/tests/session-token-cache/refresh-timer-cleanup.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        await route.continue();
+      });
+
+      // Trigger multiple touch() calls — each causes _updateClient → Session constructor
+      // → #hydrateCache → set(), which previously leaked orphaned refresh timers
+      for (let i = 0; i < 5; i++) {
+        await page.evaluate(async () => {
+          await (window as any).Clerk?.session?.touch();
+        });
+      }
```
Ephem, line 57: I think this is a multi-session app so it wont hit the 5s touch throttle? Might be worth a comment that this is necessary for the test to work.

packages/clerk-js/src/core/tokenCache.ts
Ephem: Upon reflection, while that might be something we do want to fix, let's keep that separate to keep this PR focused on one thing.

Ephem (approved):
Nice work! I like that you added a e2e-test here.

I've left a bunch of NIT-comments, mainly on the tests, but the fix itself looks good to me. I tried it in the dashboard and it does fix the accumulating timers. 👍

PR #8118, opened 2026-03-19
fix(shared): Handle missing sessionClaims and orgRole in resolveAuthState

jacekradko:
## Summary
- Hardens `resolveAuthState` to handle two edge cases that previously returned `undefined`, causing "Invalid state" errors
- When `sessionId`/`userId` exist but `sessionClaims` is missing (e.g. during client hydration before a token is fetched), returns a loading state (`isLoaded: false`) instead of throwing
- When `orgId` exists but `orgRole` is missing (e.g. membership not yet resolved), falls through to signed-in-without-org state instead of throwing
- Adds test coverage for both scenarios

## Context
This was exposed by #8101 when touch responses came back without `last_active_token`, causing `sessionClaims` to be `undefined`. While #8101 has been reverted, the underlying gap in `resolveAuthState` remains — any scenario where `lastActiveToken` is null (network errors, race conditions, backend changes) can trigger the same crash.

## Test plan
- [x] `npx vitest run packages/react/src/hooks/__tests__/useAuth.test.tsx` — 30/30 passing
- [x] `npx vitest run packages/shared/src/__tests__/authorization.spec.ts` — 4/4 passing
- [x] `npx turbo run build --filter=@clerk/shared --filter=@clerk/react --filter=@clerk/nextjs` — builds clean

### DIFF
--- a/packages/react/src/hooks/__tests__/useAuth.test.tsx
+++ b/packages/react/src/hooks/__tests__/useAuth.test.tsx
@@ -312,6 +312,49 @@ describe('useDerivedAuth', () => {
     expect(errorThrower.throw).toHaveBeenCalledWith(invalidStateError);
   });
 
+  it('returns loading state when sessionId and userId are present but sessionClaims is missing', () => {
+    const authObject = {
+      sessionId: 'session123',
+      userId: 'user123',
+      signOut: vi.fn(),
+      getToken: vi.fn(),
+    };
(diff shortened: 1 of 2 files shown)


brkalow:
@jacekradko I don't think this is a change we want to make, the error we encountered was good and revealed the underlying issue. WDYT?

PR #8152, opened 2026-03-23
fix(expo): add Kotlin metadata version compatibility flag for clerk-android

chriscanin:
## Summary

Fixes Kotlin metadata version mismatch errors during Android builds with Expo SDK 54/55.

The `clerk-android` native dependency is compiled with Kotlin 2.3.x, but Expo SDK 54 uses Kotlin 2.1.20 and SDK 55 uses Kotlin 2.1.x. When Gradle resolves `kotlinx-serialization-core:1.10.0` (a transitive dependency from clerk-android), the older Kotlin compiler can't read the 2.3.x metadata, causing `:app:compileDebugKotlin` to fail with:

```
Module was compiled with an incompatible version of Kotlin.
The binary version of its metadata is 2.3.0, expected version is 2.1.0.
```

The fix adds `-Xskip-metadata-version-check` to all project modules via `rootProject.allprojects` in the `@clerk/expo` build.gradle. This allows the older compiler to read newer metadata without failing.

Wrapped in try/catch for forward compatibility with Gradle's Isolated Projects feature, which may block cross-project configuration in future versions.

## Reported by

- Tom (internal, via Plain thread)
- Community user (https://github.com/erichli1/0319-expo-android-bug)

## Test plan

- [ ] Android build with Expo SDK 54 succeeds (no Kotlin metadata error)
- [ ] Android build with Expo SDK 55 succeeds (no Kotlin metadata error)
- [ ] Verify using the community repro: https://github.com/erichli1/0319-expo-android-bug

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents Kotlin metadata version mismatches that could cause Android build failures (e.g., “incompatible version of Kotlin” during :app:compileDebugKotlin), improving build reliability for projects using Expo SDK 54/55 and similar setups.

* **Chores**
  * Added a changeset to publish a patch release addressing the issue.

### DIFF
--- a/packages/expo/android/build.gradle
+++ b/packages/expo/android/build.gradle
@@ -54,7 +54,9 @@ android {
 
   kotlinOptions {
     jvmTarget = "17"
-    // clerk-android transitive deps (e.g. telemetry-api) still ship Kotlin 2.3.x metadata
+    // TODO: Remove once clerk-android pins its transitive deps to a Kotlin version compatible with Expo.
+    // clerk-android transitive deps (e.g. kotlinx-serialization-core:1.10.0) ship Kotlin 2.3.x metadata,
+    // but Expo SDK 54/55 uses Kotlin 2.1.x. This flag lets the 2.1.x compiler read 2.3.x metadata safely.
     freeCompilerArgs += ['-Xskip-metadata-version-check']
   }
 


packages/expo/android/build.gradle
```diff
@@ -75,7 +75,23 @@ android {
...
+      compilerOptions {
+        freeCompilerArgs.add('-Xskip-metadata-version-check')
+      }
+    }
+  }
+} catch (Exception ignored) {
+  // Future Gradle versions with Isolated Projects may block cross-project configuration.
+  // In that case, users should add '@clerk/expo' to their app.json plugins array instead,
+  // which applies the flag via the config plugin's withClerkAndroid.
+}
```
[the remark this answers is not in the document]
chriscanin: This is a good idea @coderabbitai but these fixes have been manually validated. I would like to get some testing and merge gating in as soon as possible.

chriscanin:
Adding changeset now.

packages/expo/android/build.gradle
```diff
@@ -75,7 +75,23 @@ android {
   }
 }
 
-// Note: kotlin-stdlib exclusions are handled in the clerk-android-ui dependency declaration
+// clerk-android is compiled with Kotlin 2.3.x, but Expo/RN projects typically use an older
+// Kotlin version (e.g. 2.1.x). This causes "incompatible version of Kotlin" metadata errors
+// at :app:compileDebugKotlin. Applying -Xskip-metadata-version-check allows the older
+// compiler to read 2.3.x metadata without failing.
```
brkalow, line 83: are there compatibility concerns beyond the version check here that we need to consider?

chriscanin: I have addressed these with Sam earlier last week, but the issue is that clerk-android's transitive dependencies (specifically kotlinx-serialization-core:1.10.0) are compiled with Kotlin 2.3.x, while Expo SDK 54 & 55 ships Kotlin 2.1.x so the -Xskip-metadata-version-check flag only bypasses the conservative version check on metadata reading. It does not skip the binary compatibility validation.

I believe Kotlin maintains backward compatibility for metadata across minor versions (2.1 reading 2.3). Ideally clerk-android and its peer dependencies will be compiled with the version that expo ships, but it looks like expo is behind by 2 minor versions, and although Sam and I knocked kotlin down to be compiled with 2.1 as well, some of the peer dependencies are still on 2.3, which is causing the error to pop up. 

When Sam returns from vacation, I will make sure to meet with him and discuss this, to ensure that the expo and android SDK's are synced up in the best way possible (and with testing too!)

chriscanin: This is a summary from claude that may help to clear things up as well:

**The core issue:** `clerk-android` currently compiles against Kotlin 2.3.x, while Expo SDK 54/55 ships Kotlin 2.1.x. This creates a metadata version mismatch for transitive dependencies — most notably `kotlinx-serialization-core:1.10.0`, which is compiled with Kotlin 2.3.0 metadata that a 2.1.x compiler refuses to read by default.

**Why this flag is safe:** `-Xskip-metadata-version-check` only bypasses the _conservative metadata version check_ — it does not skip binary compatibility validation. Kotlin maintains backward compatibility for metadata across minor versions, so a 2.1.x compiler reading 2.3.x metadata works correctly in practice. The strict check is just being overly cautious.

**Why the Android SDK's transitive deps are on 2.3.x:** The Android SDK doesn't currently pin or constrain its transitive dependency versions. `kotlinx-serialization-json` is declared as an `api()` dependency, which means `kotlinx-serialization-core` flows through to all consumers. Since `kotlinx-serialization:1.10.0` requires Kotlin 2.3.0 (per the [kotlinx.serialization compatibility matrix](https://github.com/Kotlin/kotlinx.serialization/blob/master/CHANGELOG.md)), any consumer on Kotlin 2.1.x will hit this metadata error. Similarly, `ktor:3.4.0` brings in its own set of 2.3.x-compiled transitive deps.

**What we're doing about it:** This flag is the right short-term fix to unblock users. Longer term, I'm working with Sam to ensure the Android SDK and its dependency tree are properly aligned with the Kotlin version that Expo ships, whether that means pinning transitive dependency versions, adding dependency constraints, or maintaining a compatible build configuration specifically for Expo consumers. We'll also be adding proper integration testing to catch these mismatches before they reach users.

brkalow: 👍 let's add a TODO here so we have a flag to come back and adjust this

chriscanin: TODO comment has been added. I am also making a linear ticket for this, will link here when completed.

PR #8162, opened 2026-03-25
fix(backend): preserve Set-Cookie headers through proxy

brkalow:
## Summary
- Use `headers.append()` instead of `headers.set()` for `set-cookie` response headers in `clerkFrontendApiProxy`, fixing OAuth callback failures caused by dropped `__client` / `__client_uat` cookies
- Add host validation guard to harden proxy

## Test plan
- [x] Added test verifying multiple `Set-Cookie` headers are preserved through the proxy (using `response.headers.getSetCookie()`)
- [x] Added test verifying protocol-relative SSRF paths are rejected with 400 before any fetch is made
- [x] All 38 existing proxy tests continue to pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Proxy now preserves and forwards multiple Set-Cookie headers from upstream to clients.

* **Security**
  * Requests that resolve to a host different from the expected frontend API host are rejected with a 400 error.

* **Tests**
  * Added tests covering Set-Cookie preservation and rejection of ambiguous/malicious proxy target resolutions.

PR #8163, opened 2026-03-25
fix(backend): harden FAPI proxy resilience and spec compliance

brkalow:
## Summary

- **Propagate client abort signal** to upstream `fetch()` to prevent zombie requests when clients disconnect
- **Strip dynamic hop-by-hop headers** listed in the `Connection` header per RFC 7230 Section 6.1, for both request and response header copying
- **Support request bodies on any HTTP method** (e.g., DELETE-with-body) by checking `request.body !== null` instead of a method allowlist
- **Add `Cache-Control: no-store`** to all error responses to prevent CDN/browser caching of transient errors
- **Only set `duplex: 'half'`** when the request actually has a body, avoiding unnecessary option on bodyless requests
- Converted `HOP_BY_HOP_HEADERS` from array to `Set` for O(1) lookups

## Test plan

- [x] Existing proxy tests continue to pass (82 tests)
- [x] New test: DELETE request with body is forwarded with `duplex: 'half'`
- [x] New test: Abort signal from incoming request is propagated to fetch
- [x] New test: Error responses (500 and 502) include `Cache-Control: no-store`
- [x] New test: Dynamic hop-by-hop headers listed in `Connection` header are stripped from forwarded requests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proxy behavior: support DELETE requests with bodies, reliable request-body detection, forward abort signals to upstream, ensure error responses use no-cache, and strip dynamically nominated hop-by-hop headers.

* **Tests**
  * Added tests for DELETE-with-body forwarding, signal propagation, cache-control on failures, and dynamic hop-by-hop header stripping.

* **Chores**
  * Added a patch release entry for the backend package.

### DIFF
--- a/packages/backend/src/proxy.ts
+++ b/packages/backend/src/proxy.ts
@@ -77,7 +77,7 @@ function getDynamicHopByHopHeaders(headers: Headers): Set<string> {
 // and Content-Length reflects the compressed size. We request identity
 // encoding upstream to avoid the double compression pass, but strip
 // these defensively since servers may ignore Accept-Encoding: identity.
-const RESPONSE_HEADERS_TO_STRIP = ['content-encoding', 'content-length'];
+const RESPONSE_HEADERS_TO_STRIP = new Set(['content-encoding', 'content-length']);
 
 /**
  * Derives the Frontend API URL from a publishable key.
@@ -322,7 +322,7 @@ export async function clerkFrontendApiProxy(request: Request, options?: Frontend
       const lower = key.toLowerCase();
       if (
         !HOP_BY_HOP_HEADERS.has(lower) &&
-        !RESPONSE_HEADERS_TO_STRIP.includes(lower) &&
+        !RESPONSE_HEADERS_TO_STRIP.has(lower) &&
         !responseDynamicHopByHop.has(lower)
       ) {
         if (lower === 'set-cookie') {


jacekradko:
### Code review

Found 3 issues:

1. SSRF via protocol-relative path resolution. The PR changes URL construction from string concatenation (`new URL(\`${fapiBaseUrl}${targetPath}\`)`) to `new URL(targetPath, fapiBaseUrl)` and removes both the explanatory comment and the host-mismatch guard (`targetUrl.host !== fapiHost`). A request to `/__clerk//evil.com/steal` causes `targetPath` to be `//evil.com/steal`, and `new URL('//evil.com/steal', 'https://frontend-api.clerk.dev')` resolves to `https://evil.com/steal` — forwarding the `Clerk-Secret-Key` to an attacker-controlled domain. The main branch had string concatenation and a host check specifically to prevent this; both are removed here along with the test "does not follow protocol-relative paths".

https://github.com/clerk/javascript/blob/e7346d8bf25680997b0a93e2667d3e0b3dccb677/packages/backend/src/proxy.ts#L236-L241

2. Multiple Set-Cookie headers silently collapsed. The PR removes the `set-cookie` special case that used `responseHeaders.append(key, value)` to preserve multiple Set-Cookie headers, replacing it with `responseHeaders.set(key, value)` for all headers. `Headers.set()` overwrites previous values for the same key, so if FAPI returns multiple Set-Cookie headers (e.g., `__client`, `__client_uat`, `__session`), only the last one survives. This regresses the fix from #8162, which was merged specifically to address this. The test "preserves multiple Set-Cookie headers from FAPI response" was also removed.

https://github.com/clerk/javascript/blob/e7346d8bf25680997b0a93e2667d3e0b3dccb677/packages/backend/src/proxy.ts#L313-L324

3. Missing changeset for user-facing changes. The PR modifies `@clerk/backend` with behavioral changes (abort signal propagation, DELETE body support, hop-by-hop stripping, Cache-Control on errors) but includes no `.changeset` file, which [`docs/CONTRIBUTING.md`](https://github.com/clerk/javascript/blob/e7346d8bf25680997b0a93e2667d3e0b3dccb677/docs/CONTRIBUTING.md) requires.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jacekradko:
### Code review (updated)

All 3 issues from the previous review have been addressed by merging main into the branch:

1. ~~SSRF via protocol-relative path resolution~~ -- Resolved. The string concatenation URL construction and host-mismatch guard from main are now present.
2. ~~Multiple Set-Cookie headers silently collapsed~~ -- Resolved. The `set-cookie` special case using `.append()` is now preserved.
3. ~~Missing changeset~~ -- Resolved via merge.

No remaining issues found.

🤖 Generated with [Claude Code](https://claude.ai/code)

packages/backend/src/proxy.ts
```diff
@@ -270,31 +292,40 @@ export async function clerkFrontendApiProxy(request: Request, options?: Frontend
...
   }
 
-  // Determine if request has a body
-  const hasBody = ['POST', 'PUT', 'PATCH'].includes(request.method);
+  // Determine if request has a body (handles DELETE-with-body and any other method)
+  const hasBody = request.body !== null;
 
   try {
     // Make the proxied request
+    // TODO: Consider adding AbortSignal.timeout(30_000) via AbortSignal.any()
```
dstaley, line 300: `AbortSignal.timeout` was added in Node 17. What other backend runtime are we waiting for support for?

packages/backend/src/proxy.ts
```diff
@@ -270,31 +292,40 @@ export async function clerkFrontendApiProxy(request: Request, options?: Frontend
...
-      // @ts-expect-error - duplex is required for streaming bodies but not in all TS definitions
-      duplex: hasBody ? 'half' : undefined,
+      signal: request.signal,
     };
 
-    // Only include body for methods that support it
-    if (hasBody && request.body) {
+    // Only set duplex when body is present (required for streaming bodies)
+    if (hasBody) {
+      // @ts-expect-error - duplex is required for streaming bodies but not in all TS definitions
```
dstaley: which definitions is it not in?

brkalow: It's not present on the `RequestInit` type, I will clarify 👀

packages/backend/src/proxy.ts
brkalow: Good Q! Seems like it's also supported in CF workers as well, so I don't think we need to worry too much about runtime compat.

I'll handle the generic top-level timeout in a follow-up

packages/backend/src/proxy.ts
```diff
@@ -270,31 +292,40 @@ export async function clerkFrontendApiProxy(request: Request, options?: Frontend
...
 
-  // Determine if request has a body
-  const hasBody = ['POST', 'PUT', 'PATCH'].includes(request.method);
+  // Determine if request has a body (handles DELETE-with-body and any other method)
+  const hasBody = request.body !== null;
 
   try {
     // Make the proxied request
+    // TODO: Consider adding AbortSignal.timeout(30_000) via AbortSignal.any()
+    // once it's available across all target runtimes.
```
brkalow:
```suggestion
```

packages/backend/src/proxy.ts
```diff
@@ -270,31 +292,39 @@ export async function clerkFrontendApiProxy(request: Request, options?: Frontend
...
+    // Build response headers, excluding hop-by-hop and encoding headers.
+    // Also strip dynamically-nominated hop-by-hop headers from the response Connection header.
+    const responseDynamicHopByHop = getDynamicHopByHopHeaders(response.headers);
     const responseHeaders = new Headers();
     response.headers.forEach((value, key) => {
       const lower = key.toLowerCase();
-      if (!HOP_BY_HOP_HEADERS.includes(lower) && !RESPONSE_HEADERS_TO_STRIP.includes(lower)) {
+      if (
+        !HOP_BY_HOP_HEADERS.has(lower) &&
+        !RESPONSE_HEADERS_TO_STRIP.includes(lower) &&
```
jacekradko: [nit] we could also convert `RESPONSE_HEADERS_TO_STRIP` to a set for the `.has()` method. But super super minor

packages/backend/src/proxy.ts
```diff
@@ -52,7 +52,25 @@ const HOP_BY_HOP_HEADERS = [
...
   'upgrade',
-];
+]);
+
+/**
+ * Parses the Connection header to extract dynamically-nominated hop-by-hop
+ * header names (RFC 7230 Section 6.1). These headers are specific to the
+ * current connection and must not be forwarded by proxies.
+ */
+function getDynamicHopByHopHeaders(headers: Headers): Set<string> {
```
jacekradko, line 62: We could probably add a unit test for this

brkalow: covered implicitly in this test case: `strips dynamic hop-by-hop headers listed in the Connection header from requests`

PR #8170, opened 2026-03-26
fix(clerk-js): static QueryClient import in native entry point

jacekradko:
## Summary

Alternative approach to fixing the React Native QueryClient issue (discussed in #8087):

- Creates a dedicated native entry point (`index.native.ts`) that statically imports `QueryClient` and overrides the lazy getter on the `Clerk` prototype
- Points the `clerkNative` rspack variant at this new entry point
- Removes the Expo-side workaround (`require('@tanstack/query-core')` override in `createClerkInstance.ts`)
- Removes `@tanstack/query-core` as a direct dependency of `@clerk/expo`

This fixes the root cause at the bundler level — only the native build's entry point is affected, leaving other dynamic imports (cookieSuffix, debug, etc.) untouched.

## Test plan

- [ ] Build passes (`pnpm build`)
- [ ] Verify on iOS simulator with NativeComponentQuickstart app
- [ ] `useOrganizationList` and other query-based hooks return real data on React Native

### DIFF
--- a/packages/clerk-js/src/__tests__/index.native.test.ts
+++ b/packages/clerk-js/src/__tests__/index.native.test.ts
@@ -40,4 +40,14 @@ describe('index.native', () => {
 
     expect(client1?.client).not.toBe(client2?.client);
   });
+
+  it('exports the same public API as the base entry point', async () => {
+    const nativeExports = await import('../index.native');
+    const baseExports = await import('../index');
+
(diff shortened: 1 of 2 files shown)


jacekradko:
Closing as this seems to be a more robust approach: https://github.com/clerk/javascript/pull/8087#issuecomment-4136550915

PR #8186, opened 2026-03-27
fix(backend): use redirect:'manual' in frontend API proxy fetch

brkalow:
## Summary
- `clerkFrontendApiProxy` uses `fetch()` to forward requests to FAPI, but without `redirect: 'manual'`, fetch defaults to `redirect: 'follow'` — chasing redirects server-side.
- When FAPI returns a 302 back to the app origin (e.g. after processing an OAuth callback), `fetch()` follows the redirect internally, hitting the app without auth cookies. This renders the sign-in page at the callback URL.
- The fix adds `redirect: 'manual'` to the fetch options so the 302 is returned to the browser as-is.

## How it was discovered
A customer's OAuth callback was resolving to their `/sign-in` page. Server logs showed the proxy request succeeding, but a second request was rendering the dashboard route without auth — this second request was invisible in browser network logs because it was `fetch()` following the redirect server-side inside the proxy.

## Test plan
- [x] Backend proxy tests pass (72/72)
- [x] Next.js clerkMiddleware tests pass (134/134)
- [ ] Manual test: OAuth flow with `frontendApiProxy` enabled completes without redirecting to sign-in

## Summary by CodeRabbit

* **Bug Fixes**
  * Proxy now returns redirect (3xx) responses to clients unchanged instead of following them server-side.
  * Prevents server-side automatic redirect handling for frontend API flows (including post‑OAuth), ensuring clients receive original redirect responses.

### DIFF
--- a/.changeset/fix-proxy-redirect-manual.md
+++ b/.changeset/fix-proxy-redirect-manual.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
+Fix frontend API proxy following redirects server-side instead of passing them to the browser. The proxy's `fetch()` call now uses `redirect: 'manual'` so that 3xx responses from FAPI (e.g. after OAuth callbacks) are returned to the client as-is, matching standard HTTP proxy behavior.


brkalow:
!snapshot

packages/backend/src/proxy.ts
```diff
@@ -278,6 +272,11 @@ export async function clerkFrontendApiProxy(request: Request, options?: Frontend
     const fetchOptions: RequestInit = {
       method: request.method,
       headers,
+      // Return redirects as-is instead of following them server-side.
+      // Without this, fetch() follows same-origin redirects internally,
+      // which bypasses the browser and breaks flows like OAuth callbacks
+      // where FAPI redirects back to the app origin.
```
brkalow:
```suggestion
```

PR #8196, opened 2026-03-28
feat(shared): Add cloudflare:workers env fallback to getEnvVariable

Chat-JPTeasdale:
## Description

Add `initCloudflareWorkersEnv()` and a `cloudflare:workers` module env check to the shared `getEnvVariable()` function in `@clerk/shared`.

This allows Clerk to read environment variables (e.g., `CLERK_SECRET_KEY`, `CLERK_PUBLISHABLE_KEY`) directly from the Cloudflare Workers runtime environment, following the same pattern already used in `@clerk/astro` (PRs #7889, #8136).

### Problem

When using Clerk with frameworks running on Cloudflare Workers (e.g., TanStack Start, Hono), `process.env` and `import.meta.env` are not available at runtime. The current `getEnvVariable()` only falls back to `globalThis[name]` for Workers, which doesn't work because Worker env vars are not set as globals — they're passed via the fetch handler's `env` parameter or accessible via the `cloudflare:workers` module.

### Solution

- Add a cached `cloudflareWorkersEnv` variable (same pattern as `@clerk/astro`'s `get-safe-env.ts`)
- Add `initCloudflareWorkersEnv()` that dynamically imports `cloudflare:workers` and caches the env object
- Check `cloudflareWorkersEnv[name]` before falling through to context-based checks
- Falls through when the key is missing (e.g., on CF Pages where dashboard secrets may not be in the module env)
- Safe to call in non-Cloudflare environments — silently no-ops

### Why in `@clerk/shared`?

The Astro package has its own `get-safe-env.ts` with this pattern, but other framework packages (`@clerk/tanstack-react-start`, `@clerk/backend`) import `getEnvVariable` from `@clerk/shared`. Adding the `cloudflare:workers` fallback to the shared utility benefits all framework packages running on CF Workers without requiring per-package fixes.

## Checklist

- [x] `pnpm build` runs as expected (no breaking changes — new export is additive)
- [x] (If applicable) JSDoc comments have been added

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change

## Summary by CodeRabbit

* **New Features**
  * Clerk environment variables can now be resolved from Cloudflare Workers runtime, enabling configuration of CLERK_SECRET_KEY and CLERK_PUBLISHABLE_KEY in Cloudflare Workers environments.

### DIFF
--- a/.changeset/cloudflare-workers-env-fallback.md
+++ b/.changeset/cloudflare-workers-env-fallback.md
@@ -1,0 +1,9 @@
+---
+'@clerk/shared': patch
+---
+
+feat(shared): Add `cloudflare:workers` env fallback to `getEnvVariable`
+
+Adds `initCloudflareWorkersEnv()` and a `cloudflare:workers` module env check to the shared `getEnvVariable()` function. This allows Clerk to read environment variables (e.g., `CLERK_SECRET_KEY`, `CLERK_PUBLISHABLE_KEY`) directly from the Cloudflare Workers runtime environment, following the same pattern used in `@clerk/astro` (PRs #7889, #8136).
+
+This is needed for frameworks running on Cloudflare Workers (e.g., TanStack Start, Hono) where `process.env` and `import.meta.env` are not available at runtime, but `cloudflare:workers` module env is.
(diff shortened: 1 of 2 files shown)


JPTeasdale:
I’m JPTeasdale and I approve this message. 

It’s also possible that the tanstack-react-start package should be changed to pass in the context or locals parameter that already exists. 

Please advise on the best way to do this.

PR #8207, opened 2026-03-31
chore(repo): Update rolldown and tsdown

dstaley:
## Description

Updates rolldown and tsdown to their latest versions. This surfaced several issues with our type system, which have been resolved in this PR:

1. We're now only building one type tree for `@clerk/shared`, using `tsc` instead of bundling types with `tsdown`. This prevents issues where the same type was pulled from `dist/types` and `dist/runtime`.
2. To support emitting types with `tsc`, the `@/` path alias was removed in favor of relative imports. We can explore using `"imports"` in `package.json` in a future PR if we know that every runtime `@clerk/shared` supports has support for that.
3. Fixed an issue with CJS type resolution in `@clerk/react` by pointing the `"require"` `"types"` field to the emitted CJS type bundle.
4. Created a structurally-similar `Ui` type in the `@clerk/react` package that allows the bundled `Ui` type and the one used by the `@clerk/ui` package to match, preventing TypeScript from complaining that the types don't match.
5. Updates `turbo.json` to factor in `tsdown.config.mts` into whether or not to invalidate the cache.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/sandbox/app.ts
+++ b/packages/clerk-js/sandbox/app.ts
@@ -1,5 +1,5 @@
 import { PageMocking, type MockScenario } from '@clerk/msw';
-import * as l from '../../localizations';
+import * as l from '../../localizations/src';
 import { dark, neobrutalism, shadcn, shadesOfPurple } from '../../ui/src/themes';
 import type { Clerk as ClerkType } from '../';
 import * as scenarios from './scenarios';
--- a/packages/expo/src/cache/index.ts
+++ b/packages/expo/src/cache/index.ts
@@ -1,4 +1,4 @@
-export { TokenCache } from './types';
+export type { TokenCache } from './types';
 
 export { MemoryTokenCache } from './MemoryTokenCache';
 export { ClientResourceCache, EnvironmentResourceCache, SessionJWTCache } from './ResourceCache';
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -29,8 +29,8 @@
         "default": "./dist/index.mjs"
       },
       "require": {
-        "types": "./dist/index.d.ts",
-        "default": "./dist/index.js"
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
       }
     },
     "./internal": {
@@ -39,8 +39,8 @@
         "default": "./dist/internal.mjs"
       },
       "require": {
-        "types": "./dist/internal.d.ts",
-        "default": "./dist/internal.js"
+        "types": "./dist/internal.d.cts",
+        "default": "./dist/internal.cjs"
       }
     },
     "./errors": {
@@ -49,8 +49,8 @@
         "default": "./dist/errors.mjs"
       },
       "require": {
-        "types": "./dist/errors.d.ts",
-        "default": "./dist/errors.js"
+        "types": "./dist/errors.d.cts",
+        "default": "./dist/errors.cjs"
       }
     },
     "./experimental": {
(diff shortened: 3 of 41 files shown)


packages/react/package.json
```diff
@@ -29,8 +29,8 @@
         "default": "./dist/index.mjs"
       },
       "require": {
-        "types": "./dist/index.d.ts",
-        "default": "./dist/index.js"
+        "types": "./dist/index.d.cts",
```
dstaley, line 32: I do think it's somewhat strange that tsdown emits both `mts` and `cts` files. I would have thought that the CJS files would be `js` since the package doesn't have `"type": "module"`. After looking into the config for tsdown it does appear that this is to be expected though. When you emit a `cjs` bundle, tsdown defaults `platform` to `"node"`, which forces the `fixedExtensions` config to be `true`. So it does seem like this is intentional for dual ESM+CJS packages.

packages/react/tsdown.config.mts
```diff
@@ -66,13 +66,15 @@ export default defineConfig(overrideOptions => {
...
+    dts: {
+      cjsReexport: true,
+    },
     onSuccess: shouldPublish ? 'pkglab pub --ping' : undefined,
     format: ['cjs', 'esm'],
     clean: true,
     minify: false,
     sourcemap: true,
-    external: ['react', 'react-dom'],
+    external: ['react', 'react-dom', '@clerk/ui/internal'],
```
nikosdouvlis: I built `@clerk/react` with this branch and checked the output. `dist/index.d.mts` has `import { Theme, Ui } from "@clerk/ui/internal"` and the global `ClerkAppearanceRegistry` augmentation right below it. `dist/internal.d.mts` has the same for `Ui`.

Since `@clerk/ui` is a devDep of `@clerk/react`, consumers wont have it in their tree. The global augmentation gets pulled in as a side effect (`import './types/appearance'` in index.ts), so any import from `@clerk/react` needs `@clerk/ui` resolvable for type checking. Same for framework SDKs since they all import `Ui` from `@clerk/react/internal`.

Making `@clerk/ui` a full dep isnt viable given its dep tree (emotion, floating-ui, solana, etc). Not sure whats the best path here - maybe keeping it not-external and finding a different fix for the unique symbol duplication?

packages/shared/src/loadScript.ts
```diff
@@ -46,7 +46,9 @@ export async function loadScript(src = '', opts: LoadScriptOptions): Promise<HTM
       });
 
       script.src = src;
-      script.nonce = nonce;
+      if (nonce) {
+        script.nonce = nonce;
+      }
```
jacekradko, line 51: Is this a type-related change?

dstaley: yeah, since the `nonce` is potentially `undefined` and the DOM types have `nonce: string`

packages/clerk-js/sandbox/app.ts
```diff
@@ -1,5 +1,5 @@
 import { PageMocking, type MockScenario } from '@clerk/msw';
-import * as l from '../../localizations';
+import * as l from '../../localizations/src';
```
dstaley, line 2: This was added to match the `ui` import below.

packages/expo/src/cache/index.ts
```diff
@@ -1,4 +1,4 @@
-export { TokenCache } from './types';
+export type { TokenCache } from './types';
```
dstaley, line 1: This prevents an error where `TokenCache` was expected to be a non-type value.

PR #8223, opened 2026-04-01
fix(clerk-js,shared): upgrade Stripe SDK from v5 to v9

jacekradko:
## Summary

- Upgrade `@stripe/stripe-js` from 5.6.0 to 9.0.0 (4 major versions)
- Upgrade `@stripe/react-stripe-js` from 3.1.1 to 6.0.0 (3 major versions)
- Add unit tests for `usePaymentElement` hook covering the `confirmSetup` flow
- Bump `stripe-vendors` bundlewatch limit from 1KB to 2KB to accommodate slightly larger v9 vendor chunk (~1.3KB gzipped)

The codebase uses a narrow slice of the Stripe API (SetupIntents + PaymentElement via a forked react-stripe-js wrapper), so the breaking changes across these versions don't affect us:

- **`redirectToCheckout` removed (v8)** — not used
- **`createSource`/`retrieveSource` types removed (v9)** — not used, we use `confirmSetup`
- **`elements.update()` returns `Promise<void>` (v9)** — return value is already ignored
- **`options.layout.radios` no longer accepts boolean (v9)** — not used
- **Checkout component renames (react-stripe-js v4-v6)** — only types imported, no Checkout usage

Note: `9.0.1` / `6.1.0` exist but are blocked by the repo's 2-day `minimumReleaseAge` policy. Can be bumped in a follow-up.

## Test plan

- [x] `pnpm build` passes for `@clerk/shared` and `@clerk/clerk-js`
- [x] All 18 billing/Stripe unit tests pass (8 new `usePaymentElement` tests + 10 existing)
- [x] New tests cover `confirmSetup` success, error handling (card_error, validation_error), reset flow, and provider readiness states
- [ ] Verify Stripe PaymentElement renders correctly in a dev environment
- [ ] Verify `confirmSetup` flow works end-to-end

## Summary by CodeRabbit

* **Chores**
  * Upgraded Stripe JavaScript library to v9.0.0
  * Upgraded Stripe React library to v6.0.0

* **Tests**
  * Added test coverage for payment element functionality


dstaley:
I believe the highest we can upgrade to is stripe.js v6.1.0, which is the last major stripe.js aligned with the acacia API versions, which is what we use on the backend. Stripe's documentation recommends using the same versions on the frontend as the backend.

>When performing API requests, each versioned Stripe.js automatically uses the API version associated with the Stripe.js version. That is, the Stripe.js acacia version uses a compatible API version such as 2024-12-18.acacia (which includes the date) to represent the release date of the API version. You can’t override the API version.
>
>We recommend that businesses keep the Stripe.js version and server-side API version on the same release train. For example, if you use Stripe.js acacia, you can safely use 2025-02-24.acacia or another acacia minor version on your application server. Gradual updates are safe, so you don’t need to release both at the same time.

jacekradko:
Thanks @dstaley ! Will adjust the PR accordingly

PR #8225, opened 2026-04-02
chore(repo): update pkglab to 0.17.1

jacekradko:
## Summary
- Updates pkglab from `0.15.0` to `0.17.1` (pinned)
- Includes upstream packument retry fix that resolves transient CI failures when fetching large package metadata (e.g. `@solana/web3.js`)

## Test plan
- [x] CI passes the `Install @clerk/clerk-js in os temp` step without `FETCH_ERROR`
- [x] `pkglab add` works correctly in integration test workflows


package.json
```diff
@@ -132,7 +132,7 @@
     "json5": "2.2.3",
     "jsonwebtoken": "9.0.2",
     "lint-staged": "^14.0.1",
-    "pkglab": "^0.15.0",
+    "pkglab": "0.17.1",
```
wobsoriano, line 135: ❓ why did we pin this again?

jacekradko: It's just more explicit this way. No need to look at the lockfile to see what version is installed

PR #8227, opened 2026-04-02
Draft: Add mechanism to manually run tests on PRs from forks

Ephem:
## Description

This PR adds a separate workflow that can be used by maintainers to trigger the integration tests on PRs from forks manually after review determines it's safe. This happens through adding a new `safe-to-test` label, which only works for people with write access to the repo.

Adding the label triggers a workflow on `pull_request_trigger`, which does have access to secrets.

The label gets removed on every push, requiring re-review and adding it again to trigger another run. The run that is triggered runs on the SHA from the _event_, not the latest SHA in the PR to avoid [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) attacks.

Note that the built in "Approve workflows to run" does _not_ apply to `pull_request_target` ([source](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories)), which is why we need the label.

The flow for a forked PR would be:

1. Fork PR is opened
1. A maintainer clicks "Approve and run" to let ci.yml's non-secret jobs run (build, lint, unit tests)
1. Those jobs pass (or the contributor fixes issues)
1. The maintainer reviews the full diff for secret safety
1. The maintainer adds the `safe-to-test` label to trigger integration tests via `ci-fork-integration.yml`

Both workflows manually post a status to a commit context with the same name, and this is what we use for a Required check, making it pass if either of the two workflows passes (normal or fork paths).

**TODO:**

- [x] Run through `actionlint`
- [ ] After merge: Test workflow from a fork - Can only be done after merge as workflow needs to live on `main`
- [ ] After merge: Adjust Required checks

**Follow ups:**

* Bundlewatch step
  * Either just skip it for forks, so Required check passes, OR
  * Add the same approach as in this PR

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI/CD infrastructure with improved fork pull request testing capabilities, including permission-based security controls and integration test result reporting in the pipeline.

### DIFF
--- a/.github/workflows/ci-fork-integration.yml
+++ b/.github/workflows/ci-fork-integration.yml
@@ -24,7 +24,7 @@ jobs:
     if: >-
       ${{ github.event.action == 'synchronize'
       && github.event.pull_request.head.repo.full_name != github.repository }}
-    runs-on: 'blacksmith-8vcpu-ubuntu-2204'
+    runs-on: 'blacksmith-2vcpu-ubuntu-2204'
     permissions:
       # Needed for removing the "safe-to-test" label
       pull-requests: write
@@ -43,7 +43,7 @@ jobs:
       ${{ github.event.action == 'labeled'
       && github.event.label.name == 'safe-to-test'
       && github.event.pull_request.head.repo.full_name != github.repository }}
-    runs-on: 'blacksmith-8vcpu-ubuntu-2204'
+    runs-on: 'blacksmith-2vcpu-ubuntu-2204'
     permissions:
       pull-requests: write
     name: Check Label Sender Permissions
@@ -288,3 +288,28 @@ jobs:
           name: playwright-traces-fork-${{ github.run_id }}-${{ github.run_attempt }}-${{ steps.sanitize.outputs.artifact-suffix }}${{ matrix.next-version && format('-next{0}', matrix.next-version) || '' }}
           path: test-results
           retention-days: 1
+
+  integration-tests-gate:
+    needs: [integration-tests]
+    if: ${{ always() && needs.integration-tests.result != 'skipped' }}
+    runs-on: 'blacksmith-2vcpu-ubuntu-2204'
+    permissions:
+      statuses: write
+    name: Fork Integration Tests Gate
+    steps:
+      - name: Report integration test status
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RESULT: ${{ needs.integration-tests.result }}
+          SHA: ${{ github.event.pull_request.head.sha }}
+          REPO: ${{ github.repository }}
+        run: |
+          if [ "$RESULT" = "success" ]; then
+            STATE="success"
+          else
+            STATE="failure"
+          fi
+          gh api "repos/$REPO/statuses/$SHA" \
+            -f state="$STATE" \
+            -f context="integration-tests" \
+            -f description="Fork integration tests: $RESULT"
--- a/.github/workflows/ci-fork-integration.yml
+++ b/.github/workflows/ci-fork-integration.yml
@@ -9,8 +9,6 @@ on:
     types: [labeled, synchronize]
     branches:
       - main
-      # For testing the workflow, remove before merging
-      - fredrik/fork-integration-tests
 
 permissions:
   contents: read
--- a/.github/workflows/ci-fork-integration.yml
+++ b/.github/workflows/ci-fork-integration.yml
@@ -1,0 +1,290 @@
+# Note: There is a copy of this workflow in ci.yml that is
+# used for triggering integration tests on forked PRs. If you tweak this
+# workflow, see if that also needs to be updated.
+
+name: 'CI: Fork Integration Tests'
+
+on:
+  pull_request_target:
+    types: [labeled, synchronize]
+    branches:
+      - main
+      # For testing the workflow, remove before merging
+      - fredrik/fork-integration-tests
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  remove-label-on-push:
+    if: >-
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-fork-integration.yml
```diff
@@ -0,0 +1,315 @@
...
+# workflow, see if that also needs to be updated.
+
+name: 'CI: Fork Integration Tests'
+
+on:
+  pull_request_target:
+    types: [labeled, synchronize]
+    branches:
+      - main
+      # For testing the workflow, remove before merging
```
Ephem: Adding a comment so I don't forget.

.github/workflows/ci-fork-integration.yml
Ephem, line 1: There's plenty of opportunity to de-duplicate parts of this with `ci.yml` later if we want, but it made sense to me to keep them separate for now to allow for easier reviewing and testing.

.github/workflows/ci.yml
```diff
@@ -476,6 +481,34 @@ jobs:
...
+          REPO: ${{ github.repository }}
+        run: |
+          if [ "$RESULT" = "success" ]; then
+            STATE="success"
+          else
+            STATE="failure"
+          fi
+          gh api "repos/$REPO/statuses/$SHA" \
+            -f state="$STATE" \
+            -f context="integration-tests" \
```
Ephem, line 509: These gates in the two files post to the same context, that way we can add that as a required check and it will pass if _either_ of the jobs pass.

.github/workflows/ci-fork-integration.yml
```diff
@@ -0,0 +1,315 @@
...
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 1
+          fetch-tags: false
+          filter: 'blob:none'
+          show-progress: false
+          clean: false
+
+      - name: Re-install dependencies for fork code
+        run: pnpm install
```
dominic-clerk, line 166: In case the review misses a change in config that had allowed a post-install script

```suggestion
        run: pnpm install --ignore-scripts
```

.github/workflows/ci-fork-integration.yml
```diff
@@ -0,0 +1,315 @@
...
+
+name: 'CI: Fork Integration Tests'
+
+on:
+  pull_request_target:
+    types: [labeled, synchronize]
+    branches:
+      - main
+      # For testing the workflow, remove before merging
+      - fredrik/fork-integration-tests
```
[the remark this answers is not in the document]
Ephem: The test branch will be removed before merging. It was a conscious decision to only include main. External contributions should always happen towards main, we backport any necessary changes ourselves. This makes the fork workflow easier to reason about as we only need to make sure the main branch workflows are safe, not the older branch workflows.

.github/workflows/ci-fork-integration.yml
```diff
@@ -0,0 +1,315 @@
...
+    runs-on: 'blacksmith-2vcpu-ubuntu-2204'
+    permissions:
+      pull-requests: write
+    name: Check Label Sender Permissions
+    steps:
+      - name: Verify label sender has write access
+        id: checkAccess
+        uses: actions-cool/check-user-permission@v2
+        with:
+          require: write
```
dominic-clerk, line 55: What about doing it like https://github.com/clerk/javascript/blob/b80eafb37ad073f85efb039a075c98cac8e12ee9/.github/workflows/release.yml#L276 where we limit to our team members. Possibly checking both if we have team members without write access.

I'd like the security gate to have as little 3rd party as possible and the way we do it in the release workflow uses only dependencies provided by GitHub.

I'm also wary of bots or 3rd party integrations that might have write access (dependabot for example) that might be hijacked... somehow. This is the paranoia speaking :)

.github/workflows/ci-fork-integration.yml
```diff
@@ -0,0 +1,315 @@
...
+          username: ${{ github.event.sender.login }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Remove unauthorized label and fail
+        if: ${{ steps.checkAccess.outputs.require-result == 'false' }}
```
dominic-clerk, line 60: The docs claim `true` | `false` are the only results, but I wonder what happens if it errors out for whatever reason? Let's fail on "not true" instead of "false".

```suggestion
        if: ${{ steps.checkAccess.outputs.require-result != 'true' }}
```

.github/workflows/ci-fork-integration.yml
```diff
@@ -0,0 +1,315 @@
...
+          fetch-depth: 1
+          fetch-tags: false
+          filter: 'blob:none'
+          show-progress: false
+
+      - name: Setup (from base branch)
+        id: config
+        uses: ./.github/actions/init-blacksmith
+        with:
+          turbo-signature: ${{ secrets.TURBO_REMOTE_CACHE_SIGNATURE_KEY }}
```
dominic-clerk, line 148: Can we have a separate cache entirely to make cache poisoning completely impossible? I know it shouldn't be possible already and would require a hash collision but it would make it completely impossible and protect us against a bug in turbo itself.

dominic-clerk (commented):
I left some mildly paranoied comments. I'll return to this again later.

.github/workflows/ci-fork-integration.yml
Ephem: We could also just not use the turbo cache. These runs don't happen often so I think it's fine if they take extra time.

dominic-clerk: Even better!

dominic-clerk:
Let's close this one and go back to the drawing board. If we must maintain a separate mirror for external contributions then we will but `pull_request_target` is too much of a danger to be used as demonstrated again today.

PR #8232, opened 2026-04-03
chore(repo): Remove jest leftovers

wobsoriano:
## Description

Removes the remaining Jest-specific tooling and config from the repo, also:

- remove root Jest dependencies and stale setup/config
- remove stale `types: ["jest"]` entries from package configs
- remove Jest scaffolding from integration templates and playground
- migrate `@clerk/chrome-extension` tests from jest-chrome to vitest-chrome

## Checklist

- [x] `pnpm test` runs as expected.
- [x] `pnpm build` runs as expected.
- [x] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Summary by CodeRabbit

* **Chores**
  * Migrated testing framework from Jest to Vitest across the codebase.
  * Removed Jest dependencies and configuration files from project configurations.
  * Updated ESLint configuration to use Vitest globals.
  * Updated test setup files and imports to use Vitest APIs.

PR #8242, opened 2026-04-06
chore(repo): Add explicit vitest imports and remove globals config

wobsoriano:
## Description

After #8232 removed `"types": ["jest"]` from tsconfigs, the ~10 test files using vitest globals lost their type backing. They still ran (vitest injects globals at runtime), but TypeScript was no longer checking the function signatures. Rather than replacing jest types with `"vitest/globals"`, this aligns all test files with the explicit import convention already used by ~390 other test files in the repo.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Chores**
  * Migrated test configuration across packages to remove global test API support, requiring explicit imports of test utilities in test files instead of relying on implicit global availability.
  * Updated test setup to include automatic cleanup after each test execution.


jacekradko (approved):
Nice! The explicit approach is preferrable

PR #8243, opened 2026-04-06
feat(expo): add native component theming via Expo config plugin

chriscanin:
Adds support for customizing native Clerk UI components (sign-in, sign-up, user profile) on both iOS and Android via a JSON theme configuration file referenced in the Expo plugin config:

  ["@clerk/expo", { "theme": "./clerk-theme.json" }]

The JSON schema supports:
- colors: 15 semantic color tokens (primary, background, danger, etc.)
- darkColors: dark mode color overrides (iOS uses @Environment colorScheme, Android uses ClerkTheme.darkColors)
- design.fontFamily: custom font family string (iOS only)
- design.borderRadius: corner radius number (both platforms)

Plugin changes:
- Reads and validates the JSON at prebuild time
- iOS: Embeds theme in Info.plist under ClerkTheme key
- Android: Copies JSON to app assets directory

The plugin does not modify `UIUserInterfaceStyle`. Dark mode switching is controlled by the developer via Expo's `userInterfaceStyle` setting in `app.json`.

Native changes:
- iOS: Parses theme from Info.plist, builds light/dark ClerkTheme objects, applies via .environment(\.clerkTheme) with colorScheme switching
- Android: Parses theme from assets JSON, sets Clerk.customTheme
- Both: AuthView now uses Clerk.customTheme instead of null

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


seanperez29:
@codex review

packages/expo/app.plugin.js
```diff
@@ -585,6 +585,120 @@ const withClerkAppleSignIn = config => {
...
+  config = withInfoPlist(config, modConfig => {
+    modConfig.modResults.ClerkTheme = themeJson;
+    console.log('✅ Embedded Clerk theme in Info.plist');
+
+    // When darkColors is provided, remove UIUserInterfaceStyle to allow
+    // the system to switch between light and dark mode automatically.
+    if (themeJson.darkColors) {
+      delete modConfig.modResults.UIUserInterfaceStyle;
+      console.log('✅ Removed UIUserInterfaceStyle to enable system dark mode');
+    }
```
seanperez29: Something codex's local review spotted on this:

`darkColors` changes the whole app's appearance policy
When darkColors is present, this deletes UIUserInterfaceStyle from Info.plist. That is an app-wide setting, not a Clerk-only one, so any app intentionally pinned to Light or Dark will silently switch back to system-controlled appearance just to theme Clerk screens.

chriscanin: Oh hmmmm I didnt consider this when doing the dark mode / light mode switch. I will look into solutions. Good callout.

chriscanin: Fixed!

packages/expo/src/__tests__/appPlugin.theme.test.js
```diff
@@ -0,0 +1,156 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+
+// The plugin is plain CJS — load the validation function by evaluating
+// just the relevant pieces.  We re-declare the constants and function
+// exactly as they appear in app.plugin.js so the test stays in sync.
```
manovotny: This test copies `validateThemeJson`, `VALID_COLOR_KEYS`, `HEX_COLOR_REGEX`, and `isPlainObject` from `app.plugin.js` instead of importing the real code. If the plugin's validation changes (eg. a new color key is added, the regex is tightened), these tests will still pass against the stale copy.

Could the validator be exported from `app.plugin.js` (eg. `module.exports.validateThemeJson = validateThemeJson`) so the test `require()`s the real implementation? That way drift between the two isn't possible.

Understand if there's an Expo plugin packaging constraint that prevents this. Just want to flag the possible maintenance risk.

packages/expo/ios/ClerkViewFactory.swift
```diff
@@ -18,6 +18,10 @@ public final class ClerkViewFactory: ClerkViewFactoryProtocol {
   private static let clerkLoadIntervalNs: UInt64 = 100_000_000
   private static var clerkConfigured = false
 
+  /// Parsed light and dark themes from Info.plist "ClerkTheme" dictionary.
+  var lightTheme: ClerkTheme?
```
manovotny, line 22: The template file at `ios/templates/ClerkViewFactory.swift` hasn't been updated with the theme changes. 

The config plugin copies from this file (`ios/ClerkViewFactory.swift`), so theming works, but the template is now significantly diverged (no `lightTheme`/`darkTheme` properties, no `loadThemes()`, no `@MainActor register()`, no wrapper view theme params).

Is the template still needed? The plugin references `ios/ClerkViewFactory.swift` directly (line 266 of `app.plugin.js`). If the template serves another purpose, maybe it should be synced.

packages/expo/android/src/main/java/expo/modules/clerk/ClerkExpoModule.kt
```diff
@@ -79,6 +85,9 @@ class ClerkExpoModule(reactContext: ReactApplicationContext) :
                     }
 
                     Clerk.initialize(reactApplicationContext, pubKey)
+                    // Must be set AFTER Clerk.initialize() because initialize()
+                    // resets customTheme to its `theme` parameter (default null).
+                    loadThemeFromAssets()
```
manovotny, line 94: `loadThemeFromAssets()` is only called here in `ClerkExpoModule`. But `ClerkViewFactory.configure()` and `ClerkUserProfileActivity.onCreate()` also call `Clerk.initialize()` without loading the theme. 

If Clerk is ever initialized through those paths first, `Clerk.customTheme` stays null and theming silently doesn't apply.

In normal Expo flow `ClerkExpoModule` initializes first, so this probably never triggers, but would it be worth either:
- Moving `loadThemeFromAssets()` into a shared companion called after every `Clerk.initialize()`, or
- Adding a comment documenting that `ClerkExpoModule` is the only supported init path and theme loading is intentionally centralized here?

packages/expo/app.plugin.js
```diff
@@ -585,6 +585,121 @@ const withClerkAppleSignIn = config => {
   });
 };
 
+/**
+ * Apply a custom theme to Clerk native components (iOS + Android).
```
manovotny, line 589: The plugin writes `ClerkTheme` to `Info.plist`, and `clerk_theme.json` to Android assets, but never cleans them up. 

If a project had a theme and later removes the `theme` prop, incremental prebuilds (`expo prebuild` without `--clean`) keep using the old embedded theme.

Maybe we could look into:
- Actively removing `ClerkTheme` from Info.plist and deleting `clerk_theme.json` when `theme` is absent, or
- Documenting that `expo prebuild --clean` is required after removing `theme`

manovotny (commented):
Nice work! I wouldn't normally peek into the JS PR, but I was here reviewing the docs for this anyway. 

A few things I pushed direct fixes for (578ac313926ef58887e92eeae4c080d81754d817):

- Removed stale JSDoc comment on `withClerkTheme` that still referenced removing `UIUserInterfaceStyle` when `darkColors` is present. That behavior was removed in 3c703efb8 and the changeset correctly says otherwise.
- Added top-level `isPlainObject(theme)` validation in `validateThemeJson`. Without it, a theme JSON file containing `null` crashes with a TypeError, and non-object values like `"hello"` silently pass validation. Added test cases for `null`, `string`, and `array`.

Let me know if those are incorrect and I can fix/adjust or revert.

I left comment for the other things that caught my eye. 

Small note on the PR description: it lists `fonts` as a top-level schema key, but the actual schema (and the changeset) puts `fontFamily` under `design`. Might be worth updating the description to match: `design.fontFamily` and `design.borderRadius`.

packages/expo/app.plugin.js
chriscanin: Hi @manovotny , good callout. I think that removing the theme when doing a prebuild if the theme is not present would be a good option. If the theme is removed, and is not present in the plugin, I can add a step that removes these themes from the json and plist and sets things back to default. 

Documenting the expo prebuild --clean as a necessary step after removal may be a bit confusing, I think the automatic removal on prebuild would be best as part of the plugin. And the --clean as a backup.

chriscanin:
> Nice work! I wouldn't normally peek into the JS PR, but I was here reviewing the docs for this anyway.
> 
> A few things I pushed direct fixes for ([578ac31](https://github.com/clerk/javascript/commit/578ac313926ef58887e92eeae4c080d81754d817)):
> 
> * Removed stale JSDoc comment on `withClerkTheme` that still referenced removing `UIUserInterfaceStyle` when `darkColors` is present. That behavior was removed in [3c703ef](https://github.com/clerk/javascript/commit/3c703efb8cbecf3e6ff5648d682e8f99a1dc81a7) and the changeset correctly says otherwise.
> * Added top-level `isPlainObject(theme)` validation in `validateThemeJson`. Without it, a theme JSON file containing `null` crashes with a TypeError, and non-object values like `"hello"` silently pass validation. Added test cases for `null`, `string`, and `array`.
> 
> Let me know if those are incorrect and I can fix/adjust or revert.
> 
> I left comment for the other things that caught my eye.
> 
> Small note on the PR description: it lists `fonts` as a top-level schema key, but the actual schema (and the changeset) puts `fontFamily` under `design`. Might be worth updating the description to match: `design.fontFamily` and `design.borderRadius`.

Always happy to have another set of eyes on these PR's, thank you for the feedback!!! 

- **PR description**: Updated to use `design.fontFamily` and `design.borderRadius`, and removed the stale `UIUserInterfaceStyle` removal claim.
- **Test copies vs imports**: Exported validation internals via `module.exports._testing` so the tests now `require()` the real implementation — no more copy drift. ([4786284](https://github.com/clerk/javascript/commit/4786284db))
- **Template file diverged**: Removed `ios/templates/ClerkViewFactory.swift` entirely — it was from an earlier iteration and isn't referenced anywhere. The plugin copies from `ios/ClerkViewFactory.swift` directly. ([257de36](https://github.com/clerk/javascript/commit/257de36b1))
- **`loadThemeFromAssets()` centralized**: Added a comment documenting that theme loading is intentionally centralized in `ClerkExpoModule`. The other init paths (`ClerkViewFactory.configure()`, `ClerkUserProfileActivity.onCreate()`) guard with `!Clerk.isInitialized.value`, so they only run if `ClerkExpoModule` hasn't already initialized + set the theme.
- **Theme cleanup on removal**: Added a troubleshooting note to the [theming docs](https://github.com/clerk/clerk-docs/pull/3285) explaining that `npx expo prebuild --clean` is required after removing the `theme` prop, since incremental prebuilds don't clean up previously embedded theme data.

I am now testing ios and android again with all of these changes. Things should be the same.

manovotny:
@chriscanin just a note to say that things LGTM! 👍 But I'll leave it to the pros to give you the official approval.

seanperez29 (approved):

swolfand (approved):

wobsoriano (approved):

PR #8260, opened 2026-04-07
fix(expo): inline AuthView OAuth + Android sign-out state cleanup

chriscanin:
## Description

Two related fixes for the inline `<AuthView>` and `<UserProfileView>` components in `@clerk/expo`.

### iOS, OAuth from forgot password screen

The inline `AuthView` was embedded as a child `UIHostingController` inside React Native's view hierarchy. This disrupted `ASWebAuthenticationSession` callbacks during OAuth flows started from the forgot password screen, causing the SSO sign-in to silently fail.

`ClerkAuthNativeView` now presents its `UIHostingController` via `UIViewController.present()` instead of embedding it as a child view. The visual appearance is unchanged (fullscreen, no animation) but the OAuth callback chain now completes correctly.

Also includes:

- `presentWhenReady()` uses `UIViewController.transitionCoordinator` to wait for any in-flight modal dismissal (for example, `UserProfileView` sign out) instead of a fixed delay, fixing a visible white flash on initial mount.
- `viewDidDisappear` checks the new session before returning a `cancelled` result so successful auth still propagates.

### Android, sign-out state cleanup

Three related fixes to prevent the `AuthView` from getting stuck on the "Get help / Email support" screen after sign out:

1. **Detect new sign-ins by session ID change** instead of "null to value". `ClerkAuthExpoView.initialSessionId` is captured at construction, but the view can be instantiated before `signOut` has finished clearing local state, causing it to capture a stale session ID. Switching to ID inequality lets subsequent sign-ins fire the auth completed event correctly.

2. **Per view `ViewModelStore` for `ClerkAuthExpoView`**. The clerk-android `AuthView`'s navigation `ViewModel` was scoped to the `MainActivity`, so its navigation state (for example, "Get help" destination) persisted across mount and unmount cycles within the same activity. Each `ClerkAuthExpoView` instance now provides its own `ViewModelStoreOwner` so the `AuthView` gets a fresh `ViewModel` scope per mount.

3. **Refresh client from server after sign out**. `Clerk.auth.signOut()` only clears the active session, not the in-progress `Clerk.client.signIn`. After sign out (whether via the JS bridge `signOut()`, the inline `UserProfileView`, or the modal `UserProfileActivity`), call `Client.getSkippingClientId()` to fetch a brand new client. The skipping variant is required because `Client.get()` echoes back the same `client_id` header, returning the same client with the stale `signIn` still attached.

### Files changed

- `packages/expo/ios/ClerkExpoModule.swift`, `ClerkAuthNativeView` modal presentation and `presentWhenReady`
- `packages/expo/ios/ClerkViewFactory.swift`, `viewDidDisappear` session detection
- `packages/expo/android/.../ClerkAuthExpoView.kt`, session ID change detection and per view `ViewModelStore`
- `packages/expo/android/.../ClerkExpoModule.kt`, `Client.getSkippingClientId()` after `signOut()`
- `packages/expo/android/.../ClerkUserProfileActivity.kt`, `Client.getSkippingClientId()` after `UserProfile` sign out
- `packages/expo/android/.../ClerkUserProfileExpoView.kt`, `Client.getSkippingClientId()` after inline `UserProfile` sign out

### How to test

iOS:

1. Build and run the `NativeComponentQuickstart` app on the iOS simulator.
2. Sign in with Google from the main sign in screen, expect a successful sign in.
3. Sign out, then sign in with Google from the forgot password screen, expect a successful sign in (this was the original bug).
4. Sign out via the authenticated screen button, expect a clean return to the sign in screen and the ability to sign in again.
5. Sign in, open `UserProfileView`, sign out from there, expect a clean return to the sign in screen.
6. Verify modal `presentAuth()` and `presentUserProfile()` flows still work.

Android:

1. Build and run the `NativeComponentQuickstart` app on an Android emulator.
2. Sign in with Google from the main sign in screen (uses Credential Manager).
3. Sign in with Google from the forgot password screen (uses OAuth redirect via Custom Tabs).
4. Sign out via the authenticated screen button, then sign back in.
5. Sign in, open `UserProfileView`, sign out from there, then sign back in. Verify the user is not stuck on the "Get help" screen.
6. Repeat sign in and sign out cycles to confirm no state leakage.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


chriscanin:
!snapshot

chriscanin:
!snapshot

seanperez29 (commented):
From codex review locally (not sure if a true issue):

presentWhenReady() queues deferred callbacks that are not invalidated when removeFromSuperview() runs. Because dismissAuthModal() clears presentedAuthVC, a stale queued callback can later pass the guard and call present(...), surfacing an orphaned fullscreen auth modal after the React Native view has already been removed. The fix is to invalidate pending retries on unmount, for example with an isInvalidated flag checked at the top of presentWhenReady()

chriscanin:
@seanperez29 both of these have now been addressed. 

I added an isInvalidated flag to ClerkAuthNativeView that gets set when removeFromSuperview() runs and is checked at the top of presentWhenReady(). Without this, queued callbacks from DispatchQueue.main.async or transitionCoordinator.animate could fire after the view was removed, passing the presentedAuthVC guard and presenting an orphaned fullscreen auth modal

seanperez29 (approved):

swolfand (approved):

PR #8275, opened 2026-04-08
feat(js): add clerk.oauthApplication.getConsentInfo

jfoshee:
## Description

Part of USER-3968

This is the first step in packaging/componentizing our OAuth Consent component.  We have a couple primary SDK functions we need: fetching consent info, and posting the user's consent choice.

Separate PRs will add the hooks and other necessary functions. I wanted to start with something small to make sure I am using a good pattern.

I also did a manual test of this function in a local project. I can share that if needed.

## Considerations

Is `OAuthApplication` a good name for the Resource? Or should it be just `OAuth` as was also suggested? 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/few-stamps-retire.md
+++ b/.changeset/few-stamps-retire.md
@@ -1,11 +1,7 @@
 ---
 '@clerk/clerk-js': minor
----
-
----
-'@clerk/clerk-js': minor
 '@clerk/react': minor
 '@clerk/shared': minor
 ---
 
-Add `OAuthApplication` resource and `fetchConsentInfo()` method for retrieving OAuth consent information, enabling custom OAuth consent flows.
+Add `OAuthApplication` resource and `getConsentInfo()` method for retrieving OAuth consent information, enabling custom OAuth consent flows.
--- a/.changeset/few-stamps-retire.md
+++ b/.changeset/few-stamps-retire.md
@@ -1,5 +1,5 @@
 ---
-'@clerk/clerk-js': patch
+'@clerk/clerk-js': minor
 ---
 
 Add OAuthApplication resource and fetchConsentInfo method
--- a/.changeset/few-stamps-retire.md
+++ b/.changeset/few-stamps-retire.md
@@ -2,4 +2,10 @@
 '@clerk/clerk-js': minor
 ---
 
-Add OAuthApplication resource and fetchConsentInfo method
+---
+'@clerk/clerk-js': minor
+'@clerk/react': minor
+'@clerk/shared': minor
+---
+
+Add `OAuthApplication` resource and `fetchConsentInfo()` method for retrieving OAuth consent information, enabling custom OAuth consent flows.
--- a/.changeset/few-stamps-retire.md
+++ b/.changeset/few-stamps-retire.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Add OAuthApplication resource and fetchConsentInfo method
--- a/packages/clerk-js/src/core/resources/OAuthApplication.ts
+++ b/packages/clerk-js/src/core/resources/OAuthApplication.ts
@@ -1,7 +1,7 @@
 import { ClerkRuntimeError } from '@clerk/shared/error';
 import type {
   ClerkResourceJSON,
-  FetchOAuthConsentInfoParams,
+  GetOAuthConsentInfoParams,
   OAuthConsentInfo,
   OAuthConsentInfoJSON,
 } from '@clerk/shared/types';
@@ -15,7 +15,7 @@ export class OAuthApplication extends BaseResource {
     return this;
   }
 
-  static async fetchConsentInfo(params: FetchOAuthConsentInfoParams): Promise<OAuthConsentInfo> {
+  static async getConsentInfo(params: GetOAuthConsentInfoParams): Promise<OAuthConsentInfo> {
     const { oauthClientId, scope } = params;
     const json = await BaseResource._fetch<OAuthConsentInfoJSON>(
       {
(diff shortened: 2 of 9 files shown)


packages/clerk-js/src/core/resources/OAuthApplication.ts
```diff
@@ -0,0 +1,44 @@
...
+    );
+
+    if (!json) {
+      throw new ClerkRuntimeError('Network request failed while offline', { code: 'network_error' });
+    }
+
+    const envelope = json;
+    const data = envelope.response ?? json;
+    return {
+      oauth_application_name: data.oauth_application_name,
```
jacekradko: I think this is deviating from the standard convention of having camelCase property being returned for SDK consumers

packages/clerk-js/src/core/resources/__tests__/OAuthApplication.test.ts
```diff
@@ -0,0 +1,132 @@
...
+  });
+
+  it('calls BaseResource._fetch with GET, encoded path, optional scope, and skipUpdateClient', async () => {
+    const fetchSpy = vi.spyOn(BaseResource, '_fetch').mockResolvedValue({
+      response: consentPayload,
+    } as any);
+
+    BaseResource.clerk = {} as any;
+
+    await OAuthApplication.fetchConsentInfo({ oauthClientId: 'my/client id', scope: 'openid email' });
```
dmoerner: I thought this was an odd client id but I think it's actually good to verify that we url escape this to avoid path traversal issues.

packages/clerk-js/src/core/resources/OAuthApplication.ts
```diff
@@ -0,0 +1,44 @@
...
+      },
+      { skipUpdateClient: true },
+    );
+
+    if (!json) {
+      throw new ClerkRuntimeError('Network request failed while offline', { code: 'network_error' });
+    }
+
+    const envelope = json;
+    const data = envelope.response ?? json;
```
dmoerner: Why do we coalesce here?

packages/shared/src/types/oauthApplication.ts
```diff
@@ -0,0 +1,54 @@
...
+  oauth_application_url: string;
+  client_id: string;
+  state: string;
+  scopes: OAuthConsentScopeJSON[];
+}
+
+export type FetchOAuthConsentInfoParams = {
+  /** OAuth `client_id` from the authorize request. */
+  oauthClientId: string;
+  /** Optional normalized scope string (e.g. space-delimited). */
```
dmoerner: "e.g." means "for example". Is space-delimited just an example of a scope string or is this something we should enforce?

packages/clerk-js/src/core/resources/OAuthApplication.ts
jfoshee: Seems to be the common pattern that callers to `_fetch` use `.response`. I assume that is for client piggybacking. We don't do that for this endpoint (it just uses `json` directly). But I guess it's good to make sure it will work either way.  I added a unit test so we cover both cases.

packages/shared/src/types/oauthApplication.ts
jfoshee: 😆  good catch

packages/clerk-js/src/core/resources/OAuthApplication.ts
jfoshee: ah yes, thanks @jacekradko

dmoerner (approved):
This looks good to me.

packages/clerk-js/src/core/resources/OAuthApplication.ts
```diff
@@ -0,0 +1,49 @@
...
+import { BaseResource } from './internal';
+
+export class OAuthApplication extends BaseResource {
+  pathRoot = '';
+
+  protected fromJSON(_data: ClerkResourceJSON | null): this {
+    return this;
+  }
+
+  static async fetchConsentInfo(params: FetchOAuthConsentInfoParams): Promise<OAuthConsentInfo> {
```
wobsoriano: Small nit: we don't use a fetch* prefix elsewhere in resources (closest precedents for non-standard verbs are `Waitlist.join` and `Passkey.registerPasskey`). 

Not blocking, just wondering if `getConsentInfo` / `retrieveConsentInfo` or similar would sit better with the rest of the codebase. Happy to keep as-is if you prefer.

.changeset/few-stamps-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/clerk-js': minor
+---
+
+Add OAuthApplication resource and fetchConsentInfo method
```
wobsoriano:
```suggestion
---
'@clerk/clerk-js': minor
'@clerk/react': minor
'@clerk/shared': minor
---

Add `OAuthApplication` resource and `fetchConsentInfo()` method for retrieving OAuth consent information, enabling custom OAuth consent flows.
```

wobsoriano (commented):
looks good already! left a tiny comment

packages/clerk-js/src/core/resources/OAuthApplication.ts
jfoshee: Oh cool, good catch. Looks like `get` is more common e.g. `getSessions`, `getToken`, `getDomains`, etc.  I'll update it!

.changeset/few-stamps-retire.md
```diff
@@ -0,0 +1,11 @@
+---
+'@clerk/clerk-js': minor
+---
+
```
alexcarpenter:
```suggestion
```

jfoshee: oops, fixing that w/ the rename

.changeset/few-stamps-retire.md
jfoshee: Ah good catch, thank you.

wobsoriano (approved):
thanks!

PR #8276, opened 2026-04-09
docs(repo): Add typedoc comments; generate typedoc output for /objects docs

alexisintech:
⚠️  In conjunction with [Generate object reference docs](https://github.com/clerk/clerk-docs/pull/3286#top)

## Description

For the [`/objects`](https://clerk.com/docs/reference/objects/overview) in clerk-docs, we want to generate the reference information. In summary, this PR's main change is that:
When typedoc goes to generate files, it creates a folder for each defined object (which we define in `reference-objects.mjs`. E.g. `clerk` object gets `.typedoc/docs/shared/clerk` folder. This folder will have a `<OBJECTNAME>-properties.mdx` file and a `<OBJECTNAME>-methods` folder that includes each of the methods.

In a little more detail:

In `reference-objects.mjs`, we define these objects by defining the Typedoc output paths and by defining the primary interface/class on each reference object page in order to resolve TypeDoc reflections.

The `extract-methods.mjs` script will extract all of the methods from that object's primary interface/class and create a dedicated file for each one. I won't go into lengthy detail here, but the file is heavily commented through.

In `custom-router.mjs`
- Adds logic to ensure that for the defined `REFERENCE_OBJECT_PAGE_SYMBOLS`, dedicated folders are created for each object. So `clerk.mdx` no longer gets added to `.typedoc/docs/shared/`, it gets added to its own folder: `.typedoc/docs/shared/clerk/`

In `custom-plugin.mjs`
- Updates all regex so that links are never added to headings (# - #######)
- Adds `applyRelativeLinkReplacements()` helper that the `extract-methods.mjs` script uses to incorporate the link replacements into the files it generates
- Adds `applyCatchAllMdReplacements()` helper that the `extract-methods.mjs` script uses to incorporate the catch-all replacements into the files it generates
- Adds `stripReferenceObjectPropertiesSection()` which removes the **Properties** section from a generated file and puts it into its own file. For example, if `shared/clerk/clerk.mdx` is generated, it pulls the **Properties** section from it and puts it into its own `shared/clerk/clerk-properties.mdx` file.

`comment-utils.mjs` is used as a helper to remove "TODO" and its following text from generated content.

`standalone-page-tag.mjs` adds support for a custom tag called `@standalonePage` that can be used in conjunction with the `@inline` tag to indicate to typedoc to still generate a dedicated page for that type, even though it is recognized as `@inline`. Using both of these tags means that the type will show up as inlined in a table, but will also still have a generated MDX file.

Adds support for a custom tag `@generateWithEmptyComment`: self-documenting placeholder for declarations intentionally left without a description.

Notable additions:
- Parameters that accept objects will include the object's properties in the table as children of that parameter. 
  See the following examples (with before on left, after on right):
  Screenshot 2026-04-13 at 18 55 29
   Screenshot 2026-04-17 at 18 09 57
- If a parameter links to a dedicated page, parameter children (a?.b, a?.c, a?.d) are not shown in the table.
  Screenshot 2026-04-17 at 18 11 16
- If a method accepts only one parameter, the parameters section will be of that type. E.g. for `joinWaitlist()`, it accepts a params object of type `JoinWaitlistParams`. Instead of a "Parameters" section, it has a "JoinWaitlistParams" section:
  Screenshot 2026-04-13 at 18 59 45

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.typedoc/reference-objects.mjs
```diff
@@ -0,0 +1,21 @@
...
+export const REFERENCE_OBJECTS_LIST = ['shared/clerk/clerk.mdx', 'shared/client-resource/client-resource.mdx'];
+
+/**
+ * Primary interface/class documented on each reference object page (used to resolve TypeDoc reflections).
+ * keys = stable MDX paths under .typedoc output
+ * values = which TypeDoc declaration to treat as that page’s “main” type for routing and per-method extraction.
+ */
+export const REFERENCE_OBJECT_PAGE_SYMBOLS = {
+  'shared/clerk/clerk.mdx': 'Clerk',
+  'shared/client-resource/client-resource.mdx': 'ClientResource',
```
SarahSoutoul: This is causing an issue in the Typedoc output - it's incorrectly linking `ClientResource` in the the `shared/clerk-options` typedoc (see below). 

Image

Causing a 404 on the doc: 

https://github.com/user-attachments/assets/57469fc6-f2b7-4d42-8a67-488bdb6084fb

When removing it, it resolves to the correct link: 

Image

.typedoc/custom-plugin.mjs
```diff
@@ -99,6 +99,10 @@ const LINK_REPLACEMENTS = [
   ['deleted-object-resource', '/docs/reference/types/deleted-object-resource'],
   ['checkout-flow-resource', '/docs/reference/hooks/use-checkout#checkout-flow-resource'],
   ['organization-creation-defaults-resource', '#organization-creation-defaults-resource'],
+  ['billing-namespace', '/docs/reference/objects/billing'],
+  ['client-resource', '/docs/reference/objects/client'],
+  ['redirect-options', '/docs/reference/types/redirect-options'],
+  ['handle-o-auth-callback-params', '/docs/reference/types/handle-o-auth-callback-params'],
```
SarahSoutoul, line 127: Needs a `['session-task', '/docs/reference/types/session-task']` to properly link this one on the `shared/clerk-options ` typedoc output.

Image

SarahSoutoul: Okay ended up fixing this in my [docs review](https://github.com/clerk/javascript/pull/8276/commits/6fef09a71a46663f417ba6c65c627a89356fe583).

.typedoc/reference-objects.mjs
SarahSoutoul: Ended up solving this in my [docs review](https://github.com/clerk/javascript/pull/8276/commits/6fef09a71a46663f417ba6c65c627a89356fe583).

SarahSoutoul:
@alexisintech I have pushed a [docs review](https://github.com/clerk/javascript/pull/8276/commits/6fef09a71a46663f417ba6c65c627a89356fe583) with a few quick / straightforward fixes, but let me know if anything doesn't look right. It does the following:

- Adds a link replacement for `SessionTask` as it wasn't properly linked and would lead to a 404 when clicking on it.
- Changed some of the logic within  `getRelativeLinkReplacements` to take into account nested object-doc links to fix the issue raised [here](https://github.com/clerk/javascript/pull/8276#discussion_r3088444134). 
- Added a replacement for `LastAuthenticationStrategy` as it wasn't linked at all, as it should comparing to the current docs.
- For the Clerk object, added a more detailed intro than `Main Clerk SDK object` which can then be used as a Typedoc output.
- Fixed the link for `authenticateWithGoogleOneTap()`, as it would appear as an external link as such and be 404 when clicked (see what I mean below). Since the destination exists on the same page we're on, it just needed to be an hashtag. 

Screenshot 2026-04-15 at 3 08 44 pm

Have found some other issues that I wasn't sure about but will make a separate comment for those.

SarahSoutoul:
@alexisintech Okay here are the issues I have found while testing:

1. The tooltip isn't translating as it should on the Clerk page - instead of being displayed as a tooltip, it gets displayed as a link and leads to 404 when clicked. See below (left locally, right live docs). 
 
    Screenshot 2026-04-15 at 2 26 30 pm

2. Another link (screenshot 1) that leads to 404, mostly because it doesn't exist. The params are now listed right below with the changes (screenshot 2) vs being separate on the current docs (screenshot 3). 
     
    **Screenshot 1**

    Screenshot 2026-04-15 at 3 12 06 pm
     
     **Screenshot 2**

    Screenshot 2026-04-15 at 5 45 46 pm
    
     **Screenshot 3**

    Screenshot 2026-04-15 at 5 45 57 pm

3. Not an issue per se, but wondering if this (screenshot 1) is purely for documentation to exist on the `javascript` repo, because it doesn't create a Typedoc output and isn't being used in the docs. That info is inputted manually in the docs (screenshot 2)

     **Screenshot 1**
   
    Screenshot 2026-04-15 at 4 38 25 pm

     **Screenshot 2**
    
    Screenshot 2026-04-15 at 4 37 26 pm

4. Some of the properties that caught my attention - first one, not sure if we want a description and second one, the type just says `any` but was expecting what we have on the current docs (screenshot 3) 

     **Screenshot 1**

    Screenshot 2026-04-15 at 3 02 11 pm

     **Screenshot 2**

    Screenshot 2026-04-15 at 3 02 00 pm
     
     **Screenshot 3**

    Screenshot 2026-04-15 at 5 44 11 pm

These are all spotted on the Clerk object page, but the Client object page looked quite good to me at first glance. The Clerk page was quite meaty to go through and test, so I may have missed some stuff but will give it a second look for peace of mind.

alexisintech:
> @alexisintech Okay here are the issues I have found while testing:
> 
> 1. The tooltip isn't translating as it should on the Clerk page - instead of being displayed as a tooltip, it gets displayed as a link and leads to 404 when clicked. See below (left locally, right live docs).
>     Screenshot 2026-04-15 at 2 26 30 pm
> 2. Another link (screenshot 1) that leads to 404, mostly because it doesn't exist. The params are now listed right below with the changes (screenshot 2) vs being separate on the current docs (screenshot 3).
>    **Screenshot 1**
>     Screenshot 2026-04-15 at 3 12 06 pm
>    **Screenshot 2**
>     Screenshot 2026-04-15 at 5 45 46 pm
>    **Screenshot 3**
>     Screenshot 2026-04-15 at 5 45 57 pm
> 3. Not an issue per se, but wondering if this (screenshot 1) is purely for documentation to exist on the `javascript` repo, because it doesn't create a Typedoc output and isn't being used in the docs. That info is inputted manually in the docs (screenshot 2)
>    **Screenshot 1**
>     Screenshot 2026-04-15 at 4 38 25 pm
>    **Screenshot 2**
>     Screenshot 2026-04-15 at 4 37 26 pm
> 4. Some of the properties that caught my attention - first one, not sure if we want a description and second one, the type just says `any` but was expecting what we have on the current docs (screenshot 3)
>    **Screenshot 1**
>     Screenshot 2026-04-15 at 3 02 11 pm
>    **Screenshot 2**
>     Screenshot 2026-04-15 at 3 02 00 pm
>    **Screenshot 3**
>     Screenshot 2026-04-15 at 5 44 11 pm
> 
> These are all spotted on the Clerk object page, but the Client object page looked quite good to me at first glance. The Clerk page was quite meaty to go through and test, so I may have missed some stuff but will give it a second look for peace of mind.

1. yeah the tooltip issue is on my agenda 😢 
2. yep - i'm still creating some of those docs, they are in a list on my sticky note. but good eye! thank you for pointing out 🙏 
3. yes it's not getting output by typedoc so i have it manually added, and then that line you highlighted we can stick in an `<If sdk="js-frontend">` so good catch! not sure what the question is though - are you wondering if i meant to do that??
4. yeah not sure what's happening with the appearance output - I can take a look but if the type is outputting as `any` then we can simply link to the appearance docs in the description of the prop instead of the type column, so that the user can still access that info. and yes let me look into the `clerkUI` one! good catches 😸💖

packages/shared/src/types/clerk.ts
```diff
@@ -1003,7 +1003,7 @@ export interface Clerk {
   redirectToTasks(opts?: TasksRedirectOptions): Promise<unknown>;
 
   /**
-   * Completes a Google One Tap redirection flow started by [`authenticateWithGoogleOneTap()`](https://clerk.com/reference/objects/clerk#authenticate-with-google-one-tap). This method should be called after the user is redirected back from visiting the Google One Tap prompt.
+   * Completes a Google One Tap redirection flow started by [`authenticateWithGoogleOneTap()`](#authenticate-with-google-one-tap). This method should be called after the user is redirected back from visiting the Google One Tap prompt.
```
alexisintech: we have to keep full links in the comments! let's revert this 🙏

alexisintech:
@SarahSoutoul should I link it on `object` or should I add a sentence: See the [Appearance](#) docs for more information.
Screenshot 2026-04-16 at 15 10 41

packages/shared/src/types/clerk.ts
alexisintech: done here: [c99e565](https://github.com/clerk/javascript/pull/8276/commits/c99e565641ef7cbe194452db697ffcaea38f91d4)

alexisintech:
fixed clerkui and appearance descriptions [c99e565](https://github.com/clerk/javascript/pull/8276/commits/c99e565641ef7cbe194452db697ffcaea38f91d4) - we can update the appearance one how we want once you reply with your thoughts @SarahSoutoul 

i also wrapped the `constructor()` method in an `If sdk="js-frontend"` in the `clerk-docs` PR [4f89cc7](https://github.com/clerk/clerk-docs/pull/3286/commits/4f89cc7aedecb515c88fac8cc91dd932c65f433f)

SarahSoutoul:
> @SarahSoutoul should I link it on `object` or should I add a sentence: See the [Appearance](#) docs for more information. Screenshot 2026-04-16 at 15 10 41

@alexisintech I would go with the second option, actually! Have a: See the [Appearance](#) docs for more information. 

For this:

> yes it's not getting output by typedoc so i have it manually added, and then that line you highlighted we can stick in an <If sdk="js-frontend"> so good catch! not sure what the question is though - are you wondering if i meant to do that??

My question was more around okay we're documenting it in the `javascript` repo, but it doesn't output any Typedoc and isn't being used in the docs. So is it just there as good documentation to exist in the `javascript` repo, aka is it needed there bcs if it isn't, it could just be removed, since we're not using it in our docs? It's not a massive issue at all and can be left alone - more of a curiosity question. 

Everything else looks resolved! I will give it a second review once you add the extra stuff.

alexisintech:
posting this here to track:
non-blocking UI issue but sometimes the `type` values, when they are links, have the gray backgrounds (like inline code usually does) but when they are in unions, e.g. `a | b`, they don't have the gray background
Screenshot 2026-04-17 at 18 16 11

alexisintech:
> posting this here to track: non-blocking UI issue but sometimes the `type` values, when they are links, have the gray backgrounds (like inline code usually does) but when they are in unions, e.g. `a | b`, they don't have the gray background Screenshot 2026-04-17 at 18 16 11

ahhh it's because of our `custom-plugin` and the way we are replacing links. sometimes the `replace` value has backticks in it, sometimes not. I vote to **not** have them as the tables will format these cells accordingly - if we do add backticks, it sometimes breaks the formatting e.g. `Web3WalletResource[]` will have a highlighted background for only `Web3WalletResource` and leave `[]` without the highlighted background:

Screenshot 2026-05-04 at 16 22 21

.typedoc/custom-plugin.mjs
```diff
@@ -132,111 +142,205 @@ const LINK_REPLACEMENTS = [
 function getRelativeLinkReplacements() {
   return LINK_REPLACEMENTS.map(([fileName, newPath]) => {
     return {
-      // Match both path and optional anchor
-      pattern: new RegExp(`\\((?:(?:\\.{1,2}\\/)+[^()]*?|)${fileName}\\.mdx(#[^)]+)?\\)`, 'g'),
+      // Match both flat links and nested object-doc links
```
alexisintech, line 149:
```suggestion
      // Match both flat links and nested object-doc links
      // Also matches optional anchors (#)
```

SarahSoutoul:
Gonna investigate but getting these errors on the corresponding [`clerk-docs` PR](https://github.com/clerk/clerk-docs/pull/3286), at first glance: 

Screenshot 2026-05-11 at 3 51 50 pm

alexisintech:
> Gonna investigate but getting these errors on the corresponding [`clerk-docs` PR](https://github.com/clerk/clerk-docs/pull/3286), at first glance:
> 
> Screenshot 2026-05-11 at 3 51 50 pm

aww those links look incorrect - i'll check it out!

alexisintech:
okay I've fixed most errors, except 
- Agent Tasks which has an open PR and I asked about it in Slack https://clerkinc.slack.com/archives/C084WHCNHCZ/p1778540498107899 - but if anything, we can remove the links and solve these errors
- These errors:
Screenshot 2026-05-11 at 16 14 50
The `shared/paginated-hook-config.mdx not found` errors are caused because it relies on `PaginatedHookConfig` which has `@inline`. And in this PR, `@inline` no longer allows generation for standalone pages. If a type is inlined, then that means in tables its getting inlined and does not need its own page generated. 
The `Typedoc shared/hook-params.mdx not found` error is basically the same issue, the `HookParams` type uses the `PaginatedHookConfig` type, which is tagged with `@inline` and so it doesn't get generated.
We can do one of two things: revert this behavior (which will cause side effects in this PR but solve the errors that are currently appearing), or come up with a fix for it in this PR. We can handle the side effects in this PR - I guess it'd take investigating which is the easier solution. I can start looking into it

SarahSoutoul:
@alexisintech just pushed this [Fix external link](https://github.com/clerk/javascript/pull/8276/commits/16d4d98f5597cffb64c82a788e8696969d765a79) as the external link wasn't working.

alexisintech:
> make frontend api url open in new tab

sorry was still generating to test, and in between tasks, but thanks for catching! 🤦‍♀️

packages/shared/src/types/apiKeys.ts
```diff
@@ -2,36 +2,95 @@ import type { CreateAPIKeyParams, GetAPIKeysParams, RevokeAPIKeyParams } from '.
...
   /**
    * Creates a new API key.
+   * @returns An [`APIKeyResource`](https://clerk.com/docs/reference/types/api-key-resource) object that includes the `secret` property.
+   * > [!WARNING]
+   * > Make sure to store the API key secret immediately after creation, as it will not be available again.
    */
   create(params: CreateAPIKeyParams): Promise<APIKeyResource>;
   /**
    * Revokes a given API key by ID.
+   * @returns An [`APIKeyResource`](https://clerk.com/docs/reference/types/api-key-resource) object.
```
manovotny, line 91: The new `reference-objects.mjs` maps `APIKeyResource` to `/docs/reference/objects/api-keys`, but these `@returns` links still point at `/docs/reference/types/api-key-resource`. 

Will that page still exist after this lands, or are these about to 404?

Same on lines 86.

turbo.json
```diff
@@ -339,7 +339,15 @@
...
-      "inputs": ["tsconfig.typedoc.json", "typedoc.config.mjs"],
+      "inputs": [
+        "tsconfig.typedoc.json",
+        "typedoc.config.mjs",
+        ".typedoc/reference-objects.mjs",
+        ".typedoc/comment-utils.mjs",
+        ".typedoc/extract-methods.mjs",
+        ".typedoc/extract-returns-and-params.mjs",
+        ".typedoc/standalone-page-tag.mjs"
+      ],
```
manovotny: The typedoc output also depends on `custom-plugin.mjs`, `custom-router.mjs`, `custom-theme.mjs`, and `prepare-markdown-renderer.mjs` — all of which this PR changes — but none are listed here. 

Cached `typedoc:generate` runs will reuse stale output after edits to renderer/router logic. Maybe broaden to a glob?

```suggestion
      "inputs": [
        "tsconfig.typedoc.json",
        "typedoc.config.mjs",
        ".typedoc/**/*.mjs"
      ],
```

.typedoc/reference-objects.mjs
```diff
@@ -0,0 +1,69 @@
...
+  'shared/api-key-resource/api-key-resource.mdx': {
+    symbol: 'APIKeyResource',
+    declarationHint: 'types/apiKeys',
+    extraMethodInterfaces: [{ symbol: 'APIKeysNamespace', declarationHint: 'types/apiKeys' }],
+  },
+  'shared/billing-namespace/billing-namespace.mdx': {
+    symbol: 'BillingNamespace',
+    declarationHint: 'types/billing',
+    extraMethodInterfaces: [{ symbol: 'BillingNamespace', declarationHint: 'types/billing' }],
+  },
```
manovotny, line 56: `extractMethodsForPage()` already runs `extractCallableMembersFromDeclaration` on the primary `symbol`, then loops over `extraMethodInterfaces`. 

Here both resolve to the same `BillingNamespace` declaration, so every method file gets written twice and the returned `count` doubles. 

Maybe a copy-paste from the `APIKeyResource` entry above? 

```suggestion
  'shared/billing-namespace/billing-namespace.mdx': {
    symbol: 'BillingNamespace',
    declarationHint: 'types/billing',
  },
```

packages/shared/src/types/clerk.ts
```diff
@@ -203,40 +217,66 @@ type EventHandler<E extends ClerkEvent> = (payload: ClerkEventPayload[E]) => voi
...
-   * - `"loading"`: Clerk is still attempting to load.
-   * - `"ready"`: Clerk singleton is fully operational.
-   * - `"degraded"`: Clerk singleton is partially operational.
+   * The status of the `Clerk` instance. Possible values are:
+   * <ul>
+   *  <li>`"error"`: Set when hotloading `clerk-js` or `Clerk.load()` failed.</li>
+   *  <li>`"loading"`: Set during initialization.</li>
+   *  <li>`"ready"`: Set when Clerk is fully operational.</li>
+   *  <li>`"degraded"`: Set when Clerk is partially operational.</li>
+   * </ul>
```
manovotny, line 275: The pre-existing pattern in this codebase (and the CONTRIBUTING styleguide examples) is markdown lists rather than HTML `<ul>`/`<li>`. There are a handful of other spots in the same vein — `client.ts:19-23`, `billing.ts:473/483/731`, `signInFuture.ts:327/366/414`, `signUpFuture.ts:48/320`. 

Intentional because the markdown won't render correctly once it's pulled into `clerk-docs`?

```suggestion
   * - `"error"`: Set when hotloading `clerk-js` or `Clerk.load()` failed.
   * - `"loading"`: Set during initialization.
   * - `"ready"`: Set when Clerk is fully operational.
   * - `"degraded"`: Set when Clerk is partially operational.
```

.typedoc/custom-theme.mjs
```diff
@@ -1,6 +1,1007 @@
...
+import { ArrayType, i18n, IntersectionType, ReferenceType, ReflectionKind, ReflectionType, UnionType } from 'typedoc';
 import { MarkdownTheme, MarkdownThemeContext } from 'typedoc-plugin-markdown';
+import {
+  backTicks,
+  heading,
+  htmlTable,
+  table,
+} from '../node_modules/typedoc-plugin-markdown/dist/libs/markdown/index.js';
+import { removeLineBreaks } from '../node_modules/typedoc-plugin-markdown/dist/libs/utils/index.js';
+import { TypeDeclarationVisibility } from '../node_modules/typedoc-plugin-markdown/dist/options/maps.js';
```
manovotny: Is there a reason these need to reach into `../node_modules/...` and `dist`? That can be fragile if the package restructures it's code or changes output patterns.

.typedoc/comment-utils.mjs
```diff
@@ -0,0 +1,93 @@
...
+  }
+  return chunks.some(text => TODO_WORD.test(text));
+}
+
+/**
+ * Truncate at the first word "TODO" (case-insensitive). Used when flattening display parts to a string.
+ *
+ * @param {string} text
+ */
+export function stripTextAfterTodo(text) {
```
manovotny: I think `stripTextAfterTodo` is unused. Delete?

.typedoc/comment-utils.mjs
```diff
@@ -0,0 +1,93 @@
+// @ts-check
+import { Comment } from 'typedoc';
+
+const TODO_WORD = /\bTODO\b/i;
+
+/**
+ * @param {import('typedoc').Comment | undefined} comment
+ */
+export function commentContainsTodo(comment) {
```
manovotny: It looks like `commentContainsTodo` is only used inside this file. Remove the `export`?

.typedoc/extract-methods.mjs
```diff
@@ -0,0 +1,1497 @@
...
+  IntersectionType,
+  OptionalType,
+  PageKind,
+  ReferenceType,
+  ReflectionKind,
+  ReflectionType,
+  UnionType,
+} from 'typedoc';
+import { MarkdownPageEvent, MarkdownTheme } from 'typedoc-plugin-markdown';
+import { removeLineBreaks } from '../node_modules/typedoc-plugin-markdown/dist/libs/utils/index.js';
```
manovotny: Same `node_modules` and `dist` comments here.

packages/shared/src/types/redirects.ts
```diff
@@ -1,12 +1,18 @@
 import type { EnterpriseSSOStrategy, OAuthStrategy } from './strategies';
 
+/**
+ *
+ */
```
manovotny: Why are there several empty comments like this one in the file?

In the following files too:

* packages/shared/src/types/resource.ts
* packages/shared/src/types/session.ts
* packages/shared/src/types/signInFuture.ts
* packages/shared/src/types/signUpFuture.ts
* packages/shared/src/types/user.ts
* packages/shared/src/types/waitlist.ts

(I probably missed a few other files)

packages/shared/src/types/clerk.ts
```diff
@@ -203,40 +217,66 @@ type EventHandler<E extends ClerkEvent> = (payload: ClerkEventPayload[E]) => voi
...
+ * @param handler - The callback function to remove.
+ */
+export type OffEventListener = <E extends ClerkEvent>(event: E, handler: EventHandler<E>) => void;
 
 /**
  * @inline
+ * @property {ClerkStatus} degraded - Set when Clerk is partially operational.
+ * @property {ClerkStatus} error - Set when hotloading `clerk-js` or `Clerk.load()` failed.
+ * @property {ClerkStatus} loading - Set during initialization.
+ * @property {ClerkStatus} ready - Set when Clerk is fully operational.
```
manovotny: `@property` documents object members, but `ClerkStatus` is a string-literal union — TypeDoc probably won't render these tags the way you want.

The new `clerkLiteralUnionAsPropertiesTable` in `custom-theme.mjs` already consumes per-arm prose via `UnionType.elementSummaries`, which means an inline `/** … */` before each `|` arm would land in the rendered Properties table... I think?

manovotny (commented):
Here's some things that caught my eye.

packages/shared/src/types/clerk.ts
alexisintech: yes, intentional for the reasoning you mentioned!

.typedoc/custom-theme.mjs
alexisintech: ahh didn't notice this at first, but i'm currently refactoring in another branch >:)

packages/shared/src/types/redirects.ts
alexisintech: it's to tell typedoc to generate them. typedoc requires jsdoc comments as a signal that something needs to get generated. these don't always require anything written - if it makes more sense to fill these with descriptions, we can (but will need to test that it doesn't break any output). should we fill these comments so that no one else is confused?

packages/shared/src/types/clerk.ts
alexisintech: yeahhhh I ended up just relying on putting the list in the JSdoc comment on line 273 - so this can get removed 🤷‍♀️

alexisintech:
i've got the refactor here ready: [refactor(repo): Inline extract-methods + tighten parameter rendering](https://github.com/clerk/javascript/pull/8637/changes#top)

@manovotny

packages/shared/src/types/redirects.ts
alexisintech: so unfortunately, filling those comments **will** affect the output files and add the descriptions to the top of the files, which is something we don't want because we use those property tables as they are

Screenshot 2026-05-22 at 17 47 18

packages/shared/src/types/apiKeys.ts
alexisintech: the apikeys object and the apikeysresource are two different things! here's the latter: https://clerk.com/docs/nextjs/reference/types/api-key-resource

.typedoc/reference-objects.mjs
alexisintech: fixed [39b8139](https://github.com/clerk/javascript/pull/8276/commits/39b81393c694e3070b94e8076ec8f711a533897e)

packages/shared/src/types/clerk.ts
alexisintech: fixed [95660d9](https://github.com/clerk/javascript/pull/8276/commits/95660d9096ce1c14875e9f10239b64135c2705ff)

.typedoc/custom-theme.mjs
alexisintech: handled in [refactor(repo): Inline extract-methods + tighten parameter rendering](https://github.com/clerk/javascript/pull/8637/changes#top)

.typedoc/comment-utils.mjs
alexisintech: nice catch! handled in [refactor(repo): Inline extract-methods + tighten parameter rendering](https://github.com/clerk/javascript/pull/8637/changes#top)

.typedoc/extract-methods.mjs
alexisintech: handled in [refactor(repo): Inline extract-methods + tighten parameter rendering](https://github.com/clerk/javascript/pull/8637/changes#top) !

alexisintech:
unsafeMetadata section added successfully:

Screenshot 2026-05-27 at 15 04 12

packages/shared/src/types/redirects.ts
alexisintech: I think something slightly better is we can change it from 
```mdx
/**
 *
 */
```
to
```mdx
/** */
```

I can do it once the other PR gets pulled in here so its not twice the work or causing any merge conflicts

manovotny:
Pushed some changes directly in 9be16f7b4.

- Broadened `typedoc:generate` cache inputs in `turbo.json` to `.typedoc/**/*.mjs` so edits to `custom-plugin.mjs`, `custom-router.mjs`, `custom-theme.mjs`, and `prepare-markdown-renderer.mjs` invalidate the cache.
- Dropped the export on `commentContainsTodo` in `.typedoc/comment-utils.mjs`.

.typedoc/custom-theme.mjs
manovotny: Handled in #8637 (commit `ff3595c80`). Resolving.

packages/shared/src/types/redirects.ts
manovotny: That would be better. I'd still like to see if we can remove them altogether or have a placeholder reason. But we can revisit both later.

manovotny (approved):
I'm good here, given the incoming changes in https://github.com/clerk/javascript/pull/8637 on top of this PR.

packages/shared/src/types/redirects.ts
alexisintech: yeah we can probably create a custom tag for this actually. like @generateTypedocWithoutCommentBody or @generateWithoutCommentBody - what do you think about that naming?

packages/shared/src/types/telemetry.ts
```diff
@@ -57,9 +57,24 @@ export interface TelemetryLogEntry {
...
+ */
 export interface TelemetryCollector {
+  /**
+   * If `true`, telemetry events are only logged to the console and not sent to Clerk.
+   */
   isEnabled: boolean;
+  /**
+   * If `true`, telemetry events are only logged to the console and not sent to Clerk.
+   */
   isDebug: boolean;
```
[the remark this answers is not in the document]
alexisintech: [apply coderabbit suggestions](https://github.com/clerk/javascript/pull/8276/commits/8c9be501f950fd34bc636a13787c80cd4422c289)

packages/shared/src/types/multiDomain.ts
```diff
@@ -60,13 +60,16 @@ export type MultiDomainAndOrProxyPrimitives =
...
-   * **Required for applications that run behind a reverse proxy**. The URL that Clerk will proxy requests to. Can be either a relative path (`/__clerk`) or a full URL (`https://<your-domain>/__clerk`).
+   * **Required for applications that run behind a reverse proxy**. The URL that Clerk will proxy requests to. Can be either a relative path (`/__clerk`) or a full URL (`https://<your-domain>/__clerk`), or a function that will be called with a `URL` made from `window.location.href`.
    */
   proxyUrl?: string | ((url: URL) => string);
   /**
-   * **Required if your application is a satellite application**. Sets the domain of the satellite application.
+   * **Required if your application is a satellite application**. Sets the domain of the satellite application. Can be either a relative path (`/__clerk`) or a full URL (`https://<your-domain>/__clerk`), or a function that will be called with a `URL` made from `window.location.href`.
    */
   domain?: string | ((url: URL) => string);
 };
```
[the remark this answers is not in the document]
alexisintech: [apply coderabbit suggestions](https://github.com/clerk/javascript/pull/8276/commits/8c9be501f950fd34bc636a13787c80cd4422c289)

alexisintech:
@jacekradko hey hey, got some scary warnings here from snapi but here's what my agent is telling me:

```mdx
this is a false positive caused by a recent snapi version bump on main, not anything in your PR.

Evidence:
- All 19 "removed" properties have an @internal JSDoc tag (verified locally)
- Your branch's source has them all (no removal in git diff main...HEAD)
- Your local build emits all 86 __internal_* symbols in dist/types/index.d.ts
- The snapi tool was bumped multiple times in the last 24h on main:
  - a3740fca3c snapi bump + cache changes (yesterday)
  - 17925fee39 enabled snapi AI reviewer
  - e42b48eceb, 5283c8cb90, d679eb1e2b further bumps
  - be63baad54 extended coverage to all SDK packages

The most likely cause: the new snapi version strips @internal-tagged symbols from the analyzed API, but the cached baseline (from before the bump) still includes them. Result: every @internal symbol appears as "removed" in any PR opened during this window.
```

just wanna make sure i'm okay 😨

.typedoc/comment-utils.mjs
```diff
@@ -0,0 +1,77 @@
...
+        const before = p.text.slice(0, match.index).trimEnd();
+        if (before.length) {
+          out.push(/** @type {import('typedoc').CommentDisplayPart} */ ({ kind: 'text', text: before }));
+        }
+        return out;
+      }
+    }
+    out.push(p);
+  }
+  return out;
```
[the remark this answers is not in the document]
alexisintech: specifically ignoring this! 

the proposed fix is risky:
- For inline-tag parts, text is the rendered label, not arbitrary prose. {@link Foo TODO bar} becomes text: 'Foo', but the part still carries tag: '@link', target: ... — render might be [Foo](...) which is OK, or might emit a half-formed link depending on the markdown plugin.
- For code parts, `someCodeTODOmore` → `someCode` is valid markup, but you've silently stripped inside a code span the author wrote literally.
- Truncating inside non-prose parts changes user-authored content in a place authors don't expect doc-tooling to edit.

PR #8280, opened 2026-04-09
fix(ui): remove back button from sign-in password compromised/pwned error screen

Ephem:
## Description

These errors are not recoverable by re-entering the password, so the back button led to a confusing dead end that would always take you back to the same error.

Before:

CleanShot 2026-04-09 at 16 24 16

After:

CleanShot 2026-04-09 at 16 23 32

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/ui/src/components/SignIn/SignInFactorOne.tsx
+++ b/packages/ui/src/components/SignIn/SignInFactorOne.tsx
@@ -155,7 +155,8 @@ function SignInFactorOneInternal(): JSX.Element {
   }
 
   if (showAllStrategies || showForgotPasswordStrategies) {
-    const canGoBack = factorHasLocalStrategy(currentFactor);
+    // Password errors are not recoverable by re-entering the password, so we hide the back button
+    const canGoBack = factorHasLocalStrategy(currentFactor) && !passwordErrorCode;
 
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/SignIn/SignInFactorOne.tsx
```diff
@@ -155,7 +155,8 @@ function SignInFactorOneInternal(): JSX.Element {
   }
 
   if (showAllStrategies || showForgotPasswordStrategies) {
-    const canGoBack = factorHasLocalStrategy(currentFactor);
+    // Password errors are not recoverable by re-entering the password, so we hide the back button
+    const canGoBack = factorHasLocalStrategy(currentFactor) && !passwordErrorCode;
```
dmoerner, line 159: `passwordErrorCode` is to me not a very clear name for this flag (it sounds more general than it is) but I confirmed that it is just a union of `'compromised' | 'pwned'` so it's exactly what we want to check.

dmoerner (approved):

wobsoriano (approved):

PR #8285, opened 2026-04-10
chore(ui): Localize OAuthConsent inline actions

alexcarpenter:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


wobsoriano (approved):
Looks good! I like the approach already where `<InlineAction>` finds the action text in the resolved string so each locale stays as one natural sentence and the tooltip still lands in the right place 👍🏼 

Image

PR #8286, opened 2026-04-10
feat(shared,nextjs,react): Introduce `useOAuthConsent` hook

jfoshee:
## Description

This PR introduces an internal (for now) `useOAuthConsent()` hook for fetching OAuth consent screen metadata for the signed-in user.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/shared/src/react/hooks/useOAuthConsent.types.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * @interface
+ */
+export type UseOAuthConsentReturn = {
+  /**
+   * OAuth consent screen metadata from Clerk, or `undefined` before the first successful fetch.
+   * Additional fields (e.g. submission helpers) may be added in the future without renaming this hook.
+   */
+  data: OAuthConsentInfo | undefined;
```
jfoshee, line 38: Might I regret typing this as `OAuthConsentInfo`? If I need to add a member for e.g. post consent response endpoint? or a method?

packages/shared/src/react/stable-keys.ts
```diff
@@ -15,6 +15,9 @@ const API_KEYS_KEY = 'apiKeys';
 // Keys for `useOrganizationCreationDefaults`
 const ORGANIZATION_CREATION_DEFAULTS_KEY = 'organizationCreationDefaults';
 
+// Keys for `useOAuthConsent` (consent metadata GET; keep distinct from future submit/mutation keys)
+const OAUTH_CONSENT_INFO_KEY = 'oauthConsentInfo';
```
jfoshee, line 19: what are these keys for? Do I need to think about future members?

packages/shared/src/react/hooks/useOAuthConsent.types.ts
```diff
@@ -0,0 +1,57 @@
...
+   */
+  error: ClerkAPIResponseError | null;
+  /**
+   * Whether the initial consent metadata fetch is still in progress.
+   */
+  isLoading: boolean;
+  /**
+   * Whether any request is still in flight, including background updates.
+   */
+  isFetching: boolean;
```
jfoshee, line 50: Do we need both isFetching and isLoading?

packages/shared/src/react/hooks/useOAuthConsent.types.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @interface
+ */
+export type UseOAuthConsentParams = Partial<Pick<GetOAuthConsentInfoParams, 'oauthClientId' | 'scope'>> & {
+  /**
+   * If `true`, the previous data will be kept in the cache until new data is fetched.
+   *
+   * @default true
+   */
+  keepPreviousData?: boolean;
```
jfoshee, line 20: do we need these?

packages/shared/src/react/hooks/useOAuthConsent.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const { oauthClientId: oauthClientIdParam, scope: scopeParam, keepPreviousData = true, enabled = true } = params;
+  const clerk = useClerkInstanceContext();
+  const user = useUserBase();
+
+  const [searchSnapshot, setSearchSnapshot] = useState(() =>
+    typeof window !== 'undefined' ? window.location.search : '',
+  );
+
+  useLayoutEffect(() => {
```
jfoshee: why not use `useEffect` here? 🤔

packages/shared/src/react/hooks/useOAuthConsent.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const getConsentInfo = clerk.oauthApplication?.getConsentInfo;
+  const hasClientId = oauthClientId.length > 0;
+  const queryEnabled = Boolean(user) && hasClientId && enabled && clerk.loaded && typeof getConsentInfo === 'function';
+
+  const queryFn = useCallback(() => {
+    if (!getConsentInfo) {
+      return Promise.reject(new Error('OAuth consent is not available in this Clerk instance'));
+    }
+    const p: GetOAuthConsentInfoParams = scope !== undefined ? { oauthClientId, scope } : { oauthClientId };
```
jfoshee: note that scope is optional

packages/shared/src/react/hooks/useOAuthConsent.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const query = useClerkQuery({
+    queryKey,
+    queryFn,
+    enabled: queryEnabled,
+    placeholderData: defineKeepPreviousDataFn(keepPreviousData),
+  });
+
+  return {
+    data: query.data,
```
jfoshee, line 86: let's change `data` to `consentInfo`

wobsoriano:
!snapshot

packages/shared/src/react/hooks/useOAuthConsent.tsx
wobsoriano: After doing another review, I'd say we should keep `data` here. Every other useClerkQuery-based hook returns `data`, not a named field.

The hook name already tells us what data is here, which is the consent info, and consumers can always destructure as `const { data: consentInfo } = useOAuthConsent() if they want a named variable`.

packages/shared/src/react/stable-keys.ts
wobsoriano: These are [tanstack query cache keys](https://tanstack.com/query/v5/docs/framework/react/guides/query-keys). Every query-based hook needs a unique stable key for caching/invalidation.

Tbh we dont really need caching/invalidation here as the page gets visited only once by user, but since we are using our tanstack query wrapper internally, we need a cache key

packages/shared/src/react/hooks/useOAuthConsent.tsx
wobsoriano: Agreed, `useEffect` would be fine here. But I think we can simplify further: the URL doesn't change after load (consent pages are full-navigation redirects), so we can drop the state + effect entirely and just read it once in a `useMemo`.

Also coderabbit flagged this for Expo users, I'll clean this up

packages/shared/src/react/hooks/useOAuthConsent.types.ts
wobsoriano: We need the `isLoading` here to check if the consent info is being fetched initially or not.

The `isFetching` is useless here as the page gets visited only once (this is for backgroudn fetches) but to follow existing hooks, we keep it for consistency

packages/shared/src/react/hooks/useOAuthConsent.types.ts
wobsoriano: Good instinct. Let's keep `data` as the raw FAPI response and put the submission helpers (even the `submissionEndpoint` property we talked about) at root level alongside `data`, similar to how  other hooks return additional properties.

That way `data` stays a clean mirror of the API response, and hook utilities live at the top level

packages/shared/src/react/hooks/useOAuthConsent.types.ts
wobsoriano: To be consistent with other hooks, yes

wobsoriano (approved):
Looks good overall! Just pushed a few cleanup commits.

Let's also export it from react and nextjs `/internal` path until public release

PR #8303, opened 2026-04-13
fix(clerk-js): prevent background token refresh from nuking sessions on mobile

chriscanin:
## Summary

Fixes an issue where iOS background thread throttling causes unexpected session destruction in Expo/React Native apps.

**The bug:** On iOS, background audio apps (e.g., sleep/dream trackers) have their JS event loop starved for hours. When the SDK's background token refresh timer eventually fires, it sends a request with stale credentials → 401 → `handleUnauthenticated()` cascades into destroying the session, even though it's still valid on the server (30-day lifetime). Users wake up to find they're signed out.

**The fix:** A single early return in `#refreshTokenInBackground()`, gated to headless/mobile runtimes only (Expo sets `runtimeEnvironment: 'headless'`). If the token has already expired when the refresh timer fires, the refresh cycle was starved by iOS background throttling — bail out instead of sending a request with stale credentials. The next foreground `getToken()` call handles token acquisition through the normal path with proper retry logic.

### Why this is safe

- **Mobile-only:** Gated to `runtimeEnvironment === 'headless'`, which is only set by `@clerk/expo` for React Native. Web, Next.js, Remix, Chrome extension, etc. are completely unaffected.
- **Only fires when the refresh is already too late:** During normal operation, the refresh timer fires at ~43s into a 60s token (before expiration). The early return only triggers when the token is ALREADY expired, meaning the timer fired late due to OS-level thread throttling.
- **No shared code paths changed:** No changes to `_baseFetch`, `Token.create`, `handleUnauthenticated`, or any other shared infrastructure. The change is entirely within the private `#refreshTokenInBackground()` method in `Session.ts`.
- **Legitimate revocations still work:** If a session is revoked, the next foreground `getToken()` call triggers `handleUnauthenticated()` through the normal path and signs the user out properly.
- **Background refresh is best-effort:** It already catches and swallows all errors. Skipping a doomed refresh is strictly better than attempting one that nukes the session as a side effect.

### Files changed

| File | Change |
|------|--------|
| `Session.ts` | Early return in `#refreshTokenInBackground()` when token is expired on headless runtime |
| `Session.test.ts` | Unit test verifying background refresh is skipped when token is expired on headless runtime |

## How was this tested

1. **Unit test**: Simulates iOS throttling by jumping the clock past token expiration, then firing the refresh timer. Verifies no API call is made on headless runtime. All 73 Session tests pass.
2. **Live iOS simulator test**: Debug panel in Expo quickstart app intercepts `/tokens` with 401, waits for background refresh timer (~43s), confirms session survives.
3. **Maestro automated test**: End-to-end flow on iOS simulator — sign in → simulate background refresh 401 → verify session survives → verify `getToken()` still works after.

## Test plan

- [x] Unit test: background refresh skipped when token expired on headless runtime
- [x] All existing Session tests pass (73/73)
- [x] Live verification on iOS simulator
- [x] Maestro end-to-end test
- [ ] Verify web apps are unaffected (runtimeEnvironment not set → early return never triggers)


packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -574,6 +574,21 @@ export class Session extends BaseResource implements SessionResource {
...
 
+    // Mobile only: skip this refresh if the token is already expired.
+    // On iOS, the OS throttles background JS threads for hours (e.g. overnight audio apps).
+    // The refresh timer fires late — well past token expiry — with stale credentials.
+    // If we send that request, the 401 response triggers handleUnauthenticated(), which
+    // destroys the session even though it's still valid on the server (30-day lifetime).
+    // Instead, bail out here and let the next foreground getToken() call recover normally.
+    const isHeadless =
+      (Session.clerk?.__internal_getOption?.('experimental') as { runtimeEnvironment?: string } | undefined)
+        ?.runtimeEnvironment === 'headless';
```
jacekradko: Just small typing fix, as `runtimeEnvironment` is typed a literal type (not a string) : `runtimeEnvironment: 'headless';` so casting it to string actually weakens type safety

```suggestion
    const experimental = Session.clerk?.__internal_getOption?.('experimental');
    const isHeadless = experimental?.runtimeEnvironment === 'headless';
```

jacekradko (approved):
LGTM

PR #8309, opened 2026-04-15
chore(clerk-js,ui): Add infinite loading to organization selection in OAuthConsent

wobsoriano:
## Description

Adds infinite loading to the organization selector inside OAuthConsent component. 

https://github.com/user-attachments/assets/b8ddae51-d309-44ec-a5ca-0c1b667331f9

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/elements/Select.tsx
```diff
@@ -376,6 +377,7 @@ export const SelectOptionList = (props: SelectOptionListProps) => {
             );
           })}
           {noResultsMessage && options.length === 0 && <SelectNoResults>{noResultsMessage}</SelectNoResults>}
+          {footer}
```
wobsoriano, line 385: There's probably a better way there but simplest is just adding an element here to keep pagination concerns out of the shared primitive. The `OrgSelect` component owns the scroll detection logic and passes the spinner in as a footer when there are more pages to load

packages/ui/src/elements/Select.tsx
```diff
@@ -294,6 +296,10 @@ export const SelectOptionList = (props: SelectOptionListProps) => {
     if (e.key === 'ArrowDown') {
       e.preventDefault();
       if (isOpen) {
+        if (onReachEnd && focusedIndex === options.length - 1) {
+          onReachEnd();
+          return;
+        }
```
wobsoriano, line 302: should not break existing components interactions like the phone number input. Only invokes the method if it exists

alexcarpenter (approved):

PR #8319, opened 2026-04-15
refactor(ui): Implement new icon set

alexcarpenter:
## Description

Migrate to new icon set https://www.figma.com/design/bTar3y0gIglvIxhmZsfHsj/Clerk-Mosaic?node-id=71-567&p=f&m=dev

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/icons/index.ts
```diff
@@ -7,69 +7,69 @@
 export { default as Add } from './add.svg';
 export { default as ApplePay } from './apple-pay.svg';
 export { default as ArrowLeftIcon } from './arrow-left.svg';
-export { default as ArrowRightButtonIcon } from './arrow-right-button.svg';
+export { default as ArrowRightButtonIcon } from './triangle-right.svg';
```
dstaley: could we add exports for the new names so that it's easier to correlate them between the SDK and Figma? For example, I assume Figma will list the icon as `triangle-right`, so we'd be expecting to use the `TriangleRight` icon. I think keeping the older name is fine for now (we'll want to clean them up later though).

alexcarpenter: yeah thats a good callout 👍🏼 will get that update.

packages/ui/src/icons/index.ts
```diff
@@ -6,72 +6,62 @@
...
 export { default as Code } from './code.svg';
-export { default as CogFilled } from './cog-filled.svg';
+export { default as Cog } from './cog.svg';
 export { default as Copy } from './copy.svg';
 export { default as CreditCard } from './credit-card.svg';
 export { default as DeviceLaptop } from './device-laptop.svg';
 export { default as DeviceMobile } from './device-mobile.svg';
-export { default as DotCircle } from './dot-circle-horizontal.svg';
-export { default as Download } from './download.svg';
+export { default as DotCircleHorizontal } from './dot-circle-horizontal.svg';
```
wobsoriano: Per the Figma mapping, should this become `three-dots-circle`?

https://www.figma.com/design/bTar3y0gIglvIxhmZsfHsj/Clerk-Mosaic?node-id=5255-1822&m=dev

wobsoriano (approved):
Looks good! left a q

packages/ui/src/icons/index.ts
alexcarpenter: thanks for catching!

wobsoriano (approved):
Okay, LGTM round 2 👍🏼 

verified the figma icon mapping against the PR diff and they all line up

PR #7450, opened 2025-12-13
feat(clerk-js,react,localizations,shared,ui): Add Solana feature support to core 3

kduprey:
## Description

This PR ports over the new Sign in with Solana feature changes currently in Core 2 over to Core 3, allowing for continued support of the new feature.

[USER-4212](https://linear.app/clerk/issue/USER-4212/add-solana-feature-support-to-core-3-release-branch)

## Checklist

- [ ] `pnpm test` runs as expected.
- [ ] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Sign in / Sign up with Solana wallets: wallet chooser flows, Solana-aware auth paths, wallet-name support, new Solana wallet buttons and selection UI.

* **Documentation**
  * Added Solana/Web3 UI copy and two Web3 error messages across 40+ locales.

* **Chores**
  * Added Solana wallet adapters and packaging updates; updated bundle size thresholds and build config for new Web3 bundle.

* **Bug Fixes**
  * Safer Web3 error handling and resilient wallet-resolution paths.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #8329, opened 2026-04-16
feat(clerk-js,shared,ui): Add Protect SDK challenge support during sign-up and sign-in

zourzouvillys:
## Summary

Adds client-side support for Clerk Protect mid-flow SDK challenges (`protect_check`) during both sign-up and sign-in. When the antifraud service gates a step, the SDK exposes the challenge, surfaces a card that loads and runs the challenge script, submits the resulting proof token, and resumes the original flow.

- New `protectCheck` field and `submitProtectCheck()` method on both `SignUp` and `SignIn` resources (and their future variants), mirrored on the `@clerk/react` state proxies.
- New `'needs_protect_check'` value on the `SignInStatus` union.
- New `protect-check` route on the prebuilt `<SignIn />` and `<SignUp />` components (standalone, `continue`, and combined-flow `create` / `create/continue` depths).

## Background

Previously anti-fraud blocks could only happen at sign-in/sign-up *create* time. This mechanism lets the service gate at any step. When gated, the response carries:

```json
{
  "protect_check": {
    "status": "pending",
    "token": "<challenge token>",
    "sdk_url": "https://.../sdk.js",
    "expires_at": 1700000000000,
    "ui_hints": { "reason": "device_new" }
  }
}
```

`expires_at` is a **Unix epoch timestamp in milliseconds** (documented on the type). The client loads the SDK at `sdk_url`, runs the challenge with `token`, and submits the proof token to `PATCH /v1/client/sign_{ins,ups}/{id}/protect_check`. The response clears the gate, issues a chained challenge, or completes the flow.

### Types (`@clerk/shared`)
- `ProtectCheckJSON` / `ProtectCheckResource` `{ status: 'pending', token, sdkUrl, expiresAt?, uiHints? }`; `expires_at` is optional on both `SignUpJSON` and `SignInJSON` (older FAPI versions omit it).
- `'protect_check'` added to `SignUpField`; `'needs_protect_check'` added to `SignInStatus`.
- `submitProtectCheck` added to the sign-up/sign-in resource + future interfaces.

### Core resources (`@clerk/clerk-js`)
- `SignUp` / `SignIn` expose `protectCheck` and `submitProtectCheck({ proofToken })`; `fromJSON` / `__internal_toSnapshot` round-trip the field; future variants mirror the API.

### SDK loader helper (`@clerk/shared/internal/clerk-js/protectCheck`)
`executeProtectCheck(protectCheck, container, { signal })` — validates `sdkUrl` (must be `https:`, no credentials, rejects `data:`/`blob:`/`javascript:`), runs the spec-compliant script contract `(container, { token, uiHints, signal })`, forwards the `AbortSignal`, and wraps failures in typed error codes without leaking the URL.

### Shared card runner (`@clerk/ui`)
Both protect-check cards share one `useProtectCheckRunner` hook so the lifecycle can't drift:
- Keys the effect on `protectCheck.token` (not object identity) so an unrelated resource refresh doesn't restart the challenge.
- Caps expired-challenge reloads and fails loud instead of spinning (a plain GET doesn't re-mint).
- Wraps the script run in a timeout, and the error state offers a **retry** control.
- Fails closed in no-RHC builds (`__BUILD_DISABLE_RHC__`) before the remote `import(sdk_url)` — the guard lives in the component layer because `@clerk/shared` is compiled once with the flag `false`.
- Finalizes (`setActive`) the `complete` case from both the normal success and the `protect_check_already_resolved` reload, so neither strands the user.
- Loading state uses a `descriptors.spinner` spinner in an `aria-live` region.

### Sign-in gate routing — single choke point
`navigateOnSignInProtectGate(res, navigate, protectCheckPath)` is the one place that turns a gated sign-in response into navigation. Every dispatch site routes through it (start ×2, passkey, password, code, alt-channel, backup-code, factor-two code, reset-password), with the protect-check path passed **per caller** (index route → `'protect-check'`, factor cards → `'../protect-check'`). Also wired into the previously-missed **email-link** result handler and the inline **web3/Solana** path (`clerk.authenticateWithWeb3`, which doesn't redirect through `_handleRedirectCallback`): it takes `protectCheckUrl` / `signUpProtectCheckUrl` params and routes a gated attempt to the sign-in **or** sign-up challenge depending on which resource the attempt resolved through (the `identifier_not_found → signUp` fallback is covered).

### OAuth / SAML callback (`clerk.ts`)
`_handleRedirectCallback` checks the gate before its transfer/missing-fields logic, **scoped to the callback intent** (`reloadResource`) so an abandoned sign-in's stale `protect_check` can't hijack a sign-up callback (and vice versa). The sign-up gate check runs before the `missing_fields` short-circuit so a gated `signUp.create({ transfer })` routes to the challenge instead of `/continue`.

### Prebuilt UI routes (`@clerk/ui`)
`protect-check` routes registered on `<SignIn />`/`<SignUp />` at every depth the flow can mount sign-up at; `SignUpProtectCheck` takes per-mount continuation paths (the `continue`-nested mounts pass `continuePath='..'`).

### Localization (`@clerk/localizations`, `@clerk/shared`)
Typed `signUp.protectCheck.{title,subtitle,loading,retryButton}` / `signIn.protectCheck.*` keys and `unstable__errors` entries for the runtime error codes (`protect_check_execution_failed`, `…_invalid_script`, `…_invalid_sdk_url`, `…_script_load_failed`, `…_timed_out`, `…_unsupported_environment`; `…_aborted` / `…_already_resolved` intentionally undefined).

## Backwards compatibility

- All new JSON fields are optional; old SDK consumers ignore them.
- `'needs_protect_check'` is type-additive — runtime behavior is unchanged (the server emits it only behind a feature gate, and `protectCheck` is the authoritative field). Strict-TypeScript consumers with an exhaustive `switch (signIn.status)` will get a new unhandled-branch hint, hence the `minor` bump.
- No existing API surface is removed.

## Risks

- **Custom flows** that switch on `signIn.status` need to handle `'needs_protect_check'` (or the `protectCheck` field). Documented on the resource interface.
- **Challenge SDK contract** — the loaded script must default-export `(container, { token, uiHints, signal }) => Promise<string>`. Coordinate with the Protect SDK team before deploying.
- **CSP** — apps with strict CSP must allow the Protect script origin via `script-src`; the load-failure error calls this out.

## Test plan

- [x] Unit (resources): `SignUp.test.ts` / `SignIn.test.ts` — serialization, optional fields, snapshot round-trip, `submitProtectCheck` path/method/body
- [x] Unit (helper): `protectCheck.test.ts` — URL validation, script contract, cancellation, error wrapping
- [x] Unit (flow): `completeSignUpFlow.test.ts` — routing priority
- [x] Unit (redirect): `clerk.test.ts` — gate routing scoped to the callback intent (stale sign-in not picked up by a sign-up callback; sign-in callback routes to the gate)
- [x] Unit (choke point): `handleProtectCheck.test.ts` — `navigateOnSignInProtectGate` / `isSignInProtectGated` (both gate signals, per-caller path, no navigation when ungated)
- [x] Integration (call site): `SignInFactorOne.test.tsx` — a gated first-factor attempt routes to `../protect-check` instead of dispatching on the underlying status
- [x] Component: `SignUpProtectCheck.test.tsx` / `SignInProtectCheck.test.tsx` — run/expiry/already-resolved/chained/abort/no-submit-on-failure, finalize-on-reload-complete, retry control
- [x] Build + type-check: `@clerk/clerk-js`, `@clerk/shared`, `@clerk/localizations`, `@clerk/ui` clean; lint clean
- [ ] Manual: drive a sign-up/sign-in on a Protect-enabled instance (challenge renders + resolves, chained challenge, expired auto-recovery, OAuth/SAML callback)

## Follow-ups (out of scope)

- Server-side ownership of re-minting an **expired** challenge on read (vs. re-running the gated step) — capped client-side so it can't loop in the meantime.
- Additional test coverage (lower priority): a dedicated `authenticateWithWeb3` sign-up-gate regression test, an email-link gate-routing test, and the hook's no-RHC / timeout branches (not exercisable in the current ui vitest setup).
- `@clerk/backend` resource model updates (the backend SDK doesn't drive end-user flows).
- Non-blocking `protect_check` (additive when the server starts emitting it).

## Summary by CodeRabbit

* **New Features**
  * Clerk Protect mid-flow challenge support for sign-up and sign-in with automatic routing in pre-built flows (including Web3 and passkey), navigation guards, and routing for chained challenges
  * Added protectCheck state, submitProtectCheck APIs, and new needs_protect_check sign-in status
  * New ProtectCheck UI components, routing steps, and a shared hook to run/retry/cancel/resume challenges

* **Localization**
  * Added protect-check UI strings and new protect-check error messages

* **Tests**
  * Extensive tests covering flows, SDK execution, cancellation, chaining, routing, and edge cases


packages/ui/src/components/SignIn/shared.ts
```diff
@@ -28,6 +31,12 @@ function useHandleAuthenticateWithPasskey(onSecondFactor: () => Promise<unknown>
   return useCallback(async (...args: Parameters<typeof authenticateWithPasskey>) => {
     try {
       const res = await authenticateWithPasskey(...args);
+      // Per spec §2.3 / §4: protect_check can fire on attempt_first_factor (which is what
```
jacekradko: [minor] is this a reference to an LLM spec document?

packages/ui/src/components/SignIn/index.tsx
jacekradko, line 104: I believe this route segment also needs a `protect-check` path. Looks like we have navigations to `create/protect-check` in the combined flow

```suggestion
 <Route path='create'>    
    <Route
      path='protect-check'
      canActivate={clerk => !!clerk.client.signUp.protectCheck}
    >                                                                                                                                                                             
      <LazySignUpProtectCheck />
    </Route>
  <Route  
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2341,11 +2347,20 @@ export class Clerk implements ClerkInterface {
       });
     }
 
+    // Per Protect spec §4.4: OAuth/SAML callbacks can result in a protect_check gate that
```
jacekradko: What spec is this referring to?

jacekradko:
@zourzouvillys The core stuff looks good. I think the biggest gap is the routing logic integration. Feels like it this is targeting the standalone `<SignIn /> / <SignUp />` , but the combined flows are not hooked up properly.

packages/ui/src/components/SignIn/shared.ts
zourzouvillys: Good catch — that `§` reference pointed at the internal Protect FAPI design doc by section number, which isn't useful to reviewers and reads like an artifact. I've reworded this and all the other `Per spec §X.X` comments across the PR to describe the behavior directly instead. Fixed in ead705968.

packages/ui/src/components/SignIn/index.tsx
zourzouvillys: You're right, this was a real gap — `handleCombinedFlowTransfer.ts:103` navigates to `create/protect-check` but no route was registered there, so a Protect gate during combined-flow sign-up would dead-end. Added the nested route in ead705968:

```tsx
<Route path='create'>
  <Route
    path='protect-check'
    canActivate={clerk => !!clerk.client.signUp.protectCheck}
  >
    <LazySignUpProtectCheck />
  </Route>
  ...
```

This also needed `LazySignUpProtectCheck` added to `lazy-sign-up.ts` and `SignUpProtectCheck` re-exported from the `SignUp` barrel (it was only used internally before). Gated on `signUp.protectCheck` since the combined flow's `create/*` segment drives the sign-up resource.

packages/clerk-js/src/core/clerk.ts
zourzouvillys: Same as above — internal Protect design doc section. Reworded to explain the behavior directly (OAuth/SAML callbacks can resolve into a protect_check gate that surfaces on the next `/v1/client` read, so we check before the transfer logic). Fixed in ead705968.

zourzouvillys:
Thanks — you were right about the combined flow. The standalone `<SignIn />` / `<SignUp />` trees had their `protect-check` routes, but the combined flow embeds sign-up under `create/*` and that nested `protect-check` route was missing, so the `create/protect-check` navigation dead-ended. Fixed in ead705968 (nested route + lazy export + barrel re-export). Also reworded the unverifiable `§` spec citations you flagged. Rebased on latest `main` and resolved the conflicts. Ready for another look.

Ephem:
@zourzouvillys This part makes this a breaking change right?

> Custom flows (non-prebuilt UI) that switch on signIn.status need to handle 'needs_protect_check' (or the protectCheck field) themselves. Without handling, the UI will appear stuck at the previous step. The new fields are documented on the SignInResource interface.

Edit: Is still a draft and if so, what's outstanding?

packages/ui/src/components/SignIn/shared.ts
```diff
@@ -28,6 +31,12 @@ function useHandleAuthenticateWithPasskey(onSecondFactor: () => Promise<unknown>
   return useCallback(async (...args: Parameters<typeof authenticateWithPasskey>) => {
     try {
       const res = await authenticateWithPasskey(...args);
+      // A protect_check gate can fire on attempt_first_factor, which is what
+      // authenticateWithPasskey calls under the hood. Detect both the explicit field and the
+      // SDK-version-gated status before dispatching on the underlying status.
+      if (isSignInProtectGated(res)) {
+        return navigate('../protect-check');
```
jacekradko: Could we pass the protect-check path in per caller, the same way `onSecondFactor` does? `SignInStart` → `'protect-check'`, `SignInFactorOnePasskey` → `'../protect-check'`.

Right now it's hard-coded to `'../protect-check'`, which works from the factor-one mount but not from `SignInStart` (index route): there `..` lands at `/protect-check` (app root, 404) instead of `/sign-in/protect-check`. Autofill makes it fire without a click, so a gated start-page passkey sign-in dead-ends.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2415,6 +2420,7 @@ export class Clerk implements ClerkInterface {
         verifyPhonePath:
           params.verifyPhoneNumberUrl ||
           buildURL({ base: displayConfig.signUpUrl, hashPath: '/verify-phone-number' }, { stringify: true }),
+        protectCheckPath: buildURL({ base: displayConfig.signUpUrl, hashPath: '/protect-check' }, { stringify: true }),
```
jacekradko: I don't think this `protectCheckPath` is reachable on a gated sign-up. A protect-gated response always carries `'protect_check'` in missing_fields, so the `if (missingFields.length)` check above short-circuits to `navigateToContinueSignUp()` and OAuth/SAML callbacks land on `/continue` instead of the challenge. Could we check the gate before that short-circuit and route to protect-check first (same for the `signUp.create({ transfer: true })` path)?

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
```diff
@@ -0,0 +1,223 @@
...
+          },
+        });
+        return true;
+      }
+      return false;
+    };
+
+    const runChallenge = async () => {
+      try {
+        const proofToken = await executeProtectCheck(protectCheck, container, {
```
jacekradko: Suggest adding a `__BUILD_DISABLE_RHC__` guard here (and in `SignUpProtectCheck`) before calling `executeProtectCheck`, bailing with a clear unsupported-environment error like turnstile/one-tap do. It has to live in the component, not in `executeProtectCheck`: that's in `@clerk/shared`, which is built once with the flag `false`, so a guard there never trips.

Without it, the no-RHC builds still run the remote `import(sdk_url)`. `@clerk/ui/no-rhc` (chrome-extension) and `clerk.no-rhc.js` would ship a live remote import (MV3 / CSP rejection, and a fail-closed gate with no way forward), and `search-for-rhc` can't catch it since the URL is a runtime value.

packages/ui/src/components/SignUp/SignUpVerificationCodeForm.tsx
```diff
@@ -46,6 +46,7 @@ export const SignUpVerificationCodeForm = (props: SignInFactorOneCodeFormProps)
           signUp: res,
           verifyEmailPath: '../verify-email-address',
           verifyPhonePath: '../verify-phone-number',
+          protectCheckPath: '../protect-check',
```
jacekradko, line 50: I think this breaks when the card is shown inside the `continue` flow. You get there from SignUpContinue, and the path then is `/sign-up/continue/verify-email-address`. From there `'../protect-check'` points at `continue/protect-check`, which isn't a real route, so if the gate fires on the verify step the user just sees a blank screen. The verify links work at that depth because `continue` has its own copies of the verify routes. Could we do the same for protect-check, add it under `continue` in `SignUp/index.tsx`? The combined flow needs it too: `create/` already has a `protect-check` route but `create/continue` doesn't.

packages/ui/src/components/SignIn/handleProtectCheck.ts
```diff
@@ -0,0 +1,12 @@
+import type { SignInResource } from '@clerk/shared/types';
+
+/**
+ * Detects whether a sign-in response is gated by Clerk Protect.
+ *
+ * The `protectCheck` field is the authoritative gating signal; new SDKs / newer servers
+ * also surface `status === 'needs_protect_check'`. Either signal triggers navigation
+ * to the protect-check route.
+ */
+export function isSignInProtectGated(signIn: SignInResource): boolean {
```
jacekradko, line 10: Since the gate is checked per call site, it's easy for one to slip through, and a missed one strands the user (the email-link and prepare/resend paths and web3 in `clerk.authenticateWithWeb3` look like they don't run it). Could we double-check every sign-in response path wires this in? Longer term a single choke point that checks the gate before dispatching would make it so new call sites can't forget.

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
```diff
@@ -0,0 +1,223 @@
...
+    // dependency changes (chained challenge with a new protectCheck reference), abort
+    // any in-flight script execution and skip downstream state updates.
+    const abortController = new AbortController();
+    let cancelled = false;
+
+    // Do not attempt to solve an expired challenge.
+    // Reload the resource so the server can mint a fresh challenge before re-routing,
+    // otherwise the local stale `signIn.protectCheck` would re-trigger this same effect
+    // and loop indefinitely with no user feedback.
+    if (protectCheck.expiresAt !== undefined && protectCheck.expiresAt < Date.now()) {
```
jacekradko: I checked the FAPI side and I think this loops forever once a challenge expires. A plain GET on the sign-in doesn't mint a new challenge (only create/attempt/prepare do that), and the serializer returns the protect_check as long as it's `pending`, expired or not. So the reload here gets the same expired challenge back, the effect runs again, sees it's expired, reloads again, and so on. The user just sees the spinner while we hit FAPI in a loop.

Getting a new challenge means either re-running the gated step or having FAPI re-mint on read, probably worth deciding with the clerk_go folks which side owns that. Either way I'd put a cap on the reloads so it can't spin. (Units are fine btw, FAPI sends `expires_at` in ms.)

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
```diff
@@ -0,0 +1,223 @@
...
+    void runChallenge();
+
+    return () => {
+      cancelled = true;
+      abortController.abort();
+      // Reset the guard so the next mount / dep-change can re-run; this is what makes
+      // chained challenges work correctly across re-renders.
+      isRunningRef.current = false;
+    };
+  }, [signIn.protectCheck]);
```
jacekradko: `fromJSON` makes a new `protectCheck` object on every resource update, so any unrelated refresh re-runs this effect and restarts the challenge under the user. Keying the dep on `protectCheck?.token` would fix that and still re-run for a real chained challenge. (Same in `SignUpProtectCheck`.)

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+        if (cancelled) {
+          return;
+        }
+        await navigateNext(updatedSignIn, navigate);
+      } catch (err: any) {
+        if (cancelled) {
+          return;
+        }
+        handleError(err, [], card.setError);
```
jacekradko: I like that failures surface a proper card error here. A small gap: the message says "please try again" but there's no retry control and nothing re-runs the effect, so recovering takes a full page reload. A retry action on the error state would round this out, and a timeout around the script run would cover the case where the challenge SDK never settles.

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
```diff
@@ -0,0 +1,223 @@
...
+          }
+          // protect_check_already_resolved is retry-safe.
+          // Reload to clear the stale local protectCheck; otherwise navigateNext would
+          // see protectCheck still set and self-loop on this route.
+          if (isClerkAPIResponseError(err) && err.errors?.[0]?.code === ERROR_CODES.PROTECT_CHECK_ALREADY_RESOLVED) {
+            await signIn.reload();
+            if (cancelled) {
+              return;
+            }
+            await navigateNext(signIn, navigate);
```
jacekradko: Can you double-check what happens here when the reload comes back `complete`? It looks like this branch skips `finalizeIfComplete`, and `navigateNext`'s complete case just goes to `'..'` expecting the caller to have run setActive, so the user would land back on the start form without the session being activated. If that's right, mirroring the success path below (finalize, then navigate) should cover it. The sign-up side already does this via `completeSignUpFlow`'s `handleComplete`.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2451,11 +2457,21 @@ export class Clerk implements ClerkInterface {
       });
     }
 
+    // OAuth/SAML callbacks can resolve into a protect_check gate that surfaces on the next
+    // /v1/client read, so check for it here before continuing with the transfer logic below.
+    // Honor either the explicit `protectCheck` field or the `needs_protect_check` status override.
+    if (si.protectCheck || si.status === 'needs_protect_check') {
```
jacekradko: I traced this through clerk_go and I think it can grab the wrong flow. This check runs before all the sign-up branches and reads `si` straight from `this.client.signIn`, with no scoping to which flow the callback is for. An abandoned sign-in sticks around on the client for a day (`AbandonAt` is now + 1 day) and keeps serializing its pending `protect_check`, and in multi-session mode a later sign-up doesn't clear it. So if someone starts a sign-in, hits the gate, bails, then signs up with Google, the callback sees the stale `protectCheck` and routes them into the old sign-in's protect-check instead of finishing the Google sign-up.

Could we scope this to the sign-in callback (e.g. gate on the same `reloadResource`/intent the rest of the function uses) so a sign-up callback won't pick it up?

packages/shared/src/types/json.ts
```diff
@@ -143,6 +143,18 @@ export interface SignUpJSON extends ClerkResourceJSON {
   legal_accepted_at: number | null;
   locale: string | null;
   verifications: SignUpVerificationsJSON | null;
+  protect_check: ProtectCheckJSON | null;
```
jacekradko: [nit] This is required here but optional on `SignInJSON` (`protect_check?`). Older FAPI versions don't send the field, so optional matches the wire format better and keeps the two consistent.

.changeset/protect-check-support.md
```diff
@@ -0,0 +1,21 @@
+---
+'@clerk/clerk-js': minor
```
jacekradko, line 2: `@clerk/react` is missing here. `stateProxy.ts` adds `protectCheck`/`submitProtectCheck` to the public state proxies, so that's user-facing surface in `@clerk/react` that needs a `'@clerk/react': minor` entry, otherwise it ships with no changelog and only an implicit dependency bump.

packages/clerk-js/src/core/resources/__tests__/SignIn.test.ts
```diff
@@ -2785,4 +2785,171 @@ describe('SignIn', () => {
       });
     });
   });
+
+  describe('protectCheck', () => {
```
jacekradko, line 2832: [nit] This block sets `BaseResource._fetch = mockFetch` but, unlike the sibling describes here, has no `afterEach` to restore it. It's harmless today since it's the last describe, but it'd leak the mock if another block is added below. Also, the `submitProtectCheck` assertion checks `method`/`body` but not the path, worth asserting it hits `.../{id}/protect_check` so the test would catch a wrong/dropped action. (Same in `SignUp.test.ts`.)

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
```diff
@@ -0,0 +1,223 @@
...
+            elementDescriptor={descriptors.main}
+            gap={6}
+          >
+            <Box
+              ref={containerRef}
+              id='clerk-protect-check'
+              style={{ display: 'block', alignSelf: 'center', minHeight: '60px' }}
+            />
+            {isRunning && !card.error ? (
+              <Box style={{ alignSelf: 'center' }}>{t(localizationKeys('signIn.protectCheck.loading'))}</Box>
```
jacekradko: [nit] This loading indicator is bare text with an inline style, no `elementDescriptor` and no `aria-live`/`aria-busy`, so screen readers won't announce that verification is running and themes can't target it. The other waiting cards use `LoadingCardContainer` with `descriptors.spinner`, might be nice to reuse that here.

packages/ui/src/components/SignIn/shared.ts
zourzouvillys: Fixed. Threaded `protectCheckPath` through `useHandleAuthenticateWithPasskey` (defaults to `'../protect-check'` for the factor-one mount); `SignInStart` now passes `'protect-check'` for both its usages. The gate detection + navigation is also now funneled through a single `navigateOnSignInProtectGate(res, navigate, path)` helper, so the start-page autofill passkey no longer dead-ends at the app root.

packages/clerk-js/src/core/clerk.ts
zourzouvillys: Good catch. `navigateToNextStepSignUp` now checks `signUp.protectCheck || missingFields.includes('protect_check')` **before** the `if (missingFields.length)` short-circuit, so a gated OAuth/SAML callback routes to `/sign-up/protect-check` instead of `/continue`. This covers the `signUp.create({ transfer: true })` path (which calls it). I also added an early `su.protectCheck` check next to the existing `si` one for a directly-gated sign-up — and scoped both to the callback intent (see the `clerk.ts:2463` thread).

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
zourzouvillys: Added. The guard lives in the component layer (in the shared `useProtectCheckRunner` hook both cards use) — it fails closed with a `protect_check_unsupported_environment` error before any `import(sdk_url)`, exactly because `@clerk/shared`'s `executeProtectCheck` is compiled once with the flag `false`. `check:no-rhc` still passes since the URL is a runtime value.

packages/ui/src/components/SignUp/SignUpVerificationCodeForm.tsx
zourzouvillys: Fixed. Registered `protect-check` routes under `continue` (in `SignUp/index.tsx`) and `create/continue` (combined flow, `SignIn/index.tsx`). Since the same card mounts at different depths, I parameterized `SignUpProtectCheck`'s continuation paths — the nested mounts pass `continuePath='..'` (the `continue` index) instead of the default `'../continue'`; the verify/self paths already resolve correctly from every depth, so the `'../protect-check'` you flagged now lands on a real route.

packages/ui/src/components/SignIn/handleProtectCheck.ts
zourzouvillys: Did both. Added `navigateOnSignInProtectGate` as the single place that turns a gated response into navigation, and routed every existing call site through it (dropping the now-dead `needs_protect_check` switch branches). For the missed paths you flagged: wired the email-link result handler and the inline web3/Solana path (`clerk.authenticateWithWeb3`, which doesn't redirect so `_handleRedirectCallback` never catches it — it now takes a `protectCheckUrl`). One honest gap remains: a sign-up via web3 wallet that gets gated — I guarded it from falling into `/continue` but didn't route it to a sign-up challenge, since that's a rarer combo we'll confirm the resource/endpoint for first. For `prepare`/`resend`: the re-issue is followed by an attempt, which is already wired — flag me if the server can gate on `prepare` itself.

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
zourzouvillys: You're right that a plain GET doesn't re-mint, so I capped it (`MAX_EXPIRED_RELOADS`) and, after a reload that comes back still-expired, fail loud with a retryable error instead of spinning. Combined with keying the effect on the token (see the `:192` thread), it can't loop. The real fix — re-running the gated step vs. FAPI re-minting on read — is a server-side decision we'll settle separately. (And confirmed: `expires_at` is Unix-epoch **milliseconds**, so the `< Date.now()` check is unit-correct; I added a JSDoc noting the unit so it's unambiguous going forward.)

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
zourzouvillys: Fixed — the effect is now keyed on `protectCheck?.token` (a primitive), so an unrelated `fromJSON` refresh no longer restarts the challenge, while a genuine chained challenge (new token) still re-runs. Applied in both cards via the shared `useProtectCheckRunner` hook.

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
zourzouvillys: Added both. The error state now renders a "Try again" control (`signIn/up.protectCheck.retryButton`) that clears the error and re-runs from scratch, and there's a 60s timeout (`Promise.race`) around the SDK run that aborts and surfaces a retryable `protect_check_timed_out`.

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
zourzouvillys: Good find — fixed. The continuation now runs through one `onResolved(resource)` path that finalizes (`setActive`) the `complete` case from both the normal success and the `protect_check_already_resolved` reload, so neither bounces to the start form with an unactivated session. Added a test covering the reload-reveals-complete case.

packages/clerk-js/src/core/clerk.ts
zourzouvillys: Great catch, and thanks for tracing it through clerk_go. Fixed: both gate checks are now scoped to the callback intent — `si.protectCheck` only routes when `params.reloadResource !== 'signUp'`, and the `su.protectCheck` check is symmetric (`!== 'signIn'`). So a Google sign-up callback no longer picks up the stale, abandoned sign-in's `protect_check`. Undefined intent keeps the legacy behavior, and transfers are unaffected since the `signIn.create({ transfer })` path checks its own fresh response. Added two `handleRedirectCallback` tests (sign-up callback completes instead of routing into the stale gate; sign-in callback routes to `/sign-in#/protect-check`).

packages/shared/src/types/json.ts
zourzouvillys: Done — `protect_check?` is now optional on `SignUpJSON`, matching `SignInJSON` and the wire format for older FAPI versions.

.changeset/protect-check-support.md
zourzouvillys: Added `'@clerk/react': minor`. You're right — `stateProxy.ts` puts `protectCheck`/`submitProtectCheck` on the public state proxies, so it's user-facing surface that needs its own changelog entry.

packages/clerk-js/src/core/resources/__tests__/SignIn.test.ts
zourzouvillys: Both fixed, in `SignIn.test.ts` and `SignUp.test.ts`: added an `afterEach` that restores `BaseResource._fetch` (saved at describe scope) so the mock can't leak, and the `submitProtectCheck` assertion now checks `path: '/client/sign_{ins,ups}/{id}/protect_check'` so a wrong/dropped action would fail the test.

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
zourzouvillys: Reworked in both cards: the bare loading text is now a `descriptors.spinner` `Spinner` inside an `aria-live='polite'` region, and the SDK container is marked `aria-busy` while running — so screen readers announce it and themes can target it. Went with the inline `Spinner` (above the SDK container) rather than full `LoadingCardContainer`, but happy to switch if you'd prefer the consistency.

zourzouvillys:
@Ephem on the breaking-change question: `'needs_protect_check'` is type-additive — existing runtime code keeps working (the server emits it only behind a feature gate, and `protectCheck` is the authoritative field). The one surface is strict-TypeScript consumers with an exhaustive `switch (signIn.status)`, who'll get a new unhandled-branch hint — which is why it's a `minor`. On the draft question: the recent review rounds are now addressed (component effect bugs, routing dead-ends, no-RHC guard, redirect-callback flow scoping), so I'm marking it ready for review.

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2798,6 +2850,14 @@ export class Clerk implements ClerkInterface {
       });
     };
 
+    // A Clerk Protect challenge can gate the inline web3 sign-in (no redirect happens, so the
+    // centralized _handleRedirectCallback check never runs). Route to the sign-in protect-check
+    // card before the status switch below, otherwise the user is stranded on the wallet step.
+    if (!viaSignUp && (signInOrSignUp.protectCheck || signInOrSignUp.status === 'needs_protect_check')) {
+      await navigateToSignInProtectCheck();
+      return;
+    }
```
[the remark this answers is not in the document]
zourzouvillys: Good catch — fixed in 3a36fc832. When the inline web3 attempt falls back to `signUp.authenticateWithWeb3()` and the sign-up is gated, we now route to the sign-up protect-check: added a `signUpProtectCheckUrl` param (combined flow passes `'create/protect-check'`, otherwise the absolute `/sign-up#/protect-check`), and the gate check is now `viaSignUp ? navigateToSignUpProtectCheck : navigateToSignInProtectCheck`.

A dedicated `identifier_not_found → signUp.authenticateWithWeb3() → protectCheck` regression test is a sensible fast-follow: `authenticateWithWeb3` has no existing unit harness and the `web3` helper comes from the `../utils` barrel (awkward to mock in isolation), so I left the test as a follow-up rather than ship a brittle one.

packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -195,6 +197,14 @@ export class SignUp extends BaseResource implements SignUpResource {
     });
   };
 
+  submitProtectCheck = (params: { proofToken: string }): Promise<SignUpResource> => {
+    debugLogger.debug('SignUp.submitProtectCheck', { id: this.id });
+    return this._basePatch({
+      action: 'protect_check',
+      body: { proof_token: params.proofToken },
+    });
+  };
```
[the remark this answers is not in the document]
zourzouvillys: Added in 3a36fc832 — JSDoc on `SignUp.submitProtectCheck` describing params, return value, and behavior. Also added it to the `SignIn` equivalent and both future-resource variants for symmetry.

packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -1141,6 +1173,15 @@ class SignUpFuture implements SignUpFutureResource {
...
   }
 
+  async submitProtectCheck(params: { proofToken: string }): Promise<{ error: ClerkError | null }> {
+    return runAsyncResourceTask(this.#resource, async () => {
+      await this.#resource.__internal_basePatch({
+        action: 'protect_check',
+        body: { proof_token: params.proofToken },
+      });
+    });
+  }
```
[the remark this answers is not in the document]
zourzouvillys: Added in 3a36fc832 (covered alongside the public method and the `SignIn` equivalents).

packages/ui/src/components/SignIn/shared.ts
```diff
@@ -10,17 +10,30 @@ import { handleError } from '@/ui/utils/errorHandler';
...
+ * @param onSecondFactor - invoked when the passkey attempt resolves to a second factor.
+ * @param protectCheckPath - route to the protect-check card relative to the caller's mount.
+ *   Defaults to the factor-one mount (`'../protect-check'`); `SignInStart` (index route) must pass
+ *   `'protect-check'`, otherwise an autofill-triggered, gated passkey sign-in dead-ends at the app
+ *   root instead of `/sign-in/protect-check`.
+ */
+function useHandleAuthenticateWithPasskey(
+  onSecondFactor: () => Promise<unknown>,
+  protectCheckPath = '../protect-check',
+) {
```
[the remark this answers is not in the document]
zourzouvillys: Added in 3a36fc832 — annotated as `(...args: Parameters<SignInResource['authenticateWithPasskey']>) => Promise<unknown>`. I used the resource type rather than `typeof authenticateWithPasskey` (which isn't in scope at the signature position), and `Promise<unknown>` since the callback's branches return non-void values.

On the related outside-diff note about the empty `useCallback` dependency array: that `[]` is the pre-existing intentional pattern in this hook (it pairs with the `__internal_WebAuthnAbortService.abort()` cleanup and the autofill flow); my change only added the stable `protectCheckPath` string to the closure. Filling the dep array would re-create the callback on every render and risks re-firing the autofill effect, so I've left it as-is.

Ephem:
@zourzouvillys Alright, that makes it clearer, thanks! Some follow up questions.

> the server emits it only behind a feature gate

So if a customer is currently using a custom flow and not handling it, flipping that feature gate is the potential breaking change? I'm guessing we do that pretty intentionally though so doesn't seem like a problem, just something to be aware of.

> The one surface is strict-TypeScript consumers with an exhaustive switch (signIn.status), who'll get a new unhandled-branch hint

So some small subset of users might _perceive_ this as a breaking change if they see that? This part of the changeset kinda explains it: "surfaced when the server-side SDK-version gate is enabled", but it's not entirely clear that's something we only enable in dialogue with the customer, so might be worth adding some extra clarity there that this is not something we'll turn on randomly. If I see that TS hint after an upgrade I might go to the changeset and will want to know what to do. Could maybe include info in the typedoc too if there is none already (didn't check)?

That's just a small NIT though. 😄

packages/ui/src/components/SignIn/__tests__/SignInFactorOne.test.tsx
```diff
@@ -106,6 +106,34 @@ describe('SignInFactorOne', () => {
...
+
+      await userEvent.type(screen.getByLabelText('Password'), '123456');
+      await userEvent.click(screen.getByText('Continue'));
+
+      await waitFor(() => {
+        expect(fixtures.router.navigate).toHaveBeenCalledWith('../protect-check');
+        // and must not fall through to a status-based route
+        expect(fixtures.clerk.setActive).not.toHaveBeenCalled();
+      });
+    });
```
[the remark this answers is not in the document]
zourzouvillys: Good catch — fixed in 15eea0e68. Switched to `f.startSignInWithEmailAddress({ supportPassword: true })` so the started flow matches the `withEmailAddress()` / `withPassword()` user fixture. (I'd mirrored the sibling password tests, which use the phone fixture, but you're right that the consistent email fixture is clearer here.)

zourzouvillys:
@Ephem Yep, exactly right — the npm upgrade itself is type-only. The runtime change only ever happens when Protect mid-flow challenges are explicitly enabled for an instance: it's currently behind a flag and **not** enabled for existing instances, so nothing turns on by itself. There's a second layer too — the server only emits the new status value to SDK versions that understand it, so older clients never receive an unknown status either way.

Good call on the changeset — "surfaced when the server-side SDK-version gate is enabled" didn't really tell an upgrading user what they needed to know. Pushed c2795e441:

- **Changeset** now says explicitly: upgrading is type-only, the feature is off by default and must be explicitly enabled per instance, and if an exhaustive `switch` on `signIn.status` flags the new value, handle it via `protectCheck` + `submitProtectCheck()` (the prebuilt components do it automatically).
- **Typedoc**: there was some on `protectCheck`/`submitProtectCheck` but nothing on the status itself — added `'needs_protect_check'` to the documented `SignInStatus` list and put the same "only when explicitly enabled for the instance; upgrading alone doesn't enable it" note on the `status`/`protectCheck` properties and `ProtectCheckResource`.

Ephem:
@zourzouvillys Sounds great, thanks! 🙏

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2423,7 +2429,18 @@ export class Clerk implements ClerkInterface {
         ),
     );
 
+    const navigateToSignUpProtectCheck = makeNavigate(
```
jacekradko, line 2433: Was working through this with Claude and it flagged that in combined flow, a Protect-gated OAuth/SAML sign-up routes to the standalone `/sign-up#/protect-check` instead of `/sign-in#/create/protect-check`, popping the user out of `<SignIn/>`. Looks like `navigateToSignUpProtectCheck` uses the absolute `displayConfig.signUpUrl` with no relative override, unlike the web3 path / `continueSignUpUrl`. Can you take a look?

jacekradko:
Trying to trace how the protect gate interacts with the MFA email-link card. The other second-factor cards route through it now, but `SignInFactorTwoEmailLinkCard` looks like it calls `setActive` directly in `handleVerificationResult` without the gate check. If Protect can gate that step, would `createdSessionId` be null here and send us down the sign-out path instead of the challenge? Wondering if it got skipped because this card doesn't go through `completeSignInFlow` like the rest.

packages/ui/src/components/SignIn/SignInFactorOneSolanaWalletsCard.tsx
```diff
@@ -67,6 +67,8 @@ const SignInFactorOneSolanaWalletsCardInner = () => {
                       customNavigate: router.navigate,
                       redirectUrl: ctx.afterSignInUrl || '/',
                       secondFactorUrl: 'factor-two',
+                      protectCheckUrl: 'protect-check',
```
jacekradko: isn't this card mounted under `choose-wallet`? so the bare `protect-check` would resolve to `/sign-in/choose-wallet/protect-check`, not the actual route.

packages/ui/src/components/SignIn/SignInFactorOneCodeForm.tsx
```diff
@@ -104,6 +105,10 @@ export const SignInFactorOneCodeForm = (props: SignInFactorOneCodeFormProps) =>
       .then(async res => {
         await resolve();
 
+        if (navigateOnSignInProtectGate(res, navigate, '../protect-check')) {
```
jacekradko, line 118: The attempt is gated here, but the resend and initial prepare just above drop the response. Can a protect gate come back on prepare (the code-send itself)? The docstring on `navigateOnSignInProtectGate` lists prepare as a call site, so these might need to route through it too, unless the server never gates that step.

jacekradko:
The routing tests mostly take the path as an argument and assert it back, so they can't catch a wrong literal when invoked. `SignInFactorOne` is the only render test pinning a real path, which is why the combined-flow OAuth sign-up gate and the Solana wallet gate both slipped through pointing at the wrong route. A render test per entry point asserting the actual `navigate(...)` would catch this class.

references/mosaic-architecture.md
```diff
@@ -58,7 +58,7 @@ Single provider that handles cascade and theme delivery. It accepts `appearance`
 import { MosaicProvider } from '../mosaic/MosaicProvider';
 
 <MosaicProvider>{children}</MosaicProvider>;
-```
+````
```
jacekradko: This Mosaic doc edit looks unrelated to the protect_check work, probably stray drift. Worth reverting it out of this PR?

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2410,6 +2412,10 @@ export class Clerk implements ClerkInterface {
         buildURL({ base: displayConfig.signInUrl, hashPath: '/reset-password' }, { stringify: true }),
     );
 
+    const navigateToSignInProtectCheck = makeNavigate(
```
jacekradko, line 2415: This one has the same gap as the sign-up nav below, and it generalizes: the factor/continue/verify navs all honor a `params.*Url` override, but both protect-check navs always use the `displayConfig` URL. A custom-mounted or path-based flow keeps every step in context except the gate. There's also no protect-check field on the callback params to override it.

packages/ui/src/hooks/useProtectCheckRunner.ts
```diff
@@ -0,0 +1,239 @@
...
+      if (reloadCountRef.current >= MAX_EXPIRED_RELOADS) {
+        failWith(ERROR_CODES.PROTECT_CHECK_TIMED_OUT, 'Protect verification expired');
+        return;
+      }
+      reloadCountRef.current += 1;
+      isRunningRef.current = true;
+      setIsRunning(true);
+      void (async () => {
+        try {
+          await reload();
```
jacekradko, line 143: After the reload, only the still-expired case is handled. If the reload instead clears the gate or completes the flow, nothing calls `onResolved`, so the route guard bounces to flow start instead of continuing (and a completed sign-in wouldn't get `setActive`). Worth routing on the refreshed resource here, not just failing when still expired.

packages/ui/src/hooks/useProtectCheckRunner.ts
```diff
@@ -0,0 +1,239 @@
...
+    // remote `import(sdk_url)` we must not perform there. This guard MUST live in the component
+    // layer — `executeProtectCheck` is in `@clerk/shared`, compiled once with the flag hard-coded
+    // `false`, so a guard there would never trip.
+    if (__BUILD_DISABLE_RHC__) {
+      failWith(
+        ERROR_CODES.PROTECT_CHECK_UNSUPPORTED_ENVIRONMENT,
+        'Protect verification is not supported in this environment',
+      );
+      return;
+    }
```
brkalow, line 128: I expect this will still get flagged, even though the `import()` is dead code. One solution would be to load `executeProtectCheck()` async behind the build-time flag.

packages/clerk-js/src/core/clerk.ts
zourzouvillys: Good catch — fixed in `48d4a00df`. `navigateToSignUpProtectCheck` now honors a new `params.signUpProtectCheckUrl` override (falling back to the `displayConfig` URL as before). The SignIn context builds it off `signUpUrl`, which is rewritten to `<signInUrl>#/create` in the combined flow, so a gated combined-flow OAuth/SAML sign-up now lands on `…/create/protect-check` and stays inside `<SignIn/>` instead of popping out to the standalone `/sign-up`.

packages/clerk-js/src/core/clerk.ts
zourzouvillys: Agreed, generalized it. Both protect-check navs now take per-caller overrides — `signInProtectCheckUrl` / `signUpProtectCheckUrl` added to `HandleOAuthCallbackParams`, mirroring the existing `firstFactorUrl`/`continueSignUpUrl` pattern. They're threaded through the callback builders (`buildOAuthCallbackParams`), the SignIn context, the standalone SignUp SSO callback, and Google One Tap, so a custom-mounted / path-based flow keeps the gate in context too. Fixed in `48d4a00df`.

packages/ui/src/components/SignIn/SignInFactorOneSolanaWalletsCard.tsx
zourzouvillys: Right — this card mounts at `/sign-in/choose-wallet`, so the bare paths were off by one level (they were copied verbatim from `SignInSocialButtons`, which lives at the sign-in root). Fixed in `48d4a00df`: `protectCheckUrl` → `../protect-check` and `signUpProtectCheckUrl` → `../create/protect-check`. I also fixed the pre-existing `secondFactorUrl`/`signUpContinueUrl`, which had the same off-by-one from #7450 — happy to pull those two back out if you'd rather keep this PR strictly protect-check. Added a render test asserting the literal params forwarded to `authenticateWithWeb3`.

packages/ui/src/components/SignIn/SignInFactorOneCodeForm.tsx
zourzouvillys: Routed both `prepare` and the initial `useFetch` prepare through `navigateOnSignInProtectGate` in `48d4a00df` (resend reuses the same `prepare` handler, so it's covered too). It's a no-op when the response isn't gated, so it's safe whether or not the server actually gates the prepare step — defensive and matches the docstring's call-site list. Added a render test driving the fetch `onSuccess` with a gated response.

packages/ui/src/hooks/useProtectCheckRunner.ts
zourzouvillys: Fixed in `48d4a00df`. After the expired-challenge reload, if the gate clears or the flow completes we now route on the refreshed resource (so a completed sign-in still gets `setActive` instead of the route guard bouncing to flow start).

One subtlety worth calling out: clearing `protectCheck` flips the effect's `token` dependency, which re-runs and *cancels* the in-flight reload effect — so I keyed the routing on a separate `mountedRef` (true until real unmount) rather than the per-run `cancelled` flag. Otherwise that re-run would abort the very routing it's supposed to trigger. Added a test for the expired-reload-clears-gate case.

packages/ui/src/hooks/useProtectCheckRunner.ts
zourzouvillys: Done in `48d4a00df` — `executeProtectCheck` is now lazy-`import()`ed inside `runChallenge`, gated on `__BUILD_DISABLE_RHC__`. In no-RHC builds the flag is `true`, so the branch (and the dynamic import below it) is dead-code-eliminated — the loader and its remote `import(sdk_url)` are tree-shaken out of those bundles entirely rather than shipped-but-unused, instead of being pulled in by the old static top-level import.

references/mosaic-architecture.md
zourzouvillys: Reverted in `48d4a00df` — yep, stray drift (a stray 4-backtick fence). Out of the PR now.

zourzouvillys:
Pushed `48d4a00df` addressing this round of feedback. Replies inline on each thread; two things that came up as top-level comments rather than inline:

**`SignInFactorTwoEmailLinkCard` finalizing without the gate check** — confirmed: it finalizes inline (doesn't go through `completeSignInFlow` like the other second-factor cards), so a gate would have hit `setActive` with a null `createdSessionId`. Added a gate check in `handleVerificationResult` before the `setActive`, routing to `../protect-check` when gated. Covered by a new render test that mocks the email-link flow resolving gated.

**Test gap (wrong-literal slips through arg-passthrough tests)** — agreed. Added per-entry-point render/assertion tests that pin the *actual* target rather than echoing the passed arg:
- `SignInFactorOneSolanaWalletsCard` — asserts the literal params forwarded to `authenticateWithWeb3` (`../protect-check`, `../factor-two`).
- `SignInFactorOneCodeForm` — drives the prepare `onSuccess` with a gated response, asserts `navigate('../protect-check')`.
- `SignInFactorTwoEmailLinkCard` — asserts `navigate('../protect-check')` + no `setActive`.
- `SignInProtectCheck` — expired-reload-clears-gate routes on the refreshed resource.
- `buildOAuthCallbackParams` — locked the new `signInProtectCheckUrl`/`signUpProtectCheckUrl` literals for every callback variant.

Verification: 124 UI + 275 clerk-js tests pass · `@clerk/ui` typecheck clean · clerk-js declarations build clean · 0 lint errors.

One open call: I also fixed the pre-existing `secondFactorUrl`/`signUpContinueUrl` off-by-one in the Solana card (same root cause from #7450) — flagged on that thread; easy to pull back out if you'd prefer to keep this PR strictly protect-check.

jacekradko (approved):
I think this is in a good enough place to merge 👍🏼

PR #8331, opened 2026-04-16
chore(backend): Harden `joinPaths` against traversal

dominic-clerk:
## Description

There is no immediate vulnerability but it could be possible that someone uses one of the functions that call `joinPath` in a way that accepts user input and allows traversal on the backend API. This adds protection to avoid this problem.

Fixes SDK-60

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/dark-bars-wink.md
+++ b/.changeset/dark-bars-wink.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
+Add path traversal protections in `joinPaths`
--- a/packages/backend/src/util/path.ts
+++ b/packages/backend/src/util/path.ts
@@ -1,14 +1,18 @@
 const SEPARATOR = '/';
 const MULTIPLE_SEPARATOR_REGEX = new RegExp('(?<!:)' + SEPARATOR + '{1,}', 'g');
+const MAX_DECODES = 10;
 
 type PathString = string | null | undefined;
 
 function isDotSegment(segment: string): boolean {
   let candidate = segment;
-  for (let i = 0; i < 3; i++) {
+  for (let i = 0; i <= MAX_DECODES; i++) {
     // After decoding, check if any slash-separated part is a dot segment
     if (candidate.split(/[/\\]/).some(p => p === '.' || p === '..')) {
       return true;
+    }
+    if (i === MAX_DECODES) {
+      throw new Error(`joinPaths: too many layers of encoding in ${segment}`);
     }
     try {
       const next = decodeURIComponent(candidate);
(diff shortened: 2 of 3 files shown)


.changeset/dark-bars-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/backend': patch
```
dominic-clerk, line 2: Minor maybe? I wasn't sure

packages/backend/src/util/path.ts
```diff
@@ -3,9 +3,27 @@ const MULTIPLE_SEPARATOR_REGEX = new RegExp('(?<!:)' + SEPARATOR + '{1,}', 'g');
...
 
+function isDotSegment(segment: string): boolean {
+  let decoded: string;
+  try {
+    decoded = decodeURIComponent(segment);
+  } catch {
+    decoded = segment;
+  }
+  return decoded === '.' || decoded === '..';
+}
```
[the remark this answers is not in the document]
dominic-clerk: I have a fix for `%2e%2e%2fusers` but does the double-encoded variant make sense to check here?

packages/backend/src/util/path.ts
```diff
@@ -3,9 +3,37 @@ const MULTIPLE_SEPARATOR_REGEX = new RegExp('(?<!:)' + SEPARATOR + '{1,}', 'g');
...
+      const next = decodeURIComponent(candidate);
+      if (next === candidate) {
+        break;
+      } // stable — no more encoding
+      candidate = next;
+    } catch {
+      break;
+    }
+  }
+  return false;
```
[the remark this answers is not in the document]
dominic-clerk: We could decode until infinity but realistically this is enough

nikosdouvlis, line 27: I think CodeRabbit is right about this one. The loop runs `i = 0,1,2`, and at `i=2` we decode a 3rd time but exit before that decoded value gets checked. So `%25252e%25252e%25252fusers` walks down to `../users` and we still return `false`. Easy fix is bumping the loop to `i <= 3` and breaking before the decode at `i === 3`, so the final decoded candidate gets checked without adding a 4th decode round

dominic-clerk: Oooh right I misread the finding and thought he meant we'd miss out on the `i == 3` loop. Thanks!

dominic-clerk: I fixed it and changed the approach a bit. We decode until it's clean and there's a max number of 10 iterations where it becomes just silly and for sure someone's doing something strange and we reject it.

nikosdouvlis (approved):

PR #8332, opened 2026-04-16
fix(backend): Verify signature before claims

dominic-clerk:
## Description

By verifying the signature first we avoid leaking information about which claims are valid if the signature isn't valid to begin with. Auth couldn't be bypassed, it leaked configuration details that could aid further attacks.

Fixes SDK-61

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/five-eagles-tap.md
+++ b/.changeset/five-eagles-tap.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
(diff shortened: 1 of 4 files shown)


wobsoriano (approved):
Thanks Dominic!

PR #8341, opened 2026-04-16
fix(ui): show claim CTA in EnableOrganizationsPrompt for keyless mode

mwickett:
## Summary

- Fixes the `EnableOrganizationsPrompt` success state in keyless mode: replaces the broken "Sign in to continue" path (which cannot reach a usable dashboard flow on an unclaimed keyless app) with a "Claim your application" anchor, plus a secondary action that delegates to `onSuccess`
- The keyless claim path now activates for any unclaimed keyless app when `isKeyless && !isClaimed`, including signed-in users, because the dashboard link is inaccessible until the app is claimed
- The real caller still reloads after the secondary action by passing `onSuccess: () => window.location.reload()` from `clerk-js`; the UI component keeps that behavior delegated to its caller instead of hardcoding reload itself
- Detection uses `useOptions()` and `useEnvironment()` already available in the component — no prop threading, no changes to `clerk-js`, shared types, or any SDK package beyond the existing caller contract

## Test plan

- [x] Release canary and verify in a keyless Next.js app:
  1. Mount `<OrganizationSwitcher />` without signing in
  2. Click "Enable Organizations"
  3. Success state should show "Claim your application" + "I'll do it later" (not "Sign in to continue")
- [x] Verify signed-in keyless user now sees "Continue" + "Claim your application" with the default org name
- [x] Verify claimed keyless app (no user) still sees "Sign in to continue"
- [x] Verify non-keyless dev instance still sees "Sign in to continue"
- [x] Unit tests: 5 new tests covering all 4 state scenarios (all passing)

### DIFF
--- a/packages/ui/src/components/devPrompts/EnableOrganizationsPrompt/__tests__/EnableOrganizationsPrompt.test.tsx
+++ b/packages/ui/src/components/devPrompts/EnableOrganizationsPrompt/__tests__/EnableOrganizationsPrompt.test.tsx
@@ -48,10 +48,14 @@ describe('EnableOrganizationsPrompt success state', () => {
       expect(claimLink).toHaveAttribute('target', '_blank');
       expect(claimLink).toHaveAttribute('rel', 'noopener noreferrer');
 
-      // href should include the claim URL with return_url appended
+      // href should stay raw until the user clicks, then pick up return_url
       const href = claimLink.getAttribute('href')!;
       expect(href).toContain('dashboard.clerk.com/claim');
-      expect(href).toContain('return_url=');
+      expect(href).not.toContain('return_url=');
+
+      await claimLink.click();
+
+      expect(claimLink.getAttribute('href')).toContain('return_url=');
 
       // Should show "I'll do it later" button
       expect(screen.getByRole('button', { name: /i.ll do it later/i })).toBeInTheDocument();
--- a/packages/ui/src/components/devPrompts/EnableOrganizationsPrompt/__tests__/EnableOrganizationsPrompt.test.tsx
+++ b/packages/ui/src/components/devPrompts/EnableOrganizationsPrompt/__tests__/EnableOrganizationsPrompt.test.tsx
@@ -98,7 +98,7 @@ describe('EnableOrganizationsPrompt success state', () => {
   });
 
   describe('unclaimed keyless, signed-in user', () => {
-    it('renders Continue button and preserves default org name', async () => {
+    it('renders claim CTA with Continue button and preserves default org name', async () => {
       const onSuccess = vi.fn();
       const mockEnableSetting = vi.fn().mockResolvedValue(undefined);
       const mockGetMemberships = vi.fn().mockResolvedValue({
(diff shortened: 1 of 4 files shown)


mwickett:
!snapshot fix-keyless-orgs-enablement

wobsoriano:
!snapshot

packages/ui/src/components/devPrompts/EnableOrganizationsPrompt/index.tsx
```diff
@@ -31,6 +31,21 @@ const EnableOrganizationsPromptInternal = ({
...
+  const isKeyless = Boolean(claimUrl) && Boolean(copyKeysUrl);
+  const isClaimed = environment.authConfig.claimedAt !== null;
+  const showKeylessClaimPath = isEnabled && isKeyless && !isClaimed;
+
+  const claimUrlWithReturnUrl = useMemo(() => {
+    if (!showKeylessClaimPath || !claimUrl) return undefined;
+    const url = new URL(claimUrl);
+    url.searchParams.append('return_url', window.location.href);
+    return url.href;
+  }, [showKeylessClaimPath, claimUrl]);
```
wobsoriano: small nit: computation here is not expensive enough so we can just put the logic inside onClick to compute at click time

wobsoriano (approved):
tested locally with the snapshot, working 👍🏼 thanks for the tests!

lgtm

packages/ui/src/components/devPrompts/EnableOrganizationsPrompt/index.tsx
mwickett: oh good call - fixed in https://github.com/clerk/javascript/pull/8341/commits/d9141033839e9ecb501497a94aa043d2ef012384

packages/ui/src/components/devPrompts/EnableOrganizationsPrompt/__tests__/EnableOrganizationsPrompt.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+      await vi.waitFor(() => {
+        expect(screen.getByText(/Organizations are now enabled/i)).toBeInTheDocument();
+      });
+
+      const laterButton = screen.getByRole('button', { name: /i.ll do it later/i });
+      await laterButton.click();
+
+      expect(onSuccess).toHaveBeenCalled();
+    });
```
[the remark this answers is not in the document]
mwickett: Checked against the current branch:

- `showKeylessClaimPath` is intentionally widened to all unclaimed keyless apps, not just `!clerk.user`, because the dashboard link is inaccessible until the app is claimed. Signed-in users now see `Continue` + `Claim your application` on purpose.
- `EnableOrganizationsPrompt` is expected to call `onSuccess`; the actual reload happens at the caller. `packages/clerk-js/src/core/clerk.ts` opens this prompt with `onSuccess: () => window.location.reload()`, so the unit test asserting `onSuccess` is called is still the right contract here.

Updated the PR description to match the current behavior.

PR #8350, opened 2026-04-17
refactor(ui): Use existing organizationMemberships and remove infinite scrolling

wobsoriano:
## Description

When the Organizations feature was disabled in the dashboard, the OAuth consent screen triggered the "Organizations feature required" dev dialog. This happened because `useOrganizationList` was called unconditionally whenever `enableOrgSelection` was set — even passing `undefined` to the hook still invokes `useAttemptToEnableOrganizations` internally.

The fix replaces `useOrganizationList` entirely with `user.organizationMemberships`, which is populated synchronously from the user object alongside the initial session load. The FAPI includes the full, unpaginated list of memberships in the user response (confirmed in the Go backend — no cap or truncation), so we get everything needed (org id, name, imageUrl) with no extra fetch, no loading state, and no spinner.

The org select is now gated on both `ctx.enableOrgSelection` and `organizationSettings.enabled`, and renders immediately without pop-in.

Resolves SDK-63

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/components/OAuthConsent/OrgSelect.tsx
```diff
@@ -113,10 +101,7 @@ export function OrgSelect({ options, value, onChange, hasMore, onLoadMore }: Org
           {selected?.label || 'Select an option'}
         </Text>
       </SelectButton>
-      <SelectOptionList
-        footer={hasMore ? <InfiniteListSpinner ref={loadMoreRef} /> : null}
-        onReachEnd={hasMore ? onLoadMore : undefined}
-      />
+      <SelectOptionList />
```
wobsoriano, line 104: I will remove the added infinite scrolling props to this shared element in a follow up PR

alexcarpenter (approved):

LauraBeatris (approved):

PR #8351, opened 2026-04-18
chore(repo): upgrade monorepo to Node.js 24

jacekradko:
## Summary
- Upgrades the monorepo development environment to Node.js 24
- Maintains Node ≥20.9.0 compatibility for published packages
- Updates CI workflows to test both Node 24 and Node 20

## Changes
- Updated root `package.json` engines to require Node ≥24.15.0
- Updated `.nvmrc` to Node 24.15.0
- Updated CI workflows to use Node 24 as primary version
- Added Node 20 to CI test matrix for published package compatibility
- Updated release workflow caching to Node 24

## Test plan
- [x] Build passes with Node 24
- [x] Tests pass with Node 24
- [x] Published packages still work with Node 20 (with expected engine warnings)
- [x] CI configuration updated to test both versions


.github/workflows/ci.yml
```diff
@@ -218,7 +218,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - node-version: 22
+          - node-version: 24.15.0
+            test-filter: "**"
+            filter-label: "**"
+          - node-version: 20
```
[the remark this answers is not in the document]
jacekradko: Keeping the Node 20 entry intentionally — every published package still declares `"engines": { "node": ">=20.9.0" }`, so testing against the lowest supported runtime catches regressions for consumers on Node 20. The root `>=24.15.0` is a dev/build requirement only, and `engine-strict=false` in `.npmrc` lets pnpm install proceed regardless. Pinned the matrix to `20.9.0` exactly so we're testing the actual minimum supported version.

.claude/scheduled_tasks.lock
bratsos, line 1: ✂️

.changeset/giant-mirrors-find.md
bratsos, line 1: ✂️

packages/backend/src/tokens/clerkRequest.ts
```diff
@@ -26,7 +26,25 @@ class ClerkRequest extends Request {
...
+      cloneInit = new Proxy(input as Request, {
+        get(target, prop) {
+          if (prop === 'signal') {
+            return undefined;
+          }
+          return Reflect.get(target, prop, target);
+        },
+      }) as unknown as RequestInit;
+    }
+    super(url, cloneInit);
```
wobsoriano, line 47: Can confirm this makes sense. Undici validates `init.signal` when constructing a `Request`, and that can reject signals coming from framework-specific request objects

https://github.com/nodejs/undici/blob/main/lib/web/fetch/request.js#L1077-L1085

packages/react-router/src/server/utils.ts
```diff
@@ -135,12 +135,14 @@ export const wrapWithClerkState = (data: any) => {
  * @internal
  */
 export const patchRequest = (request: Request) => {
+  // Omit `signal` from the clone: Node 24's bundled undici tightened the
+  // instanceof AbortSignal check, which rejects cross-realm signals (e.g.
+  // those carried by framework Request subclasses).
```
wobsoriano, line 140: We also have this in tanstack start, let's omit signal there also https://github.com/clerk/javascript/blob/7913812490027cfcb3ba36275e5e5851e2c7bf72/packages/tanstack-react-start/src/server/utils/index.ts#L77

Not sure if the underlying risk can apply here but worth checking https://github.com/clerk/javascript/blob/7913812490027cfcb3ba36275e5e5851e2c7bf72/packages/backend/src/proxy.ts#L305

wobsoriano (changes requested):

packages/backend/src/tokens/clerkRequest.ts
```diff
@@ -26,7 +26,25 @@ class ClerkRequest extends Request {
     // https://github.com/nodejs/undici/issues/2155
     // https://github.com/nodejs/undici/blob/7153a1c78d51840bbe16576ce353e481c3934701/lib/fetch/request.js#L854
     const url = typeof input !== 'string' && 'url' in input ? input.url : String(input);
-    super(url, init || typeof input === 'string' ? undefined : input);
+    // When cloning a Request by passing it as init, hide its `signal`. Undici's
+    // Request constructor in Node 24 performs a strict instanceof check on the
+    // signal and rejects ones from a different realm (e.g. NextRequest). Using a
+    // Proxy keeps property access lazy so environments that don't implement
+    // optional getters (e.g. Cloudflare Workers' Request lacks `cache`) still work.
```
bratsos, line 33: ❓ Shouldn't we also handle other call sites like:
- https://github.com/clerk/javascript/blob/dcaf694fbc7fd1b80fd10661225aa6d61eb3c2a9/packages/backend/src/proxy.ts#L305
- https://github.com/clerk/javascript/blob/dcaf694fbc7fd1b80fd10661225aa6d61eb3c2a9/packages/tanstack-react-start/src/server/utils/index.ts#L77

packages/react-router/src/server/utils.ts
jacekradko: Done!

packages/backend/src/tokens/clerkRequest.ts
jacekradko: Yes! Both call sites are now patched in the latest commit.

wobsoriano (approved):
🚀

PR #8352, opened 2026-04-18
chore(ui,shared): Remove unused OAuthConsent props

wobsoriano:
## Description

Now that we've implemented OAuthConsent component use in AP (https://github.com/clerk/accounts/pull/1418), we can safely remove the legacy callback path that was built to support the old AP integration.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


jacekradko (changes requested):
I think this makes sense to remove, but it is a breaking change even though the use case was only internal. I think we will need to move this to Core 4

wobsoriano:
> I think this makes sense to remove, but it is a breaking change even though the use case was only internal. I think we will need to move this to Core 4

@jacekradko that's fair, thanks for looking!

I'll close it and come back later

PR #8359, opened 2026-04-20
fix(backend): Clock skew of 0 should not fall back

dominic-clerk:
## Description

Because 0 is falsy the current code fell back to the default value when the clock skew was configured to 0. This changes the syntax to fall back on null-ish values which 0 is not.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/wacky-dryers-hammer.md
+++ b/.changeset/wacky-dryers-hammer.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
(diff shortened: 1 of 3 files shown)


jacekradko (commented):
Good enhancement, but we should add tests for this

wobsoriano (approved):
thanks for the tests!

jacekradko (approved):

PR #8362, opened 2026-04-20
fix(ui): Default OAuthConsent org selection to last active organization

kylemac:
## Description

The `<OAuthConsent />` org picker currently defaults to whichever org the API returns first. This change seeds the default from `clerk.session?.lastActiveOrganizationId` so users consent on behalf of the org they were most recently working in, falling back to the first membership when the last active org is null, doesn't match any membership, or was deleted. Read via the already-in-scope `useClerk()` (no new hook, no re-renders from `useSession`), and the existing render guards keep the no-memberships case unchanged.

Added three test cases covering the match, no-match, and null paths, plus a `@clerk/ui` patch changeset.

## Checklist

- [x] \`pnpm test\` runs as expected.

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/oauth-consent-last-active-org.md
+++ b/.changeset/oauth-consent-last-active-org.md
@@ -1,0 +1,5 @@
+---
+"@clerk/ui": patch
+---
+
+Default the organization selection in `<OAuthConsent />` to the user's last active organization, falling back to the first membership when it is not set or no longer available.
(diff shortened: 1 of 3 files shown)


alexcarpenter (approved):

wobsoriano (approved):
Thanks, Kyle! Should we also set the selected organization as the active one? If yes, I can open a follow-up PR.

PR #8369, opened 2026-04-21
fix(nextjs): enforce middleware authorization during keyless bootstrap

jacekradko:
## Summary

Closes [SDK-70](https://linear.app/clerk/issue/SDK-70).

When `clerkMiddleware` detected no publishable key (the client-side keyless bootstrap window), it early-returned `NextResponse.next()` and **skipped the user's handler entirely**. Any authorization logic inside the handler — including `auth.protect()` — was bypassed during that window.

The fix runs the user's handler against a synthetic signed-out `RequestState` through the **same post-authentication pipeline** (`runHandlerWithRequestState`) that the keyed path uses. Authorization fails closed during bootstrap. `<ClerkProvider/>` downstream resumes the flow once keys are provisioned client-side. **Dev-only path — production behavior unchanged.**

## Edge cases — verified, no code changes needed

- **`decorateRequest` with empty keys**: `encryptClerkRequestData` already handles this via its `isEmpty` guard + `KEYLESS_ENCRYPTION_KEY` dev fallback.
- **`handleControlFlowErrors` redirect with empty pubkey**: If the user calls `auth.protect()` with no configured `signInUrl`, `createRedirect` throws `missingPublishableKeyError` — correct fail-closed behavior.
- **Frontend API proxy**: The proxy check lives in `baseNextMiddleware`, which the no-key branch never enters. Naturally short-circuited.

## Test plan

- [x] `@clerk/nextjs` builds
- [x] `@clerk/backend` tests pass (adds coverage for `createBootstrapSignedOutState`)
- [x] `@clerk/nextjs` tests pass (pre-existing `AbortSignal` failures unchanged — unrelated)
- [x] New Playwright integration test (`integration/tests/next-middleware-keyless.test.ts`) exercises the bypass end-to-end: boots the `next-app-router` template in `withKeyless` env, hits a middleware-protected route, asserts redirect to `/sign-in` instead of the protected content rendering.
- [ ] Manual smoke: fresh Next.js app with `clerkMiddleware((auth, req) => { auth.protect() })` on a protected route, no `.env.local`. Verify the protected route redirects during the keyless bootstrap window.
- [ ] Manual smoke: public route during keyless bootstrap still renders and still triggers key provisioning.


packages/nextjs/src/server/clerkMiddleware.ts
wobsoriano, line 1: I believe this fix only needs to be applied in Next SDK right? Since it has a special keyless bootstrap path before the PK is available. The other SDKs resolve keyless keys before auth runs, so they likely not need the same fix

wobsoriano (approved):
looks good 👍🏼

packages/nextjs/src/server/clerkMiddleware.ts
jacekradko: Yes, that is correct.

wobsoriano: thanks for answering

PR #8372, opened 2026-04-22
fix(shared,clerk-js,nextjs): authorization bypass in combined-condition has() checks

nikosdouvlis:
## Why

`has()`, `auth.protect()`, and related authorization predicates can return `true` for certain calls that combined more than one condition (role, permission, feature, plan, reverification) when the result should have been `false`. When a requested dimension could not be satisfied because the underlying session data was missing or indeterminate, the combining logic ignored that dimension instead of denying. If any remaining requested dimension passed, the overall result was `true`.

Separately, `auth.protect()` in `@clerk/nextjs` silently discarded authorization params (`role`, `permission`, `feature`, `plan`, `reverification`) whenever the same argument object also contained `unauthenticatedUrl`, `unauthorizedUrl`, or `token`. TypeScript's excess-property check caught this only for inline object literals; variable-assigned or spread arguments and any JavaScript caller let every authenticated user through.

## What changed

`packages/shared/src/authorization.ts`
- Internal helper return type is now a `'pass' | 'fail' | 'skip'` tri-state. The combiner treats any asked dimension returning `'fail'` as a denial and requires at least one `'pass'`. Missing, malformed, or invalid session data now denies instead of being treated as indeterminate.
- Removed the `billingAuthorization || orgAuthorization` OR-coercion; each asked dimension is evaluated independently.
- `role` / `permission` / `feature` / `plan` values that are not strings (e.g. `null` cast through `as any`) now fail closed instead of throwing.
- `factorVerificationAge` tuple shape validation (length, numeric, `-1 | >= 0`) before comparisons.
- Graceful downgrade preserved for `strict` / `strict_mfa` per the existing docs: when no second factor is enrolled, these levels still evaluate against the first factor. `multi_factor` still requires both factors when both are enrolled.

`packages/clerk-js/src/core/resources/Session.ts`
- `session.checkAuthorization()` now uses the active organization's id (`activeMembership.organization.id`) when building authorization options, not the membership row id.

`packages/clerk-js/src/test/core-fixtures.ts`
- Test fixtures now give memberships and organizations distinct ids so downstream regressions on the above are caught.

`packages/nextjs/src/server/protect.ts`
- `getAuthorizationParams` now picks the known authorization keys from the argument via an explicit allowlist. Mixed shapes like `{ role, unauthorizedUrl }` or `{ permission, token }` enforce the authorization check; options-only objects take the fast path; unknown extra keys are ignored.

`packages/react/src/hooks/__tests__/useAuth.test.tsx`
- Removed the global `vi.mock` of `createCheckAuthorization` so the React hook tests exercise the real predicate.

Tests
- New regression matrix in `packages/shared/src/__tests__/authorization.spec.ts` covering every "asked but unsatisfiable" cell, AND semantics within and across dimensions, malformed payloads, null / non-string params, invalid reverification configs, and positive multi-requirement cases.
- SDK-68 regression tests in `packages/nextjs/src/server/__tests__/clerkMiddleware.test.ts`.
- New Playwright coverage in `integration/tests/protect.test.ts` for three new `/settings/*` routes in the existing `next-app-router` template: `auth-protect-mixed-args`, `auth-protect-mixed-token`, `auth-protect-role-and-permission`. Admin / signed-out / viewer assertions added.

Changeset
- `@clerk/shared`, `@clerk/clerk-js`, `@clerk/nextjs` at patch. Other packages pick up the fix transitively via the shared patch.

## Test plan

- [x] `pnpm --filter @clerk/shared test` - 25/25 on the new matrix
- [x] `pnpm --filter @clerk/clerk-js test` - Session.test.ts green
- [x] `pnpm --filter @clerk/clerk-react test` - useAuth.test.tsx green after removing the mock
- [x] `pnpm --filter @clerk/backend test` - green
- [x] `pnpm --filter @clerk/nextjs test` - middleware tests green (pre-existing type-only failures unrelated to this PR)
- [x] `pnpm --filter @clerk/vue test` / `pnpm --filter @clerk/astro test` - green
- [ ] `integration/tests/protect.test.ts` - new admin / viewer / signed-out assertions run in CI against the `withCustomRoles` + `withBillingJwtV2` instances.

## Backport

A Core 2 backport is prepared on the advisory's private fork and will be merged once this lands.


Ephem (approved):
I think the Rabbit had a good NIT to be extra defensive, but other than that I think this is gtg.

PR #8379, opened 2026-04-22
fix(ui): surface silent failures in component renderer

jacekradko:
## Summary

Adds diagnostic coverage to the UI component renderer so silent initialization failures produce actionable errors instead of hanging indefinitely.

- Catches rejections on the lazy module import in `ensureMounted` and logs a `[Clerk UI]` console error with the underlying cause. Currently these rejections disappear because every mount call in `clerk-js/src/core/clerk.ts` uses `void this.#clerkUI.then(...)` without a `.catch`.
- Starts a 10s timeout alongside the lazy import. If the renderer has not mounted by then, logs a `[Clerk UI]` warning naming the common causes (failed chunk load, dev-server misconfiguration, pre-hydration mount). Cleared on both success and failure paths.

Covers the full fan-out with a single change in `ensureMounted` — the ~60 `void this.#clerkUI.then(...)` call sites in `clerk-js` all funnel through this one promise, so there is no per-mount-method surface to touch.

## Context

Surfaced while debugging a regression where `pnpm dev:sandbox` loaded but components never rendered — `Clerk.loaded === true`, `#app` empty, zero console output. The root cause was an unrelated `@rspack/cli` bump enabling lazy-compilation, whose proxy chunk never resolved. With these diagnostics in place, the next person to hit a similar class of failure gets a direct signpost.

## Test plan

- [x] Verified fix renders SignIn normally when the renderer mounts successfully (no regression, no spurious logs).
- [x] Reverted the environmental workaround to reproduce the stalled-mount condition and confirmed the 10s warning fires with the expected message.
- [ ] CI builds and tests pass.


wobsoriano (approved):
👍🏼 👍🏼 👍🏼

PR #8381, opened 2026-04-22
chore(clerk-js,ui): Make OAuthConsent component public

wobsoriano:
## Description

This PR makes `OAuthConsent` publicly available across SDKs and keeps the legacy `__internal_*OAuthConsent` runtime methods in place for compatibility.

Older published SDKs still call:

  - `clerk.__internal_mountOAuthConsent`
  - `clerk.__internal_unmountOAuthConsent`

  When `clerk-js` is loaded from the CDN, users can end up with:
  - older framework SDK versions
  - newer `clerk-js` runtime assets

If we remove the legacy runtime methods now, those mixed-version deployments break at runtime with `... is not a function`.

  To avoid that, this PR:
  - keeps `mountOAuthConsent` / `unmountOAuthConsent` as the canonical API
  - preserves `__internal_mountOAuthConsent` / `__internal_unmountOAuthConsent` as deprecated aliases
  - updates SDKs as needed while maintaining backward compatibility

  ## Follow-up plan

  - SDKs can continue migrating to `mountOAuthConsent` / `unmountOAuthConsent`
  - `__internal_*OAuthConsent` can be removed in a later major once we intentionally drop compatibility with older published wrappers

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/tanstack-react-start/src/internal.ts
wobsoriano, line 1: maybe it's safe just to remove this `/internal` export that was added just for OAuthConsent. Never used in any of our repos

packages/react-router/src/internal.ts
wobsoriano, line 1: ditto https://github.com/clerk/javascript/pull/8381/changes#r3164755377

packages/vue/src/components/ui-components/OAuthConsent.vue
```diff
@@ -0,0 +1,17 @@
...
+import { useClerk } from '../../composables';
+
+const clerk = useClerk();
+const props = defineProps<__internal_OAuthConsentProps>();
+</script>
+
+<template>
+  <ClerkHostRenderer
+    :mount="clerk?.__internal_mountOAuthConsent"
+    :unmount="clerk?.__internal_unmountOAuthConsent"
```
alexcarpenter: Any of these new components should likely be using the `mountOAuthConsent` or `unmountOAuthConsent` vs deprecated `__internal` methods right?

wobsoriano: Oh, good catch 😄

wobsoriano: Updated!

alexcarpenter (approved):

PR #8385, opened 2026-04-23
fix(ui): correct OAuth redirect_url for openSignIn modal

wobsoriano:
## Description

When using `openSignIn()` (modal/virtual routing), clicking an OAuth social button was sending a `redirect_url` pointing to the Clerk accounts portal instead of the app domain.

This PR:

- Replaces `buildSSOCallbackURL` calls in the three components with `ctx.ssoCallbackUrl` directly, consistent with how `SignUpSocialButtons` already works
- Fixes `ssoCallbackUrl` computation in `SignIn.ts`: for virtual routing in non-combined flow, use `signInUrl` (resolved from `options.signInUrl`) as the base instead of `signUpUrl`, which was contaminated by the virtual router's internal `CLERK-ROUTER/VIRTUAL/sign-up` path.

 No behavior change for path or hash routing since `buildRedirectUrl` ignores baseUrl for those modes.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


wobsoriano:
!snapshot

packages/ui/src/components/SignIn/SignInStart.tsx
```diff
@@ -413,7 +412,7 @@ function SignInStartInternal(): JSX.Element {
   };
 
   const authenticateWithEnterpriseSSO = async () => {
-    const redirectUrl = buildSSOCallbackURL(ctx, displayConfig.signInUrl);
+    const redirectUrl = ctx.ssoCallbackUrl;
```
wobsoriano, line 415: Asked Claude why replace the `buildSSOCallbackURL` calls to just `ssoCallbackUrl` and here's the answer:

Image

wobsoriano:
!snapshot

wobsoriano:
!snapshot

jacekradko (commented):
Seems like the correct fix, but want to have @nikosdouvlis chime in here

nikosdouvlis (approved):

PR #8388, opened 2026-04-23
fix(ui): correct OAuth redirect_url for openSignUp modal in combined flow

nikosdouvlis:
## Why

Follow up to #8385. When a customer configures combined flow (CLERK_SIGN_IN_URL set, CLERK_SIGN_UP_URL unset, public signup mode) and calls openSignUp() directly, signUpUrl in SignUp.ts falls back to displayConfig.signUpUrl, which is the accounts portal. Customers that dont use the accounts portal end up with a broken OAuth redirect_url on the wrong origin.

The SignIn side of the same bug was fixed in #8385, but the openSignUp path wasnt touched.

## What changed

Anchor the callback base to options.signInUrl when in a combined-flow modal. The create/sso-callback and create/verify routes are mounted under the SignIn tree (packages/ui/src/components/SignIn/index.tsx), so the redirect resolves against the app origin and lands on LazySignUpSSOCallback.

Path and hash routing are unchanged since buildRedirectUrl ignores baseUrl for those modes, so this is modal-only.

## Test coverage

Added an integration test in integration/tests/oauth-flows.test.ts that exercises openSignUp with the same withSignInOrUpFlow fixture #8385 uses, asserting the redirect_url lands on /sign-in#/create/sso-callback.

## Packages affected

- @clerk/ui: SignUp context anchors callback URLs to options.signInUrl under combined-flow modal mode

### DIFF
--- a/.changeset/fix-opensignup-combined-flow-redirect.md
+++ b/.changeset/fix-opensignup-combined-flow-redirect.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Fixed OAuth `redirect_url` for `openSignUp` modal in combined flow when `CLERK_SIGN_UP_URL` is unset.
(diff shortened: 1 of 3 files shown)


Ephem (approved):
Nice find!

PR #7568, opened 2026-01-08
feat(shared): Remove SWR

jacekradko:
## Description

Removes SWR dependency and fully migrates to TanStack Query (React Query) as the sole data fetching solution. This consolidates our data fetching strategy and eliminates the need for compatibility shims and environment-based switches.

Also includes fixes from https://github.com/clerk/javascript/pull/7544 for the `useClearQueriesOnSignOut` hook.

Fixes: USER-4251

Snapshot deployed to https://github.com/clerk/dashboard/pull/7966

### Changes

- **Dependency migration**: Removed SWR, promoted TanStack Query to a runtime dependency
- **Hook consolidation**: Removed `.swr.tsx` and `.rq.tsx` file variants, consolidated logic into single files
- **Build simplification**: Removed SWR-related build switches and tsconfig virtual paths
- **Bug fixes**: Fixed query-clearing on sign-out by adjusting effect timing and guards
- **New utilities**: Added pagination utility helpers in `usePagesOrInfinite.shared.ts`

### Manual Testing

1. **Sign-out cache clearing**
   - Sign in to an application using Clerk
   - Navigate to pages that load organization/user data (triggering React Query cache)
   - Sign out
   - Sign in as a different user
   - Verify no stale data from the previous user appears

## Acceptance Criteria

- [x] All existing tests pass
- [ ] No SWR references remain in the codebase (excluding lock files)
- [ ] Sign-out properly clears React Query cache with no stale data leaks
- [ ] Pagination hooks (\`useOrganizationList\`, \`useOrganization\`) function correctly
- [ ] Billing/payment hooks initialize and query without errors
- [ ] Bundle size remains within acceptable limits (check bundlewatch)
- [ ] Dashboard snapshot (PR #7966) functions correctly with these changes

## Checklist

- [ ] \`pnpm test\` runs as expected.
- [ ] \`pnpm build\` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [x] 📖 Refactoring / dependency upgrade / documentation

## Summary by CodeRabbit

* **New Features**
  * Consolidated data-fetching on React Query for more consistent billing, pagination, and organization flows; added pagination helper utilities.

* **Bug Fixes**
  * Fixed query cleanup on sign-out and updated hook internals to comply with React Rules of Hooks for more reliable cache clearing.

[...]

PR #8395, opened 2026-04-23
fix(shared): Correct useOrganizationList/useOrganization loading state

jacekradko:
It's been flagged that after upgrading from `@clerk/nextjs` 6.32 to 7.0.12, consumers of `useOrganizationList` briefly see `isLoaded: true` alongside `userMemberships.isLoading: false` and `data: []`  before any fetch actually runs. The documented loading gate (`!isLoaded || userMemberships.isLoading`) can't catch this window, so the UI flashes an empty state on the cold path.

### Root cause

`isLoaded` flips to `true` as soon as `clerk.loaded && user`  that's independent of when the React Query client actually attaches. On slower bootstraps (prod/staging, not local dev) those two signals desync for a render.

In that gap, `useBaseQuery` short-circuits to a dummy result while it waits for the client, and that dummy used to return `isLoading: false` for everyone. `usePagesOrInfinite` then coerces the missing data to `[]`, and the two signals together read as "loaded, no data."

The fix: while the client is still attaching, report `isLoading: true` when the query is enabled. Disabled queries (e.g. `useOrganizationList()` with no opt-in params) keep returning `isLoading: false`, so existing behavior there is preserved. `useOrganization` shares the same hook stack, so it gets the fix for free.

This regressed in #7568 (SWR → React Query). Both `useClerkQuery` and `useClerkInfiniteQuery` route through `useBaseQuery`, so the single change covers both paths.

### Test plan

- [x] New `useBaseQuery.spec.tsx` exercising enabled / disabled / omitted `enabled` in the pre-client window, plus the infinite-query path
- [x] `pnpm --filter=@clerk/shared test` — 1041 tests passing
- [x] `pnpm --filter=@clerk/ui test` — 1677 tests passing (existing `useCoreOrganizationList` fixture tests still green)
- [x] `pnpm --filter=@clerk/shared build` passes

### DIFF
--- a/.changeset/fix-useorganizationlist-empty-render.md
+++ b/.changeset/fix-useorganizationlist-empty-render.md
@@ -2,4 +2,4 @@
 '@clerk/shared': patch
 ---
 
-Fix `useOrganizationList` and `useOrganization` returning `isLoaded: true` with `userMemberships.isLoading: false` and `data: []` during the brief window between clerk-js loading and the React Query client attaching. `useBaseQuery` now reports `isLoading: true` while the query client is not yet loaded, as long as the query is enabled — restoring the documented `!isLoaded || resource.isLoading` loading gate for paginated resources.
+Fix `useOrganizationList` and `useOrganization` briefly reporting paginated resources as `isLoading: false` with empty data before the query starts.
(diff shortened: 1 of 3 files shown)


wobsoriano (approved):
this makes sense

LauraBeatris (approved):

PR #8398, opened 2026-04-24
chore(express,react,shared): Support dynamic options callback in clerkMiddleware

wobsoriano:
## Description

```ts
app.use(clerkMiddleware((req) => ({
  publishableKey: req.hostname === 'domain-a.com' ? PK_A : PK_B,
})));
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


thiskevinwang (approved):

packages/shared/src/keys.ts
```diff
@@ -43,6 +43,32 @@ export function buildPublishableKey(frontendApi: string): string {
...
+ * // Express (inside clerkMiddleware callback)
+ * clerkMiddleware((req) => ({
+ *   publishableKey: publishableKeyFromHost(req.hostname, process.env.CLERK_PUBLISHABLE_KEY),
+ * }))
+ */
+export function publishableKeyFromHost(host: string, fallbackKey?: string): string {
+  if (fallbackKey && isDevelopmentFromPublishableKey(fallbackKey)) {
+    return fallbackKey;
+  }
+  return buildPublishableKey(`clerk.${host.toLowerCase()}`);
```
wobsoriano: existing helper that converts a FAPI URL into a base64-encoded publishable key

https://github.com/clerk/javascript/blob/70bd92ba85539aefb880f33e5fcd8cc1f1199756/packages/shared/src/keys.ts#L38,L44

.changeset/new-kangaroos-search.md
```diff
@@ -0,0 +1,13 @@
...
+
+Support dynamic options callback in `clerkMiddleware`:
+
+
+```ts
+app.use(clerkMiddleware((req) => ({
+ publishableKey: req.hostname === 'example.com' ? PK_A : PK_B,
+ secretKey: req.hostname === 'example.com' ? SK_A : SK_B,
+})));
+```
```
jacekradko: Do we want this in the changeset?  We usually keep those pretty high level

.changeset/brave-lions-fly.md
```diff
@@ -0,0 +1,9 @@
+---
+"@clerk/shared": patch
+"@clerk/react": patch
+"@clerk/express": patch
```
jacekradko: We should split this changeset into 2 separate ones. 1 for BE, 1 for FE, otherwise we are polluting the changelogs with unrelated changes

packages/express/src/internal.ts
```diff
@@ -0,0 +1 @@
+export { publishableKeyFromHost } from '@clerk/shared/keys';
```
jacekradko: 🤔 I am a little hesitant to add a whole new entrypoint to a package in a hotfix.. just seems a bit early for this

wobsoriano: are you saying we should just export from `@clerk/express` or let them import as-is from `@clerk/shared/keys`?

packages/shared/src/keys.ts
```diff
@@ -43,6 +43,32 @@ export function buildPublishableKey(frontendApi: string): string {
...
+ * // React
+ * <ClerkProvider publishableKey={publishableKeyFromHost(window.location.host, import.meta.env.VITE_CLERK_PUBLISHABLE_KEY)}>
+ *
+ * @example
+ * // Express (inside clerkMiddleware callback)
+ * clerkMiddleware((req) => ({
+ *   publishableKey: publishableKeyFromHost(req.hostname, process.env.CLERK_PUBLISHABLE_KEY),
+ * }))
+ */
+export function publishableKeyFromHost(host: string, fallbackKey?: string): string {
```
jacekradko, line 70: Need tests.. fine to do as a follow up

packages/shared/src/keys.ts
```diff
@@ -43,6 +43,32 @@ export function buildPublishableKey(frontendApi: string): string {
...
+ * from the host (e.g. localhost).
+ *
+ * @example
+ * // React
+ * <ClerkProvider publishableKey={publishableKeyFromHost(window.location.host, import.meta.env.VITE_CLERK_PUBLISHABLE_KEY)}>
+ *
+ * @example
+ * // Express (inside clerkMiddleware callback)
+ * clerkMiddleware((req) => ({
+ *   publishableKey: publishableKeyFromHost(req.hostname, process.env.CLERK_PUBLISHABLE_KEY),
```
jacekradko: We should be cautious with this code example. Is it safe to pass `req.hostname` directly here? Can we trust that value from any Express deployment? Perhaps I am being a little too cautious here, but I would hate to show an example that could impact the consuming apps from a security standpoint

packages/shared/src/keys.ts
wobsoriano: tests added

packages/shared/src/keys.ts
jacekradko: I am wondering if we shouldn't be sanitizing/normalizing the hosts that are passed into this helper 🤔

packages/shared/src/keys.ts
wobsoriano: Yeah fair concern. We can drop the example from the docstring or include something like this before the code

```ts
// Ensure your reverse proxy is configured to set X-Forwarded-Host correctly
```

packages/express/src/internal.ts
jacekradko: For now I think I would suggest they import from `@clerk/shared/keys` until we are sure this is something we want to re-export from an SDK package.

packages/shared/src/keys.ts
wobsoriano: yeah I imagine one can include a port and that will product a different key... pushed an update to be more defensive that isnt breaking as well

packages/shared/src/keys.ts
```diff
@@ -43,6 +43,33 @@ export function buildPublishableKey(frontendApi: string): string {
...
+ * from the host (e.g. localhost).
+ *
+ * @example
+ * // React (use window.location.hostname, not window.location.host, to avoid including the port)
+ * <ClerkProvider publishableKey={publishableKeyFromHost(window.location.hostname, import.meta.env.VITE_CLERK_PUBLISHABLE_KEY)}>
+ *
+ * @example
+ * // Express (inside clerkMiddleware callback)
+ * clerkMiddleware((req) => ({
+ *   publishableKey: publishableKeyFromHost(req.hostname, process.env.CLERK_PUBLISHABLE_KEY),
```
nikosdouvlis: Missing the trust-proxy note from the earlier thread. `req.hostname` reads from `X-Forwarded-Host` when `trust proxy` is set, so a spoofed header can swap the PK. Worth a short note above the Express example?

packages/shared/src/keys.ts
```diff
@@ -43,6 +43,33 @@ export function buildPublishableKey(frontendApi: string): string {
...
+ * }))
+ */
+export function publishableKeyFromHost(host: string, fallbackKey?: string): string {
+  if (fallbackKey && isDevelopmentFromPublishableKey(fallbackKey)) {
+    return fallbackKey;
+  }
+  const hostname = host.toLowerCase().replace(/:\d+$/, '');
+  return buildPublishableKey(`clerk.${hostname}`);
```
nikosdouvlis, line 78: What if `host` is empty? `publishableKeyFromHost('')` returns `buildPublishableKey('clerk.')` which is invalid but only fails later when clerk-js parses it. Should we guard at the top? 🤔

nikosdouvlis (approved):

packages/shared/src/keys.ts
wobsoriano: good point, will add a note!

packages/shared/src/keys.ts
wobsoriano: yeah it will fail downstream anyway but doesnt hurt to throw an error early!

wobsoriano: updated

packages/shared/src/keys.ts
wobsoriano: addressed

packages/express/src/internal.ts
wobsoriano: gotcha

PR #8415, opened 2026-04-28
feat(ui): Show OAuthConsent org selector from user:org:read scope

wobsoriano:
## Description

The org selector previously relied on the `__internal_enableOrgSelection` flag. It now also activates when `user:org:read` ([ref](https://github.com/clerk/clerk_go/pull/18352)) is present in the granted scopes, which is the long-term path forward. The flag is kept as a fallback so existing callers aren't broken. The `user:org:read` scope is filtered from the visible scopes list, consistent with how `offline_access` is handled.

will remove `__internal_enableOrgSelection` in a follow up PR once we updated our dashboard consent page which the CLI app uses.

Resolves USER-5239

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


wobsoriano:
!snapshot

packages/ui/src/components/OAuthConsent/OAuthConsent.tsx
```diff
@@ -83,6 +71,19 @@ function _OAuthConsent() {
   const oauthApplicationUrl = ctx.oauthApplicationUrl ?? data?.oauthApplicationUrl;
   const redirectUrl = ctx.redirectUrl ?? getRedirectUriFromSearch();
 
+  const hasOrgReadScope = scopes.some(s => s.scope === USER_ORG_READ_SCOPE);
+  const orgSelectionEnabled = !!((hasOrgReadScope || ctx.enableOrgSelection) && organizationSettings.enabled);
```
wobsoriano, line 75: The new scope takes precedence over `__internal_enableOrgSelection` when present, so we can safely remove the flag in a follow-up PR once confirmed working.

jfoshee (approved):

alexcarpenter (approved):

PR #8424, opened 2026-04-30
experiment: revert Node.js 24 upgrade to test e2e impact

jacekradko:
**Do not merge.** Sandbox to verify whether reverting #8351 (Node.js 24 upgrade) flips the consistently-failing express / hono / generic / nextjs / billing / machine integration suites back to green.

If the integration matrix on this branch goes mostly green compared to recent runs on \`main\`, that confirms #8351 is the regression source. Will be closed regardless of outcome.

Sibling experiment: #8423 (revert PR 8035).
Related: #8422 (the actual fix PR currently open).


jacekradko:
Closing — the same 5 `express` tests (and same hono/billing/etc. patterns) fail identically on this branch as on `main`. reverting #8351 (Node 24 upgrade) does not fix the userButton mount regression. Pattern is consistent across all four candidate branches we tested (#8422, #8423, #8424, #8426), so the cause is environmental, not in the JS source. Refocusing on infra: Clerk dashboard config on the test instances, FAPI deploys, Blacksmith runner image, or pnpm/Playwright resolution drift.

PR #8425, opened 2026-04-30
experiment: revert rspack 1.7.x pin to test e2e impact

jacekradko:
**Do not merge.** Sandbox to verify whether reverting #8378 (pin rspack to 1.7.x and disable lazy compilation) flips the consistently-failing express / hono / generic / nextjs / billing / machine integration suites back to green.

Hypothesis: \`disable lazy compilation\` + the older rspack pin may produce a chunk graph that takes longer to deliver under heavy CI parallelism, contributing to the userButton mount-timeout failures.

Sibling experiments: #8423 (revert PR 8035 auto-proxy), #8424 (revert PR 8351 Node 24).
Related: #8422 (the actual fix PR currently open).

Will be closed regardless of outcome.

PR #8426, opened 2026-04-30
experiment: upgrade to Rspack 2 to test e2e impact

jacekradko:
**Do not merge.** Sandbox to test whether upgrading from Rspack 1.7.x → 2.0.0 fixes the consistently-failing express / hono / generic / nextjs / billing / machine integration suites.

The previous experiment (#8425, now closed) was reverting the rspack pin — but that pin was a fix for an earlier rspack regression and shouldn't be reverted. This branch goes the other direction: **upgrades** to Rspack 2 (the current ecosystem version) on top of latest \`main\` (Node 24, auto-proxy, etc.).

Sibling experiments still up:
- #8423 — revert PR 8035 (auto-proxy): same failures as main
- #8424 — revert PR 8351 (Node 24): same failures as main

Will be closed regardless of outcome.


jacekradko:
Closing — the same 5 `express` tests (and same hono/billing/etc. patterns) fail identically on this branch as on `main`. upgrading to Rspack 2 does not fix the userButton mount regression. Pattern is consistent across all four candidate branches we tested (#8422, #8423, #8424, #8426), so the cause is environmental, not in the JS source. Refocusing on infra: Clerk dashboard config on the test instances, FAPI deploys, Blacksmith runner image, or pnpm/Playwright resolution drift.

PR #8432, opened 2026-04-30
fix(shared,clerk-js): raise @tanstack/query-core floor to ^5.100.6

### LINKED ISSUE #8428: @clerk/nextjs 7.2.8 breaks with e.isFetched' is undefined
After Upgrading 7.2.7 to 7.2.8 our website broke with:

`TypeError: e.isFetched is not a function. (In 'e.isFetched()', 'e.isFetched' is undefined)`

Claude investigated this:

> Root cause: The @clerk/nextjs@7.2.7 → 7.2.8 bump brought in @clerk/shared@4.8.6, which changed its @tanstack/query-core dependency from the exact pin 5.90.16 to the range ^5.90.16. That resolved to 5.100.6, which added isFetched() as a new method on the Query class and has the QueryObserver call query.isFetched().
> 
> Why only production: Clerk's production JS SDK (loaded when __client_uat cookie is present) bundles its own @tanstack/query-core internally (an older version without isFetched()). When @clerk/shared@4.8.6's QueryObserver (from 5.100.6) tries to observe queries from that bundled QueryClient, it calls query.isFetched() on a Query object that doesn't have that method → undefined is not a function.

jacekradko:
## Summary

- Raises the `@tanstack/query-core` floor from `^5.90.16` to `^5.100.6` in `@clerk/shared` and `@clerk/clerk-js`
- Pulls `query-core@5.100.6` into the lockfile so the CDN-published `clerk-js` bundle embeds it (verified: `isFetched` now ships in `query-core-vendors_clerk.browser_*.js`)

## Why

`@clerk/shared` and the production `clerk-js` CDN bundle share an implicit cross-bundle contract: shared constructs `QueryObserver` instances against `Query` objects from clerk-js's `QueryClient`. After #8417 loosened the pin, consumer-side `@clerk/shared` resolved `query-core` to `5.100.x` (which adds `Query.isFetched()` and has the Observer call it), while the CDN bundle still embedded `5.90.16`. The Observer then crashed calling `isFetched()` on the older `Query` objects.

Fixes #8428.

Raising the floor keeps consumer dedupe working with `^5.100.6` (so #8417's Vite/Bun repro should still resolve) and ensures the CDN bundle used for this release embeds the same `query-core` version that triggered the reporter's crash. The class of bug is structural — the cross-bundle contract on `query-core` shouldn't exist — and a follow-up will move `QueryClient` ownership into `@clerk/shared` to eliminate it. This PR is the patch.

## Test plan

- [x] `pnpm --filter @clerk/shared test -- --run` (1069/1069 pass)
- [x] `pnpm --filter @clerk/clerk-js build` succeeds and `query-core-vendors_*.js` chunk contains `isFetched`
- [x] Mixed-version smoke test: `5.100.6` `QueryObserver` + `5.90.16` `QueryClient` reproduces `query.isFetched is not a function`; `5.100.6` observer/client succeeds
- [x] Reproduce 7.2.8 crash with the original repro app, swap to this build, verify resolved
- [ ] Re-run the Lovable-shape repro (Vite + Bun, dedupe to `5.100.x`) — should still resolve cleanly


alexcarpenter (approved):

wobsoriano (approved):

CodingFabian:
I can confirm it appears to be fixed now. thank you

PR #8434, opened 2026-04-30
refactor(shared,clerk-js,react): move QueryClient ownership into @clerk/shared

jacekradko:
## Summary

- Moves the clerk-rq `QueryClient` from `@clerk/clerk-js` into `@clerk/shared`. Both the `QueryObserver` and the `Query` objects it observes now resolve to a single `@tanstack/query-core` instance.
- Removes the undocumented `clerk.__internal_queryClient` getter from both `@clerk/clerk-js` and `@clerk/react`'s `IsomorphicClerk`, plus the dynamic `query-core` import and `queryClientStatus` event.
- Drops `@tanstack/query-core` from `@clerk/clerk-js`'s dependencies (still a dep of `@clerk/shared`).

## Why

#8432 patched the symptom of #8428 by republishing the CDN bundle so its embedded `query-core` matches what consumers' `@clerk/shared` resolves to. That fix is correct but structural — every future `query-core` release that adds an `Observer → Query` method call could re-trigger the same crash class until both sides are forced to share one resolution.

This PR removes the cross-bundle contract entirely. The `QueryClient` is owned by an internal singleton in `@clerk/shared`, lazily instantiated on the browser only. Both the `QueryObserver` (constructor lives in `@clerk/shared`) and the `Query` objects (created via `client.getQueryCache().build(...)`) now come from one resolution — the gap #8428 exploited can no longer exist.

Key insight: clerk-js never *mutated* the QueryClient — every `setQueryData` / `invalidateQueries` / `removeQueries` call already lives in `@clerk/shared` hooks. So no push API back into clerk-js is needed; clerk-js simply stops creating the client.

## Behavior changes

- **`clerk.__internal_queryClient` removed.** Undocumented (leading underscore + `@ts-expect-error` at every call site), but technically observable. No code in this monorepo reaches for it outside the call sites I removed; possible external consumers should now construct their own `QueryClient` if they need one.
- **SSR safety preserved.** `getClerkQueryClient()` returns `undefined` server-side, so per-request renders never share a cache across requests. Hooks fall back to the same proxy mock + synthetic pending state that existed before.
- **`queryClientStatus` event removed.** It was emitted once when clerk-js's lazy `query-core` import resolved; with the singleton synchronously available on the browser there's nothing to signal. No code outside the removed listeners consumed it.

## Test plan

- [x] `pnpm --filter @clerk/shared test` — 1069/1069
- [x] `pnpm --filter @clerk/clerk-js test` — 688/688
- [x] `pnpm --filter @clerk/ui test` — 1681/1681 (incl. SubscriptionDetails / Checkout, the regression-prone billing UI tests)
- [x] `pnpm --filter @clerk/clerk-js build` — clean, no `query-core-vendors_*.js` chunk in dist (confirms the dep was actually dropped)
- [x] `pnpm --filter @clerk/shared build`

Stacked on #8432 — base will be `main` once #8432 lands.


wobsoriano (approved):
Looks good to me. I think `__internal_queryClient` was just internal migration plumbing to avoid eagerly loading query-core unless the RQ path is actually used

PR #8336, opened 2026-04-16
docs(repo): Revise and clarify isLoaded description and tab title

SarahSoutoul:
## Description

Linear ticket: https://linear.app/clerk/issue/DOCS-11007/revise-and-clarify-isloaded-description-and-tab-title 

This PR updates the description of `isLoaded` across hooks (useAuth, useUser, useSession, etc.) to better reflect its actual behavior.

Previously, the docs described `isLoaded` as:

**“Initially false, becomes true once Clerk loads.”**

This wording implies a one-time initialization phase. In practice, `isLoaded` represents whether the current authentication state is resolved, and can temporarily revert to false during auth updates (e.g. when switching organizations via setActive).

Here are the changes I applied:

- Updated isLoaded descriptions to clarify that:
    - It starts as `false` and becomes `true` once loaded BUT may temporarily return to `false` while auth state is updating
    - Replaced “initialization” language with “current authentication state” for better accuracy
    - Added a concrete example (switching organizations) to make this behaviour clearer

## Checklist

- [ ] `pnpm test` runs as expected.
- [ ] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #8435, opened 2026-04-30
docs(repo): Add new tab name for hooks

SarahSoutoul:
## Description

This PR is a follow up of this [`javascript` PR](https://github.com/clerk/javascript/pull/8336). See context [here](https://github.com/clerk/javascript/pull/8336#issuecomment-4334976448).

This PR does a name change from "Initialization" -> "Loading" for the following hooks:

- `useAuth`
- `useUser`
- `useSession`

This tab name change better reflects what's actually happening for these hooks.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


alexisintech (approved):
makes sense!

Ephem (approved):

SarahSoutoul:
@manovotny wanted to get your approval on this, as you reviewed the related PR to this recently?

manovotny:
Pushed a small follow-up directly in 75ea3dd24.

- `packages/shared/src/react/hooks/useUser.ts`: dropped "and initialized" from the docblock so it doesn't contradict the new "Loading" tab heading.

packages/shared/src/react/hooks/useSession.ts
```diff
@@ -10,7 +10,7 @@ const hookName = `useSession`;
  * The `useSession()` hook provides access to the current user's [`Session`](https://clerk.com/docs/reference/objects/session) object, as well as helpers for setting the active session.
  *
  * @unionReturnHeadings
- * ["Initialization", "Signed out", "Signed in"]
+ * ["Loading", "Signed out", "Signed in"]
```
manovotny, line 13: Other hooks with `@unionReturnHeadings` still use `["Initialization", "Loaded"]` — `useSessionList.ts:11`, and the legacy `useSignIn.ts:12` / `useSignUp.ts:12`. 

`useSessionList` has the same `isLoaded` semantics as the three hooks renamed here, so it'd be inconsistent left as-is. 

The legacy sign-in/up ones are arguably different. They describe an in-flight attempt object's lifecycle, not auth state loading. 

Is it worth extending the rename to `useSessionList` too?

SarahSoutoul: @manovotny I only updated the Tabs for the hook that were changed / edited in this [PR](https://github.com/clerk/javascript/pull/8336/changes), and the comment that justified not changing `useSessionList`: 

Screenshot 2026-05-11 at 3 14 31 pm

Happy to do the rename for `useSessionList` but want to clarify this first, as I thought it didn't apply for that hook.

manovotny: @SarahSoutoul you're right. The whole point of the rename is that `isLoaded` isn't a one-time init signal for these three hooks (it flips on `setActive`), and that doesn't apply to `useSessionList` / `useSignIn` / `useSignUp` since `client` survives transitive state changes. "Initialization" should still accurate for those.

manovotny (approved):

PR #8436, opened 2026-04-30
DIAG: instrument userButton mount path (do not merge)

jacekradko:
**Do not merge.** Diagnostic patch to capture exactly what state \`useUserBase\` and \`withCoreUserGuard\` see at the moment the userButton fails to paint.

Adds three \`[CLERK_DIAG]\` console.log lines:
- \`packages/clerk-js/src/core/clerk.ts\` \`#emit\` — every emit, with the user/session/client ids it carries
- \`packages/shared/src/react/hooks/base/useUserBase.ts\` \`getSnapshot\` — every snapshot read, with whether it fell back to initial state and what it returned
- \`packages/ui/src/contexts/CoreUserContext.tsx\` \`withCoreUserGuard\` — every render, with whether the guard passed the user through or returned null

When the express/hono userButton tests fail, the playwright trace now contains the diagnostic logs and we can definitively see whether the issue is:
- (a) clerk-js never emits with a user resource (server response shape problem),
- (b) clerk-js emits but \`__internal_lastEmittedResources\` is somehow not what useSyncExternalStore reads,
- (c) the snapshot returns a user but withCoreUserGuard still hits the null branch (closure/stale-state).

Will be closed after we read the next CI run's traces.


jacekradko:
Closing — the userButton mount regression we set out to instrument resolved itself externally between runs at 17:52 and 18:25 UTC today, with no JS-side code change in that window. That's confirmation the cause was environmental (FAPI deploy / instance config / runner image), not in this repo's source. Diagnostic isn't needed since the symptom is gone.

PR #8437, opened 2026-04-30
fix(expo): prevent session loss on Expo JS reload

souyahia:
## Summary

Fixes a bug where pressing **R** (JS reload) in Expo during development causes the user's session to be lost, forcing them to log in again. Killing and reopening the app preserves the session correctly.

### Root cause

`NativeSessionSync` calls `ClerkExpo.signOut()` whenever `isSignedIn` is falsy, without checking whether Clerk has finished loading:

```ts
const { isSignedIn } = useAuth();

useEffect(() => {
  if (!isSignedIn) {           // true when isSignedIn is `undefined` (loading)
    ClerkExpo?.signOut();       // revokes the session!
  }
}, [isSignedIn, ...]);
```

On a JS reload (pressing R in Expo):
1. All JS module state is reset, but **native modules persist**
2. `NativeSessionSync` mounts, `useAuth()` returns `isSignedIn: undefined` (loading)
3. `!undefined === true`, so it enters the signed-out branch
4. `ClerkExpo.signOut()` is called on the native module, which is still configured from the previous session
5. The native `signOut()` calls `Clerk.shared.auth.signOut(sessionId:)` (revoking the session server-side) and `Clerk.clearAllKeychainItems()` (deleting the JWT from the keychain)
6. When Clerk JS finishes loading, the token is gone and the session is invalid

On a full app restart (kill + reopen), the native module's factory is uninitialized (the process died), so `signOut()` rejects with `E_NOT_INITIALIZED` and the session is preserved correctly.

### Fix

Added an `isLoaded` guard so `ClerkExpo.signOut()` is only called when Clerk has fully loaded and confirmed the user is actually signed out, not during the initial loading phase.

```ts
const { isSignedIn, isLoaded } = useAuth();

useEffect(() => {
  if (!isSignedIn) {
    if (isLoaded) {             // only sign out when we're sure
      ClerkExpo?.signOut();
    }
  }
}, [isSignedIn, isLoaded, ...]);
```


chriscanin (approved):
Hi @souyahia , I have tested this, and it is working well. 
I appreciate you creating this PR and fix, we will get this merged in and published shortly.

chriscanin:
Hi @souyahia , I have mirrored your commit into [#8469](https://github.com/clerk/javascript/pull/8469) with you preserved as the commit author so the integration tests can run (they're gated to non-fork PRs for secret access). Closing this in favor of that one once [#8469](https://github.com/clerk/javascript/pull/8469) is green. I really appreciate you tracking this down and providing a fix for this. I will let you know here when it has been published and is available in a live version, should be some time today or tomorrow morning.

PR #8439, opened 2026-05-01
fix(nuxt): follow Nuxt module conventions for server-side imports

wobsoriano:
## Description

All h3 server runtime symbols (`createError`, `eventHandler`, `setResponseHeader`, etc.) are now imported from `'#imports'` instead of directly from `'h3'`. This matches the convention used by the official [Nuxt module builder](https://github.com/nuxt/module-builder). Nuxt server files are expected to consume auto-import symbols through `'#imports'`, which Nuxt/Nitro resolves at app build time.

`useRuntimeConfig` was already imported this way. This change extends the same pattern to all remaining h3 value imports across `clerkMiddleware.ts`, `utils.ts`, `routeMatcher.ts`, and `webhooks.ts`.

Since `h3` is no longer a direct runtime import, it has been moved from `dependencies` to `devDependencies`. Any Nuxt app already has `h3` transitively via Nuxt itself, so this is non-breaking.

### Type safety

To keep `import { ... } from '#imports'` fully typed during module development (without switching from `tsup`), a local declaration stub at `src/runtime/types/nitro-server.d.ts` re-exports the relevant h3 symbols and is mapped via `paths["#imports"]` in `tsconfig.json`.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


wobsoriano:
!snapshot

packages/nuxt/package.json
```diff
@@ -74,10 +74,10 @@
     "@clerk/shared": "workspace:^",
     "@clerk/vue": "workspace:^",
     "@nuxt/kit": "^4.1.3",
-    "@nuxt/schema": "^4.1.3",
-    "h3": "^1.15.4"
+    "@nuxt/schema": "^4.1.3"
   },
   "devDependencies": {
+    "h3": "^1.15.4",
```
jacekradko, line 80: Codex flagged `h3` type imports that still need to come from direct dependencies. Might want to double check this 

> This package still exposes public declarations that import H3Event/EventHandler from h3 through @clerk/nuxt/server, @clerk/nuxt/webhooks, and helpers like createRouteMatcher. Moving h3 to devDependencies means strict package-manager installs can leave consumers unable to resolve those published declarations from @clerk/nuxt, even if Nuxt has h3 somewhere transitively. The runtime imports can come from #imports, but the package still needs to declare h3 as a dependency or peer dependency for its public type surface.

wobsoriano: valid concern. Tested on a fresh nuxt app with pnpm/npm/yarn using the snapshot - h3 is always resolved transitively via Nuxt so no peer dep prompt or resolution failure.

Moving back to dependencies would imply we ship our own copy, which isn't accurate. Leaving in `devDeps`.

jacekradko (approved):

PR #8445, opened 2026-05-02
refactor(shared): rename clerk-rq directory to query

jacekradko:
## Summary

Renames `packages/shared/src/react/clerk-rq/` → `packages/shared/src/react/query/`. The `rq` shorthand was opaque ("react-query"); `query/` reads naturally and matches the existing function vocabulary (`useClerkQuery`, `getClerkQueryClient`, `useClerkInfiniteQuery`).

Pure directory rename — no behavior change, no public surface change. The `clerk-rq` path was internal to `@clerk/shared` (which itself is described as an internal SDK utility package); not exposed via subpath exports, not referenced in any CHANGELOG or docs.

11 files relocated, 16 import sites updated, two doc comments refreshed. The stable `__tag: 'clerk-rq-client'` runtime identifier in `mocks/clerk.ts` is intentionally untouched to avoid behavior churn.

## Stacking

Based on `jacek/clerk-rq-invert-ownership` (#8434). Will need to be retargeted to `main` once that PR merges, or merged through it.

## Test plan

- [x] `pnpm --filter @clerk/shared test` — 1076 tests pass, no type errors
- [x] `pnpm --filter @clerk/shared build` — clean
- [x] `grep clerk-rq` returns only the deliberate runtime tag and the parent PR's changeset


wobsoriano (approved):
🧹

PR #8468, opened 2026-05-04
chore(ui): Add wizard steps to `<ConfigureSSO />`

LauraBeatris:
## Description

- Introduces Wizard component with support for root and inner steps 
- Introduce basic boilerplate of modules per step
- Add different `Flow.Part` per step type 
- Handles initial loading UI state when querying enterprise connections 

To be handled in next PRs:
- Whether a step is considered "complete" (green mark)

https://github.com/user-attachments/assets/13077d6e-bf4f-4474-933e-57d6b7f465e3

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


LauraBeatris:
I think we should add element IDs for the Wizard primitives, such as the steps - this way developers can target specific parts to style 

I'll either push a change on this PR or raise another one with it, but it's on my list

packages/ui/src/components/ConfigureSSO/steps/ConfigureMapAttributesStep.tsx
```diff
@@ -0,0 +1,23 @@
+import { Flow, Text } from '@/customizables';
+import { useRegisterContinueAction, useWizard } from '@/elements/Wizard';
+
+import { StepLayout } from './StepLayout';
+
+export const ConfigureMapAttributes = (): JSX.Element => {
```
LauraBeatris: Those modules are just initial ones, next PRs will actually implement it + introduce specific stuff such as localization keys

packages/ui/src/components/ConfigureSSO/wizard/ConfigureSSOWizard.tsx
```diff
@@ -0,0 +1,466 @@
...
+        {continueLabelToShow}
+        <Icon
+          icon={CaretRight}
+          size='sm'
+          sx={theme => ({ marginInlineStart: theme.space.$1 })}
+        />
+      </Button>
+    </Flex>
+  );
+};
```
alexcarpenter, line 539: Should we move this to a profile specific element or within the SSO component? It'd be nice to keep the Wizard reusable across differing components. This makes it fairly specific to the profile components.

packages/ui/src/elements/Wizard/Wizard.tsx
```diff
@@ -0,0 +1,466 @@
...
+            key={inner.id}
+            {...(i === 0 ? { index: true } : { path: inner.path })}
+          >
+            <InnerComponent />
+          </Route>
+        );
+      })}
+    </Switch>
+  );
+};
```
alexcarpenter: Should StepRoutes be filtering step.innerSteps through shouldSkip here? Looks like Root does this via getActiveInnerSteps but StepRoutes iterates the raw unfiltered list.

alexcarpenter (commented):
We have a Wizard component within common currently. Obviously this is much more thorough implementation of a wizard, but would be nice to consider replacing in the future. Which is possibly a good reason to generalize the new Wizard component, and not include the custom UI.

packages/ui/src/elements/Wizard/Wizard.tsx
LauraBeatris: I think we're going to remove `shouldSkip` predicate in general, as the verify domain step should always be shown now 

I just had a sync with Nicolas and Tu and we're always going to require the user to verify their email address once they land on the component regardless if it was already verified

LauraBeatris:
> We have a Wizard component within common currently. Obviously this is much more thorough implementation of a wizard, but would be nice to consider replacing in the future. Which is possibly a good reason to generalize the new Wizard component, and not include the custom UI.

@alexcarpenter we're going to keep this new wizard introduced in the PR within `ConfigureSSO` since it's specific to that flow 

In the future, we can rethink on making it more generic and move to `ui/common` as well

LauraBeatris:
After talking with @iagodahlem, I'm going to push some changes to make the Wizard interface more declarative: 

```
<Wizard>
  {!user.verified && (
    <WizardStep>
      <Verify />
    </WizardStep>
  )}
  <WizardStep>
    <ConfigureStep>
      <Wizard>
        <Wizard.Step>
          <SelectConnection />
        </Wizard.Step>
        <Wizard.Step>
          <UpdateConnection />
        </Wizard.Step>
      </Wizard>
    </ConfigureStep>
  </WizardStep>
</Wizard>
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -1457,6 +1457,15 @@ export class Clerk implements ClerkInterface {
       return;
     }
 
+    if (noUserExists(this)) {
```
LauraBeatris, line 1460: Added this one that got missed on the first PR

packages/ui/src/components/ConfigureSSO/ConfigureSSO.tsx
```diff
@@ -89,12 +96,97 @@ const AuthenticatedContent = withCoreUserGuard(() => {
...
+          </ConfigureSSOWizard.Step>
+        </ConfigureSSOWizard>
+      </ConfigureSSOWizard.Step>
+      <ConfigureSSOWizard.Step
+        id='configure'
+        path='configure'
+        label='Configure'
+      >
+        <ConfigureSSOWizard>
+          {/* TODO: Implement configure steps */}
```
iagodahlem, line 163:
```suggestion
```

packages/ui/src/components/ConfigureSSO/ConfigureSSO.tsx
```diff
@@ -89,12 +96,97 @@ const AuthenticatedContent = withCoreUserGuard(() => {
...
+            borderStyle: t.borderStyles.$solid,
+            borderColor: t.colors.$borderAlpha150,
+            marginBlock: '-1px',
+            marginInlineEnd: '-1px',
+            flex: 1,
+          })}
+        >
+          <ConfigureSSOFlowProvider
+            enterpriseConnection={enterpriseConnection}
+            isLoading={isLoadingEnterpriseConnections}
```
iagodahlem, line 118: I think passing the initial `isLoading` down to the context and rendering it on the wizard might lead us to issues when invalidating the query, and also for instance, when we start determining the current step after a refresh I think it would be good to have the loading before the wizard even renders, so what do you think of lifting the initial loading state here to ConfigureSSO root? Something like:

```jsx
const AuthenticatedContent = withCoreUserGuard(() => {
  const { data, isLoading } = __internal_useUserEnterpriseConnections({ enabled: true });
  
  if (isLoading && !data) {
    return <FullSpinner />;
  }
  
  return (
    // …existing layout…
    <ConfigureSSOFlowProvider enterpriseConnection={data?.[0]}>
      <ConfigureSSOSteps />
    </ConfigureSSOFlowProvider>
  );
});
```

Happy to take this as a separate PR if you'd rather keep this one tight.

packages/ui/src/components/ConfigureSSO/steps/TestConfigurationStep.tsx
```diff
@@ -0,0 +1,16 @@
+import { Flow, Text } from '@/customizables';
+
+import { StepLayout } from './StepLayout';
+
+export const TestConfigurationStep = (): JSX.Element => {
+  return (
+    <Flow.Part part='test-sso'>
```
iagodahlem, line 7: I think the convention for `Flow.Part part=` is camelCase across Sign-In / Sign-Up / SessionTasks / UserVerification (`phoneCode2Fa`, `alternativeMethods`, `enterpriseConnections`, `chooseOrganization`, etc.). The four other ConfigureSSO steps already follow that — only `test-sso` here and `sso-confirmation` in `ConfirmationStep` are kebab-case.

iagodahlem (approved):
The declarative shape is great 👏 — I think we still need some improvements to better structure it and separate the wizard responsibilities into more layers, so we can easily extract those components for reuse later, but I think it's good enough as it is for now and we can follow up. Just two small alignment items I'd flag for this PR before the bigger refactor lands as a follow-up:

1. `Flow.Part part=` naming convention.
2. `isLoading` flowing through `ConfigureSSOFlowContext`.

PR #8470, opened 2026-05-04
fix(backend): Fix JWT array audience validation

jescalan:
## Summary

This fixes JWT audience validation for tokens whose `aud` claim is an array.

`verifyJwt` now passes the raw `aud` claim and configured `audience` option into `assertAudienceClaim`. The assertion helper already normalizes string and string-array inputs, so this lets it evaluate array audiences correctly.

## Scope

This does not mean default Clerk session tokens have an `aud` claim. They generally do not, and this PR does not change the existing behavior for tokens with no `aud`: if the token has no audience claim, `verifyJwt` still skips audience validation.

The bug only applies when both of these are true:

- the caller configured `audience`
- the token payload contains `aud` as a string array

That shape is valid JWT syntax and is already represented in this package's M2M token parsing fallback (`aud?: string[]`). Clerk-issued M2M tokens also allow custom claims, and `aud` is not reserved there. So the fix is scoped to correctly enforcing the existing `audience` option when an array `aud` is present, not to introducing audience binding for all Clerk tokens.

## Root Cause

`verifyJwt` wrapped both values before calling the assertion:

```ts
assertAudienceClaim([aud], [audience]);
```

When the token already had `aud` as a string array, the assertion received a nested array. That shape skipped both validation branches: it was not a string, and it was not an array of strings. The function returned without enforcing the configured audience.

## Tests

- Added assertion coverage for two string arrays with no intersection.
- Added verifier coverage using signed JWTs:
  - accepts an array `aud` when it includes the configured audience
  - rejects an array `aud` when it does not include the configured audience

## Validation

- `NODE_OPTIONS=--no-experimental-webstorage pnpm --filter @clerk/backend build`
- `pnpm --filter @clerk/backend build:runtime`
- `NODE_OPTIONS=--no-experimental-webstorage pnpm exec vitest run src/jwt/__tests__/assertions.test.ts src/jwt/__tests__/verifyJwt.test.ts --environment node --typecheck.enabled=false`
- `pnpm --filter @clerk/backend format:check`
- `git diff --check`


dominic-clerk (commented):
This could be for a follow-up but I think we could change the types in https://github.com/clerk/javascript/blob/f2af9ba6b174a564993731eb2c3cf55ef9c8bd13/packages/backend/src/jwt/assertions.ts#L10  from unknown to `string | string[]`

Also I understand skipping verification if we have no expected audience, but it seems odd to skip it if we do expect something and the token doesn't provide any. https://github.com/clerk/javascript/blob/f2af9ba6b174a564993731eb2c3cf55ef9c8bd13/packages/backend/src/jwt/assertions.ts#L13

wobsoriano (approved):
@dominic-clerk approving as this fixes array `aud` handling and adds targeted coverage. 

The type cleanup and missing `aud` behavior question can be a follow-up, right?

dominic-clerk:
@wobsoriano yes it can be a follow-up, what would be the best way to ensure it doesn't fall through cracks? A ticket in SDK team?

wobsoriano:
> @wobsoriano yes it can be a follow-up, what would be the best way to ensure it doesn't fall through cracks? A ticket in SDK team?

Yes, SDK team.

Created SDK-89.

PR #8474, opened 2026-05-05
feat(headless): add package foundation with Dialog primitive

alexcarpenter:
## Summary

Introduces the `@clerk/headless` package — a zero-style React component library providing accessible headless UI primitives. This is the first in a series of stacked PRs.

This PR establishes all core infrastructure and patterns:

- **Package scaffold**: Vite lib build with per-primitive entry points, Vitest with happy-dom, TypeScript config
- **Core utils**: `renderElement` (polymorphic rendering with render prop support, `data-cl-*` state attributes), `mergeProps` (event handler chaining, style/className merging)
- **Hooks**: `useControllableState`, `useTransitionStatus`, `useAnimationsFinished`, `useTransition` — the full controlled/uncontrolled + enter/exit animation lifecycle
- **First primitive**: `Dialog` — modal with focus trapping, scroll lock, scoped portal support, compound component API

## Review Stack

> **Review order** — start at the bottom and work up. Each PR builds on the one below it.

| # | PR | Primitive | Base branch |
|---|---|---|---|
| 9 | #8764 | Collapsible | ← autocomplete |
| 8 | #8481 | Autocomplete + floating-tree integration | ← menu |
| 7 | #8480 | Menu | ← select |
| 6 | #8479 | Select | ← popover |
| 5 | #8478 | Popover | ← tooltip |
| 4 | #8477 | cssVars + Tooltip | ← tabs |
| 3 | #8476 | Tabs | ← accordion |
| 2 | #8475 | Accordion | ← foundation |
| **1** | **#8474** | **Foundation** (infra + hooks + utils + Dialog) | ← main |

## Test plan

- [x] `pnpm build` succeeds with clean types
- [x] `pnpm test` — 72 tests pass (hooks, utils, Dialog)
- [ ] CI passes


alexcarpenter:
## 📋 PR Stack Overview

This is a stacked PR series breaking the `@clerk/headless` package into individually reviewable pieces. **Start reviewing here** — each subsequent PR adds exactly one primitive.

### Review order

| # | PR | Adds | Lines (src + test) |
|---|---|---|---|
| **→ 1** | **#8474** | **Foundation**: package scaffold, `renderElement`/`mergeProps`, all 4 hooks, Dialog | ~1200 + 1050 |
| 2 | #8475 | Accordion | ~350 + 420 |
| 3 | #8476 | Tabs | ~450 + 460 |
| 4 | #8477 | `cssVars` middleware + Tooltip | ~410 + 580 |
| 5 | #8478 | Popover | ~420 + 260 |
| 6 | #8479 | Select | ~625 + 700 |
| 7 | #8480 | Menu (submenus, safePolygon) | ~530 + 670 |
| 8 | #8481 | Autocomplete + cross-primitive integration tests | ~585 + 1390 |

### Key decisions to validate in this PR

1. **`renderElement` + `mergeProps`** — the rendering abstraction every primitive uses
2. **`useTransition` stack** — the enter/exit animation lifecycle (state machine + Web Animations API hook)
3. **Compound component pattern** — `Object.assign(Dialog, { Trigger, Popup, ... })`
4. **`data-cl-*` attributes** — sole styling hook, no class names emitted
5. **happy-dom** for tests (matches CI, no Playwright needed)

Once the patterns here look good, the remaining PRs are mechanical — just "one more primitive using the same patterns."

packages/headless/src/primitives/dialog/dialog.tsx
```diff
@@ -0,0 +1,377 @@
...
+  });
+
+  // When scoped to a container (Dialog.Portal has a root), skip FloatingOverlay
+  // which uses position:fixed. CSS on the backdrop element handles positioning.
+  if (scoped) {
+    return backdropElement;
+  }
+
+  return <FloatingOverlay lockScroll={lockScroll}>{backdropElement}</FloatingOverlay>;
+}
```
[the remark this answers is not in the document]
alexcarpenter: solved here https://github.com/clerk/javascript/pull/8474/commits/26a1fec0fc91a826632b8653f6a3639bed735d15

packages/headless/src/utils/render-element.tsx
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  const computedProps = { ...props, ...dataAttrs };
+
+  if (typeof render === 'function') {
+    return render(computedProps as React.HTMLAttributes<HTMLElement>);
+  }
+
+  if (render) {
+    return React.cloneElement(render, computedProps);
```
Ephem: `cloneElement` is pretty legacy by now and is [discouraged by the React docs](https://react.dev/reference/react/cloneElement).

I think it has some very real downsides, that I don't think are made up for by the DX of being able to `render={<Tag />}`. If we want to support simple tag switching without the full render prop pattern, I think `as={Tag}` or even `render={Tag}` is enough, and if you also want custom props in there, you use the explicit render prop pattern `render={(props) => <Tag customProp='value' {...props} />}`?

Subjective and not a "I will die on this hill" stance, I just happen to have an aversion to `cloneElement`. :grin:

packages/headless/package.json
```diff
@@ -0,0 +1,45 @@
+{
```
Ephem, line 1: Should this have `"sideEffects": false`?

packages/headless/README.md
```diff
@@ -0,0 +1,64 @@
...
+    </Select.Popup>
+  </Select.Positioner>
+</Select>;
+```
+
+All primitives follow the same compound component pattern. They emit zero styles — all visual styling is applied externally via `data-cl-*` attribute selectors.
+
+## Architecture
+
+- **Compound components** via `Object.assign` — each primitive is a single export with dot-accessed parts (`Select.Trigger`, `Select.Popup`, etc.)
```
Ephem: I think this approach entirely disables tree-shaking? Maybe not a huge deal but since we are setting this up from scratch and now is the best time to fix it if we want to, I wanted to raise it.

If we still want compound components, I think `export * as Dialog from './dialog';` achieves mostly the same but would tree shake if we set `sideEffects: false`?

packages/headless/README.md
```diff
@@ -0,0 +1,64 @@
...
+- **`renderElement`** — every part uses this instead of returning JSX directly, enabling consumer `render` prop overrides and automatic state-to-data-attribute mapping
+- **`data-cl-*` attributes** — structural (`data-cl-slot`), state (`data-cl-open`, `data-cl-selected`, `data-cl-active`), and animation lifecycle (`data-cl-starting-style`, `data-cl-ending-style`)
```
Ephem, line 52: I love the `data-cl-*` attributes. :heart:

Did you consider also namespacing these per primitive? Maybe that's not necessary?

packages/headless/src/utils/render-element.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+type PropsInput<Tag extends keyof React.JSX.IntrinsicElements> =
+  | React.ComponentPropsWithRef<Tag>
+  | Record<string, unknown>;
+
+/**
+ * Merges two props objects. Event handlers are chained (both fire),
+ * `style` is shallow-merged, `className` is concatenated, and
+ * everything else is overwritten by the second argument.
+ */
```
Ephem, line 39: This is one great comment, and a solid approach!

packages/headless/README.md
```diff
@@ -0,0 +1,64 @@
+# @clerk/headless
+
+Headless UI primitives for Clerk's component library. These are unstyled, accessible React components built on [Floating UI](https://floating-ui.com/) that handle positioning, keyboard navigation, focus management, and ARIA attributes.
+
+This package is **internal** (`private: true`) and consumed by `@clerk/ui`. It exists as a separate package because `@clerk/ui` uses `@emotion/react` as its JSX source, which conflicts with the standard `react-jsx` transform these primitives require.
```
Ephem, line 5: Not sure if this would help, or if we even _want_ to have `/headless` as a folder in `ui` instead, but did you try adding a `/ui/src/headless/tsconfig.json` with `"jsx": "react-jsx"`? I think most tools resolve the nested one?

Ephem, line 5: Is this intended to be private forever? I'm guessing yes and that anything we want public we export through `ui` right?

I'm trying to think through the public parts, the css `data-cl-*` attributes is an obvious one, the rest is probably only exposed through other components right?

Are there any cases for our public components where you think C1s will pass through render props, or any other decisions here that are going to shape public facing APIs down the line?

Ephem (commented):
This is very neat! Just dropping off some partway review comments, mostly me asking questions to try to wrap my head around stuff. Will continue reviewing tomorrow.

I did skim the animation parts as well, they are looking great!

packages/headless/src/utils/render-element.tsx
alexcarpenter: removed https://github.com/clerk/javascript/pull/8474/commits/82bd55b77de21d51b03a42cca5e7259e2af98b9a

packages/headless/package.json
alexcarpenter: https://github.com/clerk/javascript/pull/8474/commits/9b6aaa5d52c2df5add84f110b3372273df1b4e0d

packages/headless/README.md
alexcarpenter: I think to achieve full tree-shaking we'd need to move all subcomponents into their own files and export similar to base-ui is doing with parts exports. https://github.com/mui/base-ui/tree/master/packages/react/src/avatar

Since this is not a public library this seemed unnecessary. but happy to reconsider if you think we should. lmk what you think!

packages/headless/README.md
alexcarpenter: what do you mean per primitive?

packages/headless/README.md
alexcarpenter: I am open to moving into the ui package. I think its been helpful to isolate for the time being while reviewing to avoid any conflicts, but once we feel like were in a good spot, I can move it within. thoughts?

alexcarpenter: so the data attributes are to be used for c1s to be able to tie into and style with. the custom rendering is internal only atm. Possibly in the future that could be a mechanism we expose for custom rendering.

packages/headless/README.md
Ephem: At the time, I was thinking how a natural step would be: https://github.com/clerk/javascript/pull/8654

When we want to export subcomponents like that, maybe we do want the underlying primitives to treeshake as well?

> I think to achieve full tree-shaking we'd need to move all subcomponents into their own files and export similar to base-ui is doing with parts exports

Yes, that's exactly what I was thinking. It's annoying though so there's a real tradeoff not being able to colocate.

packages/headless/README.md
Ephem: `data-cl-select-open` etc to allow for more narrow targeting when writing CSS?

packages/headless/README.md
Ephem: Ohh yeah, I don't really have a strong opinion on where it should live, I just remember seeing your struggles with getting that to work and thought I'd share this approach if it was helpful. :) Definitely keep it as a separate package for now!

packages/headless/README.md
alexcarpenter: Trying to keep the list lean. My recommendation would be the compose the selectors together

```css
[data-cl-slot="select-root"][data-cl-open] { ... }
```

packages/headless/src/primitives/dialog/dialog.tsx
```diff
@@ -0,0 +1,381 @@
...
+      modal={modal}
+    >
+      {renderElement({
+        defaultTagName: 'div',
+        render,
+        props: mergeProps<'div'>(defaultProps, otherProps),
+      })}
+    </FloatingFocusManager>
+  );
+}
```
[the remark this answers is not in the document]
alexcarpenter: https://github.com/clerk/javascript/pull/8474/commits/2b2b8189c7d5a0b81f6c07d31b48a20112b28664

packages/headless/README.md
alexcarpenter: took the base ui approach here https://github.com/clerk/javascript/pull/8474/commits/54016b4f021f5b741f8f7ee2ea8629ae3e9e339d

packages/headless/README.md
Ephem: True that! Works for me. 😄

packages/headless/README.md
Ephem: Nice! It's a bit annoying having to split the files to support this but I think it might set us up for success down the line. I still don't have strong feelings though if you dislike it. 😄

alexcarpenter: wasn't really a problem with ai to refactor it, so I don't think its an issue in practice if we get better results 👍🏼

.changeset/dry-yaks-rats.md
```diff
@@ -0,0 +1,4 @@
+---
+---
+
+Introduce `@clerk/headless` package with Dialog primitive, hooks, and utilities.
```
Ephem: Headless package isn't public, so I don't think it needs a changeset? This will make people think they can use it directly. :smile:

packages/headless/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "exports": {
+    "./dialog": {
+      "import": "./dist/primitives/dialog/index.js",
+      "types": "./dist/primitives/dialog/index.d.ts"
+    },
+    "./utils": {
+      "import": "./dist/utils/index.js",
+      "types": "./dist/utils/index.d.ts"
+    }
+  },
```
Ephem, line 20: This is missing the /hooks

packages/headless/src/primitives/dialog/dialog-title.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useId, useLayoutEffect } from 'react';
+import { useDialogContext } from './dialog-context';
+import { type ComponentProps, mergeProps, renderElement } from '../../utils/render-element';
+
+export interface DialogTitleProps extends ComponentProps<'h2'> {}
+
+export function DialogTitle(props: DialogTitleProps) {
+  const { render, ...otherProps } = props;
+  const { setLabelId } = useDialogContext();
+  const id = useId();
```
Ephem: I think this id can get overridden by external props, in which case I think things can break.

Should it be `const id = props.id ?? useId();`?

I think this is the case in more places too.

packages/headless/src/primitives/dialog/dialog-title.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  useLayoutEffect(() => {
+    setLabelId(id);
+    return () => setLabelId(undefined);
+  }, [id, setLabelId]);
```
Ephem: This pattern has a few issues, the main one being it doesn't work well with SSR (which might or might not matter that much for dialogs, but we are establishing patterns here).

I get that this doubles as a "registration" type thing, to ensure we don't set `aria-labelledby` if a `<Dialog.Title>` was not rendered, but still feels off to me. I think Radix, MUI and React Aria all hoist the id to the root instead and just expect Title to be there, or else you need to pass in `aria-labelledby` explicitly?

I think screen readers handle `aria-labelledby` pointing to non-existing IDs quite well, so I think I'd prefer to:

* Exclude `id` from the props for Title
* Hoist `useId` / passing in manual IDs to the root component - passed down via context
* Remove layout effect

What are your thoughts?

packages/headless/src/utils/render-element.tsx
```diff
@@ -0,0 +1,156 @@
...
+      const bHandler = bVal as (...args: unknown[]) => void;
+      merged[key] = (...args: unknown[]) => {
+        aHandler(...args);
+        bHandler(...args);
+      };
+    } else if (key === 'style' && a.style && bVal) {
+      merged.style = { ...(a.style as object), ...(bVal as object) };
+    } else if (key === 'className' && a.className && bVal) {
+      merged.className = `${a.className} ${bVal}`;
+    } else {
```
Ephem: I'm getting a `no-base-to-string` lint error here because `a.className` and `bVal` are both `unknown`.

Changing the condition to this convinces the linter: `else if (key === 'className' && typeof a.className === 'string' && typeof bVal === 'string')`

packages/headless/src/utils/render-element.tsx
```diff
@@ -0,0 +1,156 @@
...
+  Tag extends keyof React.JSX.IntrinsicElements,
+  State extends Record<string, unknown> = Record<string, never>,
+>(params: RenderElementParamsConditional<Tag, State>): React.ReactElement | null;
+export function renderElement<
+  Tag extends keyof React.JSX.IntrinsicElements,
+  State extends Record<string, unknown> = Record<string, never>,
+>(params: RenderElementParamsBase<Tag, State> & { enabled?: boolean }): React.ReactElement | null {
+  const { defaultTagName, render, enabled = true, state, stateAttributesMapping, props } = params;
+
+  if (!enabled) return null;
```
Ephem: NIT: Not sure why this doesn't show up in CI, but I'm getting `Expected { after 'if' condition. eslint(curly)` here

Update: I'm getting a _bunch_ of lint errors in other files as well. Not sure if it's my IDE being in a weird state or yours?

packages/headless/src/utils/render-element.tsx
```diff
@@ -0,0 +1,156 @@
...
+ */
+export function renderElement<
+  Tag extends keyof React.JSX.IntrinsicElements,
+  State extends Record<string, unknown> = Record<string, never>,
+>(params: RenderElementParamsAlways<Tag, State>): React.ReactElement;
+export function renderElement<
+  Tag extends keyof React.JSX.IntrinsicElements,
+  State extends Record<string, unknown> = Record<string, never>,
+>(params: RenderElementParamsConditional<Tag, State>): React.ReactElement | null;
+export function renderElement<
```
Ephem, line 127: This takes input and returns jsx so it could theoretically be a React component, was that something you considered? I don't really have an opinion, just being curious.

No huge upsides, but it would grant the possibility to use hooks, it would show up in devtools (for good and bad, likely mostly bad) and possibly the React compiler could auto-memoize then. It's an internal detail that's easy to refactor later so no big deal either way, but thought I'd ask. :smile:

packages/headless/src/primitives/dialog/dialog-popup.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <FloatingFocusManager
+      context={floatingContext}
+      modal={modal}
+    >
+      {renderElement({
+        defaultTagName: 'div',
+        render,
+        props: mergeProps<'div'>(defaultProps, otherProps),
```
Ephem, line 42: Passing in `props.ref` from the outside will override the `combinedRef` that was created above and break things.

packages/headless/README.md
Ephem: NIT: Remove `Object.assign` from README now that we are not using that approach.

packages/headless/src/hooks/use-controllable-state.ts
```diff
@@ -0,0 +1,28 @@
...
+  const [uncontrolled, setUncontrolled] = useState(defaultValue);
+  const value = controlled !== undefined ? controlled : uncontrolled;
+  // eslint-disable-next-line react-hooks/exhaustive-deps -- intentionally keyed on `controlled !== undefined` to avoid re-creating when the controlled value changes
+  const setValue = useCallback(
+    (next: T) => {
+      if (controlled === undefined) setUncontrolled(next);
+      onChange?.(next);
+    },
+    [controlled !== undefined, onChange],
+  );
```
Ephem: The disable comment ended up in the wrong place here, but there's a better way to handle this:

```suggestion
  const isControlledDefined = controlled === undefined;
  const setValue = useCallback(
    (next: T) => {
      if (isControlledDefined) setUncontrolled(next);
      onChange?.(next);
    },
    [isControlledDefined, onChange],
  );
```

Ephem (approved):
This time I did a full pass and I think it's looking great! I left a bunch of comments, some like the changeset, package setup and lint errors I think we might want to fix/figure out before a merge, but most things we can iterate on later.

Seeing how this is not a package in public use yet I'm giving an ahead of time approval, feel free to resolve as you see fit. Ping me if you want me to do another pass on any changes. 😄

Nice work!

.changeset/dry-yaks-rats.md
alexcarpenter: Done in https://github.com/clerk/javascript/pull/8474/commits/40e78a900 — emptied the body, leaving the `---` delimiters only.

packages/headless/package.json
alexcarpenter: Done in https://github.com/clerk/javascript/pull/8474/commits/e21a76e26 — added `./hooks` to the exports map (backed by a new `src/hooks/index.ts` barrel + matching vite entry).

packages/headless/src/primitives/dialog/dialog-title.tsx
alexcarpenter: Resolved in https://github.com/clerk/javascript/pull/8474/commits/9dbdee9e4 — went with the hoist-to-Root approach from your follow-up comment rather than the `props.id ?? useId()` workaround, so `id` is no longer in `DialogTitle`/`DialogDescription` props at all.

packages/headless/src/primitives/dialog/dialog-title.tsx
alexcarpenter: Done exactly as you described in https://github.com/clerk/javascript/pull/8474/commits/9dbdee9e4: `labelId`/`descriptionId` are now generated once in `Dialog.Root` via `useId` and passed through context as plain `string`. Removed `useLayoutEffect`, `setLabelId`/`setDescriptionId` setters, and the `id` prop on `Title`/`Description`. Matches Radix/React Aria/MUI.

packages/headless/src/utils/render-element.tsx
alexcarpenter: Applied your suggestion in https://github.com/clerk/javascript/pull/8474/commits/b8ec122ce — narrowed both `a.className` and `bVal` with `typeof === "string"`.

packages/headless/src/utils/render-element.tsx
alexcarpenter: Root cause: this package had no `lint` script, so Turbo silently skipped it. Wired the script up in https://github.com/clerk/javascript/pull/8474/commits/e21a76e26 and fixed the remaining errors in https://github.com/clerk/javascript/pull/8474/commits/75b95bb3f. `pnpm turbo lint --filter @clerk/headless` is now clean.

packages/headless/src/utils/render-element.tsx
alexcarpenter: Considered it but going to leave it as a plain function for now — the parts already wrap it via their own components, so making `renderElement` a component would add a layer in devtools without unlocking new behavior. Happy to revisit if we ever want hooks inside it.

packages/headless/src/primitives/dialog/dialog-popup.tsx
alexcarpenter: Fixed in https://github.com/clerk/javascript/pull/8474/commits/2f4d10174 — pulled `ref` out of the consumer props on both `Dialog.Popup` and `Dialog.Trigger` (same bug there) and fed it into `useMergeRefs` alongside the internal refs so it merges instead of clobbering.

packages/headless/README.md
alexcarpenter: Done in https://github.com/clerk/javascript/pull/8474/commits/7c130affa.

packages/headless/src/hooks/use-controllable-state.ts
alexcarpenter: Applied your suggestion in https://github.com/clerk/javascript/pull/8474/commits/b0f0911a1 — hoisted into `const isControlled = controlled !== undefined`, used in both the callback and the dep array, dropped the disable comment.

packages/headless/src/utils/render-element.tsx
Ephem: Sounds good to me, I think that's the right move! I think making it a component likely makes the types a bit of a pain too. 👌

PR #8487, opened 2026-05-06
fix(nextjs): Improve auth() misimport from main entry point

alexcarpenter:
## Summary
Improved developer experience when `auth` is accidentally imported from `@clerk/nextjs` instead of `@clerk/nextjs/server`. The `auth` export now includes a `@deprecated` JSDoc tag with a clear message guiding developers to the correct import path, visible on hover in IDEs.

Ref: SDK-65

## Test plan
- [x] Added `auth-misimport.test.ts` covering both function call and property access scenarios
- [x] Existing server exports snapshot test still passes
- [ ] Verify in a Next.js app that `import { auth } from '@clerk/nextjs'` shows the new error instead of the bundler error


alexcarpenter:
!snapshot

packages/nextjs/src/index.ts
```diff
@@ -96,4 +96,4 @@ export const Show = ComponentsModule.Show as ServerComponentsServerModuleTypes['
  * The `auth` function is only available in server-side contexts:
  * API Routes, Server Components, Server Actions, and Middleware.
  */
-export declare const auth: never;
+export { auth } from './server-only-stubs';
```
Ephem: Does this approach move the error from being a build time one to being a runtime one? I'm not sure we should do that, I'd rather take a less specific error at build time than a great one at runtime or else users risk deploying bugs to production before catching the error?

I didn't think much about feasibility or where the current message was coming from when I wrote the issue. :/

alexcarpenter: yeah, I agree the tradeoff probably isn't worth it. losing the build-time check to improve a runtime error risks letting misimports ship to production for that subset of users.

wdyt about adding a @deprecated jsdoc notice? https://github.com/clerk/javascript/pull/8487/commits/140d3345c9f0c9db326e96b5d6142159ee09b0cb

Ephem: That's clever and likely the best we can do, I like it!

Ephem (approved):
I like it! 🙏 

NIT: Maybe update(/remove) the PR description before merging to avoid someone being confused in the future.

jacekradko:
FYI: I considered using the `@deprecated` tag, but I think it's misleading as it's not actually deprecated here: https://github.com/clerk/javascript/pull/8358

PR #8493, opened 2026-05-06
chore(ui): Layer architecture for `<ConfigureSSO />`

iagodahlem:
## Summary

Refactors the `<__experimental_ConfigureSSO />` UI surface into a layered architecture that separates the navigation primitive (Wizard) from the layout primitives (Step, Stepper, ProfileCard) from the per-step content. No public component API change — the `<__experimental_ConfigureSSO />` export keeps its current shape.

## Why

The wizard primitive that landed in #8468 ships a working declarative API plus an imperative `useRegisterContinueAction` registry, but it owns its own chrome (Header, Footer, Body), couples step gating to a wizard-wide `FooterActionsContext`, and runs URL routing through `Switch` / `Route`. That made the breadcrumb non-reusable, blocked per-step footer customization, and required a `deepestWizard` workaround to dispatch correctly across nested wizards.

This PR reshapes the surface so each layer has a single job:

- **Wizard** — pure state machine. No chrome, no `Switch` / `Route`, no URL coupling. Steps register themselves via `useLayoutEffect` and render self when active. `goNext` / `goPrev` bubble through nested wizards naturally.
- **Stepper** — purely presentational step indicator. Each `<Stepper.Item>` is a self-rendering button + bullet + label. `<Stepper.Skeleton>` mirrors it for the loading state.
- **Step** — compound layout primitives (`Step.Header`, `Step.Body`, `Step.Section`, `Step.Counter`, `Step.Footer` with `Footer.Previous` / `Footer.Continue`). All UI-only, no wizard coupling — the same primitives drive the live shell and the skeleton.
- **ProfileCard** — semantic `<header>` / `<footer>` chrome with the consistent border + padding the surface uses.

The host (`ConfigureSSO.tsx`) wires the layers together. Each step file is short, opinionated about its own behavior, and assembles the primitives from above.

## What changed

New primitives at `packages/ui/src/components/ConfigureSSO/elements/`:

- `Wizard/` — state-driven, children-rendering wizard with an optional `initialStepId` + `Wizard.Step` registration via `useLayoutEffect`. Drops the previous routing layer entirely.
- `Stepper/` — declarative `<Stepper>` + `<Stepper.Item>` + `<Stepper.Skeleton>`. The container only handles flex layout + separator insertion between items; current/completed/reachable computation moves to the host.
- `Step.tsx` — the `<Step>` layout wrapper plus the compound members listed above. The Footer button compounds (`Step.Footer.Previous`, `Step.Footer.Continue`) take their own `onClick` / `isDisabled` / `isLoading` props — no `useWizard` inside, which is why the same Footer primitives work in `ConfigureSSOSkeleton`.
- `ProfileCard.tsx` — header / footer chrome.

Surface-level wiring:

- `ConfigureSSO.tsx` adopts the new shape end-to-end. The inline ProfileCard + NavBar layout collapses into composed primitives. The wizard's old internal chrome is gone.
- `ConfigureSSOHeader` reads `useWizard()` and maps the active steps into `<Stepper.Item>`s; it also filters `select-provider` out of the breadcrumb by id.
- Each step file destructures `useWizard()` directly, owns its own `<Step.Footer>` with explicit `goPrev` / `goNext` handlers, and renders its body via `Step.Body` + `Step.Section`.
- `SelectProviderStep` boilerplate is added as the first wizard step. It's filtered out of the breadcrumb in `ConfigureSSOHeader` by id (no `hidden` flag on the wizard primitive — kept the primitive clean). Real provider-tile UI lands separately.
- The previous `FooterActionsContext` registry (push / pop / `deepestWizard` plumbing) is gone — per-step footer rendering removes the need for it.

## Out of scope — planned follow-ups

- Real Select Provider tile UI + `FlowProvider.provider` state
- `deriveInitialStep(connection)` and wiring to `Wizard.initialStepId`
- Per-step query lifecycle (Test step's runs, Configure step's POST result handling)
- Reset connection + Configure-again flows
- Element-descriptor sweep — `// TODO: add descriptor` comments mark the placements
- Localization keys for `Previous` / `Continue` / `Step X / Y`

## Test plan

- [ ] Sandbox renders `<__experimental_ConfigureSSO />` end-to-end with the placeholder content
- [ ] Breadcrumb shows 4 visible items (Verify domain / Configure / Test / Confirmation). Select Provider doesn't appear in the breadcrumb at any time
- [ ] When on Select Provider, no breadcrumb bullet is rendered as current. Previous is disabled, Continue is enabled
- [ ] When on Verify Domain with the ProvideEmail conditional sub-step hidden (user has a verified primary email), `Step.Counter` doesn't render (single inner step → badge auto-hides)
- [ ] Continue / Previous navigate within nested wizards, then bubble to the outer wizard at the inner-last-step boundary
- [ ] Initial loading shows `ConfigureSSOSkeleton` once on cold start; subsequent re-fetches don't re-trigger the full skeleton

## Type of change

- [x] 📖 Refactoring


packages/ui/src/components/ConfigureSSO/elements/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { CaretRight } from '@/icons';
+
+import type { StepperItemProps, StepperProps } from './types';
+
+const Root = ({ children }: StepperProps): JSX.Element => {
+  const items = React.Children.toArray(children).filter(child => React.isValidElement(child));
+
+  return (
+    <Flex
+      // TODO: add descriptor
```
LauraBeatris: Do you think it's feasible to add descriptors on this PR? or on a follow up?

packages/ui/src/components/ConfigureSSO/elements/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,163 @@
...
+    <Button
+      // TODO: add descriptor
+      isActive={isCurrent}
+      variant='unstyled'
+      isDisabled={!isReachable}
+      onClick={onClick}
+      sx={theme => ({
+        gap: theme.space.$1x5,
+        padding: 0,
+        color: isCurrent || isCompleted ? theme.colors.$colorForeground : theme.colors.$colorMutedForeground,
```
LauraBeatris, line 56: Should we already introduce the "green mark" style for the step item here, if completed? Also can do in a follow up

LauraBeatris (approved):

packages/ui/src/components/ConfigureSSO/elements/Stepper/Stepper.tsx
iagodahlem: Done here: https://github.com/clerk/javascript/pull/8493/commits/7bd72fe43a86d363eb567699c0a88f67bd85f22a

packages/ui/src/components/ConfigureSSO/elements/Stepper/Stepper.tsx
iagodahlem: Oh, that's actually already in place but it is defined in another layer: https://github.com/clerk/javascript/pull/8493/changes#diff-600fd890fab67d8574cbd047af3e1b1444eeda5aa871469eca2c82e5ecbb3eadR31 I'll refactor it to make more clear.

PR #8494, opened 2026-05-06
fix(clerk-js): reset Core 3 OAuth retry state

jacekradko:
Fixes [SDK-75](https://linear.app/clerk/issue/SDK-75/oauth-sign-in-stuck-in-pending-state-after-browser-back-button-core-3).

`signIn.__internal_future.sso()` reused an existing sign-in resource any time one had an id. After a user backed out of an OAuth provider (browser back button on Google's consent screen), the resource still held the previous provider's `externalVerificationRedirectURL`, so a later click on a different provider sent them back to the original consent screen.

The fix scopes reuse to enterprise SSO only, where `prepare_first_factor` refreshes server state for ticket-based and identifier-discovery flows that compose `ticket() → sso()` against a single sign-in. OAuth strategies have no equivalent refresh, so they now always start fresh. This closes the cross-provider regression. Also resets the async resource fetch status when restoring a pending task from BFCache, which was leaving subsequent attempts stuck. Coverage spans the resource-level invariants plus a Playwright test in the custom-flows-react-vite template that exercises the abandoned-redirect retry end-to-end.


EandrewJones:
FWIW: I'm quite sure this same issue applies to `signUp` future as well. If a user backs out of the pop up or the redirect page, future attempts fail unless `signUp.reset()` is called.

Moreover, the interface definitions for `signUp/signIn.sso()` claim the promise returns an object with a `ClerkError`:
```js
  sso: (params: SignUpFutureSSOParams) => Promise<{
    error: ClerkError | null;
  }>;
```
However, when the above error occurs, it bubbles up a JSON syntax error instead, so it's not possible to condition logic based on the clerk error attributes.
image

Additionally, it appears users can get stuck in this state even if they don't back out of SSO.

jacekradko:
Thanks for the report, @EandrewJones. I looked into this and the sign-up case appears to be a separate issue from the one this PR fixes, not the same underlying cause.

The JSON SyntaxError is a separate problem worth fixing on its own: the { error: ClerkError | null } contract should hold regardless of what's happening underneath.  The raw parse error escaping that contract is its own bug.

I created an internal ticket to investigate both the retry behavior and the error-shape issue. I'll follow up here once we have something to share.

dstaley (approved):
this makes sense to me! basically if we call sso and there was an existing firstFactorVerification URL we call create if the strategies don't match, which seems correct.

PR #8503, opened 2026-05-07
feat(ui,shared,localizations): provider selection step for `<__experimental_ConfigureSSO />`

iagodahlem:
## Summary

Implements the first wizard step of `<__experimental_ConfigureSSO />` — the provider selection screen — and the state machine foundation that makes it work end-to-end. The user picks one of two SAML providers (Okta Workforce, Custom SAML Provider), Continue creates the enterprise connection, and the wizard advances. On reload, `deriveInitialStep` lands the user on the correct step based on the connection's state. All user-visible copy is wired through `@clerk/localizations` with translations for every supported locale. No public component API change.

Linear: [ORGS-1462](https://linear.app/clerk/issue/ORGS-1462)

## Why

After the layered architecture landed in #8493, `SelectProviderStep` was registered as a regular `<Wizard.Step>` with a placeholder body. The step is structurally a wizard step (not a pre-wizard gate), but is filtered out of the breadcrumb in `ConfigureSSOHeader` by id so the visible step count stays at 4 (Verify domain → Configure → Test → Confirmation). This PR fills in the body with the real provider tiles, selection behavior, the localization plumbing — and the flow-context plumbing the rest of the wizard depends on. The `createConnection` call lands here as an interim location; a small follow-up will move it into Verify Domain's `EnterVerificationCodeStep.onResolve` once Laura's #8520 merges.

### Select Provider step UI

- `steps/SelectProviderStep.tsx` — replaces the placeholder body with:
  - Sub-header block (`Select your identity provider` + helper copy)
  - `SAML` group containing two equal-width tiles laid out via `Grid columns={2}`
  - Warning callout (`<Alert variant='warning'>`) about provider lock-in
  - Local `useState<ProviderType | null>` selection state
  - Inline error display sourced from `useCardState`
  - `Step.Footer.Continue` gated on `isDisabled={isLastStep || !selected || card.isLoading}`; shows a loading state during the async create call
  - `Step.Footer.Previous` left as-is (auto-disabled by `isFirstStep`)
  - Every user-visible string consumed via `localizationKeys('configureSSO.selectProviderStep.*')` (no inline literals)
- `ProviderCard` (private to the file) — `<SimpleButton variant='outline'>` with a column-flex icon + label, height `$32`, gap `$2`, padding `$1x5`. Selected state renders a 4px `boxShadow` ring using `theme.colors.$colorRing`. `aria-pressed` mirrors the selected state for assistive tech.
- Provider icons sourced from the `img.clerk.com` static CDN via `iconImageUrl('okta')` and `iconImageUrl('saml')`. Rendered as an inline `<Span aria-hidden>` with `backgroundImage`, mirroring `ProviderIcon`'s non-mask-image branch — we don't wrap with `<ProviderIcon>` directly because its `id` prop union doesn't include `'saml_custom'`.
- `elements/Step.tsx` — minor `Step.Header` polish: top-align the inner row (`align='start'`) and bump the title/description column gap to `$2` to match the Figma. No behavior change.

### State machine foundation

- `ConfigureSSOContext.tsx` — extends the flow context with `provider`, `setProvider`, `clearProvider`, and `createConnection`. `provider` derives from `enterpriseConnection?.provider ?? localProvider`. `createConnection` is wrapped in `useReverification`, idempotent (no-ops when a connection already exists), and assembles the payload from the user's primary email domain + active organization id.
- `deriveInitialStep.ts` (new) — pure helper that maps a connection state to the wizard step the user should land on: no connection → `select-provider`, no `samlConnection.idpSsoUrl` → `configure`, otherwise → `confirmation`. Test step derivation is deferred (no synchronous "last test passed" signal).
- `ConfigureSSO.tsx` — wires `Wizard.initialStepId={deriveInitialStep(enterpriseConnection)}` so reloads land on the right step.
- `SelectProviderStep` Continue handler now `setProvider(selected) → createConnection() → goNext()`, all wrapped in `useCardState` (`setLoading` / `setIdle` / `setError`) and routed through `handleError` on rejection.
- `ProviderType` renamed from UI-side (`'okta' | 'custom_saml'`) to backend values (`'saml_okta' | 'saml_custom'`) so the flow context, the resource call, and the resource's own `provider` field all agree.

### Localization

- `@clerk/shared` — adds the typed `selectProviderStep` namespace to `__internal_LocalizationResource` (under `__experimental_configureSSO`).
- `@clerk/localizations` — adds English values in `en-US.ts` and translations for every other supported locale (~49 files).

## Out of scope — planned follow-ups

- **Move `createConnection` into Verify Domain.** Once #8520 merges, the create call moves to `EnterVerificationCodeStep.onResolve` so the connection is created after email + OTP verification, not on Select Provider Continue. The `createConnection` idempotency guard in this PR is what enables that move safely.
- **OIDC provider tile** — explicitly deprioritized. SAML covers v1 enterprise demand.
- **Reset semantics** returning the user back to Select Provider — wired from the Confirmation step in the same follow-up.

## Test plan

- [x] Sandbox renders `<__experimental_ConfigureSSO />` and lands on Select Provider on cold start (`deriveInitialStep` returns `select-provider`)
- [x] Breadcrumb shows 4 items (Verify domain / Configure / Test / Confirmation), all greyed because Select Provider is filtered out
- [x] Both SAML tiles render side-by-side with the real Okta + SAML logos loaded from `img.clerk.com`
- [x] Continue is disabled on entry; selecting a tile enables Continue and renders the 4px ring on the chosen tile
- [x] `aria-pressed` flips on the selected tile (verified via DevTools accessibility tree)
- [x] Clicking Continue shows loading state → enterprise connection created in backend → wizard advances to Verify Domain
- [x] Previous stays disabled throughout (this is the first step)
- [x] Backend failure on Continue → error rendered in the step body, wizard does not advance
- [x] Unit tests cover: mount, both tiles render, Continue gating, click-to-select, click-Continue → `setProvider` + `createConnection` + `goNext` invoked in order, loading state, error path
- [x] `deriveInitialStep` table-driven tests cover undefined connection / no `samlConnection` / no `idpSsoUrl` / has `idpSsoUrl`
- [ ] Sandbox: reload mid-flow with connection but no metadata → mounts on Configure
- [ ] Sandbox: reload mid-flow with connection + metadata → mounts on Confirmation
- [ ] Sandbox visual pass with the localized strings resolved (pending manual QA)

## Type of change

- [x] ✨ New feature (`__experimental_*` surface — no public API change)


packages/ui/src/components/ConfigureSSO/elements/Step.tsx
```diff
@@ -54,11 +54,11 @@ const Header = ({ title, description, children }: StepHeaderProps): JSX.Element
       })}
     >
       <Flex
-        align='center'
+        align='start'
         justify='between'
         sx={theme => ({ gap: theme.space.$4 })}
       >
-        <Col sx={theme => ({ gap: theme.space.$1x5, minWidth: 0 })}>
+        <Col sx={theme => ({ gap: theme.space.$2, minWidth: 0 })}>
```
iagodahlem, line 61: Updates the badge display on the header.

packages/ui/src/components/ConfigureSSO/deriveInitialStep.ts
```diff
@@ -0,0 +1,25 @@
...
+ * passed" signal synchronously from the resource. Users can re-test from
+ * Confirmation.
+ */
+export const deriveInitialStep = (connection: EnterpriseConnectionResource | undefined): WizardStepId => {
+  if (!connection) {
+    return 'select-provider';
+  }
+  if (!connection.samlConnection?.idpSsoUrl) {
+    return 'configure';
+  }
```
LauraBeatris, line 23: This would translate as "minimum IdP configuration" -> we could extract to a boolean as well or add a comment, but just to highlight that this can also be extended in the future for OIDC, we'll prob reuse the same function here

packages/ui/src/components/ConfigureSSO/steps/SelectProviderStep.tsx
```diff
@@ -13,27 +101,144 @@ export const SelectProviderStep = (): JSX.Element => {
...
+  isSelected?: boolean;
+  onClick?: () => void;
+};
+
+const ProviderCard = ({ iconId, label, isSelected, onClick }: ProviderCardProps): JSX.Element => {
+  const { t } = useLocalizations();
+  const labelText = t(label);
+
+  return (
+    <SimpleButton
```
alexcarpenter: This really should be a radio input styled as a button.

We lose the selected state boxShadow when clicking the button, then clicking elsewhere on the page

packages/ui/src/components/ConfigureSSO/steps/SelectProviderStep.tsx
```diff
@@ -13,27 +101,144 @@ export const SelectProviderStep = (): JSX.Element => {
...
+    >
+      <Span
+        aria-hidden
+        sx={theme => ({
+          width: theme.sizes.$8,
+          height: theme.sizes.$8,
+          backgroundImage: `url(${iconImageUrl(iconId)})`,
+          backgroundSize: 'contain',
+          backgroundPosition: 'center',
+          backgroundRepeat: 'no-repeat',
```
alexcarpenter: Similar to sso providers, we'll need to map the white provider icons  to flip in dark mode.

Image

iagodahlem:
@alexcarpenter fixed both: 
CleanShot 2026-05-13 at 17 21 34@2x

LauraBeatris (approved):

PR #8524, opened 2026-05-12
fix(repo): Add standalonePage tag for JSDoc comments

alexisintech:
## Description

We were getting errors that certain files weren't being generated anymore. More information here: https://github.com/clerk/javascript/pull/8276#issuecomment-4425942784

This PR adds a `@standalonePage` tag that can be used in conjunction with the `@inline` tag to indicate to typedoc to still generate a dedicated page for that type, even though it is recognized as `@inline`. Using both of these tags means that the type will show up as inlined in a table, but will also still have a generated MDX file.
For example:
`HookParams` will be inlined in a table:
Screenshot 2026-05-11 at 17 22 38
but will still get a generated `shared/hook-params.mdx` file that can then be used in the docs:
Screenshot 2026-05-11 at 17 23 22

## Checklist

## Type of change

- [X] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.typedoc/standalone-page-tag.mjs
+++ b/.typedoc/standalone-page-tag.mjs
@@ -1,8 +1,7 @@
 // @ts-check
 
 /**
- * Modifier tag: keep a dedicated `.mdx` page even when `@inline` is present (TypeDoc + our router
- * otherwise drop inline-marked declarations; the theme also expands references instead of linking).
+ * Modifier tag: keep a dedicated `.mdx` page even when `@inline` is present (TypeDoc + our router otherwise drop inline-marked declarations; the theme also expands references instead of linking).
  */
 export const STANDALONE_PAGE_MODIFIER_TAG = '@standalonePage';
 
--- a/.typedoc/standalone-page-tag.mjs
+++ b/.typedoc/standalone-page-tag.mjs
@@ -1,0 +1,25 @@
+// @ts-check
+
+/**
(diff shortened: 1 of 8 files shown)


.typedoc/standalone-page-tag.mjs
```diff
@@ -0,0 +1,25 @@
+// @ts-check
+
+/**
+ * Modifier tag: keep a dedicated `.mdx` page even when `@inline` is present (TypeDoc + our router
+ * otherwise drop inline-marked declarations; the theme also expands references instead of linking).
```
alexisintech:
```suggestion
 * Modifier tag: keep a dedicated `.mdx` page even when `@inline` is present (TypeDoc + our router otherwise drop inline-marked declarations; the theme also expands references instead of linking).
```

SarahSoutoul (approved):
This looks good! 

One non-blocking follow-up: we may want to add `.typedoc/standalone-page-tag.mjs` to the Typedoc inputs in `turbo.json` so Turbo invalidates the cache when that script changes.

alexisintech:
> This looks good!
> 
> One non-blocking follow-up: we may want to add `.typedoc/standalone-page-tag.mjs` to the Typedoc inputs in `turbo.json` so Turbo invalidates the cache when that script changes.

oh cool! done [9e29238](https://github.com/clerk/javascript/pull/8524/commits/9e29238e449bdc4ea376e1f6bc01463be5e35f97)

PR #8525, opened 2026-05-12
chore(e2e): retire tailwind-merge trusted-publisher override

jacekradko:
Follow-up to #8522. Bumps `tailwind-merge` direct dependencies to `^3.4.0` in the two integration templates and one playground that were still on v2.5/v3.3, then drops the `tailwind-merge@<3.4.0` entry from root `pnpm.overrides` and from `TRUSTED_OVERRIDES` in `integration/constants.ts`.

v3.4.0 is the first version with `trustedPublisher` evidence, so once everything resolves to 3.4 or above, pnpm 10's trust-downgrade check is satisfied without needing the override. The only source-level consumers use `twMerge(clsx(...))`, which is API-identical in v2 and v3.


Ephem (approved):

jacekradko:
superseded by https://github.com/clerk/javascript/pull/8536

PR #8527, opened 2026-05-12
feat(clerk-js,shared,ui): Add seats info to payment attempts page

aeliox:
Adds seat info (limits, usage, and costs) to the payment attempts page of an OrganizationProfile's billing section.

Screenshot 2026-05-15 at 1 43 27 PM


packages/ui/src/components/PaymentAttempts/PaymentAttemptPage.tsx
```diff
@@ -198,18 +197,25 @@ export const PaymentAttemptPage = () => {
...
+  const { subscriptionItem } = paymentAttempt;
+
   const fee =
     subscriptionItem.planPeriod === 'month'
       ? // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         subscriptionItem.plan.fee!
       : // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         subscriptionItem.plan.annualMonthlyFee!;
 
+  const seatsTotal = subscriptionItem.seats != null ? getSeatsPerUnitTotal(paymentAttempt.totals) : undefined;
```
dstaley, line 214: I think in this situation a `null` value in seats represents "Unlimited seats".

packages/ui/src/components/PaymentAttempts/PaymentAttemptPage.tsx
```diff
@@ -225,6 +231,28 @@ function PaymentAttemptBody({ subscriptionItem }: { subscriptionItem: BillingSub
...
+                planSeatLimit != null
+                  ? localizationKeys('billing.seatsWithLimit', { limit: planSeatLimit })
+                  : localizationKeys('billing.seats')
+              }
+              description={(() => {
+                const seatLabel = `${seatsChargeable} ${seatsChargeable === 1 ? 'seat' : 'seats'}`;
+                const rate = `${seatSummary.paidTier.feePerBlock.currencySymbol}${seatSummary.paidTier.feePerBlock.amountFormatted}/mo`;
+                return seatSummary.included > 0
+                  ? `${seatSummary.used} used − ${seatSummary.included} included → ${seatLabel} at ${rate}`
+                  : `${seatLabel} at ${rate}`;
```
dstaley: This should be done through the localization system.

packages/ui/src/utils/billingPlanSeats.ts
```diff
@@ -24,70 +25,53 @@ export const getSeatUnitPrice = (plan: { unitPrices?: BillingPlanUnitPrice[] }):
...
-    unitTotal.tiers[0].feePerBlock.amount === 0 &&
-    unitTotal.tiers[1].feePerBlock.amount > 0
-  ) {
-    return unitTotal.tiers[1];
-  }
-
-  return null;
+export type SeatChargeSummary = {
+  /** Sum of `quantity` across all tiers (paid + included). */
+  used: number;
```
dstaley: I think calling this "used" is misleading. This number represents the number of seats an organization was charged for, not how many seats are used. Maybe "total" instead since it represents both included and purchased seats?

packages/ui/src/utils/billingPlanSeats.ts
```diff
@@ -96,7 +80,8 @@ export const getPlanSeatLimit = (plan: BillingPlanResource): number | null | und
     return undefined;
   }
 
-  return seatUnitPrice.tiers[seatUnitPrice.tiers.length - 1]?.endsAfterBlock;
+  const lastTier = seatUnitPrice.tiers[seatUnitPrice.tiers.length - 1];
+  return lastTier.endsAfterBlock != null ? lastTier.endsAfterBlock * seatUnitPrice.blockSize : null;
```
dstaley, line 150: good catch!

packages/ui/src/components/PaymentAttempts/PaymentAttemptPage.tsx
```diff
@@ -225,6 +231,28 @@ function PaymentAttemptBody({ subscriptionItem }: { subscriptionItem: BillingSub
...
+              title={
+                planSeatLimit != null
+                  ? localizationKeys('billing.seatsWithLimit', { limit: planSeatLimit })
+                  : localizationKeys('billing.seats')
+              }
+              description={(() => {
+                const seatLabel = `${seatsChargeable} ${seatsChargeable === 1 ? 'seat' : 'seats'}`;
+                const rate = `${seatSummary.paidTier.feePerBlock.currencySymbol}${seatSummary.paidTier.feePerBlock.amountFormatted}/mo`;
+                return seatSummary.included > 0
+                  ? `${seatSummary.used} used − ${seatSummary.included} included → ${seatLabel} at ${rate}`
```
dstaley: I think a comma instead of an arrow here would probably be better (especially since people will be writing this string in their localizations):

```
5 total - 2 included, 3 seats at $10.00/mo
```

packages/shared/src/types/json.ts
```diff
@@ -740,6 +740,19 @@ export interface BillingStatementGroupJSON extends ClerkResourceJSON {
   items: BillingPaymentJSON[];
 }
 
+/**
+ * @experimental This is an experimental API for the Billing feature that is available under a public beta, and the API is subject to change. It is advised to [pin](https://clerk.com/docs/pinning) the SDK version and the clerk-js version to avoid breaking changes.
+ *
+ * Per-payment cost breakdown including optional base fee and per-unit (for example, seats) subtotals.
+ */
+export interface BillingPaymentTotalsJSON {
```
dstaley, line 748: I think this is actually our `CommerceTotalsResponse` type, which is shared between checkout, invoices, and payments, rather than a payments specific shape. I have one in my checkout PR, so we can probably just land this as-is in your PR, and then I'll update it to use the same type as checkout in mine. Just calling it out in case the checkout PR lands before this one.

packages/localizations/src/en-US.ts
```diff
@@ -168,6 +167,7 @@ export const enUS: LocalizationResource = {
     },
     reSubscribe: 'Resubscribe',
     seats: 'Seats',
+    seatsWithLimit: 'Seats ({{limit}} limit)',
```
dstaley: would it make sense to use our existing `upToSeats` localization value? It'd read a little different `Up to X seats`

packages/ui/src/components/PaymentAttempts/PaymentAttemptPage.tsx
aeliox: I think that would be represented by `subscriptionItem.seats.quantity` being `null`, right?

dstaley: ah yeah my bad, forgot it was a `{ quantity: number | null }`!

packages/ui/src/utils/billingPlanSeats.ts
aeliox: I agree, rolling with `totalSeats` to be explicit

packages/ui/src/components/PaymentAttempts/PaymentAttemptPage.tsx
aeliox: Hmm, what do you think about an equals sign? I want it to be a bit more clear that it's an equation

packages/localizations/src/en-US.ts
aeliox: I tried it out and it doesn't sufficiently feel like a label

packages/ui/src/components/PaymentAttempts/PaymentAttemptPage.tsx
aeliox: or maybe parens? `(5 total - 2 included) 3 seats at $10.00/mo`

aeliox: Maybe reversing the layout – a little less like equation but flows pretty logically as a line item...

Screenshot 2026-05-15 at 1 35 10 PM

packages/localizations/src/en-US.ts
aeliox: Ooo but I do kinda like `Seats (up to 20)` better than using the word `limit`

packages/ui/src/components/PaymentAttempts/PaymentAttemptPage.tsx
dstaley: I think I like that one the most! I didn't realize it was formatted like an equation until you pointed it out.

PR #8520, opened 2026-05-11
chore(ui,shared,localizations): Add verify/add email address step to `<ConfigureSSO />`

LauraBeatris:
## Description

- Handles warning when there's already a connection created for the given domain when changing to an organization context
- Handles adding an primary email address when user account is created with username/phone as identifiers 
- Handles verifying email address if not performed previously via sign-in/sign-up 

#### Add + verify

https://github.com/user-attachments/assets/c74da068-0294-4537-97e4-4411d6d3e6a4

#### Verify only

https://github.com/user-attachments/assets/a3460d63-e423-48d2-b9e7-dbca6887d1ff

## Checklist

- [X] `pnpm test` runs as expected.
- [X] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #8535, opened 2026-05-12
feat(ui): build out Okta Configure step UI in ConfigureSSO

iagodahlem:
## Description

Builds out the Configure step UI for `<__experimental_ConfigureSSO />` (Okta provider). The Configure step is now a nested 4-sub-step inner wizard, mirroring the pattern Laura established for the Verify Domain step:

1. **Create app** — onboarding walkthrough for the Okta admin console + read-only SP-side copy rows (ACS URL, Audience URI) sourced from `connection.samlConnection.{acsUrl, spEntityId}`.
2. **Configure attributes** — SAML attribute mapping table (Email Required / First Name Optional / Last Name Optional with `user.email` / `user.firstName` / `user.lastName` claim names) + guidance on setting up Okta's Attribute Statements, with name/expression pairs rendered as badges.
3. **Assign users** — Okta admin steps for assigning users or groups to the enterprise app.
4. **Submit SAML config** — IdP metadata URL submission. The user pastes the URL from Okta and the wizard advances on a successful `PATCH /me/enterprise_connections/{id}` with `{ saml: { idpMetadataUrl } }`.

The mutation is hoisted onto `ConfigureSSOProvider` and wrapped in `useReverification`, matching the convention for sensitive `user.*` mutations in `@clerk/ui`. `useCardState` drives the loading state; `handleError` routes backend errors inline under the field when the API returns `idp_metadata_url`, or to the card-level error surface otherwise. The hook owns query revalidation after a successful mutation, so the in-context `enterpriseConnection` reflects the new IdP metadata without an explicit refetch.

Locale keys live under `configureSSO.configureStep`. Shared keys (attribute mapping table, badges, SP field labels) sit at the top of the namespace; provider-specific copy lives under `configureSSO.configureStep.samlOkta`. Future Custom SAML and OIDC providers can drop in a sibling namespace without colliding.

Also fixes `toMeEnterpriseConnectionBody` in `@clerk/clerk-js` to emit a flat snake-case body (`saml_idp_metadata_url`) instead of the nested shape (`saml: { idp_metadata_url }`) the previous helper produced. The backend expects flat-prefixed fields, so the prior shape was silently rejected for `user.createEnterpriseConnection` and `user.updateEnterpriseConnection`.

Linear: ORGS-1456

### Follow-ups (separate PR)

The "Configure manually" segmented-control mode of sub-step 4 ships in a follow-up PR. That adds:

- Tabbed segmented control between metadata-URL and manual-entry modes
- Sign on URL + Issuer + Signing certificate form fields (with new `FieldId` entries)
- Optional file upload + client-side XML parse for the metadata-file mode shown in the Figma

Tests for the Configure sub-steps ship alongside Laura's #8520 follow-up sweep once both land.

### How to test

In a sandbox with the ConfigureSSO wizard, enter through Select Provider → Verify Domain → land on Configure. Walk through sub-steps 1 through 3 (Previous / Continue navigate the inner wizard), then on sub-step 4 paste a valid Okta SAML metadata URL → Continue → connection is patched and the wizard advances to Test. Backend errors keyed `idp_metadata_url` render inline; others surface at the card level.

### Base

Stacked on #8503. GitHub will auto-retarget this PR to \`main\` once that merges.

## Checklist

- [ ] \`pnpm test\` runs as expected.

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/ui/src/components/ConfigureSSO/steps/ConfigureStep.tsx
+++ b/packages/ui/src/components/ConfigureSSO/steps/ConfigureStep.tsx
@@ -450,11 +450,11 @@ export const ConfigureAttributesSubStep = (): JSX.Element => {
                 <Col
                   as='ul'
                   sx={theme => ({
-                    gap: theme.space.$1,
+                    gap: theme.space.$1x5,
                     margin: 0,
-                    marginTop: theme.space.$1,
+                    marginTop: theme.space.$1x5,
                     paddingInlineStart: theme.space.$5,
-                    listStyleType: 'disc',
+                    listStyleType: '"- "',
                   })}
                 >
                   {ATTRIBUTE_PAIRS.map(pair => (
--- a/packages/ui/src/components/ConfigureSSO/steps/ConfigureStep.tsx
+++ b/packages/ui/src/components/ConfigureSSO/steps/ConfigureStep.tsx
@@ -1,12 +1,57 @@
-import { descriptors, Flow } from '@/customizables';
+import { useReverification, useUser } from '@clerk/shared/react';
+import type { UpdateMeEnterpriseConnectionParams } from '@clerk/shared/types';
+
+import { descriptors, Flow, localizationKeys } from '@/customizables';
+import { useCardState } from '@/elements/contexts';
+import { Form } from '@/elements/Form';
+import { handleError } from '@/utils/errorHandler';
+import { useFormControl } from '@/utils/useFormControl';
 
 import { useConfigureSSO } from '../ConfigureSSOContext';
 import { Step } from '../elements/Step';
 import { useWizard } from '../elements/Wizard';
 
 export const ConfigureStep = (): JSX.Element => {
-  const { goNext, goPrev, isFirstStep, isLastStep } = useWizard();
+  const card = useCardState();
+  const { user } = useUser();
+  const { goNext, goPrev, isFirstStep } = useWizard();
   const { enterpriseConnection } = useConfigureSSO();
+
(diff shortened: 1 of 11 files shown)


packages/ui/src/components/ConfigureSSO/steps/ConfigureStep.tsx
```diff
@@ -1,39 +1,826 @@
...
+    isRequired: false,
+    attribute: localizationKeys('configureSSO.configureStep.attributeMapping.rows.lastName.attribute'),
+    claim: localizationKeys('configureSSO.configureStep.attributeMapping.rows.lastName.claim'),
+  },
+] as const;
+
+const ATTRIBUTE_PAIRS = [
+  {
+    id: 'email',
+    name: localizationKeys('configureSSO.configureStep.samlOkta.configureAttributes.pairs.email.name'),
```
LauraBeatris, line 102: In follow-up PRs, are we going to refactor those to receive the IdP type as a param in order to return different localization keys?

LauraBeatris (approved):

packages/ui/src/components/ConfigureSSO/steps/ConfigureStep.tsx
iagodahlem: Yep, that's the idea. The provider is already available from the ConfigureSSOContext. For now, it will always render as if Okta was selected.

PR #8536, opened 2026-05-12
chore(repo): drop tailwind-merge from templates and playground

jacekradko:
Nothing in `packages/*` uses `tailwind-merge`. The only consumers were two `cn()` helpers in the browser-extension playground and the custom-flows-react-vite integration template (and one stale dep declaration in the tanstack-react-start template). This drops the package and inlines `cn()` to `clsx`, and removes the trusted-publisher floor entries that pinned it.

Tradeoff worth noting: `cn(defaults, className)` in the affected shadcn-style components no longer resolves duplicate Tailwind utilities, so if a consumer passes a conflicting class, CSS source order wins instead of the call-site override. Acceptable for these template/demo surfaces.


wobsoriano (approved):
🧹

PR #8541, opened 2026-05-13
chore(repo): Add script to run integration tests in Docker

dstaley:
## Description

This PR adds a new script to allow running a provided integration test suite in Docker, preventing the installation of packages not tied to a lock file on a developer's local machine. Currently, our integration test suites do not utilize package lock files, meaning that when they're inflated as part of the integration test suite they're susceptible to recently-compromised third party packages. With that change, that vulnerability is contained to a Docker container with no access to the underlying host through bind mounts or host-mode networking. This approach also allows developer's machines to use a global `minimum-release-age` configuration that is incompatible with `pkglab`.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


integration/playwright.config.ts
```diff
@@ -38,10 +38,10 @@ export default defineConfig({
       testMatch: /global\.teardown/,
     },
     {
-      name: 'chrome',
+      name: process.env.E2E_BROWSER || 'chrome',
```
dstaley, line 41: Playwright doesn't contain a Chrome ARM64 browser; just Chromium. We can _probably_ just always use Chromium, but I've done it this way for now to maintain the previous behavior.

scripts/test-integration-docker.mts
```diff
@@ -0,0 +1,311 @@
...
+  const trackedFiles = trackedFilesOutput
+    .split('\0')
+    .filter(Boolean)
+    .filter(repoRelativePath => isExistingPath(repoRoot, repoRelativePath));
+  const dependencyFiles = trackedFiles.filter(isDependencyFile);
+  const buildFiles = trackedFiles.filter(isBuildFile);
+
+  await copyFiles(repoRoot, trackedFiles, path.join(contextDir, 'repo'));
+  await copyFiles(repoRoot, dependencyFiles, path.join(contextDir, 'deps'));
+  await copyFiles(repoRoot, buildFiles, path.join(contextDir, 'build'));
```
dstaley, line 226: The three different sets of files here is to support layer caching, preventing building everything from scratch on each execution of the script.

PR #8545, opened 2026-05-13
chore(ci): remove cache restore in OIDC publish jobs

jacekradko:
Closes [SDK-78](https://linear.app/clerk/issue/SDK-78/defuse-cache-in-oidc-publish-jobs).

Adds a `cache-enabled` input to the `init` and `init-blacksmith` composite actions, defaulting to `false` so privileged jobs can't accidentally restore poisonable cache state. The three `id-token: write` jobs in `release.yml` inherit the safe default; every other CI caller now opts in with `cache-enabled: true`. Also flips the monorepo install to `--frozen-lockfile` and drops the misleading `playwright-enabled: true` flag from release jobs.

First step in the supply-chain hardening project; the principle is to keep cache restoration out of any job that holds an OIDC publish token, which is the primitive the recent TanStack compromise exploited.


.github/actions/init-blacksmith/action.yml
```diff
@@ -111,14 +115,14 @@ runs:
...
-        cache: pnpm
+        cache: ${{ inputs.cache-enabled == 'true' && 'pnpm' || '' }}
         node-version: ${{ inputs.node-version }}
         registry-url: ${{ inputs.registry-url }}
 
     - name: Install PNPM Dependencies
       env:
         CYPRESS_INSTALL_BINARY: 0
-      run: pnpm install
+      run: pnpm install --frozen-lockfile
```
dominic-clerk, line 125: Definitely no harm in have it explicit, but for your information `--frozen-lockfile` is implied by `pnpm` when it detects a CI environment https://pnpm.io/cli/install#--frozen-lockfile

dominic-clerk (approved):
Love it! LGTM

.github/actions/init-blacksmith/action.yml
jacekradko: Yep 👍🏼

PR #8549, opened 2026-05-14
chore(shared): replace telemetry postinstall with runtime notice

### LINKED ISSUE #8540: security: clerk runs third party posinstall scripts
### Preliminary Checks

### Reproduction

n/a

### Publishable key

pk_test_123456789

### Description

npm postinstall scripts are a significant security risk in the JavaScript ecosystem and are frequently used as an attack vector in supply-chain incidents. Given the current threat landscape, many teams are moving toward blocking or heavily restricting these scripts by default.

Clerk currently uses a postinstall script for **telemetry**. We do not believe telemetry justifies the security tradeoff of running install-time scripts, especially when users and organizations are actively trying to reduce exposure to this class of risk.

We plan to block these scripts in our environment. For the security of your users, and in recognition of the ongoing risks in the JavaScript ecosystem, we ask that Clerk remove the use of postinstall scripts for telemetry.

### Environment

```Shell
As a security practice I no longer run npx, pnpm dlx or equivalent one off package execution commands on my machine. 

latest macOS, latest clerk SDKs
```

jacekradko:
Removes the postinstall script from `@clerk/shared` and drops the `std-env` dependency. Disclosure now happens at runtime when the telemetry collector boots on a development instance, server-side only, deduped per process via a `globalThis` flag. The browser console intentionally doesn't get the notice, keeping the noise profile similar to the old install-time message. The runtime path doesn't touch `node:fs`, `process.versions`, or any dynamic-code APIs, so it stays safe for Edge Runtime, Workers, and browser bundles.

Resolves SDK-84. Closes #8540.


brkalow:
@jacekradko my hunch is that this is going to trigger more than desired in the browser, and we'll get requests to disable the warning altogether similar to what happened with the dev mode warning. How can we get ahead of that? 🤔

jacekradko:
@brkalow I think you are right. It's tricky in the browser. Are we ok with this only on the server?

dominic-clerk:
Just dropping a comment here to say that I love this.

packages/shared/src/__tests__/telemetry.notice.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+  'GITHUB_ACTIONS',
+  'GITLAB_CI',
+  'CIRCLECI',
+  'TRAVIS',
+  'BUILDKITE',
+  'JENKINS_URL',
+  'TF_BUILD',
+  'DRONE',
+  'CODEBUILD_BUILD_ID',
+];
```
wobsoriano: There's this keyless [PR](https://github.com/clerk/javascript/pull/8676) that introduces shared CI env vars ([ref](https://github.com/clerk/javascript/pull/8676/changes/BASE..c11afee992047d7f14428a93a3b7401964636e86#r3319137259)). Should we re-export this somewhere to be used by that PR?

wobsoriano (approved):
looks good, left a q

packages/shared/src/__tests__/telemetry.notice.spec.ts
jacekradko: Nice catch! Didn't see the keyless PR before. I think we should actually use that PR's list of env vars. We can wait for that to merge before using it here

wobsoriano: That PR is merged 👍🏼

PR #7926, opened 2026-02-25
chore(express): Remove deprecated `enableHandshake` option and `req.auth` property access

wobsoriano:
## Description

This PR removes deprecated `enabledHandshake` option and direct `req.auth` property access.

Resolves USER-4803

## Checklist

- [x] `pnpm test` runs as expected.
- [x] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed the deprecated handshake option from middleware — update middleware calls accordingly.
  * Direct access to request auth as an object was removed; retrieve authentication via the getAuth-style call.

* **Refactor**
  * Simplified authentication flow and removed legacy handshake/compatibility pathway.

* **Tests**
  * Removed obsolete tests covering the deprecated auth/handshake behavior.

* **Documentation**
  * Updated upgrade notes documenting the removal and migration examples.

PR #8560, opened 2026-05-15
fix(fastify): Add __internal_enableHandshake option to skip handshake flow

jescalan:
## Summary
- add an internal `__internal_enableHandshake` option to `clerkPlugin()` (defaults to `true`)
- when set to `false`, strip handshake cookies (`__clerk_handshake`, `__clerk_handshake_nonce`) and query params before calling `authenticateRequest`, and skip handshake redirects
- keep dev-browser handshake redirects (`dev-browser-missing`, `dev-browser-sync`) even when disabled, since development instances cannot establish auth state without them
- fall back to a signed-out auth object when a handshake redirect is skipped (a handshake state's `toAuth()` is `null`), and drop the already-applied `Location`/`Cache-Control` headers from the reply
- add focused regression coverage for the disabled flow, the dev-browser exception, and the default behavior

## Context
Forced handshake nonce transport stores a short `__clerk_handshake_nonce` instead of the large `__clerk_handshake` payload. Server SDKs then need to exchange that nonce through the Backend API client attached to `authenticateRequest`.

On an API-only backend that cannot deliver `Set-Cookie` headers back to the browser (e.g. Fastify on Lambda behind API Gateway, where multi-value `Set-Cookie` headers are dropped), the browser can never clear that nonce cookie. Every subsequent request replays the already-consumed nonce, and the exchange returns `404 resource_not_found`, logging `Clerk: HandshakeService: error getting handshake payload` on each request. `authenticateRequest` already tolerates the failed exchange and falls through to cookie verification, so disabling the handshake removes the per-request error noise, the wasted Backend API call, and handshake redirects that an API client can never complete.

The option is `__internal_`-prefixed rather than public because `@clerk/express` deprecated (#6499) and removed (#7926) its public `enableHandshake` option, and we don't want to reintroduce public API surface for this. The affected customer has been running the equivalent behavior in production via snapshot since May 22 with the errors fully gone.

## Notes
- The runtime-keys client construction changes originally on this branch shipped separately in #8640, so the branch was rebuilt on `main` and now contains only the handshake option.
- The dev-browser reasons are compared as string literals because `AuthErrorReason` is not exported from `@clerk/backend/internal`; exporting it is a one-line `@clerk/backend` change if preferred.

<span title="This PR description was written by AI using employee-skills/write-pr-description">✎</span>


jescalan:
!snapshot

jeremy-clerk (approved):
LGTM

wobsoriano:
@copilot can you add changes from https://github.com/clerk/javascript/pull/8563 to this PR?

wobsoriano:
!snapshot

wobsoriano:
!snapshot

wobsoriano:
!snapshot

jorgevrgs:
It looks like `enableHandshake: false` doesn't prevent the API to request `GET /v1/client/handshake` always if the cookies `__clerk_handshake` and/or `__clerk_handshake_nonce` exist. That causes a noisy log:

```
ERROR Clerk: HandshakeService: error getting handshake payload: ClerkAPIResponseError: Not Found
  at ClientAPI.request (.../@clerk/backend/src/api/request.ts:278:21)
  at async HandshakeService.getCookiesFromHandshake (.../@clerk/backend/src/tokens/handshake.ts:187:34)
  at async HandshakeService.resolveHandshake (.../@clerk/backend/src/tokens/handshake.ts:220:26)
  at async authenticateRequestWithTokenInCookie (.../@clerk/backend/src/tokens/request.ts:457:16)
  at async Object.<anonymous> (.../@clerk/fastify/src/withClerkMiddleware.ts:88:26)
```

Quick solution: strip/ignore cookies/queries when the config is `enableHandshake: false`.

wobsoriano:
!snapshot

wobsoriano:
@jorgevrgs thanks, generated another snapshot for you

`npm i @clerk/fastify@3.1.28-snapshot.v20260516134813 --save-exact`

wobsoriano:
!snapshot

jescalan:
Where are we at with this one? Don't want it to linger as a single customer hotfix

wobsoriano:
!snapshot

jescalan:
Bumping this again

jescalan:
Another bump here - this is debt that I feel that we need to clear out 😬

wobsoriano:
!snapshot

wobsoriano (approved):

PR #8580, opened 2026-05-18
chore(repo): Pin renovate config validator version

dominic-clerk:
## Description

This pins the validator version and removes the unnecessary init step to harden against supply chain compromises. In the previous workflow we would fetch the latest version and bypass other security mechanisms we have around dependency management.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] other: CI

### DIFF
--- a/.github/workflows/validate-renovate-config.yml
+++ b/.github/workflows/validate-renovate-config.yml
@@ -16,14 +16,5 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
 
-      - name: Setup
-        id: config
-        uses: ./.github/actions/init
-        with:
-          cache-enabled: true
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-renovate-config.yml
```diff
@@ -16,14 +16,5 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
 
-      - name: Setup
```
dominic-clerk, line 19: we use `npx` so setting up the whole environment, injecting secrets and running `pnpm install` didn't do anything as `renovate` isn't in the dependencies we install.

wobsoriano (approved):

PR #8587, opened 2026-05-18
chore(backend, clerk-js): Deprecate metadata updates in user update methods

brunol95:
## Description

- deprecates metadata fields in update user methods and refactors implementation to route to dedicated metadata endpoints
- adds a new replaceUserMetadata method 

## Checklist

- [x] update `SUPPORTED_BAPI_VERSION` and `SUPPORTED_FAPI_VERSION` constants

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.changeset/deprecate-update-user-metadata.md
wobsoriano, line 1: Let's split this into separate changesets so each package changelog stays focused 👍🏼 

1. for backend
2. for clerk-js and shared

packages/clerk-js/src/core/resources/User.ts
```diff
@@ -235,9 +237,42 @@ export class User extends BaseResource implements UserResource {
...
+    // *replace* semantics of `user.update({ unsafeMetadata })` by
+    // diffing the locally-cached value against the desired one and sending
+    // an RFC 7396 merge patch (null-deletes for removed keys).
+    if (hasRest) {
+      await this._basePatch({
+        body: normalizeUnsafeMetadata(rest as UpdateUserParams),
+      });
+    }
+
+    const patch = computeMergePatch(this.unsafeMetadata, unsafeMetadata);
```
wobsoriano, line 265: Does this preserve replacement semantics if `this.unsafeMetadata` is not fresh at the moment of the call? Touch-on-focus can refresh the user in common tab-switching flows, but `user.update({ unsafeMetadata })` previously sent the caller-provided value as the full replacement regardless of local cache state. With this client-side diff, a stale local value can produce an incomplete patch, or `{`} and skip the request entirely.

Correct me if Im wrong 😄

wobsoriano (changes requested):
Left a small suggestion and a question.

Also if possible, I think we should add an e2e test. We can place the test here https://github.com/clerk/javascript/blob/main/integration/tests/unsafeMetadata.test.ts

packages/clerk-js/src/core/resources/User.ts
brunol95: Ah yes you are right . There are two scenarios:

1. if we execute the  call to PATCH `v1 /me` endpoint first that returns a fresh `user` object. No issue here since the diff will be using the latest data.
2. If the request is a metadata only update, then theres a chance the `user` object could be stale. In this case, I think we'll need to perform a reload first.

.changeset/deprecate-update-user-metadata.md
brunol95: done

packages/clerk-js/src/core/resources/User.ts
brunol95: @wobsoriano I pushed a fix. Let me know what you think!

.changeset/route-unsafe-metadata-to-merge-endpoint.md
```diff
@@ -0,0 +1,6 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/shared': patch
+---
+
+Deprecate `unsafeMetadata` on `user.update()`. Use `user.updateMetadata({ unsafeMetadata })` for partial updates (deep merge) instead. The parameter continues to work — metadata is now routed through `PATCH /v1/me/metadata`.
```
wobsoriano:
```suggestion
Deprecate passing `unsafeMetadata` to `user.update()`.

Use `user.updateMetadata()` when you want to partially update unsafe metadata with deep-merge semantics:

~~~ts
await user.updateMetadata({
  unsafeMetadata: { onboardingComplete: true },
});
~~~

user.update({ unsafeMetadata })` continues to work for now and preserves its existing full-replacement behavior:

~~~ts
await user.update({
  unsafeMetadata: { theme: 'dark' },
});
~~~

New code should prefer `user.updateMetadata({ unsafeMetadata })` for metadata-only updates.
```

.changeset/deprecate-update-user-metadata.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/backend': minor
+---
+
+Add `clerkClient.users.replaceUserMetadata(userId, params)` for full-replacement metadata updates. Deprecate the `publicMetadata`, `privateMetadata`, and `unsafeMetadata` parameters on `clerkClient.users.updateUser` — use `updateUserMetadata` for partial updates (deep merge) or `replaceUserMetadata` for full replacement.
```
wobsoriano:
```suggestion
Add `clerkClient.users.replaceUserMetadata(userId, params)` for replacing a user's metadata fields in full.

Use `replaceUserMetadata` when the provided metadata should become the complete value for that metadata field:

~~~ts
await clerkClient.users.replaceUserMetadata(userId, {
  publicMetadata: { plan: 'pro' },
});
~~~

Use `clerkClient.users.updateUserMetadata(userId, params)` when you want to partially update metadata with deep-merge semantics:

~~~ts
await clerkClient.users.updateUserMetadata(userId, {
  publicMetadata: { onboardingComplete: true },
});
~~~

The `publicMetadata`, `privateMetadata`, and `unsafeMetadata` parameters on `clerkClient.users.updateUser()` are now deprecated. They continue to work, but new code should use `updateUserMetadata()` for partial updates or `replaceUserMetadata()` for full replacement.
```

integration/tests/unsafeMetadata.test.ts
```diff
@@ -68,4 +68,105 @@ testAgainstRunningApps({ withEnv: [appConfigs.envs.withEmailCodes] })('unsafeMet
...
+
+  test('user.update({ unsafeMetadata }) preserves replace semantics end-to-end', async ({ page, context }) => {
+    const { u, fakeUser, bapiUser } = await signUpAndGetUser({ page, context });
+
+    // Drive the deprecated path from the browser. The SDK should route
+    // metadata through PATCH /v1/me/metadata after computing a merge patch
+    // against the locally-cached value; the server-side outcome must match
+    // a true replace (the original `position` key is gone).
+    await page.evaluate(async () => {
+      await (window as any).Clerk.user.update({ unsafeMetadata: { city: 'Toronto' } });
```
wobsoriano: nit: we can remove the assertions here as we have correct types for `window.Clerk` already https://github.com/clerk/javascript/blob/a043a15cdd87a49de3fcf49d5ef929907753983e/integration/types.d.ts#L3,L7

packages/clerk-js/src/utils/mergePatch.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+    const aKeys = Object.keys(a as Record<string, unknown>);
+    const bKeys = Object.keys(b as Record<string, unknown>);
+    if (aKeys.length !== bKeys.length) {
+      return false;
+    }
+    return aKeys.every(k => deepEqual((a as Record<string, unknown>)[k], (b as Record<string, unknown>)[k]));
+  }
+  return false;
+}
```
wobsoriano: I missed this but we have a `dequal` package installed in clerk-js so you can just import that 👍🏼

packages/clerk-js/src/utils/mergePatch.ts
```diff
@@ -0,0 +1,84 @@
...
+      patch[key] = null;
+    }
+  }
+
+  return patch;
+}
+
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
```
wobsoriano: ❓  Should this check only match real POJOs?  Before this PR, `user.update({ unsafeMetadata })` sent metadata through `JSON.stringify`, so values like `Date` were treated as atomic values.

I think this Date, Map or class interfaces will pass this check and produce an empty sub-patch and skip the update 🤔

packages/clerk-js/src/core/resources/User.ts
```diff
@@ -235,9 +237,50 @@ export class User extends BaseResource implements UserResource {
...
+    // When `hasRest` is true the PATCH /me below refreshes `this` in place
+    // via `fromJSON` before we read `this.unsafeMetadata` for the diff.
+    // When it's false (only-metadata update), no upstream call refreshes the cache,
+    // so we `reload()` explicitly.
+    if (hasRest) {
+      await this._basePatch({
+        body: normalizeUnsafeMetadata(rest as UpdateUserParams),
+      });
+    } else {
+      await this.reload();
```
wobsoriano, line 262: thanks for adding! added latency, but its the cost of compatibility 👍🏼

wobsoriano (changes requested):
a couple more comments, but this is looking good! Thank you for resolving the other requests

.changeset/route-unsafe-metadata-to-merge-endpoint.md
brunol95: done

.changeset/deprecate-update-user-metadata.md
brunol95: done

packages/clerk-js/src/utils/mergePatch.ts
brunol95: thanks!

packages/clerk-js/src/utils/mergePatch.ts
brunol95: I updated the function to calculate the prototype of the object. This way we reject Dates, Map and class interfaces from the check. Added tests to verify behavior.thanks!

wobsoriano (approved):
Thanks for addressing the comments!

I think the main tradeoff I see is that mixed metadata + non-metadata updates are now split into two requests, so they are no longer atomic if the second request fails.

Approving!

brunol95:
!snapshot

wobsoriano:
!snapshot

PR #8597, opened 2026-05-19
fix(ui): Handle instant password recovery when strict enumeration protection is enabled

wobsoriano:
## Description

When a password manager autofills both email and password on the sign-in start screen and the credentials are wrong, the recovery that silently retries with just the identifier and navigates to factor-one (where "Forgot password?" is visible) never triggered if strict user enumeration protection was enabled.

This is because enumeration protection causes the backend to return `form_password_or_identifier_incorrect` instead of `form_password_incorrect`, and the former was missing from the `instantPasswordError` check.

Screenshot 2026-05-19 at 2 08 31 PM

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/tidy-plums-exist.md
+++ b/.changeset/tidy-plums-exist.md
@@ -1,0 +1,6 @@
+---
+"@clerk/shared": patch
+"@clerk/ui": patch
+---
+
+Handle instant password recovery when strict enumeration protection is enabled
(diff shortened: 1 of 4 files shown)


wobsoriano:
!snapshot

PR #8598, opened 2026-05-20
chore(shared): Remove internal tag from publicly re-exported error helpers

wobsoriano:
## Description

- Removes the `@internal` JSDoc tag from error helpers in `@clerk/shared` that are explicitly re-exported for public use by SDK packages (`isKnownError`, `isReverificationCancelledError`, `isMetamaskError`, `isUserLockedError`, `isEmailLinkError`)
- Helpers that are genuinely internal and not re-exported retain the `@internal` tag

The `@internal` tag was originally applied because these helpers live in `@clerk/shared`, which users shouldn't import from directly. However, since they are explicitly re-exported from the SDK packages, the tag was misleading.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/orange-pandas-burn.md
+++ b/.changeset/orange-pandas-burn.md
@@ -1,0 +1,5 @@
+---
+"@clerk/shared": patch
+---
+
+Remove internal tag from publicly re-exported error helpers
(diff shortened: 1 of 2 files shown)


jacekradko (approved):
Makes sense for these not to be internal

PR #8601, opened 2026-05-20
chore(ui,shared,localizations): UX improvements for `<ConfigureSSO />`

LauraBeatris:
## Description

- Moves verify domain step as the first one, to clarify which domain the user is setting up the enterprise connection
- Update select provider step, to only include header title/description 
- Add UI descriptors to each step
- Improve test step structure (new empty CTA, open test URL button)

CleanShot 2026-05-20 at 19 24 55

CleanShot 2026-05-20 at 19 24 26

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/components/ConfigureSSO/ConfigureSSONavbar.tsx
```diff
@@ -100,3 +101,31 @@ const OrganizationSidebarSubtitle = (): JSX.Element | null => {
...
+  }
+
+  return (
+    <Col
+      sx={t => ({
+        gap: t.space.$6,
+        padding: `${t.space.$none} ${t.space.$3}`,
+      })}
+    >
+      <Hr sx={{ backgroundColor: 'borderAlpha100' }} />
```
alexcarpenter: this needs to pull from the theme value right?

packages/ui/src/components/ConfigureSSO/ConfigureSSONavbar.tsx
```diff
@@ -77,6 +77,7 @@ export const ConfigureSSONavbar = ({ children, contentRef }: ConfigureSSONavbarP
         title={localizationKeys('configureSSO.navbar.title')}
         routes={[]}
         contentRef={contentRef}
+        footer={<PrimaryEmailDomainFooter />}
```
alexcarpenter: we now have three slots that all render in the same spot. Do we need to introduce another? footer also feels like the wrong name here, as this floats towards the top of the sidebar

```tsx
{header}
{items}
{footer}
```

LauraBeatris: I'll check with Tu and Jeremy again to see if we can find out another option to tackle this domain UX issue, but I'll rever the changes from the sidebar for the time being

LauraBeatris: We decided to drop that domain from the navbar, since we're going to introduce domain TXT verification step in the next weeks https://github.com/clerk/javascript/pull/8601/commits/96a227697604d95de8ad79c1416c0292a8737632

packages/ui/src/components/ConfigureSSO/steps/ConfigureStep.tsx
```diff
@@ -450,76 +418,106 @@ export const ConfigureAttributesSubStep = (): JSX.Element => {
...
 };
 
-const OktaConfigureAttributesStepContent = (): JSX.Element => {
-  return (
-    <>
-      <Text
-        as='p'
-        colorScheme='secondary'
-        localizationKey={localizationKeys('configureSSO.configureStep.attributeMapping.paragraph')}
-      />
```
iagodahlem, line 460: Noticed this paragraph here was removed, but it is still on the new designs: 

Image

packages/ui/src/components/ConfigureSSO/steps/ConfigureStep.tsx
```diff
@@ -450,76 +418,106 @@ export const ConfigureAttributesSubStep = (): JSX.Element => {
...
-              paddingInlineStart: theme.space.$5,
-              listStyleType: '"- "',
-            })}
-          >
-            {ATTRIBUTE_PAIRS.map(pair => (
+        </Th>
+        <Th>
+          <Text
+            sx={theme => ({ fontSize: theme.fontSizes.$xs })}
+            localizationKey={config.columns.second}
```
iagodahlem, line 474: In the designs, this column is now "Expression", currently as Value, but I think we can update it later, not sure if it is intended.

packages/ui/src/components/ConfigureSSO/steps/ConfigureStep.tsx
```diff
@@ -450,76 +418,106 @@ export const ConfigureAttributesSubStep = (): JSX.Element => {
...
-              </Text>
-            ))}
-          </Col>
-        </Text>
-      </Col>
-    </>
-  );
-};
+                as='span'
+                colorScheme={config.monoFirst ? undefined : 'secondary'}
```
iagodahlem, line 491: Noticed this column is always secondary even with mono in the designs, adding as a note to check what is the right one, we can fix later too.

iagodahlem (approved):

packages/ui/src/components/ConfigureSSO/steps/ConfigureStep.tsx
LauraBeatris: I'll refactor this one on a next PR where I'll introduce better support for UI per IdP!

packages/ui/src/components/ConfigureSSO/steps/ConfigureStep.tsx
LauraBeatris: I recognize this abstraction is bad, so I'll likely remove it on a next PR 

Reason why, custom SAML attribute mapping table doesn't have a fixed attribute name to use mono, as it could be anything on the custom provider, so we use labels.


Okta uses the actual attribute name on the first column, therefore mono.


CleanShot 2026-05-25 at 16 41 00


CleanShot 2026-05-25 at 16 41 36

PR #8607, opened 2026-05-20
fix(expo): Fix expo-auth-session import leak and env var detection

chriscanin:
## Summary

This PR fixes an optional-dependency leak in `@clerk/expo`'s `ClerkProvider`.

Previously, `ClerkProvider` synchronously referenced `expo-web-browser` behind an `isWeb()` runtime check in order to call `maybeCompleteAuthSession()` for OAuth/SSO flows on web. In Expo/Metro, that pattern can still cause `expo-web-browser` to be picked up during native bundling even though it is only needed on web and is declared as an optional peer dependency.

This change moves that logic into a platform-specific helper:
- `maybeCompleteAuthSession.web.ts` contains the web implementation
- `maybeCompleteAuthSession.ts` is a native no-op

That keeps web behavior unchanged while ensuring native bundles no longer reference `expo-web-browser` through `ClerkProvider`.

Note: this is distinct from the `useSSO` / `expo-auth-session` issue fixed separately https://github.com/clerk/javascript/pull/8720. This PR only addresses the `ClerkProvider` `expo-web-browser` leak.


wobsoriano (approved):
tested this locally 👍🏼

wobsoriano:
!snapshot

PR #8628, opened 2026-05-22
fix(astro): avoid invalid ui constructor in injected scripts

jescalan:
## Summary

Fix two Astro integration initialization paths that could pass an invalid `ui.ClerkUI` value into `clerk.load()`:

```ts
import clerk from "@clerk/astro";
import { ui } from "@clerk/ui";

clerk({ ui });
```

The bundled UI path was serializing injected script params with `JSON.stringify(...)`, which strips `ui.ClerkUI` because it is a constructor. The client-side initialization then saw `options.ui`, skipped CDN UI loading, and passed an object without a usable `ClerkUI` constructor into `clerk.load()`.

The `prefetchUI: false` path had a related issue: Astro returned `Promise<undefined>` from the UI resolver, then passed that promise as `ui.ClerkUI`. `clerk-js` treats the promise object as truthy and later attempts to instantiate the resolved `undefined` value.

This PR now:

- imports the live bundled `ui` object from `@clerk/ui` in injected snippets when the Astro integration is configured with `ui`
- returns plain `undefined`, not `Promise<undefined>`, for no-UI paths like `prefetchUI: false` and UI marker objects without constructors
- keeps default/non-bundled snippets on the previous serialized behavior

## Testing

- `pnpm --filter @clerk/astro test -- src/internal/__tests__/create-clerk-instance.test.ts src/integration/__tests__/snippets.test.ts`
- `pnpm --filter @clerk/astro lint` (passes with existing warnings)
- `pnpm exec prettier --check packages/astro/src/internal/create-clerk-instance.ts packages/astro/src/internal/__tests__/create-clerk-instance.test.ts packages/astro/src/integration/snippets.ts packages/astro/src/integration/__tests__/snippets.test.ts .changeset/astro-bundled-ui-serialization.md`
- `git diff --check`

Note: `pnpm --filter @clerk/astro format:check` still fails on pre-existing copied `components/**/*.astro` formatting issues, but it does not flag the files touched in this PR.


jescalan:
@codex review

MrAlex94:
Any chance this could be pulled in soon?

wobsoriano (approved):
Looks good, great catch!

The only failure I see is if the injected `import { ui } from "@clerk/ui"` can't resolve from the user's project, but that won't happen because passing ui requires importing it from `@clerk/ui` in `astro.config` in the first place, so the package is always installed

PR #8632, opened 2026-05-22
chore(repo): pin GitHub Actions to SHAs and drop persisted tokens

jacekradko:
This is the repo-wide follow-up to a security finding that flagged a single workflow for using mutable action tags and leaving `GITHUB_TOKEN` persisted in `.git/config` after `actions/checkout`. Fixing one file would have left the broader posture unchanged, so this PR addresses the whole `.github/` surface in one pass.

Three coordinated pieces:

1. Every third-party action under `.github/workflows/**` and `.github/actions/**` is now pinned to an immutable commit SHA, with the original tag preserved as a `# v4` comment. 73 `uses:` references across 12 workflows and 4 composite actions.
2. Renovate's `includePaths` now covers `.github/**`, and a new package rule applies `pinDigests: true` plus a `GitHub Actions` group, so the SHAs stay current automatically and won't quietly rot. `renovate.json5` is regenerated from `scripts/renovate-config-generator.mjs`; do not hand-edit.
3. `persist-credentials: false` is set on every `actions/checkout` that doesn't need the persisted token (20 sites). The one site that does, `release.yml` line 45, is intentionally left alone: it sets `token:` explicitly so `changesets/action` can push the Version PR back to the repo. The existing comment block above that `token:` input already documents why.

`step-security/harden-runner` would be the natural next layer but needs an audit-mode rollout before enforcement, deferred to a separate PR.


dominic-clerk (approved):
Thank you!

PR #8637, opened 2026-05-22
refactor(repo): Inline extract-methods + tighten parameter rendering

alexisintech:
Stacked on top of #8276 — review that one first.

Five independent refactors of `.typedoc/` (each its own commit) plus snapshot tests to lock the contract for future changes.

## Summary

- **Run `extract-methods` inside the markdown render pass** — folds `extract-methods.mjs` into a `MarkdownPageEvent.END` listener that queues a `preWriteAsyncJob`. Removes the second `Application.bootstrapWithPlugins` + `convert()` and the `prepare-markdown-renderer.mjs` workaround it required.
- **Consolidate unwrap + slug helpers** — three unwrap-optional helpers + a peel-all while-loop collapse into a single `unwrapOptional(t, { deep })`. The two `toKebabCase` flavors (one for method filenames, one for page URLs) move into `.typedoc/slug.mjs` — they intentionally differ on acronyms, so the module exports both rather than unifying.
- **Flatten params with optionality-aware separator** — `clerkParametersTable.flattenParams` now joins `parent.child` with `?.` when the parent is optional and `.` when required. Lets us delete `formatMethodParametersTable` (the regex post-process that did a blanket `.` → `?.` rename plus a no-longer-needed reorder).
- **Inline markdown helpers, drop `node_modules` imports** — `custom-theme.mjs` and `extract-methods.mjs` were reaching into `node_modules/typedoc-plugin-markdown/dist/libs/...` for `backTicks`, `heading`, `htmlTable`, `table`, `removeLineBreaks`, and the `TypeDeclarationVisibility` enum — none of which the plugin exports publicly, so that coupling broke on any dependency update. The five utilities (plus their `escapeChars` / `formatTableCell` dependencies) are now inlined in `.typedoc/markdown-helpers.mjs`, and the enum check is replaced with the underlying string literal (`'compact'`).
- **Consolidate property merge across intersection/union/generic arms** — three branches of `resolveDeclarationWithObjectMembers` duplicated the same "filter to `Property`, dedupe by name, sort" pattern, each wrapping the result in a synthetic holder (`__intersectionMerged` / `__unionMerged` / `__referenceMerged`) just so callers could read `.children`. Extracts `mergePropertyArms(arms, options)`; union arms pass `{ skipNever: true, pickBetter: true }` to drop `prop?: never` discriminators. The resolver now returns `DeclarationReflection[]` directly, and all four callers consume the array.

Also adds the snapshot dir to `.prettierignore`: lint-staged was silently transforming the committed snapshots after `vitest run` passed, leaving them out of sync with the raw output that `extract-methods.mjs` writes.

## Output changes

Output is byte-identical to the pre-refactor baseline (`diff -r`, 695/695 files) for the first, second, fourth, and fifth commits. The third commit deliberately changes **8 files** to fix a latent inconsistency:

- `shared/clerk/methods/handle-email-link-verification.mdx`: parent `params` is required, children now render as `params.foo` (was `params?.foo`).
- `backend/get-auth-fn.mdx`, `backend/get-auth-fn-no-request.mdx`, `nextjs/current-user.mdx`, `shared/api-keys-namespace.mdx`, `shared/custom-navigation.mdx`, `shared/get-memberships.mdx`, `shared/on-event-listener.mdx`: parent is optional, children now render as `parent?.foo` (was `parent.foo`).

All eight match the rule "separator reflects how the property would be accessed at runtime."

## Test plan

- [ ] `pnpm typedoc:generate` succeeds with no errors.
- [ ] `cd .typedoc && vitest run` passes (11 new file-snapshot tests + the existing file-structure tests, 16 total).
- [ ] Spot-check the 8 changed files in the rendered docs preview to confirm they read correctly.


manovotny:
Pushed some changes directly in 3f45eb58e.

- Hoisted `unwrapOptional` into `.typedoc/type-utils.mjs` and dropped the duplicate `unwrapOptionalType` from `custom-theme.mjs`.
- `formatTypeScriptSignature` now uses the bare last segment for dotted method names (`function sendCode(...)` instead of `function emailCode.sendCode(...)`, which wasn't valid TS). Heading still shows the qualified name. Snapshot refreshed.
- Added a description after `// @ts-check` in `markdown-helpers.mjs` and `slug.mjs` to satisfy `@typescript-eslint/ban-ts-comment`.
- Added an empty changeset per `AGENTS.md`.

manovotny (approved):

alexisintech:
there's a changeset in the branch this is getting merged into so I'm going to remove the changeset here! @manovotny

PR #8642, opened 2026-05-26
feat(cli-auth): add @clerk/cli-auth package

nicolas-angelo:
## Summary

Adds `@clerk/cli-auth` to the monorepo — a package for adding Clerk authentication to Node.js CLIs and the backend routes those CLIs talk to.

Ships two entry points:

- **`@clerk/cli-auth`** — CLI runtime: browser sign-in via PKCE + localhost callback, token storage, refresh, revocation, credential resolution.
- **`@clerk/cli-auth/server`** — backend route handlers for verifying CLI bearers via `clerk.authenticateRequest`. Drop into any framework speaking Web `Request`/`Response` (Next.js App Router, Hono, Cloudflare Workers, Bun, Deno).

### `ClerkCliAuth`
```ts
import { ClerkCliAuth } from '@clerk/cli-auth';

const auth = new ClerkCliAuth({
  clientId: process.env.CLERK_OAUTH_CLIENT_ID!,
  issuer: process.env.CLERK_ISSUER!,
  scopes: ['profile', 'email', 'openid', 'offline_access'],
  storage: 'keychain', // or 'file', or 'memory'
  // Optional: enables auth.verifyToken() against your backend
  identityEndpoint: 'https://myapp.com/api/cli/identity',
  // Optional: the environment variable that resolveToken() reads from
  tokenEnvVar: 'MYAPP_API_KEY',
});
```

### Usage examples 
```ts
// Sign in: opens browser, runs localhost callback, exchanges code, stores tokens
const { tokens, user } = await auth.login();

// Get a valid access token (auto-refreshes within 30s of expiry)
const token = await auth.getAccessToken();

// Fresh whoami against /oauth/userinfo
const me = await auth.whoami();

// Resolve whichever token is available from auth
// source: 'arg' (from --token), 'env' (from tokenEnvVar), or 'oauth' (from login)
const { token, source } = await auth.resolveToken({ tokenFromArg: argv.token });

// Revoke at the issuer, then clear local credentials
await auth.logout();
// Local logout only (skip token revocation)
await auth.logout({ revoke: false })
```

## Identity
The SDK provides helper methods for use cases that require resolving identity from an auth token — for example, a `whoami` command that works with both OAuth and API key auth. All related utilities conform to the `Identity` interface, covering four of Clerk's subject types:

```ts
type Identity = UserIdentity | OrgIdentity | MachineIdentity | ScimIdentity;
// narrow w/ type guards
// isUserIdentity / isOrgIdentity / isMachineIdentity / isScimIdentity 
```

### `cliAuth`
Returns an instance with helper methods that can be used to resolve auth info and secure endpoints that your CLI communicates with.

```ts
// lib/clerk-cli.ts
import { cliAuth } from '@clerk/cli-auth/server';
import { clerkClient } from '@clerk/nextjs/server';

export const auth = cliAuth()
// or optionally pass in a pre-configured clerk client
export const auth = cliAuth({ client: clerkClient });
```

### Identity resolution handler (for tokens)
The CLI can fetch the authorized party's identity by calling a backend endpoint you host.

For convenience, the SDK exports `handle()`, which takes a `cliAuth` instance and returns a route handler that verifies the bearer token on the `Authorization` header, resolves the identity of it's subject and responds with a JSON `Identity` payload.

Drop into any Fetch API compatible framework (Next.js App Router, Hono, Cloudflare Workers, Bun, Deno).

```ts
// app/api/cli/identity/route.ts
import { handle } from '@clerk/cli-auth/server';
import { auth } from '@/lib/clerk-cli';

export const GET = handle({ auth, accepts: ['api_key', 'oauth_token'] });
```

Don't forget to point `identityEndpoint` (in your `ClerkCliAuth` configuration) at this URL to opt-in.

### Accepted credentials

| `accepts` value | Matches                                                          |
| --------------- | ---------------------------------------------------------------- |
| `'api_key'`     | `ak_*` Clerk API keys with user, org, **or** machine subjects    |
| `'oauth_token'` | `oat_*` opaque OAuth access tokens or JWTs   |
| `'any'`         | Either of the above                                              |

> The SDK's verification helpers do not support session JWTs and M2M tokens. To enable authorization for machine identities, mint an API key with a `mch_*` subject.

For other routes your CLI communicates with use `auth.verifyTokenFromRequest()` directly inside your handler:

```ts
// app/api/cli/projects/route.ts
import { NextResponse } from 'next/server';
import { auth } from '@/lib/clerk-cli';

export async function GET(request: Request) {
  const tokenInfo = await auth.verifyTokenFromRequest(request, {
    accepts: ['api_key', 'oauth_token'],
  });

  const projects = await db.projects.findMany({
    where: { ownerId: tokenInfo.subject },
  });

  return NextResponse.json({ projects });
}
```

> `verifyTokenFromRequest` throws if the header is missing, the token is invalid, or its type isn't accepted. 

The returned `tokenInfo` has subject, type, scopes, and claims — use any of them to authorize the request.

## Test plan

- [x] Unit tests pass — `pnpm test` (20 tests, runs offline)
- [x] Integration suite passes — `pnpm test:integration` (16 tests, requires `CLERK_SECRET_KEY` and `CLERK_PUBLISHABLE_KEY`; skips if unset)
- [x] Integration suite provisions throwaway Clerk resources (user, org, api keys, etc) and tears them all down in `afterAll`
- [x] `pnpm build`, `pnpm lint`, `pnpm lint:attw`, `pnpm lint:publint` success


packages/cli-auth/src/lib/credential-store.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  async get(key: string): Promise<string | null> {
+    try {
+      const keyring = await this.loadKeyring();
+      if (!keyring) {
+        return this.fallback.get(key);
+      }
+
+      const entry = new keyring.Entry(this.service, this.account(key));
+      return entry.getPassword() ?? (await this.fallback.get(key));
```
Railly, line 160: This fallback can return stale file credentials after keychain was the successful write path. `set()` only writes to keychain when keychain works, but `get()` falls back to file on null/failure, so an old fallback token can be resurrected. Can we either pick one active backend after probing, or mirror writes/deletes into the fallback so both stores stay coherent?

packages/cli-auth/src/lib/token-exchange.ts
```diff
@@ -0,0 +1,216 @@
...
+    tokenSet.expiresAt = Date.now() + data.expires_in * 1000;
+  }
+
+  return tokenSet;
+}
+
+async function requestTokens(issuer: string, body: URLSearchParams): Promise<TokenSet> {
+  let response: Response;
+  try {
+    response = await fetch(endpoint(issuer, '/oauth/token'), {
```
Railly: This package now has several raw `fetch` calls without a timeout, here and in `verify-api-key.ts`. The callback server timeout does not cover token exchange, refresh, revoke, userinfo, or API-key verification, so CLI commands can hang indefinitely. Can we route these through one small internal HTTP helper with an `AbortController`, shared body parsing, and consistent error mapping?

packages/cli-auth/src/clerk-cli-auth.ts
```diff
@@ -0,0 +1,273 @@
...
+      ...refreshed,
+      refreshToken: refreshed.refreshToken ?? tokens.refreshToken,
+    };
+    await this.setJson('tokens', nextTokens);
+    return nextTokens.accessToken;
+  }
+
+  async whoami(): Promise<UserInfo | null> {
+    const cachedUser = await this.getJson<UserInfo>('user');
+    if (cachedUser) {
```
Railly: `whoami()` returns cached user data before checking whether the stored access token is still fresh or refreshable. That makes identity freshness ambiguous after expiry, revocation, or scope/org changes.

nicolas-angelo:
Going to close this and split it into two PRs for better seperation of core CLI lib from additional utilities.
See:

- PR #8709 (base) — core CLI SDK
- PR #8710 (consumer backend helpers for communication w/ CLI)

PR #8643, opened 2026-05-26
feat(clerk-js,shared): Backport Session Minter SDK changes to core-2

nikosdouvlis:
## Why

Session Minter (edge token minting) shipped to clerk-js on `main` between March and May 2026 across five small PRs. Until now, only the touch-intent change ([#8135](https://github.com/clerk/javascript/pull/8135)) was backported to `release/core-2`. The other five SDK changes have to land on core-2 too so v5 consumers can opt into edge-minted tokens once their instance flag is flipped on.

This is a behavior-preserving backport: every code path is gated behind `auth_config.session_minter`, which defaults to `false`. With the gate off, clerk-js v5 sends the same `/tokens` request body and runs the same retry path it does today.

## What changed

Five upstream PRs ported, one per commit:

- [#8107](https://github.com/clerk/javascript/pull/8107) - add optional `oiat` field to `JwtHeader`
- [#8097](https://github.com/clerk/javascript/pull/8097) - monotonic cross-tab token replacement via `pickFreshestJwt`, so a stale edge-minted broadcast cannot clobber a fresher cached token
- [#8105](https://github.com/clerk/javascript/pull/8105) - send the previous session token in the `/tokens` POST body when `auth_config.session_minter` is true
- [#8108](https://github.com/clerk/javascript/pull/8108) - skip the `missing_expired_token` retry when Session Minter is enabled (the token is now in the body, so the retry-with-expired-token fallback is unnecessary). Legacy retry preserved when the flag is off.
- [#8106](https://github.com/clerk/javascript/pull/8106) - send `forceOrigin=true` in the body when `skipCache` is true, so FAPI routes to origin instead of the Session Minter edge

A sixth commit cleans up a review-pass finding: rather than introduce a new `__internal_environment` getter on Clerk (which exists on `main` but not on core-2), the code now reads the existing `__unstable__environment` getter that has been on core-2 since v3. Same commit also backports the Session test coverage from #8105 and #8106 that wasn't worth porting until the env-getter question was settled.

## Packages affected

- `@clerk/clerk-js`: AuthConfig gains a `sessionMinter` field, Session sends the new body fields when the flag is on, tokenCache compares broadcast tokens with `pickFreshestJwt`. All behavior is gated.
- `@clerk/shared`: `AuthConfigResource.sessionMinter`, `AuthConfigJSON.session_minter?`, `JwtHeader.oiat?`. All additive.

## Test plan

- [x] `pnpm vitest run src/core` in clerk-js: 380/380 pass
- [x] `pnpm test` in shared: 641/641 pass
- [x] New unit tests in `tokenFreshness.test.ts` cover the broadcast-side guard's edge cases (legacy tokens, equal oiat, tie-break)
- [x] New `Session.test.ts` blocks assert the `/tokens` body shape for both `sessionMinter: true` and `sessionMinter: false`
- [ ] Manual: smoke test in a v5 host app against a staging instance with the flag on

### DIFF
--- a/.changeset/remove-expired-token-retry.md
+++ b/.changeset/remove-expired-token-retry.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Skip `expired_token` retry flow when Session Minter is enabled. When `sessionMinter` is on, the token is sent in the POST body, so the retry-with-expired-token fallback is unnecessary. The retry flow is preserved for non-Session Minter mode.
--- a/.changeset/session-minter-force-origin.md
+++ b/.changeset/session-minter-force-origin.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Send `force_origin=true` body param on `/tokens` requests when `skipCache` is true, so FAPI Proxy routes to origin instead of Session Minter.
--- a/.changeset/session-minter-monotonic-guard.md
+++ b/.changeset/session-minter-monotonic-guard.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Add monotonic token replacement based on `oiat` to prevent edge-minted tokens with stale claims from overwriting fresher DB-minted tokens in multi-tab scenarios.
--- a/.changeset/session-minter-oiat-type.md
+++ b/.changeset/session-minter-oiat-type.md
@@ -1,0 +1,5 @@
+---
+'@clerk/shared': patch
+---
+
+Add `oiat` (original_issued_at) field to `JwtHeader` type for Session Minter monotonic token freshness checks.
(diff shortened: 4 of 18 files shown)


nikosdouvlis:
The two failing checks (`Integration Tests (machine, chrome)` and `Integration Tests (machine, chrome, RQ)`) are pre-existing on `release/core-2`, not introduced by this PR.

[#8633](https://github.com/clerk/javascript/pull/8633), the most recently merged PR against `release/core-2` (a pure CI version-bump with no code changes), has the same two failures. The earlier PRs [#8321](https://github.com/clerk/javascript/pull/8321), [#8310](https://github.com/clerk/javascript/pull/8310), [#8239](https://github.com/clerk/javascript/pull/8239) all had these checks passing, so the M2M test regressed somewhere between then and now.

The failure is `m2m.test.ts:139` "authorizes M2M requests when sender machine has proper access to receiver machine" - it sends an M2M bearer token to a protected endpoint and expects 200, but gets 401. M2M tokens are a separate auth flow from user session tokens; nothing in this port touches the M2M code path.

Static analysis (including bundlewatch) is now green after I bumped the `clerk.headless*.js` limit from 67 KB to 68 KB.

PR #8649, opened 2026-05-26
fix(shared): handle banned/deactivated 403 as unauthenticated

jacekradko:
#8004 narrowed ClerkJS token/session sign-out from any 4xx to `isUnauthenticatedError`, but that helper only matched 401 and 422. Terminal 403s (`user_banned`, `user_deactivated`) stopped triggering sign-out.

This adds those two codes into the predicate. Generic 403 and 429 stay out of the sign-out path on purpose.

```ts
return status === 401 || status === 422 || (status === 403 && hasTerminalUserErrorCode);
```

Cross-checked against `clerk_go`: `apierror.UserBanned()` / `UserDeactivated()` return `http.StatusForbidden` with codes `user_banned` / `user_deactivated`, serialized as `{ errors: [{ code, ... }] }`, and are emitted on the session-token / handshake path in `api/shared/sessions/service.go`. So the predicate matches the API contract 1:1.

Test additions cover both directions: terminal codes return true, other 403 codes (e.g. `not_allowed_access`) still return false.


nikosdouvlis (approved):
lgtm, restores the 403 banned/deactivated sign-out behavior that we lost in #8004

PR #8651, opened 2026-05-26
chore(ui): Layer architecture for configure steps per IdP and protocol

LauraBeatris:
## Description

Introduces new architecture for adding configuration steps per IdP and protocol, making it more flexible to add Microsoft Entra + Google Workspace for SAML, and later OpenID connect.

Now, each protocol has it's folder within the provider types, and each provider should have it's inner steps created within a module, which allows for flexibility when composing the UI.

#### How to add a new SAML provider

1. Update the `ProviderType` union 
2. Once the union gets updated, then it'll require to update the step mapper within `ConfigureStep/index.tsx` 
3. Implement the steps component for that identity provider according to it's instructions 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/components/ConfigureSSO/steps/ConfigureStep/saml/shared/IdentityProviderMetadataForm.tsx
```diff
@@ -0,0 +1,427 @@
...
+ *   - Mode defaults to `manual` if the connection already has any IdP
+ *     config persisted, otherwise `metadataUrl`.
+ *   - On submit, the connection is updated via `useReverification` and
+ *     the wizard advances; field/card errors are surfaced from the API.
+ *
+ * All copy is provided by the caller as resolved `LocalizationKey`s so
+ * this primitive stays decoupled from any specific localization
+ * namespace.
+ */
+export const IdentityProviderMetadataForm = ({
```
iagodahlem, line 55: I feel like this abstraction here should not define the entire step, and really just the segmented control + forms, and also avoid using the high level hooks like `useConfigureSSO` and `useWizard` directly in the shared components, I think we should cross a boundary where only the actual steps should use those hooks.

iagodahlem (approved):

PR #8654, opened 2026-05-26
feat(ui): Expose profile sub components

alexcarpenter:
## Summary

Ships a composable `UserProfile` / `OrganizationProfile` from `@clerk/ui/experimental`. Consumers render individual panels and sections (Account, Security, Members, Billing, …) inline in their own app instead of inside Clerk's modal/page flow.

The API is a set of **flat, top-level named exports** — not a namespace object like `UserProfile.Account`. Each named export of a `'use client'` module becomes its own React Server Component client reference, so consumers can render these directly in a Server Component tree without adding their own `'use client'` boundary. (Property access on a namespace object would not survive the RSC boundary.)

```tsx
import {
  UserProfileProvider,
  UserProfileAccountPanel,
  UserProfileSecurityPanel,
  UserProfileEmailSection,
  UserProfilePasswordSection,
} from '@clerk/ui/experimental';

export default function Page() {
  return (
    <UserProfileProvider>
      {/* A panel with no children renders the full standard page… */}
      <UserProfileAccountPanel />
      {/* …or you compose the sub-sections yourself. */}
      <UserProfileSecurityPanel>
        <UserProfilePasswordSection />
      </UserProfileSecurityPanel>
    </UserProfileProvider>
  );
}
```

## Design decisions worth scrutiny

| Area | Decision | Why |
|---|---|---|
| Flat named exports | Every panel/section is a top-level named export from `@clerk/ui/experimental`, not a property on a namespace object. `flat-exports.test.ts` pins the surface and asserts the namespace objects are `undefined`. | RSC client references (see Summary). |
| Provider shell (`ProfileProviderShell.tsx`) | Collapse `LazyProviders` + `LazyComponentRenderer` / `LazyModalRenderer` (portal path) into one shell. `ClerkContextProvider` omitted. | Composed mounts bypass the portal. Consumer's `<ClerkProvider>` already supplies `clerk` via `useClerk()`. |
| Emotion cache | Shared per clerk instance via `WeakMap` (`internal/styleCacheStore.ts`). | Sibling composed roots would otherwise create two `cl-internal` caches and double-insert rules. Keyed on the clerk instance object, so it auto-GCs on teardown. |
| `moduleManager` | clerk-js exposes `__internal_moduleManager` getter on `Clerk` (`clerk.ts:288`); IsomorphicClerk forwards it (`isomorphicClerk.ts:298`). Composed shell reads `clerk.__internal_moduleManager` directly. | Makes `<ClerkProvider ui={ui} />` optional for composed (Web3 works out of box). Plain property access is the only channel that crosses the bundle boundary — clerk-js loads standalone with its own inlined `@clerk/shared`, so `@clerk/react`/`@clerk/ui` reading `@clerk/shared` from `node_modules` see a different copy. |
| Options threading | Pull `localization` + `supportEmail` + `nonce` from `clerk.__internal_getOption`, pass via `OptionsProvider` / `SharedStyleCacheProvider`. | Only keys reachable from the composed subtree (`makeLocalizable`, `parseLocalization`, `useSupportEmail`, Emotion CSP nonce). `nonce` needs an `as any` cast (typed on `IsomorphicClerkOptions`, not `ClerkOptions`). |
| `cssLayerName` | Apply `extractCssLayerNameFromAppearance` on `globalAppearance` before reading. | Matches portal path normalization (`Components.tsx:209`). Without it, nested `appearance.theme.cssLayerName` is ignored. |
| Stub router (`stubRouter.ts`) | `navigate` / `baseNavigate` / `resolve` are real and SSR-safe. `currentPath` is a static empty string (no `popstate` listener). `matches()` / `refresh()` / `getMatchData()` throw in dev, no-op in prod. | Composed sections render outside the AIO Route/Switch tree. Dev throw catches accidental AIO API use, prod no-op keeps consumer apps from crashing on the same call. Contracts pinned by `__tests__/stub-limitations.test.ts`. |
| Dual-mode panels | A panel with no children passes through to the full AIO page (e.g. `UserProfileAccountPanel` → `AccountPage`). With children, sub-sections render directly under `PageContext` + `CardStateProvider`. | Lets consumers either drop in a section as-is or compose sub-parts. Passthrough only breaks if a sub-component calls `matches()`/`refresh()`/`getMatchData()` (dev throw, prod no-op). |
| Security panel | `OrganizationProfileSecurityPanel` renders the whole standard `OrganizationSecurityPage` (SSO overview + configuration wizard). It takes no composable children — the security tab is not section-based, unlike General. There is no standalone ConfigureSSO composable. | Mirrors what `<OrganizationProfile />` shows on its security route. |
| Section dedupe | `createSection(name, Component)` factory + `PageContext` guard via `useRequirePage`. | Each leaf is a one-liner instead of repeating wrapper boilerplate. |

## Suggested review order

1. `packages/ui/src/composed/ProfileProviderShell.tsx`: the architectural heart.
2. `packages/clerk-js/src/core/clerk.ts` (`__internal_moduleManager` getter, `clerk.ts:288`) + `packages/react/src/isomorphicClerk.ts` (forwards to the wrapper, `:298`): confirm the cross-bundle wiring and the ABI tradeoff.
3. `packages/ui/src/composed/stubRouter.ts` + `__tests__/stub-limitations.test.ts`: router contract and dev/prod asymmetry.
4. `packages/ui/src/composed/{UserProfile,OrganizationProfile}/*Provider.tsx`: section wiring + `__internal_environment` cast.
5. `packages/ui/src/composed/__tests__/`: parity, wiring, style-cache, flat-export, and tree-shaking tests pin the design.
6. `integration/tests/composed-components.test.ts`: e2e parity — the composed exports run the same trusted sign-in / profile flows as the standard components against a real clerk-js runtime.

## `__internal_moduleManager` ABI tradeoff

New getter on clerk-js's `Clerk` class. Once shipped, removing it breaks every pinned SDK version that reads it.

Why accept the cost:

- IsomorphicClerk already proxies every other piece of clerk-js state through public-ish properties. One more is consistent.
- `__internal_` prefix signals "may change between majors" — same contract as the rest of `__internal_*` on Clerk.
- Only channel that reliably crosses the bundle boundary.

Alternatives considered and rejected:

- **`@clerk/shared` WeakMap registry** (a prior design, now deleted): doesn't cross the bundle boundary. clerk-js inlines its own `@clerk/shared`, so the WeakMap clerk-js writes isn't the one `@clerk/react`/`@clerk/ui` read. Under version skew / no-dedupe the registry splits and dynamic-import features silently fall back to the rejecting manager.
- **`Symbol.for(...)`-keyed `globalThis` WeakMap**: crosses bundles, but hides the contract in an unnamed channel. Action-at-a-distance, lifecycle issues with test workers and HMR, still on the hook to keep the symbol stable forever — same cost, less inspectable.

Resolves PD-56

## Test plan

- [ ] CI green
- [ ] Smoke test in a Next.js sandbox app without `<ClerkProvider ui={ui} />`: confirm Web3 Add-Coinbase flow resolves the SDK
- [ ] Verify a `UserProfileProvider` and an `OrganizationProfileProvider` sibling-mounted on one page: single `cl-internal` cache, localized strings render
- [x] e2e parity in `integration/tests/composed-components.test.ts`: the composed UserProfile / OrganizationProfile exports pass the same trusted flows as the standard components (account edits, security, org general/members/security)


alexcarpenter:
!snapshot

alexcarpenter:
!snapshot

alexcarpenter:
!snapshot

alexcarpenter:
!snapshot

alexcarpenter:
!snapshot

packages/ui/src/composed/moduleManagerStore.ts
```diff
@@ -0,0 +1,11 @@
+import type { ModuleManager } from '@clerk/shared/moduleManager';
+
+let storedModuleManager: ModuleManager | undefined;
+
+export function setModuleManager(mm: ModuleManager): void {
```
alexcarpenter: Module-level store so composed components can access the `ModuleManager` that `ClerkUI` receives at mount time. Composed components render outside the normal component tree so they cant inherit it through context from the mount call — `ClerkUI.init()` calls `setModuleManager()` and providers read it back via `getModuleManager()`.

packages/ui/src/composed/stubRouter.ts
```diff
@@ -0,0 +1,30 @@
+import type { RouteContextValue } from '../router/RouteContext';
+
+export function createComposedRouter(clerkNavigate: (to: string) => Promise<unknown> | void): RouteContextValue {
```
alexcarpenter: Composed components render individual pages, not full routed flows — but many child components still call `useRouter()` internally. This satisfies that contract with a minimal implementation that delegates to `clerk.navigate` (or `window.location.assign` for the static fallback). No actual routing; just enough to prevent crashes from missing context.

packages/ui/src/styledSystem/StyleCacheProvider.tsx
```diff
@@ -15,6 +13,7 @@ type StyleCacheProviderProps = React.PropsWithChildren<{
 
 export const StyleCacheProvider = (props: StyleCacheProviderProps) => {
   const cache = useMemo(() => {
+    const el = typeof document !== 'undefined' ? document.querySelector('style#cl-style-insertion-point') : null;
```
alexcarpenter: Moved `document.querySelector` inside the component body. Top-level call ran at import time which breaks SSR / non-browser environments. Composed components may be imported in SSR-capable frameworks so this needs to be safe.

packages/ui/src/primitives/Spinner.tsx
```diff
@@ -6,6 +6,7 @@ const { size, thickness, speed } = createCssVariables('speed', 'size', 'thicknes
 const { applyVariants, filterProps } = createVariants(theme => {
   return {
     base: {
+      boxSizing: 'border-box',
```
alexcarpenter, line 9: Fixing a sizing issue uncovered where the border rendered outside the size defined causing layout shifts.

packages/ui/src/hooks/useSafeState.ts
```diff
@@ -13,6 +13,7 @@ export function useSafeState<S>(initialState?: S | (() => S)) {
   const isMountedRef = React.useRef(true);
 
   React.useEffect(() => {
+    isMountedRef.current = true;
```
alexcarpenter, line 16: Ref initializes as `true` but under StrictMode the cleanup runs between double-invocation, setting it to `false` — without this line the second effect never resets it, so `setState` becomes a permanent no-op.

jacekradko:
I like this from a directional standpoint and I agree that exposing subcomponents makes sense for flexibility and customization. At a quick glance, though, this seems to duplicate/add another Profile component stack that then needs extra context plumbed into it to work. That expands our public/internal contract and likely increases the chance of drift and defects over time. Could we instead look at extracting the shared parts from the current Profile components so the existing implementation and the new subcomponent API compose the same internals differently? What do you think? Happy to collab on this if you want or drop a high level proposal for internals

alexcarpenter:
@jacekradko here is an update with shared parts targeting this branch https://github.com/clerk/javascript/pull/8714

packages/ui/src/composed/OrganizationProfile/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { OrganizationProfileProvider } from './OrganizationProfileProvider';
+import {
+  GeneralDeleteOrganization,
+  GeneralLeaveOrganization,
+  GeneralOrganizationProfile,
+  GeneralVerifiedDomains,
+} from './sectionWrappers';
+import { ConfigureSSO } from './ConfigureSSO';
+
+export const OrganizationProfile = Object.assign(OrganizationProfileProvider, {
```
Ephem: Same comment as here: https://github.com/clerk/javascript/pull/8474#discussion_r3195843656

These _are_ meant to be publicly exported, so do we optimize for treeshaking early even though that's annoying for compound components?

packages/ui/src/composed/OrganizationProfile/sectionWrappers.tsx
```diff
@@ -0,0 +1,45 @@
...
+    }
+    return false;
+  }
+  return true;
+}
+
+export function GeneralOrganizationProfile(): ReactNode {
+  if (!useRequirePage('GeneralOrganizationProfile')) return null;
+  return <OrganizationProfileSection />;
+}
```
Ephem: Incredibly niche thing to zoom in on, not important at all in the grand scheme, you should almost definitely ignore this, but I wanted to share a neat trick in case it comes in handy some other time.

```tsx
if (!useRequirePage('GeneralOrganizationProfile')) return null;
```

Looks and is fine today, but it would be pretty simple to accidentally add a new hook call under that conditional, and now you can run into the rules of hooks errors (in this case, likely only due to hotloading in dev, if even that).

While this makes it clearer and easier to avoid a regression, it's annoying to type the thing out though:

```tsx
const hasRequiredContext = useRequirePage(...);
if (!hasRequiredContext) return null;
```

The neat trick: `use` is not a real hook, but can access context. It's allowed to call it _inside regular functions_ as long as they are called inside the render loop.

```tsx
function hasRequiredContext(componentName: string): boolean {
  const page = use(PageContext);
  if (!page) {
    if (typeof __DEV__ !== 'undefined' && __DEV__) {
      console.warn(`${componentName} must be used inside a page component (e.g. OrganizationProfile.General)`);
    }
    return false;
  }
  return true;
}
```

This is now not a hook, and you are allowed to call it conditionally. I think this is little known, but makes sense when you think about how you are allowed to call `use` conditionally as well.

packages/ui/src/composed/UserProfile/APIKeys.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+const APIKeysPage = lazy(() =>
+  import('../../components/UserProfile/APIKeysPage').then(m => ({
+    default: m.APIKeysPage,
+  })),
+);
+
+export const APIKeys = (): ReactNode => (
+  <CardStateProvider>
+    <Suspense fallback={''}>
```
Ephem: The `<Suspense>` fallbacks are interesting. 🤔 I think designing the loading and error experience should be front and center here.

At what levels do we handle it? (Page, section, component?) How do we make it good out of the box, while also allowing devs to tweak fallbacks?

This is tricky since we really only use Suspense for the `initialState` now, so maybe the thinking and right move is to just do '' for now? Maybe we _don't_ allow for tweaking loading states besides CSS in the first iteration?

It would be interesting to figure out the best possible long term DX here even if we don't implement it right away. I think that's Suspense+Error boundaries for everything, so the user can control it.

Just thinking out loud, but maybe that's also composable? Also maybe they want a top level loading state for their entire page/section of the page where Clerk is a small part, and that loading state lives far away from the Clerk code? They don't want the rest of the page to reveal first, show Clerk loading state, reveal Clerk, they want to wait for everything to load before revealing, including Clerk? Especially now that they can compose together their UI.

This of course means we can't render the components in their own React application, it needs to compose together with the host app, but I don't think this PR is doing the standalone render right?

packages/ui/src/composed/UserProfile/sectionWrappers.tsx
```diff
@@ -0,0 +1,164 @@
...
+export function AccountProfile(): ReactNode {
+  if (!useRequirePage('AccountProfile')) return null;
+  return <UserProfileSection />;
+}
+
+export function AccountUsername(): ReactNode {
+  if (!useRequirePage('AccountUsername')) return null;
+
+  const { attributes } = useEnvironment().userSettings;
+  const { immutableAttributes } = useUserProfileContext();
```
Ephem: Conditional hooks here. No big deal, but also easy to fix so we might as well.

(As per the other comment, since all of these only access context, none of them actually needs to be hooks and could be called conditionally if migrated. I don't think we should do that though.)

Btw, why did you go with the `return null` for these compared to `assertContextExists` which we use everywhere else? Was that a better error experience for these? (Don't have an opinion, just curious)

packages/ui/src/composed/UserProfile/UserProfileProvider.tsx
```diff
@@ -0,0 +1,83 @@
...
+  import: () => Promise.resolve(undefined) as any,
+};
+
+const composedOverrides: Elements = {
+  profilePageContent: { padding: 0 },
+};
+
+type UserProfileProviderProps = React.PropsWithChildren<{
+  appearance?: Appearance;
+  additionalOAuthScopes?: Partial<Record<OAuthProvider, OAuthScope[]>>;
```
Ephem, line 14: If these components keep rendering inside the host app, as opposed to as a separate React app, maybe SSR becomes as "easy" as passing in an `initialState` prop here, which we put on a context the hooks read from further down? 🤔

Ephem (commented):
I need to think more on the clerk-js/ui boundary, but I thought I'd leave some random thoughts.

packages/ui/src/composed/OrganizationProfile/sectionWrappers.tsx
alexcarpenter: Thanks for the edification! ❤️

alexcarpenter: Note on use(): React 19's use() isn't available yet, so I kept useContext but paired it with assertContextExists.

packages/ui/src/composed/UserProfile/sectionWrappers.tsx
alexcarpenter: https://github.com/clerk/javascript/pull/8654/commits/40e270ddf302ccf3f91223a60fea7b9d396ce424

packages/ui/src/composed/UserProfile/UserProfileProvider.tsx
alexcarpenter: Possibility! I think @jacekradko is looking at bringing these back under the same runtime here https://github.com/clerk/javascript/pull/8718

Ephem: I was looking at that, and if I understood it correctly, the components were still rendering as part of the host app, and not as a separate app?

packages/ui/src/composed/OrganizationProfile/index.tsx
alexcarpenter: handled here https://github.com/clerk/javascript/pull/8654/commits/6f53bc4f0bd4c49c8e82b1062d50067c3544cbd6

packages/ui/src/composed/ProfileProviderShell.tsx
```diff
@@ -0,0 +1,153 @@
...
+import { extractCssLayerNameFromAppearance } from '@/ui/utils/extractCssLayerNameFromAppearance';
+
+import { EnvironmentProvider } from '../contexts/EnvironmentContext';
+import { ModuleManagerProvider } from '../contexts/ModuleManagerContext';
+import { OptionsProvider } from '../contexts/OptionsContext';
+import { AppearanceOverrides } from '../elements/AppearanceOverrides';
+import type { Elements } from '../internal/appearance';
+import { createComposedRouter } from './stubRouter';
+
+export const fallbackModuleManager: ModuleManager = {
```
alexcarpenter, line 39: Real `moduleManager` registered at `clerk.ts:478` via `setModuleManager(this, this.#moduleManager)`. This fallback hits when registration missed. Means Web3 silently no-ops if the shared WeakMap channel breaks. Identity pinned by `composed-provider-wiring.test.tsx:172`.

packages/ui/src/internal/styleCacheStore.ts
```diff
@@ -0,0 +1,12 @@
+// eslint-disable-next-line no-restricted-imports
+import type { EmotionCache } from '@emotion/cache';
+
+const store = new WeakMap<object, EmotionCache>();
```
alexcarpenter, line 4: WeakMap keyed on clerk instance object, not `instanceId` string. Sibling `<UserProfile.Root>` + `<OrganizationProfile.Root>` reuse one Emotion cache, single `cl-internal` style block. Auto-GC on clerk teardown.

packages/ui/src/composed/stubRouter.ts
```diff
@@ -0,0 +1,57 @@
...
+// instead of silently rendering null.
+function unsupported(api: string): never {
+  throw new Error(
+    `[@clerk/ui composed] router.${api} is not supported inside composed sections. ` +
+      `Composed sections are leaf-only — <Route>, navigateToFlowStart(), router.refresh(), ` +
+      `getMatchData() and urlStateParam are AIO-only. Move routing to the consumer.`,
+  );
+}
+
+function maybeUnsupported<T>(api: string, prodValue: T): T {
```
alexcarpenter, line 18: Dev throws, prod no-ops. Asymmetry intentional: catches accidental AIO API use (`matches`/`refresh`/`getMatchData`) in dev without crashing consumer prod. Contracts pinned by `stub-limitations.test.ts`.

packages/ui/src/composed/UserProfile/UserProfileProvider.tsx
```diff
@@ -0,0 +1,50 @@
...
+}>;
+
+export const UserProfileProvider = (props: UserProfileProviderProps): ReactNode => {
+  const { children, appearance, additionalOAuthScopes } = props;
+  const clerk = useClerk();
+  const { isLoaded, user } = useUser();
+
+  const environment = (clerk as any).__internal_environment as EnvironmentResource | null | undefined;
```
alexcarpenter, line 23: Cast deliberate. `__internal_environment` not on typed `LoadedClerk`. Mirror cast in `OrganizationProfileProvider.tsx`. Both providers null out at line 24 if renamed in clerk-js. `context-parity.test.tsx` covers.

packages/ui/src/composed/__tests__/composed-provider-wiring.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+
+      expect(screen.getByTestId('path-probe').dataset.current).toBe('');
+    });
+  });
+
+  // Registering through @clerk/shared/moduleManager is the same channel
+  // clerk-js's constructor uses (`setModuleManager(this, this.#moduleManager)`),
+  // so a passing test here proves composed UserProfile picks up the
+  // bootstrap-registered moduleManager without `<ClerkProvider ui={ui} />`.
+  it('resolves a moduleManager registered through @clerk/shared, not just via ClerkUI', async () => {
```
alexcarpenter: Load-bearing test for the no-`ui={ui}` claim. Same registration channel as `clerk.ts:478`. Identity equality check. If `@clerk/shared` ever splits the module instance, this catches it before Web3 silently breaks.

packages/ui/src/composed/__tests__/tree-shaking.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  ] as const) {
+    it(`${label}/parts.ts contains only re-export statements`, () => {
+      const content = readFileSync(join(dir, 'parts.ts'), 'utf-8');
+      const lines = content
+        .split('\n')
+        .map(l => l.trim())
+        .filter(l => l.length > 0);
+
+      for (const line of lines) {
+        expect(line).toMatch(/^export \{.+\} from '.+';$/);
```
alexcarpenter: Regex enforces `parts.ts` is re-exports only. Any helper, type, or barrel import here defeats per-section tree-shaking and pulls every section into a consumer importing just one. Don't weaken.

alexcarpenter (commented):
Inline guideposts to speed review across the 83-file surface. Each comment is a navigation aid, not a request for changes.

alexcarpenter:
!snapshot

alexcarpenter:
!snapshot

alexcarpenter:
!snapshot

packages/ui/src/composed/UserProfile/APIKeys.tsx
alexcarpenter: Agree this deserves a dedicated design pass. For v1 the loading experience is intentionally minimal: the only Suspense boundaries are around the lazily-loaded pages (`APIKeysSection`, `BillingSection`), both with `fallback={null}`, so there's no consumer-tweakable loading UI yet — CSS only.

The longer-term DX you describe is the right target: composable Suspense + error boundaries so the host controls fallbacks, including a single top-level loading state that spans Clerk and the rest of the page. Because these render inside the host app's tree (not a standalone React root), that composition stays open to us. Tracking as a follow-up rather than blocking this PR — resolving here.

packages/ui/src/composed/UserProfile/UserProfileProvider.tsx
alexcarpenter: Right — these render inside the host app's React tree, not a separate app, so SSR via an `initialState` prop threaded onto a context the hooks read from is a viable path. Not in scope for this PR (the providers render nothing until clerk + environment resolve client-side today), but it's a clean follow-up. Resolving to track separately.

packages/ui/src/components/OrganizationProfile/OrganizationGeneralPage.tsx
```diff
@@ -58,34 +57,22 @@ const DeleteOrganizationScreen = () => {
...
+        </Protect>
+        <OrganizationLeaveSection />
+        <OrganizationDeleteSection />
+      </ProfileCard.PageSection>
     </ProfileCard.Page>
   );
 };
 
-const OrganizationProfileSection = () => {
+export const OrganizationProfileSection = () => {
```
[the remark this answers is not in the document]
alexcarpenter: Addressed in 9a131f031b — added explicit `JSX.Element | null` return types and JSDoc to all four exported section components.

packages/ui/src/components/UserProfile/__tests__/AccountSections.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+  describe('AccountEnterpriseAccounts', () => {
+    it('returns null when enterprise SSO is disabled', async () => {
+      const { wrapper } = await createFixtures(f => {
+        f.withUser({ email_addresses: ['test@clerk.com'] });
+      });
+
+      const { container } = render(<AccountEnterpriseAccounts />, { wrapper });
+      expect(container.innerHTML).toBe('');
+    });
+  });
```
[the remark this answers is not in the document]
alexcarpenter: Addressed in 9a131f031b — added an enabled-branch test (enterprise SSO on + a user with an active enterprise account) asserting the section renders.

packages/ui/src/composed/ProfileProviderShell.tsx
```diff
@@ -0,0 +1,76 @@
...
+  import: () => Promise.resolve(undefined) as any,
+};
```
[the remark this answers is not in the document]
alexcarpenter: Addressed in 9a131f031b — `fallbackModuleManager.import()` now rejects with a `ClerkRuntimeError` (code `composed_module_manager_unavailable`) instead of resolving `undefined`, so a missing module manager fails at the root cause.

alexcarpenter:
!snapshot

alexcarpenter:
!snapshot

alexcarpenter:
closing in favor of https://github.com/clerk/javascript/pull/9144

PR #8665, opened 2026-05-27
chore(repo): add conductor.json for Conductor workspaces

kylemac:
## Description

Adds a `conductor.json` at the repo root so [Conductor](https://www.conductor.build/docs/reference/conductor-json) workspaces share a consistent setup/run config. `setup` runs `pnpm install`; `run` starts the clerk-js sandbox via `pnpm run dev:sandbox`, passing `$CONDUCTOR_PORT` through to the rspack dev server so multiple workspaces can run concurrently without port collisions. Includes an empty changeset since the change is repo-tooling only.

## Type of change


jacekradko (approved):
👍🏼

PR #8666, opened 2026-05-27
chore(repo): add react-grab to clerk-js sandbox

kylemac:
## Description

Adds [`react-grab`](https://www.npmjs.com/package/react-grab) to the `clerk-js` sandbox so component locations can be jumped to from the rendered Clerk UI during local development. Installed as a `devDependency` on `@clerk/clerk-js` and loaded in `sandbox/template.html` via the unpkg global script. The sandbox has no production build (only `pnpm dev:sandbox`), so the tag is intrinsically development-only.

To test, run `pnpm dev:sandbox`, open http://localhost:4000, and use the default React Grab activation key on any rendered Clerk component.

## Checklist

- [ ] \`pnpm test\` runs as expected.

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/package.json
```diff
@@ -114,6 +114,7 @@
     "bundlewatch": "^0.4.2",
     "jsdom": "26.1.0",
     "minimatch": "^10.2.5",
+    "react-grab": "^0.1.37",
```
alexcarpenter: since your loading from unpkg, whats the reason for needing it as devDependencies?

kylemac: no probably not. i didn't catch that. i'll remove from dev dep

alexcarpenter (approved):

PR #8691, opened 2026-05-28
ci(repo): extend break-check coverage to all SDK packages

jacekradko:
Extends break-check API-change tracking from the original 6 packages to all 19 publishable SDKs (skipping `@clerk/upgrade`, `@clerk/dev-cli`, and `@clerk/msw`, which either have no public JS API or are private).

Most of the work is mechanical: each newly-tracked package gets a 3-line `tsconfig.declarations.json` and a `"build:declarations": "tsc -p tsconfig.declarations.json"` script, mirroring the pattern already in `backend`, `clerk-js`, `nextjs`, `react`, `shared`, and `ui`. `@clerk/vue` uses `vue-tsc` instead of `tsc` so SFCs resolve correctly. The choice to standardize on a TypeScript-compiler-emitted per-file `.d.ts` tree (vs tsup's bundled `.d.ts`) is deliberate: per-file output is what API Extractor / break-check can analyze without inventing duplicate-index-signature errors.

Two non-mechanical bits worth a look:

- `packages/nuxt/src/runtime/components/uiComponents.ts` adds explicit `typeof BaseUserProfile` / `typeof BaseOrganizationProfile` annotations. Without them tsc emits TS2742 ("inferred type cannot be named without a reference to '../../../node_modules/@clerk/vue/dist/types'") because `Object.assign(component, { Page, Link })` produces a structural type that can only be written by reaching into a transitive dep. Same emitted type, just spelled out.

- `packages/tanstack-react-start/tsconfig.declarations.json` adds an `exclude` for `__tests__` / `*.test.{ts,tsx}`. The pre-existing config didn't exclude tests, so its `build:declarations` was actually failing on a test file's type assertion. It just hadn't been wired into the workflow yet, so nobody noticed.

`break-check.config.json` and `.github/workflows/api-changes.yml` are expanded to cover the new packages in both the path filter and `BREAK_CHECK_FILTERS`. Verified locally: `pnpm turbo build:declarations` across all 19 packages runs clean (38 tasks, ~57s).

The tool itself was also renamed upstream from snapi to break-check (package `@clerk/break-check`, binary `break-check`), so this branch moves with it: the workflow points at the new pkg.pr.new build, `snapi.config.json` becomes `break-check.config.json`, and the `SNAPI_*` env/secret names become `BREAK_CHECK_*`. No backwards-compat shims left behind.


packages/nuxt/src/runtime/components/uiComponents.ts
```diff
@@ -58,13 +58,13 @@ const wrapComponentWithRouting = <T extends Component>(baseComponent: T, compone
...
 const _UserProfile = wrapComponentWithRouting(BaseUserProfile, 'UserProfile');
-export const UserProfile = Object.assign(_UserProfile, {
+export const UserProfile: typeof BaseUserProfile = Object.assign(_UserProfile, {
   Page: BaseUserProfile.Page,
   Link: BaseUserProfile.Link,
 });
 
 const _OrganizationProfile = wrapComponentWithRouting(BaseOrganizationProfile, 'OrganizationProfile');
-export const OrganizationProfile = Object.assign(_OrganizationProfile, {
+export const OrganizationProfile: typeof BaseOrganizationProfile = Object.assign(_OrganizationProfile, {
```
wobsoriano, line 67: nice catch! ty

wobsoriano (approved):

Issue #8659, opened 2026-05-18
AuthView 'Use phone number' toggle does not switch identifier field when embedded via Expo native


Blakefolgado:
### Summary

In Clerk iOS `AuthView`, tapping **Use phone number** can appear to do nothing when the view is embedded through the Expo native bridge / React Native host. The tap is received, but the identifier input stays on the email/username field instead of switching to the phone-number field.

### Environment

- `clerk-ios`: `1.1.1`
- `@clerk/expo`: `3.2.7`
- Expo SDK: `55`
- React Native: `0.83.6`
- iOS native `AuthView` rendered via `@clerk/expo/native`
- AuthView usage: `mode="signIn"`, `isDismissable={false}`
- Clerk instance has phone number and email sign-in enabled, so the switcher is shown

### Steps to reproduce

1. Configure a Clerk instance with both email and phone-number sign-in enabled.
2. Render native iOS `AuthView` from `@clerk/expo/native`.
3. On the initial identifier screen, tap **Use phone number**.

### Expected behavior

The identifier field switches from email/username to phone-number entry.

### Actual behavior

The button press does not visibly change the field. The user remains on the email/username identifier field.

### Source-level finding

In `Sources/ClerkKitUI/Components/Auth/AuthStartView.swift`, the active identifier mode is stored as scene storage:

```swift
@SceneStorage("phoneNumberFieldIsActive") private var phoneNumberFieldIsActive = false
```

Changing this local UI toggle state to `@State` makes the phone-number switch work reliably in the Expo native embedded AuthView:

```swift
@State private var phoneNumberFieldIsActive = false
```

That state appears to be purely local to the current auth-start view, so scene persistence may be unnecessary and can break in embedded/native-host contexts where SwiftUI scene storage is not behaving like a normal app scene.

### Notes

This does not appear to be a Clerk Dashboard strategy/configuration issue: the phone strategy is enabled, the switcher renders, and phone OTP APIs are available. The failure is the SwiftUI field-mode toggle not updating the rendered identifier field.

seanperez29:
Hey @Blakefolgado 👋 Thanks for reporting this! We will look into this and hope to get an update out shortly. Will keep you updated. Thanks again!

mikepitre:
Hey @Blakefolgado this should be fixed in the latest expo sdk version! Thanks for reporting!

PR #8699, opened 2026-05-28
feat(expo): align native prebuilt components with mobile SDKs

mikepitre:
## Summary

This PR brings Expo's native prebuilt components closer to the iOS and Android SDKs:

- `AuthView` and `UserProfileView` are now renderable content. App code owns whether they appear in a modal, sheet, route, or full-screen root.
- `UserButton` is backed by the platform-native Clerk button instead of a JS avatar/button implementation.
- Native and JS auth state sync by refreshing both SDK clients toward the Clerk backend client, not by mirroring sign-out or interpreting component-level auth events.
- Android native registration moved to Expo Modules. iOS stays on the existing app-target bridge until Expo has a clean SPM-compatible path for this setup.

Updated Docs PR https://github.com/clerk/clerk-docs/pull/3424

### Prebuilt Views Are App-Presented Content

**Before:** Expo owned presentation through imperative helpers and Android activities, and profile presentation used `useUserProfileModal()`.

```tsx
const { presentUserProfile } = useUserProfileModal();

<AuthView isDismissable={false} />;
<Button title="Manage account" onPress={presentUserProfile} />;
```

**After:** `AuthView` and `UserProfileView` render as content. Apps present them in their own modal, sheet, route, or root view.

```tsx
<Modal visible={open} presentationStyle="pageSheet" onRequestClose={() => setOpen(false)}>
  <AuthView onDismiss={() => setOpen(false)} />
</Modal>

<UserProfileView onDismiss={() => setOpen(false)} />
<AuthView isDismissible={false} />
```

**Motivation:** This matches the native SDK model: the prebuilt component owns its content and built-in dismissal affordance, while the app owns presentation.

### UserButton Is The Native Button Surface

**Before:** `UserButton` was a JS-managed avatar/profile launcher and accepted JS layout props.

```tsx
<UserButton style={{ width: 40, height: 40 }} />
```

**After:** `UserButton` wraps the platform-native Clerk button directly.

```tsx
<UserButton />
```

**Motivation:** The Expo API now mirrors the mobile SDKs: tapping the native button opens the native user profile, and Expo does not reimplement avatar/profile behavior in JS.

### Native And JS Sync Through Client Refresh

**Before:** Native prebuilt views emitted auth-specific events such as sign-in/sign-out completion, and Expo JS interpreted those events or mirrored sign-out across SDKs.

**After:** Both sides refresh toward the Clerk backend client:

- Native-to-JS: iOS observes Clerk iOS client changes; Android observes `Clerk.clientFlow`; both emit an internal `refreshClient` event.
- JS-to-native: `ClerkProvider` observes Clerk JS resource changes, syncs the current JS client token to native when one exists, and otherwise asks the native SDK to refresh its current client.

**Motivation:** The backend client endpoint stays the source of truth. Expo no longer needs component-level auth semantics, duplicate sign-out calls, or one-off session syncing paths.

### Android Uses Expo Modules

**Before:** Android used React Native package/view-manager registration plus presentation activities/factory paths for the prebuilt surfaces.

**After:** Android registers the Clerk module and native prebuilt views through Expo Modules. JS uses `requireNativeModule` and `requireNativeView` for Android.

**Motivation:** Expo Modules are the cleaner fit for Android now. iOS intentionally stays on the existing React Native bridge plus app-target native bridge because Clerk iOS is supplied to the consuming app through SPM, and the CocoaPods-backed Expo module target cannot cleanly import that app-target SPM dependency yet.


packages/expo/android/src/main/java/expo/modules/clerk/ClerkExpoModule.kt
```diff
@@ -34,46 +32,56 @@ private fun debugLog(tag: String, message: String) {
...
+            if (listenerCount <= 0) {
+                return
+            }
 
-        // Pending promises for activity results
-        private var pendingAuthPromise: Promise? = null
-        private var pendingProfilePromise: Promise? = null
+            val event = Arguments.createMap().apply {
+                putString("type", type)
+                if (sessionId == null) {
```
swolfand: You're putting session id in this map either way. Could we simplify the expression?

packages/expo/android/src/main/java/expo/modules/clerk/ClerkUserButtonExpoView.kt
```diff
@@ -0,0 +1,116 @@
...
+            Log.w(USER_BUTTON_TAG, "Client refresh after UserButton sign-out failed: ${e.message}")
+          }
+          ClerkExpoModule.emitAuthStateChange("signedOut", null)
+        }
+        if (session != null) {
+          hadSession = true
+        }
+      }
+
+      val content = @androidx.compose.runtime.Composable {
```
swolfand: Nit: could we import this package? Should be able to do `@Composable`

packages/expo/android/src/main/java/expo/modules/clerk/ClerkUserButtonExpoView.kt
```diff
@@ -0,0 +1,116 @@
...
+        MaterialTheme {
```
swolfand: Would be curious what this is actually doing. It may be needed I'm just curious

packages/expo/android/src/main/java/expo/modules/clerk/ClerkExpoModule.kt
mikepitre: Yep, agreed. I simplified this to write the nullable `sessionId` directly with `putString("sessionId", sessionId)`, which keeps the same JS shape without the explicit null branch.

packages/expo/android/src/main/java/expo/modules/clerk/ClerkUserButtonExpoView.kt
mikepitre: Yep, done. I imported `Composable` and changed the local lambda to `val userButtonContent: @Composable () -> Unit = { ... }` so this reads like the other Compose wrappers.

packages/expo/android/src/main/java/expo/modules/clerk/ClerkUserButtonExpoView.kt
mikepitre: This is the Compose host setup for rendering Clerk Android UI inside a React Native view. The Clerk Compose components expect lifecycle/view model/saved-state owners, but RN-hosted views do not reliably inherit those locals, so we provide the Activity owners explicitly. I added a short comment here to make that intent clearer.

chriscanin (commented):
Hi Mike, things are looking really good in here, I have tested each screen thourouhgly side by side with the native ios quickstart application. **This review is for iOS only right now**. I am happy to test the android, but I wanted to put this out there first before moving on. 

In my testing I found 3 (minor) issues that could be considered not 1:1: 
1. No grabber pill on the top of the RN Modal for sign in, the user button sheet has one.
2. After sign-out, AuthView modal auto-opens instead of returning to the unauthenticated Sign In button state.
3. Phone number sign-in/up tab tapping `Use phone number` does nothing, but this is a known issue: https://github.com/clerk/javascript/issues/8659
**Side note:** Docs and quickstarts will need updates to go along with this branch. I am happy to provide those once we are reviewed and satisfied with this PR. 

I have also left 3 code comments / questions in this review as well.

packages/expo/ios/ClerkExpoModule.swift
```diff
@@ -264,129 +252,93 @@ public class ClerkAuthNativeView: UIView {
...
   private var hasInitialized: Bool = false
+  private var didSignOut = false
+  private var dismissalEventSent = false
 
   @objc var onProfileEvent: RCTBubblingEventBlock?
 
   @objc var isDismissable: NSNumber? {
     didSet {
-      currentDismissable = isDismissable?.boolValue ?? true
+      currentDismissable = isDismissable?.boolValue ?? false
```
seanperez29: Should this get the same `guard newDismissable != currentDismissable else { return }` that line 237 has?

packages/expo/ios/ClerkExpoModule.swift
```diff
@@ -216,15 +130,95 @@ class ClerkExpoModule: RCTEventEmitter {
   }
 }
 
+// MARK: - Inline View: ClerkUserButtonNativeView
+
+public class ClerkUserButtonNativeView: UIView {
```
seanperez29: Is it worth pulling each native view out of this file into it's own?

packages/expo/ios/ClerkExpoModule.swift
```diff
@@ -216,15 +130,95 @@ class ClerkExpoModule: RCTEventEmitter {
...
+  }
+
+  private func updateView() {
+    detachHostingController()
+
+    guard let factory = clerkViewFactory else { return }
+
+    guard let returnedController = factory.createUserButton(
+      onEvent: { eventName, data in
+        if eventName == "signedOut" {
```
seanperez29: Could these events be typed in some manner rather than the raw strings?

packages/expo/ios/ClerkExpoModule.swift
```diff
@@ -216,15 +130,95 @@ class ClerkExpoModule: RCTEventEmitter {
...
+          let sessionId = data["sessionId"] as? String
+          ClerkExpoModule.emitAuthStateChange(type: "signedOut", sessionId: sessionId)
+        }
+      }
+    ) else { return }
+
+    attachHostingController(returnedController)
+  }
+
+  private func attachHostingController(_ controller: UIViewController) {
```
seanperez29: I saw this method along with `detachHostingController` and `findViewController` are recreated a bit. I asked codex what can be done and it recommended either composition with a small hosting coordinator/helper object or superclass. Superclass I feel eh about. From a high level the composition approach doesn't sound too bad, if it doesn't end up being weird once implemented

packages/expo/ios/ClerkExpoModule.swift
mikepitre: Yep, good catch. Added the same guard here so `UserProfileNativeView` only rebuilds when `isDismissable` actually changes, matching the `AuthView` behavior.

packages/expo/ios/ClerkExpoModule.swift
mikepitre: Went with the composition approach you suggested. Added a small `ClerkNativeHostingCoordinator` helper for attach/detach/layout/view-controller lookup and avoided introducing a superclass.

packages/expo/ios/ClerkExpoModule.swift
mikepitre: Agreed. Moved each native view into its own file so `ClerkExpoModule.swift` only owns the module/protocol surface now.

packages/expo/ios/ClerkExpoModule.swift
mikepitre: Added typed Swift enums for this. View-wrapper events now use `ClerkNativeViewEvent`, and auth-state sync uses `ClerkNativeAuthStateEvent`; we only stringify at the JS bridge boundary.

swolfand (approved):

seanperez29 (approved):

packages/expo/ios/ClerkNativeViewHost.swift
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  override public func didMoveToWindow() {
+    super.didMoveToWindow()
+
+    if window != nil && !hasInitialized {
+      hasInitialized = true
+      updateHostedView()
+    }
+  }
```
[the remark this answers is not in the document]
mikepitre: Agreed. I added the `window == nil` detach path and reset initialization so a detached native host releases its hosted controller and can reinitialize cleanly when attached again.

packages/expo/src/hooks/useNativeClientEvents.ts
```diff
@@ -0,0 +1,62 @@
...
+      return;
+    }
+
+    let subscription: { remove: () => void } | null = null;
+
+    try {
+      const eventEmitter: RefreshClientEventEmitter =
+        Platform.OS === 'android'
+          ? (new EventEmitter(ClerkExpo as never) as unknown as RefreshClientEventEmitter)
+          : (new NativeEventEmitter(ClerkExpo) as RefreshClientEventEmitter);
```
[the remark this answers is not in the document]
mikepitre: Agreed. Removed this cast. Android now subscribes through the Expo module’s own `addListener`, while iOS continues using `NativeEventEmitter`.

packages/expo/android/src/main/java/expo/modules/clerk/ClerkUserProfileViewModule.kt
```diff
@@ -0,0 +1,74 @@
...
+          }
+        )
+      }
+    }
+  }
+
+  private fun sendEvent(type: String) {
+    onProfileEvent(mapOf("type" to type))
+  }
+}
```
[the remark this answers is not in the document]
mikepitre: Good catch. I mirrored the `AuthView` pattern here: `UserProfileView` now gets a per-view `ViewModelStoreOwner`, and the store is cleared when the host detaches so profile navigation state does not leak across remounts.

packages/expo/src/native/AuthView.tsx
```diff
@@ -1,60 +1,10 @@
-import { ClerkRuntimeError } from '@clerk/shared/error';
-import { useCallback, useRef } from 'react';
+import { useCallback } from 'react';
```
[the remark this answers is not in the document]
mikepitre: Added a focused test for native `dismissed` events calling `onDismiss`. I also tightened the handler type by deriving it from the generated native view prop instead of hand-rolling the event shape.

packages/expo/src/provider/ClerkProvider.tsx
```diff
@@ -54,147 +53,160 @@ const SDK_METADATA = {
...
+  const jsActiveSessionId = clerkInstance.session?.id;
+
+  if (
+    nativeActiveSessionId &&
+    nativeActiveSessionId !== jsActiveSessionId &&
+    typeof clerkInstance.setActive === 'function'
+  ) {
+    await clerkInstance.setActive({ session: nativeActiveSessionId });
+  }
+}
```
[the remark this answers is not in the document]
mikepitre: Cleaned this up with a local `SyncableClerkInstance` type and explicit return types around the native client sync helpers.

.changeset/remove-expo-present-auth.md
```diff
@@ -0,0 +1,17 @@
...
+
+Align `AuthView` dismissal with the native SDK lifecycle. App code now owns presentation while the native auth view requests `onDismiss` for user dismissal and successful auth completion after native post-auth routing is complete.
+
+Separate native view events from auth-state sync events. Expo host views now forward only dismissal through the view event channel, while sign-in and sign-out updates use the internal native auth-state sync channel.
+
+When auth, session, or user state changes on one side of the Expo bridge, the opposite client now refreshes from Clerk instead of issuing a second sign-out. This keeps JS and native prebuilt components synced after native profile updates, foreground client refreshes, and sign-out.
+
+Align `UserButton` with the native Clerk SDKs by wrapping the platform-native user button, letting it present the user profile from the button itself.
+
+Bump the bundled clerk-android prebuilt UI dependency to `1.0.20` for the native `AuthView` dismissal callbacks used by the Expo wrapper.
```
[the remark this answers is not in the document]
mikepitre: I’m intentionally not making this a major. These native prebuilt components are still beta, and this PR is aligning that beta surface with the native SDK usage model rather than preserving the older beta API shape.

packages/expo/src/provider/ClerkProvider.tsx
```diff
@@ -54,147 +53,160 @@ const SDK_METADATA = {
...
+          .finally(() => {
+            isRefreshingNativeFromJsRef.current = false;
+          });
+      },
+      { skipInitialEmit: true },
+    );
+  }, [clerkInstance, effectiveTokenCache, isSyncingNativeClientToJsRef, publishableKey]);
 
   return null;
 }
```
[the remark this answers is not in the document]
mikepitre: Cleaned this up with a local `SyncableClerkInstance` type and explicit return types around the native client sync helpers.

packages/expo/src/native/index.ts
```diff
@@ -23,14 +23,13 @@
...
  *
  * @module @clerk/expo/native
  */
 
 export { AuthView } from './AuthView';
 export type { AuthViewProps, AuthViewMode } from './AuthView.types';
 export { UserButton } from './UserButton';
-export type { UserButtonProps } from './UserButton';
 export { UserProfileView } from './UserProfileView';
 export type { UserProfileViewProps } from './UserProfileView';
```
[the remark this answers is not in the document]
mikepitre: I’m intentionally not restoring this for compatibility. The native prebuilt components are still beta, and `UserButton` should stay the native button surface rather than a JS-styled wrapper with legacy props.

packages/expo/src/specs/NativeClerkGoogleSignIn.android.ts
```diff
@@ -0,0 +1,13 @@
...
+
+type NativeMap = Record<string, unknown>;
+
+interface Spec {
+  configure(params: NativeMap): void;
+  signIn(params: NativeMap | null): Promise<NativeMap>;
+  createAccount(params: NativeMap | null): Promise<NativeMap>;
+  presentExplicitSignIn(params: NativeMap | null): Promise<NativeMap>;
+  signOut(): Promise<void>;
+}
```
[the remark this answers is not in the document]
mikepitre: Leaving this as-is. This spec is an internal native bridge contract rather than user-facing API docs, and adding method-by-method JSDoc here feels like noise for this PR.

packages/expo/src/specs/NativeClerkModule.android.ts
```diff
@@ -0,0 +1,16 @@
...
+interface Spec {
+  // addListener/removeListeners are present on the iOS RN event emitter module.
+  // Android uses Expo Modules EventEmitter instead.
+  addListener?(eventName: string): void;
+  configure(publishableKey: string, bearerToken: string | null): Promise<void>;
+  getSession(): Promise<NativeMap | null>;
+  getClientToken(): Promise<string | null>;
+  refreshClient(): Promise<void>;
+  removeListeners?(count: number): void;
+}
```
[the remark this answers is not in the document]
mikepitre: Leaving this as-is. This spec is an internal native bridge contract rather than user-facing API docs, and adding method-by-method JSDoc here feels like noise for this PR.

packages/expo/android/src/main/java/expo/modules/clerk/ClerkUserProfileViewModule.kt
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  @Composable
+  override fun Content() {
+    debugLog(TAG, "setupView - isDismissible: $isDismissible")
+
+    MaterialTheme {
+      Surface(
+        modifier = Modifier.fillMaxSize(),
+        color = MaterialTheme.colorScheme.background
```
swolfand: is there a reason we want to always default to a background color here?

.changeset/quiet-android-bridge.md
wobsoriano, line 1: 🪓

packages/expo/android/src/main/java/expo/modules/clerk/ClerkUserProfileViewModule.kt
mikepitre: Good call. There was not a strong reason for Expo to force a Material background here. The Clerk Android prebuilt views already own their themed surfaces, so I removed the Expo-side wrappers and now render `AuthView` / `UserProfileView` directly.

I also did a follow-up pass for the same kind of unnecessary presentation glue: removed the extra `MaterialTheme` wrapper around Android `UserButton`, and removed redundant iOS clear-background plumbing now that the SwiftUI hosting controller is already cleared in one place.

.changeset/remove-expo-present-auth.md
wobsoriano:
```suggestion
Update Expo's beta native prebuilt components to more closely match the behavior of Clerk's native iOS and Android SDKs.

Previously, native auth and profile views relied on Expo-specific presentation behavior. `AuthView` and `UserProfileView` are now app-presented components, with dismissal handled through `onDismiss`. This also improves session synchronization between Clerk's JavaScript and native layers.

**Note:** This includes native changes, so rebuild your native app after upgrading (`expo prebuild --clean` or a new EAS build).
```

Let's make it user facing instead of the implementation details 👍🏼

.changeset/bump-expo-native-sdks.md
wobsoriano, line 1: let's 🪓 and keep one changeset

packages/expo/src/provider/ClerkProvider.tsx
```diff
@@ -54,147 +53,170 @@ const SDK_METADATA = {
...
+  const maxAttempts = 30;
+  const intervalMs = 100;
+
+  for (let attempt = 0; attempt < maxAttempts; attempt++) {
+    const nativeClientToken = await ClerkExpo.getClientToken();
+    if (nativeClientToken) {
+      return nativeClientToken;
+    }
+    await new Promise(resolve => setTimeout(resolve, intervalMs));
+  }
```
wobsoriano, line 81: ~This polls up to 3s for a token. On sign-out the device token still exists, so it returns fast. Is there a startup case where there's no token yet and it waits the full 3s? What does the user see then?~

Tested sign-out + cold start (signed out) and it's always sub-second, so this never hits the 30-attempt ceiling in practice

packages/expo/src/native/AuthView.tsx
```diff
@@ -83,49 +36,14 @@ export async function syncNativeSession(sessionId: string): Promise<void> {
...
-        } else if (__DEV__) {
-          console.warn('[AuthView] Auth event received but no sessionId in data:', data);
-        }
+    (event: AuthNativeEvent) => {
+      if (event.nativeEvent.type === 'dismissed') {
+        onDismiss?.();
       }
     },
-    [syncSession],
+    [onDismiss],
```
wobsoriano, line 46: Testing this on a local Expo app and I think there's an iOS dismissal bug, want to confirm. Rendering `<AuthView onDismiss={...} />` inside an app-owned `<Modal presentationStyle="pageSheet">` (the pattern from the PR description) and pressing the native close button closes the sheet but then freezes the whole app (touches stop working). Same with `UserProfileView.`

wobsoriano (changes requested):
Left minor suggestions and one bug I encountered on iOS (havent tested Android)

packages/expo/app.plugin.js
```diff
@@ -214,7 +214,7 @@ const withClerkIOS = config => {
     return config;
   });
 
-  // Inject ClerkViewFactory.register() call into AppDelegate.swift
+  // Inject ClerkNativeBridge.register() call into AppDelegate.swift
```
seanperez29, line 217: I was able to reproduce this one. Upgraded prebuilt iOS apps can still have `ClerkViewFactory.register()` and `ClerkViewFactory.swift` from the previous plugin. This adds `ClerkNativeBridge`, but can fail with `ClerkViewFactory.swift:14:32: error: cannot find type 'ClerkViewFactoryProtocol' in scope` due to the old registration and file references.

packages/expo/src/native/AuthView.tsx
wobsoriano: Added native logs to dig in. On the close button, only `ClerkNativeViewHost.didMoveToWindow(window=nil)` fires, `ClerkNativeHostingController.dismiss()` and the `onEvent(.dismissed)` relay never run

packages/expo/app.plugin.js
mikepitre: Added a note to the changeset devs may need to run `expo prebuild --clean`

packages/expo/src/native/AuthView.tsx
mikepitre: Pushed a fix for this in 217e4c72fce031ec0c141e96a3b0dbe8932882a9

wobsoriano: thanks mike! fix works 👍🏼

wobsoriano (approved):
Thanks for improving and fixing the native components! 👍🏼 

Attaching my local testing for reference

https://github.com/user-attachments/assets/0223b542-b5de-410e-b34d-2cfb29ac62f4

PR #8700, opened 2026-05-28
chore(js,ui,shared): Correctly display OAuth consent redirect root domains

wobsoriano:
## Description

The OAuth consent page was computing the display domain client-side by splitting the hostname on dots, which breaks for ccTLDs like `.co.uk` and hosting platforms like `github.io`. This is a security concern since users make trust decisions based on that domain.

The backend already has full PSL support, so this PR threads `redirect_uri` through to `GET /v1/me/oauth/consent/{clientID}` and uses the resolved `redirectDomain` from the response instead. The client-side heuristic is kept as a fallback for IPs, localhost, and older FAPI versions.

Part of USER-5440

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/components/OAuthConsent/OAuthConsent.tsx
```diff
@@ -50,9 +50,11 @@ function _OAuthConsent() {
 
   // Public path: fetch via hook. Disabled on the accounts portal path
   // (which already has all data via context) to avoid a wasted FAPI request.
+  const redirectUri = getRedirectUriFromSearch();
   const { data, isLoading, error } = useOAuthConsent({
     oauthClientId,
     scope,
+    redirectUri: redirectUri || undefined,
```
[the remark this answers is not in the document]
jfoshee: Is ctx just used for testing?

packages/ui/src/components/OAuthConsent/OAuthConsent.tsx
```diff
@@ -50,9 +50,11 @@ function _OAuthConsent() {
 
   // Public path: fetch via hook. Disabled on the accounts portal path
   // (which already has all data via context) to avoid a wasted FAPI request.
+  const redirectUri = getRedirectUriFromSearch();
```
jfoshee: Based on coderabbit analysis, should this get the redirect from ctx? This is used in AP?

```suggestion
  const redirectUri = ctx.redirectUrl ?? getRedirectUriFromSearch();
```

jfoshee (approved):
Did you test w/ AP locally?  You should be able to test it locally now using `make up/local-js`

.changeset/quiet-terms-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/ui': patch
```
jacekradko, line 2: Looks like this also touches shared and clerk-js

jacekradko (changes requested):
Need to update the changeset

.changeset/quiet-terms-drum.md
wobsoriano: thanks! will update after some more local testing later

wobsoriano:
> Did you test w/ AP locally? You should be able to test it locally now using `make up/local-js`

I tested via `pnpm dev:sandbox` since the backend change is live and can confirm it strips correct root domain 👍🏼 

Screenshot 2026-06-01 at 10 36 42 AM

wobsoriano:
@jacekradko thoughts on the API changes report here https://github.com/clerk/javascript/pull/8700#issuecomment-4594573068 ?

The `OAuthConsentInfo` update is backwards compat

jacekradko:
@wobsoriano Looked at the break-check output. Those are indeed false positives. Will update `break-check` to better handle these types of changes

jacekradko (approved):

wobsoriano:
> @wobsoriano Looked at the break-check output. Those are indeed false positives. Will update `break-check` to better handle these types of changes

Thanks! Feel free to merge this or use it as test PR when break-check is updated

PR #8701, opened 2026-05-29
docs(shared): use @hidden over @internal for internal Clerk/Client members

jacekradko:
Targets the #8276 branch.

Those 18 `__internal_*` members on `Clerk` and `ClientResource` got `@internal` tags so they'd stay out of the generated `/objects` docs. The problem is snapi reads `@internal` as a TSDoc release tag and drops them from the API snapshot, so the API Changes check sees them disappear and reports 18 "removed" breaking changes. But there's no `stripInternal` in the build, so these members are still emitted in the `.d.ts` and at runtime. They're genuinely still part of the contract and shouldn't leave snapi's radar.

Swapping `@internal` to `@hidden` on just those members keeps typedoc excluding them from the docs (`@hidden` is a removal tag), while snapi keeps tracking them since `@hidden` isn't a release tag. The 14 pre-existing `@internal` tags are left alone. I ran the typedoc generation locally to confirm none of the 18 leak into the output and the build is clean.

One heads up: the API Changes check won't run on this PR, that workflow only triggers on PRs into `main`/`release/*`, so it'll re-validate on #8276 once these commits land on the branch.

### DIFF
--- a/packages/shared/src/types/clerk.ts
+++ b/packages/shared/src/types/clerk.ts
@@ -332,7 +332,7 @@ export interface Clerk {
    * Entrypoint for Clerk's Signal API containing resource signals along with accessible versions of `computed()` and
    * `effect()` that can be used to subscribe to changes from Signals.
    *
-   * @internal
+   * @hidden
    * @experimental This experimental API is subject to change.
    */
   __internal_state: State;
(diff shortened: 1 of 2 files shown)


wobsoriano (approved):

alexisintech:
ahhh you're a genius thank you!

PR #8703, opened 2026-05-29
fix(ui): scope active-devices cache by user to prevent cross-user leak

dominic-clerk:
## Description

The `UserProfile` active-devices list was cached in a module-scoped fetch cache keyed only by a constant string, so a session switch or a sign-out/sign-in on a shared device could render the previous user's device activity (IP, location, browser/device). Key the cache by `user.id`.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/tangy-ducks-end.md
+++ b/.changeset/tangy-ducks-end.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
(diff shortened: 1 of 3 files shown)


jacekradko (approved):
👍🏼

PR #8704, opened 2026-05-29
feat(eslint-plugin): Add initial @clerk/eslint-plugin package and rule

Ephem:
## Description

> This PR also contains this one that was merged into it and implements suggestions and fixing-capabilities: https://github.com/clerk/javascript/pull/8828

Adds `@clerk/eslint-plugin`, a package for eslint-rules. Adds a first rule, `require-auth-protection` that enforces auth protections for the Next app router at the page/route/server action level.

The rule flags any `page`, `layout`, `template`, `default`, `route`, or Server Action under folders configured as `protected` that doesn't guard itself with `await auth.protect()` (or an equivalent early-exit `auth()` check).

## What's included

- New package `packages/eslint-plugin` (dual CJS/ESM, type-only deps, `eslint >=9` peer, ships at `0.0.0` → `0.1.0`)
- `require-auth-protection` rule with `protected` (required), `public`, and `mixedScopeLayouts` options
- Full test suite
- CI setup

### Config

```js
import clerkNextPlugin from '@clerk/eslint-plugin/next';

export default [
  {
    plugins: { '@clerk/next': clerkNextPlugin },
    rules: {
      '@clerk/next/require-auth-protection': [
        'error',
        {
          protected: ['app/**'],
          public: ['app/sign-in/**', 'app/sign-up/**'],
        },
      ],
    },
  },
];
```

Also see README for more options. It is possible to be extra strict by providing explicit `mixedScopeLayouts`, and it's possible to turn off checking for `routeHandlers`, `serverFunctions` or `serverComponentEntrypoints`.

## Errors

* missingProtect:
  * 'Expected `await auth.protect()` at the top of {{subject}} in a folder configured as protected. Add the call to the top of the function, move the file into a public folder, or configure this folder as public.',
* exportImported:
  * "This {{subject}} is exported from '{{source}}'. The rule cannot follow imports across files. Add a wrapper with `await auth.protect()`, or ensure the imported function calls it and add an eslint-disable comment with a reason.",
* unverifiableExport:
  * 'This {{subject}} could not be verified as being protected, likely because it is assigned from a call expression (e.g. `const handler = withAuth(impl)`). Inline a function literal that calls `await auth.protect()`, or add an eslint-disable comment with a reason.',
* unlistedMixedScopeLayout (only if an explicit `mixedScopeLayouts` was provided in config):
  * "This {{fileKind}} at '{{folder}}/' wraps both protected and public descendants but is not listed in `mixedScopeLayouts`. Either add '{{folder}}' to the list to acknowledge the mixed scope, or restructure so the {{fileKind}} wraps only public or protected descendants.",

## Notes

- Verified against our internal dashboard repo
- Other tooling is upcoming
- ✅ Before merging and releasing, we need to double check first time publish via OIDC will work
  - `@clerk/eslint-plugin` package has been published and set up for trusted publishing, so should work
  - `pkg-pr-new` does not work since current npm pkg does not have the correct setup, will work after first publish

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Added `@clerk/eslint-plugin-next` with an experimental require-auth-protection rule for Next.js App Router to enforce auth guards on pages, routes, layouts, and server functions.

* **Documentation**
  * Added README describing installation, configuration options, recognized auth checks, and usage examples.

* **Tests**
  * Comprehensive test suites covering folder classification, pattern matching, protection detection, rule behavior, and schema validation.

* **Chores**
  * Package metadata, build/test configs, license, and CI labeler updates for publishing.


packages/eslint-plugin-next/src/__tests__/__snapshots__/plugin-shape.test.ts.snap
```diff
@@ -0,0 +1,9 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
```
jacekradko: Now that we have break-check, these snapshot tests are duplicative

packages/eslint-plugin-next/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+`@clerk/eslint-plugin-next` is a static analysis aid, not a runtime guard. It's provided to help you catch missing protections and it does error on the side of caution, but there are no guarantees there might not be edge cases it fails to detect.
+
+_For more information and to report security issues, please refer to our [security documentation](https://github.com/clerk/javascript/blob/main/docs/SECURITY.md)._
+
+## License
+
+This project is licensed under the **MIT license**.
+
+See [LICENSE](https://github.com/clerk/javascript/blob/main/packages/eslint-plugin-next/LICENSE) for more information.
```
jacekradko: I don't see a LICENSE file in this PR

packages/eslint-plugin-next/src/__tests__/__snapshots__/plugin-shape.test.ts.snap
Ephem: Removed!

packages/eslint-plugin-next/README.md
Ephem: Great catch, added!

packages/eslint-plugin/src/next/require-auth-protection.ts
```diff
@@ -0,0 +1,322 @@
...
+    return {
+      Program(programNode) {
+        const ast = programNode as TSESTree.Program;
+        const isServerFunction = isServerFunctionModule(ast);
+        const isClient = isClientModule(ast);
+
+        const sourceClass = classifyFolder(folder, config);
+        if (sourceClass !== 'protected') {
+          return;
+        }
```
jacekradko, line 143: Right now, if a folder doesn't match either the `protected` or `public` globs, the rule just skips it and says nothing. That's the part that worries me. Say someone sets `protected: ['app/dashboard/**']` and later adds an `app/admin` section. None of those admin pages match the glob, so the rule stays quiet, and the whole section ships with no auth checks and no warning that anything was missed.

The catch is that people will install this rule specifically to feel confident their protected areas are covered, but the rule only checks the folders they remembered to list. When they forget one, that's exactly when they need a heads-up, and that's exactly when they get silence.

Could we add an option to flag resource files that live in folders matching neither list? Even off by default, it gives people a way to catch the folders they forgot. At a minimum, I think the README should say plainly that anything not listed as protected is simply not checked.

packages/eslint-plugin/src/next/lib/protection-checks.ts
```diff
@@ -0,0 +1,288 @@
...
+      }
+      if (stmt.type === 'ThrowStatement') {
+        return true;
+      }
+      if (stmt.type === 'ExpressionStatement' && isExitCall(stmt.expression)) {
+        return true;
+      }
+    }
+  }
+  return false;
```
jacekradko, line 243: Question on this one: the guard counts as good as soon as there's a `return` or `throw` in the block, even if something runs before it. So this passes:

```tsx
const { userId } = await auth();
if (userId === null) {
  await db.write({ event: 'unauth-access', userId }); // runs while signed out
  return null;
}
```

The `await` runs for a signed-out user before we bail. Is that a pattern you'd expect people to hit, or rare enough not to worry about? Curious how you're thinking about it.

packages/eslint-plugin/src/next/lib/protection-checks.ts
```diff
@@ -0,0 +1,288 @@
...
+
+function isExitCall(expr: TSESTree.Expression | null | undefined): boolean {
+  if (!expr || expr.type !== 'CallExpression') {
+    return false;
+  }
+  if (expr.callee.type !== 'Identifier') {
+    return false;
+  }
+  return EXIT_FUNCTIONS.has(expr.callee.name);
+}
```
jacekradko, line 210: Question on how wrappers are meant to work here. Exit functions are matched by name, so the rule lands in an odd spot in both directions:

```tsx
const redirect = (url) => {};                  // not next/navigation
if (!userId) redirect('/sign-in');             // no-op, but counts as an exit

if (!userId) perhapsRedirOrThrow();            // real wrapper, but gets flagged
```

A local no-op named `redirect` passes, while a real wrapper with any other name doesn't. The note above says name-matching is there to support wrappers, but it only works if the wrapper is named exactly `redirect`/`notFound`/etc., which people won't do.

Since we can't really tell whether `perhapsRedirOrThrow()` exits without following it, would it be cleaner to recognize only the actual `next/navigation` imports, and let people opt wrappers in through a config allowlist? Curious how you're thinking about the wrapper case.

packages/eslint-plugin/src/next/lib/protection-checks.ts
```diff
@@ -0,0 +1,288 @@
...
+
+  const captured = capturedAuthBindings(stmts[first], authNames);
+  if (captured) {
+    const second = nextExecutable(stmts, first + 1);
+    if (second < stmts.length && isAuthGuardWithExit(stmts[second], captured)) {
+      return true;
+    }
+  }
+
+  return false;
```
jacekradko, line 308: The guard has to be in a pretty specific shape to count, and a few safe patterns get flagged because of it. Flagging since I think they'll come up a lot:

```tsx
// 1. reading params first (normal in Next 15) - protect() still guards everything below
const { id } = await params;
await auth.protect();

// 2. reading auth for the page, then protecting
const { userId } = await auth();
await auth.protect();

// 3. the same check the rule accepts, just written with !
const { userId } = await auth();
if (!userId) return null;   // rule wants `userId === null`
```

All three are protected, but each one trips the rule. I know there are a lot of valid coding styles out there that could lead to false positives like these, so we may need to do some extra design work on this before it's solid. That said, it's also fair to take the stance that a strict rule just requires people to stick to one blessed way of writing the check, as long as that's a deliberate call and we document it clearly.

packages/eslint-plugin/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+export default [
+  {
+    plugins: { '@clerk/next': clerkNext },
+    rules: {
+      '@clerk/next/require-auth-protection': [
+        'error',
+        {
+          protected: ['app/**'],
+          public: ['app/sign-in/**', 'app/sign-up/**'],
```
jacekradko, line 71: One thing about this example: folders get matched by their path on disk, not by the URL. It works as written, but if someone puts their auth pages in a route group like `app/(auth)/sign-in/`, the folder becomes `app/(auth)/sign-in` and `app/sign-in/**` stops matching it. Then `app/**` picks it up as protected and the sign-in page starts failing the rule. Route groups around auth pages are pretty common, so I think people will run into this. Might be worth mentioning in the docs, or stripping the group/slot/private segments before matching so it follows the URL instead.

packages/eslint-plugin/src/next/require-auth-protection.ts
Ephem: That's fair. 🤔

I've been thinking that the option you are talking about is just choosing to protect everything with `['/app/**']` and specifically opting out of routes, which is what I think we should primarily document.

I could definitely see something like the `mixedScopeLayouts` option but that warns for indeterminable routes instead. I think that should be on by default with opt out being possible, or else it kind of defeats the purpose.

On the other hand, it's pretty obvious that if you don't configure it, it wont be covered, so might also be a follow up? I'll feel it out, don't think it's that hard to implement!

packages/eslint-plugin/src/next/lib/protection-checks.ts
Ephem: I hadn't thought about it that much honestly, so this is a good catch. 🤔

I have mostly gone with safety first, like how the protection has to be at the very top of the pages etc, so my instinct is that we should do that here as well. If being able to for example log before the exit becomes an ask, we could implement some kind of `safeStatements` option?

packages/eslint-plugin/src/next/lib/protection-checks.ts
Ephem: Yeah, the same applies to `auth.protect` actually, it matches by name, not by scope walking to an import.

It's all a tradeoff of the implementation complexity, the upsides and how easy it is to understand those tradeoffs for end users. I want to keep the first version of the lint rule as simple as possible so we can fully understand it without too much mental overhead.

It's easy enough for an agent to implement scope walking, I tried it but ended up reverting because I didn't like the complexity it brought _at this stage_.

If you have a `redirect` or `notFound` function that doesn't actually DO those things, you are kind of on your own. 😅 My thinking is that the names being _indicative_ is good enough until we hear otherwise.

For `auth.protect`, same thing. It's extremely unlikely someone would import some other function with the same signature, that _didn't_ actually protect.

---

All this is just how I've approached it and my loose thinking on it, not firm opinions by any means!

What are your thoughts? Do you think the correctness argument is worth the complexity?

Ephem: Btw, custom wrappers is likely more important for `auth.protect()` than for these "exit-functions":

```tsx
export async function Page() {
  await checkSignedIn();
}
```

or:

```tsx
async function InnerPage() { ... }
export const Page = withProtect(InnerPage);
```

but again, wanted to keep complexity down to start with.

packages/eslint-plugin/src/next/lib/protection-checks.ts
Ephem: For sure! The first version had a list of things that were allowed before the protect call (like `await params` etc). In the end I tore that out because I think it's hard to explain, and will lead to a lot of "can't you allow this too?" requests. A good case can be made for _only_ allowing `params` and `searchParams`, but immediately you run into `headers()` and `cookies()` etc. After that you run into "why aren't other Clerk functions allowed" like your case 2. and things just snowball (case 2 should just be `const { userId } = await auth.protect();` ofc).

I've always thought we'd bring this back later on, together with some kind of `safeStatements` option to let users configure their own allowed things.

A strict stance seems a lot easier to document and understand. We do need to document things like multiple checks:

```tsx
export async function Page({ params }) {
  // Must be at the top for the lint rule
  await auth.protect();

  const edit = (await params).edit;
  if (edit) {
    // Any extra narrowing checks that rely on other resources happens later,
    // fine to have multiple checks
    await auth.protect({ role: 'admin' });
  }
  
  // ...
}
```

Case 3 - `if (!userId) return null;` is interesting. The reason I didn't support that was that the equivalent client check is _not_ safe, so I didn't want to encourage that pattern:

```tsx
const { userId } = useAuth();

// This will redirect on load and the transitive state
if (!userId)
  return <RedirectToSignIn />

// This is the correct check
if (userId === null) 
  return <RedirectToSignIn />
```

Maybe we should just support it though?

packages/eslint-plugin/README.md
Ephem: This is very much by design. We should document it for sure (but I've intentionally kept the readme small for now).

Resolving paths to urls is surprisingly hard in Next (as we've seen), you also have route intercepts and a ton of other things to deal with. Plus, if we ever want to move to also support folder-based config, sticking to folder over url protection makes everything a ton easier.

The case you mention is actually very easy to reason about. I moved my pages, lint rule says boo, I notice the pattern is not correct and fix it.

Plus, the route groups is a feature. If we strip that by default, I can no longer group routes by `(unauthenticated)` or `(protected)` which is pretty nice:

```tsx
{
  protected: ['app/**'],
  // /sign-in and /sign-up lives here now:
  public: ['app/(unauthenticated)/**'],
}
```

packages/eslint-plugin/src/next/require-auth-protection.ts
Ephem: I'm leaning we punt on this? What you configure is what you get is a fair stance and the way most rules etc work.

packages/eslint-plugin/src/next/lib/protection-checks.ts
Ephem: I implemented this, so return have to be at the top, but I think we should revert that actually.

You have just asserted there is no `userId`, so it's kind of on you to not run unsafe code in that if right? I think strictness here hurts us more than it helps, because you might want to construct some custom UI in that if for example, only allowing early returns makes it too strict to be useful?

packages/eslint-plugin/src/next/require-auth-protection.ts
jacekradko: Agreed, let's punt for now. Long term though, I'd really like to see a warning for paths that aren't covered by any glob. If someone's setup is incomplete I think the rule should at least make them aware of it, otherwise it falls a bit short on giving the user full feedback. This could be configurable, off by default so we don't annoy everybody, but I would love to see something like `strict-mode` where everything needs to be accounted for

packages/eslint-plugin/src/next/lib/protection-checks.ts
jacekradko: Fair points. I feel like this is the tightrope we are trying to walk.. be strict enough to catch most issues, but not too strict to have users just disable the rule or drop the plugin.

jacekradko: So yea, let's revert it. The branch still has to exit at the top level either way, so falling through into protected content stays caught, and that's the strict part worth keeping. What someone does inside their own signed-out branch feels like the right place to be lenient.

packages/eslint-plugin/src/next/lib/protection-checks.ts
jacekradko: This is something we can iterate on in the future.. not a blocker for getting this out. So no, I don't think the correctness is worth the complexity right at this juncture. When we flesh out some of the trusted wrapper ideas, that's probably the natural point to revisit import tracing for the built-ins too, since it would likely be similar/same machinery. Until then maybe just a line in the readme that names are trusted, not verified?

packages/eslint-plugin/src/next/require-auth-protection.ts
Ephem: Yeah it's a good feature for sure! Can probably throw this together pretty quickly in a follow up, just trying to be mindful of scope creep for the initial version.

packages/eslint-plugin/src/next/lib/protection-checks.ts
Ephem: 🤝

I'm not opposed to that complexity in principle, correctness is important for a lint rule and now that I'm adding fixers in a follow up, it's even more so. It's just now initially that I'm trying to keep it down. 😄

> Until then maybe just a line in the readme that names are trusted, not verified?

👍 I've been thinking we can add a section with implementation details at the bottom where we can document specifics like this.

packages/eslint-plugin/src/next/lib/protection-checks.ts
Ephem: I decided this is not the correct place to educate and added support for `!userId` and `!sessionId` as well.

packages/eslint-plugin/src/next/lib/protection-checks.ts
Ephem: Reverted.

jacekradko (approved):
:shipit:

PR #8705, opened 2026-05-29
fix(ui): Align legal consent checkbox in simple theme

dmoerner:
## Description

In the simple theme, the legal consent checkbox would display a large square when wrapped to a second line. This is because it's inside a flex container which defaults align to `stretch`. Explicitly set it to `align='start'` to match the behavior of `CheckboxField`.

Some other themes don't have this problem because they set an explicit height and width. But that approach doesn't seem appropriate for `simple`. This fix also fixes `neobrutalism` which had the same issue.

This doesn't break existing themes because if you already have a height, the default align of `stretch` falls back to `start`. I also verified this visually by checking each theme before and after the change.

Also add `simple` to the Sandbox for easy verification of the fix.

CleanShot 2026-05-29 at 09 23 49@2x

Fixes USER-5441.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/fair-singers-consent.md
+++ b/.changeset/fair-singers-consent.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Fix the legal consent checkbox growing in size when its label wraps to a second line while using the `simple` theme. The checkbox is now aligned to the start of the row so it no longer stretches to match the label height.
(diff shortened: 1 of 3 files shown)


alexcarpenter (approved):
🫶

PR #8706, opened 2026-05-29
feat(ui): Introduce dialog to reset enterprise connection

iagodahlem:
Implements ORGS-1588: a dedicated reset connection dialog and a restyled confirmation step for `<ConfigureSSO />`.

## Summary

- **`<ResetConnectionDialog />`** — new modal-based, type-to-confirm gated dialog that scopes its backdrop to the wizard's content container via `portalRoot`. Wraps the destructive flow behind `useReverification(user.deleteEnterpriseConnection)`, clears the local provider selection on success, and asks the wizard to navigate back to the provider selection step.
- **Confirmation step restyle** — unified `Step.Header` with an inline status badge, grouped Enable SSO + Domain rows, two-column configuration details rendered through `ProfileSection.ItemList`, outlined `Configure again`, destructive `Reset connection`, and a sticky info banner inside the step footer when the connection is inactive.
- **`Step.Header` badge prop** — new `badge?: ReactNode` prop so a step can render an inline status pill next to its title without crowding the existing right-aligned children slot.
- **`<OrganizationProfile />` content ref forwarding** — propagates the shared `ProfileCard` content ref into `<ConfigureSSO />` so the dialog portals into the wizard chrome (not the document body) when the component is embedded under `<OrganizationProfile />`.
- **Unit tests for `<ResetConnectionDialog />`** covering type-to-confirm gating, cancel, and the successful submit chain.

The wizard-wide reset entry (a `Step.Footer.Reset` compound part on Verify Domain / Configure / Test) ships in a follow-up PR — see ORGS-1550.

## Screenshots

_Drag-drop the latest captures from the running sandbox here — active + inactive confirmation, plus the reset dialog open._

| Confirmation — active | Confirmation — inactive | Reset dialog |
| --- | --- | --- |
| CleanShot 2026-06-01 at 10 05 25@2x | CleanShot 2026-06-01 at 10 05 05@2x | CleanShot 2026-06-01 at 10 05 48@2x |

## Test plan

- [ ] **Confirmation — inactive**: title `SSO Successfully configured` + `Inactive` badge inline, grouped Enable SSO + Domain rows, two-column Configuration details with Sign on URL / Issuer / Certificate, outlined `Configure again`, destructive solid `Reset connection`, sticky `(i) SSO is inactive and you need to enable it to authenticate` banner in the footer.
- [ ] **Confirmation — active**: same layout, `Active` badge, no inactive banner.
- [ ] **Dialog open**: clicking Reset opens the dialog centered inside the wizard card; backdrop covers only the card, not the rest of the page; no close `X` in the corner.
- [ ] **Gating**: empty input → Reset button disabled; wrong text → still disabled; the exact organization name → enables.
- [ ] **Cancel + dismiss**: clicking Cancel closes the dialog without side effects.
- [ ] **Successful submit**: type the org name and click Reset → reverification challenge → `user.deleteEnterpriseConnection` runs and the enterprise connections query is invalidated.
- [ ] **Embedded under `<OrganizationProfile />`**: same behavior; the dialog portals into the wizard chrome, not the page body.
- [ ] **Unit tests**: `pnpm vitest run packages/ui/src/components/ConfigureSSO/__tests__/ResetConnectionDialog.test.tsx` green.

Linear: [ORGS-1588](https://linear.app/clerk/issue/ORGS-1588)


.changeset/sso-confirmation-restyle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Restyle the ConfigureSSO confirmation step with a unified header, two-column details layout, dedicated Configure again and Reset connection buttons, and an inactive-state banner.
```
LauraBeatris: We can keep all in the same changeset for a single patch on clerk/ui + use `pnpm run changeset` for the file naming convention 

then separate those items with bullet points

packages/ui/src/components/ConfigureSSO/__tests__/ResetConnectionDialog.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+  goToStep.mockReset();
+  setProvider.mockReset();
+  deleteEnterpriseConnection.mockReset();
+  deleteEnterpriseConnection.mockResolvedValue({} as DeletedObjectResource);
+  goToStep.mockResolvedValue(undefined);
+  connectionMockState.current = { id: 'idn_connection_1' };
+};
+
+describe('ResetConnectionDialog', () => {
+  it('returns null when isOpen is false', async () => {
```
LauraBeatris:
```suggestion
  it('does not render when `isOpen` is `false`', async () => {
```

packages/ui/src/components/ConfigureSSO/__tests__/ResetConnectionDialog.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+      expect(deleteEnterpriseConnection).toHaveBeenCalledWith('idn_connection_1');
+    });
+    expect(setProvider).toHaveBeenCalledWith(undefined);
+    expect(goToStep).toHaveBeenCalledWith('select-provider');
+    await waitFor(() => {
+      expect(onClose).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  it('does not delete when the active enterprise connection is missing', async () => {
```
LauraBeatris: Is there a case for it? Once we select the provider, then the connection gets created 

I thought of not displaying this button if the connection doesn't exist

packages/ui/src/components/ConfigureSSO/ConfigureSSOContext.tsx
```diff
@@ -30,9 +30,10 @@ export interface ConfigureSSOData {
   provider: ProviderType | undefined;
   /**
    * Sets the local provider selection used by Select Provider before a
-   * connection has been created.
+   * connection has been created. Pass `undefined` to clear the selection
+   * (e.g. after deleting the connection from the reset dialog).
    */
-  setProvider: (provider: ProviderType) => void;
+  setProvider: (provider: ProviderType | undefined) => void;
```
LauraBeatris: I think we'll have to adjust this JSDocs based on the new flow 

Once the user selects the provider, then the connection already gets created for that organization

Previously we were relying on verifying the email address, to then create the connection, so I think the comment is just stale

packages/ui/src/components/ConfigureSSO/ResetConnectionDialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+type ResetConnectionDialogProps = {
+  isOpen: boolean;
+  onClose: () => void;
+  /**
+   * The string the user must type to enable the destructive submit button.
+   * Today this is the organization name; long-term it may swap to the
+   * connection's domain — the dialog stays agnostic so callers own the
+   * canonical value.
+   */
```
LauraBeatris:
```suggestion
```

I believe it's fine if we remove this, especially if we want to change later the underlying value used for confirmation - the prop has enough semantics

packages/ui/src/components/ConfigureSSO/ResetConnectionDialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+  );
+};
+
+const ResetConnectionDialogContent = withCardStateProvider((props: ResetConnectionDialogProps) => {
+  const { onClose, confirmationValue } = props;
+  const card = useCardState();
+  const { enterpriseConnection, deleteEnterpriseConnection, setProvider } = useConfigureSSO();
+  const { goToStep } = useWizard();
+
+  const deleteConnection = useReverification((id: string) => deleteEnterpriseConnection(id));
```
LauraBeatris, line 55: Do we need reverification here? I'm not sure if all FAPI mutations rely on it

iagodahlem: Keeping it — `useReverification` wraps every sensitive `user.*` mutation in `@clerk/ui` (delete account, delete passkey, etc.), and the pre-refactor inline `ResetConnectionForm` used it for the same destructive delete call. Skipping it here would diverge from the rest of the component surface.

LauraBeatris: Once we refactor those endpoints to the organization scoped instead of the user scope, then I think those won't be needed

LauraBeatris (approved):

packages/ui/src/components/ConfigureSSO/ResetConnectionDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+                <Form.SubmitButton
+                  block={false}
+                  colorScheme='danger'
+                  isDisabled={!canSubmit}
+                  localizationKey={localizationKeys('configureSSO.resetConnectionDialog.resetButton')}
+                />
+                <Form.ResetButton
+                  block={false}
+                  localizationKey={localizationKeys('configureSSO.resetConnectionDialog.cancelButton')}
+                  onClick={onClose}
```
LauraBeatris: Should we add descriptors to this modal?

iagodahlem: Oh, for sure! 👍

PR #8711, opened 2026-06-01
feat(ui): Add reset connection action on footer

iagodahlem:
Implements ORGS-1550: wires a wizard-wide \"Reset connection\" entry into every ConfigureSSO step footer (except Select Provider / Confirmation) so the reset action is reachable from anywhere in the flow.

**Stacked on top of #8706** — please review/merge that one first. This PR will rebase onto \`main\` after #8706 lands.

## Summary

- Adds a new \`Step.Footer.Reset\` compound part to the local \`Step\` element. It renders a destructive ghost button in the leading edge of the footer (via \`marginInlineEnd: 'auto'\`) and opens the existing \`<ResetConnectionDialog />\` from #8706.
- The slot owns its own \`isOpen\` state and reads the current connection + organization from context — so steps just opt in by rendering \`<Step.Footer.Reset />\` with no extra wiring.
- Gates itself on \`enterpriseConnection\` being defined, so it stays hidden on Select Provider (no connection yet).
- Integrated on:
  - Verify Domain (verified-on-mount, ProvideEmail, EnterVerificationCode footers)
  - Configure (Okta — create app / attribute mapping / assign users / IdP metadata)
  - Configure (Custom SAML — same four sub-step footers)
  - Test
- Adds the \`configureSSOFooterResetButton\` element descriptor + appearance entry.

Reset stays out of the Confirmation step on purpose — that step already has its own destructive Reset button row inside the body (shipped in #8706).

## Test plan

- [ ] **Verify Domain → Configure → Test**: a \`Reset connection\` ghost button appears on the left of every footer.
- [ ] **Select Provider**: no Reset button (no connection yet).
- [ ] **Confirmation**: still uses the in-body destructive button, footer has no Reset.
- [ ] **Click Reset from any step**: opens the same \`<ResetConnectionDialog />\` (modal scoped to the wizard container).
- [ ] **Successful reset**: dialog deletes the connection and revalidates the cache (wizard nav back to Select Provider is racy — covered separately by the wizard refactor).
- [ ] **Embedded under \`<OrganizationProfile />\`**: dialog still portals into the wizard chrome.

Linear: [ORGS-1550](https://linear.app/clerk/issue/ORGS-1550)


packages/ui/src/components/ConfigureSSO/elements/Step.tsx
```diff
@@ -188,11 +192,42 @@ const FooterContinue = ({ onClick, isDisabled, isLoading, label = 'Continue' }:
 };
 FooterContinue.displayName = 'Step.Footer.Continue';
 
+const FooterReset = (): JSX.Element | null => {
+  const { enterpriseConnection } = useConfigureSSO();
+  const { organization } = useOrganization();
```
LauraBeatris: Instead of `useOrganization`, we should use `__internal_useOrganizationBase` to avoid opening the enable orgs prompt

LauraBeatris (approved):

PR #8712, opened 2026-06-01
fix(ui): update the free trial badge spacing and rendering

l-armstrong:
## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

Screenshot 2026-06-02 at 3 29 13 PM

### DIFF
--- a/packages/ui/src/elements/LineItems.tsx
+++ b/packages/ui/src/elements/LineItems.tsx
@@ -124,7 +124,16 @@ const Title = React.forwardRef<HTMLTableCellElement, TitleProps>(({ title, descr
             />
           ) : null}
           <Span localizationKey={title} />
-          {badge}
+          {badge ? (
+            <Span
+              sx={{
+                display: 'inline-flex',
+                marginBlock: '-2px',
+              }}
+            >
+              {badge}
+            </Span>
+          ) : null}
         </Span>
       ) : null}
       {descriptionElements.length > 0 ? (
--- a/packages/ui/src/elements/LineItems.tsx
+++ b/packages/ui/src/elements/LineItems.tsx
@@ -124,16 +124,7 @@ const Title = React.forwardRef<HTMLTableCellElement, TitleProps>(({ title, descr
             />
           ) : null}
           <Span localizationKey={title} />
-          {badge ? (
-            <Span
-              sx={{
-                display: 'inline-flex',
-                marginBlock: '-2px',
-              }}
-            >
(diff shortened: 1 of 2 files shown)


packages/ui/src/elements/LineItems.tsx
```diff
@@ -124,7 +124,16 @@ const Title = React.forwardRef<HTMLTableCellElement, TitleProps>(({ title, descr
             />
           ) : null}
           <Span localizationKey={title} />
-          {badge}
+          {badge ? (
+            <Span
```
dstaley: instead of wrapping in a span with custom styles, I think wrapping in `<Box>` should fix this, since the Box will inherit the height of the flex container, and the Badge will then be allowed to shrink to its intrinsic height.

l-armstrong: So just something like this?

```Javascript
{badge ? (
  <Box>
    {badge}
  </Box>
) : null}
```

dstaley: yup!

mauricioabreu (approved):

wobsoriano (approved):

packages/ui/src/elements/LineItems.tsx
```diff
@@ -124,7 +124,7 @@ const Title = React.forwardRef<HTMLTableCellElement, TitleProps>(({ title, descr
             />
           ) : null}
           <Span localizationKey={title} />
-          {badge}
+          {badge ? <Box>{badge}</Box> : null}
```
alexcarpenter, line 127: @l-armstrong why'd we add a Box wrapper here? This shouldn't be needed from my testing. We are just adding more spacing between the title and the badge?

wobsoriano: didn't see your comment here @alexcarpenter sry

l-armstrong: @alexcarpenter Previously, I tried fixing an issue I noticed with the free trial badge during checkouts when testing in staging. The badge occupied more space than necessary. Initially, I tried fixing it with a span and negative margins, but it was [suggested](https://github.com/clerk/javascript/pull/8712#discussion_r3336273376) by @dstaley to use a Box wrapper instead. 
Screenshot 2026-05-29 at 4 54 45 PM

PR #8715, opened 2026-06-01
refactor(ui): drive ConfigureSSO wizard navigation with a state machine

iagodahlem:
## What

Refactors `<ConfigureSSO />` wizard navigation onto a guard-based state machine (ORGS-1599, from Alex Carpenter's named-guards + reducer-based-navigation feedback). The public component API is unchanged — this is an internal architecture change so navigation derives from connection state instead of imperative per-step routing.

## Why

The previous wizard had four structural problems that would calcify as more providers and DNS/TXT verification land:

1. **Steps routed on data** — navigation decisions lived inside step bodies, not the wizard.
2. **Data was scattered and duplicated** — connection / email / test-run reads were repeated across steps.
3. **Mutations threaded through props with inconsistent reverification** — some sensitive writes were unwrapped.
4. **Footer / submit lifecycle was copy-pasted** per step, patched over with effect sentinels.

## The model

Each step carries one inline `guard: () => boolean` — an *entry precondition* ("may navigation land here now?"). Everything derives from it, uniformly:

- **Initial step (and reload):** the furthest contiguously-reachable step, derived from the connection's state — so a reload lands where the user was.
- **Next / Previous:** sequential; advance or retreat only when the target step's guard holds, otherwise no-op.
- **Stepper:** jumps to a step only when its guard holds — the same predicate gates the click and the disabled state, so they can't diverge.
- **Reset:** deletes the connection; the wizard self-corrects to the furthest-reachable step when the current step's guard breaks — no imperative navigation, no remount. This also heals an external/mid-flow delete.

The reducer is pure and unit-tested without React.

## What's in here

- A single umbrella data hook — connection + domain-derived state + reverification-wrapped mutations + two-tier test-run loading (cold skeleton vs. background table refetch) — gated above the provider so the context never observes loading.
- The generic wizard primitive (pure reducer + machine + `<Wizard.Match>`), with the machine hidden behind `useWizard()`; nested SAML / verify sub-flows preserved.
- All step bodies migrated to the new navigation.

## Scope / follow-ups

- The verify-domain step is unchanged (still email-OTP); the DNS/TXT rewrite and the `domains[]` create-param work land as a follow-up on top of this.
- `@clerk/ui` patch changeset (no public API change).

## Tests

Unit tests cover the reducer (sequential / positional navigation, reachability, self-correction, the nested-flow boundary), the domain derivation, and the data hooks. The flow was manually QA'd end-to-end: initial-state derivation, create → configure, breadcrumb gating, positional back, reload, and reset (including from inside the nested configure sub-flow).

## Summary by CodeRabbit

* **New Features**
  * Improved SSO configuration wizard with smarter step progression based on connection state and configuration completeness.
  * Enhanced enterprise connection test-run management with better pagination and polling controls.

* **Bug Fixes**
  * Fixed step navigation to maintain correct position after page reload.
  * Improved connection reset flow to properly navigate to the appropriate configuration step.

* **Documentation**
  * Updated wizard component documentation to reflect new guard-based navigation behavior.


packages/ui/src/components/ConfigureSSO/ConfigureSSOSteps.tsx
```diff
@@ -0,0 +1,108 @@
...
+ * - `test` is satisfied once there is a successful test run (`tested`); an
+ *   active connection likewise counts it satisfied.
+ * - `confirmation` is the terminal step.
+ *
+ * `verify-domain` is terminal when the domain is taken by another org
+ * (`domainTaken`) — the host suppresses its footer in that state.
+ *
+ * RESET lands on `select-provider`, force-enabled (a just-deleted connection's
+ * stale `hasConnection` must not gate it out).
+ */
```
LauraBeatris: I'd suggest to remove this JSDocs here as it's too verbose, I believe the code is semantic enough

As we keep adding guards, the implementation might change as well which affects the "guard semantics" bit

packages/ui/src/components/ConfigureSSO/ConfigureSSOSteps.tsx
```diff
@@ -0,0 +1,108 @@
...
+const ConfigureSSOStepsGraph = (): JSX.Element => {
+  const { facts } = useConfigureSSO();
+
+  // Named guards the graph references by name, sourced directly from the
+  // umbrella hook's derived state (via context). Each is a trivial boolean read
+  // — the wizard engine only ever sees the resolved value, never the meaning.
+  // An active connection counts configure + test satisfied so a live connection
+  // short-circuits to confirmation (matching the legacy active → confirmation
+  // initial step).
```
LauraBeatris: Same here

packages/ui/src/components/ConfigureSSO/domain/enterpriseConnection.ts
```diff
@@ -0,0 +1,66 @@
...
+ * in — i.e. the domain is already claimed elsewhere.
+ *
+ * Pure relationship between a connection and an organization id; the caller
+ * decides whether this matters (e.g. only once the user's email is verified).
+ * A `null`/`undefined` connection is never "taken".
+ */
+export const isOwnedByOtherOrganization = (
+  connection: EnterpriseConnectionResource | undefined | null,
+  organizationId: string | null | undefined,
+): boolean => Boolean(connection && connection.organizationId !== (organizationId ?? null));
```
LauraBeatris: I believe we can remove this one here - also might need to delete the UI for that on Figma

Reason why, we now are scoping connections per organization, so if the user is a member of that org, there's no case where it would lead to conflicting usage. 

Also if they try to add a domain that it's currently used, it would give an error.

packages/ui/src/components/ConfigureSSO/domain/enterpriseConnection.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Only meaningful once the email backing the domain is verified — an unverified
+ * email can never have its domain "taken", so the verified flag gates the
+ * ownership check. A pure function of its inputs.
+ */
+export const isDomainTakenByOtherOrganization = (
+  connection: EnterpriseConnectionResource | undefined | null,
+  organizationId: string | null | undefined,
+  isEmailVerified: boolean,
+): boolean => isEmailVerified && isOwnedByOtherOrganization(connection, organizationId);
```
LauraBeatris: Same here

.changeset/configuresso-wizard-navigation.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clerk/ui": patch
+---
+
+`<ConfigureSSO />` wizard navigation is now driven by a guard-based state machine, making step navigation more predictable: the step you land on (including after a reload) and which steps you can move to are derived from the connection's state, the connection reset flow lands you on the right step, and enterprise-connection writes stay reverification-wrapped. The public component API is unchanged.
```
LauraBeatris:
```suggestion
Internal refactor for self-serve SSO wizard navigation to leverage a guard-based state machine

It makes the step navigation more predictable: the step you land on (including after a reload) and which steps you can move to are derived from the connection's state, the connection reset flow lands you on the right step.
```

Let's remove `ConfigureSSO` from the changeset and mention self-serve SSO instead + start mentioning that it's an internal refactoring as well

LauraBeatris: `configuresso-wizard-navigation.md` doesn't seem to use the unique generated changeset naming format from `pnpm run changeset` - could we generate with that?

packages/shared/src/react/hooks/useOrganizationEnterpriseConnectionTestRuns.tsx
```diff
@@ -33,6 +34,16 @@ export type UseOrganizationEnterpriseConnectionTestRunsParams = {
    * @default true
    */
   enabled?: boolean;
+  /**
+   * When `true`, a background refetch keeps the previously-loaded page visible
+   * (`isFetching` stays `true`, `isLoading` does not flip back to `true`) instead
+   * of clearing to a cold-load state. Mirrors the `keepPreviousData` option on
+   * `__internal_useOrganizationEnterpriseConnections` so paginating an
+   * existing-at-load list never tears the surrounding UI down to a skeleton.
```
LauraBeatris:
```suggestion
   * When `true`, a background refetch keeps the previously-loaded page visible
   * (`isFetching` stays `true`, `isLoading` does not flip back to `true`) instead
   * of clearing to a cold-load state.
```

packages/ui/src/components/ConfigureSSO/__tests__/ConfigureSSO.test.tsx
```diff
@@ -67,4 +67,67 @@ describe('ConfigureSSO', () => {
       expect(queryByText(/you do not have permission to manage single sign-on/i)).not.toBeInTheDocument();
     });
   });
+
+  describe('state machine mounts on the right step', () => {
```
LauraBeatris, line 71: I think we're missing other states here, such as mounting on the test runs page if the connection is inactive without a successful test

For `mounts on select-provider with a verified email and no connection` we could also leave a TODO comment to replace with a case for TXT verification

packages/ui/src/components/ConfigureSSO/domain/organizationEnterpriseConnection.ts
LauraBeatris, line 1: We could clean it up a couple of comments from this file

packages/ui/src/components/ConfigureSSO/elements/Stepper/types.ts
```diff
@@ -1,17 +1,14 @@
 import type React from 'react';
 
 export interface StepperItemProps {
-  /** The number shown in the bullet — typically the 1-indexed position. Replaced by a check icon when `isCompleted && !isCurrent`. */
+  /** Replaced by a check icon when `isCompleted && !isCurrent`. */
```
LauraBeatris:
```suggestion
  /** The number shown in the bullet */
```

packages/ui/src/components/ConfigureSSO/hooks/useEnterpriseConnectionMutations.ts
```diff
@@ -0,0 +1,117 @@
...
+  deleteEnterpriseConnection: UseOrganizationEnterpriseConnectionsReturn['deleteEnterpriseConnection'];
+}
+
+/**
+ * Returns the {@link EnterpriseConnectionMutations} surface: a stable, memoized
+ * object of {@link useReverification}-wrapped writes. Each function owns only
+ * reverification and the name derivation; the calling UI keeps its own
+ * loading/error lifecycle.
+ */
+export const useEnterpriseConnectionMutations = ({
```
LauraBeatris: I believe we don't need to wrap those mutations with `useReverification` so this module/hook won't be needed 

In FAPI, only the endpoints with `EnsureReverification` middleware, specifically under the `/me` scope that requires it due to `session_reverification_required` error 

Since the enterprise connection methods are under `/organizations` now, then this isn't required

LauraBeatris:
https://github.com/user-attachments/assets/0442f840-d021-4298-8b4c-f38db1a4de69


`goNext` is not moving to the next step after creating the connection, I believe it might be something related to the guards conditions - perhaps with that `useRef`

LauraBeatris:
https://github.com/user-attachments/assets/e073a7db-ace6-4d40-aedc-0f02cbd98599

When successfully adding the IdP metadata, it doesn't navigate to the next step, it requires an extra click on "Continue"

LauraBeatris:
I noticed we aren't pooling for successful test runs anymore but relying on "Refresh logs" only - was this an expected change as part of this PR?

Also on mount we're performing 4 requests


https://github.com/user-attachments/assets/7f07c800-47a9-4bb7-9790-1c0753fd2ecf

LauraBeatris (approved):

PR #8690, opened 2026-05-28
feat(ui): Add Google Workspace SAML to self-serve SSO

LauraBeatris:
## Description

Introduces Google Workspace SAML provider with tailored configure steps 

https://github.com/user-attachments/assets/8a3ffe51-b0ed-4726-9c15-fa4ce6ffb117

#### Unrelated changes/fixes

- Update stepper line-height to fix alignment issue 
- Update test runs query to not refetch on window focus, which was causing the 'Refresh logs' button to display a spinner on mount

## Checklist

- [X] `pnpm test` runs as expected.
- [X] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #8717, opened 2026-06-01
build(shared): nest internal build chunks under _chunks/

jacekradko:
`@clerk/shared` exposes `"./*"`, which resolves to `dist/runtime/*`. Rolldown was emitting its content-hashed shared chunks flat in that directory, so they got picked up as public subpaths and surfaced as phantom breaking changes in break-check whenever a chunk hash shifted (#8690 recommended a spurious `@clerk/shared` major bump off the back of two such chunk renames). This routes the chunks into `dist/runtime/_chunks/` via `outputOptions.chunkFileNames`, so the single-segment wildcard glob only ever matches real entry points.

Nothing changes for consumers: entries reference chunks by relative path, so rolldown just rewrites `./index-<hash>` to `./_chunks/index-<hash>`. `attw` (node16, all green) and `publint` still pass, and a clean build now globs to 50 real subpaths with zero hashed-chunk phantoms.


wobsoriano (approved):
looks sound to me

PR #8720, opened 2026-06-01
fix(expo): Load SSO deps via require to fix Metro async-chunk failure

### LINKED ISSUE #8288: @clerk/expo: useSSO/useOAuth dynamic import() of expo-auth-session/expo-web-browser fails under Metro, masked by empty catch
## Summary

`useSSO` / `useOAuth` in `@clerk/expo` use a dynamic `import()` for `expo-auth-session` and `expo-web-browser` wrapped in an empty `catch {}`. When Metro's async-chunk resolution fails (not uncommon in monorepo / bun setups, or when `@expo/metro-runtime` isn't imported at the entry), every underlying error is swallowed and surfaces as the misleading:

> expo-auth-session and expo-web-browser are required for SSO. Install them: npx expo install expo-auth-session expo-web-browser

…even though both packages are installed, listed in `package.json`, and being bundled by Metro.

## Offending code

`packages/expo/src/hooks/useSSO.ts` (built as `dist/hooks/useSSO.js`):

```js
let AuthSession;
let WebBrowserModule;
try {
  [AuthSession, WebBrowserModule] = await Promise.all([
    import("expo-auth-session"),
    import("expo-web-browser"),
  ]);
} catch {
  return errorThrower.throw(
    "expo-auth-session and expo-web-browser are required for SSO. Install them: npx expo install expo-auth-session expo-web-browser"
  );
}
```

Two problems:

1. **Dynamic `import()` from a published bundle is fragile under Metro.** It forces Metro to emit async chunks; if the runtime loader isn't set up (e.g. `@expo/metro-runtime` not imported), the promise rejects and the real error never surfaces. A synchronous `require()` would be bundled into the main chunk and "just work". This is the same class of issue fixed for the RN QueryClient in clerk/javascript#8087 — dynamic import from inside the published bundle broken under Metro, sync setup resolves it.
2. **`catch {}` hides the real error.** Even when the underlying failure is `Unable to resolve module`, `@expo/metro-runtime is not installed`, or a genuine `TypeError`, users see a generic "install these packages" message that sends them in the wrong direction.

## Environment

- `@clerk/expo@3.1.9`
- Expo SDK 55 (`expo@^55.0.11`, `expo-router@~55.0.10`)
- Metro (default `@expo/metro-config`)
- `expo-auth-session@~55.0.13`, `expo-web-browser@~55.0.14` both installed
- bun workspaces monorepo
- iOS (dev client + simulator)

## Repro

1. Bun-workspaces monorepo, `@clerk/expo` + `expo-auth-session` + `expo-web-browser` all correctly declared in `apps/mobile/package.json`.
2. Don't import `@expo/metro-runtime` at the entry (it's not a declared dep of `@clerk/expo` or `expo-router` in this setup, so it's easy to miss).
3. Call `startSSOFlow({ strategy: 'oauth_google' })` from a screen.
4. Metro logs show both packages bundling successfully:

   ```
   iOS Bundled 321ms …/expo-auth-session/build/index.js (1018 modules)
   iOS Bundled 347ms …/expo-web-browser/build/WebBrowser.js (1145 modules)
   ```

5. At runtime, the dynamic `import()` inside `useSSO` rejects and Clerk throws the "required for SSO" error.

Root cause in our case: `@expo/metro-runtime` wasn't imported at the app entry, so async chunks produced by Metro couldn't be loaded. Adding `import '@expo/metro-runtime'` at the top of the root layout fixed the runtime loading — but only after hours of wrong turns because the Clerk error pointed at package installation, not bundler/runtime.

## Current workaround

We ship a `postinstall` patch that rewrites the dynamic imports to `require()`:

```js
// scripts/patch-clerk-expo-sso.mjs
const SNIPPET_FROM =
  '[AuthSession, WebBrowserModule] = await Promise.all([import("expo-auth-session"), import("expo-web-browser")]);';
const SNIPPET_TO = `AuthSession = require("expo-auth-session");
      WebBrowserModule = require("expo-web-browser");`;
```

Applied to `node_modules/@clerk/expo/dist/hooks/useSSO.js` and `useOAuth.js`. This eliminates the problem entirely on native because Metro inlines both modules into the main bundle instead of emitting async chunks.

## Proposed fix

Either (or both):

1. **Prefer synchronous `require()` on native** (same spirit as clerk/javascript#8087). `expo-auth-session` / `expo-web-browser` are tiny and already listed as optional peer deps — nothing is gained by deferring them. `require()` is Metro-safe and removes the async-chunk failure mode entirely.

2. **At minimum, surface the real error:**

   ```js
   try {
     [AuthSession, WebBrowserModule] = await Promise.all([
       import("expo-auth-session"),
       import("expo-web-browser"),
     ]);
   } catch (e) {
     return errorThrower.throw(
       \`expo-auth-session and expo-web-browser are required for SSO. If they are installed, this usually means Metro failed to resolve the dynamic import() — ensure \\\`@expo/metro-runtime\\\` is imported at your app entry. Underlying error: \${e?.message ?? e}\`
     );
   }
   ```

   The `catch {}` is actively hostile to debugging — it converts every possible failure mode (bundler, runtime, native module, version mismatch) into a single wrong-suggestion error.

Happy to open a PR if the direction is agreed.

## Related

- clerk/javascript#8087 — `fix(expo): synchronous QueryClient for React Native` (same class of dynamic-import-in-Metro fix)

wobsoriano:
## Description

Fixes https://github.com/clerk/javascript/issues/8288

This swaps `useSSO` to synchronous `require()` (the same pattern already used in `ClerkProvider.tsx` for `expo-web-browser`) and surfaces the underlying error instead of swallowing it.

Per Metro's [Module API](https://metrobundler.dev/docs/module-api/), a string-literal `require()` is statically resolved and added to the bundle with no async runtime, while `import()` relies on `asyncRequire` / [`@expo/metro-runtime`](https://docs.expo.dev/guides/customizing-metro/). On native there's no bundle-size cost: the modules ship in the single bundle either way, and Expo's `inlineRequires` keeps evaluation lazy. Same class of fix as [#8087](https://github.com/clerk/javascript/pull/8087).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/expo/src/hooks/useSSO.ts
```diff
@@ -47,17 +47,22 @@ export function useSSO() {
...
+      AuthSession = require('expo-auth-session');
+      // eslint-disable-next-line @typescript-eslint/no-require-imports
+      WebBrowserModule = require('expo-web-browser');
+    } catch (err) {
       return errorThrower.throw(
-        'expo-auth-session and expo-web-browser are required for SSO. ' +
-          'Install them: npx expo install expo-auth-session expo-web-browser',
+        'Unable to load expo-auth-session and expo-web-browser, which are required for SSO. ' +
+          'If they are not installed, run: npx expo install expo-auth-session expo-web-browser. ' +
+          `Original error: ${err instanceof Error ? err.message : String(err)}`,
```
wobsoriano: Keeps the install hint but no longer hides the real error

packages/expo/src/hooks/__tests__/useSSO.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    expect(response.createdSessionId).toBe(null);
+  });
+
+  test('surfaces the underlying error when an auth-session dependency fails to load', async () => {
+    const { result } = renderHook(() => useSSO());
+
+    await expect(result.current.startSSOFlow({ strategy: 'oauth_google' })).rejects.toThrow(
+      /Unable to load expo-auth-session and expo-web-browser.*Original error:/s,
+    );
+  });
```
wobsoriano, line 98: these dont test the actual fix, `import()` vs `require()` is Metro resolution behavior that Vitest/Node can't reproduce but the last test does pin down the behavior change we care about

packages/expo/src/hooks/useSSO.ts
```diff
@@ -47,17 +47,22 @@ export function useSSO() {
...
+      AuthSession = require('expo-auth-session');
+      // eslint-disable-next-line @typescript-eslint/no-require-imports
+      WebBrowserModule = require('expo-web-browser');
+    } catch (err) {
       return errorThrower.throw(
-        'expo-auth-session and expo-web-browser are required for SSO. ' +
-          'Install them: npx expo install expo-auth-session expo-web-browser',
+        `Unable to load expo-auth-session and expo-web-browser, which are required for SSO: ${
+          err instanceof Error ? err.message : 'Unknown error'
+        }. If they are not installed, run: npx expo install expo-auth-session expo-web-browser`,
```
wobsoriano, line 65: Keeps the install hint but no longer hides the real error

wobsoriano:
!snapshot

wobsoriano:
!snapshot

wobsoriano:
!snapshot

wobsoriano:
I tested this locally against the snapshot in an Expo 54 app.

I removed `expo-auth-session` and `expo-web-browser`, deleted `node_modules`/lockfile, reinstalled, and verified both packages were actually absent. With those packages missing, the app still booted successfully while importing `@clerk/expo` normally. The failure only happened when I explicitly exercised the SSO path, where the runtime `require('expo-auth-session')` threw `Cannot find module 'expo-auth-session'`.

Also, the SDK already uses guarded runtime `require()` for optional Expo auth deps in `ClerkProvider` (`expo-web-browser` on web), so this is not a brand new pattern in the package. ([ref](https://github.com/clerk/javascript/blob/1c41b929c36510f291086c9c8fad3c8d9c1e8c7d/packages/expo/src/provider/ClerkProvider.tsx#L379))

Screen capture of the repro:


https://github.com/user-attachments/assets/13709d02-4ac4-4972-974c-f4c495565e53

jacekradko (approved):

PR #8721, opened 2026-06-01
build(repo): drop redundant tsc declaration builds

jacekradko:
| `break-check` failure is expected as the baseline is using the duplicate type tree

These nine packages (react, ui, express, fastify, hono, localizations, nuxt, react-router, tanstack-react-start) emit their type declarations straight from the bundler via tsup/tsdown `dts: true`, so the separate `tsc -p tsconfig.declarations.json` never shipped: release runs only `turbo build`. The tsc step only ran inside the api-changes break-check job, where it redundantly overwrote the bundler's `.d.ts`.

So this drops `build:declarations` and `tsconfig.declarations.json` from all nine, plus react's slot in the root filter and ui's now-stale turbo input. break-check coverage is unchanged: each package's `build` still runs in that job and produces the same dts the snapshot reads. Confirmed locally with attw (node16/esm-only) and publint green on a bundler-only build, every export's type target present.

Deliberately left alone: nextjs, backend, chrome-extension, expo, expo-passkeys, testing, vue and clerk-js, where the bundler emits no declarations and tsc is the actual source (astro builds both and wants its own look). Empty changeset since nothing published changes. Follow-up to #8719.


wobsoriano (approved):
🫰

PR #8727, opened 2026-06-02
feat(js): support new checkout totals shape

mauricioabreu:
Add `discounts` (prorated seat discount + total), `totals_due_per_period` (full recurring renewal breakdown), `total_due_per_period`, and top-level `base_fee` to checkout totals. The top-level totals now reflect only what is actively being purchased, while `totals_due_per_period` conveys the recurring charge. Parse the new fields and surface the prorated discount and renewal totals in the checkout form

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.changeset/checkout-totals-new-shape.md
```diff
@@ -0,0 +1,17 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/shared': patch
+'@clerk/ui': patch
+'@clerk/localizations': patch
+'@clerk/msw': patch
```
dstaley:
```suggestion
```

this is an internal-only package, so we don't need to bump it

packages/shared/src/types/json.ts
```diff
@@ -876,24 +876,101 @@ export interface BillingCreditsJSON {
...
+  /**
+   * Discounts applied to this checkout such as mid-cycle prorated seat discounts.
+   * The key is always present; the value is `null` when no discounts apply.
+   */
+  discounts: BillingCheckoutDiscountsJSON | null;
+  /**
+   * The expected recurring payment for each future billing period.
+   * Kept for backwards compatibility. Prefer `totals_due_per_period` for the full breakdown.
+   */
+  total_due_per_period: BillingMoneyAmountJSON;
```
dstaley, line 966: why are we adding a deprecated field? maybe we just expose `totals_due_per_period`?

packages/shared/src/types/json.ts
```diff
@@ -876,24 +876,101 @@ export interface BillingCreditsJSON {
...
+   * Legacy credit field. Kept for backwards compatibility; prefer the unified `credits` breakdown.
+   */
   credit: BillingMoneyAmountJSON | null;
   credits: BillingCreditsJSON | null;
   account_credit: BillingMoneyAmountJSON | null;
   past_due: BillingMoneyAmountJSON | null;
   total_due_after_free_trial: BillingMoneyAmountJSON | null;
+  /**
+   * Discounts applied to this checkout such as mid-cycle prorated seat discounts.
+   * The key is always present; the value is `null` when no discounts apply.
```
dstaley: not sure we need this disclaimer since we don't have something similar elsewhere

dstaley (approved):
approving, just drop the backwards compatible field and the package bump

packages/shared/src/types/json.ts
mauricioabreu: We are not actually adding it. it existed before my changes, it just moved from line 891 to 867.

packages/shared/src/types/json.ts
mauricioabreu: Agree. removed via 7fab36a8c

PR #8734, opened 2026-06-03
fix(clerk-js): Bypass captcha on OAuth backport for Core 2

dmoerner:
## Description

Backport of the OAuth captcha bypass fix from #8030 to Core 2. `shouldBypassCaptchaForAttempt` bailed out early (therefore showing captcha) when `params.strategy` was unset, which is the case on an OAuth transfer to sign up (the strategy lives on the sign in verification). As a result a captcha was unconditionally requested for brand-new OAuth users. Normally this was a no-op, just wasting cycles to Cloudflare and then our backend would ignore captcha results for bypassed providers anyway. However, we got a report that it was producing a bug on Safari where the frontend would hang indefinitely on /sign-in/sso-callback.

The transfer bypass is now checked before any strategy-based early return, and the redundant strategy-injection block in `create` is removed. Enterprise SSO transfers still correctly show a captcha.

Only the captcha bypass fix is backported; the sign_up_if_missing feature from #8030 is intentionally excluded.

For discussion of the regression introduced in 2024, see https://clerkinc.slack.com/archives/C01KTTYKWBU/p1772566383559539.

Written by Claude under Daniel's supervision.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where new OAuth users could hang on the SSO callback during signup transitions. The system now properly bypasses captcha verification for pre-configured OAuth providers instead of showing an inaccessible captcha, providing a smoother user experience.

### DIFF
--- a/packages/clerk-js/src/core/resources/SignUp.ts
+++ b/packages/clerk-js/src/core/resources/SignUp.ts
@@ -182,13 +182,6 @@ export class SignUp extends BaseResource implements SignUpResource {
       finalParams = { ...finalParams, ...captchaParams };
     }
 
-    if (finalParams.transfer && this.shouldBypassCaptchaForAttempt(finalParams)) {
-      const strategy = SignUp.clerk.client?.signIn.firstFactorVerification.strategy;
-      if (strategy) {
-        finalParams = { ...finalParams, strategy: strategy as SignUpCreateParams['strategy'] };
-      }
-    }
-
     return this._basePost({
       path: this.pathRoot,
       body: normalizeUnsafeMetadata(finalParams),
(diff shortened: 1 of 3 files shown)


dmoerner:
@coderabbitai review

packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -182,13 +182,6 @@ export class SignUp extends BaseResource implements SignUpResource {
       finalParams = { ...finalParams, ...captchaParams };
     }
 
-    if (finalParams.transfer && this.shouldBypassCaptchaForAttempt(finalParams)) {
```
dmoerner, line 185: On the previous code path, this branch would never hit: On a transfer we would have no strategy which would then early return false from `shouldBypassCaptchaForAttempt`. Clean it up, and it was useless even if it did hit, because the backend deliberately ignores the strategy sent with a transfer in favor of fetching the original strategy.

wobsoriano (approved):
looks good 👍🏼 

Im not sure about the release process of Core 2, let's ask @clerk/team-sdk

dmoerner:
Tested with a dummy Core 2 app on Safari.

PR #8741, opened 2026-06-03
ci(repo): Scope core-2 release dispatch to sdk-infra-workers

jacekradko:
The Release (Core 2) job was failing on "Trigger workflows on related repos". `release/core-2` is a maintenance branch, but it was still dispatching `prepare-nextjs-sdk-update.yml` to `clerk/dashboard` and `typedoc-core-2.yml` to `clerk/clerk-docs`, both of which have moved on to Core 3. Those dispatches now return 404 / 403, and since the calls were fire-and-forget promises with no error handling, the rejection bubbled up as an unhandled error and failed the whole job (which `main` doesn't hit because its version of this step was already hardened).

This keeps only the `sdk-infra-workers` clerk-js CDN version bump, `await`s it instead of firing-and-forgetting, and marks the step `continue-on-error: true` so a downstream-notification hiccup can't fail a release that already published to npm.

Failing run: https://github.com/clerk/javascript/actions/runs/26906104288


wobsoriano (approved):
niceee

PR #8747, opened 2026-06-04
chore(clerk-js,shared): Remove deprecated enterprise methods from `User` resource

LauraBeatris:
## Description

This PR removes some enterprise connection methods from the `User` resource, leaving one for retrieval which is currently used on `OrganizationProfile` for account linking.

**There aren't existing GA consumers for those**

The only previous consumer of those were previous versions of `clerk/ui` on `<ConfigureSSO />` internals. Removing those from `clerk-js` won't break existing clients are this component also didn't get released GA. 

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed three enterprise connection management methods from the User resource: `createEnterpriseConnection()`, `updateEnterpriseConnection()`, and `deleteEnterpriseConnection()`. To manage enterprise connections, use the Organization-scoped equivalents instead. The read-only method `getEnterpriseConnections()` remains available. Update your integrations to leverage Organization-based APIs for connection write operations.


packages/shared/src/types/user.ts
```diff
@@ -307,14 +303,6 @@ export interface UserResource extends ClerkResource, BillingPayerMethods {
    */
   leaveOrganization: (organizationId: string) => Promise<DeletedObjectResource>;
   getEnterpriseConnections: (params?: GetEnterpriseConnectionsParams) => Promise<EnterpriseConnectionResource[]>;
```
alexisintech, line 305: getEnterpriseConnections is still there - did you mean to leave it? If so, can you document it, and document the GetEnterpriseConnectionsParams and EnterpriseConnectionResource

and then I will review!

LauraBeatris: Those aren't public methods - we have a feature flagged functionally within `OrganizationProfile` that uses those for a couple of customers

I'll update their JSDocs with `@hidden` and add a description to clarify it

iagodahlem (approved):

PR #8750, opened 2026-06-04
feat(ui): support Electron native OAuth redirects

wobsoriano:
## Description

Adds experimental support for native OAuth redirects in Clerk JS/UI so Electron can use Clerk prebuilt components without forking SignIn, SignUp, or UserProfile connected account flows.

The implementation keeps the existing browser APIs as the public entry point. SignIn and SignUp components still call `authenticateWithRedirect()`, and the SignIn/SignUp resources continue through the existing `authenticateWithRedirectOrPopup` path. When a native OAuth handler is registered on the Clerk instance, that path swaps only the transport layer:

- uses `__internal_nativeOAuthHandler.getRedirectUrl()` as the FAPI redirect URL, for example `myapp://sso-callback`
- calls `__internal_nativeOAuthHandler.open(url)` instead of navigating the current browser context
- reloads the active SignIn/SignUp resource from the returned callback URL
- completes the normal Clerk redirect callback handling through `__internal_handleNativeOAuthCallback`

This PR is intentionally scoped to the Clerk component flows that need OAuth/SSO transport changes for Electron:

- Social OAuth buttons in `SignIn`
- Enterprise SSO in `SignIn`
- Social OAuth buttons in `SignUp`
- Connected account flows in `UserProfile`, including enterprise account connections

Default web behavior is unchanged. If `__internal_nativeOAuthHandler` is not registered, the components continue using the existing browser redirect or popup flow.

The planned `@clerk/electron` SDK will provide the Electron-specific runtime pieces and register this internal handler from its renderer provider, expected to be imported from `@clerk/electron/react`. That SDK layer is responsible for main/preload setup, opening the system browser, waiting for the custom-scheme callback, and secure token persistence. The JS/UI packages only define and consume the internal native OAuth transport contract.

This also adds tests for native OAuth success and cancellation/error handling so provider cancellation can stop the loading state, surface the verification error, and reset stale SignIn/SignUp attempts.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring / dependency upgrade / documentation
- [ ] Other


wobsoriano:
!snapshot

wobsoriano:
!snapshot

wobsoriano:
!snapshot

wobsoriano:
!snapshot

wobsoriano:
!snapshot

wobsoriano:
!snapshot

wobsoriano:
!snapshot

PR #8751, opened 2026-06-04
fix(backend): correct createOrganizationInvitationBulk return type to PaginatedResourceResponse

### LINKED ISSUE #8740: createOrganizationInvitationBulk return type does not match Backend API response shape
### Package + version

`@clerk/backend@3.5.0`

### Description

`clerkClient.organizations.createOrganizationInvitationBulk()` is typed as returning `Promise<OrganizationInvitation[]>`, but the Backend API endpoint it wraps returns an object with a `data` array and `total_count`.

The latest published declaration in `@clerk/backend@3.5.0` still has:

```ts
createOrganizationInvitationBulk(
  organizationId: string,
  params: CreateBulkOrganizationInvitationParams,
): Promise<OrganizationInvitation[]>;
```

The compiled implementation appears to return the raw request result without unwrapping `data`:

```js
async createOrganizationInvitationBulk(organizationId, params) {
  this.requireId(organizationId);
  return this.request({
    method: "POST",
    path: joinPaths(basePath19, organizationId, "invitations", "bulk"),
    bodyParams: params
  });
}
```

However, the Backend API reference for `POST /v1/organizations/{organization_id}/invitations/bulk` shows a response shaped like:

```json
{
  "data": [
    {
      "object": "organization_invitation",
      "id": "string"
    }
  ],
  "total_count": 1
}
```

Docs reference: https://clerk.com/docs/reference/backend-api/tag/organization-invitations/POST/organizations/%7Borganization_id%7D/invitations/bulk

SDK reference currently says `Promise<OrganizationInvitation[]>`: https://clerk.com/docs/reference/backend/organization/create-organization-invitation-bulk

### Expected behavior

Either:

1. the SDK should unwrap and return `OrganizationInvitation[]`, matching the current TypeScript declaration, or
2. the TypeScript declaration and SDK docs should reflect the actual returned object shape, likely something like `Promise<PaginatedResourceResponse<OrganizationInvitation[]>>` or `{ data: OrganizationInvitation[]; totalCount: number }`.

### Actual behavior

Runtime returns an object with a `data` array, while TypeScript says the method returns an array.

This makes code like this type-check but fail at runtime:

```ts
const invitations = await clerkClient.organizations.createOrganizationInvitationBulk(orgId, params);

for (const invitation of invitations) {
  // invitations is not iterable if the runtime response is { data: [...] }
}
```

### Workaround

```ts
const response = await clerkClient.organizations.createOrganizationInvitationBulk(orgId, params) as
  | OrganizationInvitation[]
  | { data: OrganizationInvitation[]; totalCount?: number; total_count?: number };

const invitations = Array.isArray(response) ? response : response.data;
```

### Notes

The response shape shown in the Backend API reference is also consistent with other list-style Clerk responses that expose `data` and count metadata.

VihAMBR:
## Description

Fixes #8740

`clerkClient.organizations.createOrganizationInvitationBulk()` is typed as returning `Promise<OrganizationInvitation[]>`, but the Backend API endpoint it wraps returns a paginated `{ data, total_count }` envelope. The request goes through the same deserializer as every other list endpoint, so at runtime the method already resolves to `{ data, totalCount }` — only the type was wrong, which made the declared type impossible to use correctly.

This aligns the return type with `PaginatedResourceResponse<OrganizationInvitation[]>`, matching the sibling `getOrganizationInvitationList()` method directly above it.

**Verification of the API shape** — confirmed against the Backend API OpenAPI spec rather than assuming. The `CreateOrganizationInvitationBulk` operation's `200` response references the `OrganizationInvitations` component, which is:

```yaml
type: object
properties:
  data:
    type: array
    items:
      $ref: '#/components/schemas/OrganizationInvitation'
  total_count:
    type: integer
```

(Note: this is intentionally different from the instance-level `invitations.createInvitationBulk()`, whose response is a bare array — see #7702 — which is why the two bulk methods have different return types.)

This is consistent with the prior return-type corrections shipped as patches, e.g. `getOrganizationMembershipList()` (#2681) and `JwtTemplatesApi.list` (#7868).

### Note on impact

The runtime value is unchanged; this only corrects the static type. It is technically a type-level breaking change for anyone who wrote code against the old (non-functional) `OrganizationInvitation[]` type — but since that type never matched the value returned at runtime, such code could not have worked. Consumers now access the invitations via `.data` and the count via `.totalCount`, the same as the sibling list methods.

### How to test

A unit test was added at `packages/backend/src/api/__tests__/OrganizationApi.test.ts` that:

- mocks the bulk endpoint returning a `{ data, total_count }` payload,
- asserts the SDK forwards the invitations as a snake_cased array body,
- asserts the response is surfaced as `{ data, totalCount }`, and
- asserts (via `expectTypeOf`) the return type equals `PaginatedResourceResponse<OrganizationInvitation[]>` (this assertion fails type-check without the fix).

Full `pnpm test` for `@clerk/backend` passes across all three environments (node, edge-runtime, miniflare) and `pnpm build` succeeds.

## Checklist

- [ ] (If applicable) JSDoc comments have been added or updated for any package exports

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/fix-bulk-invitation-return-type.md
+++ b/.changeset/fix-bulk-invitation-return-type.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
+Fix the return type of `clerkClient.organizations.createOrganizationInvitationBulk()` to `PaginatedResourceResponse<OrganizationInvitation[]>`. The Backend API returns the bulk-created invitations in a `{ data, totalCount }` envelope (the same shape as `getOrganizationInvitationList()`), but the method was typed as `OrganizationInvitation[]`, which did not match the value returned at runtime.
--- a/packages/backend/src/api/__tests__/OrganizationApi.test.ts
+++ b/packages/backend/src/api/__tests__/OrganizationApi.test.ts
@@ -1,0 +1,73 @@
+import { http, HttpResponse } from 'msw';
+import { describe, expect, expectTypeOf, it } from 'vitest';
+
+import { server, validateHeaders } from '../../mock-server';
+import { createBackendApiClient } from '../factory';
+import type { PaginatedResourceResponse } from '../resources/Deserializer';
+import type { OrganizationInvitation } from '../resources/OrganizationInvitation';
+
+describe('OrganizationAPI', () => {
+  const apiClient = createBackendApiClient({
+    apiUrl: 'https://api.clerk.test',
+    secretKey: 'deadbeef',
+  });
+
+  const organizationId = 'org_123';
+
+  const mockInvitation = {
+    object: 'organization_invitation',
+    id: 'orginv_1',
+    email_address: 'one@example.com',
+    role: 'org:member',
+    role_name: 'Member',
+    organization_id: organizationId,
+    status: 'pending',
+    public_metadata: {},
+    private_metadata: {},
+    url: null,
+    created_at: 1640995200,
+    updated_at: 1640995200,
+    expires_at: 1643673600,
+  };
+
+  describe('createOrganizationInvitationBulk', () => {
+    it('returns a paginated { data, totalCount } response matching the Backend API shape', async () => {
+      server.use(
+        http.post(
+          `https://api.clerk.test/v1/organizations/${organizationId}/invitations/bulk`,
+          validateHeaders(async ({ request }) => {
+            // The SDK should forward the invitations as a snake_cased array body.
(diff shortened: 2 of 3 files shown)


wobsoriano (approved):
thanks for the fix!

PR #8752, opened 2026-06-04
ci(repo): enforce bundlewatch bundle-size budgets

jacekradko:
Adds a bundlewatch budget for `ui.shared.browser.js`, the shared-React variant clerk-js loads when the host app brings its own React, which had no size guard before.

The load-bearing change is making bundle size actually enforced. The bundlewatch step ran with `continue-on-error: true`, so it never failed CI. It now lives in its own `Bundle size` job without that flag, so an over-budget chunk fails a standalone required check. Keeping it a separate job rather than a step in `static-analysis` means a future step can't quietly re-bypass it. No fork conditional is needed: bundlewatch only exits non-zero on a genuine overage, and its GitHub/service calls no-op when the token is absent.

One follow-up has to happen outside the repo: branch protection needs `Bundle size` added as a required check and the old `bundlewatch` commit-status removed. That status is racy in a monorepo, since every passing package writes the same context, a sibling's pass can overwrite a failing one green.

The `web3-solana` budget goes 79KB to 85KB for headroom. It's ~78.6KB in a clean build, so this is drift cushion, not masking a regression.


wobsoriano (approved):
makes sense!

PR #8756, opened 2026-06-05
ci(e2e): harden staging e2e against deterministic and flaky reds

jacekradko:
Fixing the deterministic failures in staging e2e. 

Two always-failing tests simply cannot pass on staging today, and no retry helps. `whatsapp-phone-code` waits on a WhatsApp button the staging instance never renders (the channel isn't enabled there), and `custom-pages` "survives a parent rerender" validates an unreleased `@clerk/react` fix (#8604) while the staging leg installs published `@latest`. Both now skip only where they can't pass (staging, and `E2E_SDK_SOURCE=latest` respectively); PR CI still runs them against branch builds, so coverage there is unchanged.

The change that most deserves a look is concurrency. It was keyed on `ref` (effectively always `main`) with `cancel-in-progress`, so every new staging deploy cancelled the in-flight run and no commit could report its own result. It now keys on the clerk_go commit SHA, falling back to `run_id` for manual dispatches. The rest is a timeout bump above the 25-minute test step, plus a JSON Playwright report upload for later failure classification.

Deliberately out of scope and coming next: making `validate-staging-instances` actually gate, a small `@smoke` gating leg with the full suite informational, and reducing FAPI load on the shared `with-email-codes` instance.

## Summary by CodeRabbit

* **Chores**
  * Improved CI concurrency to avoid canceling distinct staging deploys and added a small changeset entry.
  * Made integration test timeouts configurable with a longer default for CI.

* **Tests**
  * Enhanced CI test reporting by producing JSON reports for non-cancelled runs.
  * Made E2E tests more resilient: added environment-aware skips to avoid deterministic failures on staging and silenced related lint warnings.


wobsoriano (approved):
nice

PR #8762, opened 2026-06-05
fix(clerk-js): Correct locale inconsistencies when creating SignUpFuture

dmoerner:
## Description

We have five different ways to create a `SignUp` with `SignUpFuture` and they have inconsistent handling of params and the browser locale:

- `create`: Uses a param or browser fallback to set locale.
- `password`: This does not apply the browser-locale fallback but does use the param. Fixed in this PR to use the fallback only on POST.
- `web3`: This delegates to `_create` but did not pass along `locale` or first or last name, meaning that it always got the fallback behavior. Fixed in this PR to pass all three params.
- `ticket`: Handles correctly
- `sso`: This doesn't apply any `locale` at all. (It also doesn't take first and last name but since we generally get this from the SSO provider I didn't include it.)

These changes now align our API with the contract in our types `SignUpFutureAdditionalParams` which lists `locale` as a parameter and says we will fall back to the browser's locale.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * `signUp.password()` and `signUp.sso()` now default to the browser locale when creating new sign-ups, matching `signUp.create()`.
  * Explicitly provided locale values are preserved.
  * `signUp.web3()` now forwards `firstName`, `lastName`, and `locale` when creating a sign-up.

* **Tests**
  * Added coverage verifying locale behavior across create vs. update flows and web3 parameter forwarding.

### DIFF
--- a/.changeset/signup-future-password-browser-locale.md
+++ b/.changeset/signup-future-password-browser-locale.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Fix `signUp.password()` on the future SignUp API not defaulting the sign-up's `locale` to the browser locale when it creates a new sign-up, matching the documented behavior and `signUp.create()`. An explicitly passed `locale` still takes precedence, and updates to an existing sign-up remain unaffected.
--- a/.changeset/signup-future-password-browser-locale.md
+++ b/.changeset/signup-future-password-browser-locale.md
@@ -2,4 +2,7 @@
 '@clerk/clerk-js': patch
 ---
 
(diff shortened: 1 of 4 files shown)


wobsoriano (approved):
ahh this makes sense! thanks for adding

dmoerner:
Thanks @wobsoriano for the approval, and thanks to Support Rotation for leading me to look into this code for a customer 👍

PR #8771, opened 2026-06-07
fix(backend): return IdPOAuthAccessToken JWT timestamps in milliseconds

jacekradko:
`IdPOAuthAccessToken.fromJwtPayload` set `expiration`, `createdAt`, and `updatedAt` from the JWT's raw second-based `exp`/`iat`, but `M2MToken` and the API response expose those fields in milliseconds. So the same OAuth token's timestamps came out 1000x off depending on whether you verified it as a JWT or fetched it from the API, and `token.expiration < Date.now()` was always true.

The fix is the `* 1000` conversion that was already being done everywhere else:

```ts
payload.exp * 1000  // was: payload.exp
```

Token validity itself was never affected since the `expired` boolean already used `* 1000`, only the exposed timestamps were wrong. The OAuth-JWT verify test didn't assert these fields, which is why it slipped through, so I added that coverage.

## Summary by CodeRabbit

* **Bug Fixes**
  * OAuth access token verification now returns timestamp fields (expiration, createdAt, updatedAt) in milliseconds instead of seconds, ensuring consistency with other token types and API responses.

### DIFF
--- a/.changeset/idp-oauth-token-ms-timestamps.md
+++ b/.changeset/idp-oauth-token-ms-timestamps.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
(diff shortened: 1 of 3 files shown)


wobsoriano (approved):
thanks for catching!

PR #8772, opened 2026-06-08
feat(backend): validate JWT issuer (iss) claim when an issuer option is provided

jacekradko:
`verifyJwt` never validated the `iss` claim. `VerifyJwtOptions` had no `issuer` field, the `IssuerResolver` type in `jwt/assertions.ts` was unused, and eleven tests in `verifyJwt.test.ts` passed an `issuer:` that nothing read, so they passed regardless. This wires the option up for real: pass a string or a list of strings, the token's `iss` must match one exactly, and mismatches surface as a new `token-invalid-issuer` error. Omitting the option skips the check, the same opt-in shape as `audience` and `azp`.

Two decisions worth scrutiny. The option is `string | string[]` rather than a predicate, because middleware options cross a JSON serialization boundary in Next.js (`encryptClerkRequestData`) where a function value would be silently dropped. And machine token verification now passes an explicit options object to `verifyJwt` instead of spreading the caller's options through, so an `issuer` (or `audience`/`authorizedParties`) meant for session tokens can't reject M2M and OAuth tokens, whose `iss` differs. Both paths have regression tests, and the previously inert tests now exercise the assertion.

## Release Notes

* **New Features**
  * Added optional issuer validation for token verification with `verifyToken()` and `verifyJwt()`
  * Issuer parameter accepts a string or array of strings for flexible token validation
  * Issuer validation is opt-in and backward compatible; omitting it preserves existing behavior
  * Machine tokens (M2M, OAuth access tokens, API keys) are unaffected by issuer validation

### DIFF
--- a/packages/backend/src/jwt/verifyMachineJwt.ts
+++ b/packages/backend/src/jwt/verifyMachineJwt.ts
@@ -52,9 +52,12 @@ async function resolveKeyAndVerifyJwt(
       };
     }
 
+    // Pass only the options declared on JwtMachineVerifyOptions. Callers (e.g. authenticateRequest)
+    // hand us wider option objects whose session-token claim options (issuer, audience,
+    // authorizedParties) must not be asserted against machine tokens, which carry different claims.
     const { data: payload, errors: verifyErrors } = await verifyJwt(token, {
-      ...options,
       key,
+      clockSkewInMs: options.clockSkewInMs,
       ...(headerType ? { headerType } : {}),
     });
 
--- a/.changeset/backend-issuer-validation.md
+++ b/.changeset/backend-issuer-validation.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': minor
+---
+
+`verifyToken()` and `verifyJwt()` now support an `issuer` option to validate a token's `iss` claim. Pass a string for an exact match, or a list of strings of which one must match. Validation is opt-in: when `issuer` is not provided, the `iss` claim is not checked and existing behavior is unchanged. The option only applies to session-token verification; machine tokens (M2M, OAuth access tokens, API keys) are unaffected.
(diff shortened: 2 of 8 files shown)


packages/backend/src/jwt/verifyMachineJwt.ts
```diff
@@ -52,9 +52,12 @@ async function resolveKeyAndVerifyJwt(
       };
     }
 
+    // Pass only the options declared on JwtMachineVerifyOptions. Callers (e.g. authenticateRequest)
+    // hand us wider option objects whose session-token claim options (issuer, audience,
+    // authorizedParties) must not be asserted against machine tokens, which carry different claims.
     const { data: payload, errors: verifyErrors } = await verifyJwt(token, {
-      ...options,
       key,
+      clockSkewInMs: options.clockSkewInMs,
```
dominic-clerk, line 60: **[MEDIUM] Replacing the option spread with an allowlist silently drops `audience` validation for M2M JWTs**

Dropping `issuer` and `authorizedParties` is harmless (machine tokens carry no `azp`, so `assertAuthorizedPartiesClaim` no-ops). Dropping `audience` is not: for Clerk M2M JWTs `aud` is the list of machine IDs the token may call — the minter sets `aud: token.scopes` ("Use JWT audience to represent which machines this token is valid for") and the verify endpoint enforces it (`bapi-proxy` checks `!jwtOpaqueToken.scopes?.includes(authenticatedMachineID)`). JWT-format M2M tokens presented to `authenticateRequest` are verified **locally** and never reach that endpoint, so this option was the only local enforcement: it flowed `createClerkClient({ audience })` → `AuthenticateContext` → `verifyMachineAuthToken` (`tokens/request.ts:820,847`) → here → `assertAudienceClaim`.

Concretely: machines A, B, C; A's key is scoped to C so its tokens carry `aud: ['mch_C']`. Service B calling `authenticateRequest` with `audience: 'mch_B'` previously rejected A's token and now accepts any signature-valid M2M JWT from the instance. `verifyJwt.test.ts` on `main` already pins this ("rejects JWT when array aud does not include the configured audience", added in `ee25cf258`), so the changeset's claim that machine tokens are unaffected isn't accurate.

Suggest keeping `audience: options.audience` in the forwarded set and widening `JwtMachineVerifyOptions` accordingly.

PR #8781, opened 2026-06-09
ci(repo): adopt the official @clerk/break-check action for API checks

jacekradko:
Now that clerk/break-check is public, this swaps the hand-rolled api-changes.yml (snapshot/detect/baseline/comment by hand) for the official Action, pinned to the 0.1.1 commit. Reviewer-visible behavior is unchanged: same tracked packages, still advisory (continue-on-error, fail-on-breaking off), AI reviewer still on with downgrades applied. That downgrade policy moved into break-check.config.json so it travels with the repo, and the Anthropic key is scoped to the detect step, which runs the published CLI rather than PR build code.

The baseline moves from actions/cache to an artifact: publish-baseline uploads a snapshot on every push to main/release/v4/release/core-2, and PRs download the one matching their base.sha (one name covers all three branches). The turbo remote cache is preserved via a setup-command running the filtered build:declarations, with TURBO_* flowing in from init-blacksmith through $GITHUB_ENV. The old turbo-hash skip-gate is dropped, so qualifying PRs always build the head and run detect instead of short-circuiting when no declarations changed.

One edge to know about: a new branch's first PR, before its base has a published baseline (or after the 30-day artifact expiry), makes the Action rebuild the base unauthenticated. That is fine for a public repo, and the job is advisory, so a rare miss does not block merge.

## Summary by CodeRabbit

* **Chores**
  * Improved API change detection workflow and artifact handling for publish/check processes.
  * Enabled AI-assisted break-check behavior (automatic downgrade handling and AI comment support).
  * Added a new release changeset to record versioning/metadata updates.


wobsoriano (approved):
🚀

PR #8783, opened 2026-06-09
fix(shared): pin apiUrlFromPublishableKey return type for deterministic declaration emit

jacekradko:
`apiUrlFromPublishableKey` had an inferred return type, and TypeScript orders an inferred union's members by global type-id rather than by source. Those ids shift whenever unrelated code changes, so break-check kept reporting a phantom reorder of this export's union on PRs that never touched it (#8771 was the latest).

Pinning the return type with an explicit literal union fixes it at the source: the annotation is emitted verbatim, so the order can't drift. I used literal strings rather than `typeof PROD_API_URL | ...` because the `typeof` form pulls a `./constants` import into the `.d.ts`; the literals keep it self-contained, and the function still returns the constants so tsc flags any value drift. The public type is unchanged, so break-check will show one last reorder and then go quiet.

Root cause tracked in clerk/break-check#85 (the differ flags reorder-only union changes); this pins the order at the source until that lands.


wobsoriano (approved):
Makes sense. Let's go break check!

PR #8791, opened 2026-06-09
feat(electron): harden token storage encryption

jeremy-clerk:
Adds guarded async `safeStorage` support for Electron versions that expose it.

Preserves encrypted token entries when decryption is temporarily unavailable or fails, avoids downgrading to plaintext when encryption is available but encryption fails, and adds explicit raw/encrypted storage prefixes with fallback behavior coverage.

### DIFF
--- a/packages/electron/src/storage/index.ts
+++ b/packages/electron/src/storage/index.ts
@@ -4,29 +4,223 @@ import Store from 'electron-store';
 import type { TokenStorage } from '../shared/types';
 
 type StorageOptions = {
+  /**
+   * The name of the file (without extension) used to persist tokens.
+   *
+   * @default 'clerk-tokens'
+   */
   name?: string;
+  /**
+   * The directory in which the token file is stored. Maps to `electron-store`'s `cwd` option.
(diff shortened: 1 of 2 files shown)


packages/electron/src/storage/index.ts
```diff
@@ -4,30 +4,224 @@ import Store from 'electron-store';
...
+  };
+
+  let warned = false;
+  const warnOnce = (message: string) => {
+    if (warned) {
+      return;
+    }
+    warned = true;
+    console.warn(message);
+  };
```
wobsoriano, line 154: Was the local `warnOnce` intentionally scoped to suppress all later storage warnings?

If not, we can use `logger.warnOnce` from `@clerk/shared/logger`, which dedupes by message and still allows distinct warnings

wobsoriano (approved):
Looks good! left a q

packages/electron/src/storage/index.ts
jeremy-clerk: yeah it's scoped for just storage warnings, but I also did not know that existed

PR #8796, opened 2026-06-10
feat(ui,localizations,shared): rename the organization profile SSO page to Security

iagodahlem:
The `<OrganizationProfile />` self-serve SSO surface is becoming a "Security" page that will host more organization security settings over time, mirroring UserProfile's Security tab — SSO is its first section. This PR is the nomenclature half of that reframe:

- Navbar entry renamed "Single Sign-On (SSO)" → "Security" with the shield icon; a new `organizationProfile.navbar.security` localization key replaces `organizationProfile.navbar.selfServeSSO`
- Route path `organization-self-serve-sso` → `organization-security`; the page component is now `OrganizationSecurityPage`
- `<OrganizationProfile.Page label="security" />` reordering now targets this page

The self-serve SSO feature gates (`shouldShowSelfServeSSO`, environment settings) and the ConfigureSSO wizard itself intentionally keep their names — only the hosting page's identity changes. The old localization key and route path are removed rather than aliased since the page is still feature-gated/pre-GA.

ORGS-1606

## Summary by CodeRabbit

* **Chores**
  * Rebranded the Single Sign-On feature section in Organization Profile as "Security" with updated navigation labels, icons, and route paths to better reflect the feature's purpose.

### DIFF
--- a/packages/ui/src/components/OrganizationProfile/OrganizationProfileRoutes.tsx
+++ b/packages/ui/src/components/OrganizationProfile/OrganizationProfileRoutes.tsx
@@ -37,9 +37,9 @@ const OrganizationPaymentAttemptPage = lazy(() =>
   })),
 );
 
-const OrganizationSelfServeSSOPage = lazy(() =>
-  import(/* webpackChunkName: "op-self-serve-sso-page"*/ './OrganizationSelfServeSSOPage').then(module => ({
-    default: module.OrganizationSelfServeSSOPage,
+const OrganizationSecurityPage = lazy(() =>
+  import(/* webpackChunkName: "op-security-page"*/ './OrganizationSecurityPage').then(module => ({
+    default: module.OrganizationSecurityPage,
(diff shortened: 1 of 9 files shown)


packages/ui/src/components/OrganizationProfile/OrganizationProfileRoutes.tsx
```diff
@@ -155,11 +155,11 @@ export const OrganizationProfileRoutes = ({ contentRef }: OrganizationProfileRou
           </Protect>
         )}
         {shouldShowSelfServeSSO ? (
```
LauraBeatris, line 157:
```suggestion
        {shouldShowSecurityPage ? (
```

We only have `organization.self_serve_sso_enabled` for now to determine this, but once we have per org settings such as MFA, etc, we can move this condition to within the page content

LauraBeatris (approved):

PR #8798, opened 2026-06-10
fix(ui): show user:org:read scope in OAuth Consent list

jfoshee:
so that the user is made aware that they are sharing org information with 3rd party

Fixes USER-5442

The OAuth Consent dialog shows a list of scope descriptions so that the user is clearly informed about what they are granting access to.  We were omitting the display of the `user:org:read` scope because we show an org selector in that case.

We determined that is not clear enough to users. Users should be explicitly informed that org information is being shared.

CleanShot 2026-06-10 at 09 50 06@2x

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed OAuth consent scope display to show additional relevant scopes, including the `user:org:read` scope when available.

### DIFF
--- a/packages/ui/src/components/OAuthConsent/__tests__/OAuthConsent.test.tsx
+++ b/packages/ui/src/components/OAuthConsent/__tests__/OAuthConsent.test.tsx
@@ -393,7 +393,7 @@ describe('OAuthConsent', () => {
       });
     });
 
-    it('does not display user:org:read in the scopes list', async () => {
+    it('displays user:org:read in the scopes list', async () => {
       const { wrapper, fixtures, props } = await createFixtures(f => {
         f.withUser({
           email_addresses: ['jane@example.com'],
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/OAuthConsent/__tests__/OAuthConsent.test.tsx
```diff
@@ -410,7 +410,7 @@ describe('OAuthConsent', () => {
       const { queryByText } = render(<OAuthConsent />, { wrapper });
 
       await waitFor(() => {
-        expect(queryByText('Access your organizations')).toBeNull();
+        expect(queryByText('Access your organizations')).toBeVisible();
```
dmoerner, line 413: This isn't the description that was in the screenshot? But the test passes so I'm guessing this is some sort of internal description we also have.

dmoerner (approved):

packages/ui/src/components/OAuthConsent/__tests__/OAuthConsent.test.tsx
jfoshee: yeah it's from a mock setup

wobsoriano (approved):

PR #8804, opened 2026-06-10
fix(express): prevent auth bypass from foreign req.auth values

jacekradko:
`clerkMiddleware()` opened its handler by bailing out early if `req.auth` was already truthy, to avoid re-authenticating when the middleware runs twice. The problem is that `req.auth` is a property name shared with express-jwt, Passport, and express-openid-connect, so in any app that pairs one of those with `@clerk/express`, Clerk authentication was skipped on every request and `getAuth()` returned the other library's unverified value. `requireAuth()` and the `requestHasAuthObject` type guard had the same blind spot.

The fix brands the `req.auth` handler Clerk installs with a `Symbol.for('@clerk/express.auth')` key, and the "did Clerk already run?" checks now gate on that brand rather than on the property's presence:

```ts
export const requestHasAuthObject = (req): req is ExpressRequestWithAuth => {
  const auth = (req as Partial<ExpressRequestWithAuth>).auth;
  return typeof auth === 'function' && auth[clerkAuthBrand] === true;
};
```

A foreign `req.auth` no longer satisfies the guard, so Clerk authenticates the request and overwrites the value (warning once when it does). The public `req.auth(options)` surface is unchanged and the legitimate double-`clerkMiddleware` no-op still holds, since only Clerk's own handler carries the brand. Worth a close look at the three gated call sites: the skip guard in `authenticateRequest.ts`, `getAuth`, and `requireAuth.ts:52`.

Fixes SDK-76.


packages/express/src/authenticateRequest.ts
```diff
@@ -140,10 +148,21 @@ export const authenticateAndDecorateRequest = (options: ClerkMiddlewareOptions =
...
+    // authentication.
+    if (requestHasAuthObject(request)) {
       return next();
     }
 
+    if ('auth' in request) {
+      logger.warnOnce(
+        'Clerk: another middleware has already set `req.auth` on this request. Clerk authentication will run anyway and overwrite it. To use another auth library alongside Clerk, configure it to store its state on a different request property.',
+      );
+    }
```
wobsoriano, line 164: nice 👍🏼

packages/express/src/utils.ts
```diff
@@ -4,8 +4,22 @@ import { Readable } from 'stream';
...
 
+/**
+ * Brand attached to the `req.auth` handler installed by Clerk middleware.
+ * `req.auth` is a property name shared with other auth libraries (express-jwt,
+ * passport, express-openid-connect), so its mere presence proves nothing about
+ * whether Clerk authenticated the request. `Symbol.for` registers the brand
+ * globally so it stays stable across multiple copies of this package loaded in
+ * the same process.
+ */
+const clerkAuthBrand = Symbol.for('@clerk/express.auth');
```
wobsoriano, line 15: 🧠

wobsoriano (approved):
looks good 👍🏼

PR #8177, opened 2026-03-27
chore(repo): Upgrade to TypeScript 6.0

dstaley:
## Description

Updates the repository to TypeScript 6.0, and replaces usage of tsup with tsdown where it was necessary for the packages to build correctly. (There are a few remaining packages using tsup, but those don't break when built with TypeScript 6.0, so I'm saving them for a future PR.)

The majority of this PR was reviewed as [individual PRs](https://github.com/clerk/javascript/pulls?q=sort%3Aupdated-desc+is%3Apr+base%3Ads.chore%2Fts-6.0).

## Checklist

- [ ] `pnpm test` runs as expected.
- [ ] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #8811, opened 2026-06-10
fix(shared): emit unbundled type declarations at stable public paths

jacekradko:
The 2 remaining breaking changes are actually fixes, since they are repairing the broken types, but break-check cannot verify the types are actually reachable so it fails-closed

break-check on #8177 flagged 8 breaking changes: after the tsdown 0.22 bump, rolldown-plugin-dts started code-splitting shared's declarations into content-hashed chunks with mangled aliases, so downstream declaration emit (vue, react, ui, testing) hardcoded specifiers like `@clerk/shared/_chunks/index-Cr_OtBLq`. That subpath is export-blocked, and the hash churns every build. Published `@clerk/shared@4.17.0` has none of this, so it would have shipped broken types in the next release.

The fix splits shared's build in two: the runtime build is unchanged (bundled, `_chunks/` routing intact), and a separate dts-only pass emits declarations unbundled, so every symbol lives at a real-named path that resolves through the existing `./*` export. Two spots deserve scrutiny: the `external` regex in the dts config (bare imports stay external, matching how 4.17.0 publishes, while the `@/*` alias is excluded so it resolves back to relative imports) and the dropped declaration sourcemaps (the plugin cannot emit them in this mode).

Verified by rebuilding every downstream package (zero `_chunks` references remain in any declaration) and by type-checking a consumer app against packed tarballs with `skipLibCheck: false`: `useClerk`, `UserButton`, and `SignInWithMetamaskButton` resolve to real types again, and attw/publint are green.

## Release Notes

* **Bug Fixes**
  * Restored proper TypeScript type resolution for packages dependent on Clerk shared utilities by fixing declaration emission to use stable public paths, preventing silent type degradation.

### DIFF
--- a/.changeset/shared-unbundled-declarations.md
+++ b/.changeset/shared-unbundled-declarations.md
@@ -1,0 +1,5 @@
+---
+'@clerk/shared': patch
+---
+
+Restore resolvable TypeScript declarations. Type declarations are now emitted per-module at stable public paths instead of being bundled into content-hashed internal chunk files. This fixes type resolution failures (or silent `any` degradation) in packages whose declarations reference `@clerk/shared` types, such as `@clerk/vue`, `@clerk/react`, `@clerk/ui`, and `@clerk/testing`, which previously pointed at unresolvable `@clerk/shared/_chunks/*` specifiers.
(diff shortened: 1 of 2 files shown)


wobsoriano (approved):
break check passed 👍🏼

PR #8813, opened 2026-06-10
feat(ui,localizations,shared): add the organization Security page SSO overview

iagodahlem:
The Security page in `<OrganizationProfile />` now renders an SSO overview instead of dropping straight into the configuration wizard:

- **Connection state at a glance** — a status badge (`Unconfigured` / `In Progress` / `Active` / `Inactive`) driven by the connection entity's `status`, with per-state copy and actions (Start configuration / Continue configuration)
- **Enable SSO toggle** — optimistic, gated on a successful test run
- **Inline configuration details** — Provider (icon + name), Domain chips (multi-domain wraps), Sign on URL, Issuer, and Certificate, truncating with ellipsis + full-value tooltips for long SAML values
- **⋯ menu** — Edit (re-enters the wizard) and Delete (type-to-confirm using the organization name; lands back on the overview without a remount)

Start / Continue / Edit switch the page into the existing wizard via local view state — the temporary navigation until the wizard gets its back-to-Security header in the integration follow-up.

Under the hood: the page reads only the umbrella enterprise-connection hook and passes data down as props (the wizard's provider stays internal to the wizard), and `ResetConnectionDialog` is now context-free (hosts pass the delete action and portal root), so the page reuses it outside the wizard. Wizard behavior is unchanged.

States verified against design in light and dark. Known gaps deferred to the wizard-integration follow-up: Edit currently lands on the guard-derived step (the confirmation step for an active connection) rather than forcing step 1, and there's no in-wizard way back to the overview yet.

## Summary by CodeRabbit

* **New Features**
  * Added an SSO security overview to the Organization Profile with status badges (Unconfigured, In Progress, Active, Inactive).
  * Introduced an overview-to-wizard flow for configuring SSO, including provider/domain/sign-on details and contextual actions (edit, activate/deactivate, remove).

* **Localization & Customization**
  * Expanded Security SSO localization (badge states, descriptions, remove confirmation, and action labels).
  * Added new appearance selectors to customize the Security SSO UI.

* **Tests**
  * Added coverage for SSO states, navigation, menus, and remove/activate behaviors.


packages/ui/src/components/ConfigureSSO/elements/Step.tsx
```diff
@@ -229,6 +229,10 @@ const FooterReset = (): JSX.Element | null => {
         isOpen={isOpen}
         onClose={() => setIsOpen(false)}
         confirmationValue={organization?.name ?? ''}
+        // The footer self-hides without a connection (`hasConnection` above), so the resource is set.
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        onDelete={() => mutations.deleteConnection(enterpriseConnection!.id)}
```
LauraBeatris: We could also do `!enterpriseConnection` above so avoid `!` here, what would be the difference with `hasConnection`?

packages/ui/src/components/ConfigureSSO/ResetConnectionDialog.tsx
```diff
@@ -60,18 +66,13 @@ const ResetConnectionDialogContent = withCardStateProvider((props: ResetConnecti
...
 
     try {
-      // Reset is a pure delete — no navigation. Dropping `hasConnection` breaks
-      // the active step's entry guard, so the wizard self-corrects to the
-      // furthest-reachable step. The mutation is already reverification-wrapped.
-      // No `useWizard()` here — that lets this dialog be triggered from ANY
-      // footer (including the nested SAML configure footers) without binding to
-      // a nested wizard.
-      await mutations.deleteConnection(enterpriseConnection.id);
+      // A pure delete, no navigation — the wizard self-corrects once the active step's entry guard breaks.
```
LauraBeatris:
```suggestion
```

packages/ui/src/components/OrganizationProfile/OrganizationSecurityPage.tsx
```diff
@@ -24,28 +30,62 @@ export const OrganizationSecurityPage = ({ contentRef }: OrganizationSecurityPag
...
-        primaryEmailAddress={primaryEmailAddress}
-      />
+      {view === 'overview' ? (
+        <ProfileCard.Page>
+          <Col
+            elementDescriptor={descriptors.page}
+            sx={t => ({ gap: t.space.$8 })}
+          >
+            <Col
+              elementDescriptor={descriptors.profilePage}
```
LauraBeatris, line 57: Do we need this nested `Col` here for layout purposes? Or it is used to apply a different descriptor?

packages/ui/src/components/OrganizationProfile/OrganizationSecurityPage.tsx
```diff
@@ -24,28 +30,62 @@ export const OrganizationSecurityPage = ({ contentRef }: OrganizationSecurityPag
...
+              <Header.Root>
+                <Header.Title
+                  localizationKey={localizationKeys('organizationProfile.securityPage.title')}
+                  sx={t => ({ marginBottom: t.space.$4 })}
+                  textVariant='h2'
+                />
+              </Header.Root>
+              <SecuritySsoSection
+                connection={organizationEnterpriseConnection}
+                enterpriseConnection={enterpriseConnection}
```
LauraBeatris, line 69: Do we need both `organizationEnterpriseConnection` and `enterpriseConnection`? Is it possible to access `enterpriseConnection` within `organizationEnterpriseConnection`?

packages/ui/src/components/OrganizationProfile/SecuritySsoSection.tsx
```diff
@@ -0,0 +1,506 @@
...
+import { useFetchRoles, useLocalizeCustomRoles } from '../../hooks/useFetchRoles';
+import type {
+  OrganizationEnterpriseConnection,
+  OrganizationEnterpriseConnectionStatus,
+} from '../ConfigureSSO/domain/organizationEnterpriseConnection';
+import type { EnterpriseConnectionMutations } from '../ConfigureSSO/hooks/useOrganizationEnterpriseConnection';
+import { ResetConnectionDialog } from '../ConfigureSSO/ResetConnectionDialog';
+import type { ProviderType } from '../ConfigureSSO/types';
+
+/** Props-driven: the Security page owns the data — no wizard context below the page. */
```
LauraBeatris:
```suggestion
```

packages/ui/src/components/OrganizationProfile/SecuritySsoSection.tsx
```diff
@@ -0,0 +1,506 @@
...
+    label: localizationKeys('organizationProfile.securityPage.ssoSection.badge__active'),
+  },
+  inactive: {
+    id: 'inactive',
+    colorScheme: 'danger',
+    label: localizationKeys('organizationProfile.securityPage.ssoSection.badge__inactive'),
+  },
+};
+
+/** Keep in sync with the select-provider step's MONOCHROMATIC_PROVIDER_ICONS. */
```
LauraBeatris:
```suggestion
```

Perhaps a shared constant for the steps? Also think it's not an issue to keep duplicated

LauraBeatris (approved):

LauraBeatris:
CleanShot 2026-06-15 at 14 55 51

Is this horizontal padding expected or do we expect the divider to take the whole space?

LauraBeatris:
"Let members sign in through your identity provider using their domain email" -> "domain email" reads weirdly... I think it should be email domain instead but waiting on confirmation

packages/ui/src/components/OrganizationProfile/OrganizationSecurityPage.tsx
iagodahlem: Yeah, it mirrors the OrganizationGeneralPage (Col descriptors.page → Col descriptors.profilePage); the inner Col carries the profilePage descriptor, so I kept it for consistency.

packages/ui/src/components/OrganizationProfile/OrganizationSecurityPage.tsx
iagodahlem: Yeah, that's a bit confusing. Today, the entity drives the status/menu decisions, and the raw connection supplies the display fields it doesn't carry (domains, samlConnection.*, id). Rather than fold them hastily, I've opened a follow-up to consolidate to a single prop in its own PR so we get the shape right.

packages/ui/src/components/ConfigureSSO/elements/Step.tsx
iagodahlem: No difference, good call! 👍

packages/ui/src/components/OrganizationProfile/SecuritySsoSection.tsx
iagodahlem: 👍 Left it duplicated for now

PR #8817, opened 2026-06-10
feat(ui): add Mosaic hover, motionSafe, and alpha CSS utilities

alexcarpenter:
## Summary

- Adds `packages/ui/src/mosaic/utils.ts` with CSS-in-JS helpers:
  - `hover(styles)` — wraps styles in `@media (hover: hover) { &:hover { ... } }` so hover states only fire on pointer devices (mirrors Tailwind's `hover:` variant, avoids sticky-hover on touch)
  - `motionSafe(styles)` — wraps styles in `@media (prefers-reduced-motion: no-preference)` so animations only run when the user hasn't requested reduced motion
  - `alpha(color, opacity)` — applies opacity to any CSS color string via `color-mix` (not limited to theme token keys)
- `theme.alpha()` in `variables.ts` now delegates to the standalone `alpha()` utility
- Adds ESLint rules scoped to `src/mosaic/` that ban bare `'&:hover'`, `'@media (hover: hover)'`, and `'@media (prefers-reduced-motion: no-preference)'` as object keys — enforcing use of the utilities instead

## Test plan

- [ ] `pnpm --filter=@clerk/ui exec vitest run src/mosaic/__tests__/utils.test.ts`
- [ ] `pnpm eslint packages/ui/src/mosaic` — should report zero `no-restricted-syntax` errors

## Summary by CodeRabbit

* **New Features**
  * Added style utilities for responsive hover behavior, reduced-motion–aware rules, and color opacity mixing.

* **Tests**
  * Added tests covering the new style utilities with varied color and style inputs.

* **Refactor**
  * Theme variables now use the new opacity helper; button hover styling updated to use the new hover helper.

* **Chores**
  * Added initial changeset file.

* **Style**
  * Lint rules added to enforce using the helpers instead of raw hover/motion media.


packages/ui/src/mosaic/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+export function motionSafe(styles: StyleRule): StyleRule {
+  return {
+    '@media (prefers-reduced-motion: no-preference)': styles,
+  };
+}
+
+/** Applies opacity to any CSS color via `color-mix`. */
+export function alpha(color: string, opacity: number): string {
+  return `color-mix(in oklab, ${color} ${opacity}%, transparent)`;
+}
```
maxyinger, line 22: I guess 2 Qs here.

1. did you consider relative colors for fading instead of mix? I think in oklab this will still look good, but it's mixing with black at 0 opacity vs strictly fading the color, so you'll get slightly different results at some of the midpoints. What you have above will still look good cus it's in oklab, but might not exactly match something like figma that's strictly altering opacity.
```
return `oklab(from ${color} l a b / ${opacity})`;
```

~2. I'm guessing this is just more preference, but any apis make you lean towards a percentage vs 0 to 1? I guess I normally assume it's 0 to 1 unless I explicitly pass in a string with %~ 

nvmd. looked below and it's a backwards compat thing with a the mosaic alpha. ignore 2

.changeset/real-chicken-build.md
```diff
@@ -0,0 +1,2 @@
+---
```
maxyinger, line 1: does the changeset just need an empty entry?

maxyinger (approved):
Looks good to me. I'm not sure how to tophat this, but the code all makes sense

.changeset/real-chicken-build.md
alexcarpenter: yeah, since these are don't have any user facing changes atm we can use an empty changeset.

packages/ui/src/mosaic/utils.ts
alexcarpenter: great flag, I have been following the approach tw does.

so `bg-red-500/50` -> `color-mix(in oklab, var(--color-red-500) 50%, transparent)`

maxyinger: ahh that makes sense. Yea probably better to be consistent. feels like a framework level css decision anyway. right on

PR #8832, opened 2026-06-11
ci(e2e): report captcha drift in staging instance validation

jacekradko:
The staging-instance validator has been ignoring `captcha_enabled` and `captcha_widget_type` as "expected to differ", which is exactly how the Turnstile drift that kept the staging `generic` leg red for a week stayed invisible (legal-consent sign-ups blocked on an unsolvable headless captcha, see run 27344100979). Bot protection changes e2e behavior, so it belongs in the comparison.

The script is report-only on main, so this only surfaces drift. A live run currently flags ~11 staging instances with captcha on while their prod counterparts have it off; 5 others have it on in both environments and correctly report as matched.

## Release Notes

* **Tests**
  * Added test coverage for environment validation logic.

* **Chores**
  * Updated environment validation configuration to adjust which settings are compared between staging and production deployments.


wobsoriano (approved):

jacekradko:
Superseded by #8757 (merged as 09054eb962), which carried this change in full so the two could land in either order. Everything here (captcha ignore-removal, isIgnored export, tests) is already on main.

PR #8827, opened 2026-06-11
fix(shared): Added support for global navigator check to isValidBrowser

royanger:
## Checklist

- [x] `pnpm test` runs as expected.
- [x] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed connectivity checks in MV3 extension background service workers so online/offline status is detected correctly; SSR without a navigator still reports offline.
  * Tightened WebAuthn detection to avoid false positives in environments that lack a window object.

* **Tests**
  * Added service-worker-like tests and expanded online/offline coverage for browser detection.

* **Documentation**
  * Updated changelog entry documenting the connectivity heuristic fix.

PR #8840, opened 2026-06-12
fix(shared): exclude server runtimes from the worker navigator fallback

jacekradko:
Follow-up to #8827. While reviewing it we checked how the new worker-scope navigator fallback behaves on server runtimes with worker-like globals. Running the probe in real workerd (via miniflare, old and current compatibility dates) showed Cloudflare Workers expose the `WorkerGlobalScope` constructor but `self instanceof WorkerGlobalScope` evaluates to false, so the merged gate excludes them today only through that quirk of workerd's prototype chain. If workerd ever aligns with the spec (where `ServiceWorkerGlobalScope` inherits from `WorkerGlobalScope`), CF would pass the gate, and since it implements no `navigator.onLine`, `isBrowserOnline` would default to true.

Server worker runtimes self-identify in `navigator.userAgent` (`Cloudflare-Workers`, `Node.js/24`, `Deno/2.5.0`, `Bun/1.3.9`), so the fallback now rejects those before returning a worker navigator. Browser web/service workers (the MV3 case #8827 fixed) are unaffected; tests cover both today's workerd shape and a simulated spec-compliant one.

## Bug Fixes
* Fixed server-side worker environments (Cloudflare Workers, Node.js, Deno, Bun) from being incorrectly identified as valid browsers in runtime validation checks.

## Tests
* Added test coverage for browser detection in worker-like environments to verify server runtimes are properly excluded.


tmilewski (approved):
Good catch

PR #8882, opened 2026-06-16
feat(ui): replace the SSO setup confirmation step with an Activate step

iagodahlem:
## Summary

Replaces the final step of the self-serve SSO setup wizard — previously a static "Confirmation" summary — with an explicit **Activate** step, so configuring a connection and turning it on become distinct, intentional actions.

After the Test step, the user lands on Activate:

- Confirms the connection is configured ("SSO connection configured"), noting that once activated, anyone signing in with the verified domain must use the identity provider
- **Activate SSO** — activates the connection and returns to the Security overview
- **Skip for now** — returns to the overview without activating (they can activate later)

Also renames the wizard step labels to **Domains → Connection → Test → Activate**.

Follows #8864 (Security ↔ wizard navigation).

## Test plan

- [ ] Configure and test a connection, land on Activate, click **Activate SSO** → connection becomes active, returns to the Security overview
- [ ] On Activate, click **Skip for now** → returns to the overview, connection stays inactive
- [ ] `ConfigureSSO` + `OrganizationSecurityPage` unit suites pass

## Summary by CodeRabbit

* **New Features**
  * Added an explicit **Activate** step to the SSO setup flow after configuration and testing, offering **Activate SSO**/**Skip for now**, plus an already-active state with **Done**.
  * Updated wizard step labels to **Domains → Connection → Test → Activate**.
  * Added theming hooks for the SSO **Activate** UI (title/subtitle and button labels).
* **Refactor**
  * Replaced the prior confirmation-style experience with the interactive activation flow.
* **Tests**
  * Updated and added navigation/component tests to cover activation and already-active behavior.


packages/ui/src/components/ConfigureSSO/steps/ActivateStep.tsx
```diff
@@ -0,0 +1,126 @@
...
+        <Step.Body>
+          <Step.Section
+            elementDescriptor={descriptors.configureSSOActivate}
+            fill
+            gap={5}
+            sx={{ alignItems: 'center', justifyContent: 'center' }}
+          >
+            <Col
+              align='center'
+              sx={t => ({ textAlign: 'center', maxWidth: '332px', gap: t.space.$3x5 })}
```
LauraBeatris: I don't see references on the codebase of relying on `px` as unit for `maxWidth`, perhaps we should rely on `sizes` from the theme variable or REM

LauraBeatris (approved):
We just need to follow up with product design to get the UI prototype updated for when the user lands on this screen with an already enabled connection

PR #8885, opened 2026-06-16
fix(shared): drop docsUrl for Enstall OAuth provider

SarahSoutoul:
This PR is paired with a [`dashboard` PR](https://github.com/clerk/dashboard/pull/9497).

### Problem

The JS SDK was shipping a `docsUrl` for Enstall that points to a docs page that does not exist: `https://clerk.com/docs/authentication/social-connections/enstall`. That link surfaces to end users as a 404.

Enstall is a private, single-customer OIDC provider rather than a general public social connection, so we should not ship a public docs link for it.

This PR removes the public `docsUrl` for the Enstall OAuth provider from `clerk/javascript`.

### Changes

- Made `docsUrl` optional on `OAuthProviderData`
- Removed `docsUrl` from the Enstall provider entry in:
  - `packages/shared/src/oauth.ts`
  - `packages/shared/src/types/runtime-values.ts`
- Audited current shared/UI consumers in this repo and did not find any UI surface that assumes `docsUrl` is always present for OAuth providers

### Notes

- This is the SDK-side follow-up only for Enstall. 
- The long-lived docs redirect added [here](https://github.com/clerk/clerk-docs/pull/3391) still needs to remain in `clerk-docs` for already-shipped SDK versions that still contain the old URL

## Summary by CodeRabbit

* **Bug Fixes**
  * OAuth provider `docsUrl` is now optional, allowing provider metadata without documentation links.
  * Removed the documentation URL for the Enstall OAuth provider so it no longer returns `docsUrl` in its provider data.

* **Release Management**
  * Updated the package changeset to publish a patch/minor version bump for `@clerk/shared`.


SarahSoutoul:
@manovotny a bit worried about this warning. Making `OAuthProviderData.docsUrl` optional fixes the Enstall 404 cleanly on the SDK side, but it also changes a public `@clerk/shared` type from `string` to `string | undefined`, which the analyzer is classifying as a breaking change. That would imply a major bump just for this fix, which feels heavier than what we want here.

Wanted to sanity-check whether we’re okay treating this as a real breaking change, or if we'd rather preserve the current type shape and use a safer fallback for Enstall instead? 

Screenshot 2026-06-16 at 5 44 00 pm

wobsoriano:
> @manovotny a bit worried about this warning. Making `OAuthProviderData.docsUrl` optional fixes the Enstall 404 cleanly on the SDK side, but it also changes a public `@clerk/shared` type from `string` to `string | undefined`, which the analyzer is classifying as a breaking change. That would imply a major bump just for this fix, which feels heavier than what we want here.
> 
> Wanted to sanity-check whether we’re okay treating this as a real breaking change, or if we'd rather preserve the current type shape and use a safer fallback for Enstall instead?
> 
> Screenshot 2026-06-16 at 5 44 00 pm

This is expected result with our new break check tool. But this type is only used by docs correct? Tagging @jacekradko here

packages/shared/src/oauth.ts
```diff
@@ -161,7 +161,6 @@ export const OAUTH_PROVIDERS: OAuthProviderData[] = [
     provider: 'enstall',
     strategy: 'oauth_enstall',
     name: 'Enstall',
-    docsUrl: 'https://clerk.com/docs/authentication/social-connections/enstall',
```
manovotny, line 164: I ran a full org-wide sweep of `docsUrl` / `OAUTH_PROVIDERS` consumers (in-repo + `gh search code --owner clerk`), and there is the one spot worth a second look — everything appears safe. 

`clerk/cli` reads `OAUTH_PROVIDERS[].docsUrl` but defensively (`?.docsUrl ?? fallback`), and the in-repo UI consumers only read `.strategy`/`.provider`/`.name`, never `.docsUrl`, so we're good here.

The tricky one is `clerk/dashboard`. Its `getSocialSettings.ts` imports `OAUTH_PROVIDERS` from `@clerk/shared/oauth` and reads `item.docsUrl` with a name-based fallback:

```ts
item.docsUrl ?? `https://clerk.com/docs/guides/configure/auth-strategies/social-connections/${item.provider}`
```

Then, it renders that as the provider's docs link on the SSO connection page. 

With `docsUrl` gone, Enstall falls through to `…/guides/configure/auth-strategies/social-connections/enstall`, which currently **404s** — whereas the URL being removed here **308-redirects** to the live social-connections overview (via the redirect added in clerk-docs#3391). So on the Enstall-enabled instance, this swaps a working (redirected) link for a hard 404 on the surface that actually renders it.

Since the dashboard is what surfaces this link, should we pair this with a dashboard change that drops the link when `docsUrl` is absent (or skips the fallback for private providers), or add a clerk-docs redirect for the new `…/auth-strategies/social-connections/enstall` path so the fallback resolves?

SarahSoutoul: Nice catch! Made a sibling [`dashboard` PR](https://github.com/clerk/dashboard/pull/9497). @manovotny

SarahSoutoul:
> This is expected result with our new break check tool. But this type is only used by docs correct? Tagging @jacekradko here

Makes sense! But I should prob make it a major bump then? @wobsoriano

manovotny (approved):
I am good with the changes here and in https://github.com/clerk/dashboard/pull/9497 (this PR needs to land first). 

I would appreciate another set of eyes and approval before merge to ensure we're not overlooking anything else.

SarahSoutoul:
> > @manovotny a bit worried about this warning. Making `OAuthProviderData.docsUrl` optional fixes the Enstall 404 cleanly on the SDK side, but it also changes a public `@clerk/shared` type from `string` to `string | undefined`, which the analyzer is classifying as a breaking change. That would imply a major bump just for this fix, which feels heavier than what we want here.
> > Wanted to sanity-check whether we’re okay treating this as a real breaking change, or if we'd rather preserve the current type shape and use a safer fallback for Enstall instead?
> > Screenshot 2026-06-16 at 5 44 00 pm
> 
> This is expected result with our new break check tool. But this type is only used by docs correct? Tagging @jacekradko here

@jacekradko @wobsoriano could I follow up on this?

wobsoriano (approved):
Given that this is internal, we can disregard the API breaking change break check 👍🏼

NWylynko (approved):

PR #8887, opened 2026-06-17
fix(eslint-plugin): wrap parenthesized concise arrow bodies without leaving stray parens

jacekradko:
`insertProtectCallFix` wrapped a concise arrow body with `fixer.replaceText(body, ...)`, but a body node's range excludes the parentheses around it. So `() => ({ ok: true })` and `() => (<div/>)` came out as `() => ({ ...block... })`, which doesn't parse. Through the bulk `clerk-next-fix-auth-protection` runner (which applies suggestions by raw splice, no parse guard) that means a silent write of broken source, reported as a clean fix, with the protection dropped.

The fix replaces the whole span after `=>` instead of just the body node, so the wrapping parens go with the old body. Added RuleTester cases for the parenthesized object and JSX shapes, and confirmed end to end that the CLI now rewrites both to a valid `async () => { await auth.protect(); return ...; }` and that a re-run is a no-op.

Empty changeset since the rule is still unreleased on this branch; the existing feature changeset already covers it.


Ephem (approved):
Nice find! Thanks for tackling the fix. 😄

PR #8890, opened 2026-06-17
fix(react-router): scope getAuth to the request to prevent cross-user auth bleed

jacekradko:
clerkMiddleware stores each request's auth on the React Router `context` and getAuth reads it back. Apps that return a single shared `RouterContextProvider` from a custom server or `getLoadContext` reuse one context across every request, so under concurrency a request can be served another user's auth. I reproduced it end to end (a real RR v7 app with `v8_middleware` and a custom server sharing one context: ~45% of 1000 concurrent requests got the wrong user) and confirmed `@clerk/backend`'s `authenticateRequest` is not the cause.

getAuth (and `rootAuthLoader`) now re-derive the current request's auth from `args.request` instead of reading a value cached on the context. Identity is a pure function of the request's cookies, so a shared context can never return another user, including across React Router's action-to-loader revalidation (which mints a fresh `Request`). The middleware stashes only identity-free options (keys, urls) on the context for the re-derivation to reuse, the same way `@clerk/nextjs` resolves auth. There is no `node:async_hooks` dependency, so behavior is uniform across Node, Workers, Deno and Bun.

clerkMiddleware also surfaces a context reused across requests: it throws in development and logs once in production, since a shared context can still leak the application's own per-request data.

Worth a look: the re-derivation in `authenticateFromRequest`, and that getAuth re-runs `authenticateRequest` per call (networkless once the middleware has warmed the JWKS cache). Verified end to end: re-derivation holds 0/1000 wrong-user on a shared context where the previous cached read bled ~45%.

(Supersedes the earlier AsyncLocalStorage approach on this branch, per review: re-derivation is fully runtime-portable, needs no new dependency, and matches the pattern our other server SDKs already use.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed potential authentication leakage between users when a React Router context/provider instance is reused across concurrent requests.
  * `getAuth()` and `rootAuthLoader()` now resolve auth from the current request, preventing incorrect auth during overlapping middleware execution and action→loader revalidation.
  * Added a one-time warning when context reuse across requests is detected.
* **Tests**
  * Added/updated test coverage for auth isolation and shared-context warning behavior.
* **Documentation**
  * Updated the patch release note for the fix and the warning behavior.


packages/react-router/src/server/requestAuthStorage.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Router gives each request regardless of how the context object was
+ * constructed, so `getAuth` reads the right user even when the context is
+ * shared. It also survives the action -> loader revalidation hop, where React
+ * Router mints a fresh `Request` object (so a request-keyed map would miss) but
+ * the async scope is unchanged.
+ *
+ * Note: this module imports `node:async_hooks`, matching `@clerk/nextjs`'s
+ * middleware storage. On a runtime without async storage, `getAuth` falls back
+ * to the RR context slot (today's behavior).
+ */
```
nikosdouvlis: Not related to the fix per-se, but lets avoid the comment pollution in `javascript` as we're trying to do in the `cloudflare-workers` repo. Ill add an AGENTS rule accordingly 

(same for the rest of the comments as well)

PR #8896, opened 2026-06-17
fix(react-router): resolve getAuth per-request to prevent cross-user auth bleed

jacekradko:
`@clerk/react-router` derived per-request auth from the React Router `context`: `clerkMiddleware` cached the resolved auth on `args.context` and `getAuth` / `rootAuthLoader` read it back. That is correct only while the context is unique per request. If any layer reuses the context across requests, a custom server, a shared `getLoadContext`, or a serverless adapter that reuses it on a warm instance, one request can be served another user's auth.

This hardens that path. The middleware still resolves auth once, but now caches it keyed by the `Request` object (which the runtime creates per request and an app cannot alias), and `getAuth` / `rootAuthLoader` read that, re-deriving from the request's own cookies only when the `Request` instance differs (React Router's action-to-loader hop). So a reused context can no longer return the wrong user, and token verification and refresh still run once per request. `clerkMiddleware` also logs a one-time warning when it detects a reused context.

Worth reviewing: `resolveRequestState` (the `WeakMap<Request, ...>` memo plus the re-derive fallback) and that the cache key is the `Request`, never the context.

Scope: this protects against a reused context. A layer that reuses the `Request` object itself (stale cookies and all) would be an adapter bug this can't paper over. This came out of a cross-user bleed investigation and is defensive hardening; it is not claimed to be the root-cause fix for any specific incident. Supersedes #8890.

## Release Notes

* **Bug Fixes**
  * Fixed per-request auth resolution to prevent cross-user auth leakage when a shared React Router context is reused across concurrent requests.
  * Updated auth revalidation (including action → loader) so auth is re-derived from the current request.

* **Tests**
  * Expanded concurrency, revalidation, and cache-miss coverage, including “no options bleed” and shared-context warning scenarios.

* **Versioning**
  * Released a patch update for `@clerk/react-router`.


jacekradko:
**Perf note (updated for the per-request memo design)**

Earlier numbers here assumed `getAuth` re-runs `authenticateRequest` on every call. That's no longer how it works: the middleware authenticates once and memoizes the result keyed by the `Request`; `getAuth`/`rootAuthLoader` read from that memo.

Re-measured locally (real RS256 networkless, JWKS warm):

| path | per `getAuth` call | extra `authenticateRequest` |
| --- | --- | --- |
| common case (loader on the Request the middleware saw) | ~0.17µs (WeakMap hit + `toAuth`) | 0 |
| fresh-Request miss (action→loader revalidation) | ~0.5ms one-time, then cached | 1 |

The common path is effectively free and matches the previous cached behavior (both just call `toAuth`). The only added cost vs. before is one re-derive on a fresh `Request` instance, which is then cached, so repeat reads on it are hits again. That ~0.5ms is `loadOptions` rebuild ~0.35ms + RS256 verify ~0.19ms. Negligible for typical apps; the earlier "0.6-1.9ms per request" estimate assumed a re-derive per `getAuth` call, which no longer happens.

nikosdouvlis:
!snapshot rr-auth-bleed

mikecousins:
Reviewed this against a production incident and wanted to confirm the **Fork 3** boundary in `crossUserDiagnostic.test.ts` isn't hypothetical — a real cross-user bleed maps exactly to it.

**Stack:** React Router v7 framework mode, `future.v8_middleware`, `clerkMiddleware()` + `getAuth()`, Vercel serverless (Node) on warm instances with in-instance request concurrency.

**How we localized it to Fork 3:** an in-request guard decoded the `sub` from the request's own `__session` cookie (read straight off `args.request.headers`) and compared it to `getAuth()`'s `userId`. On every bled request those two were **equal and both the *wrong* user**. That:
- rules out **Fork 1** (would need cookie `sub` ≠ resolved `userId`);
- rules out **Fork 2** (also cookie ≠ `userId`; and we don't share a `RouterContextProvider` — no custom server / `getLoadContext` — and a context-identity probe stayed silent);
- leaves **Fork 3** — the loader is handed *another in-flight request's* `Request`, so the cookie and the resolved auth both originate from that request and therefore agree.

Consistent with that, an in-request guard structurally can't catch it (once `args.request` itself is the wrong one, there's no value left with independent provenance), and the only thing that stopped it was removing in-instance concurrency at the adapter layer (disabling Vercel Fluid Compute) — dependency rollbacks within Clerk did not.

So the hardening here is welcome, but for our case it confirms the real fix has to be in the serverless adapter that reuses/aliases the `Request` (or its loader args) across concurrent invocations on a warm instance — exactly the limit you've documented. Thanks for the clear fork breakdown; it made the diagnosis crisp. Happy to share a minimal repro if useful.

PR #8903, opened 2026-06-17
fix(expo): align iOS Google sign-in defaults

mikepitre:
## Summary

Follow-up to #8901 for the two CodeRabbit comments we deferred there.

- Treat blank or whitespace-only iOS Google client IDs as unconfigured so sign-in preserves the existing NOT_CONFIGURED path.
- Default iOS Google One Tap signIn to filter by previously authorized accounts, matching Android and the documented default.

## Testing

- pnpm --filter @clerk/expo format:check
- pnpm --filter @clerk/expo test

## Summary by CodeRabbit

* **Bug Fixes**
  * iOS Google One Tap sign-in now properly validates and rejects blank Google client IDs.
  * Account filtering for iOS Google One Tap now defaults to enabled, prioritizing previously authorized accounts.

### DIFF
--- a/.changeset/bright-lamps-sign.md
+++ b/.changeset/bright-lamps-sign.md
@@ -1,0 +1,5 @@
+---
+'@clerk/expo': patch
+---
+
+Fixes iOS Google One Tap sign-in to reject blank Google client IDs and to default to filtering by previously authorized accounts.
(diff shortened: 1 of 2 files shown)


wobsoriano (approved):
looks 👍🏼

PR #8905, opened 2026-06-17
fix(expo): restore iOS Google sign-in default

mikepitre:
## Summary

Restore the previous iOS Google sign-in default for `filterByAuthorizedAccounts`.

On iOS, callers must now explicitly pass `filterByAuthorizedAccounts: true` to enable this behavior.

## Testing

- pnpm --filter @clerk/expo format:check
- pnpm --filter @clerk/expo test

## Bug Fixes
* Restored iOS Google sign-in default behavior. Sign-in now displays all available accounts instead of filtering to only previously authorized ones. Users can explicitly enable account filtering if needed.

### DIFF
--- a/.changeset/steady-google-signin.md
+++ b/.changeset/steady-google-signin.md
@@ -1,0 +1,5 @@
+---
+'@clerk/expo': patch
+---
+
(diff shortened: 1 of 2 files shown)


wobsoriano (approved):
🚀

PR #8907, opened 2026-06-17
feat(ui): spotlight interactive captcha during sign-in/sign-up

alexcarpenter:
## What

Spotlight the interactive bot-protection challenge during sign-in/sign-up.

When Turnstile escalates to an interactive "Verify you are human" challenge, the start card now brings it to the foreground — collapses + `inert`s the rest (social/form), keeps header/footer/passkey reachable — until solved. Invisible challenges unaffected.

## How

- `CaptchaElement`: new `onInteractiveChange` (MutationObserver on `#clerk-captcha` `maxHeight` surfaces interactive ↔ resolved) + `gapless` (drop out of flow while collapsed, no flex gap gutter).
- Sign-in/up start: hoist captcha to single slot; collapse `main` column (`display:none`) + `inert` while interactive; passkey action relocated outside the collapsed subtree.
- `@clerk/ui` patch changeset.

## Test

- `pnpm vitest run` in `packages/ui` (+ turnstile unit test in `clerk-js`).
- Sandbox: `cd packages/clerk-js && pnpm dev:sandbox`, open `http://localhost:4000/sign-up?captcha=interactive` (or `/sign-in`). `?captcha=interactive` forces the CF test sitekey + inline render + disables OAuth bypass. Click the checkbox → form snaps back on resolve, no gap during challenge. Drop the param for normal invisible captcha.

## Summary by CodeRabbit

* **New Features**
  * Interactive bot-protection (Turnstile) challenges now foreground during sign-in and sign-up by collapsing/inerting other fields and actions until the challenge is solved, while keeping passkey access available.
* **Bug Fixes**
  * Improved inline “captcha spotlight” behavior across older/newer runtimes, and ensured invisible challenges don’t trigger spotlight expansion.
* **Tests**
  * Added regression and component tests for interactive vs. invisible Turnstile flows, spotlight/inert-state behavior in sign-in/sign-up, and interactive-change callback handling in the captcha element.


anagstef (changes requested):
Looks good in general. But I have a main concern:

The implementation on the `CaptchaElement` is hooked on a CSS implementation detail that the `before-interactive-callback` is triggering (max-height value). This is fragile and may silently break in the future. 

I would suggest we go with another signal, like toggling `data-cl-interactive="true"` on `#clerk-captcha` when `before-interactive-callback` is called and then reset its value on the `finally` block.

What do you think?

packages/ui/src/components/SignUp/SignUpStart.tsx
```diff
@@ -464,8 +474,11 @@ function SignUpStartInternal(): JSX.Element {
                   />
                 )}
               </SocialButtonsReversibleContainerWithDivider>
-              {!shouldShowForm && <CaptchaElement />}
             </Flex>
+            <CaptchaElement
+              gapless
+              onInteractiveChange={setCaptchaIsInteractive}
+            />
```
anagstef, line 481: This may be breaking customers that have applied layout CSS changes on the Card.Content component, because it adds one more child to it.

The impact may be really small here, but let's make sure first.

Same applies for `SignInStart.tsx`

anagstef (approved):

anagstef:
!snapshot

PR #8909, opened 2026-06-18
fix(react): Use globalThis instead of bare global in the browser Clerk loader

### LINKED ISSUE #8871: TanStack Start ClerkProvider can fail when global shim is split into lazy chunk
### Description

In a TanStack React Start + Vite production build, `@clerk/tanstack-react-start` can fail during `<ClerkProvider>` startup with:

```text
ReferenceError: global is not defined
    at getClerkJsEntryChunk (...)
```

The emitted browser bundle still reads `global.Clerk` inside `getClerkJsEntryChunk()`, but the code that creates the browser `window.global` alias can be emitted into a different lazy chunk and may not run before the Clerk loader.

This started after updating `@clerk/tanstack-react-start` from `1.3.3` to `1.4.2`.

### Versions observed

- Failing: `@clerk/tanstack-react-start@1.4.2`
  - `@clerk/react@6.9.1`
  - `@clerk/shared@4.17.1`
  - `@clerk/backend@3.7.0`
- Previously working lockfile: `@clerk/tanstack-react-start@1.3.3`
  - `@clerk/react@6.7.3`
  - `@clerk/shared@4.15.0`
  - `@clerk/backend@3.5.0`
- `@tanstack/react-start@1.168.25`
- `@tanstack/react-router@1.170.15`
- `react@19.2.7` / `react-dom@19.2.7`
- `vite@8.0.16`
- `pnpm@11.6.0`

I also inspected `@clerk/react@6.10.0`; it still appears to emit the same `global.Clerk` lookup in the browser loader, so this may still affect the latest TanStack package path.

### Bundle comparison

Both good and bad builds contain `global.Clerk` in the Clerk loader. The important difference is where the `window.global` compatibility side effect lands.

Known-good build (`@clerk/tanstack-react-start@1.3.3`): the same main `index-*` chunk contains both the shim and the Clerk loader:

```js
if (typeof window < "u" && !window.global) {
  window.global = typeof global > "u" ? window : global
}

// same chunk later:
if (!global.Clerk) throw Error("Failed to download latest ClerkJS...")
return global.Clerk
```

Failing build (`@clerk/tanstack-react-start@1.4.2`): the main `index-*` chunk contains the Clerk loader:

```js
if (!global.Clerk) throw Error("Failed to download latest ClerkJS...")
return global.Clerk
```

But the `window.global` shim moved to a lazy `dist-*` chunk:

```js
if (typeof window < "u" && !window.global) {
  window.global = typeof global > "u" ? window : global
}
```

If the main Clerk loader runs before the lazy chunk, `global` is not defined in the browser and Clerk initialization fails.

### Expected behavior

The browser Clerk loader should not depend on an unbound Node-style `global` identifier, or the package should guarantee the `window.global` alias is installed before any `global.Clerk` lookup can run.

### Actual behavior

`getClerkJsEntryChunk()` can execute in the browser before `window.global` exists, causing `ReferenceError: global is not defined` and preventing Clerk from finding/using the global Clerk object.

### Workaround

Installing the alias before app scripts run avoids the crash:

```js
if (typeof window !== "undefined" && !window.global) {
  window.global = typeof global === "undefined" ? window : global
}
```

But this feels like something the Clerk browser loader or package chunking should handle internally.

### Why I think this may be related to recent bundling changes

The package code has used `global.Clerk` across multiple `@clerk/react` versions, but the production output changed so the shim and the loader are no longer guaranteed to be in the same eagerly-loaded chunk. That makes the previous implicit side effect ordering unreliable.

jacekradko:
Fixes #8871. `<ClerkProvider>` crashed in production builds (TanStack Start, React Router) with `ReferenceError: global is not defined`. The loader read a bare `global.Clerk` that only resolved via the `window.global` polyfill on the package's main entry, but the provider comes in through `@clerk/react/internal`, which never imports that polyfill; since the tsdown migration (#8177) split the package into chunks, a production bundle can reach the loader with the polyfill absent. Using `globalThis` drops the dependency.

The bit worth scrutiny is the type change: `declare const global` only typed the `global` name, so `globalThis.Clerk` needs a `declare global` augmentation. It's intentionally not emitted into the public `.d.ts`, so consumers don't gain an ambient `Clerk` global. The `window.global` polyfill is kept for older clerk-js runtimes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a runtime initialization error that occurred in production environments during ClerkProvider startup, particularly in applications using TanStack Start and React Router. The fix improves compatibility across different bundler configurations and ensures reliable operation in all supported environments.

### DIFF
--- a/.changeset/silly-pumas-cheer.md
+++ b/.changeset/silly-pumas-cheer.md
@@ -1,0 +1,5 @@
+---
+'@clerk/react': patch
+---
+
+Fix `ReferenceError: global is not defined` thrown from the browser Clerk loader during `<ClerkProvider>` startup in production builds (seen in TanStack Start and React Router apps). The loader now reads the Clerk global via `globalThis` instead of relying on a `window.global` polyfill side-effect, whose load order across bundler chunks was not guaranteed.
--- a/packages/react/src/isomorphicClerk.ts
+++ b/packages/react/src/isomorphicClerk.ts
@@ -91,12 +91,10 @@ const SDK_METADATA = {
   environment: process.env.NODE_ENV,
 };
 
-export interface Global {
-  Clerk?: HeadlessBrowserClerk | BrowserClerk;
-  __internal_ClerkUICtor?: ClerkUIConstructor;
+declare global {
+  var Clerk: HeadlessBrowserClerk | BrowserClerk | undefined;
+  var __internal_ClerkUICtor: ClerkUIConstructor | undefined;
 }
-
-declare const global: Global;
 
 type GenericFunction<TArgs = never> = (...args: TArgs[]) => unknown;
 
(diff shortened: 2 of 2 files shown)


wobsoriano (approved):
Nice!

PR #8920, opened 2026-06-18
fix(expo): initialize Android native components without event methods

mikepitre:
## Summary
- Fix Android native component initialization by validating the actual Clerk bootstrap contract instead of assuming every Expo module also exposes React Native event-emitter bookkeeping methods.
- On Android, the Expo module object used for native bootstrap can provide configure/getClientToken/syncFromJsClientToken without addListener/removeListeners. The previous guard treated those event methods as required, rejected the module, and skipped configure, leaving AuthView mounted against an uninitialized native SDK.
- Keep native client event subscription separate from bootstrap validation, so modules without an event-emitter contract are not called as emitters.
- Keep the injected iOS bridge internal to the app target while using the same plain ClerkExpo import style as Expo's generated files.

## Repro
On the native-components quickstart using the published 3.4.6 package, Android can render the AuthView shell without fields because the bootstrap code rejects the native module before calling configure. Packing this branch into that app restores the expected AuthView fields and social buttons.

## Validation
Use the native-components quickstart with a package tarball from this branch, then run clean native prebuilds for both platforms. iOS and Android both build/run, and Android AuthView renders the expected fields/buttons.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Android native component initialization and auth syncing when React Native event listener APIs are unavailable.
  * Improved native module/event handling to support modules that omit listener management methods.
  * Fixed Android client-token hydration and sync behavior after JS token updates.
* **Chores**
  * Updated the iOS native bridge for improved Swift compatibility and import visibility handling.

### DIFF
--- a/.changeset/android-native-client-sync.md
+++ b/.changeset/android-native-client-sync.md
@@ -2,4 +2,4 @@
 "@clerk/expo": patch
 ---
 
-Fix Android native component auth state when JS syncs the current client token back to the native SDK.
+Fix native component auth state when syncing JS and native client changes.
--- a/.changeset/android-native-client-sync.md
+++ b/.changeset/android-native-client-sync.md
@@ -1,0 +1,5 @@
+---
+"@clerk/expo": patch
+---
+
+Fix Android native component auth state when JS syncs the current client token back to the native SDK.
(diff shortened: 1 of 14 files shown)


wobsoriano (approved):
Looks good 👍🏼 

Tested locally in ios and android simulators

Issue #8921, opened 2026-06-18
clerk_active_context cookie is not partition-aware (non-deterministic session-cookie write gate in CHIPS/partitioned cross-origin iframes)


jacekradko:
## Summary

The `clerk_active_context` cookie is the one auth cookie in clerk-js that is **not** partition-aware. In a partitioned / blocked-third-party-cookie context (Chrome CHIPS), it becomes unreadable inside a cross-origin iframe, which silently degrades the cross-tab/cross-context arbitration that gates the session-cookie write.

## Details

`createActiveContextCookie()` sets only `SameSite=None`, with no `Partitioned` attribute and no `usePartitionedCookies` plumbing:

```ts
// packages/clerk-js/src/core/auth/cookies/activeContext.ts
const attributes = { secure: getSecureAttribute('None') };
```

By contrast, the session cookie (and `__client_uat` / dev-browser cookies) compute a `Partitioned` attribute when partitioned cookies are enabled or when running in a cross-origin iframe:

```ts
// packages/clerk-js/src/core/auth/cookies/session.ts
const isPartitioned = options.usePartitionedCookies();
const sameSite = isPartitioned || inCrossOriginIframe() || requiresSameSiteNone() ? 'None' : 'Lax';
const partitioned = isPartitioned && secure;
```

`AuthCookieService.updateSessionCookie()` gates the cookie write on `document.hasFocus()` OR `isCurrentContextActive()`, and `isCurrentContextActive()` reads `clerk_active_context`. In a partitioned cross-origin iframe the session cookie lives in the partitioned jar while `clerk_active_context` does not, so:

- the active-context read returns `undefined` and `isCurrentContextActive()` falls back to `true` (which happens to *unblock* the write today), or
- in configurations where the cookie is readable-but-mismatched, the gate falls closed and drops writes.

Either way the multi-context arbitration the gate is meant to provide does not behave the same inside a partitioned iframe as it does in a first-party tab. The behavior is config-dependent rather than deterministic.

## Impact

Today this is mostly latent because the absent-cookie branch falls open. But it makes the session-cookie write-gate non-deterministic in partitioned/CHIPS embedded contexts, and it is a latent correctness hazard for the multi-tab/multi-context active-organization logic when an app runs inside a cross-origin iframe.

## Suggested fix

[...]

PR #8923, opened 2026-06-18
fix(clerk-js): refresh session cookie on visibilitychange in cross-origin iframes

jacekradko:
Apps running inside a cross-origin iframe (Replit's preview pane is the reported case) could get 401s after the user tabbed away and came back, until a manual page refresh. clerk-js refreshed the session cookie only on the window `focus` event, which never fires inside a cross-origin iframe on tab re-activation, and it gated the cookie write on `document.hasFocus()`, which a visible-but-embedded frame never reports. So the token went stale and the app's refetch-on-focus hit it.

This adds a `visibilitychange` trigger next to the existing `focus` one (visibilitychange does propagate into iframes) and lets a visible cross-origin iframe write the refreshed cookie. The gate change is additive: top-level multi-tab ownership still keys on `document.hasFocus()`, so the active-organization handoff from #3786 is untouched. The subtlety worth a look is the `typeof document !== 'undefined'` guard on the new listener. The `mockNativeRuntime` path constructs `AuthCookieService` with a `window` but no `document`, and an unguarded `document.addEventListener` throws there.

The cross-origin focus/visibility semantics this relies on are pinned down by a real-browser test (`iframe-session-refresh.spec.ts`): a visible cross-origin iframe reports `hasFocus() === false` and only receives `focus` when explicitly clicked into. The behavioral wiring (refresh on `visibilitychange` when visible-but-unfocused-in-iframe) is covered by new `AuthCookieService` unit tests, which had no coverage before. One honest caveat: the full "background the tab until the cookie expires, then watch it heal" leg cannot run in automated Chromium (it won't background a page, and the worker poller keeps the token fresh), so that end-to-end path is being verified manually against a real embedded app.

Follow-up filed as #8921: `clerk_active_context` is not partition-aware, which makes the write-gate arbitration non-deterministic in CHIPS iframes. Left out of scope here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session-cookie refresh so it also occurs when a tab becomes visible again (via `visibilitychange`), reducing stale-token `401` errors.
  * Visible cross-origin embedded iframes can now safely write refreshed session cookies to maintain valid authentication.
* **Tests**
  * Added Playwright integration coverage for focus/visibility behavior in cross-origin iframe scenarios.
  * Added unit coverage to verify cookie refresh and write eligibility for focused vs hidden/visible embedded contexts.


thiskevinwang (approved):
LGTM. Validated with a rather crude test in replit.com:

Before the PR, I "reproed" the broken state in the iframed app.
- Observed the session token refresh
- Tabbed away
- Closed my computer for 5 minutes
- Came back, and saw that the session token never refreshed.
- Even interacting with the iframed app did not refresh the token.

Update the app to PR package fix, and re-tested:
- Observed the session token refresh
- Tabbed away
- Closed my computer for 5 minutes
- Came back, saw that the session token was expired
- Then saw that the session token was refreshed, without interacting with the iframed app

PR #8931, opened 2026-06-19
feat(js): expose loaded @clerk/ui version via Clerk.uiVersion

jacekradko:
Adds `Clerk.uiVersion` so the loaded `@clerk/ui` version is reachable from the `Clerk` object, next to the existing `Clerk.version` (clerk-js). Until now the only way to read it was `window.__internal_ClerkUICtor.version`. The getter returns the UI bundle's version once it has loaded (reading the global the UI bundle publishes, with a fallback to the constructor passed via `options.ui.ClerkUI` for bundled / no-RHC builds), or `undefined` before then.

Because `@clerk/ui` and `@clerk/clerk-js` are versioned and loaded independently, `uiVersion` can differ from `version`. Verified in a real browser through the dev sandbox: `Clerk.uiVersion` returned `1.18.1` while `Clerk.version` was `6.18.1`.

The one thing worth a deliberate look: this lands `uiVersion` on the public `Clerk` interface in `@clerk/shared/types`, so it's a committed member across every SDK. React's `IsomorphicClerk` forwards it and returns `undefined` when running against older clerk-js runtimes that predate the getter.

## Summary by CodeRabbit

* **New Features**
  * Added a `uiVersion` property to expose the currently loaded `@clerk/ui` bundle version, returning `undefined` until the UI has loaded.

* **Tests**
  * Added coverage to verify `uiVersion` presence and correct behavior before and after the UI version becomes available.


.changeset/clerk-ui-version.md
```diff
@@ -0,0 +1,7 @@
+---
+'@clerk/clerk-js': minor
+'@clerk/shared': minor
+'@clerk/react': minor
```
anagstef, line 4: Should it just be a `patch` bump maybe?

anagstef (approved):

PR #8933, opened 2026-06-19
feat(clerk-js,ui): Auto-activate exclusive organization in choose-organization task

NicolasLopes7:
## What

When the `choose-organization` session task fires for a member of an **exclusive-membership** organization, clerk-js now skips the org picker entirely: it shows the loading spinner and auto-calls `setActive` on that org. The generic instance-wide force-choose-org flow is unchanged for everyone else.

Implements the clerk-js half of exclusive membership "after-auth".

## Changes

- **Expose `Organization.exclusiveMembership`** — parse the FAPI `exclusive_membership` field (`@clerk/shared` types + `clerk-js` Organization resource `fromJSON` / `__internal_toSnapshot`). Field is `undefined` for non-adopting instances.
- **`TaskChooseOrganization`** — detect the membership whose `organization.exclusiveMembership === true` (an exclusive member belongs to exactly one org). If found: render only the spinner and auto-`setActive` once (`useRef` single-fire guard), reusing the existing `navigateOnSetActive` / `redirectUrlComplete` wiring. On failure, surface the error via card state and fall back to the normal flows so the user can recover. No exclusive membership → existing behavior, zero change.

## Release Notes

**New Features**
* Organizations can now enforce exclusive membership, restricting users to a single organization at a time
* Users in exclusive-membership organizations are automatically activated during sign-in, eliminating the need to manually select an organization

### DIFF
--- a/.changeset/exclusive-membership-auto-activate.md
+++ b/.changeset/exclusive-membership-auto-activate.md
@@ -1,0 +1,9 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/shared': patch
+'@clerk/ui': patch
+---
+
+When the `choose-organization` session task is triggered for a member of an organization that enforces exclusive membership, Clerk now automatically activates that organization instead of rendering the organization picker. The user sees a loading spinner while the organization is set as active; if activation fails, the regular picker is shown so they can recover. The generic force-choose-organization flow is unchanged for everyone else.
+
+Additionally, `Organization.exclusiveMembership` is now exposed on the Organization resource, reflecting the `exclusive_membership` field returned by the Frontend API.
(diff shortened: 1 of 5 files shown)


.changeset/exclusive-membership-auto-activate.md
```diff
@@ -0,0 +1,9 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/shared': patch
+'@clerk/ui': patch
+---
+
+When the `choose-organization` session task is triggered for a member of an organization that enforces exclusive membership, Clerk now automatically activates that organization instead of rendering the organization picker. The user sees a loading spinner while the organization is set as active; if activation fails, the regular picker is shown so they can recover. The generic force-choose-organization flow is unchanged for everyone else.
+
+Additionally, `Organization.exclusiveMembership` is now exposed on the Organization resource, reflecting the `exclusive_membership` field returned by the Frontend API.
```
LauraBeatris: Let's start with "Introduces organization membership feature" and then explain the feature below 

Also use `pnpm run changeset` and apply `minor` on each package

LauraBeatris (approved):

PR #8935, opened 2026-06-19
feat(clerk-js): Backport interactive captcha spotlight for Core 2

anagstef:
Backports the interactive-captcha spotlight from #8907 to the Core 2 release line.

There is no `@clerk/ui` package in Core 2 — the sign-in/sign-up UI lives entirely inside `clerk-js` (`packages/clerk-js/src/ui`). So the UI changes from the original PR are applied directly to clerk-js's bundled components here, alongside the clerk-js runtime change.

## What it does

When a Cloudflare Turnstile challenge escalates to an interactive "Verify you are human" check, the start card brings it to the foreground: the rest of the card (social buttons, form) collapses and is `inert`-ed until the challenge is solved, while the header, footer, and passkey action stay reachable. Invisible challenges — the vast majority — are unaffected.

## Changes

- **`utils/captcha/turnstile.ts`** — sets a `data-cl-interactive` attribute on `#clerk-captcha` when a challenge becomes interactive (removed on resolve), giving the UI a reliable signal.
- **`ui/elements/CaptchaElement.tsx`** — new optional `onInteractiveChange` and `gapless` props; the MutationObserver now also reads `data-cl-interactive`, with a `maxHeight` fallback for older runtimes. Other render sites that use a bare `<CaptchaElement />` are unaffected.
- **`ui/components/SignIn/SignInStart.tsx`** and **`ui/components/SignUp/SignUpStart.tsx`** — hoist the captcha into a single slot outside the collapsing column and wire up the spotlight.
- **`ui/components/SignUp/SignUpForm.tsx`** — no longer renders its own captcha; the start card's slot owns it now.

## Note

`SignUpForm` is shared by the start and continue flows, so removing its captcha means `SignUpContinue` no longer renders a captcha element. This matches the original PR's behavior — the captcha is only needed on the initial sign-up create.

### DIFF
--- a/.changeset/captcha-interactive-spotlight.md
+++ b/.changeset/captcha-interactive-spotlight.md
@@ -1,0 +1,5 @@
+---
+'@clerk/clerk-js': minor
+---
+
+Spotlight interactive bot-protection (Cloudflare Turnstile) challenges during sign-in and sign-up. When a challenge escalates to an interactive "Verify you are human" check, the start card now brings it to the foreground — collapsing and `inert`-ing the rest of the form until the challenge is solved — while keeping the header, footer, and passkey action reachable. Invisible challenges are unaffected.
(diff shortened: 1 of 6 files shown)


anagstef:
!snapshot

nikosdouvlis (approved):

PR #8941, opened 2026-06-22
fix(eslint-plugin): Fix initial lint rule bugs

Ephem:
## Description

I bundled 3 small fixes into this PR, I recommend reviewing per commit.

* Make fixer wrap existing return type in `Promise<>` when making the function `async`
  * Previously `function Component(): React.ReactElement` would become `async function Component(): JSX.Element` which would fail TS
  * Now becomes `function Component(): Promise<React.ReactElement>`
  * Does not wrap if it's already a Promise
  * `type Return = Promise<...>;` can get double-wrapped via `Promise<Return>` - we can't follow it, but this should be very rare and fine in practice, worst case TS complains and you have to fix it manually
* Make the `import { auth } from "@clerk/nextjs/server"` match the quote style of already existing imports
  * Default to `"` to match Prettier, ESLint and `create-next-app` defaults
* Support `||` in custom checks, like `if (!isAuthenticated || !has(...))`

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * The `require-auth-protection` rule now recognizes authentication guards expressed with OR-conditions (e.g., `if (!isAuthenticated || otherCondition)`).
  * The fixer now respects the file's existing import quote style when adding new imports.
  * Enhanced TypeScript type handling: function return types are wrapped in `Promise<>` when converting functions to `async`.

### DIFF
--- a/packages/eslint-plugin/src/next/__tests__/require-auth-protection.test.ts
+++ b/packages/eslint-plugin/src/next/__tests__/require-auth-protection.test.ts
@@ -360,6 +360,59 @@ ruleTester.run('require-auth-protection', rule, {
       options: [config],
     },
     {
+      name: 'manual check: !isAuthenticated || otherCondition with return',
+      code: `
+        import { auth } from '@clerk/nextjs/server';
+        export default async function Page() {
+          const { isAuthenticated } = await auth();
+          if (!isAuthenticated || someFlag) return null;
+          return <div />;
+        }
+      `,
+      filename: abs('app/dashboard/page.tsx'),
+      options: [config],
+    },
+    {
+      name: 'manual check: otherCondition || !isAuthenticated with return',
+      code: `
+        import { auth } from '@clerk/nextjs/server';
+        export default async function Page() {
+          const { isAuthenticated } = await auth();
+          if (someFlag || !isAuthenticated) return null;
+          return <div />;
+        }
+      `,
+      filename: abs('app/dashboard/page.tsx'),
+      options: [config],
+    },
+    {
+      name: 'manual check: userId === null || otherCondition with redirect',
+      code: `
+        import { auth } from '@clerk/nextjs/server';
+        import { redirect } from 'next/navigation';
+        export default async function Page() {
+          const { userId } = await auth();
+          if (userId === null || someFlag) redirect('/sign-in');
+          return <div />;
+        }
+      `,
+      filename: abs('app/dashboard/page.tsx'),
+      options: [config],
+    },
+    {
+      name: 'manual check: !isAuthenticated || a || b with return',
+      code: `
+        import { auth } from '@clerk/nextjs/server';
+        export default async function Page() {
+          const { isAuthenticated } = await auth();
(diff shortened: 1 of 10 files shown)


packages/eslint-plugin/src/next/lib/fixers.ts
```diff
@@ -120,19 +118,48 @@ function getLineIndent(sourceCode: TSESLint.SourceCode, node: TSESTree.Node): st
...
     return null;
   }
-  const firstToken = sourceCode.getFirstToken(fn);
-  if (!firstToken) {
+  const typeAnnotation = returnType.typeAnnotation;
+  // If return type is a predicate like `value is boolean`, we still add `async`,
+  // but don't edit the return type. This will produce invalid TS so user can
+  // go fix it. Should be exceedingly rare.
+  if (typeAnnotation.type === 'TSTypePredicate' || isPromiseTypeAnnotation(typeAnnotation)) {
     return null;
```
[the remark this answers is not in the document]
Ephem: This is intentional. This should be an exceedingly rare case for the type of resources we check, if it exists at all. If we exclude running the fix entirely, we now have to deal with listing these skipped files in the bulk auto-fixer etc, complexity that was not deemed to be worth it.

Generating broken TS here forces the user to look at it and fix it anyway, which should be a fine stance when it's a pattern that pretty much does not exist in real code.

packages/eslint-plugin/src/next/__tests__/require-auth-protection.test.ts
```diff
@@ -359,6 +359,59 @@ ruleTester.run('require-auth-protection', rule, {
       filename: abs('app/dashboard/page.tsx'),
       options: [config],
     },
+    {
+      name: 'manual check: !isAuthenticated || otherCondition with return',
+      code: `
+        import { auth } from '@clerk/nextjs/server';
+        export default async function Page() {
+          const { isAuthenticated } = await auth();
+          if (!isAuthenticated || someFlag) return null;
```
jacekradko, line 368: Directional question.. should we allow mix of auth/not-auth conditional logic here or should it be strictly auth.* || auth.* ?

jacekradko (approved):
Looks good. Just an exploratory question on logical checks

packages/eslint-plugin/src/next/__tests__/require-auth-protection.test.ts
Ephem: This is a good question and I asked myself the same, but I don't see why not? As long as it _always_ happens for `!isAuthenticated` I think it's fine.

There is a small risk of `if (!isAuthenticated || doDangerousThing())`, but I feel that should be rare and obvious enough that DX wins here. We already allow arbitrary code above the early exit now which is not exactly the same (we are already in an "unauthed state" there, which we are not when evaluating the if), but close enough?

PR #8960, opened 2026-06-23
feat(expo): add hosted auth flow

mikepitre:
## Summary

Adds the Expo client SDK surface for hosted auth through Account Portal.

Expo apps can now start a hosted sign-in or sign-up flow in the system browser, return through a native callback, and activate the single Clerk session that ClerkGo attaches to the native client.

## What Changed

- Adds `useHostedAuth()` through the opt-in `@clerk/expo/hosted-auth` subpath.
- Opens Account Portal with `expo-web-browser` / `expo-auth-session`.
- Registers the canonical `clerk://<android.package>.callback` intent filter through the Expo config plugin.
- Generates and validates `state` for callback correlation.
- Generates a PKCE verifier/challenge pair and sends only the challenge when creating the hosted-auth attempt.
- Redeems the returned rotating token nonce from the Expo hosted-auth helper.
- Applies the returned `/client` payload to the existing Clerk client and activates the created native session. The payload is applied before the created session is validated, so local client state stays in sync with the rotated server state even when validation fails.
- If an abandoned handoff leaves the native client stale, retries hosted-auth creation once after the response hook stores ClerkGo's replacement device token.
- Rejects overlapping `startHostedAuth` calls while a flow is in progress.
- Warns once in development builds when the default Android callback is used, since it only works after the Clerk config plugin registers the intent filter and the native project is rebuilt. The hook docs spell out the same prerequisite.

## Developer Experience

```ts
import { useHostedAuth } from '@clerk/expo/hosted-auth';

const { startHostedAuth } = useHostedAuth();

await startHostedAuth();
await startHostedAuth({ mode: 'sign-up' });
```

Apps may pass `mode: 'sign-in' | 'sign-up'`, a custom non-HTTP `redirectUrl`, or `authSessionOptions`. By default, the hook uses the canonical callback registered for a native build and `AuthSession.makeRedirectUri()` in Expo Go.

## Flow

1. The hook creates a redirect URL, random `state`, and PKCE verifier/challenge.
2. It asks ClerkGo to create a hosted-auth Account Portal URL.
3. It opens that URL with `openAuthSessionAsync`.
4. On callback, it validates the redirect URL and `state`.
5. It redeems the returned `rotating_token_nonce` with the original PKCE verifier.
6. It updates the current Clerk client instance and calls `setActive` with the created native session.

## Security

- Uses a random `state` and verifies it on callback so the app only accepts the browser result for the auth attempt it initiated.
- Uses PKCE so ClerkGo stores the challenge at creation time and requires the matching verifier at redemption time.
- Sends the PKCE verifier in the FAPI request body, not in the URL.
- Validates callback protocol, host, and path against the initiated redirect URL before redeeming anything.
- Does not create or keep a web session in the SDK; the browser only drives Account Portal and the completed session is attached to the native client.
- Leaves production redirect-url allowlisting enforced by ClerkGo.

## Implementation Notes

Hosted auth is exposed from `@clerk/expo/hosted-auth` so apps that import the default `@clerk/expo` package do not need to resolve optional Expo auth/browser/crypto peer dependencies unless they opt into this flow.

The completion request is a physical `POST /v1/client` with `_method=GET` in the form body. That lets the backend route to the existing `/client` read/rotation path while keeping the PKCE verifier out of query strings and logs.

The verifier-bound completion stays scoped to `packages/expo/src/utils/hostedAuth.ts`. Generic `Client.reload()` remains the normal GET reload path, so ClerkJS core does not gain hosted-auth-specific branches.

## Screen Recording

https://github.com/user-attachments/assets/f1d40506-eef1-4bbf-81b9-139b0bf120fa

## Related PRs

- ClerkGo: https://github.com/clerk/clerk_go/pull/19852
- Accounts: https://github.com/clerk/accounts/pull/1429
- iOS: https://github.com/clerk/clerk-ios/pull/521
- Android: https://github.com/clerk/clerk-android/pull/803


packages/expo/src/hooks/useHostedAuth.ts
```diff
@@ -0,0 +1,251 @@
...
+    if (rotatingTokenNonce) {
+      updatedClient = await clerk.client?.reload({ rotatingTokenNonce, codeVerifier: pkce.codeVerifier });
+      if (updatedClient) {
+        getClientUpdater(clerk)?.(updatedClient);
+      }
+    }
+    const createdSessionId =
+      callbackParams.get('created_session_id') ??
+      updatedClient?.lastActiveSessionId ??
+      clerk.client?.lastActiveSessionId ??
```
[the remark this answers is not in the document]
mikepitre: Already addressed in the current branch. `useHostedAuth` now only derives `createdSessionId` from the callback or from a client reloaded with `rotating_token_nonce`; it no longer falls back to the pre-existing active client when callback params are missing. Covered by `does not fall back to an existing active session when callback session params are missing`.

packages/expo/src/hooks/useHostedAuth.ts
```diff
@@ -0,0 +1,270 @@
...
+  return {
+    codeVerifier,
+    codeChallenge,
+  };
+}
+
+function loadExpoCrypto(): typeof ExpoCrypto {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    return require('expo-crypto') as typeof ExpoCrypto;
```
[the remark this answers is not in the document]
mikepitre: Fixed in 63f5baa by moving hosted auth behind a dedicated @clerk/expo/hosted-auth subpath. The root hooks and types barrels no longer re-export useHostedAuth, so apps that import @clerk/expo without using hosted auth will not pull the optional Expo auth/crypto/web-browser modules into the default graph. I also added the directory-level package.json fallback to match the existing google/apple subpath pattern.

packages/expo/src/hooks/useHostedAuth.ts
```diff
@@ -0,0 +1,294 @@
...
+  };
+}
+
+function getDefaultRedirectUrl(AuthSessionModule: typeof AuthSession): string {
+  const appIdentifier = getExpoAppIdentifier();
+  if (appIdentifier && Platform.OS === 'ios') {
+    return `${appIdentifier}://callback`;
+  }
+  if (appIdentifier && Platform.OS === 'android') {
+    return `clerk://${appIdentifier}.callback`;
```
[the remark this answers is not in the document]
mikepitre: Fixed in [9e88ac2](https://github.com/clerk/javascript/commit/9e88ac2384e66b2c40c386995f8ebe413ce4e588). The Expo config plugin now adds an idempotent VIEW/BROWSABLE filter for `clerk://<android.package>.callback` to the main activity. I also added coverage for registration and duplicate prevention.

*Written by Codex*

packages/expo/src/hooks/useHostedAuth.ts
```diff
@@ -0,0 +1,294 @@
...
+    return Platform.OS === 'ios' ? expoConfig?.ios?.bundleIdentifier : expoConfig?.android?.package;
+  } catch {
+    return undefined;
+  }
+}
+
+function getHostedAuthClerk(clerk: ReturnType<typeof useClerk>): HostedAuthClerkInstance {
+  const hostedAuthClerk = clerk as ReturnType<typeof useClerk> & Partial<HostedAuthClerkInstance>;
+  if (typeof hostedAuthClerk.getFapiClient !== 'function') {
+    return errorThrower.throw('Hosted auth requires a Clerk instance that can make FAPI requests.');
```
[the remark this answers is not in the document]
mikepitre: Fixed in [9e88ac2](https://github.com/clerk/javascript/commit/9e88ac2384e66b2c40c386995f8ebe413ce4e588). `useHostedAuth()` now uses the existing native Clerk singleton for FAPI requests and keeps the React wrapper for client state and `setActive()`. The test mocks them as separate objects to match the real provider setup.

*Written by Codex*

packages/expo/src/hooks/useHostedAuth.ts
```diff
@@ -0,0 +1,301 @@
...
+  };
+}
+
+function getDefaultRedirectUrl(AuthSessionModule: typeof AuthSession): string {
+  const appIdentifier = getExpoAppIdentifier();
+  if (appIdentifier && Platform.OS === 'ios') {
+    return `${appIdentifier}://callback`;
+  }
+  if (appIdentifier && Platform.OS === 'android') {
+    return `clerk://${appIdentifier}.callback`;
```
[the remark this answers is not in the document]
mikepitre: Fixed in [5720b903](https://github.com/clerk/javascript/commit/5720b9034b3369e439cad8d47ef65e3ce3ae3a7d). Expo Go now returns early when `expo-constants` reports either `executionEnvironment === 'storeClient'` or the legacy `appOwnership === 'expo'`, so redirect generation falls through to `AuthSession.makeRedirectUri` even when the app config includes a bundle identifier or package name. I also added coverage for both metadata shapes.

*Written by Codex*

packages/expo/src/hooks/useHostedAuth.ts
```diff
@@ -0,0 +1,362 @@
...
+      authSessionResult,
+    };
+  }
+
+  return {
+    startHostedAuth,
+  };
+}
+
+function getDefaultRedirectUrl(AuthSessionModule: typeof AuthSession): string {
```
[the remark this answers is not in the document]
mikepitre: I checked the iOS side against the generated Expo config before changing it. Expo prebuild already registers `ios.bundleIdentifier` in `CFBundleURLTypes`; the SDK56 app's generated plist includes `com.clerk.exposimsdk56`, so `com.clerk.exposimsdk56://callback` is handled without Clerk adding the scheme again. I kept that behavior and added a comment next to the default callback explaining the dependency.

The Android part was valid. The plugin now only treats the callback filter as registered when it includes `VIEW`, `DEFAULT`, and `BROWSABLE`, along with the expected scheme and host. I added regression coverage for each missing requirement in [`1892027`](https://github.com/clerk/javascript/commit/189202795eb24bcc9979147cf121ac65844f9c0b).

*Written by Codex*

packages/expo/src/hooks/useHostedAuth.ts
```diff
@@ -0,0 +1,357 @@
...
+    if (!createdSessionId) {
+      return errorThrower.throw('Hosted auth completion did not include a created session id.');
+    }
+    if (!clientJSON.sessions.some(session => session.id === createdSessionId)) {
+      return errorThrower.throw('Hosted auth created session was not found on the redeemed client.');
+    }
+
+    await clerk.setActive({
+      session: createdSessionId,
+    });
```
mikepitre, line 197: @wobsoriano one thing I was unsure of on this one was if we wanted to just return the session and let the caller call setActive manually or do it for them like I've done here.

wobsoriano: Will look into the best DX for this

packages/expo/src/hooks/useHostedAuth.ts
```diff
@@ -0,0 +1,342 @@
...
+
+function getDefaultRedirectUrl(AuthSessionModule: typeof AuthSession): string {
+  const appIdentifier = getExpoAppIdentifier();
+  if (appIdentifier && Platform.OS === 'ios') {
+    // Expo prebuild registers ios.bundleIdentifier as an iOS URL scheme.
+    return `${appIdentifier}://callback`;
+  }
+  if (appIdentifier && Platform.OS === 'android') {
+    warnAndroidDefaultRedirectOnce();
+    return `clerk://${appIdentifier}.callback`;
```
wobsoriano: Why `clerk://<package>.callback here` instead of `<package>://callback`? Prebuild already registers the package name as a scheme (same as the iOS branch above), which would make this plugin-free like iOS, right?

packages/expo/src/hooks/useHostedAuth.ts
wobsoriano: I think auto-activating here is the right call and we should keep it.

Other hooks like `useSSO` returns `setActive` since SSO can end in intermediate states (transferable sign-in, missing requirements, MFA) where the caller has to inspect `signIn/signUp` before a session exists. I believe hosted auth has none of the those, AP resolves severything in the browser

wobsoriano (commented):
Left a q!

Everything looks good here, great job! Thanks for following existing conventions

packages/expo/src/hooks/useHostedAuth.ts
mikepitre: I think prebuild only does that on iOS actually. The android side of `withScheme` just registers whatever's in `scheme`, it never adds the package name, so `<package>://callback` wouldn't make it back to the app either. Since android needs an intent filter no matter what, figured we might as well use `clerk://<package>.callback` since that's what the server already registers for the android app (and what clerk-android defaults to), so nothing needs to be allowlisted.

wobsoriano (approved):

packages/expo/src/hooks/useHostedAuth.ts
wobsoriano: Confirmed the prebuild behavior locally, Android only registers what's in scheme, so `<package>://callback` is a dead end

I found a problem with `clerk://<package>.callback` in Expo during further testing though. clerk-android already registers `SSOReceiverActivity` for the same URI, so 2 activities in the app claim the callback and on a fresh install Android shows the "Open with" prompt at callback time:

Screenshot 2026-07-23 at 8 42 36 PM

Picking the clerk-android one silently loses the sign-in. The callback goes into native SSO code, `startHostedAuth()` resolves with `createdSessionId: null`, and the nonce is spent. Tapping "Always" pins the device permanently, which is why this is easy to miss in testing.

I tried manually patching locally, switching the callback to a distinct host (`clerk://<package>.hosted-callback`), and the app opens instantly with no prompt

wobsoriano (approved):
Re-tested and the collision issue is fixed 👍🏼

packages/expo/src/hooks/useHostedAuth.ts
wobsoriano: thanks for fixing!

PR #8961, opened 2026-06-23
fix(clerk-js): validate window navigation protocols and honor allowedRedirectProtocols

jacekradko:
`windowNavigate` in `@clerk/shared` assigned its argument straight to `window.location.href` on the assumption that callers had already validated it. This change routes every internal browser navigation in `clerk-js` and `@clerk/ui` through one chokepoint, `clerk.__internal_windowNavigate`, which checks the resolved URL against the configured `allowedRedirectProtocols` (the `http`/`https`/`wails`/`chrome-extension` defaults plus any customer additions) and rejects scheme-relative inputs like `//host` before parsing.

The part worth a close look is the version-skew path. `@clerk/ui` can run against an older `clerk-js` that does not expose `__internal_windowNavigate`, so `clerkWindowNavigate` falls back to the bare `@clerk/shared` helper and the static allowlist, still rejecting the bad cases. `@clerk/react`'s `IsomorphicClerk` mirrors that on its own forwarder: it delegates to the loaded `clerk-js` when the method is present, and otherwise falls back to the shared helper (honoring `allowedRedirectProtocols` from its options, browser-only), so the `__internal_windowNavigate` the React SDK exposes via `useClerk()` stays consistent instead of silently doing nothing on an older `clerk-js`. The bare `windowNavigate` export stays for that reason and is now `@deprecated` in favor of the wrapper.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed missing redirect URL protocol validation for browser navigations, including multi-session add-account, so redirects only use permitted protocols (“fail closed” for unsafe schemes).
  * Improved consistency of full-page navigation across sign-in/sign-up and external verification flows, including when mixed component versions are in use.
  * Updated UI flows to prefer the Clerk instance-aware navigation path when available.

* **New Features**
  * Introduced an internal navigation chokepoint with centralized protocol allowlist handling.

* **Tests**
  * Added/expanded coverage for allowed/blocked protocols, scheme-relative inputs, and fallback behavior when the internal navigation hook is unavailable.


packages/shared/src/internal/clerk-js/windowNavigate.ts
```diff
@@ -11,13 +11,48 @@ export const ALLOWED_PROTOCOLS = [
...
+ * `options.allowedProtocols` so legitimate custom protocols (Wails, Tauri, etc.) are honored.
+ *
+ * @deprecated Use `clerk.__internal_windowNavigate` instead. It honors the customer-supplied
+ * `allowedRedirectProtocols` option by default, so internal call sites can't accidentally
+ * bypass it by forgetting to pass `options.allowedProtocols`. The bare export will be removed
+ * in the next major version.
  */
-export function windowNavigate(to: URL | string): void {
+export function windowNavigate(to: URL | string, options?: WindowNavigateOptions): void {
+  if (typeof to === 'string' && SCHEME_RELATIVE_PREFIX.test(to.trim())) {
```
dominic-clerk: **Scheme-relative open-redirect bypass via control characters**

`trim()` only strips leading/trailing whitespace, but `new URL()` (next line) strips tab/LF/CR from anywhere and leading C0 controls — so an interior or leading control char defeats this regex while still resolving to a scheme-relative URL. It then inherits the base `https:` scheme (allowlisted), so navigation proceeds cross-origin.

Bypasses against base `https://victim-app.com/`:
```js
"/\t/evil.attacker.com"   // navigates cross-origin
"\x00//evil.attacker.com" // navigates cross-origin
"//evil.attacker.com"     // correctly blocked (only case covered)
```
The added tests only cover leading/trailing whitespace, which `trim()` handles, so this slips through.

Fix — strip parser-ignored chars before the test:
```js
if (typeof to === 'string' && SCHEME_RELATIVE_PREFIX.test(to.replace(/[\u0000-\u001F]/g, ''))) {
```

- Generated by Claude with @dominic-clerk's supervision

dominic-clerk (changes requested):

packages/shared/src/internal/clerk-js/windowNavigate.ts
jacekradko: Addressed in [6b919f9](https://github.com/clerk/javascript/pull/8961/commits/6b919f9e78ba35f96e9e487c0d09b472f65f29ef)

dominic-clerk (approved):

PR #8969, opened 2026-06-23
fix(clerk-js): prevent broadcast failure from evicting a cached session token

jacekradko:
`SessionTokenCache.setInternal` runs the cross-tab broadcast as the last step inside the token-resolution `.then`, and that whole chain ends in `.catch(() => deleteKey())`. So if `postMessage` (or `getRawString`) throws, the failure lands in that catch and evicts the token that was just cached and timer-scheduled, turning the next `getToken()` into an avoidable network fetch.

This isolates the broadcast side-effect in its own try/catch (logs a warning, keeps the cached token). The regression test reproduces the bug: with the guard removed, `size()` drops to 0 after a throwing `postMessage`; with the guard the token stays cached.

Fixes SDK-119. It is pre-existing in shipped clerk-js and a backport candidate, so it is scoped to just the broadcast guard, independent of the TokenCache decouple in #8860.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where cross-tab token broadcasting failures would unnecessarily evict cached session tokens, preventing unexpected follow-up network requests on token retrieval.

* **Tests**
  * Added test coverage for broadcast resilience to ensure cached tokens remain intact when cross-tab synchronization encounters message delivery failures.


wobsoriano (approved):
looks straightforward

PR #8970, opened 2026-06-23
fix(ui): re-seat nested SSO wizard to provider selection when connection is reset

iagodahlem:
## What

Resetting (deleting) a connection from the **first** per-provider configure step in `<ConfigureSSO />` left the user on a blank screen instead of returning to provider selection.

## Why

The wizard self-corrects to the furthest-reachable step when the active step's reachability predicate breaks, but that correction was gated to top-level (non-nested) wizards only. The provider-selection → per-provider-config flow runs as a nested wizard whose `configure-provider` step is reachable only while a connection exists. On reset the connection is deleted and that step becomes unreachable, but because the wizard is nested the correction never ran — it stayed on `configure-provider`, which renders nothing without a connection, producing the blank.

Resetting from a later step (e.g. Test) already worked, because that path re-seats the outer wizard and remounts the nested wizard fresh. Resetting from inside the nested wizard needed the nested wizard itself to re-seat in place.

## How

Remove the `!isNested` gate on the reachability clamp in `useWizardMachine.ts` so the clamp runs for nested wizards too. The nested wizard re-derives to its furthest reachable local step (`select-provider`) in place, via the same `initialState` derivation used on mount — a local state update, never bubbling to the parent. Scope is contained: the only nested step today with a reachability predicate is `configure-provider`; the per-provider inner wizards are guard-less, so the clamp is inert for them.

## Testing

- New integration test: resetting from the first configure step lands on `select-provider` (no blank), mirroring the existing reset-from-test-step case.
- Updated the nested-clamp unit test to assert the wizard re-seats in place **and** does not call the parent's navigation (proving it's local, not a bubble).
- Full `ConfigureSSO` + `OrganizationSecurityPage` suites pass (193 tests); existing flows confirmed unchanged (reset-from-test, resume-skips-selection, forward nav, change-provider).
- Manually verified in the sandbox: reset from the first configure step now returns to provider selection with a fresh card.

Resolves ORGS-1675

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue in the self-serve SSO configuration wizard where resetting from the initial setup step could show a blank page. The wizard now reliably returns to the provider selection step.
* **Tests**
  * Added an integration test covering the reset-from-first configure-step flow and verifying the correct step is displayed after reset.
* **Behavior Updates**
  * Updated nested wizard reachability handling so the wizard re-seats to the furthest locally reachable step instead of leaving it on an unreachable step.

### DIFF
--- a/packages/ui/src/components/ConfigureSSO/__tests__/ConfigureSSO.navigation.test.tsx
+++ b/packages/ui/src/components/ConfigureSSO/__tests__/ConfigureSSO.navigation.test.tsx
@@ -163,13 +163,6 @@ describe('ConfigureSSO wizard navigation (integration)', () => {
     expect(queryByText(/test your sso connection/i)).not.toBeInTheDocument();
   });
 
-  // ORGS-1675: resetting from the FIRST per-provider configure step (not the test
-  // step) must also land on select-provider. The configure sub-flow runs inside a
-  // NESTED wizard whose `configure-provider` step is gated on `hasConnection`;
-  // deleting the connection breaks that guard. Before the fix the nested clamp was
-  // skipped (gated on `!isNested`), so the wizard stranded on `configure-provider`
-  // — which renders nothing without a provider — leaving a blank pane. The clamp
-  // now runs for the nested wizard too and re-seats it in place to select-provider.
   it('reset → re-derive: deleting from the first configure step lands on select-provider (no blank)', async () => {
     const { wrapper, fixtures } = await createFixtures(withAdminOrgUser);
 
--- a/packages/ui/src/components/ConfigureSSO/__tests__/ConfigureSSO.navigation.test.tsx
+++ b/packages/ui/src/components/ConfigureSSO/__tests__/ConfigureSSO.navigation.test.tsx
@@ -163,6 +163,52 @@ describe('ConfigureSSO wizard navigation (integration)', () => {
     expect(queryByText(/test your sso connection/i)).not.toBeInTheDocument();
   });
 
+  // ORGS-1675: resetting from the FIRST per-provider configure step (not the test
+  // step) must also land on select-provider. The configure sub-flow runs inside a
+  // NESTED wizard whose `configure-provider` step is gated on `hasConnection`;
+  // deleting the connection breaks that guard. Before the fix the nested clamp was
+  // skipped (gated on `!isNested`), so the wizard stranded on `configure-provider`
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/ConfigureSSO/__tests__/ConfigureSSO.navigation.test.tsx
```diff
@@ -163,6 +163,52 @@ describe('ConfigureSSO wizard navigation (integration)', () => {
     expect(queryByText(/test your sso connection/i)).not.toBeInTheDocument();
   });
 
+  // ORGS-1675: resetting from the FIRST per-provider configure step (not the test
+  // step) must also land on select-provider. The configure sub-flow runs inside a
+  // NESTED wizard whose `configure-provider` step is gated on `hasConnection`;
+  // deleting the connection breaks that guard. Before the fix the nested clamp was
+  // skipped (gated on `!isNested`), so the wizard stranded on `configure-provider`
+  // — which renders nothing without a provider — leaving a blank pane. The clamp
+  // now runs for the nested wizard too and re-seats it in place to select-provider.
```
LauraBeatris:
```suggestion
```

packages/ui/src/components/ConfigureSSO/elements/Wizard/__tests__/useWizardMachine.test.tsx
```diff
@@ -550,26 +550,35 @@ describe('useWizardMachine — reachability clamp (self-correct on a broken isRe
...
-  it('does not clamp a nested wizard (isNested) even when its active isReachable breaks', () => {
-    // A nested machine has no isReachable in practice, but the clamp is gated on
-    // !isNested regardless: a nested wizard never re-seats, it bubbles instead.
+  it('re-seats a nested wizard in place when its active isReachable breaks (does not bubble)', () => {
+    // ConfigureSSO's middle wizard is nested AND has a guarded step
+    // (`configure-provider`, reachable only while a connection exists). When that
+    // connection is deleted from inside the flow, the guard breaks; the clamp must
+    // re-seat to the furthest LOCAL reachable step (n0 here, select-provider in the
+    // app) IN PLACE rather than bubbling to the parent — otherwise the nested
+    // wizard strands on an impossible step (the ORGS-1675 blank pane).
```
LauraBeatris:
```suggestion
```

LauraBeatris (approved):

PR #8971, opened 2026-06-23
fix(ui): hide OrganizationProfile Security tab when lacking manage enterprise connections permission

iagodahlem:
## Summary

`ConfigureSSO` is embedded in `<OrganizationProfile />` → Security tab, which shows when self-serve SSO is enabled at the org level. A member without the `org:sys_entconns:manage` permission previously landed on the Security tab and hit a missing-permission state that was built as a wizard step, so it rendered wizard chrome (header/footer) that doesn't fit someone who never entered a flow.

Since SSO is currently the only section in Security, this hides the Security tab entirely when the current member lacks `org:sys_entconns:manage` — matching how the Members, Billing, and API keys tabs are permission-gated.

## Changes

- `OrganizationProfileNavbar.tsx` — filter out the Security nav item via `useProtect(has => has({ permission: 'org:sys_entconns:manage' }))`, mirroring the existing Members/Billing filters.
- `OrganizationProfileRoutes.tsx` — wrap the Security route in `<Protect>` so a direct navigation is gated too (matches the Billing/API keys routes).
- `OrganizationSecurityPage.tsx` — drop the now-redundant page-level `ConfigureSSOProtect` wrapper; the route guard covers it. `ConfigureSSOProtect` stays in place for the standalone `<ConfigureSSO />` component.
- Tests updated to assert the Security tab is hidden when the permission is missing. The "hide when self-serve SSO is disabled at the org/instance level" behavior is unchanged and still covered by tests.

## Follow-up

When the Security page gains non-SSO settings, the permission-error state should be redesigned to scope to the SSO section only (so the tab can stay visible for other settings). Tracked separately.

Resolves ORGS-1676.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened access control for the Organization Security tab so it stays hidden unless `org:sys_entconns:manage` is granted.
  * Protected the self-serve SSO “organization-security” route with the same permission gate.
  * Updated the Security page to show the correct content (or centered loading spinner) only when access is allowed, instead of a permission-denied message.
  * Refined navbar/route filtering to include Security only when permitted (and keep the existing API keys visibility rule consistent).
* **Tests**
  * Updated and refactored route/SSO permission assertions to verify “hidden vs. warning” behavior and guarded routing.

### DIFF
--- a/.changeset/olive-pans-sin.md
+++ b/.changeset/olive-pans-sin.md
@@ -1,5 +1,4 @@
 ---
-'@clerk/clerk-js': patch
 '@clerk/ui': patch
 ---
 
--- a/.changeset/olive-pans-sin.md
+++ b/.changeset/olive-pans-sin.md
@@ -1,0 +1,6 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/ui': patch
+---
+
+The Security tab in `<OrganizationProfile />` is now hidden for members who lack the manage enterprise connections permission (`org:sys_entconns:manage`), instead of rendering a permission-denied state. This matches how the Members, Billing, and API keys tabs are gated.
--- a/packages/ui/src/components/OrganizationProfile/OrganizationProfileNavbar.tsx
+++ b/packages/ui/src/components/OrganizationProfile/OrganizationProfileNavbar.tsx
@@ -29,6 +29,8 @@ export const OrganizationProfileNavbar = (
       }) || has({ permission: 'org:sys_billing:manage' }),
   );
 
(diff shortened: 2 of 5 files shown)


LauraBeatris (approved):

.changeset/olive-pans-sin.md
```diff
@@ -0,0 +1,6 @@
+---
+'@clerk/clerk-js': patch
```
LauraBeatris:
```suggestion
```

PR #8977, opened 2026-06-24
feat(clerk-js, localizations, ui): Add credits information in the user/org profile

l-armstrong:
Display the C2's remaining account credit and their account credit history in the user/org profile. 
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Added an Account Credits section to user and organization billing/subscriptions views, showing the current credit balance.
  * Added a Credit History page with navigation, loading state, and a localized table of credit ledger entries.
* **Bug Fixes**
  * Billing subscription data now includes the payer reference needed to load credit views correctly.
* **Documentation**
  * Added new localization strings for the Account Credits section and Credit History page (titles and table headers).


packages/clerk-js/src/core/resources/BillingSubscription.ts
```diff
@@ -36,6 +36,7 @@ export class BillingSubscription extends BaseResource implements BillingSubscrip
   nextPayment?: BillingSubscriptionNextPayment | null;
   subscriptionItems!: BillingSubscriptionItemResource[];
   eligibleForFreeTrial!: boolean;
+  payerId!: string;
```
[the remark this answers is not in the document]
l-armstrong: the payer_id field is guaranteed by the API contract and this pattern is consistent with the existing patterns.

packages/ui/src/components/AccountCredits/CreditHistoryPage.tsx
```diff
@@ -0,0 +1,149 @@
...
+}
+
+function formatCreditAmount(amount: number, currencySymbol: string): string {
+  const absAmount = Math.abs(amount);
+  const dollars = (absAmount / 100).toFixed(2);
+  const prefix = amount >= 0 ? '+' : '-';
+  return `${prefix}${currencySymbol}${dollars}`;
+}
+
+export const CreditHistoryPage = () => {
```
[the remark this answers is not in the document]
l-armstrong: I added the explicit return.

.changeset/fancy-rats-stick.md
```diff
@@ -0,0 +1,8 @@
+---
+'@clerk/localizations': minor
+'@clerk/clerk-js': minor
+'@clerk/shared': minor
+'@clerk/ui': minor
+---
+
+Display the C2s Account Credits in the user/org profile.
```
mauricioabreu: 💭  Maybe we should mention Billing somewhere here

dstaley: also keep in mind that "C2" is an internal term, and likely not recognizable to our end users

packages/clerk-js/src/core/modules/billing/namespace.ts
```diff
@@ -1,5 +1,9 @@
 import type {
   BillingCheckoutJSON,
+  BillingCreditBalanceJSON,
```
mauricioabreu, line 3: ❔  Should it be `BillingAccountCreditBalance`?
cc @dstaley

packages/shared/src/react/stable-keys.ts
```diff
@@ -60,6 +63,9 @@ export const STABLE_KEYS = {
 
   // Keys for `useOAuthConsent`
   OAUTH_CONSENT_INFO_KEY,
+
+  // Keys for `useCreditBalance`
```
mauricioabreu, line 67: 🔧  The comment does not add much to the context

packages/shared/src/types/billing.ts
```diff
@@ -926,6 +940,59 @@ export interface BillingSubscriptionResource extends ClerkResource {
    * Whether the payer is eligible for a free trial.
    */
   eligibleForFreeTrial: boolean;
+
+  /**
+   * The ID of the payer for this subscription.
```
mauricioabreu: _subscription item_?

packages/clerk-js/src/core/modules/billing/namespace.ts
dstaley: No, the internal type is `CommerceCreditBalanceResponse`, so `BillingCreditBalanceJSON` would be the correct version of the JavaScript type.

packages/shared/src/types/billing.ts
dstaley: this is on the subscription, not the subscription item

packages/ui/src/components/AccountCredits/AccountCredits.tsx
```diff
@@ -0,0 +1,53 @@
...
+      })}
+    >
+      <ProfileSection.ItemList
+        id='accountCredits'
+        disableAnimation
+      >
+        <ProfileSection.Item id='accountCredits'>
+          <Text variant='subtitle'>
+            {creditBalance.balance.currencySymbol}
+            {creditBalance.balance.amountFormatted}
```
dstaley: should use the new `$` helper from `useLocalizations` for multi-currency handling

packages/ui/src/components/AccountCredits/CreditHistoryPage.tsx
```diff
@@ -0,0 +1,149 @@
+import { Header } from '@/ui/elements/Header';
+import { ProfileCard } from '@/ui/elements/ProfileCard';
+
+import { useCreditBalance, useCreditHistory, useSubscriberTypeLocalizationRoot } from '../../contexts';
+import { Box, descriptors, localizationKeys, Spinner, Text } from '../../customizables';
+import { useRouter } from '../../router';
+
+function formatCreditDate(date: Date): string {
+  return new Intl.DateTimeFormat('en-US', {
```
dstaley: can you check if we're doing locale-aware date formatting anywhere? might be a gap we need to address in a future PR

packages/ui/src/components/AccountCredits/CreditHistoryPage.tsx
```diff
@@ -0,0 +1,149 @@
...
+    year: 'numeric',
+  }).format(date);
+}
+
+function formatCreditAmount(amount: number, currencySymbol: string): string {
+  const absAmount = Math.abs(amount);
+  const dollars = (absAmount / 100).toFixed(2);
+  const prefix = amount >= 0 ? '+' : '-';
+  return `${prefix}${currencySymbol}${dollars}`;
+}
```
dstaley: let's use our `$` formatter from `useLocalizations`

packages/ui/src/components/AccountCredits/CreditHistoryPage.tsx
```diff
@@ -0,0 +1,149 @@
...
+                  variant='caption'
+                  colorScheme='secondary'
+                  localizationKey={localizationKeys(
+                    `${localizationRoot}.billingPage.creditHistoryPage.tableHeader__date`,
+                  )}
+                />
+              </Box>
+            </Box>
+          </Box>
+          <Box as='tbody'>
```
dstaley: `@/ui/customizables` has actual table elements like `Tbody` and `Td`, so no need to use `Box` with `as`.

packages/shared/src/react/hooks/useCreditBalance.tsx
```diff
@@ -0,0 +1,103 @@
...
+        userId: user?.id,
+        orgId: safeOrgId,
+      },
+      untracked: {
+        args: { orgId: safeOrgId },
+      },
+    });
+  }, [user?.id, organization?.id, params?.for]);
+
+  const queriesEnabled = Boolean(user?.id && billingEnabled);
```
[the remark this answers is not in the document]
l-armstrong: Good catch, I applied the changes.

packages/clerk-js/src/core/resources/BillingSubscription.ts
dstaley: do we still need this now that we're not passing it to the endpoints?

packages/shared/src/react/hooks/useCreditBalance.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { createCacheKeys } from './createCacheKeys';
+import { useBillingIsEnabled } from './useBillingIsEnabled';
+import { useClearQueriesOnSignOut } from './useClearQueriesOnSignOut';
+
+const HOOK_NAME = 'useCreditBalance';
+
+export type UseCreditBalanceParams = {
+  for?: ForPayerType;
+  keepPreviousData?: boolean;
+  enabled?: boolean;
```
dstaley, line 21: what is this parameter used for? why would a caller pass it as false?

packages/shared/src/react/hooks/useCreditBalance.tsx
```diff
@@ -0,0 +1,103 @@
...
+  error: Error | undefined;
+  isLoading: boolean;
+  isFetching: boolean;
+  revalidate: () => Promise<void> | void;
+};
+
+/**
+ * @internal
+ */
+export function useCreditBalance(params?: UseCreditBalanceParams): CreditBalanceResult {
```
dstaley: does this need the `__internal_` prefix the other hook is using? I don't understand why the balance hook is public, but the ledger history hook is internal.

packages/shared/src/types/billing.ts
```diff
@@ -926,6 +940,51 @@ export interface BillingSubscriptionResource extends ClerkResource {
...
+   */
+  payerId: string;
```
dstaley: I don't think we need this anymore?

packages/shared/src/types/billing.ts
```diff
@@ -926,6 +940,51 @@ export interface BillingSubscriptionResource extends ClerkResource {
...
+};
+
+/**
+ * @experimental This is an experimental API for the Billing feature that is available under a public beta, and the API is subject to change. It is advised to [pin](https://clerk.com/docs/pinning) the SDK version and the clerk-js version to avoid breaking changes.
+ */
+export interface BillingCreditLedgerResource {
+  id: string;
+  payerId: string;
+  amount: number;
+  currency: string;
```
dstaley: This isn't our standard BillingMoneyAmount way of representing currency denominated values in our APIs. This is either something that's wrong on the type level or wrong in FAPI. It should be consistent with other endpoints so we can use shared helpers like our currency formatting helper, which expects the standard BillingMoneyAmount type.

packages/shared/src/types/billing.ts
```diff
@@ -926,6 +940,51 @@ export interface BillingSubscriptionResource extends ClerkResource {
...
+   */
+  orgId?: string;
+};
+
+/**
+ * @experimental This is an experimental API for the Billing feature that is available under a public beta, and the API is subject to change. It is advised to [pin](https://clerk.com/docs/pinning) the SDK version and the clerk-js version to avoid breaking changes.
+ */
+export interface BillingCreditLedgerResource {
+  id: string;
+  payerId: string;
```
dstaley: let's remove this from the TypeScript types (it's fine that it's in the FAPI response for now)

packages/ui/src/components/AccountCredits/CreditHistoryPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+                <Td sx={t => ({ padding: t.space.$3 })}>
+                  <Text
+                    variant='subtitle'
+                    sx={{ color: entry.amount >= 0 ? undefined : 'inherit' }}
+                  >
+                    {`${entry.amount >= 0 ? '+' : '-'}${$({
+                      amount: Math.abs(entry.amount),
+                      amountFormatted: '',
+                      currency: entry.currency,
+                      currencySymbol: '',
```
dstaley: once we update the API to return our standard BillingMoneyAmount type we can simplify this

dstaley (changes requested):

packages/shared/src/react/hooks/useCreditBalance.tsx
l-armstrong: yes, you're right. I've updated it the have the `__internal_ prefix ` to be consistent.

packages/shared/src/types/billing.ts
l-armstrong: Yes, this was previously wrong in FAPI. It's been corrected to return the standard BillingMoneyAmount.

packages/shared/src/react/hooks/useCreditBalance.tsx
l-armstrong: `enabled` is the pattern we are using for conditionally disabling fetches. All of the billing hooks pass through `useBillingHookParams()` which sets `enabled: false` for org members lacking `org:sys_billing:read/manage permissions`. Just following the same patterns as the other hooks.

packages/ui/src/components/AccountCredits/CreditHistoryPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+                })}
+              >
+                <Td sx={t => ({ padding: t.space.$3 })}>
+                  <Text
+                    variant='subtitle'
+                    sx={{ color: entry.amount.amount >= 0 ? undefined : 'inherit' }}
+                  >
+                    {`${entry.amount.amount >= 0 ? '+' : '-'}${$({
+                      ...entry.amount,
+                      amount: Math.abs(entry.amount.amount),
```
dstaley: Our `$` helper will automatically use a negative symbol in the correct location, so we don't need to manually add one (plus, as written, this will render two negative symbols). I think it's fine to lose the + symbol as well. We also don't need to take the absolute value of the amount field. So, taken together, the following should be okay:

```
{$(entry.amount)}
```

packages/ui/src/components/AccountCredits/CreditHistoryPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+              <Th
+                sx={t => ({ padding: t.space.$3, textAlign: 'left' })}
+                localizationKey={localizationKeys(
+                  `${localizationRoot}.billingPage.creditHistoryPage.tableHeader__date`,
+                )}
+              />
+            </Tr>
+          </Thead>
+          <Tbody>
+            {creditHistory?.data?.map(entry => (
```
dstaley: We can fix this in a follow-up PR, but I believe we'll want to support paginating this list. Not sure if the endpoint supports that or not yet, so we can tackle this later.

dstaley (changes requested):

packages/ui/src/components/AccountCredits/CreditHistoryPage.tsx
l-armstrong: ah okay. makes sense. fixed.

packages/ui/src/components/AccountCredits/CreditHistoryPage.tsx
l-armstrong: I can add support for pagination in a separate PR.    
https://linear.app/clerk/issue/BILL-1844/add-pagination-to-listing-a-payers-credit-history

packages/ui/src/components/AccountCredits/AccountCredits.tsx
```diff
@@ -0,0 +1,44 @@
...
+        disableAnimation
+      >
+        <ProfileSection.Item id='accountCredits'>
+          <Text variant='subtitle'>{$(creditBalance.balance)}</Text>
+        </ProfileSection.Item>
+        <ProfileSection.Button
+          id='accountCredits'
+          localizationKey={localizationKeys(`${localizationRoot}.billingPage.accountCreditsSection.viewHistory`)}
+          sx={[t => ({ justifyContent: 'start', height: t.sizes.$8 })]}
+          onClick={() => void navigate('credit-history')}
```
dstaley, line 39: I think this is fine for now, but we should iterate on it because we now have three different types of buttons on the same page:
1. Icon, label, no arrow
2. Icon, label, arrow on hover
3. No icon, label, no arrow on hover

Would be nice to have some consistent logic being used here for when a button needs an icon, and when it should show an arrow.

dstaley (approved):

dstaley (approved):

packages/ui/src/components/AccountCredits/AccountCredits.tsx
l-armstrong: Makes sense. I made a ticket for it to track this https://linear.app/clerk/issue/BILL-1845/improve-ui-consistency-for-account-credits-component

PR #9014, opened 2026-06-26
test(electron): Initial Electron SDK integration test

wobsoriano:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Added an Electron + Vite preset and a new Electron-Vite starter template with authentication UI, secure preload bridge, and baseline styling.
  * Wired the Electron preset into the integration preset catalog for quicker scaffolding.
* **Bug Fixes**
  * Improved CI reliability for Electron integration tests by running them under a virtual display when needed.
  * Hardened Electron renderer asset loading to safely serve bundled routes.
* **Tests**
  * Added Playwright Electron E2E tests for sign-in/sign-out and session persistence, plus fixtures and `test:integration:electron`/Turbo task support.


.github/workflows/ci.yml
```diff
@@ -486,10 +487,24 @@ jobs:
...
         run: ls -la && pwd
 
+      - name: Ensure Xvfb is installed
+        if: ${{ matrix.test-name == 'electron' }}
+        run: |
+          if ! command -v xvfb-run &> /dev/null; then
+            sudo apt-get update
+            sudo apt-get install -y xvfb
+          fi
+          xvfb-run --help > /dev/null
```
wobsoriano: We need `xvfb` because the Electron E2E launches a real BrowserWindow in Linux CI, where there's no display server by default. `xvfb-run` provides a virtual display, and it's only installed/used for the Electron matrix job

integration/tests/electron/fixtures.ts
```diff
@@ -0,0 +1,87 @@
...
+      await app.withEnv(env);
+      await app.setup();
+      await run('node node_modules/electron/install.js', { cwd: app.appDir });
+      await app.build();
+      await setupClerkTestingEnv(env);
+
+      await provide(app);
+      await app.teardown();
+    },
+    { scope: 'worker', timeout: 120_000 },
```
jeremy-clerk: This reuses the chrome-extension worker timeout (120s), but electron does strictly more work in that budget: pnpm install + the ~100MB electron binary download (`node install.js`) + vite build. The margin is tighter here, so watch for timeout flakes on cold CI runners and consider bumping it.

.github/workflows/ci.yml
jeremy-clerk: Some `xvfb-run` versions do not recognize `--help` and exit non-zero, which would fail this step even when xvfb is functional. `command -v xvfb-run` (already used just above) is a safer functional probe.

integration/templates/electron-vite/main.mjs
```diff
@@ -0,0 +1,71 @@
...
+
+function registerClerkAppProtocol() {
+  protocol.handle(RENDERER_SCHEME, async request => {
+    const url = new URL(request.url);
+
+    if (url.host !== RENDERER_HOST) {
+      return new Response('Not found', { status: 404 });
+    }
+
+    const requestedPath = decodeURIComponent(url.pathname);
```
jeremy-clerk: `decodeURIComponent` throws on malformed percent-encoding (e.g. a stray `%` in the path), rejecting the handler with no 4xx fallback. Low risk since the renderer only requests its own bundled assets, but wrapping this in try/catch and returning a 400 would make the handler defensive.

integration/presets/electron.ts
```diff
@@ -0,0 +1,17 @@
+import { applicationConfig } from '../models/applicationConfig';
+import { templates } from '../templates';
+import { PKGLAB } from './utils';
+
+const vite = applicationConfig()
+  .setName('electron-vite')
+  .useTemplate(templates['electron-vite'])
+  .setEnvFormatter('public', key => `VITE_${key}`)
+  .addScript('setup', 'pnpm install')
+  .addScript('dev', 'pnpm build')
```
jeremy-clerk: `dev` (and `serve` below) are dead config here, since the fixtures launch Electron directly and never call `app.dev()`/`serve`. Harmless, but aliasing `dev` to a one-shot build can imply a watch server exists; a clearer noop or a short comment would avoid the confusion.

jeremy-clerk (commented):
A few low-severity reliability/cleanup notes from a review pass. No correctness bugs; the Electron-specific details (privileged scheme timing, bundled clerk-js vs. served UI, pnpm-blocked binary install) all look correct.

jeremy-clerk (approved):
minor nits but looks good 👍 let me know what we need to do for passkey test support

.github/workflows/ci.yml
wobsoriano: addressed

integration/templates/electron-vite/main.mjs
wobsoriano: addressed

integration/presets/electron.ts
wobsoriano: addressed

integration/tests/electron/fixtures.ts
wobsoriano: addressed

PR #9015, opened 2026-06-26
feat(expo): Move Google Sign to a separate package

wobsoriano:
## Description

Moves the native Google Sign-In implementation out of `@clerk/expo` and into a new optional package, `@clerk/expo-google-signin`.

The Expo SDK currently includes the native Google Sign-In module even for apps that never call `useSignInWithGoogle()`. That means every Expo app using `@clerk/expo` can pick up the extra native iOS/Android Google Sign-In dependencies during prebuild/autolinking. This split keeps the native Google module opt-in while preserving the existing JS-facing API through a re-export.

This is intended for the next major version because apps using native Google Sign-In will need to install the new package and add its Expo config plugin.

### Migration

Apps that use native Google Sign-In now need to install the new package:

```sh
npx expo install @clerk/expo-google-signin
```

and add its config plugin:

```json
  {
    "expo": {
      "plugins": [
        "@clerk/expo",
        "@clerk/expo-google-signin"
      ]
    }
  }
```

The existing import path can continue to work:

```ts
import { useSignInWithGoogle } from '@clerk/expo/google';
```

but the native module is no longer included unless `@clerk/expo-google-signin` is installed and prebuilt into the app.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [x] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated `@clerk/expo-google-signin` package with a matching Expo config plugin.
  * Exposed `useSignInWithGoogle` and related flow types via the new package.
* **Breaking Changes**
  * Native Google Sign-In support is no longer bundled in `@clerk/expo`; install `@clerk/expo-google-signin`, add its config plugin, and rebuild.
  * The combined Expo config plugin no longer configures the iOS Google Sign-In URL scheme.
* **Bug Fixes**
  * Improved missing-native-module guidance and removed the prior migration warning behavior.
* **Documentation / Tests**
  * Updated import examples and refreshed Google sign-in hook tests.


wobsoriano:
!snapshot

packages/expo-google-signin/package.json
```diff
@@ -0,0 +1,73 @@
+{
+  "name": "@clerk/expo-google-signin",
```
mikepitre, line 2: Should we do anything for the first-publish path here, or is the plan to force this check once and publish immediately after merge?

`pkg-pr-new --compact --pnpm ./packages/*` is failing because `@clerk/expo-google-signin` does not exist on npm yet. If this might sit on `main` before the first publish, should we exclude it from compact previews or make that workflow tolerate brand-new packages?

.changeset/expo-google-signin-package.md
```diff
@@ -0,0 +1,8 @@
+---
+'@clerk/expo': major
+'@clerk/expo-google-signin': minor
```
mikepitre: Should we add `@clerk/expo-google-signin` to the API Changes/break-check coverage too? `@clerk/expo` and `@clerk/expo-passkeys` are already tracked there, so it seems like this new public package should get the same declaration snapshot coverage.

packages/expo-google-signin/expo-module.config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "platforms": ["apple", "android"],
```
mikepitre, line 2: Should the Expo native matrix exercise this package as part of the split? The current workflow still only triggers on `packages/expo/**` and only packs/installs `@clerk/expo`, so it proves the main package still prebuilds but not that the new native Google Sign-In package works in a clean app.

packages/expo-google-signin/package.json
wobsoriano: Good question! We need a manual publish for a pre-release version so that pkg-pr-new can pick it up (it wants a package available on NPM already).

@dominic-clerk similar to electron and electron-passkeys, can I request for `@clerk/expo-google-signin` to be manually published? 🙏🏼

.changeset/expo-google-signin-package.md
wobsoriano: Good catch, I will add it

packages/expo-google-signin/expo-module.config.json
wobsoriano: Oh yeah I think that's one way of testing it! will add

packages/expo-google-signin/package.json
dominic-clerk: @wobsoriano Done! ✅

wobsoriano: Thanks Dominic!

packages/expo-google-signin/expo-module.config.json
wobsoriano: added!

.changeset/expo-google-signin-package.md
wobsoriano: added!

wobsoriano:
!allow-major

mikepitre (approved):
LGTM!

wobsoriano:
!snapshot

wobsoriano:
!snapshot

PR #9029, opened 2026-06-28
fix(ui): keep truncateWithEndVisible code-point-safe in short-width fallback

mahirhir:
`truncateWithEndVisible` keeps the start and end of a string and puts an ellipsis in the middle. The main path builds the result from `Array.from(str)`, so it slices on code points and never breaks a character. The short-width fallback (taken when `maxLength <= endChars + 3`) instead returns `ELLIPSIS + str.slice(-endChars)`, which slices on UTF-16 code units.

For characters outside the BMP (CJK Extension B kanji, emoji) each character is a surrogate pair of two code units, so `str.slice(-endChars)` can cut one in half and leave a lone surrogate:

```ts
truncateWithEndVisible('𠮷𠮷𠮷𠮷𠮷', 8, 5)
// before: '...\uDFB7𠮷𠮷'  (broken leading character)
// after:  '...𠮷𠮷𠮷𠮷𠮷'
```

The fallback now slices by code point, the same way the main path already does. ASCII inputs are unaffected and the existing tests still pass. Added a test covering the fallback with astral characters.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text truncation so very short widths no longer split multi-code-point characters.
  * Fixed fallback truncation to preserve emoji and other non-BMP characters, avoiding broken replacement symbols.
  * Added test coverage to verify truncation remains correct for Unicode characters in edge cases.

### DIFF
--- a/.changeset/truncate-surrogate-safe.md
+++ b/.changeset/truncate-surrogate-safe.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Stop `truncateWithEndVisible` from splitting characters outside the BMP (such as CJK Extension B kanji and emoji) into a broken replacement character when truncating to a very small width. The short-width fallback now slices by code point, matching the main truncation path.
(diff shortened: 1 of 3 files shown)


alexcarpenter:
closing in favor of https://github.com/clerk/javascript/pull/9047

thanks for the PR @greymoth-jp

PR #9030, opened 2026-06-29
fix(clerk-js): apply session tokens monotonically on a single tab

nikosdouvlis:
A stale edge-minted session token (older `oiat` header) could overwrite a fresher one on the same tab. The freshness guard only ran in the cross-tab broadcast receiver, so the `__session` cookie, the session's `lastActiveToken`, and the token cache each took whatever token landed last. A slow or out-of-order edge read could revert a newer token, and a stale `last_active_token` on a client response could clobber a fresher cached token during hydration.

Freshness is ranked by the JWT `oiat` header (origin-issued-at), then `iat`, via `pickFreshestJwt`, and each store now enforces it where it is written:

- The cookie service drops a token only when it is strictly staler than the current `__session` cookie for the same session and organization. Missing `oiat`, decode failures, and cross-context tokens always write through, so sign-ins and session or org switches are unaffected and the guard fails open.
- `Session` only advances `lastActiveToken` when the dispatched token is at least as fresh as the current one.
- The token cache keeps an internal per-key baseline: the freshest claims-valid token it has seen, chained across `set()` calls and folded in whenever a resolver settles. A staler resolve can no longer overwrite a fresher token in the slot, and a fresher token from a resolver that was replaced mid-flight still lands once the slot resolves.

`getToken` semantics are otherwise untouched: a call arriving while a fetch is in flight still awaits that fetch (a pending cache entry never serves a token synchronously), `skipCache` callers still resolve to their own freshly minted token, and an empty token response still throws so the retry logic fires. An own fetch that fails or resolves unusable still drops the cache slot. Two tokens without `oiat` keep the existing newest-wins behavior.


packages/clerk-js/src/core/tokenFreshness.ts
```diff
@@ -50,3 +50,24 @@ export function pickFreshestJwt<T extends TokenResource | JWT>(existing: T, inco
...
+export function tokenOrgId(input: TokenResource | JWT): string {
+  const claims = asJwt(input)?.claims;
+  return (claims?.org_id as string) || ((claims?.o as { id?: string } | undefined)?.id ?? '');
+}
+
+export function normalizeOrgId(orgId?: string | null): string {
+  return orgId || '';
+}
+
+export function pickFreshestOrIncoming<T extends TokenResource | JWT>(existing: T | null | undefined, incoming: T): T {
```
nikosdouvlis: we dont need this - pickFreshestJwt can return non null value

bratsos (approved):

octoper (approved):

packages/clerk-js/src/core/auth/AuthCookieService.ts
```diff
@@ -214,6 +220,43 @@ export class AuthCookieService {
...
+
+    // Only a same session+org cookie is a comparable freshness baseline; write through otherwise.
+    if (
+      tokenSid(current) !== tokenSid(incoming) ||
+      normalizeOrgId(tokenOrgId(current)) !== normalizeOrgId(tokenOrgId(incoming))
+    ) {
+      return false;
+    }
+
+    return pickFreshestJwt(current, incoming) === current;
```
bratsos, line 253: ❓ Shouldn't we also consider `exp` for this comparison? What happens if we get a higher `oiat` but expired `exp` claim?

packages/clerk-js/src/core/resources/Session.ts
```diff
@@ -520,7 +521,7 @@ export class Session extends BaseResource implements SessionResource {
 
     eventBus.emit(events.TokenUpdate, { token });
 
-    if (token.jwt) {
+    if (token.jwt && pickFreshestJwt(this.lastActiveToken, token) === token) {
```
bratsos: ❓ What happens during an org-switch here? Could a new org token with a lower `oiat` lose to the previous org's token? Sounds edge-casey but you never know.

packages/clerk-js/src/core/tokenCache.ts
```diff
@@ -321,48 +330,84 @@ const MemoryTokenCache = (prefix?: string): TokenCache => {
...
+        }
+
+        clearTimeout(live.timeoutId);
+        clearTimeout(live.refreshTimeoutId);
 
-        const expiresAt = claims.exp;
-        const issuedAt = claims.iat;
+        const expiresAt = winnerClaims.exp;
+        const issuedAt = winnerClaims.iat;
         const expiresIn: Seconds = expiresAt - issuedAt;
```
bratsos, line 404: ❓ Should timers here be scheduled from the winner’s remaining TTL rather than its full lifetime?

Example:
at `t=0` we cache token A with `iat=0, exp=60, oiat=300`.
at `t=45`, refresh returns token B with `iat=45, exp=105, oiat=250`, so A wins.
This computes `expiresIn = A.exp - A.iat = 60` and schedules deletion/refresh 60s from now, at `t=105`, even though A actually expires at `t=60`.

nikosdouvlis:
!snapshot

PR #8853, opened 2026-06-13
docs(backend,shared): Generate typedoc output for backend docs

alexisintech:
## Summary

- Adds JSDoc comments to resources in `packages/backend/src/api/endpoints/**/*`.
- Refines JSDoc on the existing `packages/backend/src/api/resources/**/*` model classes (e.g. `User`, `Session`, `Organization`, `OrganizationInvitation`, `Verification`, `AllowlistIdentifier`, `Role`, `Permission`, `BillingPlan`, `BillingSubscription{,Item}`, `EnterpriseAccount`, `ExternalAccount`, `OAuthApplication`, `WaitlistEntry`, …). Recurring patterns:
  - `The date when …` → `The Unix timestamp when …` on every `createdAt` / `updatedAt` / `expiresAt` / `lastActiveAt` / etc.
  - `A boolean indicating whether …` → `Whether …`.
  - Adds `{{ target: '_blank' }}` to Backend API links and converts bare mentions of Roles / Permissions / Organizations / Sessions into proper doc links.
  - Fills in property-table tombstones — `M2MToken`, `Machine`, `MachineScope` go from a one-line description stub to a full properties table.
- Adds a custom `@expandProperties` modifier tag (handled by `.typedoc/custom-plugin.mjs`, registered in `typedoc.config.mjs`) so generic-instantiation aliases like `UserListParams = ClerkPaginationRequest<{...}>` render as a single merged properties table instead of an empty page. typedoc-plugin-markdown otherwise has no rendering branch for top-level `ReferenceType`s. Tag applied to `ClerkPaginationRequest` and `ClerkPaginationParams`.
- Wires backend API endpoint classes into the existing `extract-methods.mjs` pipeline via a new `BACKEND_API_CONFIG` in `.typedoc/reference-objects.mjs`. Each endpoint class registered here gets a `backend/<kebab-name>/` folder containing `<name>.mdx` (main page) and `methods/<method>.mdx` (one file per public method, with description + TypeScript signature + parameters).
- `extract-methods.mjs` now supports two output styles via a `methodFormat` option plumbed from the config:
  - **`reference`** (existing reference-object pages — Clerk, UserResource, etc.): `### foo()` title + `#### Parameters` heading. Designed for pages aggregating many methods.
  - **`page`** (new, used by `BACKEND_API_CONFIG`): no method-name title, `## Parameters` heading. Designed for one-method-per-docs-page consumption.
[...]

PR #9036, opened 2026-06-29
fix(backend): guard deleteOrganization against empty organization ID

jacekradko:
`OrganizationAPI.deleteOrganization()` was the lone ID-based method on that client without a `requireId()` guard. Since `joinPaths()` filters out falsy segments, `deleteOrganization('')` quietly built `/organizations` and fired a `DELETE` at the collection route instead of throwing locally. This adds the guard so an empty ID fails fast with `A valid resource ID is required.`, matching every sibling method.

Pre-existing on `main` (spotted by CodeRabbit on #8853); pulling it out as its own one-liner rather than folding a behavior change into that docs-only PR. The added test confirms the empty-ID call now rejects locally (it fails without the guard).

## Summary by CodeRabbit

* **Bug Fixes**
  * Organization deletion now validates that an organization ID is provided before sending the request.
  * If an empty ID is used, the app now returns a clear error message instead of attempting the action.
  
* **Tests**
  * Added coverage for the new validation behavior to ensure the error is shown as expected.
  
* **Chores**
  * Updated the release notes entry to document the behavior change.

### DIFF
--- a/.changeset/delete-organization-require-id.md
+++ b/.changeset/delete-organization-require-id.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
+`organizations.deleteOrganization()` now validates that an organization ID was provided. Calling it with an empty ID throws `A valid resource ID is required.` locally instead of issuing a `DELETE` request to the organizations collection endpoint, matching the other ID-based methods on the API.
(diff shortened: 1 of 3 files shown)


wobsoriano (approved):
nice catch

PR #9043, opened 2026-06-30
fix(electron): default allowedRedirectProtocols to the renderer scheme

nicolas-angelo:
## Description

Small DX win: today apps have to manually set `allowedRedirectProtocols={['<scheme>:']}` on `<ClerkProvider>` so Clerk's prebuilt-UI cross-links aren't rejected (i.e. the sign-up link on the SignIn component and the sign-in link on the SignUp component).

This PR makes it zero-config by deriving the scheme from the renderer's own origin (`window.location.protocol`, i.e. the scheme registered via `createClerkBridge`) and appending it to `allowedRedirectProtocols` internally by default.

The only exception to this new behavior is honoring explicit values (including `[]`) which will override.

**Testing:** drop `allowedRedirectProtocols` from `<ClerkProvider>` and confirm in-app cross-links (sign-in ↔ sign-up) on auth components work properly.

> **Note on changeset level:** scoped as a `patch` because the current default (no custom scheme in the allowlist) leaves the SDK's cross-links broken out of the box - so this feels more like a fix. 

## Checklist

- [x] JSDoc comments have been added for the new behavior.

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change

## Summary by CodeRabbit

* **New Features**
  * Electron apps now automatically allow the renderer’s own custom protocol for redirect flows when applicable (e.g., `clerk:`).
  * You can still explicitly control redirect protocols via an `allowedRedirectProtocols` setting, including setting it to none.
* **Bug Fixes**
  * Improved default redirect handling so custom renderer schemes work with no extra configuration.
  * `file:` renderers remain unchanged (not auto-allowed), and `https:` defaults are not added automatically.
* **Tests**
  * Added coverage for defaulting and override behavior of redirect protocols.


wobsoriano (approved):
looks good, thanks!

PR #9047, opened 2026-06-30
fix(ui): keep truncateWithEndVisible code-point-safe in short-width fallback

alexcarpenter:
`truncateWithEndVisible` keeps the start and end of a string and puts an ellipsis in the middle. The main path builds the result from `Array.from(str)`, so it slices on code points and never breaks a character. The short-width fallback (taken when `maxLength <= endChars + 3`) instead returned `ELLIPSIS + str.slice(-endChars)`, which slices on UTF-16 code units.

For characters outside the BMP (CJK Extension B kanji, emoji) each character is a surrogate pair of two code units, so `str.slice(-endChars)` can cut one in half and leave a lone surrogate:

```ts
truncateWithEndVisible('𠮷𠮷𠮷𠮷𠮷', 8, 5)
// before: '...\uDFB7𠮷𠮷'  (broken leading character)
// after:  '...𠮷𠮷𠮷𠮷𠮷'
```

The fallback now slices by code point, the same way the main path already does. ASCII inputs are unaffected and the existing tests still pass. Added a test covering the fallback with astral characters.

---

Re-creates #9029 from a branch in this repo so CI runs with the necessary secrets. Authored by @greymoth-jp; original commit authorship is preserved in git history.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text truncation so very short widths no longer split emoji or other non-BMP characters into broken symbols.
  * Made character counting and fallback truncation behave consistently with Unicode code points, preserving readable output.
  * Added coverage for Unicode edge cases to prevent regressions.

### DIFF
--- a/.changeset/truncate-surrogate-safe.md
+++ b/.changeset/truncate-surrogate-safe.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Stop `truncateWithEndVisible` from splitting characters outside the BMP (such as CJK Extension B kanji and emoji) into a broken replacement character when truncating to a very small width. The short-width fallback now slices by code point, matching the main truncation path.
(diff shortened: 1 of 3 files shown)


wobsoriano (approved):
👍🏼

PR #9065, opened 2026-07-01
fix(clerk-js): fail fast when the origin is slow at load

nikosdouvlis:
Why
When Clerk's Frontend API is slow or unreachable, a cold Clerk.load() hangs for minutes. The /client request has no timeout, and on a 5xx or network error the recovery path awaits getToken({ skipCache: true }), which burns the full ~162s retry budget before Clerk is marked loaded. So during an origin outage the app sits there unresponsive instead of degrading and rendering.

What changed
Bound the browser load with a timeout in two spots: around the /client fetch and around the recovery token mint. The /client fetch is aborted on timeout (a new abortSignal option threaded through Client.fetch and the resource GET path) so it can't keep the load hanging. On a slow or failed /client, recovery renders identity from the __session cookie, clears the session's token cache (via the existing clearCache()) so the mint bypasses the cache (no skipCache, so no force_origin), and keeps the cookie identity if the mint times out. Also adds a timeLimit util to @clerk/shared/utils.

The timeout is 7s, grounded in production traffic rather than picked by feel. Across ~1B GET /v1/client requests over 7 days measured at the Cloudflare edge, server-side latency is p99 ~0.55s and p99.9 ~1.5-2.2s (stable day over day). 7s clears that p99.9 with roughly 3x margin, leaving room for the client's last-mile network that edge metrics don't capture, so it fires only on a genuinely stuck origin and not on a slow-but-working load. Erring high is deliberate: a false trip only drops the user to the cookie-derived client for one tick before the background refetch reconciles, so there is little downside to the extra margin.

One known gap: the recovery mint's timeout stops waiting but doesn't cancel the underlying request, so a timed-out mint can keep retrying in the background and briefly write a slightly stale token before the poller reconciles. Low severity and self-correcting. This is phase 1; later phases cover fail-fast /environment and render-first init.

## Summary by CodeRabbit

* **New Features**
  * Added timeout-bounded, fail-fast client startup with improved background recovery so initialization can complete even when requests hang.
  * Improved cold-load identity availability using session-token-based fallback when available.
* **Bug Fixes**
  * Abort in-flight requests on initialization timeouts; prevent overlapping recovery and polling.
  * Discard stale late client responses and improve degraded-state recovery, including correct stub-user behavior until full user data loads.
  * Stop retries when requests are aborted; added abort/cancellation support for client fetch and resource GET reads.
* **Tests**
  * Expanded coverage for degraded startup, background client recovery, and timeout/abort scenarios.


nikosdouvlis:
!snapshot

nikosdouvlis:
!snapshot

nikosdouvlis:
!snapshot

PR #9069, opened 2026-07-01
feat(expo): Add support for Expo SDK 57

wobsoriano:
## Description

Adds support for [Expo SDK 57](https://expo.dev/changelog/sdk-57). SDK 57 is a small, focused release (React Native 0.86, React unchanged at 19.2) that Expo explicitly ships with no breaking changes from SDK 56.

I dropped 56 rather than an older row because 56 and 57 are the redundant pair. SDK 57 ships with no breaking changes from 56, so a native build passing on one effectively passes on the other. Keeping 54, 55, and 57 gives evenly spread React Native coverage (0.81, 0.83, 0.86) instead of two adjacent versions.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Added support for Expo SDK 57.
* **Updates**
  * Bumped app and package versions to align with the newer Expo SDK, including related React and React Native updates.
  * Expanded compatibility ranges so the packages work with the latest supported Expo releases.
* **Bug Fixes**
  * Updated the native build workflow to include the newer Android/Expo SDK combination.


.github/workflows/expo-native-build.yml
```diff
@@ -45,10 +45,10 @@ jobs:
           - expo-sdk: 55
             platform: ios
             runner: blacksmith-6vcpu-macos-26
-          - expo-sdk: 56
+          - expo-sdk: 57
             platform: android
             runner: blacksmith-8vcpu-ubuntu-2204
-          - expo-sdk: 56
+          - expo-sdk: 57
```
wobsoriano, line 51: I dropped 56 rather than an older row because 56 and 57 are the redundant pair. SDK 57 ships with no breaking changes from 56, so a native build passing on one effectively passes on the other. Keeping 54, 55, and 57 gives evenly spread React Native coverage (0.81, 0.83, 0.86) instead of two adjacent versions.

mikepitre: makes sense!

mikepitre (approved):

PR #9072, opened 2026-07-02
test(js): add ProtectChallenge sandbox scenario for protect-check flows

mwickett:
## What

A sandbox scenario for exercising the #8329 `protect_check` step without any backend — MSW mocks the FAPI responses (`protect_check` on sign-up create and after sign-in first-factor attempt, PATCH resolution, chaining) and serves a mock challenge module implementing the script contract.

Targets `theo/protect-check-sdk-support` so it rides with #8329 but stays separately reviewable.

## How to use

```
pnpm dev:sandbox
```

- `/sign-in?scenario=ProtectChallenge&protectChallengeMode=manual`
- `/sign-up?scenario=ProtectChallenge&protectChallengeMode=manual`

`protectChallengeMode=manual` pauses on a "Complete challenge" button (omit for auto-resolve). Add **`protectChallengeWidget=turnstile`** to render the real Cloudflare Turnstile widget inside the protect-check card using Cloudflare's universal test sitekeys (always-pass in auto mode, force-interactive in manual mode) — production pixels, mocked proof token. That mode's only network dependency is Cloudflare's CDN.

## Why

@alex-carpenter asked in [#team-protect](https://clerkinc.slack.com/archives/C0AJBKMHB51/p1782416473832539) how the challenge renders within the flow — this makes that a 30-second local check and gives the PR Playwright coverage for auto/manual resolution on both flows. Verified against a full local clerk_go + warden-go E2E run; the mocked response shapes match real FAPI.

## Testing status

The Playwright spec (`sandbox/integration/protect-challenge.spec.ts`) encodes the **intended** post-resolution behavior: after a challenge resolves, sign-up continues to email verification and sign-in activates the session. All three tests are annotated `test.fail()` because that behavior is currently broken on the base branch — once a check resolves, the flow never advances (details in the #8329 review discussion). The moment that's fixed, these flip to "unexpected pass"; delete the annotations and the spec becomes an enforced regression guardrail for exactly that bug.

Run it locally (the sandbox scenario mocks FAPI, so dummy keys suffice):

```bash
cd packages/clerk-js
PORT=4021   INTEGRATION_INSTANCE_KEYS='{"sandbox":{"pk":"pk_test_…","sk":"sk_test_dummy"}, …one entry per name referenced by integration/presets/envs.ts…}'   CLERK_TESTING_TOKEN=dummy   pnpm playwright test sandbox/integration/protect-challenge.spec.ts
```

Open question for the branch owner: `test:sandbox:integration` isn't part of CI today, so this spec only guards locally until it's wired into a workflow — happy to follow whatever the plan is for the other sandbox specs.

## Notes

- The scenario grafts a full `UserJSON` onto the mock session (the msw session template ships `user: null`), so the sign-in test asserts a genuinely hydrated `Clerk.user.id` rather than a placeholder.
- Empty changeset — sandbox/tooling only, no published-package changes.
- Two UI bugs found while testing the flow E2E will be left as review comments on #8329 (post-resolution auto-advance; retry duplicating the challenge widget).


mwickett:
Ok, closing this out because I don't like the idea of mocking this with a service worker. The real fix is to get the actual Integration Testing instance in staging configured to have Protect enabled.

PR #9099, opened 2026-07-07
fix(ui): hide protect-check spinner while challenge widget is visible

mwickett:
## Description

Fixes two UX issues in the Protect check card reported while dogfooding the challenge flow ([Slack report](https://clerkinc.slack.com/archives/C081A8SDN9F/p1783385907265719?thread_ts=1783381315.790449&cid=C081A8SDN9F)):

1. **The loading spinner kept spinning below the Turnstile widget** once it appeared — two competing progress indicators.
2. **The empty SDK container reserved `minHeight: 60px`**, leaving a hole of dead space between the header and the spinner before any widget rendered.

Root cause: the card only knows "the check is running" (`isRunning` spans the whole lifecycle — SDK load → widget render → solve → verify → proof submit), and the challenge-script contract had no way to say "I'm showing UI now".

### What changed

**`@clerk/shared`** — the `executeProtectCheck` contract gains an optional **visibility handshake**: the host passes `setWidgetVisible(visible: boolean): Promise<void>` in the script's init payload. The script calls it right before revealing UI in the container (and with `false` once its widget is done). The promise resolves only after the host has committed the change to the DOM, so the script can reveal its widget with **no overlap frame**; a script that knows its widget is imminent can call it immediately so the spinner never flashes. Scripts must treat the field as optional (older hosts don't send it).

**`packages/ui`** — `useProtectCheckRunner` implements the host side and exposes the signal-driven `isWidgetVisible`:

- The callback commits via `flushSync` (the same commit-before-returning guarantee `BaseRouter` relies on), then resolves — that ordering is what makes the "no overlap frame" promise real.
- It is constructed **per run**, closing over that run's `cancelled` flag, abort signal, and mount state: aborting a script is best-effort by contract, so a zombie script from a timed-out, retried, or superseded run can still call it late — those stale signals resolve as silent no-ops instead of flipping visibility under the active run.
- The spinner renders only while the check runs **and no widget is signalled visible**; the widget-visibility *decision* lives in the challenge script (the Turnstile implementation), not in DOM observation — no ResizeObserver/MutationObserver.
- The spinner's *entrance* is debounced with the existing `useSpinDelay` hook (300ms): a check that completes inside the window — or a script that signals its widget immediately — never flashes a spinner at all (the card header carries the first moments). The error and widget-visibility gates sit **outside** the delay hook, so its minimum-visible duration can never outrank the handshake's spinner-gone-at-resolve guarantee or hold a spinner next to the retry button.
- The empty container drops out of flex flow (`position: absolute`, the `CaptchaElement` `gapless` idiom) instead of reserving 60px.
- Each new run clears the container and resets visibility, so a retry or chained challenge starts on a clean spinner.
- Per design-system feedback in the thread: no visible "Loading…" text — the localized label rides on the `Spinner` as `aria-label` (the `Button` loading idiom; `Spinner` already announces via `aria-busy`/`aria-live`).

### Deploy ordering

The speed-bump script's side of the handshake is [clerk/protect#213](https://github.com/clerk/protect/pull/213) (guarded fallback, safe on older hosts). **Specter deploys before this reaches production**: this host listens only for the signal, so against an unsignaling old script the spinner stays under the widget and the container never re-enters layout flow.

### How to test

- `pnpm vitest run src/components/SignIn/__tests__/SignInProtectCheck.test.tsx src/components/SignUp/__tests__/SignUpProtectCheck.test.tsx` in `packages/ui` (visibility swap driven through the captured signal, async-confirm semantics, layout flip, retry hygiene, zombie-run guard, spin-delay entrance and its gates), and `src/internal/clerk-js/__tests__/protectCheck.test.ts` in `packages/shared` (contract forwarding).
- Manually: any sign-in gated by a `protect_check` against a Specter serving the #213 script — spinner alone during invisible phases, widget alone when interaction is required, spinner back during the verify round-trip.

Fixes PROT-863

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary

* **Bug Fixes**
  * Improved Protect check UI so the loading spinner no longer appears while the challenge widget is visible, and responds correctly to widget visibility changes in both Sign In and Sign Up.
  * Reduced layout “reserved” space so the cards stay compact until the widget renders.
  * Refined retry behavior to clear stale widget content and restore the correct loading state.
  * Updated loading presentation for improved accessibility (centered spinner with an aria label).

* **Tests**
  * Added deterministic tests covering challenge-widget visibility transitions and retry cleanup, with proper synchronization.


alexcarpenter (approved):
take a look at https://github.com/clerk/javascript/pull/9099#discussion_r3538789996 before merging but else lgtm

.changeset/protect-check-spinner-widget-visibility.md
```diff
@@ -0,0 +1,6 @@
+---
+'@clerk/shared': patch
+'@clerk/ui': patch
+---
+
+Polish the Protect check card: the loading spinner now hides while a challenge widget (e.g. Turnstile) is visible instead of spinning alongside it, only appears after a short delay so near-instant checks never flash it, and the card no longer reserves empty space above the spinner before a widget has rendered. Challenge scripts drive this through a new optional `setWidgetVisible` callback in the `executeProtectCheck` init payload; its promise resolves once the host has committed the change, so scripts can reveal their UI without overlap.
```
alexcarpenter:
```suggestion
Polish the Protect check card: the loading spinner now hides while a challenge widget (e.g. Turnstile) is visible instead of spinning alongside it, only appears after a short delay so near-instant checks never flash it, and the card no longer reserves empty space above the spinner before a widget has rendered.
```

focus on the user facing changes vs implementation details

alexcarpenter (approved):
one nit about changeset but LGTM

zourzouvillys (approved):

PR #9102, opened 2026-07-07
feat(ui): add experimental flag to gate the OIDC self-serve flow

iagodahlem:
## Summary

Adds an experimental `oidcSelfServe` flag to the ClerkProvider `experimental` options, laying the groundwork to gate a forthcoming OIDC self-serve configuration flow in `<ConfigureSSO />`.

- Opt in via `<ClerkProvider experimental={{ oidcSelfServe: true }} />` (default off).
- Surfaced to the ConfigureSSO wizard and all of its steps as `isOIDCFlowEnabled` via `useConfigureSSO()`.
- No user-facing surface is gated yet — the OIDC provider path lands in a follow-up PR. The flag lives in the SemVer-exempt `experimental` namespace, so there is no public API impact.

ORGS-1741

## Summary by CodeRabbit

* **New Features**
  * Added support for a new experimental self-serve OIDC option in SSO configuration.
  * The SSO setup flow now recognizes and passes through an additional enabled/disabled state for this experience.
* **Chores**
  * Included a changeset entry to track this update.


.changeset/empty-ducks-tease.md
```diff
@@ -0,0 +1,2 @@
+---
```
LauraBeatris, line 1: I believe we still need to include the version changes here to bump the packages and build it

Let's write that it's an internal change

LauraBeatris (approved):

PR #9103, opened 2026-07-07
feat(expo): add experimental useSSO with new hooks

swolfand:
## Summary

- Add an experimental `useSSO` export for `@clerk/expo/experimental` that uses the new `useClerk`, `useSignIn`, and `useSignUp` hooks.
- Preserve the existing Expo SSO behavior for redirect URL creation, browser auth sessions, rotating token nonce reloads, and transfer sign-ups.
- Add focused coverage for OAuth, enterprise SSO, transfer sign-up, dismissed browser sessions, error handling, and the experimental export.

Linear: MOBILE-582

## Testing

- `corepack pnpm --filter @clerk/expo exec vitest run src/hooks/__tests__/useSSO.experimental.test.ts src/hooks/__tests__/useSSO.test.ts`
- `corepack pnpm --filter @clerk/expo exec eslint src/experimental.ts src/hooks/useSSO.experimental.ts src/hooks/ssoDependencies.ts src/hooks/__tests__/useSSO.experimental.test.ts`
- `corepack pnpm --filter @clerk/expo format:check`
- Manual device test with `https://pkg.pr.new/@clerk/expo@9103`: Google SSO from the native development build returned to the app and signed in successfully.

## Notes

- Local `build:declarations` was blocked by workspace dependency resolution / pnpm non-interactive module purge behavior in this shell, before reaching this change.


packages/expo/src/hooks/useSSO.experimental.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  async function startSSOFlow(startSSOFlowParams: StartSSOFlowParams): Promise<StartSSOFlowReturnType> {
+    if (
+      !loaded ||
+      !client ||
+      !signIn ||
+      !signUp ||
+      signInFetchStatus === 'fetching' ||
+      signUpFetchStatus === 'fetching'
+    ) {
```
wobsoriano:
```suggestion
    if (!client || !signIn || !signUp) {
```

We can slim this down. The 6 conditions become truthy at the same moment (when Clerk is loaded). We need the remaining ones to narrow the resource types.

packages/expo/src/hooks/useSSO.experimental.ts
```diff
@@ -0,0 +1,151 @@
...
+  if (typeof error === 'object' && error !== null && 'message' in error && typeof error.message === 'string') {
+    return error.message;
+  }
+
+  return 'An unknown SSO error occurred';
+}
+
+function assertNoSSOError(result: { error: unknown }): void {
+  if (result.error) {
+    return errorThrower.throw(getSSOErrorMessage(result.error));
```
wobsoriano: `result.error` here is typed `ClerkError`. Passing it here flattens it to a plain `Error` losing `.code/.errors[]` that custom flows rely on

I think we can just do `throw result.error` to keep it structured (and both helpers can go). This matches the rest of the codebase as well. Let me know if that makes sense!

packages/expo/src/hooks/useSSO.experimental.ts
```diff
@@ -0,0 +1,151 @@
...
+      !signUp ||
+      signInFetchStatus === 'fetching' ||
+      signUpFetchStatus === 'fetching'
+    ) {
+      return {
+        createdSessionId: null,
+        authSessionResult: null,
+        signIn,
+        signUp,
+        setActive,
```
wobsoriano: Since this hook is built on the future resources, could we finalize internally and drop `setActive` from the return? `finalize()` also completes the future resource lifecycle, while raw `setActive` does not.

`HandleSSOCallback` follows this pattern [(ref)](https://github.com/clerk/javascript/blob/2148b4deff8471283e223c026bf4c979c066d0d4/packages/react/src/components/HandleSSOCallback.tsx#L116-L156), using `finalize()` for new sessions and `clerk.setActive()` only for eixsting session

wobsoriano (changes requested):
Looking good with minor changes!

wobsoriano (approved):

PR #9133, opened 2026-07-10
feat(ui): add composed profile shell, providers infra and stub router

alexcarpenter:
Stack 4/7. New (unexported) infra for the composed profile API: `ProfileProviderShell`, `stubRouter`, `useBillingRouter`, `useRequirePage`, `PageContext`, `createSection`, plus the shared `APIKeysSection`/`BillingSection` used by both User and Org composed dirs.

Nothing is exported from the package yet.

_Changesets live only on the release PR #9144; this PR carries none._

---

### Stack

Reviewed as a stack. The 7 PRs merge bottom-up into release branch #9144, which integrates into `main` as a single unit.

| # | PR | Layer |
|---|----|-------|
| – | #9144 | release → `main` (integration) |
| 1 | #9130 | moduleManager getter across build boundary |
| 2 | #9131 | shared profile UI infra |
| 3 | #9132 | extract Section components from profile pages |
| 4 | **#9133** 👈 | **composed shell + providers infra** |
| 5 | #9134 | composed UserProfile |
| 6 | #9135 | composed OrganizationProfile |
| 7 | #9136 | expose `@clerk/ui/experimental` (public API) |

### DIFF
--- a/packages/ui/src/composed/ProfileProviderShell.tsx
+++ b/packages/ui/src/composed/ProfileProviderShell.tsx
@@ -9,6 +9,7 @@
 // `styleCacheStore` so sibling composed roots don't duplicate style insertions.
 
 import { ClerkRuntimeError } from '@clerk/shared/error';
+import { logger } from '@clerk/shared/logger';
 import type { ModuleManager } from '@clerk/shared/moduleManager';
 import type { EnvironmentResource, LoadedClerk } from '@clerk/shared/types';
 // eslint-disable-next-line no-restricted-imports
@@ -19,7 +20,7 @@ import { useMemo } from 'react';
 import { AppearanceProvider } from '@/ui/customizables/AppearanceContext';
 import { FlowMetadataProvider } from '@/ui/elements/contexts';
 import type { Appearance, Elements } from '@/ui/internal/appearance';
-import { getStyleCache, setStyleCache } from '@/ui/internal/styleCacheStore';
+import { getStyleCacheEntry, setStyleCache } from '@/ui/internal/styleCacheStore';
 import { RouteContext } from '@/ui/router/RouteContext';
 import { InternalThemeProvider } from '@/ui/styledSystem';
 import { createEmotionCache } from '@/ui/styledSystem/createEmotionCache';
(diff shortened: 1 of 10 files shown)


packages/ui/src/composed/useBillingRouter.ts
```diff
@@ -0,0 +1,98 @@
...
+    case 'statement':
+      return { statementId: route.statementId };
+    case 'payment-attempt':
+      return { paymentAttemptId: route.paymentAttemptId };
+    default:
+      return {};
+  }
+}
+
+export function useBillingRouter(): { router: RouteContextValue; route: BillingRoute } {
```
brkalow, line 68: is this a pattern we're going to need elsewhere? We have the `VirtualRouter` component already, which could potentially be usable here in some fashion, but that one is sort of geared towards modal routing so might not slot in as cleanly.

Food for thought, don't think we need to adjust now but I expect to see this pattern emerge more than once.

packages/ui/src/composed/useRequirePage.ts
```diff
@@ -0,0 +1,16 @@
+import { useContext } from 'react';
+
+import { PageContext } from './PageContext';
+
+export function useRequirePage(componentName: string): boolean {
+  const page = useContext(PageContext);
+  if (!page) {
+    if (typeof __DEV__ === 'undefined' || __DEV__) {
```
brkalow: this would only throw when the package is being developed, not in Clerk dev mode. I think this is intentional, but calling it out regardless

packages/ui/src/internal/styleCacheStore.ts
```diff
@@ -7,13 +7,25 @@ import type { EmotionCache } from '@emotion/cache';
 // same key = duplicate style inserts, broken emotion dedup/ordering). AIO gets this
 // free from its single portal tree; composed reconstructs it by keying the cache
 // on the clerk instance here so sibling roots reuse one cache instead of each making
-// their own.
-const store = new WeakMap<object, EmotionCache>();
+// their own. The cache is stored with the nonce/cssLayerName it was built from so a
+// later change to either rebuilds it (matching the AIO StyleCacheProvider) instead
+// of pinning whatever the first-mounted root happened to see.
```
brkalow: not sure if this is really needed. I would recommend that the comment is trimmed down to more clearly explain why this construct exists, or removed.

packages/ui/src/composed/ProfileProviderShell.tsx
```diff
@@ -0,0 +1,202 @@
...
+  appearance,
+}: ProfileProviderShellProps): ReactNode {
+  // currentPath is left empty: composed has no Clerk-internal navigation. Each
+  // section owns its own CardStateProvider, so errors clear on section unmount
+  // (single-section mounting). Side-by-side sections keep independent error
+  // state — a consumer URL change wouldn't be a meaningful signal to clear
+  // either, so observing it would only cause spurious clears.
+  const router = useMemo(() => createComposedRouter(clerk.navigate), [clerk]);
+  // Match the portal path's normalization (Components.tsx:209) so a cssLayerName
+  // nested inside appearance.theme gets hoisted to top-level for @layer wrapping.
```
brkalow: this reference file / line no. will get out of date quickly

packages/ui/src/composed/ProfileProviderShell.tsx
```diff
@@ -0,0 +1,202 @@
...
+ * runtime is a real version mismatch and gets a one-time warning instead of a
+ * silent blank render.
+ */
+export function resolveComposedClerkRuntime(
+  clerk: LoadedClerk,
+  clerkLoaded: boolean,
+): { environment: EnvironmentResource | null | undefined; moduleManager: ModuleManager } {
+  // SAFETY: __internal_environment is a real getter on the clerk-js Clerk class
+  // but not on the shared LoadedClerk interface. Narrowing to an optional field
+  // (not `any`) keeps the access typed and forces callers to handle `undefined`.
```
brkalow:
```suggestion
```

brkalow (approved):

packages/ui/src/composed/useBillingRouter.ts
alexcarpenter: Agreed it'\''ll recur. Not reusing VirtualRouter: that'\''s BaseRouter <Route> matching, but composed is leaf-only (stubRouter throws on <Route>/matches). Typed in-memory state is the intended shape here. Added a comment noting why. Extract when a 2nd composed router shows up.

packages/ui/src/composed/useRequirePage.ts
alexcarpenter: Good catch — was dead for consumers (`__DEV__` baked false in published @clerk/ui). Now gated on `isDevelopmentSDK(clerk)`, same as createCustomPages: throws in consumer dev, null in prod. Throw tests mark the fixture as a dev SDK.

packages/ui/src/internal/styleCacheStore.ts
alexcarpenter: Trimmed to 3 lines.

packages/ui/src/composed/ProfileProviderShell.tsx
alexcarpenter: Dropped the file:line ref.

packages/ui/src/composed/ProfileProviderShell.tsx
alexcarpenter: Compressed to one line (kept a terse why for the cast) + dropped the duplicate type-def comment above it.

PR #9135, opened 2026-07-10
feat(ui): add composed OrganizationProfile panels and section components

alexcarpenter:
Stack 6/7. Composed `OrganizationProfile` panels and section components, plus the cross-provider wiring and style-cache-sharing tests (they need both providers, so they land here).

Not yet wired to the public entrypoint.

_Changesets live only on the release PR #9144; this PR carries none._

---

### Stack

Reviewed as a stack. The 7 PRs merge bottom-up into release branch #9144, which integrates into `main` as a single unit.

| # | PR | Layer |
|---|----|-------|
| – | #9144 | release → `main` (integration) |
| 1 | #9130 | moduleManager getter across build boundary |
| 2 | #9131 | shared profile UI infra |
| 3 | #9132 | extract Section components from profile pages |
| 4 | #9133 | composed shell + providers infra |
| 5 | #9134 | composed UserProfile |
| 6 | **#9135** 👈 | **composed OrganizationProfile** |
| 7 | #9136 | expose `@clerk/ui/experimental` (public API) |

### DIFF
--- a/packages/ui/src/composed/__tests__/composed-provider-wiring.test.tsx
+++ b/packages/ui/src/composed/__tests__/composed-provider-wiring.test.tsx
@@ -1,0 +1,569 @@
+import { act } from '@testing-library/react';
+import { useContext } from 'react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { bindCreateFixtures } from '@/test/create-fixtures';
+import { render, screen, waitFor } from '@/test/utils';
+import { useModuleManager, useOptions } from '@/ui/contexts';
+import { useAppearance } from '@/ui/customizables/AppearanceContext';
+import { useRouter } from '@/ui/router';
+
+import { OrganizationProfileContext } from '../../contexts/components/OrganizationProfile';
+import { UserProfileContext } from '../../contexts/components/UserProfile';
+import { clearFetchCache } from '../../hooks';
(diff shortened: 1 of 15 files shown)


packages/ui/src/composed/__tests__/composed-provider-wiring.test.tsx
```diff
@@ -0,0 +1,569 @@
...
+  return (
+    <div
+      data-testid='router-probe'
+      data-has-navigate={typeof router.navigate === 'function'}
+      data-has-base-navigate={typeof router.baseNavigate === 'function'}
+    />
+  );
+}
+
+describe('UserProfileProvider wiring', () => {
```
brkalow, line 22: These tests are leaning a bit too far towards testing implementation details for my liking. Would rather see tests that assert on the output of the components, e.g. what a user would see

brkalow (approved):

packages/ui/src/composed/__tests__/composed-provider-wiring.test.tsx
alexcarpenter: Reworked. Appearance + localization now assert rendered output (computed $primary500 color / visible French text) instead of probing context. Dropped the pure impl-detail probes (typeof navigate, currentPath decoupling, apiKeysProps/afterLeave context shape, weak !!mm). Kept getter-resolution + fallback: bundle-boundary contract with no user-visible surface. 269 lines removed.

PR #9140, opened 2026-07-11
fix(expo): deduplicate native client startup requests

mikepitre:
Expo native-client startup could trigger repeated client refreshes across the JS and native runtimes. In the captured SDK 56 launch, that produced 11 Clerk requests, including eight identical `GET /v1/client` calls.

The duplicate requests came from native configuration racing JS client hydration, observer callbacks echoing bootstrap and JS-originated changes, and device-token updates being followed by a second explicit refresh. This change waits for the JS Clerk instance before reading its initial token, configures native once with that token, and reconciles the two clients after both sides are ready. Native observers now seed their initial snapshot and suppress bootstrap or JS-originated echoes without dropping later native changes.

Token-cache notifications are deduplicated, client changes that occur before native is ready are coalesced, and configure failures still release the synchronization barrier. Focused tests cover initial bootstrap, token rotation, remount recovery, configure failure, and native-to-JS reconciliation.

With the SDK 56 hosted-auth example unchanged, process-cold launches on iOS and Android each produce four Clerk requests: one JS and one native request for `/v1/environment`, plus one JS and one native request for `/v1/client`.

## Summary by CodeRabbit

- **Improvements**
  - Reduced redundant native and JavaScript client refreshes during Expo startup and reconfiguration.
  - Improved device-token and client-state synchronization, including better handling of reloads and avoiding unnecessary updates.
  - Prevented missing or duplicate client-change notifications during initialization and sync flows.

- **New Features**
  - Added an `enabled` option to control whether native client event syncing runs, including queuing and merging refresh actions until readiness.

- **Tests**
  - Expanded hook and provider coverage to verify conditional subscription and synchronization behavior.


wobsoriano (approved):
Awesome! The Expo E2E test passed 🙏🏼

PR #9145, opened 2026-07-13
fix(backend): Fix cross-origin handshake bypass

dominic-clerk:
## Description

Fix a cross-origin handshake bypass where `isKnownClerkReferrer()` trusted overly broad referrer hosts as Clerk-owned: any `accounts.*` host (e.g. `accounts.attacker.com`), plus dev account-portal domains (`*.accounts.dev` and legacy suffixes) on production instances. These let unrelated origins skip the handshake and its session-freshness check. The referrer is now trusted only for the accounts portal derived from the instance's frontend API, plus dev account-portal domains on non-production instances.

Fixes SDK-143

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/backend/src/tokens/__tests__/request.test.ts
```diff
@@ -2058,7 +2058,8 @@ describe('tokens.authenticateRequest(options)', () => {
       });
     });
 
-    test('does not trigger handshake when referer is from production accounts portal', async () => {
```
dominic-clerk, line 2088: We revert the logic of this test because it tested that example.com's account portal wouldn't trigger a handshake on another (primary.com's) instance.

wobsoriano (approved):
Had to rebase because break check somehow reports a breaking change, even when there's none

👍🏼

PR #9153, opened 2026-07-14
fix(backend): Add required provider field to CreateEnterpriseConnectionParams

manovotny:
## Description

`CreateEnterpriseConnectionParams` was missing the `provider` field, but the Backend API requires it (`validate:"required"` on BAPI's `CreateParams`). Calls to `createEnterpriseConnection()` without it type-checked fine and then failed validation at runtime. The type now carries a required `provider`, typed as the `OrganizationEnterpriseConnectionProvider` union from `@clerk/shared/types` — the same narrow-union pattern `SamlConnectionApi` already uses — so both missing and unsupported values are caught at compile time.

Verified against the clerk_go source: `provider` is required on create, and the shared union exactly matches the server allowlist — `saml_custom`, `saml_okta`, `saml_google`, `saml_microsoft`, `oidc_custom`, `oidc_github_enterprise`, and `oidc_gitlab`.

Two side effects worth calling out:

- Making the field required is a compile-time break for TypeScript callers that omit it — but those calls were already broken at runtime, so this converts a runtime failure into a type error. Shipping as a patch for that reason.
- The tanstack-start enterprise SSO integration test was already passing `provider` behind an `as` cast to work around the missing field. The cast is now removed.

To test: `pnpm --filter @clerk/backend test:node EnterpriseConnectionApi.test.ts` — both create tests pass `provider` and assert it reaches the request body, and type-level assertions lock in that `provider` is required and rejects unsupported values.

A follow-up will align the remaining create-params gaps with the BAPI contract (`name` and `domains` are also required server-side).

Discovered while reviewing clerk/clerk#2887, where a docs code sample calling `createEnterpriseConnection` without `provider` would fail BAPI validation despite type-checking.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Bug Fixes**
  * Enterprise connection creation now requires a supported provider, preventing invalid requests and related runtime failures.
  * Added validation for supported SAML and OIDC provider values.

* **Tests**
  * Expanded coverage to verify provider values are included for SAML and OIDC connections.
  * Added type-level checks to reject unsupported provider values.

### DIFF
--- a/.changeset/heavy-melons-argue.md
+++ b/.changeset/heavy-melons-argue.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
+Add the required `provider` field to `CreateEnterpriseConnectionParams`. The Backend API has always required `provider` when creating an enterprise connection, so calls to `createEnterpriseConnection()` without it type-checked but failed at runtime. The parameter type now reflects the API contract.
--- a/.changeset/heavy-melons-argue.md
+++ b/.changeset/heavy-melons-argue.md
@@ -2,4 +2,4 @@
 '@clerk/backend': patch
 ---
 
-Add the required `provider` field to `CreateEnterpriseConnectionParams`. The Backend API has always required `provider` when creating an enterprise connection, so calls to `createEnterpriseConnection()` without it type-checked but failed at runtime. The parameter type now reflects the API contract.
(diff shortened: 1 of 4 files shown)


manovotny:
Folded into #9155, which now targets `main` and contains these commits plus the rest of the params-side alignment (required `name`/`domains`, deprecations, and the missing optional params). Collapsing the stack to two PRs — params (#9155) and response (#9156) — to make review and merge order simpler.

PR #9154, opened 2026-07-14
feat(expo): Part 1 - add native logo max height theme option

swolfand:
## Summary

- add a `logoMaxHeight` prop to Expo's native `AuthView`
- forward the prop through the Android and iOS native view bridges
- preserve the configured Android theme while overriding `ClerkDesign.logoMaxHeight` for that `AuthView`
- apply the value through `.clerkAppIcon(maxHeight:)` on iOS
- add prop-forwarding coverage and a patch changeset

## Why

The native Clerk Android and iOS SDKs support configuring the managed application logo height, but Expo apps could not pass that value to `AuthView`.

Because the setting only affects authentication screens, it is exposed as a component prop rather than as part of the global native theme configuration:

```tsx
<AuthView logoMaxHeight={64} />
```

simulator_screenshot_B014A7EA-7CC6-4F73-8E35-8E2FE6E4433A

image

## Summary by CodeRabbit

* **New Features**
  * Added an optional `logoMaxHeight` setting to Expo’s native `AuthView`.
  * Allows controlling the maximum managed logo height on Android and iOS (density-independent pixels), defaulting to 44.
* **Tests**
  * Added coverage confirming `logoMaxHeight` is forwarded to the native authentication view.


mikepitre:
Is this approach better than a param on the AuthView? This feels very AuthView specific to be sticking in the theme

swolfand:
@mikepitre good call, switched that

packages/expo/ios/ClerkAuthNativeView.swift
```diff
@@ -22,6 +23,13 @@ public class ClerkAuthNativeView: ClerkNativeViewHost {
     setNeedsHostedViewUpdate()
   }
 
+  func setLogoMaxHeight(_ logoMaxHeight: Double?) {
```
seanperez29: Can this parameter directly take a ``CGFloat?`` instead of conversion after?

packages/expo/.npmrc
```diff
@@ -0,0 +1 @@
+
```
seanperez29: Can this be removed?

wobsoriano (approved):
Looks good, thanks for adding screenshots!

PR #9155, opened 2026-07-14
fix(backend): Align enterprise connection create and update params with the Backend API

manovotny:
## Description

Aligns `CreateEnterpriseConnectionParams` and `UpdateEnterpriseConnectionParams` with the Backend API contract (`clerk_go` `api/bapi/v1/enterprise_connections/service.go`). Folds in #9153 — the provider fix and this work are one thought: the create/update params didn't match what BAPI accepts. Every claim was verified against the Go source.

Discovered while reviewing clerk/clerk#2887, where a docs code sample calling `createEnterpriseConnection` without `provider` would fail BAPI validation despite type-checking.

**Required `provider` field (from #9153)**

- `CreateEnterpriseConnectionParams` was missing `provider`, which BAPI validates as required. Calls without it type-checked and failed at runtime. The field is typed as the `OrganizationEnterpriseConnectionProvider` union from `@clerk/shared/types` (matching the sibling `SamlConnectionApi` pattern), with type-level tests so it can't silently become optional or widen back to `string`.
- The tanstack-start enterprise SSO integration test was already passing `provider` behind an `as` cast to work around the missing field. The cast is removed.

**Required `name` and `domains`, plus deprecations**

- `name` and `domains` are now required on create. BAPI validates both (`validate:"required"`, `min=1` on domains), so calls without them already failed at runtime — the types now catch it at compile time. Type-level tests lock each field individually.
- `syncUserAttributes` on create is deprecated. BAPI's `CreateParams` has no `sync_user_attributes` field; the JSON decoder drops it silently. It works on update, and the deprecation notice points there.
- `provider` on update is deprecated. BAPI's `UpdateParams` has no `provider` field — the provider can't be changed after creation.

**Missing optional params**

- Adds the optional fields BAPI accepts that the SDK types omitted: `allowOrganizationAccountLinking`, `customAttributes`, `authenticatable`, `disableJitProvisioning` (create and update), `disableAdditionalIdentifications` (update only), and `saml.loginHint`.
- `loginHint` is a discriminated union: `source` is required exactly when `mode` is `'custom_attribute'` and rejected otherwise, matching BAPI's validation. Custom attribute shapes (`name` and `key` required) match the Go structs, and all camelCase names were confirmed to snake-case to the exact wire keys.

Making previously-optional fields required is a compile-time change, but only for calls that already failed BAPI validation at runtime — it converts a runtime failure into a type error. Ships as a patch for that reason.

To test: `pnpm --filter @clerk/backend test:node EnterpriseConnectionApi.test.ts` — wire-format tests assert every field reaches the request body in snake_case, alongside the type-level assertions.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


wobsoriano (approved):
this looks legit to me

PR #1544, opened 2023-08-01
Update the resources / types to include has_Image

dimkl:
## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Packages affected

- [x] `@clerk/clerk-js`
- [ ] `@clerk/clerk-react`
- [ ] `@clerk/nextjs`
- [ ] `@clerk/remix`
- [x] `@clerk/types`
- [ ] `@clerk/themes`
- [ ] `@clerk/localizations`
- [ ] `@clerk/clerk-expo`
- [x] `@clerk/backend`
- [ ] `@clerk/clerk-sdk-node`
- [ ] `@clerk/shared`
- [ ] `@clerk/fastify`
- [ ] `@clerk/chrome-extension`
- [ ] `gatsby-plugin-clerk`
- [ ] `build/tooling/chore`

## Description

- [ ] `npm test` runs as expected.
- [ ] `npm run build` runs as expected.

PR #9156, opened 2026-07-14
fix(backend): Align EnterpriseConnection resource with the Backend API response

manovotny:
## Description

Stacked on #9155. The `EnterpriseConnection` resource classes and JSON types drifted from what BAPI's serializer returns (`clerk_go` `api/serialize/enterprise_connection.go`). Some properties read fields BAPI never sends — typed `string`/`boolean` but always `undefined` at runtime — and several returned fields were missing from the SDK.

What changed:

- `EnterpriseConnection` now exposes `provider`, `logoPublicUrl`, `allowOrganizationAccountLinking`, `authenticatable`, `disableJitProvisioning`, and `customAttributes`.
- `EnterpriseConnectionSamlConnection` now exposes `active`, `forceAuthn`, and `loginHint` (required and non-null — BAPI's serializer always sends it).
- `EnterpriseConnectionOauthConfig` now exposes `providerKey`, `authUrl`, `tokenUrl`, `userInfoUrl`, and `requiresPkce`.
- Phantom fields are deprecated, not removed: top-level `allowSubdomains` (only exists on the nested SAML connection) and nested `idpMetadata`/`syncUserAttributes`. Removing them would break positional constructor callers, so they keep their slots and gain `@deprecated` notices pointing at the real field.
- `organizationId` is normalized to `null` when BAPI omits the key, making the declared `string | null` true at runtime.
- Fields backed by pointer+`omitempty` in Go (`oauthConfig.clientId`, the SAML `idp_*` fields, certificate timestamps) are retyped `| undefined` to match runtime behavior.

New constructor params are appended after existing ones. Adding required constructor params to resource classes in a non-major release follows repo precedent (#1544 added `hasImage` mid-constructor as a minor); this ships as a **minor** for the same reason.

One review call worth a second opinion: the `| undefined` retyping is a compile-time change for consumers with strict null checks. It converts silent runtime `undefined`s into visible type errors, but it will surface in downstream builds on upgrade.

> [!NOTE]
> Review surfaced an upstream gap: BAPI's `EnterpriseConnectionResponse` is missing the `object` discriminator its sibling serializers all send, so the SDK's deserializer (which switches on `object`) currently returns raw snake_case JSON for enterprise connection responses in production — `fromJSON` only runs in tests, where the mock supplies `object`. The mock keeps `object` deliberately: it exercises the SDK path that the existing `ObjectType.EnterpriseConnection` deserializer case defines, and matches the contract once clerk_go adds the field. The upstream fix is filed separately — [ORGS-1765](https://linear.app/clerk/issue/ORGS-1765).

To test: `pnpm --filter @clerk/backend test`

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


wobsoriano (approved):
Looks good!

Re `| undefined question`: The old types were wrong. Any downstream code that breaks on upgrade was already unsafe at runtime.

If there's  a compiler error then it's just surfacing existing bugs rather than creating new ones. Minor bump is perfect!

PR #9167, opened 2026-07-15
fix(electron): validate IPC sender frame for token-cache and OAuth handlers

dominic-clerk:
## Description

Token-cache and OAuth-transport IPC handlers accepted requests from any renderer frame. Untrusted content in a subframe or <webview> sharing the Clerk preload could read the persisted client JWT or drive the OAuth transport. Guard both against non-main-frame senders, mirroring the existing passkey handler check via a shared isMainFrameEvent helper.

Fixes SDK-142

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

## Summary by CodeRabbit

* **Security**
  * Restricted token-cache, OAuth transport, and passkey IPC operations to requests originating from the app’s main window frame.
  * Blocked untrusted subframes and embedded webviews from accessing stored tokens or initiating OAuth/passkey flows.
  * Prevented redirect handling and external URL launching when triggered from non-main-frame requests.
* **Bug Fixes**
  * Strengthened validation and expanded test coverage to ensure these main-frame protections are correctly enforced and cleaned up.

### DIFF
--- a/packages/electron/src/main/validate-sender.ts
+++ b/packages/electron/src/main/validate-sender.ts
@@ -1,0 +1,6 @@
+import type { IpcMainInvokeEvent } from 'electron';
+
+// Restricts privileged IPC to a window's top-level frame so untrusted subframes/webviews sharing the Clerk preload cannot reach it.
+export function isMainFrameEvent(event: IpcMainInvokeEvent): boolean {
+  return Boolean(event.senderFrame) && event.senderFrame === event.sender.mainFrame;
+}
--- a/packages/electron/src/main/validate-sender.ts
+++ b/packages/electron/src/main/validate-sender.ts
@@ -1,6 +1,10 @@
 import type { IpcMainInvokeEvent } from 'electron';
 
-// Restricts privileged IPC to a window's top-level frame so untrusted subframes/webviews sharing the Clerk preload cannot reach it.
+// Restricts privileged IPC to a top-level BrowserWindow's main frame. The frame check alone
+// only proves the sender is the top document of *its own* WebContents, so a <webview> (a separate
+// WebContents) sharing the Clerk preload would still pass; requiring getType() === 'window' rejects
+// <webview> guests and BrowserViews, and the frame check rejects in-page iframes.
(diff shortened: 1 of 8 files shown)


packages/electron/src/main/validate-sender.ts
```diff
@@ -0,0 +1,6 @@
+import type { IpcMainInvokeEvent } from 'electron';
+
+// Restricts privileged IPC to a window's top-level frame so untrusted subframes/webviews sharing the Clerk preload cannot reach it.
+export function isMainFrameEvent(event: IpcMainInvokeEvent): boolean {
+  return Boolean(event.senderFrame) && event.senderFrame === event.sender.mainFrame;
```
dominic-clerk: **[MEDIUM] `isMainFrameEvent` checks frame hierarchy, not WebContents identity — `<webview>` bypass**

`isMainFrameEvent` returns `event.senderFrame === event.sender.mainFrame`, which only proves the sender is the top frame of *its own* WebContents — it never ties the request to the trusted app window. Since the token-cache/OAuth handlers are registered globally (`setupTokenCacheIpcHandlers` / `setupOAuthTransportIpcHandlers`) with no `event.sender` identity check, a `<webview>` (a separate WebContents) satisfies `senderFrame === mainFrame` for its top document and passes the guard. Untrusted content in a `<webview>` sharing the Clerk preload can then call `TOKEN_CACHE_CHANNELS.getToken` to read the persisted client JWT or drive `OAUTH_TRANSPORT_CHANNELS.open` — the exact `<webview>` case the changeset claims to block (only in-page iframes are actually blocked). Recommend additionally verifying `event.sender` against the registered main-window `WebContents`.

jeremy-clerk:
Good call. Looks good from reading the code. Will run validation e2e to ensure no breakage. Thank you!

jeremy-clerk (approved):
👍 want to get @wobsoriano's input as well though

wobsoriano (approved):
Nice fix and thanks for the test coverage!

PR #9173, opened 2026-07-16
fix(expo): prevent initialization retry leaks

mikepitre:
## Description

When the Frontend API (FAPI) is unavailable, Expo falls back to cached resources and schedules `__internal_reloadInitialResources()`. Fast Refresh and repeated reloads can run this path several times during local development. Each cache miss could start an independent recursive retry timer, and old timers were not cancelled when the Clerk instance was replaced.

I observed hundreds of simulator TCP connections in Proxyman, eventually exhausting macOS ephemeral ports.

## Current approach

This PR:

- Allows only one pending retry timer or request per Clerk instance.
- Cancels the previous instance's retry timer when the instance is replaced.
- Stops retrying after resources load successfully.
- Uses exponential backoff capped at 30 seconds.

The retry timing is:

- The initial cache-miss retry remains 3 seconds.
- Network failures retry after 2, 4, 8, 16, and then 30 seconds.
- Other retryable failures, including 5xx responses, retry after 10, 20, and then 30 seconds.
- Retries continue indefinitely at the 30-second cap.

Healthy initialization is unchanged. During a prolonged outage, a real app may take up to 30 seconds to detect that FAPI has recovered.

## Alternative approach

Keep the same single-flight and cancellation behavior, but preserve the existing fixed retry timing:

- Network failures retry every 2 seconds.
- Other retryable failures retry every 10 seconds.
- Retries stop after resources load successfully.

### Difference

- The current approach sends fewer requests during a long outage, but recovery may be detected later.
- The alternative preserves the existing production recovery timing, but continues calling FAPI every 2 or 10 seconds while it is unavailable.
- Both approaches prevent stacked retry loops and stale instances from continuing to retry.

The lifecycle tests cover duplicate cache misses, an in-flight request, replacement after a failed attempt, successful recovery, and capped backoff.

## Checklist

- [x] `pnpm --filter @clerk/expo test`
- [x] `pnpm --filter @clerk/expo build`
- [x] No package exports changed
- [x] No documentation update is needed for this internal fix

## Type of change

- [x] 🐛 Bug fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Expo resource initialization when the Clerk API is unavailable or cached resources are missing.
  * Prevented duplicate recovery requests during app reloads or repeated resource misses.
  * Added cancellable retries with increasing delays, capped at 30 seconds.
  * Stopped retries after successful initialization or when the active instance changes.

* **Chores**
  * Prepared a patch release for `@clerk/expo`.

### DIFF
--- a/packages/expo/src/provider/singleton/createClerkInstance.ts
+++ b/packages/expo/src/provider/singleton/createClerkInstance.ts
@@ -44,6 +44,9 @@ type ResolvedClerkRuntimeOptions = Omit<ClerkRuntimeOptions, 'publishableKey'> &
   publishableKey: string;
 };
 
+const RESOURCE_RETRY_MAX_DELAY_MS = 30_000;
+const RESOURCE_RETRY_MAX_EXPONENT = 4;
+
 function hasOwnOption<Key extends keyof BuildClerkOptions>(
   options: BuildClerkOptions | undefined,
   key: Key,
@@ -53,6 +56,7 @@ function hasOwnOption<Key extends keyof BuildClerkOptions>(
 
 let __internal_clerk: HeadlessBrowserClerk | BrowserClerk | undefined;
 let __internal_clerkOptions: ClerkRuntimeOptions | undefined;
+let __internal_cancelResourceRetries: (() => void) | undefined;
 // Token IO can change without recreating the native singleton.
 let __internal_tokenCache: TokenCache = MemoryTokenCache;
 
@@ -110,6 +114,9 @@ export function createClerkInstance(ClerkClass: typeof Clerk) {
     if (!__internal_clerk || hasConfigChanged) {
       assertValidProxyUrl(proxyUrl);
 
+      __internal_cancelResourceRetries?.();
+      __internal_cancelResourceRetries = undefined;
+
       if (hasConfigChanged) {
         void __internal_tokenCache.clearToken?.(CLERK_CLIENT_JWT_KEY);
       }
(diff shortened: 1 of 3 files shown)


packages/expo/src/provider/singleton/createClerkInstance.ts
```diff
@@ -162,17 +170,64 @@ export function createClerkInstance(ClerkClass: typeof Clerk) {
...
-                // Retry after 2 seconds if the error is a network error
-                // Retry after 10 seconds if the error is a 5xx FAPI error
-                const timeout = isClerkNetworkError(err) ? 2000 : 10000;
-                setTimeout(() => void retryInitilizeResourcesFromFAPI(), timeout);
+                const initialRetryTimeout = isClerkNetworkError(err) ? 2000 : 10000;
+                nextRetryTimeout = Math.min(
+                  initialRetryTimeout * 2 ** resourceRetryFailureCount,
+                  RESOURCE_RETRY_MAX_DELAY_MS,
+                );
+                resourceRetryFailureCount = Math.min(resourceRetryFailureCount + 1, RESOURCE_RETRY_MAX_EXPONENT);
```
wobsoriano, line 223: Follow-up idea: `@clerk/shared/retry` (already used elsewhere in the monorepo for this kind of backoff) could replace this and gives us jitter for free. It needs to have an option to cancel first... happy to help add that in a follow-up PR

wobsoriano (approved):
Thanks for chasing this down!

packages/expo/src/provider/singleton/createClerkInstance.ts
mikepitre: sounds good!

PR #9174, opened 2026-07-16
feat(clerk-js): Use FAPI for OAuth popup callbacks

brkalow:
OAuth popup flows currently depend on the Account Portal hostname even when an application sends all Frontend API traffic through its own proxy. That leaves a hostname dependency outside the proxy and prevents the popup flow from staying on the application's domain.

This changes Clerk.js to use the new FAPI popup callback route by default for both existing and future SignIn/SignUp flows. Clerk.js generates a random state value before starting the flow and accepts the callback message only when its origin matches the effective FAPI origin, its source is the opened popup, and its state matches. Direct FAPI and proxied FAPI URLs use the same path-building behavior.

Depends on https://github.com/clerk/clerk_go/pull/20467 being deployed before this Clerk.js version is released.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OAuth and SSO popup authentication reliability.
  - Popup callbacks now use the Frontend API callback route instead of the Account Portal hostname.
  - Added secure state validation to prevent unrelated messages from completing authentication.
  - Preserved popup routing information across sign-in and sign-up verification flows.
- **Tests**
  - Expanded coverage for popup redirects, callback URLs, state handling, origin validation, and successful navigation.


packages/clerk-js/src/utils/authenticateWithPopup.ts
```diff
@@ -1,8 +1,28 @@
-import { buildAccountsBaseUrl } from '@clerk/shared/buildAccountsBaseUrl';
 import type { AuthenticateWithPopupParams, AuthenticateWithRedirectParams } from '@clerk/shared/types';
 
 import type { Clerk } from '../core/clerk';
 
+const popupCallbackPath = '/popup_auth_callback';
```
dominic-clerk, line 5: **[MEDIUM] FAPI popup callback leaks the active session ID to any origin; `state` is not validated server-side**

Pointing the popup flow at `/popup_auth_callback` makes the FAPI handler the default callback for all popup sign-in/sign-up. That handler (`api/fapi/v1/auth_popup_callback/http.go` in clerk_go) posts its payload with `targetOrigin` `"*"` and only echoes back the `state` it was given, never checking that it minted it. The route needs no auth (`router.go:622`) and is in the `ignoredExactPaths` set that `validateRequestOrigin` skips (`auth_middleware.go:111`), and `__client` is `SameSite=Lax`, so it rides along on the top-level GET that opening a popup performs. Any site can therefore call `window.open('https://clerk.<app>.com/v1/popup_auth_callback?state=x')` and read the reply, learning whether the visitor is signed in and what their active session ID is.

The `state` added here correctly stops a spoofed message from being accepted by clerk-js, but it cannot stop an attacker who initiates the callback themselves. The fix belongs on the FAPI side: bind `state` to a server-side pending popup attempt so an unknown state returns nothing, and post to a concrete allowlisted origin instead of `"*"`.

PR #9182, opened 2026-07-16
feat(shared): Add cancellation and per-error delay options to retry

wobsoriano:
## Description

Follow-up to #9173, which added a hand-rolled exponential backoff to Expo's initialization resource recovery because the shared retry helper couldn't express two things it needed. This adds them:

- `signal?: AbortSignal`: cancels retrying. Aborting rejects the returned promise with the signal's abort reason and immediately interrupts a pending backoff delay. It does not abort a callback that is already executing.
- `initialDelay` now also accepts a function `(error, iteration) => number`, deriving the backoff base delay from the error that triggered the retry (e.g. 2s for network errors, 10s for 5xx). The exponential factor, cap, and jitter still apply.

Both changes are optional and additive. Defaults and existing behavior are unchanged, and no current retry consumer is affected.

A follow-up PR will migrate the Expo recovery logic from #9173 onto this helper, gaining jitter on its retry delays.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/shared/src/retry.ts
```diff
@@ -120,9 +162,9 @@ export const retry = async <T>(callback: () => T | Promise<T>, options: RetryOpt
       }
```
mikepitre, line 168: Should we handle aborts while `onBeforeRetry` is running too? Right now, if the signal aborts while an async hook is pending, `retry()` waits for the hook to finish. If the hook rejects, its error also wins over the abort reason. Could we make this path follow the same cancellation behavior as the retry delay?

wobsoriano: Thanks for checking this out! Pushed a fix ([ref](https://github.com/clerk/javascript/pull/9182/changes/be31b94b0804a9312afcdb624e9e4cf86c3d2ed6)) for the error precedence. If the hook rejects after an abort, the abort reason now wins, same as the callback path

I left the pending-hook behavior as is. We don't interrupt user code, same as the callback, and the abort still wins as soon as the hook settles. I also updated the JSDoc to document this.

PR #9184, opened 2026-07-16
feat(ui): add UserButton view component

alexcarpenter:
## Description

The Mosaic `UserButton` surface, presentation only. The controller that connects it to Clerk is a follow-up.

`user-button.view.tsx` exports the parts (`UserButtonRoot`, `UserButtonTrigger`, `UserButtonPopup`) and the all-in-one `UserButtonView`, which frees the bare `UserButton` name for the connected container. `user-button.types.ts` holds the data contract that container has to satisfy, and `userButtonBusyKeys` names every affordance that can spin.

What the surface offers:

- A workspace list of organization memberships, suggestions, and invitations, with a personal workspace row an app can withhold.
- The signed-in accounts, with sign-out for one account and for all of them.
- `mode` narrows the surface to organizations or to accounts. `modePriority` picks which one a combined surface leads with, in its header and in the trigger.
- `renderTriggerLabel` names the active workspace beside the avatar, `renderPlanBadge` shows its plan.
- `customMenuItems` and `menuItemOrder` add the app's own rows to the foot of the menu, each row either an action or a link.
- `pendingKey` spins the affordance that owns the action in flight and stands the others down.
- A loading row that holds the workspace list's place until its first page lands.
- The Clerk logo in the footer, linking to Clerk.

`@clerk/headless` gains what the surface needs from its popover: contents stay mounted while it closes, `Popover.Root` takes an `alignOffset`, and Escape inside a menu no longer bubbles to the floating element around it.

Swingset documents the surface and renders it as a live prototype.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/swingset/src/lib/registry.ts
```diff
@@ -105,6 +105,12 @@ import {
 } from '../stories/text.stories';
 import { meta as tooltipMeta } from '../stories/tooltip.stories';
 import { meta as useDataTableMeta } from '../stories/use-data-table.stories';
+import {
+  Default as UserButtonDefault,
+  meta as userButtonMeta,
+  MultipleSessions as UserButtonMultipleAccounts,
+  Personal as UserButtonPersonal,
+} from '../stories/user-button.stories';
```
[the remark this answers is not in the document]
alexcarpenter: Not a bug. The MDX passes the module namespace straight to `<Story>` (`import * as UserButtonStories from './user-button.stories'`), and `StoryEmbed` reads `__source` off that prop (`StoryEmbed.tsx:58`), so the code footers do render. `registry.ts` only ever supplies `.meta` (`DocsViewer.tsx:87`) and no module in it sets `__source`.

packages/swingset/src/stories/user-button.stories.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+// Exposes this file's own source (via the `?raw` webpack rule) so each `<Story>` example
+// renders a code footer with its function's source. See `StoryModule.__source`.
+export { default as __source } from './account-button.stories?raw';
+
+export const meta: StoryMeta = {
+  group: 'Components',
+  title: 'AccountButton',
+  source: 'packages/ui/src/mosaic/components/account-button.view.tsx',
+};
```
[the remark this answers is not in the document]
alexcarpenter: Not applicable. `UserButton` is compound and has no flat CVA recipe, so it follows archetype A-compound: no `meta.styles`, no `<Preview>`, no `<PropTable>`. There are no knobs for the stories to spread. The suggested `accountButtonRecipe` does not exist.

packages/headless/src/utils/freeze.tsx
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Holds its subtree's DOM at the last committed frame while `frozen`. Renders keep
+ * happening, they just don't reach the DOM; the pending one commits when `frozen` flips
+ * back to `false`.
+ *
+ * Use it to stop content from visibly changing under an exit animation — a popover that
+ * closes because the thing it was showing changed would otherwise swap its contents on the
+ * way out.
+ */
+export function Freeze({ frozen, children }: FreezeProps) {
```
maxyinger, line 31: Just had this idea, so might be a little malformed, but I wonder if there's a simpler version of this that just uses view transitions for the exit 🤔 ... would need to alter the root transition logic that's holding this in the dom, but kinda opts into the exact behavior we're replicating here with snapshotting the dom.

would put everything exiting on a top layer, so that might have ramifications... but might have legs 🤷 

anyway, lets stick with this for now to not block this, but just wanted to mention as something that we can play with later

packages/ui/src/mosaic/components/avatar/avatar.tsx
```diff
@@ -88,6 +98,9 @@ const AvatarImage = React.forwardRef<HTMLImageElement, AvatarImageProps>(functio
       ref={ref}
       src={src}
       alt={alt}
+      // An avatar is an identity mark, not content to pull out of the page — dragging one
+      // only ever produces a stray ghost image mid-interaction.
+      draggable={false}
```
maxyinger, line 104: nice

packages/ui/src/mosaic/components/avatar/avatar.tsx
```diff
@@ -61,18 +61,28 @@ const AvatarImage = React.forwardRef<HTMLImageElement, AvatarImageProps>(functio
...
     const image = new window.Image();
+    image.src = src;
+
+    // An image the browser already holds is complete the moment it is asked for. Resolving it here
+    // rather than off an event keeps a remount (a row changing shape) or a swap between two avatars
+    // already on screen from dropping to the initials and back for a frame.
+    if (image.complete) {
+      onStatusChange('loaded');
+      return;
+    }
```
maxyinger, line 81: I know we're doing this in a layout effect, but will it still have a tick where this component returns null no matter what even when the image is already loaded because of the state update?

I guess if we wrap in like a `flsuhSync` it might mediate the issue without having to use a ref or something

packages/ui/src/mosaic/components/badge/badge.styles.ts
```diff
@@ -27,7 +33,7 @@ export const colors = stylex.create({
     color: colorVars['--cl-color-primary-foreground'],
   },
   neutral: {
-    backgroundColor: colorVars['--cl-color-neutral'],
+    backgroundColor: neutralScrim,
```
maxyinger, line 36: not needed for this pr, but used some neutral colors in the button variants. might make sense to hoist these up at some point.

packages/ui/src/mosaic/components/item/item.styles.ts
```diff
@@ -22,16 +27,37 @@ export const item = stylex.create({
...
+        default: null,
+        ':hover': colorVars['--cl-color-neutral'],
+      },
+    },
     cursor: 'pointer',
+    // A row that is standing down while another action runs keeps its place and its look, but
+    // stops answering the pointer — one declaration takes the cursor and the hover states with it.
+    pointerEvents: {
+      default: null,
+      ':disabled': 'none',
```
maxyinger: this kind of has some different behaviors in various parts we've written. maybe I'm just not picking up on more context here, but feels like something we should align on.

I left pointer events on disabled on the button and did `cursor: not-allowed;`. I'm ok with either that or this, but just feels like something that should be consistent on.

packages/ui/src/mosaic/components/item/item.styles.ts
```diff
@@ -87,12 +113,6 @@ export const description = stylex.create({
 
 export const label = stylex.create({
   base: {
-    // keyed to the row's hover, not the text's, so pointing anywhere in the row promotes the label
-    color: {
-      default: null,
-      [stylex.when.ancestor(':not(:hover)', itemScope)]: colorVars['--cl-color-neutral-faded'],
-      [stylex.when.ancestor(':hover', itemScope)]: colorVars['--cl-color-neutral'],
-    },
```
maxyinger, line 95: nice to remove these when calls 😌

packages/ui/src/mosaic/components/popover/popover.tsx
```diff
@@ -25,10 +25,14 @@ export type PopoverTitleProps = MosaicComponentProps<'h2'>;
...
 /** The anchor. Renders a `<button>`; `render` swaps in another element. */
-const Trigger = React.forwardRef<HTMLButtonElement, PopoverTriggerProps>(function PopoverTrigger(props, ref) {
+const Trigger = React.forwardRef<HTMLButtonElement, PopoverTriggerProps>(function PopoverTrigger(
+  { className, style, ...props },
+  ref,
+) {
   return (
     <Primitive.Trigger
       ref={ref}
+      {...mergeStyleProps(themeProps('popover-trigger'), className, style)}
```
maxyinger, line 35: gah. ty

packages/ui/src/mosaic/user-button/user-button.styles.ts
```diff
@@ -0,0 +1,88 @@
...
+    borderStyle: 'none',
+    outline: {
+      default: 'none',
+      ':focus-visible': `2px solid ${colorVars['--cl-color-primary']}`,
+    },
+    backgroundColor: 'transparent',
+    cursor: 'pointer',
+    display: 'inline-flex',
+    outlineOffset: '2px',
+  },
```
maxyinger, line 18: what was your thinking on excluding hover and popover open styles here?

packages/ui/src/mosaic/user-button/user-button.styles.ts
```diff
@@ -0,0 +1,88 @@
...
+    maxWidth: '12rem',
+  },
+
+  // The workspace list scrolls; the header and footer stay put. The scroll area carries the
+  // overflow itself, along with the edge fades and the scrollbar, so only the cap and the scroll
+  // padding are ours. The padding answers those fades: tabbing to a row below the fold would
+  // otherwise land it flush against the edge the mask fades out, on the one row you just moved to.
+  scroll: {
+    scrollPaddingBlockEnd: scrollFadeVars['--cl-scroll-fade-size'],
+    scrollPaddingBlockStart: scrollFadeVars['--cl-scroll-fade-size'],
```
maxyinger: ahh good catch... almost wonder if these should just be part of the scroll area classes. does it feel too opinionated to be in there?

packages/ui/src/mosaic/tokens.stylex.ts
```diff
@@ -59,18 +59,15 @@ export const colorVars = stylex.defineVars(colorDefaults);
...
 const radiusDefaults = {
   '--cl-radius-none': '0rem',
-  '--cl-radius-inner': '0.25rem',
-  '--cl-radius-control': '0.375rem',
-  '--cl-radius-element': '0.5rem',
-  '--cl-radius-container': '0.75rem',
+  '--cl-radius-sm': '0.25rem',
+  '--cl-radius-md': '0.375rem',
+  '--cl-radius-lg': '0.5rem',
+  '--cl-radius-xl': '0.75rem',
```
maxyinger, line 70: tyty

packages/ui/src/mosaic/user-button/user-button.layout.ts
```diff
@@ -0,0 +1,96 @@
...
+    accountRow: false,
+    signOutInHeader: true,
+    listsOrganizations: false,
+    listsAccounts: true,
+    accountsHeading: false,
+    createOrganizationInFooter: false,
+    addAccountInFooter: true,
+    signOutAllInFooter: true,
+  },
+} as const satisfies Record<UserButtonMode, ModeSlots>;
```
maxyinger: I was chatting with claude about how to make this more ergonomic, it mentioned using this shape instead:

```
type Slot = 'header' | 'account-row' | 'accounts-heading' | 'footer' | 'none';

const placement = {
  signOut:            { combined: 'account-row',      organization: 'none',   user: 'header' },
  createOrganization: { combined: 'account-row',      organization: 'footer', user: 'none'   },
  addAccount:         { combined: 'accounts-heading', organization: 'none',   user: 'footer' },
  signOutAll:         { combined: 'footer',           organization: 'none',   user: 'footer' },
  inviteMembers:      { combined: 'header',           organization: 'header', user: 'none'   },
} as const satisfies Record<string, Record<UserButtonMode, Slot>>;
```

Does this feel any better to you? or nah? reads a little easier to me. Still trying to wrap my head around the logic here tho, so ignore if this makes things worse

maxyinger: idt this was typechecked but this is the sample it came up with:

```ts
import type { UserButtonData, UserButtonMode, UserButtonModePriority } from './user-button.types';

/**
 * The containers a surface has. The data only ever takes one away — an empty list is no list — so
 * this is the half of the layout that mode and data both have a say in.
 */
interface Regions {
  /** The active account's own row, at the head of the workspace list. */
  accountRow: boolean;
  /** The workspace list, and something to put in it. */
  organizationList: boolean;
  /**
   * The Accounts group. `headed` carries a heading, so it tells its rows apart and has somewhere to
   * hang "Add account"; `bare` is just a list of somewhere else to go.
   */
  accountList: 'headed' | 'bare' | 'none';
}

const modeRegions = {
  combined: { accountRow: true, organizationList: true, accountList: 'headed' },
  // Not a surface about the account: it neither lists other accounts nor carries the row that holds
  // the account-wide actions.
  organization: { accountRow: false, organizationList: true, accountList: 'none' },
  // Nothing to tell the account rows apart from, so they stand unheaded.
  user: { accountRow: false, organizationList: false, accountList: 'bare' },
} as const satisfies Record<UserButtonMode, Regions>;

/** Where an affordance can land. `none` is the surface not offering it at all. */
type Slot = 'header' | 'account-row' | 'accounts-heading' | 'footer' | 'none';

/**
 * Which slot each movable affordance takes, per mode. Pure mode: whether the data leaves that slot's
 * region standing is `resolveUserButtonLayout`'s to settle, and all it can do about it is move the
 * affordance somewhere that does stand.
 */
const placement = {
  // Every other surface hangs it off the account's own row. An account-only one has no such row, so
  // it takes the labelled header slot **Invite** occupies elsewhere.
  signOut: { combined: 'account-row', organization: 'none', user: 'header' },
  // Shares the footer slot with "Add account" wherever it reaches the footer: whichever of the two
  // the surface has nowhere else to put.
  createOrganization: { combined: 'account-row', organization: 'footer', user: 'none' },
  addAccount: { combined: 'accounts-heading', organization: 'none', user: 'footer' },
  signOutAll: { combined: 'footer', organization: 'none', user: 'footer' },
  inviteMembers: { combined: 'header', organization: 'header', user: 'none' },
} as const satisfies Record<string, Record<UserButtonMode, Slot>>;

export type UserButtonAffordance = keyof typeof placement;

export interface UserButtonLayout extends Regions {
  /** Which workspace the trigger names and the header leads with. */
  leadWith: 'organization' | 'user';
  /** Where each affordance landed, once the absent regions are accounted for. */
  slots: Record<UserButtonAffordance, Slot>;
}

export function resolveUserButtonLayout(
  mode: UserButtonMode,
  modePriority: UserButtonModePriority,
  data: UserButtonData,
): UserButtonLayout {
  const regions = modeRegions[mode];

  // A pending invitation or suggestion counts: it has to be reachable before there is a membership.
  // Loading does not count, so an account with none never opens a section that then disappears
  // under it.
  const organizationList =
    regions.organizationList && (data.hasOrganizations || data.suggestions.length > 0 || data.invitations.length > 0);
  const accountList = data.additionalSessions.length > 0 ? regions.accountList : 'none';

  // A slot has to exist before anything can land in it. The Accounts heading is the only one the
  // data can take away, and the footer is where its occupant goes when it does.
  const demote = (slot: Slot): Slot => (slot === 'accounts-heading' && accountList !== 'headed' ? 'footer' : slot);

  return {
    // Only a combined surface has two things to choose between; the other two are what they are.
    leadWith: mode === 'combined' ? modePriority : mode,
    accountRow: regions.accountRow,
    organizationList,
    accountList,
    slots: slots: {
      signOut: demote(placement.signOut[mode]),
      createOrganization: demote(placement.createOrganization[mode]),
      addAccount: demote(placement.addAccount[mode]),
      signOutAll: demote(placement.signOutAll[mode]),
      inviteMembers: demote(placement.inviteMembers[mode]),
    },
  };
}
```

maxyinger (commented):
lgtm. just posted some initial thoughts. custom actions api makes sense to me with all the variations there

One fun thing with the view separation is potentially being able to import the view on the marketing site or docs. can run a postcss step to convert psuedo selectors like `:hover` to `data-hover` or something like that and be able to control everything with js

packages/headless/src/utils/freeze.tsx
maxyinger: would make it tough to alter exit animations from css as a user consuming this tho 🤔

packages/ui/src/mosaic/components/avatar/avatar.tsx
alexcarpenter: One extra render. Not painted. Layout effects run in the commit, before paint. React flushes their state update in the same commit.

`flushSync` here warns and adds nothing: `flushSync was called from inside a lifecycle method.`

Your question found a gap. The old cached test used Testing Library `render`. That wraps in `act`, which drains passive effects. It passed with `useEffect` too. It pinned nothing.

New test drives the root directly. `flushSync` stops before passive effects. Fails on `useEffect`. Passes on `useLayoutEffect`.

packages/ui/src/mosaic/components/badge/badge.styles.ts
alexcarpenter: yeah, will follow up with a fix. 👍🏼

packages/ui/src/mosaic/components/item/item.styles.ts
alexcarpenter: good catch, moved Item to cursor: not-allowed

packages/ui/src/mosaic/user-button/user-button.styles.ts
alexcarpenter: I am open to adding some hover interaction. What do you think it should be?

maxyinger: hrmph. I feel more strongly about having a sticky open state like the ring in this pic. for hover I feel like it should either be a lighter ring or excluded entirely. maybe now that I think about it, I might lean more towards excluding a hover state. idk. wdyt?

Screenshot 2026-08-10 at 11 53 27 AM

packages/ui/src/mosaic/user-button/user-button.styles.ts
alexcarpenter: good call moved to the scroll area. wasnt sure if thatd support inline scrolling or if it was only block.

maxyinger: yea rn block only. happy to add support for inline if we want, but rn figured all we needed was block, so didn't want to add the complexity

packages/ui/src/mosaic/user-button/user-button.layout.ts
alexcarpenter: updated https://github.com/clerk/javascript/pull/9184/commits/c68d0930448e7450f01f41d9b9164a2421952e33

packages/headless/src/utils/freeze.tsx
alexcarpenter: if you want to take a stab in a follow up PR id be open to it.

maxyinger: Yea maybe I'll give it a shot after I wrap up dialog work and we can see how it feels. I'm a little skeptical, but also curious

maxyinger (approved):

packages/ui/src/mosaic/user-button/user-button.layout.ts
Ephem, line 1: This file was hard to read, it's hard to get a sense for what the layout looks like or how it works. That reflects elsewhere as well, mainly parts of the `.view.` file.

I think part of it is some of the naming, part of it is some of the structure, I'll leave a few separate comments below so we can keep discussions separate.

This is subjective of course so I wanted to leave these comments first to discuss, but I'll also work on a more concrete suggestion we can look at together.

packages/ui/src/mosaic/user-button/user-button.layout.ts
```diff
@@ -0,0 +1,98 @@
+import type { UserButtonData, UserButtonMode, UserButtonModePriority } from './user-button.types';
+
+/** Where an affordance lands. `none` is a mode not offering it at all. */
+type Slot = 'account-row' | 'accounts-heading' | 'footer' | 'header' | 'none';
```
Ephem, line 4: Reading this at first confused me since I thought this was a complete list of slots/sections, but they don't list all the slots, only the ones that are possible targets for affordances (shoutout to "The Design of Everyday Things").

Affordance feels weird here though, "action" is used in the `.view.` file and feels clearer, so maybe `ActionSlots` or something would make it clear what this actually is?

packages/ui/src/mosaic/user-button/user-button.layout.ts
```diff
@@ -0,0 +1,98 @@
...
+interface Sections {
+  /** The workspace list carries the active account's row, and the account-wide actions with it. */
+  accountRow: boolean;
+  /** There is a workspace list at all, and something to put in it. */
+  listsOrganizations: boolean;
+  /** There is an Accounts group, and another account to switch to. */
+  listsAccounts: boolean;
+  /** The Accounts group is headed, so it tells its rows apart and carries "Add account". */
+  accountsHeading: boolean;
+}
```
Ephem, line 16: Ignoring the `account` naming and `accountRow` confusion mentioned in other comments, the naming here also feels disjoint given they are all booleans. Some read as nouns, some as verbs/adjective-like.

If it's sections, `listsOrganizations` and `listsAccounts` feels off. What I'd argue is NOT treating them as sections though, given that the headers don't feel like "sections" and that these leak out into the `.view.` code without preserving the section context. My suggestion is to instead name them as booleans, like `showOrganizationsSection`.

packages/ui/src/mosaic/user-button/user-button.layout.ts
```diff
@@ -0,0 +1,98 @@
+import type { UserButtonData, UserButtonMode, UserButtonModePriority } from './user-button.types';
+
+/** Where an affordance lands. `none` is a mode not offering it at all. */
+type Slot = 'account-row' | 'accounts-heading' | 'footer' | 'header' | 'none';
+
+/** Which sections a mode opens. Whether the data fills them is settled in `resolveUserButtonLayout`. */
+interface Sections {
+  /** The workspace list carries the active account's row, and the account-wide actions with it. */
+  accountRow: boolean;
```
Ephem, line 9: I find `accounts-row` and `accounts-heading` to be confusing, given that `accounts-row` is actually `organization-heading` structurally and visually? 

Semantically, it's the `activeAccount`, or as I'd argue, `activeSession`, see next point.

I find the accounts naming confusing. It's a great word to use in user facing text, but we already have `users` and `sessions` in code, introducing a third `accounts` feels hard to reason about (I know we've used it in a limited fashion in other places before, but not as much as it's embraced here).

packages/ui/src/mosaic/user-button/user-button.layout.ts
```diff
@@ -0,0 +1,98 @@
...
+  /** There is a workspace list at all, and something to put in it. */
+  listsOrganizations: boolean;
```
Ephem, line 11: Workspace is something we introduced in the dashboard to disambiguate from organizations. It might be a good word to use in translations/user facing text, but I find it confusing to introduce in the code when we already use organizations everywhere else?

The `.view.` is what uses it in code but I thought I'd comment here to keep naming discussions in one place.

packages/ui/src/mosaic/user-button/user-button.layout.ts
```diff
@@ -0,0 +1,98 @@
...
+  // Nothing to tell the account rows apart from, so they stand unheaded.
+  user: { accountRow: false, listsOrganizations: false, listsAccounts: true, accountsHeading: false },
+} as const satisfies Record<UserButtonMode, Sections>;
+
+/**
+ * The slots a mode can name: the two every mode has, plus whichever sections it opens above. A mode
+ * cannot put an affordance in a section it does not render, so the table below cannot drift out of
+ * step with `sections` without failing to compile.
+ */
+type ModeSlot<M extends UserButtonMode> =
```
Ephem, line 31: Type safety is nice, but it makes this file pretty hard to read compared to something like:

```tsx
type Action = 'inviteMembers' | 'signOut' | 'createOrganization' | 'addAccount' | 'signOutAll';

interface ModeLayout {
  sections: Sections;
  placements: Record<Action, Slot>;
}

const modeLayouts = {
  combined: {
    sections: { accountRow: true, listsOrganizations: true, listsAccounts: true, accountsHeading: true },
    placements: {
      inviteMembers: 'header',
      signOut: 'account-row',
      createOrganization: 'account-row',
      addAccount: 'accounts-heading',
      signOutAll: 'footer',
    },
  },
  organization: {
    sections: { accountRow: false, listsOrganizations: true, listsAccounts: false, accountsHeading: false },
    placements: {
      inviteMembers: 'header',
      signOut: 'none',
      createOrganization: 'footer',
      addAccount: 'none',
      signOutAll: 'none',
    },
  },
  user: {
    sections: { accountRow: false, listsOrganizations: false, listsAccounts: true, accountsHeading: false },
    placements: {
      inviteMembers: 'none',
      signOut: 'header',
      createOrganization: 'none',
      addAccount: 'footer',
      signOutAll: 'footer',
    },
  },
} as const satisfies Record<UserButtonMode, ModeLayout>;
```

Not sure where I stand on that tradeoff but wanted to note that it comes at a cost.

Ephem (commented):
Really nice work here, this is really great!

I did stumble when trying to understand the layout and how different sections render etc. Maybe I'm just unfamiliar and some parts takes some getting used to, and I don't have fully formed opinions yet, but I left a few comments as discussion points below.

packages/ui/src/mosaic/user-button/user-button.view.tsx
```diff
@@ -0,0 +1,1025 @@
...
+
+/**
+ * What the surface leads with: named in the trigger and headed in the popup, so the two always
+ * agree. Only an organization-led surface with an organization actually active resolves to one.
+ */
+function leadWorkspace({ layout, activeOrganization, activeSession }: UserButtonContextValue): ActiveWorkspace {
+  const organization = layout.leadWith === 'organization' ? activeOrganization : null;
+  return organization
+    ? { name: organization.name, imageUrl: organization.imageUrl, shape: 'square', organization }
+    : { name: activeSession.name, imageUrl: activeSession.imageUrl, shape: 'circle' };
```
Ephem, line 104: @alexcarpenter Say mode is organization, organizations are optional and the user has none, and hidePersonal is true.

The existing organizationSwitcher would say "No organization selected" in that case, but this falls back to the user, which might feel off in organization mode?

PR #9185, opened 2026-07-16
feat(ui): add UserButton controller

alexcarpenter:
## Description

> The PR https://github.com/clerk/javascript/pull/9191 was folded into this one for a final review. See that PR description for additional details.

Stacked on #9184. Connects the `UserButton` view to live Clerk data.

`useUserButtonController()` returns a `'loading' | 'hidden' | 'ready'` union. When it is `ready` it carries the view's data contract and the callback behind every row. `user-button.tsx` is the connected container that owns the popover.

Where the data comes from:

- `activeSession` from `useUser()` and `useSession()`. The name follows Clerk's own order: first and last name, then username, then the identifier.
- `activeOrganization` from `useOrganization()`, where `null` is the personal workspace.
- `memberships`, `suggestions`, and `invitations` from `useOrganizationList()`, paged as the list scrolls.
- `hasOrganizations` from the user resource, so it can answer before those lists load.
- `additionalSessions` from the client, without the active one.
- The membership request count only on the active organization row, and only with `org:sys_memberships:manage`.

What the rows do:

- Picking a workspace or an account calls `setActive`. Signing out calls `signOut`. Invitations and suggestions accept in place and revalidate the list.
- Manage account, manage organization, create organization, and invite members open Clerk's own modals, portalled out of the popover so they outlive it. Passing a URL routes to a page of your own instead, and that is the whole opt-in: `userProfileUrl`, `organizationProfileUrl`, `createOrganizationUrl`. `afterSelectOrganizationUrl` and `afterSelectPersonalUrl` say where picking a workspace lands.
- Create organization is withheld from a user without the permission to create one. The personal row is withheld where an organization is required, and `hidePersonal` withholds it by request.
- `customMenuItems` reach the menu through the container, and a custom action closes the popover behind whatever it opens.

The button renders nothing until Clerk answers. While it loads, a signed-out visitor and a session still resolving are indistinguishable, so anything rendered then is a button promised to people who are never going to get one. `<ClerkLoading>` is where an app that knows its own nav puts a placeholder.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/user-button/user-button.tsx
```diff
@@ -0,0 +1,155 @@
...
+  // the immediate `pendingKey`; only the view's feedback is delayed.
+  const displayPendingKey = useSpinDelay(pendingKey);
+
+  // Nothing stands in for the button until Clerk answers: while it is loading, a signed-out visitor
+  // is indistinguishable from a session still resolving, so anything rendered here is a button
+  // promised to people who are never going to get one. `<ClerkLoading>` is where an app that knows
+  // its own nav puts a placeholder.
+  if (controller.status !== 'ready') {
+    return null;
+  }
```
Ephem: Would we expose `<Avatar.Fallback>` so it's easy to get a matching loading state immediately if you want that? Or put as a more general question, what is the story for people who do want a loading state immediately?

Note that `<ClerkLoading>` is already bundled, and since this is too, no matter where you put that loading state, I think the loading state works for SSR too out of the box! 🤯

packages/ui/src/mosaic/user-button/user-button.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * Everything `<UserButton />` takes: where its profile surfaces open (`UserButtonControllerOptions`),
+ * what the trigger shows (`UserButtonTriggerProps`), and the app's own rows at the foot of the menu
+ * (`UserButtonMenuProps`).
+ */
+export type UserButtonProps = UserButtonControllerOptions &
+  UserButtonTriggerProps &
+  UserButtonMenuProps &
+  Pick<UserButtonModeProps, 'modePriority'>;
```
Ephem: Something that popped to mind when pondering the public API is the previous discussions about providing instrumentation callbacks to let users log their own events etc when actions are taken within the button.

Not saying that's something we need to include in a first version, just thought I'd mention it.

packages/ui/src/mosaic/user-button/user-button.controller.tsx
```diff
@@ -0,0 +1,299 @@
...
+      UserButtonAsyncCallbacks &
+      UserButtonBrandingProps & {
+        status: 'ready';
+        /** Whether the instance has organizations turned on at all. False forces the button to `user` mode. */
+        organizationsEnabled: boolean;
+      });
+
+// Mirrors the `<OrganizationSwitcher>` `afterSelectOrganizationUrl` prop: a full URL/path, a `:token`
+// path template resolved against the organization, or a builder function.
+type AfterSelectUrl<T> = ((entity: T) => string) | string;
```
Ephem: It's been a source of confusion to me in the past that some url props support builder functions, and some do not. I now get that the AfterSelect ones _needs it_ since the data you want to build from is not present before the click, but that form could be useful for `userProfileUrl` and `organizationProfileUrl` too for pure DX gains (not having to grab the current user/org first), so maybe this is an opportunity to streamline since you've built such a nice abstraction around it?

An total aside of course, not sure the consistency gains are worth it, feel free to ignore, just thought I'd mention it. 😄

packages/ui/src/mosaic/user-button/user-button.controller.tsx
```diff
@@ -0,0 +1,299 @@
...
+  const createOrganization = openOrNavigate({
+    url: options?.createOrganizationUrl,
+    mode: options?.createOrganizationMode,
+    openModal: () => clerk.openCreateOrganization({ getContainer }),
+    buildUrl: () => clerk.buildCreateOrganizationUrl(),
+    navigate: router.navigate,
+  });
+
+  // Organizations, single-session and forced selection all come off the environment, and it
+  // hydrates on its own schedule. Waiting for it beats guessing at three answers and rearranging.
```
Ephem: I'm guessing this is an LLM. I find so many of the comments in these PRs 1. Unnecessary and 2. Very hard to read. I'm honestly having a harder time parsing some of the comments than I have reading the code. 😬

If we even need a comment:

```suggestion
  // These effect layout, so we wait for all to avoid UI shifts
```

packages/ui/src/mosaic/user-button/user-button.controller.tsx
```diff
@@ -0,0 +1,299 @@
...
+    identifier: getIdentifier(user),
+    imageUrl: user.imageUrl,
+  };
+}
+
+export function useUserButtonController(options?: UserButtonControllerOptions): UserButtonController {
+  const { isLoaded: isUserLoaded, user } = useUser();
+  const { isLoaded: isSessionLoaded, session } = useSession();
+  const { isLoaded: isOrgLoaded, organization } = useOrganization();
+  const { userMemberships, userInvitations, userSuggestions, ref } = useOrganizationListInView();
```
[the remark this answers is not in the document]
Ephem: @alexcarpenter I think this coderabbit feedback is correct, ~~also for `useOrganization`~~. Currently, we'll make these calls for _all_ instances, including the ones that don't have orgs enabled, which seems wasteful. I think calling these will also trigger the prompt to enable organizations in dev?

The downside is, we'll now have a request waterfall where we first have to wait to check if orgs are enabled which is not great. I also don't think we have a good way to enable/disable these conditionally right now.

This looks like a key piece to figure out, let's chat!

Update: I realized `useOrganization` without extra args does _not_ result in an extra API call, it reads from `clerk`, so that one should be fine if we can fix the enable orgs prompt.

packages/ui/src/mosaic/user-button/user-button.controller.tsx
```diff
@@ -0,0 +1,299 @@
...
+    invitations,
+    additionalSessions,
+    paging: {
+      ref,
+      hasMore: Boolean(userMemberships.hasNextPage || userInvitations.hasNextPage || userSuggestions.hasNextPage),
+    },
+    onSelectOrganization: organizationId =>
+      clerk.setActive({ organization: organizationId, redirectUrl: afterSelectUrl(organizationId) }),
+    onSwitchSession: sessionId =>
+      clerk.setActive({ session: sessionId, redirectUrl: displayConfig.afterSwitchSessionUrl }),
```
Ephem: I think this needs to handle session tasks as well? Current implementation lives in `handleSessionClicked` in [useMultisessionActions](https://github.com/clerk/javascript/blob/main/packages/ui/src/components/UserButton/useMultisessionActions.tsx).

packages/ui/src/mosaic/user-button/user-button.controller.tsx
```diff
@@ -0,0 +1,299 @@
...
+  const { isLoaded: isSessionLoaded, session } = useSession();
+  const { isLoaded: isOrgLoaded, organization } = useOrganization();
+  const { userMemberships, userInvitations, userSuggestions, ref } = useOrganizationListInView();
+
+  const clerk = useClerk();
+  const router = useMosaicRouter();
+  // An app can mount the button inside its own dialog or popover; the modal has to portal into that
+  // same root or it renders behind the surface that opened it.
+  const getContainer = usePortalRoot();
+  const environment = useMosaicEnvironment();
```
Ephem: Not related to this PR, but this not being reactive feels a bit iffy long term. The hook has a bunch of good comments with constraints that are not obvious in this location where you are actually consuming it.

This is more a general problem with the `clerk-js`->"host app" bridge though and not worth tackling right now.

packages/ui/src/mosaic/user-button/user-button.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const runAction = <Args extends unknown[]>(
+    keyFor: (...args: Args) => string,
+    fn: ((...args: Args) => void | Promise<unknown>) | undefined,
+    closeOnSuccess = false,
+  ) =>
+    fn
+      ? (...args: Args) => {
+          if (pendingKey) {
+            return;
+          }
```
Ephem: This makes the UI feel frozen when an action is in progress, especially on throttled mobile where there's no cursor to indicate things. I saw this changed in the next PR though so will re-review there, just leaving this comment to not forget. 😄

Ephem (commented):
Feels great overall and this one was easy to follow. Left a few comments and will review the machine PR next to get a feel for the entire thing.

packages/ui/src/mosaic/user-button/user-button.tsx
alexcarpenter: Likely want to expose the fallback prop like we currently do on components?

packages/ui/src/mosaic/user-button/user-button.tsx
alexcarpenter: yeah, thats a neat idea!

packages/ui/src/mosaic/user-button/user-button.controller.tsx
alexcarpenter: Resolved here https://github.com/clerk/javascript/pull/9185/commits/44895824f0

packages/ui/src/mosaic/user-button/user-button.controller.tsx
alexcarpenter: Resolved here https://github.com/clerk/javascript/pull/9185/commits/80a1968c09

Two differences from `handleSessionClicked`: it calls `decorateUrl` (the old `redirectUrl` path decorated for us, and clerk-js warns in dev when the callback does not), and it skips the virtual-router branch, since Mosaic never mounts inside one.

packages/ui/src/mosaic/user-button/user-button.tsx
alexcarpenter: Added here https://github.com/clerk/javascript/pull/9185/commits/809fb4242e

Matches the `fallback` prop the existing components carry, so it stands in while Clerk is still answering and is dropped once nobody is signed in (signed out is an answer, not a wait). Rendering a matching placeholder is left to the app.

Exposing `Avatar.Fallback` / `Skeleton` so the placeholder lines up without hand-sizing is a separate export-surface question, so I left it for #9338.

packages/ui/src/mosaic/user-button/user-button.controller.tsx
```diff
@@ -0,0 +1,206 @@
...
+  // Every action here is a network round trip, so we can start the
+  // pending state immediately, we use this for the minDuration
+  const displayPendingKey = useSpinDelay(context.pendingKey, {
+    delay: 0,
+    minDuration: context.closeOnSuccess ? 0 : undefined,
+  });
+
+  if (model.status !== 'ready') {
+    return { status: model.status };
+  }
```
Ephem: Since this happens before respecting the frozen state I think this is going to revert back to `'loading'` during the transitive state when a navigation is involved after `setActive`.

That means we'll pop the fallback/null during that window.

Reverting to `'loading'` also isn't covered by tests currently.

Image

packages/ui/src/mosaic/user-button/user-button.controller.tsx
```diff
@@ -0,0 +1,206 @@
...
+    modePriority,
+    customMenuItems: menuItems,
+    menuItemOrder,
+    open: value !== 'closed',
+    onOpenChange: next => send(next ? { type: 'OPEN' } : { type: 'CLOSE' }),
+    pendingKey: displayPendingKey,
+    onSelectOrganization: runAction(userButtonBusyKeys.selectOrganization, onSelectOrganization, true),
+    onSwitchSession: runAction(userButtonBusyKeys.switchSession, onSwitchSession),
+    onSignOutSession: runAction(userButtonBusyKeys.signOutSession, onSignOutSession),
+    onSignOutAll: runAction(userButtonBusyKeys.signOutAll, onSignOutAll),
```
Ephem: I think `onSignOutAll` should have `closeOnSuccess: true`, and `onSignOutSession` should have  `data.additionalSessions.length === 0`.

Otherwise, the state will still be 'open', but the component unmounts. If the user signs in again while `UserButton` is still mounted, the popup will open again after the sign in which feels off.

packages/ui/src/mosaic/user-button/user-button.model.tsx
```diff
@@ -0,0 +1,303 @@
...
+    }
+    return [toSession(s.id, sessionUser)];
+  });
+
+  const afterSelectUrl = (organizationId: string | null): string | undefined => {
+    if (!organizationId) {
+      return resolveAfterSelectUrl(options?.afterSelectPersonalUrl, user);
+    }
+    const selected = membershipData.find(m => m.organization.id === organizationId)?.organization;
+    return selected ? resolveAfterSelectUrl(options?.afterSelectOrganizationUrl, selected) : undefined;
```
Ephem, line 237: This is called via `onSelectOrganization` and one of the callsites for that is when switching to an recently accepted invitation. When it's an invitation, there's no guarantee the organization is in `membershipData` by the time this gets called, resulting in an undefined url and no navigation.

This can happen on fast clicks on slow networks because we don't await the revalidates further down, see separate comment.

packages/ui/src/mosaic/user-button/user-button.model.tsx
```diff
@@ -0,0 +1,303 @@
...
+    onAcceptSuggestion: suggestionId => {
+      const suggestion = suggestionData.find(s => s.id === suggestionId);
+      return Promise.resolve(suggestion?.accept()).finally(() => void userSuggestions.revalidate?.());
+    },
+    // Accepting joins the organization, so memberships are stale too. A suggestion joins nothing.
+    onAcceptInvitation: invitationId => {
+      const invitation = invitationData.find(i => i.id === invitationId);
+      return Promise.resolve(invitation?.accept()).finally(() => {
+        void userInvitations.revalidate?.();
+        void userMemberships.revalidate?.();
```
Ephem: These should not be voided, instead return a `Promise.all(...)` from the finally so we wait for the revalidate. This fixes a bug mentioned above, but it also fixes a case where the UI goes back to showing "Join" again right after already having joined, and the possibility of trying to accept invitations twice.

Or put differently, by not waiting for the revalidation, we are not getting the benefits of `frozen`.

Not sure if we also want to add a separate catch with a noop for the revalidations, otherwise the action will be treated as failed just because revalidations failed. 🤔 Both versions has tradeoffs.

packages/ui/src/mosaic/user-button/user-button.controller.tsx
```diff
@@ -0,0 +1,206 @@
...
+            closeOnSuccess: event.closeOnSuccess,
+          })),
+        },
+      },
+    },
+    // Reached only from `open`, so a busy popup that is not open is unrepresentable, and RUN going
+    // unhandled here is what stops a second action starting while one is in flight. Dismissing the
+    // popup abandons the action: the request finishes, but nothing is left for its result to land in.
+    busy: {
+      on: { CLOSE: { target: 'closed', actions: assign(() => settled) } },
```
Ephem: What if you reopen the popup while the action you were taking is still ongoing? We'd want to still show pending then right?

I think we might need to split it into two separate states?

```
popup: open | closed
action: idle | busy
```

packages/ui/src/mosaic/user-button/user-button.model.tsx
```diff
@@ -0,0 +1,303 @@
...
+      clerk.setActive({
+        session: sessionId,
+        navigate: async ({ session, decorateUrl }) => {
+          const task = session.currentTask;
+          if (task) {
+            await router.navigate(buildTaskUrl(task, { base: clerk.buildSignInUrl() }));
+            return;
+          }
+          // `redirectUrl` decorated for us; taking the callback takes the Safari ITP refresh with it.
+          await router.navigate(decorateUrl(displayConfig.afterSwitchSessionUrl));
```
Ephem: This tries to navigate even if `afterSwitchSessionUrl` is empty, which might navigate to the same page as we are on, old behavior in `useMultisessionActions` was to call it conditionally.

Also, are we missing a `afterSwitchSessionUrl` prop on the component? Is that intentional?

packages/ui/src/mosaic/user-button/user-button.model.tsx
```diff
@@ -0,0 +1,303 @@
...
+  CreateOrganizationMode & {
+    afterSelectOrganizationUrl?: AfterSelectUrl<OrganizationResource>;
+    /** Where selecting the personal workspace lands. Resolved against the user, not an organization. */
+    afterSelectPersonalUrl?: AfterSelectUrl<UserResource>;
+    /**
+     * Leaves the personal workspace out. An instance that forces organization selection withholds it
+     * either way, so this cannot opt back in.
+     */
+    hidePersonal?: boolean;
+  };
```
Ephem, line 71: I had AI go through and map props to the old button, these are the discrepancies it found. Some of this might be intentional, or saved for follow-ups but it's worth being explicit about so I thought I'd post the list, maybe you can shed some light on which are missing intentionally and not?

### Likely worth adding

- **`defaultOpen`**
  - Supported by `UserButtonRoot`, but not exposed by the connected `UserButton` or controller.
- **`signInUrl`**
  - Legacy uses the component override for “Add account” and session-task URLs.
  - Mosaic always uses `clerk.buildSignInUrl()`.
- **Additional `userProfileProps`**
  - Mosaic only accepts `customPages` and `pageOrder`.
  - Legacy also forwards:
    - `additionalOAuthScopes`
    - `apiKeysProps`
    - `appearance` for the opened UserProfile modal
- **Custom menu `open` / profile start path**
  - Legacy custom items can open a specific UserProfile page.
  - Mosaic items only support `href` or `onClick`.

Because Mosaic UserButton also replaces OrganizationSwitcher functionality, these are missing too:

- **`afterCreateOrganizationUrl`**
  - Controls where Clerk navigates after creating an organization.
  - This differs from `createOrganizationUrl`, which controls whether clicking Create opens Clerk’s modal or navigates to an app page.
- **`skipInvitationScreen`**
  - Configures the Clerk create-organization modal.
- **`afterLeaveOrganizationUrl`**
  - Used after leaving an organization through OrganizationProfile.
- **`organizationProfileProps`**
  - Legacy forwards custom pages and appearance into the opened OrganizationProfile.

packages/ui/src/mosaic/user-button/user-button.model.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+  return {
+    status: 'ready',
+    organizationsEnabled,
+    renderBranding: displayConfig.branded,
+    activeSession: toSession(session.id, user),
+    activeOrganization: organization ? toMembership(organization) : null,
+    // The user resource settles this before the paginated list answers; the count covers a stale resource.
+    hasOrganizations: user.organizationMemberships.length > 0 || (userMemberships.count ?? 0) > 0,
+    hidePersonal: forceOrganizationSelection || (options?.hidePersonal ?? false),
```
Ephem, line 248: There's a state I don't think is modelled here which is "No organization selected". I think this is when `forceOrganizationSelection` is true, but `organization` is `null`. Not sure when this happens (deleted orgs?), but I know the old UserButton handles it.

Requires changes in the view too, but this is within the PR diff and I don't think we are providing everything we need from here (`forceOrganizationSelection`).

Ephem (commented):
I did some AI-assisted reviewing and found a few things to dig into. I'll follow this up with stepping through everything myself as well.

Happy to tackle a few of these if you'd like after I've gotten through the full review.

packages/ui/src/mosaic/user-button/user-button.tsx
```diff
@@ -0,0 +1,150 @@
...
+export type UserButtonProps = UserButtonModelOptions &
+  UserButtonTriggerProps &
+  UserButtonMenuProps &
+  UserButtonModeProps & {
+    userProfileProps?: UserButtonUserProfileProps;
+    /**
+     * Stands in while Clerk is still answering, so the space the button will take is held rather
+     * than appearing under whatever is beside it. Dropped once nobody is signed in, since that is
+     * an answer and not a wait.
+     */
```
Ephem, line 39: This is a very AI-comment.

```suggestion
    /**
     * Fallback while loading.
     * 
     * Note that the UserButton renders nothing when the user is signed out, so using this on
     * pages that are reachable while both signed-out and signed-in can result in Fallback->Nothing.
     */
```

packages/ui/src/mosaic/user-button/user-button.model.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+  const afterSelectUrl = (organizationId: string | null): string | undefined => {
+    if (!organizationId) {
+      return resolveAfterSelectUrl(options?.afterSelectPersonalUrl, user);
+    }
+    const selected = membershipData.find(m => m.organization.id === organizationId)?.organization;
+    return selected ? resolveAfterSelectUrl(options?.afterSelectOrganizationUrl, selected) : undefined;
+  };
+
+  return {
```
Ephem, line 240: Reading this again, I still like the idea of segmenting this a bit:

```
{
  status: 'ready',
  data: { ... },
  mutations/actions: { ... },
}
```

Highly subjective, no real reason besides creating some structure, very much feel free to ignore/disagree.

packages/ui/src/mosaic/user-button/user-button.controller.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+interface UserButtonMachineContext {
+  /** Which action is currently pending. */
+  pendingKey: string | null;
+  /**
+   * The model the action started from. `setActive` swaps the active organization while its
+   * promise is still in flight, so the live model would rearrange the popup mid-action.
+   * The view renders this instead until the action settles.
+   */
+  frozen: UserButtonReadyModel | null;
```
Ephem: NIT: I still like something like `frozenModel` here just for some extra clarity.

packages/ui/src/mosaic/user-button/user-button.model.tsx
```diff
@@ -0,0 +1,303 @@
...
+          }
+          // `redirectUrl` decorated for us; taking the callback takes the Safari ITP refresh with it.
+          await router.navigate(decorateUrl(displayConfig.afterSwitchSessionUrl));
+        },
+      }),
+    onSignOutSession: sessionId =>
+      clerk.signOut({
+        sessionId,
+        // Other accounts stay signed in, so this is a single sign out rather than a full one.
+        redirectUrl:
```
Ephem: This seems to be using a different strategy than the old UserButton which used [clerk.redirectWithAuth](https://github.com/clerk/javascript/blob/23d22125124102bc45bf2d156b72ff02ee685d7b/packages/ui/src/contexts/components/UserButton.ts#L32), is that intentional?

packages/ui/src/mosaic/user-button/user-button.pages.tsx
```diff
@@ -0,0 +1,160 @@
...
+    // clerk-js decides what an item *is* from which callbacks are present, and drops one missing an
+    // icon pair as invalid. So the icon callbacks go out whether or not there is an icon to put
+    // through them; without them, leaving `icon` off would silently cost you the page.
+    const icon = bind(`icon:${id}`);
+    const content = isLink(item) ? undefined : bind(`content:${id}`);
+
+    return {
+      label: item.label,
+      // A page is routed to by its path; a link is followed to wherever it points.
+      url: isLink(item) ? item.href : item.path,
```
Ephem, line 147: `item.href` here could be something like `javascript:`, which can be dangerous. Previously we passed this through `navigate()` which rejects dangerous protocols, configurable by the option `allowedRedirectProtocols`.

This is provided by developers of the host app, so one might argue it's their own fault, and it's unlikely to happen etc, but since we have that option, I _think_ it should apply here too?

Ephem (commented):
Did a full manual pass and I found one security hardening thing and some very minor NITs.

This is starting to feel very good to me now. The last review uncovered some real issues, but most of it is easily fixed, and none of it is about the structure or approach itself, that feels really solid.

* We might want to update the Mosaic docs to match this new structure (can be follow-up)
* The portals _into_ the Clerk app and that whole thing feels a bit scary just because of the nature of portals and the whole thing. I don't have any notes there and we'll give it some proper testing after this PR is merged, just wanted to point it out as an area to give some proper tire kicking.

packages/ui/src/mosaic/user-button/user-button.controller.tsx
Ephem: Fixed in b9ded7b421fbd6fc55cc1ac4c40e332766749dec

packages/ui/src/mosaic/user-button/user-button.model.tsx
Ephem: Turned out we only needed `hidePersonal` to do it, added the view part and a swingset story in e92f42bf9fa9b040f2d84a59356701145d9cfa11

@alexcarpenter This is one you might want to review so I'll leave the comment open.

CleanShot 2026-08-27 at 11 03 03

packages/ui/src/mosaic/user-button/user-button.model.tsx
Ephem: Fixed and added prop in e2f08b213b5c3b48a0c80778c6a29cfbb1b9dda9

packages/ui/src/mosaic/user-button/user-button.model.tsx
Ephem: Fixed by using `allSettled` to swallow the errors here: 3bf3eac2742d5be29fbbebc6265a7e711444d2de

packages/ui/src/mosaic/user-button/user-button.model.tsx
Ephem: Fixed by awaiting the revalidation in 3bf3eac2742d5be29fbbebc6265a7e711444d2de

packages/ui/src/mosaic/user-button/user-button.controller.tsx
Ephem: Status is now derived from the frozen model in: cdb6f8128b58fd8dd19791d1f49cc61588665511

packages/ui/src/mosaic/user-button/user-button.tsx
Ephem: 3f2827f997b88473771b215994d38958b8d45a98

packages/ui/src/mosaic/user-button/user-button.controller.tsx
Ephem: Did it: 3cbe8f3d273c2bc7451c14698b137afc29c5319e

packages/ui/src/mosaic/user-button/user-button.pages.tsx
Ephem: I think it still goes through navigate later on. Closing.

packages/ui/src/mosaic/user-button/user-button.model.tsx
Ephem: Thinking about it some more, I don't really see the point, at least right now. I think I was just overthinking the structure. Closing.

packages/ui/src/mosaic/user-button/user-button.model.tsx
Ephem: Added this list as a TODO here: d685fc70b5713036141993de229aa9efbaa5bbca

packages/ui/src/mosaic/user-button/user-button.controller.tsx
Ephem: I refactored this in 88eef780067f8528fd5d31d2e090a521a153f642 so open/closed is now a context value instead of an explicit state.

@alexcarpenter I think this is a pattern we'll likely want to reuse for other popups so closing during a pending action doesn't throw away the action state. Often we can't truly cancel the action, so either we want to block the popup from closing until it's done, or we want to keep the action state when re-opening, depending on usecase.

This is resolved, but keeping comment open for visibility.

Ephem (approved):

PR #9186, opened 2026-07-16
ci(e2e): Add Expo native E2E testing

wobsoriano:
## Description

This PR adds E2E coverage for our `@clerk/expo` SDK (finally!) using Maestro.

Each flow is a single scenario, and every native interaction is followed by an assertion that the JS layer observed it, so the native-to-JS session sync (and the reverse) is what's actually under test:

- **sign-in.yaml:** signs in through the native `AuthView` with email + password, asserts the JS side sees the session (useAuth state and user id rendered by the fixture), then signs out from the JS side and asserts the native session is gone.

- **session-persists-across-restart.yaml**: signs in, fully kills and relaunches the app without clearing state, and asserts the session is restored from secure-store with no re-auth.

- **authview-detach-reattach.yaml**: opens the `AuthView`, dismisses it with the platform-native gesture (Android back button, iOS sheet close), reopens it, and completes a full sign-in on the remounted view.

- **user-button-sign-out-re-sign-in.yaml:** opens the `UserButton` account sheet, signs out natively, asserts JS observes it, then signs in again in the same app process.

Cold launch and initialization are exercised implicitly.

Rather than a new workflow, this reuses the existing `expo-native-build.yml`. The SDK 57 matrix entries get `run-e2e` flag and run the Maestro suite after their normal fixture build. If we want e2e on other SDK versions later, it's just a matter of adding the same flag to their matrix entries.

## Summary by CodeRabbit

* **New Features**
  * Added Expo Native end-to-end authentication coverage on iOS and Android (AuthView open/close, sign-in/sign-out, re-sign-in, and session persistence) using Maestro.
  * Added a dedicated integration test runner script and an npm command to execute these flows.
  * Updated the Expo Native test fixture UI with explicit, testable sign-out controls.
* **Bug Fixes**
  * Improved CI E2E reliability with warmup, per-flow retries, conditional E2E setup, safer debug/artifact handling, and robust cleanup on failures.


.github/workflows/expo-native-build.yml
```diff
@@ -48,9 +53,11 @@ jobs:
           - expo-sdk: 57
             platform: android
             runner: blacksmith-8vcpu-ubuntu-2204
+            run-e2e: true
```
wobsoriano, line 56: We just need to add this in other SDK matrix if we want to E2E test them

package.json
```diff
@@ -45,6 +45,7 @@
     "test:integration:custom": "pnpm test:integration:base --grep @custom",
     "test:integration:deployment:nextjs": "pnpm playwright test --config integration/playwright.deployments.config.ts",
     "test:integration:electron": "pnpm test:integration:base --grep @electron",
+    "test:integration:expo-native": "integration/tests/expo-native/run-flows.sh",
```
wobsoriano, line 48: will document this, but for now to run locally you'd need Maestro CLI installed and this

```bash
CLERK_TEST_EMAIL=your_email CLERK_TEST_PASSWORD=your_passowrd pnpm test:integration:expo-native
```

mikepitre (approved):
awesome!

PR #9191, opened 2026-07-17
feat(ui): connect the Mosaic UserButton and its custom UserProfile pages

alexcarpenter:
## Description

Stacked on #9185. Covers the connected `UserButton` through the real view and the real controller against a mocked Clerk, settles the popover behaviour that coverage exposed, and carries the app's own pages into the UserProfile the button opens.

**One flow, one machine.** The popover's open state and the one action in flight are the same flow: an action that ends the interaction closes the surface, so they settle together or not at all. `user-button.machine.ts` holds both instead of two `useState`s in the container. Re-entry, clearing busy, and closing on success stop being hand written there: `RUN` is simply unhandled while busy, and busy is only reachable from open. Dismissing the popover mid action now abandons the result rather than letting it land in a surface that is already gone.

**The surface holds still while an action runs.** The view renders the controller the action started from. `setActive` swaps the active organization while its promise is in flight, so the live controller would otherwise rearrange the popup under the pointer: the header renaming itself, the check jumping rows, Invite coming and going as the permission is re-read. The result lands in one step when the action settles.

**The popover closes behind whatever it opens.** A modal or another page takes over from there, so there is nothing left for the popover to show. Left up, it would sit over the very surface it just opened.

**A failure leaves the popover open** so the row can be clicked again. Nothing reports what went wrong yet; the error surface is its own change.

**`mode` reaches the connected button.** Organizations turned off at the instance leave nothing for an organization surface to lead with or to list, so the button is the account's whatever `mode` asked for. clerk-js withholds its own `<OrganizationSwitcher>` at the mount boundary, and nothing mounts this one, so the gate lives in the container.

### Custom UserProfile pages

`userProfileProps` puts your own pages and links into the UserProfile the button opens, and orders that profile's navigation.

```tsx
<UserButton
  userProfileProps={{
    customPages: [
      { label: 'Terms', path: 'terms', icon: <FileIcon />, content: <Terms /> },
      { label: 'Docs', path: 'docs', href: 'https://clerk.com/docs', icon: <BookIcon /> },
    ],
    pageOrder: ['account', 'terms', 'security', 'docs'],
  }}
/>
```

Plain objects rather than the `<UserButton.UserProfilePage>` children the pre-Mosaic button takes. Every entry has a `path`, which is what identifies it: with `content` it is a page inside the profile, with `href` it is a link out of it. `icon` is optional on both. `pageOrder` entries are ids: `account`, `security`, `billing` or `apiKeys` for a built-in page, or a custom entry's `path`.

Objects also drop the `'use client'` requirement the children API carries. `UserButton.UserProfilePage` means dotting into a client module, which RSC forbids ("Cannot access UserProfilePage on the server"), so describing config forced the directive onto the consumer's file and pulled the page into the client bundle. A prop has nothing to dot into: this renders from a server component, and `content` can stay a server component too.

The profile still opens in clerk-js's own React root, which cannot render a node from the host app's tree, so `useCustomPages` bridges the two: each entry goes out as the `mount`/`unmount` pair clerk-js already takes, and the host tree portals the content into the element it hands back. The portals hang off the container rather than the popover, since they have to outlive whatever opened the profile. When a Mosaic UserProfile lands in-tree, the bridge goes and the props stay.

None of this is public surface. The experimental entry exports `UserButton` and `UserButtonProps` and nothing else, so the hooks and the page types stay internal.

Two things worth a look in review:

- The icon callbacks go out whether or not there is an icon. clerk-js decides what an item _is_ from which callbacks are present and drops one missing an icon pair as invalid, so without them, leaving `icon` off would silently cost you the page.
- clerk-js puts every built-in page it was _not_ asked to move ahead of everything it was, so `pageOrder` has to send the built-ins that were left out too, appended after the ones that were named. That means knowing which built-ins the instance actually shows, which is what `useUserProfilePages` computes, behind the same shared guards clerk-js uses. It cannot see `shouldShowBilling` (paid plans / past subscriptions, fetched inside the profile), so an instance with billing enabled but no visible plans gets one dev-only log and the page is dropped, which is what should happen to it anyway.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [x] other: test coverage


kylemac:
!snapshot

packages/ui/src/mosaic/user-button/user-button.tsx
```diff
@@ -55,10 +75,14 @@ export type UserButtonProps = UserButtonControllerOptions &
...
  * @example
- * `customMenuItems` adds your own rows to the foot of the menu, each one either an `onClick` action
- * or an `href` link, and `menuItemOrder` names the order the foot's rows run in.
+ * `customPages` adds your own pages to the profile this button opens; `customMenuItems` adds your
+ * own rows to the foot of the menu, each one either an `onClick` action or an `href` link.
  * ```tsx
  * <UserButton
+ *   userProfileProps={{
+ *     customPages: [{ path: 'usage', label: 'Usage', icon: <ChartIcon />, content: <UsagePage /> }],
+ *     pageOrder: ['account', 'usage', 'security'],
```
Ephem, line 90: It was not very clear to me from reading this that `pageOrder` (and `menuItemOrder`) could contain the built in pages by using magic strings only available in the TS types?

I'm not sure how I feel about that API, and also how defining a custom page with one of those ids will kinda silently fail today. We could export a `BUILT_IN_PAGES` constant or something, but I'm not sure that feels better.

If nothing else, I think this doc-comment can be clearer, and maybe we should log a warning or something on conflicts?

packages/ui/src/mosaic/user-button/user-button.machine.ts
```diff
@@ -0,0 +1,82 @@
...
+export type UserButtonReadyController = Extract<UserButtonController, { status: 'ready' }>;
+
+export interface UserButtonMachineContext {
+  /** The affordance that owns the action in flight: it spins, and every other one stands down. */
+  pendingKey: string | null;
+  /**
+   * The controller the action started from. `setActive` swaps the active organization while its
+   * promise is still in flight, so the live controller would rearrange the popup mid-action: the
+   * header renaming itself, the check jumping rows, Invite coming and going as the permission is
+   * re-read. The view renders this instead until the action settles.
```
Ephem: This is where using a React transition for the setActive would have been soo nice, but alas.

packages/ui/src/mosaic/user-button/user-button.machine.ts
```diff
@@ -0,0 +1,82 @@
...
+            frozen: event.frozen,
+            run: event.run,
+            closeOnSuccess: event.closeOnSuccess,
+          })),
+        },
+      },
+    },
+    // Reached only from `open`, so a busy popup that is not open is unrepresentable, and RUN going
+    // unhandled here is what stops a second action starting while one is in flight. Dismissing the
+    // popup abandons the action: the request finishes, but nothing is left for its result to land in.
```
Ephem: We probably have to do this because of how `setActive` works, but it's unfortunate the UI feels locked up meanwhile.

What happens if you close the popup during a pending action, reopen it and trigger a run? That would lead to the duplicate runs we are trying to avoid right?

packages/ui/src/mosaic/user-button/user-button.tsx
```diff
@@ -68,24 +92,48 @@ export type UserButtonProps = UserButtonControllerOptions &
...
+  const builtInPages = useUserProfilePages();
+  const { customPages, portals } = useCustomPages({
+    items: userProfileProps?.customPages,
+    order: userProfileProps?.pageOrder,
+    builtInPages,
+  });
+  const controller = useUserButtonController(options, customPages);
+  // The popover's open state and the one action in flight are the same flow: an action that ends the
+  // interaction closes the surface, so they settle together or not at all.
+  const [{ value, context }, send] = useMachine(userButtonMachine);
```
Ephem: Something feels a tad off here considering the flirtation with the MVC model. I think the controller here is actually a model (data+mutations) and the machine is actually the controller (tying the data/business layer to the UI layer, aka state)? 🤔

Also, what if we make the machine a custom hook and pass in the model? That way we could move runAction and handoff into there and encapsulate things a bit better?

packages/ui/src/mosaic/user-button/user-button.pages.tsx
Ephem, line 1: 😮😬

This is clever, but it's going to be very nice when we can move the pages over to the host app just like the new user button and get rid of it. 😂 Is this like a reverse [ClerkHostRenderer](https://github.com/clerk/javascript/blob/83f50f68619205f16541439fd27ca653686ba6df/packages/react/src/components/ClerkHostRenderer.tsx)?

I haven't reviewed it fully yet, will take another pass when my brain is on point.

packages/ui/src/mosaic/user-button/user-button.machine.ts
alexcarpenter: in this commit we reset on close https://github.com/clerk/javascript/pull/9191/changes/27b86e6a4fb8c072b3725b003b8df4b91424d844

packages/ui/src/mosaic/user-button/user-button.tsx
alexcarpenter: open to other ideas here. kind of how it currently works https://clerk.com/docs/nextjs/guides/customizing-clerk/adding-items/user-button#reorder-default-items but imo this feels better.

Ephem:
Folded into #9185.

PR #9198, opened 2026-07-20
feat(ui): add experimental flag to gate the OIDC self-serve flow

NicolasLopes7:
## Summary

Adds an experimental `oidcSelfServe` flag to the ClerkProvider `experimental` options, laying the groundwork to gate a forthcoming OIDC self-serve configuration flow in `<ConfigureSSO />`.

- Opt in via `<ClerkProvider experimental={{ oidcSelfServe: true }} />` (default off).
- Surfaces the flag to the ConfigureSSO wizard and its steps as `isOIDCFlowEnabled` via `useConfigureSSO()`.
- No user-facing surface is gated yet; the OIDC provider path lands separately. The flag lives in the SemVer-exempt `experimental` namespace.


.changeset/empty-ducks-tease.md
```diff
@@ -0,0 +1,2 @@
+---
```
dstaley, line 1: you'll want to add a changeset or else it won't release the change to a new version

dstaley (approved):

PR #9222, opened 2026-07-22
fix(expo): Prevent stale native clients from replacing signed-in sessions

wobsoriano:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/expo/src/provider/nativeClientSync.tsx
```diff
@@ -204,44 +206,85 @@ function getDefaultSignedInSession(client: ClientResource | null | undefined): S
...
+  try {
+    if (shouldSyncDeviceToken) {
+      await syncNativeDeviceTokenToCache({
+        deviceToken: nativeDeviceToken,
+        suppressTokenCacheNotificationsRef,
+        tokenCache,
+      });
+    }
+
+    refreshedClient = await fetchRefreshedJsClient(clerkInstance);
```
mikepitre, line 274: I think there's a window here worth closing before we ship: the native token lands in the shared cache before the fetch resolves, and if a session token refresh 401s in that window, the intercepted `handleUnauthenticated` recovery https://github.com/clerk/javascript/blob/c9a13a1fd632054a6fabd1091c3c341d5302c69e/packages/expo/src/provider/nativeClientSync.tsx#L778-L788 runs `refreshJsClientFromNativeState` without the new guard — so it adopts the foreign sessionless client via `updateClient`. The outer validation then restores the token but not the client/session state, which leaves a good token in the cache with `session` null. I managed to reproduce it with a regression test, so it seems worth handling.

A couple of ideas, curious what you think:
1. Share the in-flight validation promise and have the recovery path await it — once the good token is restored the 401 should mostly resolve itself.
2. Or simply thread `rejectForeignSessionlessClient` + previous client/token into the recovery call too, so every path gets the same protection.

The fully airtight version (validating the candidate token without publishing it first) would need `client.fetch` to take an explicit token, which is clerk-js plumbing — totally fine as a follow-up.

packages/expo/src/provider/nativeClientSync.tsx
```diff
@@ -434,12 +502,27 @@ async function syncNativeClientToJs({
     return;
   }
 
+  const previousDeviceToken = didChangeDeviceToken ? await getCachedDeviceToken(tokenCache) : undefined;
```
mikepitre, line 484: Small edge case: `getCachedDeviceToken`'s 1s timeout resolves to `null`, same as "confirmed no token". If the cache read is just slow and we later hit the rejection/error path, the restore ends up calling `clearToken` on a token that actually exists. Maybe return `undefined` on timeout so the restore stays a no-op? Could also argue for skipping native adoption entirely in that case since we don't have a reliable snapshot — but I don't feel strongly about that part.

packages/expo/src/provider/nativeClientSync.tsx
```diff
@@ -402,6 +445,31 @@ async function getCachedDeviceToken(tokenCache: TokenCache | undefined): Promise
   }
 }
 
+function getDeviceTokenIssuedAt(deviceToken: string | null): number | null {
```
mikepitre: One thing I noticed while checking the server side: client cookie JWTs are signed with exactly `{id, rotating_token}`
(https://github.com/clerk/clerk_go/blob/01fd589c0a2a283317ddfb89674bcd5d62b71e4f/model/client_cookie.go#L22-L25) and nothing injects `iat`, so I believe this branch can't fire with real tokens.

mikepitre (changes requested):
Really nice work tracking this down — I went through the FAPI side to sanity-check the assumptions and the foreign-sessionless rejection holds up well against what the server actually does (a stale token on `GET /v1/client` re-mints a lazy client reusing the same id, so the validation fetch is deterministic). I found one interaction with the 401 recovery path that I think can still reproduce the original bug, plus two smaller things — details inline.

packages/expo/src/provider/nativeClientSync.tsx
wobsoriano: nice catch, thanks! I went with option 2, the recovery path now passes `rejectForeignSessionlessClient` and a previous token snapshot so it goes through the same guard as the event path

packages/expo/src/provider/nativeClientSync.tsx
wobsoriano: fixed with your suggestion 🙏🏼 `getCachedDeviceToken` now returns undefined on timeout and null only for a confirmed missing token

packages/expo/src/provider/nativeClientSync.tsx
wobsoriano: haha yeah, confirmed and removed the whole branch

mikepitre (approved):
@wobsoriano I didn't test manually but if all looks good with your manual testing then the code itself looks good to me!

PR #9231, opened 2026-07-23
feat(headless): add useRender hook alongside renderElement

alexcarpenter:
## Why

`renderElement` (function) → `useRender` (hook). Wins:

**1. render-as-element.** Today every override must be a function wrapper. Now pass the element directly.

```tsx
// before
<Trigger render={props => <Link {...props} href="/home" />} />
// after
<Trigger render={<Link href="/home" />} />
```

**2. ref merging moves inside.** Drop the per-call-site `useMergeRefs`.

```tsx
// before
const combinedRef = useMergeRefs([popupRef, refs.setFloating, ref]);
if (!mounted) return null;
return renderElement({ defaultTagName: 'div', render, props: { ref: combinedRef, ...rest } });
// after
const el = useRender({ defaultTagName: 'div', render, enabled: mounted, ref: [popupRef, refs.setFloating, ref], props: rest });
if (!el) return null;
return el;
```

**3. Mosaic: one path, no cast.**

```tsx
// before
if (render) return renderElement({ defaultTagName: 'h2', render, props: { ref, ...root, ...rest } as Record<string, unknown> });
return <h2 ref={ref} {...root} {...rest} />;
// after
return useRender({ defaultTagName: 'h2', render, ref, props: { ...root, ...rest } });
```

Also aligns with base-ui (swingset already uses real base-ui `useRender`).

## Scope

PR1 of a phased series. Adds `useRender` (own `utils/use-render.tsx`), keeps `renderElement` exported for the rest of the migration. Migrates Mosaic `Box`/`Heading`/`Text` + `dialog-trigger`/`dialog-popup` as representative patterns. Follow-ups convert remaining ~60 primitives by family; final PR deletes `renderElement`.

Reuses Clerk's `mergeProps` + `useMergeRefs` (`@floating-ui/react`, already a dep). No new runtime dep, no base-ui engine vendored.

Empty changeset: `@clerk/headless` private, migrated Mosaic parts not in `@clerk/ui` public exports (matches #9208).

## Verified

`@clerk/headless` typecheck + 601 tests pass. `@clerk/ui` type-check clean.


wobsoriano (approved):
clean!

PR #9233, opened 2026-07-23
feat(headless): remove renderElement, consolidate into useRender

alexcarpenter:
Stacked on #9232. Merge that first.

Final PR of the `renderElement` → `useRender` migration. After PR2, `renderElement` had zero callers, so this removes it.

## What changed

`render-element.tsx` was doing double duty: it held the dead `renderElement` function **and** the shared building blocks every primitive still uses (the prop types + `mergeProps`). So this is a move-then-delete, not a bare delete:

- Moved `RenderProp`, `ComponentProps`, `DefaultProps`, `StateAttributesMapping`, and `mergeProps` into `use-render.tsx` (its sole consumer now).
- Repointed the `utils` barrel to re-export them from `./use-render`. Public barrel surface is unchanged except `renderElement` is gone (nothing imported it).
- Deleted `render-element.tsx`, `render-element.test.tsx`, `render-element.test-d.ts`.
- Ported the `mergeProps` describe block into `use-render.test.tsx` and the `ComponentProps` type test into `use-render.test-d.ts` so no coverage is lost. The `renderElement`-specific behavioral tests are dropped with the function (their behavior is already covered by the `useRender` cases from PR1).

## Verification

- `grep renderElement` / `grep render-element` across `packages/headless/src` + `packages/ui/src` → zero matches.
- `pnpm --filter @clerk/headless typecheck` → clean.
- `pnpm --filter @clerk/headless exec eslint src` → exit 0, zero warnings.
- `pnpm --filter @clerk/headless test` → 591 passed, 1 skipped.

Empty changeset (`@clerk/headless` is private).

The two remaining `useMergeRefs` imports (`tooltip-arrow.tsx`, `popover-arrow.tsx`) are correct — those parts render `FloatingArrow` directly and never used `renderElement`.


wobsoriano (approved):
🧹

PR #9238, opened 2026-07-24
docs(repo): codify StyleX best practices in the Mosaic skill

alexcarpenter:
## Summary

Expands `.claude/skills/mosaic/references/stylex.md` into an evidence-backed do/don't StyleX authoring guide, from a full sweep of the astryx design system (~120 components).

New/rewritten sections:
- **Breaking up `stylex.create`** — base object + prop-keyed variant maps + small state objects, composed in one `stylex.props(...)`; don't monolith, don't per-property, don't inline-ternary a variant (that's a missing map).
- **Dynamic styles: dos and don'ts** — dynamic functions are for continuous/derived runtime values only (not closed enums). Key sub-distinction: write a `--var` and let a static rule consume it (one atom, survives consumer `@media` overrides, inherits to descendants) vs. set the property raw (per-instance only).
- **Scoped ancestor states** — `stylex.when.ancestor()` + `defineMarker()` in place of parent-selector hacks.
- Tightened tokens, conditions/state, and composition guidance throughout.

## Corrections / decisions
- Marker files are `<comp>.markers.stylex.ts` (a `.stylex.ts` define-primitive file), not `.markers.ts` — verified against the `@stylexjs/eslint-plugin` `enforce-extension` rule source.
- File-layout section now separates **rule-forced facts** (`.stylex.ts` = `define*` only; `create` may not live there) from **Mosaic house convention**: `stylex.create` lives in a co-located `<comp>.styles.ts`, never inline in the `.tsx`. This is a deliberate divergence from astryx (which inlines), recorded as our choice rather than a research finding.

## Verification
Key claims spot-checked against astryx source: `when.ancestor`/`defineMarker` (real), dynamic functions (47 uses), `firstThatWorks`/`attrs` (0 uses), the CodeBlock/AvatarGroup var-writing pattern, and the `enforce-extension` behavior.

## Changeset
Empty — repo/tooling-only doc change, no user-facing package impact.


kylemac (approved):
stamp made of rubbah

PR #6486, opened 2025-08-07
feat(clerk-js): Introduce `navigate` for `setActive`

LauraBeatris:
## Description

This PR introduces a new API to navigate on `setActive`, that can be leveraged for after-auth flows to handle pending sessions, featuring a callback mechanism that enables developers to integrate their application logic, such as navigation. 

Previously, our flows were mistakenly navigating after `setActive`, which could cause issues with session revalidation and re-render of pages.

From now on, we're proposing two options to deal with after-auth: `taskUrls` or `navigate`

### `navigate`

A callback function that can be provided as an `setActive` option.

```ts
setActive({ 
  navigate: async ({ session }) => { 
     const currentTask = session.currentTask; 

     if (currentTask){ 
      // Developer has full control over navigation logic
      // (this also allows our AIO components to pass their navigation context such as modal routing)
       await router.push(`/onboarding/${currentTask.key}`)
       return 
     }

    await router.push('/dashboard')
  }
})
```

### `taskUrls` (existing)

A map of URLs per session task key that can be provided as a Clerk option. On `setActive`, it'll be used for navigation.

```tsx
<ClerkProvider 
   taskUrls={{'choose-organization': '/onboarding/choose-organization' }}
/>
```

### AIOs behavior 

https://github.com/user-attachments/assets/5a0c5f52-f8fd-45d0-9b56-1b82a52a1e87

## Checklist

- [X] `pnpm test` runs as expected.
- [X] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [X] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

## Summary by CodeRabbit

* **New Features**
  * Added per-task routing APIs (buildTasksUrl, redirectToTasks) and optional navigate callback for setActive; sign-in/sign-up contexts and session-tasks expose navigateOnSetActive and taskUrl; new HOCs to redirect to sign-in/sign-up tasks.

* **Refactor**
  * Renamed RedirectToTask→RedirectToTasks and SessionTask→SessionTasks; centralized task-based navigation and removed legacy internal navigate-to-task plumbing.

* **Tests**
[...]

PR #9254, opened 2026-07-27
fix(clerk-js): apply the Safari ITP cookie refresh on sign-out

dmoerner:
Signing out re-issues the client cookie from a fetch, which Safari's ITP caps at 7 days. The client outlives sign-out — persistClient defaults to true, so `signOut()` removes the sessions and keeps the client — and the client is what Client Trust uses to recognise a known device. Leaving the cap in place meant a user who signed out and returned more than a week later had lost the cookie, arrived as a new client, and was challenged for a second factor.

Route the sign-out redirect through `/v1/client/touch` when the cookie is close to expiring, which is the workaround `setActive()` already applies after sign-in: touch is reached by a top-level navigation, which ITP does not cap, so it restores the full lifetime. The redirect is returned unchanged when no refresh is needed, so this only adds a hop on Safari — every other browser gets a 400 day cookie and is never eligible.

This covers the `signOut()` paths that navigate to a redirect URL, including <UserButton /> and <SignOutButton />. `signOut(callback)` navigates on the caller's behalf and is left alone.

Fixes USER-5769.

Depends on https://github.com/clerk/clerk_go/pull/20738 (merged)

✅ This fix was tested in staging E2E.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -657,25 +657,22 @@ export class Clerk implements ClerkInterface {
   }
 
   /**
-   * Routes the post-sign-out navigation through `/v1/client/touch` when the client
-   * cookie is close to expiring, and returns `redirectUrl` unchanged otherwise.
+   * Routes `url` through `/v1/client/touch` when the client cookie is close to expiring,
+   * and returns it unchanged otherwise.
    *
-   * In practice this only adds a redirect on Safari. Signing out re-issues the client
-   * cookie from a fetch, which ITP caps at 7 days; every other browser gets 400 days
-   * and is never eligible. The cap matters because the client outlives sign-out and is
-   * what Client Trust uses to recognize a known device, so a user returning a week
-   * later would be challenged for a second factor. Touch is a top-level navigation,
-   * which ITP does not cap, so it restores the full lifetime.
+   * In practice this only adds a redirect on Safari. A cookie re-issued from a fetch is
+   * capped at 7 days by ITP; every other browser gets the full 400 days and is never
+   * eligible. Touch is a top-level navigation, which ITP does not cap, so it restores
+   * the full lifetime.
    */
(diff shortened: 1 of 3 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -656,6 +656,28 @@ export class Clerk implements ClerkInterface {
...
+   * which ITP does not cap, so it restores the full lifetime.
+   */
+  #buildSignOutRedirectUrl(redirectUrl: string): string {
+    // In React Native, window exists but window.location does not, so the redirect
+    // cannot be resolved to an absolute URL. ITP does not apply there either.
+    if (!inBrowser() || typeof window.location === 'undefined' || !this.client?.isEligibleForTouch()) {
+      return redirectUrl;
+    }
+
+    const absoluteRedirectUrl = new URL(redirectUrl, window.location.href);
```
dstaley: should we consolidate this logic into a helper shared between setActive and buildSignOutRedirectUrl? It looks like the logic is exactly the same, so feels like a good candidate for that.

dmoerner: Sure! Abstracted in [ab39dee](https://github.com/clerk/javascript/pull/9254/commits/ab39deec2fddd9a0907ee41e2d73e4026d908855)

There is a third path where we could also deduplicate but I believe that there may have been a regression introduced in https://github.com/clerk/javascript/pull/6486/changes#diff-1952d6b8ac4b5495e1dba57c32bff9ce72d43f83595c1bc8532a76cf31feb753L1315 (we dropped the `else` and so we immediately clobbered ITP fix navigate). Fixing that seems out of scope here, I'll make a separate ticket.

dstaley (approved):

PR #9255, opened 2026-07-27
fix(ui): remove ticket parameters on sign up continue

dstaley:
## Description

This PR updates the `handleComplete` callback on `SignUpContonue` to remove Clerk's ticket query parameter before calling `setActive`. This prevents a remount attempting to consume the already consumed ticket.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/components/SignUp/SignUpContinue.tsx
```diff
@@ -180,13 +181,16 @@ function SignUpContinueInternal() {
           verifyEmailPath: './verify-email-address',
           verifyPhonePath: './verify-phone-number',
           protectCheckPath: '../protect-check',
-          handleComplete: () =>
-            clerk.setActive({
+          handleComplete: () => {
+            removeClerkQueryParam('__clerk_ticket');
```
tmilewski, line 185: This cleanup only runs when `signUp.update()` completes immediately. `completeSignUpFlow` forwards the ticket into email/phone verification and protect-check routes, whose terminal handlers call `setActive` without removing it, so those flows can still remount and attempt to consume an already-used ticket. Please apply the cleanup to every terminal sign-up completion path (preferably via a shared helper) and cover a continue → verification → activation case.

~ 🤖

dstaley: I was a bit nervous to impact ~all the sign up routes, but I feel like removing the tickets universally _should_ be safe. So I updated to move the removal logic into `completeSignUp` instead of forcing each caller to individually remove the parameters.

tmilewski: We don't HAVE to, I just wanted to call it out. Feel free to ignore for now.

tmilewski (approved):

PR #9258, opened 2026-07-27
chore(nuxt): Remove createRouteMatcher support

wobsoriano:
## Description

This PR removes `createRouteMatcher` from `@clerk/nuxt`, following the same removal we did for Astro in #8974.

Another thing, since Nuxt 3 reaches [end-of-life on July 31, 2026](https://nuxt.com/blog/v4-5#nuxt-3-end-of-life), we now require Nuxt 4.

Docs PR: https://github.com/clerk/clerk/pull/3043

## Checklist

- [ ] `pnpm test` runs as expected.
- [ ] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #9269, opened 2026-07-29
docs(nuxt): Use isAuthenticated in createRouteMatcher migration examples

wobsoriano:
## Description

Follow-up to #9258. The migration examples in the pending `@clerk/nuxt` major changeset checked `userId` truthiness, while the docs (clerk/clerk#3043) standardize on the `isAuthenticated` property for authentication checks. This updates the changeset's event handler example and agent migration prompt to match, so the changelog and docs recommend the same check.

No code changes, changeset content only.

## Checklist

## Type of change


wobsoriano:
!allow-major

SarahSoutoul (approved):

PR #9272, opened 2026-07-29
fix(expo): Sanitize stray hash fragment in SSO callback URL

wobsoriano:
## Description

Native SSO sign-ins can fail with a 401 signed_out at the sign-in reload when the OAuth callback deep link carries a stray `#`. Some iOS redirect chains leave a fragment on the redirect URL, either literal or percent-encoded into the `rotating_token_nonce` value, so the nonce reaches FAPI as `<nonce>#` and the exact-match check fails.

`useSSO` now extracts the nonce through a helper that strips fragments from the callback URL and trims a # and anything after it from the extracted value.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/expo-callback-nonce-hash.md
+++ b/.changeset/expo-callback-nonce-hash.md
@@ -1,0 +1,5 @@
+---
+'@clerk/expo': patch
+---
+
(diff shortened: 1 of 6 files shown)


mikepitre (approved):

wobsoriano:
closed in favor of logging an error if the redirectUrl they pass in `useSSO` is missing a path

PR #9274, opened 2026-07-29
fix(ui): guard clipboard data in paste handlers

brkalow:
## Description

Prevents `PhoneInput` and `TagInput` from assuming clipboard data is present on every paste event. Programmatically dispatched paste events can expose null clipboard data, which caused an unhandled `TypeError` in Clerk UI.

The affected handlers now ignore events without clipboard data while preserving normal paste behavior. Regression coverage exercises both the sign-in phone number field and the organization invite multi-email field.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/calm-taxis-paste.md
+++ b/.changeset/calm-taxis-paste.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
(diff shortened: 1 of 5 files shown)


wobsoriano (approved):
makes sense

PR #9278, opened 2026-07-29
fix(electron): handle Linux OAuth deep links

jeremy-clerk:
## Description

Acquire Electron’s single-instance lock for OAuth deep links and forward secondary-instance callbacks to the primary process. Hosts that manage the process-wide lock can opt out with `manageSingleInstanceLock: false`.

## Checklist

- [x] (If applicable) [JSDoc comments](https://jsdoc.app/) have been added or updated for any package exports

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/calm-taxis-return.md
+++ b/.changeset/calm-taxis-return.md
@@ -2,4 +2,4 @@
 '@clerk/electron': patch
 ---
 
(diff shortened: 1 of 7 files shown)


wobsoriano (approved):
thanks for the quick solution 🥇

PR #9292, opened 2026-07-30
refactor(ui): migrate Mosaic Card to StyleX

austincalvelage:
## Description

https://swingset-git-austin-mosaic-card-component.clerkstage.dev/components/card

Updates the Mosaic Card to match the current [Figma](https://www.figma.com/design/yt5qKlfNyIIdaKaMoav1v2/Clerk-Mosaic-NEW?node-id=8230-12146):

- Migrates Card styling from Emotion to StyleX.
- Aligns the Card root, elevation, shadow, radius, and footer treatment with the design.
- Updates the API to use the compound `Card.Root`, `Card.Header`, `Card.Content`, and `Card.Footer` structure.
- Adds `raised`, `flush`, and `overlay` elevation options.

```tsx
import { Card } from '@clerk/ui/mosaic/components/card';

<Card.Root elevation='raised'>
  <Card.Header>Heading</Card.Header>
  <Card.Content>Body content.</Card.Content>
  <Card.Footer>Footer actions</Card.Footer>
</Card.Root>;
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/components/card/card.styles.ts
```diff
@@ -0,0 +1,76 @@
...
+    boxShadow: 'none',
+  },
+  overlay: {
+    borderRadius: radiusVars['--cl-radius-container'],
+    overflow: 'hidden',
+    backgroundColor: colorVars['--cl-color-card'],
+    boxShadow: `0 12px 12px -7px oklch(0.2046 0 0 / 12%),
+                0 24px 24px -10px oklch(0.2046 0 0 / 4%),
+                0 0 0 1px oklch(0.2046 0 0 / 4%)`,
+  },
```
maxyinger, line 63: this a placeholder for now? or a backwards compatibility thing? (noticing it mirrors raised)

packages/ui/src/mosaic/components/card/card.tsx
```diff
@@ -0,0 +1,109 @@
...
+    ref,
+    props: {
+      ...rest,
+      ...mergeStyleProps(
+        themeProps('card-root', { alignment, elevation }),
+        stylex.props(styles.root, elevations[elevation]),
+        className,
+        style,
+      ),
+    },
```
maxyinger: small thing, but we've been doing mergeStyleProps before rest in case someone want to override data props. probably not a scenario we'll run into much, but what has been going on in the previous components we've done

packages/ui/src/mosaic/components/card/card.styles.ts
```diff
@@ -0,0 +1,76 @@
...
+    flexBasis: 'auto',
+    flexGrow: '1',
+    flexShrink: '1',
+  },
+  footer: {
+    gap: space['2'],
+    paddingBlock: space['4'],
+    paddingInline: space['6'],
+    alignItems: 'center',
+    backgroundColor: `color-mix(in oklab, ${colorVars['--cl-color-card-foreground']} 4%, ${colorVars['--cl-color-card']})`,
```
maxyinger: is this moving to just white with the latest mocks?

maxyinger (commented):
What do we want to do with darkmode here? my gut says to move all shadows to just a  1px spread white shadow so users are not dealing with dark shadows on potentially black bgs.

Only thing is we want to explicitly move away from the 1px spread shadows acting as outer borders right?

idk. any thoughts?

packages/ui/src/mosaic/components/card/card.tsx
austincalvelage: Good call!

packages/ui/src/mosaic/components/card/card.styles.ts
austincalvelage: So in the Figma, these currently mirror each other. I'd assume these eventually would differ?

maxyinger: cool cool. yea I'm guessing so too 🤷 . so basically just a placeholder for now I guess while a shadowing system is still tbd

maxyinger: I think we want to move away from that `0 0 0 1px oklch(0.2046 0 0 / 4%)` shadow tho.. maybe worth syncing up with @stvhayes here

packages/ui/src/mosaic/components/card/card.styles.ts
austincalvelage: Good call! I was just going off the Figma I'll adjust

austincalvelage:
> What do we want to do with darkmode here? my gut says to move all shadows to just a 1px spread white shadow so users are not dealing with dark shadows on potentially black bgs.
> 
> Only thing is we want to explicitly move away from the 1px spread shadows acting as outer borders right?
> 
> idk. any thoughts?

> What do we want to do with darkmode here? my gut says to move all shadows to just a 1px spread white shadow so users are not dealing with dark shadows on potentially black bgs.

Yeah that's fair I removed the shadows on dark mode and just kept the ring.

> Only thing is we want to explicitly move away from the 1px spread shadows acting as outer borders right?

Do we have the shadows figured out yet? Yesterday Alex mentioned just build to the Figma spec as shadows weren't set in stone yet. (Totally down to switch to borders If we are ready for that)

maxyinger:
> Do we have the shadows figured out yet?
nah idt so. just wasnt' sure if there was anything more recent in the figma

cool, yea lets just ship the figma spec for now then. works for me

maxyinger (approved):

PR #9299, opened 2026-07-31
feat(js): attach an optional server-configured session token to sign-in

zourzouvillys:
## Description

Adds an optional, server-configured session token to `@clerk/clerk-js`, attached to sign-in and sign-up requests.

It is inert unless an instance's loader configuration opts into it. An instance not using it keeps today's behaviour exactly, stores nothing in the browser, and sends no additional parameters.

- **Placeholder interpolation in loader config.** A closed set — `{cid}`, `{pid}`, `{rid}`, `{instance_id}`, `{sdkver}` — substituted into a loader's attribute values and `textContent` before the element is appended. An unrecognised `{…}` is left verbatim, so an older SDK loading the same configuration stays compatible.
- **An opaque, random correlation id**, minted with `crypto.getRandomValues` as lowercase unpadded RFC 4648 base32. Nothing is derived from the device — no fingerprinting input, no clock, no user data.
- **One acquisition per browser session, shared across tabs.** Runs under the existing `SafeLock` primitive (`core/auth/safeLock.ts`) that `SessionCookiePoller` already uses, with a re-check inside the lock so concurrent tabs do not each acquire in turn. A wedged leader delays nobody, and a leader tab closing mid-run releases its lock automatically.
- **Prefetched at `Clerk.load()`** rather than at sign-in, so acquisition stays off the latency-critical path. It is bounded by a deadline and can never block or fail a sign-in: when no token is obtained, a status travels in its place and the request proceeds unchanged.
- **Parameters ride in the form-encoded body** of sign-in and sign-up POSTs. Not the query string, since a signed credential should not land in access logs along the path; and not a header, since a custom header reintroduces the CORS preflight that breaks cookie dropping in Safari — the same constraint that already forces `_method` into the query. No request header is added anywhere. The merge is an explicit step in `fapiClient.request` on a path allowlist, because the `onBeforeRequest` callbacks fire after the body is stringified.

`ProtectLoader` gains two optional fields, `tokenUrl` and `tokenTimeoutMs`, both inert unless set.

`rollout` moved out of `applyLoader` into a module-level helper so `load()` can decide participation before it starts applying loaders. Same semantics, evaluated once.

Bundle thresholds were bumped on the affected bundles to cover the addition.

## Checklist

- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated — the persistent pseudonymous identifier needs a mention in customer-facing privacy docs; tracked separately.

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/protectSession.ts
```diff
@@ -0,0 +1,733 @@
...
+  }
+
+  async #acquire(): Promise<ProtectStatus> {
+    if (!this.#cid) {
+      return 'unsupported';
+    }
+    if (this.#freshStoredToken()) {
+      // Step 1: valid and not near expiry. No lock, no loader, no network.
+      return 'ok';
+    }
```
dominic-clerk, line 566: **[MEDIUM] A page-writable localStorage entry suppresses the Protect loader entirely and makes the client report `status: ok`**

Pre-PR every loader was injected on every page load. Now the `{cid}` loader is injected only when `__clerk_protect_st` lacks a fresh token, and `#freshStoredToken()` accepts an entry that merely satisfies the `rid` regex, `typeof token === 'string' && length <= 4096` (no format check), and an `exp` within 24h. One `localStorage.setItem` therefore keeps the detection script out of the DOM — the new test asserts `elements.length === 0` on this path — and makes `getRequestParams()` emit `__clerk_protect_status: 'ok'`.

Network-blocking the loader yields the distinguishable `script_error`; this path does not, so it is a strictly quieter suppression. The actionable half is server-side: `__clerk_protect_status` is client-asserted and unauthenticated, so `ok` accompanied by a token the fraud service cannot verify should read as adversarial rather than benign.

Suggest binding freshness to a token the client cannot mint (e.g. verify the `v1.payload.mac` shape before trusting the entry) and treating unverifiable `ok` as absent server-side.

zourzouvillys: `__clerk_protect_status` is telemetry — it exists so we can diagnose *why* acquisition didn't produce a token (blocked, timed out, unsupported browser), and it's a debugging aid, not a security control. It's client-asserted and treated that way: nothing is granted on the strength of it.

The stored entry likewise gates a cache, not trust. The token is verified server-side, so an entry that doesn't verify reaches the same place as no entry at all — and writing one isn't cheaper than simply sending the same values directly, which was always possible. That's why binding freshness to an unmintable token can't be the mitigation: the client can't do the verification, and the route it would close isn't the cheap one.

Taking the shape check regardless, for a different reason: `validateToken` documents that intent and doesn't implement it, and a corrupt or truncated entry suppressing acquisition for a real user is worth closing on its own. It'll be version-agnostic so an older SDK doesn't reject a future mint.

Server-side handling of unverifiable tokens is already covered in the paired internal PR.

dominic-clerk (approved):

.changeset/lucky-pandas-observe.md
```diff
@@ -0,0 +1,12 @@
...
+
+Acquisition can never block or fail a sign-in: it is bounded by a deadline, and when no token can be obtained a status (`no_token`, `timeout`, `script_error`, `fetch_error`, `unsupported`, `http_<n>`) travels in its place and the request proceeds unchanged. Only the loader carrying the correlation id is governed by the shared token; any other configured loader is still applied on every page load. Nothing is stored in the browser unless a loader references `{cid}`, `{pid}` or `{rid}`.
+
+`ProtectLoader` gains two optional fields: `tokenTimeoutMs`, and `tokenUrl` for instances that opt into fetching the token from a dedicated endpoint instead of taking the one served with the loader.
```
Ephem: This describes the technical implementation, but it doesn't really explain to an end user reading the release changelog what it _does_.

I'd favor a two sentence summary of the capabilities this gives instead, no need for this level of technical detail here.

packages/clerk-js/src/core/protectSession.ts
```diff
@@ -0,0 +1,743 @@
...
+  readonly #lock: ReturnType<typeof SafeLock>;
+
+  /** One attempt at a time; re-armed once it has settled without leaving a usable token. */
+  #acquisition?: Promise<ProtectStatus>;
+  #acquisitionSettled = false;
+  #lastAttemptAt = 0;
+
+  /**
+   * Returns a session only when at least one loader references a placeholder — an instance not
+   * using any of them keeps today's behaviour exactly.
```
Ephem: NIT: "Today's behavior" wont age well. 😄

packages/clerk-js/src/core/protectSession.ts
```diff
@@ -0,0 +1,743 @@
...
+      return Promise.resolve('error');
+    }
+
+    return new Promise<LoaderOutcome>(resolve => {
+      const timer = setTimeout(() => resolve('timeout'), Math.max(0, deadline - Date.now()));
+      const settle = (outcome: LoaderOutcome) => () => {
+        clearTimeout(timer);
+        resolve(outcome);
+      };
+      element.addEventListener('load', settle('loaded'), { once: true });
```
Ephem, line 747: Not sure it's something that can happen in practice, but wanted to point out that since `load` only fires if `src` is set, this wont ever trigger if it's a pure inline script via `textContent`, which would put things in a bad state.

packages/clerk-js/src/core/protectSession.ts
```diff
@@ -0,0 +1,743 @@
...
+
+/** Where the loader script leaves the token it was served with. */
+const INLINE_TOKEN_GLOBAL = '__clerk_specter';
+
+/** A token this close to expiry is treated as absent, so a fresh run starts before it lapses. */
+const TOKEN_REFRESH_MARGIN_MS = 60 * 1_000;
+/** Acquisition deadline when the instance does not configure one. */
+const DEFAULT_TOKEN_TIMEOUT_MS = 5 * 1_000;
+/** Ceiling on the instance-configured deadline, so no server value can stall a sign-in. */
+const MAX_TOKEN_TIMEOUT_MS = 10 * 1_000;
```
Ephem, line 36: These are per API request, not per sign-in/up flow right? So a combined sign in/up that transitions from sign-in to sign-up and fires multiple API requests might stall for twice the time in niche scenarios?

Feels like an edge case so not necessarily something we need to tackle, just wanted to highlight it.

packages/clerk-js/src/core/protect.ts
```diff
@@ -23,31 +26,57 @@ export class Protect {
     // here rather than at end to mark as initialized even if load fails.
     this.#initialized = true;
 
-    for (const loader of config.loaders) {
+    // The config is server-controlled and cached, so nothing it can contain may take `Clerk.load()`
+    // down with it.
+    try {
+      this.#apply(config.loaders, config.id || undefined);
```
Ephem: `id` is typed as required, but treated as optional, all the way through to:

```
const suffix = instanceId ? `.${instanceId}` : '';
this.#tokenStorageKey = `${TOKEN_STORAGE_KEY}${suffix}`;
```

which made me start thinking about conflicting tokens stored under the same key etc before I realized it's probably expected to always be available?

If that's the case but we still want to be defensive at runtime (which I like), maybe we should validate its presence here and fail early instead (logging + early exit) to make the error predictable?

packages/clerk-js/src/core/protectSession.ts
```diff
@@ -0,0 +1,743 @@
...
+
+/**
+ * The store is writable by anything running on the origin, so a value that could not have come
+ * from a mint of ours is discarded rather than trusted to suppress the loaders.
+ *
+ * The shape check is hygiene, not a security boundary: only the server can tell a real token from a
+ * well-formed forgery, and anything that can write the store can send the same values to the API
+ * directly. What it buys is that a corrupt or truncated entry starts a fresh run immediately
+ * instead of suppressing acquisition until it expires.
+ */
```
Ephem, line 316: This is a great comment 👍

Ephem (approved):
I don't have a ton of context on the protect stuff and this is a lot to wrap ones head around, but I think I've managed to form a decent mental model.

I wont pretend I understand all the details here, but I've focused especially on understanding and verifying the mechanisms that makes sure existing clients aren't affected, and that seems 👌. Also tried it in our dashboard, no issues.

The approach seems good in general too, I see no obvious blockers. There's a bit of overlap with how we refresh session tokens which has been a thorny area historically, but this thankfully has a bit less complexity.

.changeset/lucky-pandas-observe.md
zourzouvillys: Agreed — rewritten as what the release does rather than how. Two sentences: what it improves, and that it's inert unless the instance opts in and can never block or delay a sign-in. The placeholder names and the status list are gone; they're implementation detail and this repo is public.

packages/clerk-js/src/core/protectSession.ts
zourzouvillys: Ha, fair. Reworded to "an instance using none of them is unaffected", and did the same at the one other site that had it.

packages/clerk-js/src/core/protectSession.ts
zourzouvillys: Good catch, and it's real. A classic inline script executes during `appendChild` and fires neither `load` nor `error`, so this would have waited out the whole deadline, reported `timeout`, and discarded a token the script body had already assigned.

Worth flagging one wrinkle, because the obvious fix is wrong: it isn't "no `src` ⇒ no event". An inline **module** does fire `load` — it evaluates off a module graph the browser still has to fetch — so short-circuiting on `src` alone would settle before the module body ran and report no token instead. The guard now resolves early only for an element that fetches nothing *and* isn't an inline module. Both paths have a test, and the fixtures moved to `type=module` to match what's actually served.

packages/clerk-js/src/core/protectSession.ts
zourzouvillys: Correct, and worth having written down. Two things keep it off the common path: acquisition is prefetched at `Clerk.load()` rather than at sign-in, so a second request in the same flow reads the already-shared store rather than starting a run; and once an attempt settles without producing a token there's a cooldown before a fresh one is allowed, so back-to-back requests can't each pay a full deadline.

The worst case is still reachable — first load, acquisition failing, two requests far enough apart to clear the cooldown. Accepting it: the deadline is a ceiling on a request that would proceed anyway, and it never fails the sign-in.

packages/clerk-js/src/core/protect.ts
zourzouvillys: You were right that the type and the code disagreed, and chasing it turned up more than expected: `protect_config.id` isn't sent at all. So `instanceId` was `undefined` on every load — the storage/lock key suffix never applied, and `{instance_id}` was never substituted, just left verbatim in the URL.

Resolved by removing it from the client rather than validating it: the instance id is the server's to place into the config it serves, not something the SDK interpolates. `ProtectSession` no longer takes it, `{instance_id}` is out of the placeholder set, and the store is explicitly one per origin. `ProtectConfigResource` also stops narrowing `id` to required — that narrowing was the mismatch you spotted. The JSON type keeps it, since `ClerkResourceJSON` requires an id of every resource (there's a TODO there asking the same question you did).

One consequence worth stating plainly: per-instance store namespacing is gone. It never actually applied, so nothing changes in practice, and the token names the instance that minted it and is verified server-side — two instances on one origin costs a rejected token, never one honoured for the wrong instance.

PR #9303, opened 2026-07-31
test(expo): Restore green expo native Maestro flows and enforce them

wobsoriano:
## Description

Android Maestro e2e had failed on every run since the switch to maestro-runner (#9264) (maestro open source fork). This restores the suite to green on both platforms and removes the burn-in `continue-on-error`, so e2e failures now fail the job.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


integration/templates/expo-native/plugins/withClerkOkHttpAlignment.js
wobsoriano, line 1: clerk-android is built against OkHttp 5, while React Native pins apps to OkHttp 4. That mismatch causes the native instance config fetch to fail at runtime, so sign-in never loads.

This CI fixture forces OkHttp 5, along with the last `okhttp-urlconnection` alpha that still ships `JavaNetCookieJar`, to match clerk-android. This is a known ecosystem issue. Expo hit the same problem (expo/expo#44848), and the same workaround is documented on StackOverflow: https://stackoverflow.com/questions/72885577/how-to-implement-okhttp-5-0-0-in-react-native-module

This only affects the CI fixture. Consumer apps are unchanged.

swolfand (approved):

PR #9304, opened 2026-07-31
feat(ui): grow the Mosaic Button touch target without growing the button

maxyinger:
## Description

Under a coarse pointer `Button` floored at `--cl-target-coarse` via `min-height`, which made every
button visibly taller than its `size`. It now grows an invisible `::after` overlay instead — the hit
area reaches the target, the rendered box stays put. Default shape grows on the block axis only;
square and circle grow on both.

`touchTarget={false}` opts out where neighbors sit close enough that the overlays would overlap.
`variant='link'` never takes one.

`overflow: hidden` comes off `base`: the label span carries its own, and the overlay has to escape
the box to be hit-testable.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/rare-pumas-repeat.md
+++ b/.changeset/rare-pumas-repeat.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': minor
+---
+
+Mosaic `Button` no longer grows on touch devices to reach the recommended hit target. It keeps the size its `size` prop gives it and extends only the area that responds to a tap. Add `touchTarget={false}` to turn the larger tap area off where buttons sit close enough that theirs would overlap.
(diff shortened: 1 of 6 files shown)


.changeset/rare-pumas-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/ui': minor
+---
+
+Mosaic `Button` no longer grows on touch devices to reach the recommended hit target. It keeps the size its `size` prop gives it and extends only the area that responds to a tap. Add `touchTarget={false}` to turn the larger tap area off where buttons sit close enough that theirs would overlap.
```
alexcarpenter:
```suggestion
---
---
```

alexcarpenter (approved):

PR #9307, opened 2026-07-31
feat(ui): fade Mosaic Button icons until hover

alexcarpenter:
## Description

An icon inside a `Button` now rests a step below its label and comes up to it on hover, so the label leads and the icon reads as supporting.

Preview:
- https://swingset-git-carp-mosaic-button-icon-color.clerkstage.dev/components/button
- https://swingset-git-carp-mosaic-button-icon-color.clerkstage.dev/components/menu

Each `variant × color` cell sets `--_cl-icon-color` to an opaque faded form of that cell's own text color, and lifts it to the full text color on hover and while a disclosure trigger is open. `Icon` reads the var with a `currentColor` fallback, so an icon outside a button is unchanged. StyleX can't emit a descendant rule, so the value crosses the element boundary as a custom property; it sits per cell rather than once in `base` because StyleX resolves a property to the last style that declares it, and a cell would otherwise drop the base hover branch wholesale.

The faded steps are opaque rather than an alpha of the label color: a translucent icon picks up whatever sits behind the button and drifts per surface. The achromatic foregrounds share the house faded gray; the two that carry hue fade toward the tint that belongs to them; the light-on-fill pairs fade toward their own fill, which is the only backdrop they can sit on.

`MenuTrigger` was rendering the raw registry glyph rather than the `Icon` component, so its ellipsis stayed full-strength next to icons that faded. It now renders `<Icon name='ellipsis' />`, which also makes that glyph swappable via `appearance.icons`. Its `triggerIcon` atom is gone — it was `space['4']` square, exactly what `Icon`'s default `md` size gives.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


maxyinger (approved):
I must not be using a custom svg in this story or something. mind updating it to use icon so it picks up on these style changes. other than that looks great

Image

PR #9308, opened 2026-07-31
fix(clerk-js): complete the Safari ITP touch hop in redirect

dmoerner:
The `else if (redirectUrl)` branch navigated to `/v1/client/touch` and then immediately navigated again to the undecorated redirect URL. Where both resolve to a hard navigation, the second assignment supersedes the first and the touch request is aborted, so the client cookie is never extended past ITP's 7 day cap.

This was a regression that was accidentally introduced in https://github.com/clerk/javascript/pull/6486/changes#diff-1952d6b8ac4b5495e1dba57c32bff9ce72d43f83595c1bc8532a76cf31feb753L1314 11 months ago, when an `else` in a nested control flow was accidentally dropped. I suspect it was not reported because Safari use is less common and the sign out would have only occurred after 7 days.

Route the redirect through the shared #decorateUrlWithTouch helper instead, which returns the URL unchanged when the client is not eligible.

The existing test asserted only navigate.mock.calls[0][0], so it stayed green across the regression; it now also asserts navigate is called exactly once.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/smart-jars-repeat.md
+++ b/.changeset/smart-jars-repeat.md
@@ -1,0 +1,9 @@
+---
+'@clerk/clerk-js': patch
+---
+
+Complete the Safari ITP cookie refresh when `setActive({ redirectUrl })` navigates.
+
+Safari's ITP caps the client cookie at 7 days when it is re-issued from a fetch, so `setActive()` routes its redirect through `/v1/client/touch` to restore the full lifetime. That navigation was immediately followed by a second one to the undecorated redirect URL, which superseded it and aborted the touch request before it completed, leaving the cookie capped.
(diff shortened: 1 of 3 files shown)


wobsoriano (approved):
great catch!!

dmoerner:
> great catch!!

Definitely partial credit goes to Dylan who in a previous PR told me to reduce code duplication, and then when I added this `#decorateUrlWithTouch` helper and was looking for places to use it, found this no-op!

PR #9310, opened 2026-07-31
feat(headless): hold Popover contents while it closes

alexcarpenter:
## Description

Stacked on #9184 (`carp/account-button-switcher`).

A popover's popup outlives `open` by the length of its exit animation, and whatever closed it has usually changed the data behind it — switching account or organization from the UserButton closes the menu and you watch its contents swap while it animates out.

`Popover.Popup` now wraps its children in a new `Freeze` utility (`@clerk/headless/utils`): a Suspense boundary plus a sibling that throws a never-settling promise, so React keeps rendering the subtree but holds the commit. The DOM keeps painting its last frame until `frozen` flips back off, at which point the pending render commits. The popup element itself is outside the freeze, so `data-closed` / `data-ending-style` still land and the exit animation is unchanged.

The technique is from https://barvian.me/react-exit-animations, adapted for React 18, which this package still supports (`react: ^18.0.0 || ~19.0.3 || …`):

- `React.use(infinitePromise)` → `throw never`.
- React 19.2 Fragment Refs → a `display: contents` wrapper the component owns. React hides a suspended boundary's host children with `display: none !important`; an insertion effect restores `display: contents` on the same commit, before paint. React also detaches refs of hidden children, so the wrapper's node is captured through a ref callback that ignores `null` rather than read back from a plain ref.

Trade-offs, both documented in the popover README:

- Every popover gains a `display: contents` element between the popup and its content.
- Refs inside popup content are `null` while the popover is closing, until it reopens.

`Freeze` is unit tested. The popover wiring is not covered end to end: happy-dom has no animations, so `mounted` flips to `false` immediately on close and there is no exit window to assert against.

BEFORE

https://github.com/user-attachments/assets/b9763b75-9bcd-48c4-8e68-f5759dd5a175

AFTER

https://github.com/user-attachments/assets/2fb6c4ad-b786-4dbc-bec5-eed756af5644

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


austincalvelage (approved):
Such a clever trick

PR #9313, opened 2026-08-01
feat(clerk-js,shared,react): supply a Protect assertion from the application

zourzouvillys:
## Description

A **Protect assertion** is a short-lived, signed token an application mints from its own backend with the Clerk Backend API, carrying key/value pairs its Protect rules can read.

Until now the only way to deliver one to Protect was the `__clerk_protect_assertion` cookie. That requires the app and the Frontend API to be same-site — true with a production CNAME setup, but **not on development instances**, where it is cross-site. It also means an application that has just minted a token has to work out a cookie domain scope in order to hand it over.

The token can now be given to Clerk directly:

```ts
// A token you already have.
Clerk.load({ protectAssertion: token });

// A function, re-read for each request.
Clerk.load({ protectAssertion: () => sessionStorage.getItem('protect_assertion') ?? undefined });

// Or set it later, once your app has fetched one.
clerk.setProtectAssertion(token);
```

Clerk attaches it to sign-in and sign-up requests. The cookie continues to work unchanged; where both are present, the value given to the SDK wins.

### Why a function as well as a string

Assertions are short-lived by design and a single-page app routinely outlives one. A string captured at configuration time silently stops applying at expiry; a function is re-read per request, so a token refreshed in the background takes effect without re-configuring Clerk.

### Why a body param and not a header

`fapiClient` already tunnels `PATCH`/`DELETE` through a `_method` query param to keep requests CORS-simple, because the preflight otherwise breaks cookie dropping in Safari (see the existing comments in that file). A custom request header would reintroduce exactly that. The param rides the form-encoded body instead, on the requests Protect gates and nothing else.

The merge has to happen inside `request()`, before the body is stringified — `onBeforeRequest` callbacks run *after* stringification, so a callback cannot add a body param.

## Behaviour worth reviewing

**Nothing here can fail a sign-in.** That is the invariant the tests are built around:

- a resolver that throws, rejects, or returns anything other than a non-empty string → no assertion, one warning, request proceeds
- a body that is not a plain object (`FormData`, a string, a `Blob`) is left completely alone — spreading it would *discard* the caller's payload rather than add to it
- with nothing configured the request is byte-for-byte what it was before, which is why the params resolve to `undefined` rather than `{}`

**`setProtectAssertion` tracks that it was called, separately from the value.** Without that flag, clearing with `undefined` would silently fall back to the `protectAssertion` option, and a call made *before* `Clerk.load()` would be overwritten by it.

**`IsomorphicClerk`** implements it through the existing `premountMethodCalls` map, which is keyed by method name — so a second call before load replaces the first, which is the semantics a setter wants.

## Checks

- `vitest run src/core` in `@clerk/clerk-js`: **790 passed**, including 16 new for the resolver and 12 new for the request wiring
- `@clerk/react`: 203 passed. One suite (`ClerkProvider.test.tsx`) fails to import in my local worktree because `@clerk/localizations` isn't built there — unrelated to this change
- `build:declarations` clean for `@clerk/clerk-js`; `tsc --noEmit` clean for `@clerk/react`
- `eslint` clean on every file touched
- **bundlewatch**: `clerk.native.js` gains ~0.2KB gzipped and crossed its 74KB ceiling, so it is raised to 75KB — the same headroom the other entries carry. Everything else passes unchanged.

## Note on ordering

This depends on a matching server-side change that must be **deployed first** — until then the new param is not accepted and would be rejected. Happy to hold this until that has shipped.

There is also an in-flight PR touching `fapiClient`'s request body for a separate Protect feature. Both introduce the same `getProtectParams` seam; whichever lands first, the other is a trivial resolve (keep one copy of the hook).


zourzouvillys:
### On the two `nextjs` integration failures

First run: **145 passed, 1 failed, 1 flaky, 2 skipped**. Both affected specs are session-task navigation (`session-tasks-multi-session`, `session-tasks-sign-in-reset-password`) — Playwright itself classified the second as flaky in that same run. I've re-run the failed jobs rather than assert a cause either way.

Worth stating the one way this change *could* plausibly have touched them, and why it can't:

`request()` now `await`s the Protect params before building the URL, which adds a microtask on gated requests. The only value read after that point is `options.getSessionId()`, used to append `_clerk_session_id` — and that is skipped for any path starting with `/client`:

```ts
const unauthorizedPathPrefixes = ['/client', '/waitlist'];
// ...
if (path && !unauthorizedPathPrefixes.some(p => path.startsWith(p)) && sessionId) {
  searchParams.append('_clerk_session_id', sessionId);
}
```

Both gated prefixes (`/client/sign_ins`, `/client/sign_ups`) start with `/client`, so the one thing read after the new await is unused on precisely the paths that have it. The added microtask cannot change the request that goes out.

Beyond those paths the hook never runs: `isProtectGatedRequest` returns `false` for `GET` and for any non-matching path, so every other request takes the identical code path it did before.

wobsoriano (approved):
pushed a bundle config fix and cleaned up comments ff. agents rule. looks good overall 👍🏼

PR #9314, opened 2026-08-03
fix(react): make ClerkProvider publishableKey optional

SarahSoutoul:
## Description

[clerk/javascript#7634](https://github.com/clerk/javascript/pull/7634) added the runtime fallback that allows `@clerk/react` to read `VITE_CLERK_PUBLISHABLE_KEY` or `CLERK_PUBLISHABLE_KEY` when `publishableKey` isn't passed to `<ClerkProvider>`. That PR also updated the React Vite templates to omit the explicit prop.

However, the corresponding public type was not updated. `ClerkProviderProps` continued to inherit the required `publishableKey: string` field from `IsomorphicClerkOptions`.

As a result, applications using the environment-variable fallback work correctly at runtime, but TypeScript incorrectly reports that the `publishableKey` prop is missing.

This PR makes `publishableKey` optional specifically in `ClerkProviderProps`, aligning the public type with the runtime behavior introduced by #7634. It also updates the type tests to cover using `<ClerkProvider>` without an explicit key.

The focused React tests pass with no type errors, and the generated declaration exposes `publishableKey?: string`.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


manovotny:
Pushed a fix directly in 7fccba0c4.

Making `publishableKey` optional in `@clerk/react` leaked into `@clerk/chrome-extension`, whose provider intersects the base props without redeclaring the field. Extensions have no env-var fallback and assert the key at runtime, so a user could drop `publishableKey`, get no type error, and hit a runtime throw.

- Omit and redeclare `publishableKey: string` on `ChromeExtensionClerkProviderProps` so it stays required — the same pattern the Next.js, React Router, TanStack, Expo, and Electron wrappers already use.
- Added a type test so it can't silently regress.
- No changeset: `@clerk/chrome-extension` already gets a patch bump as a dependent of this PR's `@clerk/react` change, and the resulting type matches `main`.

manovotny (approved):
This LGTM, but if someone on the JS/SDK team could look, that'd be great!

wobsoriano (approved):
thanks for catching this

PR #9320, opened 2026-08-04
feat(ui): wire Mosaic Button to the headless Button with focusableWhenDisabled

alexcarpenter:
## Description

> [!NOTE]
> Stacked on #9319. Review that one first; this PR's diff is against it.

Mosaic's `Button` now renders the headless `Button` instead of a bare `<button>`, and forwards a new `focusableWhenDisabled` prop:

```tsx
<Button disabled={submitting} focusableWhenDisabled type='submit'>Save</Button>
```

A button that disables itself mid-interaction drops focus to the body, and the user loses their place on the page. With the prop set, the button keeps its position in the tab order and is marked `aria-disabled` instead. `type='button'` is dropped from the call site since the headless component already defaults it.

### Regating the interactive states

The variant styles gated every hover/active/`[data-open]` state on `:enabled`. That gate stops holding under `focusableWhenDisabled`, which by design omits the `disabled` attribute — a disabled button would have kept lighting up on hover. The button is also deliberately left hit-testable while disabled, so `cursor: not-allowed` renders and a wrapping tooltip still receives the pointer, which rules out suppressing the states with `pointer-events`.

All 52 selectors are regated on `:not([data-disabled])`. `themeProps` reflects `data-disabled` from the prop, so it is present whether or not the native attribute is, and one selector covers both. Compiled output is equivalent: 27 deduped atoms, no `:enabled` left in `styles.css`.

Mosaic components are internal — only `styles.css` is exported from the package — so there is no public API change and no call site opts into the prop yet. Hence the empty changeset.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


maxyinger (approved):
🚀

PR #9328, opened 2026-08-04
fix(ui): Fix email link race with sign up if missing

dmoerner:
Follow up fix for https://github.com/clerk/javascript/pull/7928 after additional testing in staging.

With development instances, there can be races between the polling and new email link tabs when processing a sign up through the sign-in-or-up flow with strict enumeration protection enabled. Late in the review process on that PR, I added an attempt to avoid these kinds of races, but it was a regression on dev instances when the sign up completes without extra requirements. https://github.com/clerk/javascript/pull/7928/changes/ec83f7cfdcc459246de799c519e3a0ed1226f1d5

Revert that fix, and also add improved redirect handling in the case where the flow completes immediately.

This has the cost that if someone closes the original tab, they are unable to complete the sign up. (That was what we were trying to fix during PR review.) However, there is no easy solution to that problem (and the solution that we attempted to implement to fix it was not successful.) Since email links are so little used, I think we should accept that as a cost for now. If someone closes the original tab, they will just need to start the flow again.

This was tested locally against staging and I was no longer able to produce races with dev instances.

End to end validation I performed on this branch:

- Email link sign in with existing user
- Email link sign in transfer to sign up missing requirements
- Email link sign in transfer to sign up complete (this was the regression)
- Email link sign up
- Email code sign in transfer to sign up missing requirements
- Email code sign in with existing user

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/fifty-ads-smell.md
+++ b/.changeset/fifty-ads-smell.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+fix(ui): Avoid races between email link tabs when using sign up if missing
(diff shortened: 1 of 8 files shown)


wobsoriano (approved):
i got you

PR #9332, opened 2026-08-04
fix(ui): Give each Mosaic component a minimal CSS reset

alexcarpenter:
## Description

`MosaicProvider` injects a reset keyed on `data-cl-slot`, but only the slot-recipe components (`dialog`, `tabs`, `input`) emit that attribute. The StyleX components style through `themeProps` + `stylex.props` and never carried it, so they inherited user-agent defaults — most visibly `<Text>` and `<Heading>`, which rendered with the browser's paragraph and heading margins. Meanwhile `boxSizing` and `fontFamily: inherit` were hand-copied into six separate `.styles.ts` files.

An injected stylesheet can't fix this: the StyleX tier builds through the deliberately Emotion-free `styles/index.ts` barrel, which has no provider to inject anything. So the reset has to be a StyleX atom.

`components/reset.styles.ts` exports `reset.base`, composed as the first argument of every `stylex.props(...)` call — the same way `typography.styles.ts` already shares `sizes` / `colors` / `truncationStyles`:

```ts
export const reset = stylex.create({
  base: {
    boxSizing: 'border-box',
    color: 'inherit',
    fontFamily: 'inherit',
    fontSize: 'inherit',
    fontWeight: 'inherit',
    lineHeight: 'inherit',
    margin: 0,
    padding: 0,
  },
});
```

`margin` and `padding` are shorthands on purpose. StyleX sorts shorthands into an earlier priority layer than longhands, so a component's own `paddingInline` outranks the reset no matter how the arguments are ordered — visible in the built `styles.css`:

```
margin: 0        → priority2   (reset, shorthand)
padding-inline:  → priority3   (components, longhand)
```

The `inherit` declarations tie with a component's own value, so those do depend on the reset going first.

The declarations the reset now owns are deleted from `button`, `badge`, `avatar`, `card`, `item` and `menu`. Each one still appears exactly once in the emitted CSS.

Two things are deliberately left alone. `box.tsx` keeps its local plain-object reset and `input.tsx` keeps its own `inherit` declarations — both sit on the Emotion path and can't consume a StyleX atom, and `MOSAIC_RESET_CSS` remains the reset for that tier. `Popover.Trigger` / `Close` / `Title` / `Description` are pure pass-throughs with no styles at all; giving a raw `<button>` `padding: 0` while leaving its native chrome would look worse than leaving it, and that needs a `control` atom this PR doesn't introduce.

**This is a visual change.** `<Text>` and `<Heading>` no longer carry the browser's default margins, so they now sit flush with their container. Layouts that leaned on that spacing need to set it explicitly.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.changeset/mosaic-component-reset.md
```diff
@@ -0,0 +1,7 @@
+---
+'@clerk/ui': patch
+---
+
+Mosaic components now carry a minimal reset, so they no longer inherit user-agent margins, padding, or fonts. Previously only the components built on slot recipes were reset; the rest picked up browser defaults, most visibly `<Text>` and `<Heading>`, which rendered with the browser's paragraph and heading margins.
+
+Text and headings now sit flush with their container. A layout that relied on the browser's default spacing around them needs to set that spacing explicitly.
```
alexcarpenter:
```suggestion
---
---
```

austincalvelage (approved):
Looks good!

PR #9342, opened 2026-08-05
feat(ui): add Mosaic SubmitButton

alexcarpenter:
## Description

Preview: https://swingset-git-carp-mosaic-submit-button.clerkstage.dev/components/button

Adds `SubmitButton` to Mosaic: a `Button` that defaults `type` to `submit` and takes an `isPending` prop.

While pending, the label fades to zero opacity and a spinner centers over it. The label stays mounted rather than being swapped out, so the button holds the width its content gives it and nothing around it reflows when the state flips. Every child sits in one box, so an icon fades with its label instead of hanging on beside the spinner. Text runs still get their own truncatable box inside that wrapper.

The pending *state* is immediate — `aria-busy`, `aria-disabled`, `data-pending`, pointer events dropped, press cancelled — but the spinner waits 300ms before it's drawn and stays up at least 200ms once it is, so an action that resolves quickly never flashes one. `spinDelay` moves both numbers, or removes the wait entirely with `{ delay: 0 }`.

Accessibility:

- The indicator enters the accessibility tree as an indeterminate `progressbar` the moment `isPending` flips, including during the delay when it's mounted but not yet drawn. That's why the delay is `opacity` and not conditional rendering — not rendering it removes it from the tree just as `display: none` would.
- The button goes inert via `aria-disabled` rather than the `disabled` attribute, so it stays focusable and doesn't drop focus mid-action just as the progressbar is announced. The press is cancelled in JS instead. Marked with a TODO to fold into the headless button's `focusableWhenDisabled` from #9319 / #9320 once those land.
- The progressbar is named in its own right via `pendingLabel` rather than folded into the button's name: `progressbar` is a range role, so name computation reads its value — absent, since it's indeterminate — over its label, and a descendant one contributes nothing to the button above it.

`Button` gates its hover and pressed fills on `:enabled`, which a pending button still is. Pointer events are dropped for the pointer half; the 9 active-fill selectors now exclude `[data-pending]` for the keyboard half, since a focused button takes `:active` from space and enter with no pointer involved.

`Spinner` and `useSpinDelay` come over from `carp/account-button-switcher`. One divergence to reconcile when that branch merges: `spinner.styles.ts` derives both colors from `currentColor` instead of the fixed `--cl-color-card-foreground` pair, which would be invisible on a `filled-primary` button.

Swingset's button page documents all of it, with a pressable example and a slow/fast pair showing the delay.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/components/button/submit-button.tsx
```diff
@@ -0,0 +1,123 @@
...
+  { isPending = false, pendingLabel = 'pending', size = 'md', spinDelay, className, children, onClick, ...rest },
+  ref,
+) {
+  const { delay = DEFAULT_SPIN_DELAY, minDuration } = spinDelay ?? {};
+
+  // `isPending` drives the semantics, this drives the pixels. The split is deliberate: the button
+  // has to go inert and start announcing the moment the action does, or a fast action gets
+  // submitted twice and assistive tech misses it — but drawing a spinner that fast only produces
+  // a flash, so the visual waits out the delay and then sticks around long enough to be read.
+  const showPending = useSpinDelay(isPending || null, { delay, minDuration }) !== null;
```
austincalvelage, line 70: Love this

packages/ui/src/mosaic/components/button/submit-button.tsx
```diff
@@ -0,0 +1,123 @@
...
+// The spinner scale stops at `md`, and a `lg` button's label is only one step up, so both take
+// the larger ring rather than `lg` asking for one the spinner cannot render.
+const spinnerSizes = { sm: 'sm', md: 'md', lg: 'md' } as const;
+
+// Long enough that a request served from cache or a local mutation never draws a spinner, short
+// enough that a press which is going to take a while doesn't sit there looking ignored. Set here
+// rather than on `useSpinDelay` itself: a button is pressed and watched, so it wants a tighter
+// window than a hook shared with background loads. `minDuration` has no such tension, so it takes
+// the hook's default.
+const DEFAULT_SPIN_DELAY = 300;
```
austincalvelage, line 44: I am curious if this delay could be a little shorter, but I think we will discover that through usage

austincalvelage (approved):

packages/ui/src/mosaic/components/button/submit-button.tsx
alexcarpenter: yeah, I am inclined to reduce it as well! will tweak values once we start implemnting in the user button 👍🏼

PR #4789, opened 2024-12-16
feat(types): Add `oidcPrompt` and `oidcLoginHint`

octoper:
## Description

This PR introduces two newly added params: `oidcLoginHint` and `oidcPrompt`

## Checklist

- [ ] `pnpm test` runs as expected.
- [ ] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #9354, opened 2026-08-06
fix(js): Send oidcPrompt and oidcLoginHint in external account requests

### LINKED ISSUE #9177: External account flows silently drop oidcPrompt despite public type support
### Preliminary checks

- [x] I reviewed the Clerk documentation.
- [x] I searched the `clerk/javascript` issue tracker, including open and closed issues.
- [x] This is a JavaScript SDK bug report rather than a general support question.

### Summary

The public JavaScript types for connected-account OAuth flows accept `oidcPrompt`, but ClerkJS silently drops it when constructing the Frontend API request.

This affects both:

- `ExternalAccountResource.reauthorize(...)`
- `UserResource.createExternalAccount(...)`

For Google integrations that need background access, this means `oidcPrompt: 'consent'` does not reach the OAuth authorization request. Clerk/Google can return a short-lived access token without a refresh token when an existing Google grant is reused. The connection initially works, then `getUserOauthAccessToken` can fail after the access token expires with `oauth_missing_refresh_token`.

### Reproduction

Call reauthorization on an existing Google external account:

```ts
await externalAccount.reauthorize({
  redirectUrl: '/oauth/callback',
  additionalScopes: [
    'https://www.googleapis.com/auth/calendar.readonly',
  ],
  oidcPrompt: 'consent',
});
```

Inspect the request to the external-account reauthorization endpoint.

**Expected request body**

```json
{
  "redirect_url": "/oauth/callback",
  "additional_scope": [
    "https://www.googleapis.com/auth/calendar.readonly"
  ],
  "oidc_prompt": "consent"
}
```

**Actual request body**

```json
{
  "redirect_url": "/oauth/callback",
  "additional_scope": [
    "https://www.googleapis.com/auth/calendar.readonly"
  ]
}
```

The generated Google authorization URL consequently uses Clerk's/default provider prompt behavior rather than the requested `prompt=consent`.

The same omission occurs when passing `oidcPrompt` to `user.createExternalAccount(...)`.

### Source-level evidence on current main

The parameter is part of the public reauthorization type:

https://github.com/clerk/javascript/blob/bcbdda6d7d6c6e12cf33febe17fd148c69788716/packages/shared/src/types/externalAccount.ts#L6-L10

But the implementation only reads and serializes `additionalScopes` and `redirectUrl`:

https://github.com/clerk/javascript/blob/bcbdda6d7d6c6e12cf33febe17fd148c69788716/packages/clerk-js/src/core/resources/ExternalAccount.ts#L37-L43

Likewise, `CreateExternalAccountParams` declares `oidcPrompt` and `oidcLoginHint`:

https://github.com/clerk/javascript/blob/bcbdda6d7d6c6e12cf33febe17fd148c69788716/packages/shared/src/types/user.ts#L397-L403

But `User.createExternalAccount` omits both from its destructuring and request body:

https://github.com/clerk/javascript/blob/bcbdda6d7d6c6e12cf33febe17fd148c69788716/packages/clerk-js/src/core/resources/User.ts#L165-L177

The original type PR added these fields to type files, but did not update the ClerkJS resource implementations:

https://github.com/clerk/javascript/pull/4789

For comparison, Clerk's iOS SDK now sends `oidc_prompt` to the same external-account creation and reauthorization endpoints:

https://github.com/clerk/clerk-ios/blob/d41365666ac3be138737a57b33b30b6291b5c65b/Sources/ClerkKit/Domains/User/ExternalAccount/ExternalAccountService.swift#L21-L41

https://github.com/clerk/clerk-ios/blob/d41365666ac3be138737a57b33b30b6291b5c65b/Sources/ClerkKit/Domains/User/UserService.swift#L117-L138

### Expected behavior

`oidcPrompt` should be serialized as `oidc_prompt` for external-account creation and reauthorization. If `oidcLoginHint` is supported by these endpoints, it should likewise be serialized; otherwise it should not be exposed by these public parameter types.

### Version check

Observed in an app using:

```text
@clerk/nextjs 6.38.2
```

Confirmed in the latest published ClerkJS package at the time of filing:

```text
@clerk/clerk-js 6.25.4
```

The minified `6.25.4` bundle contains the same implementation:

```js
reauthorize=e=>{
  let{additionalScopes:t,redirectUrl:a}=e||{};
  return this._basePatch({
    action:"reauthorize",
    body:{additional_scope:t,redirect_url:a}
  })
}
```

It is also present on current `main` at commit `bcbdda6d7d6c6e12cf33febe17fd148c69788716`.

### Environment

```text
System:
  OS: macOS 26.5.1
  CPU: arm64 Apple M4
Binaries:
  Node: 25.2.1
  npm: 11.6.2
Browsers:
  Chrome: 149.0.7827.201
npmPackages:
  @clerk/nextjs: 6.38.2
```

### Related issue

https://github.com/clerk/javascript/issues/1827 reports a similar “works initially, then cannot refresh the Google access token” symptom, but it does not identify this parameter-serialization bug and was closed for inactivity in 2023.

wobsoriano:
## Description

The public types for `user.createExternalAccount()` and `externalAccount.reauthorize()` have accepted `oidcPrompt` and `oidcLoginHint` since they were added in #4789, but the ClerkJS resource implementations never read them, so both parameters were silently dropped from the Frontend API request. FAPI supports `oidc_prompt` and `oidc_login_hint` on both `POST /v1/me/external_accounts` and `PATCH /v1/me/external_accounts/{id}/reauthorize`.

This meant `oidcPrompt: 'consent'` never reached the OAuth authorization request. For Google connections needing offline access, an existing grant could be reused and return an access token without a refresh token, causing `getUserOauthAccessToken` to later fail with `oauth_missing_refresh_token`.

Fixes #9177

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


HAYASHIDAKodai:
@wobsoriano 
We're hitting this in production and can confirm the behavior described in [#9177](https://github.com/clerk/javascript/issues/9177) is still present as of clerk-js 6.30.1.

What we observed:

We call ExternalAccount.reauthorize({ additionalScopes: ['https://www.googleapis.com/auth/calendar.readonly'], oidcPrompt: 'consent', redirectUrl: ... }). This type-checks and our unit tests assert the param is passed.
Watching the network, the SDK sends PATCH /v1/me/external_accounts/{id}/reauthorize, but the Google authorize URL generated from it still contains prompt=select_account — prompt=consent never makes it into the OAuth request.

Why this hurts more than it looks:

Because prompt=consent never reaches Google, the consent screen is skipped on reauthorization, so Google never issues a new refresh token covering the incrementally added scope (refresh tokens are only issued on consent, and their scopes are fixed at issuance). Clerk then keeps refreshing with the original sign-in refresh token (base scopes only). The result: the access token works right after reauthorization, but ~1 hour later (access token expiry) the Backend API token endpoint returns a token whose actual grant lacks the added scope — while the reported scopes metadata still lists it — and Google returns 403 insufficientPermissions. Users are forced to "reconnect" every hour, and the reconnect can never fix itself.

This effectively makes incremental OAuth scopes unusable on web until this PR lands. Would love to see it merged — happy to test a canary. Thanks!

jeremy-clerk (approved):

PR #9355, opened 2026-08-06
fix(ui): hide unavailable alternative verification methods

tmilewski:
## Summary

- hide “Use another method” during Device Trust verification when only one factor is available
- apply the same behavior to regular second-factor verification
- add status-accurate component coverage and a Device Trust integration assertion

## Root cause

The Device Trust and second-factor containers always passed an alternative-method callback to their verification cards. The cards use the callback’s presence to decide whether to render the action, so the link appeared even when there was no alternative factor.

## Impact

Users no longer see a dead-end “Use another method” action when only one verification factor is available. The action remains visible when multiple factors exist.

## Validation

- `pnpm --filter @clerk/ui exec vitest run src/components/SignIn/__tests__/SignInClientTrust.test.tsx src/components/SignIn/__tests__/SignInFactorTwo.test.tsx`
- `E2E_APP_ID=next.appRouter.withNeedsClientTrust pnpm playwright test --config integration/playwright.config.ts integration/tests/client-trust.test.ts`
- `pnpm --filter @clerk/ui format:check`
- targeted ESLint (0 errors)
- `pnpm turbo build --filter=@clerk/ui`


tmilewski:
@codex review

wobsoriano (approved):
🫰

PR #9356, opened 2026-08-06
feat(ui): upgrade affiliation domains for SSO

NicolasLopes7:
## Summary
- surface existing affiliation-verified organization domains in Configure SSO
- let admins start TXT ownership verification for those domains
- attach only ownership-verified domains to enterprise connections


packages/ui/src/components/ConfigureSSO/hooks/useOrganizationEnterpriseConnection.ts
```diff
@@ -243,7 +288,7 @@ export const useOrganizationEnterpriseConnection = (): UseOrganizationEnterprise
         await deleteEnterpriseConnection(enterpriseConnection.id);
       }
 
-      const domains = enterpriseConnection?.domains ?? organizationDomains?.map(domain => domain.name);
+      const domains = enterpriseConnection?.domains ?? verifiedOwnershipDomains?.map(domain => domain.name);
```
gabrielmeloc22, line 291: should it be filtering the non-verified ones ? not sure if it would make much difference to filter it out, since they'd need to prove it anyways, but just curious if it was intentional

Image

gabrielmeloc22 (approved):
LGTM

some AI findings ⬇️ 

## 🤖 AI Review Findings

Reviewed the ownership-upgrade changes (`@clerk/ui` + `@clerk/shared`). Core gating logic is correct and well-tested — a few things worth a look before merge.

### Main: two-hook fetch looks redundant with the new `createEnrollmentMode` param

`useOrganizationEnterpriseConnection` instantiates `__internal_useOrganizationDomains` **twice** — one unfiltered (all domains) and one filtered to `enterprise_sso` — then merges + de-dupes by `id`. But `enterpriseSSODomains` is a strict subset of the unfiltered list, so the merge is a no-op for reads; the second hook only supplies `createDomain` + `revalidate`, which is exactly what the newly-added `createEnrollmentMode` param is for. A single hook should suffice:

```ts
const { data: organizationDomains, createDomain, prepareOwnershipVerification,
        attemptOwnershipVerification, revalidate, isLoading }
  = __internal_useOrganizationDomains({
      createEnrollmentMode: 'enterprise_sso',
      onOwnershipVerified,
    });
```

Costs of keeping two hooks:
- **Duplicate `getDomains` fetch** where one result is a subset of the other.
- **Duplicate ownership polling (medium):** `enterprise_sso` domains live in both hooks, so both run the `attemptOwnershipVerification` poll and both fire `onOwnershipVerified`. Not a correctness bug (`handleDomainOwnershipVerified` guards on `hasNewDomains`), but it can issue concurrent `attemptOwnershipVerification` / `updateEnterpriseConnection` calls for the same domains.

If a single unfiltered hook genuinely doesn't work here, a comment explaining why would help — the merge currently reads as scaffolding left over from before `createEnrollmentMode` existed.

### Smaller points
- **Remove path for a domain with both affiliation + ownership verification** (`OrganizationDomainsStep.tsx` `handleRemoveDomain`): it detaches from the connection and skips `domain.delete()`, but keeps `ownershipVerification`. Since `createEnterpriseConnection` auto-includes all `verifiedOwnershipDomains`, that domain would be silently re-attached to any *new* connection. Confirm that's intended — the dialog copy only mentions affiliation being preserved.
- **`prepareOwnershipVerification` wrapper** revalidates only `enterpriseSSODomains`; it works only because the underlying affiliation hook already revalidates its own query. Redundant, not wrong — another artifact of the two-hook split.
- **Merge keeps the last duplicate** (`Array.from(new Map(...).values())`) → the `enterprise_sso` copy wins for shared IDs. Fine if the two copies are always identical; worth a one-line comment on that assumption.

PR #9370, opened 2026-08-09
fix(js): keep native OAuth transport callbacks inside the component router

wobsoriano:
## Description

This PR fixes native OAuth transport flows (e.g. `@clerk/electron`) breaking out of modal components. When a callback needed an intermediate step such as a sign-in to sign-up transfer, the continue step, or an MFA factor, clerk-js navigated the app window to an internal Clerk route like `/CLERK-ROUTER/VIRTUAL/sign-up`. In Electron apps this reloaded the renderer on a route the app does not have, and the leaked URL was later submitted as the completion redirect, which production instances reject with a "Redirect url mismatch" error ([ref](https://clerkinc.slack.com/archives/C068ZP01R7A/p1786222968731609)).

The prebuilt UI now hands its own router to the transport callback via an internal param so those steps render inside the component, and `_authenticateWithTransport` now submits the registered transport callback URL as the completion redirect while the app's real destination is navigated client-side. Both changes are internal and optional, so older UI versions keep the existing behavior.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/clerk.ts
+++ b/packages/clerk-js/src/core/clerk.ts
@@ -2389,11 +2389,14 @@ export class Clerk implements ClerkInterface {
     const signIn = 'identifier' in (signInOrUp || {}) ? (signInOrUp as SignInResource) : _signIn;
     const signUp = 'missingFields' in (signInOrUp || {}) ? (signInOrUp as SignUpResource) : _signUp;
 
-    const resolvedNavigate = customNavigate ?? params.__internal_navigate;
+    // The component router expects raw component-relative paths; buildUrlWithAuth would absolutize
+    // them on development instances and push the navigation back out to the window.
     const navigate = (to: string) =>
-      resolvedNavigate && typeof resolvedNavigate === 'function'
-        ? resolvedNavigate(this.buildUrlWithAuth(to))
-        : this.navigate(this.buildUrlWithAuth(to));
+      customNavigate && typeof customNavigate === 'function'
+        ? customNavigate(this.buildUrlWithAuth(to))
+        : params.__internal_navigate && typeof params.__internal_navigate === 'function'
+          ? params.__internal_navigate(to)
+          : this.navigate(this.buildUrlWithAuth(to));
 
     return this._handleRedirectCallback(params, {
       signUp,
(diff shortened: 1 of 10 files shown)


packages/clerk-js/src/core/clerk.ts
```diff
@@ -2389,9 +2389,10 @@ export class Clerk implements ClerkInterface {
     const signIn = 'identifier' in (signInOrUp || {}) ? (signInOrUp as SignInResource) : _signIn;
     const signUp = 'missingFields' in (signInOrUp || {}) ? (signInOrUp as SignUpResource) : _signUp;
 
+    const resolvedNavigate = customNavigate ?? params.__internal_navigate;
     const navigate = (to: string) =>
-      customNavigate && typeof customNavigate === 'function'
-        ? customNavigate(this.buildUrlWithAuth(to))
+      resolvedNavigate && typeof resolvedNavigate === 'function'
+        ? resolvedNavigate(this.buildUrlWithAuth(to))
```
wobsoriano: backwards compatible

jeremy-clerk: `__internal_navigate` should receive the raw component-relative path here. Native transport reaches this method through `_authenticateWithTransport` → `__internal_handleResourceCallback`; with a development publishable key, `buildUrlWithAuth("continue")` produces a root-level absolute URL such as `http://host/continue`. The virtual router then sees `/continue` as outside `/CLERK-ROUTER/VIRTUAL` and delegates to window-level navigation, so test-instance Electron flows still break out of the modal. Could we keep the existing decorated behavior for `customNavigate`, but pass raw `to` to `params.__internal_navigate`? A regression test should exercise `__internal_handleResourceCallback` with `developmentPublishableKey`; the current test calls `handleRedirectCallback` with a production key, so it does not cover this path.

wobsoriano: ty for catching!

jeremy-clerk (approved):
LGTM 👍 but touches core, would like to see another set of eyes

PR #9375, opened 2026-08-10
fix(e2e): more robust user deletion

dstaley:
## Description

This PR updates the logic we use to create fake users to embed a run key, which is a hash of the unique data for a single test run. After integration tests finish, we manually clean up users that were created with the run key in their username, email address, or metadata.

This also includes a fix for `deleteIfExists` that will delete all matching users rather than just the first. This handles the case were we generate the same matching email/phone number across multiple tests. This shouldn't be as much of an issue now that we're storing the run key in the identifier, but it doesn't hurt!

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


wobsoriano (approved):
Extra changesets but everything looks good!

PR #9378, opened 2026-08-10
feat(ui): add Avatar button composition

austincalvelage:
## Description

https://swingset-git-austin-user-profile-03-avatar-button.clerkstage.dev/components/avatar

Adds support for `Avatar` button this will eventually be used in a component like 

`<AvatarUploader  />` as the trigger.

- Adds  `render` support to `Avatar.Root`, allowing the avatar to render as a button or dialog trigger without an additional wrapper.
- Adds the `Avatar.Icon` 
- Adds the `pen` glyph to the Mosaic icon registry.
- Adds the `--cl-color-avatar-icon` token so the edit icon can be themed independently.

```tsx
import { Avatar } from '@clerk/ui/mosaic/components/avatar';
import { Button } from '@clerk/ui/mosaic/components/button';
import { Icon } from '@clerk/ui/mosaic/components/icon';

<Avatar.Root
  size='lg'
  render={
    <button
      aria-label='Edit profile picture'
    />
  }
>
  <Avatar.Image
    src='https://avatars.githubusercontent.com/u/51144033?v=4'
    alt='Colin'
  />
  <Avatar.Fallback>CL</Avatar.Fallback>
  <Avatar.Icon>
    <Icon name='pen' />
  </Avatar.Icon>
</Avatar.Root>;
```


alexcarpenter (commented):
Stacked a PR with some suggested changes here https://github.com/clerk/javascript/pull/9413

- when using a button trigger, we'd likely just use a Headless Button vs our styled Button to avoid the overrides you needed to add
- icon viewbox is wonky, we shouldn't need to override viewboxes and keep them consistent at 0 0 16 16. not sure whats up with the pen icon, looks like it came from ceramic which has some goofy stuff going on with strokes

alexcarpenter (approved):

PR #9381, opened 2026-08-10
fix(shared): decode JWT v2 permission masks exactly

jeremy-clerk:
## Description

JWT v2 organization permission masks were parsed through JavaScript numbers, which could lose precision above the safe integer limit and cause `auth().has()` to incorrectly grant or deny permissions.

Decode permission masks directly from their decimal strings with `decimalToBinaryBits`. This avoids relying on `BigInt`, which may be incompatible with ClerkJS's legacy browser bundle, while preserving the existing bit ordering for normal masks. Malformed and negative masks now fail closed.

The backend now emits one permission mask per feature, using `0` for features without permissions. The v2 authorization regressions cover this positional contract, including a targeted feature that follows a permission-less plan feature.

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/shared/src/jwtPayloadParser.ts
```diff
@@ -62,7 +88,7 @@ function buildOrgPermissions({
       continue;
     }
 
-    for (let permIndex = 0; permIndex < permissionBits.length; permIndex++) {
+    for (let permIndex = 0; permIndex < permissionBits.length && permIndex < permissions.length; permIndex++) {
```
dominic-clerk, line 91: **[MEDIUM] `fea`/`fpm` index misalignment attaches a feature's permission mask to the wrong feature when feature targeting is active**

`buildOrgPermissions` assumes `featurePermissionMap[i]` corresponds to `features[i]`, but the encoder does not emit a mask per feature — `pkg/auth/v2.go:223` skips any feature not in `featuresInPermissions`, so `fpm` is a *compacted* list while `fea` is the full one. The two only stay aligned when the permission-bearing features happen to form a prefix of `fea`. When `params.Plan != nil` (`v2.go:69-95`) `featureSet` is seeded from plan features and targeted features are appended after, breaking that prefix invariant — a plan feature the member has no permissions on then inherits the next mask in the list, so `has({ permission: 'org:<wrong-feature>:manage' })` returns `true` for a permission the user was never granted.

This predates the diff, but it is in the function this PR rewrites and the PR's stated goal is decoding these masks exactly — the new `permIndex < permissions.length` clamp bounds the inner loop without fixing the outer index mapping. Suggest having the encoder emit a mask for every entry in `fea` (zero for features with no permissions), or emitting the feature name alongside each mask so the decoder can key on it rather than on position.

jeremy-clerk: Confirmed — this is a valid pre-existing encoder issue. I reproduced the false grant through the JWT v2 auth-object conversion and `has()` with a permission-bearing plan feature, a plan feature without permissions, and a permission-bearing targeted feature.

There isn't enough information in the current claims for ClerkJS to repair the mapping safely: once zero masks are omitted, the decoder cannot tell whether the next compacted mask belongs to the next `fea` entry or a later targeted feature. Failing closed whenever `fea.length !== fpm.length` would also deny valid permissions for existing plan tokens.

The correct fix is in the backend: emit one `fpm` entry for every `fea` entry, using `0` for features without permissions, and cover the plan-plus-targeting case there. I'm keeping this PR scoped to exact decimal mask decoding and undeclared-bit handling rather than adding an ambiguous decoder heuristic.

— Comment generated with Codex with @jeremy-clerk's supervision

dominic-clerk (approved):

packages/shared/src/jwtPayloadParser.ts
dominic-clerk: 👍 I'll follow-up on this with a backend PR

dominic-clerk: https://github.com/clerk/clerk_go/pull/21138

jeremy-clerk: Verified that the backend fix is merged. This PR is now updated to match the new positional contract: fixtures include `0` masks for features without permissions, and the regression covers `fea = repositories,impersonation,billing` with `fpm = 6,0,3` through the v2 auth object and `has()`.

— Comment generated with Codex with @jeremy-clerk's supervision

wobsoriano (approved):
awesome, looks good here 👍🏼

PR #9386, opened 2026-08-10
feat(ui): compose user profile panel

austincalvelage:
## Description
https://swingset-git-austin-user-profile-08-panel-composition.clerkstage.dev/user/user-profile-profile-panel

CleanShot 2026-08-13 at 12 53 59@2x

CleanShot 2026-08-13 at 12 54 07@2x


packages/ui/src/mosaic/user-profile/user-profile-profile-panel.view.tsx
```diff
@@ -0,0 +1,114 @@
...
+  onSetPrimaryPhone,
+  onRemovePhone,
+  onConnectAccount,
+  onManageConnectedAccount,
+  onRemoveConnectedAccount,
+  onConnectWeb3Wallet,
+  onManageWeb3Wallet,
+  onSetPrimaryWeb3Wallet,
+  onRemoveWeb3Wallet,
+  onDeleteAccount,
```
maxyinger, line 62: sheesh these get long

packages/ui/src/mosaic/user-profile/user-profile-profile-panel.styles.ts
```diff
@@ -22,7 +22,13 @@ export const styles = stylex.create({
     width: '20px',
   },
   root: {
-    gap: space['6'],
+    gap: space['4'],
+    display: 'flex',
+    flexDirection: 'column',
+    width: '100%',
```
maxyinger: I'm not super opposed, but does this break without the width: 100% on these? just wondering if it's redundant on here and the sections styles

packages/ui/src/mosaic/user-profile/user-profile-profile-panel.view.tsx
```diff
@@ -0,0 +1,114 @@
...
+        />
+        {connectedAccounts.length > 0 ? (
+          <UserProfileConnectedAccountsSectionView
+            accounts={connectedAccounts}
+            onConnect={onConnectAccount}
+            onManage={onManageConnectedAccount}
+            onRemove={onRemoveConnectedAccount}
+          />
+        ) : null}
+        {web3Wallets.length > 0 ? (
```
maxyinger, line 101: we might want to be able to delineate on these in the future between an empty state and a null state. fine with leaving this for now tho and we can alter when we need it

maxyinger (approved):

packages/ui/src/mosaic/user-profile/user-profile-profile-panel.styles.ts
austincalvelage: Good call. Definitely redundant

PR #9399, opened 2026-08-11
fix(backend): decode cookie values after parsing

### LINKED ISSUE #9333: URIError: URI malformed — ClerkRequest.parseCookies throws on any malformed percent-escape in the Cookie header
### Preliminary Checks

- [x] I have not already reached out to Clerk support via email or Discord
- [x] This issue is not a question, general help request, or anything other than a bug report directly related to Clerk

### Reproduction

https://gist.github.com/davidbarratt/08d1334f2c9dde6fa5cde0da7c3a4030

### Publishable key

Not applicable. The throw happens inside the `ClerkRequest` constructor, in a string function, before any Clerk API call — the reproduction runs with no instance, no key, and no network access.

### Description

`ClerkRequest.decodeCookieValue` runs `decodeURIComponent` over the raw `Cookie` header with no `try`/`catch`, in [`packages/backend/src/tokens/clerkRequest.ts#L96-L103`](https://github.com/clerk/javascript/blob/438f2e5220e82dfefb1dfbe9e6ea4c989aac5194/packages/backend/src/tokens/clerkRequest.ts#L96-L103):

```ts
private parseCookies(req: Request) {
  const cookiesRecord = parse(this.decodeCookieValue(req.headers.get('cookie') || ''));
  return new Map(Object.entries(cookiesRecord));
}

private decodeCookieValue(str: string) {
  return str ? str.replace(/(%[0-9A-Z]{2})+/g, decodeURIComponent) : str;
}
```

`decodeURIComponent` throws `URIError: URI malformed` on any percent-escape that isn't valid UTF-8. Because `parseCookies` is called from the constructor ([L53](https://github.com/clerk/javascript/blob/438f2e5220e82dfefb1dfbe9e6ea4c989aac5194/packages/backend/src/tokens/clerkRequest.ts#L53)), the error escapes `createClerkRequest`, which is the first thing `authenticateRequest` does — so the request fails before any auth logic runs.

Two properties widen the blast radius well past the one cookie that is malformed:

- **The decode is applied to the whole header, not per value.** Any cookie on the domain can trigger it, including ones Clerk never set and never reads — analytics, ad tooling, third-party scripts.
- **The bad value persists in the browser until it expires.** Every subsequent request from that client fails, not just one.

**Steps to reproduce:**

1. `npm install @clerk/backend@3.15.1`
2. Call `createClerkRequest` with a request whose `Cookie` header contains a malformed escape:
   ```js
   import { createClerkRequest } from '@clerk/backend/internal';

   createClerkRequest(new Request('https://example.com/', {
     headers: { cookie: '__session=abc; analytics_id=%E2%9' },
   }));
   ```
3. Observe `URIError: URI malformed`.

**Expected behavior:**

A cookie value that cannot be percent-decoded is left as its raw value or skipped, and the rest of the header parses normally. A cookie unrelated to Clerk should not be able to fail the request.

**Actual behavior:**

```
URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at String.replace (<anonymous>)
    at ClerkRequest.decodeCookieValue (.../@clerk/backend/dist/chunk-DDFIPK3V.mjs:6491:22)
    at ClerkRequest.parseCookies (.../@clerk/backend/dist/chunk-DDFIPK3V.mjs:6487:57)
    at new ClerkRequest (.../@clerk/backend/dist/chunk-DDFIPK3V.mjs:6450:25)
    at createClerkRequest (.../@clerk/backend/dist/chunk-DDFIPK3V.mjs:6496:37)
```

**Inputs that throw**, all verified against 3.15.1:

| `Cookie` header | Why |
| --- | --- |
| `x=%E2%9` | escape truncated mid-sequence; `%E2` decodes alone as an incomplete UTF-8 lead byte |
| `x=%98` | lone continuation byte |
| `x=%C0%80` | overlong UTF-8 encoding |

The truncated case is the one we hit in production. A client-side script writing a value that gets clipped at the browser's ~4096-byte per-cookie limit lands mid-escape, and that alone is enough to make every later request 500.

**How it surfaces in a Next.js app.** `@clerk/nextjs` 7.x on Vercel (Node.js 22 runtime), `clerkMiddleware()` in `proxy.ts`. The proxy returns HTTP 500 and the request never reaches our handler. The minified frames map onto the same four methods:

```
Error running the exported Web Handler: URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at String.replace (<anonymous>)
    at aJ.decodeCookieValue (.next/server/chunks/[root-of-the-server]__0tiorp-._.js:49:40399)
    at aJ.parseCookies (.next/server/chunks/[root-of-the-server]__0tiorp-._.js:49:40317)
    at new aJ (.next/server/chunks/[root-of-the-server]__0tiorp-._.js:49:39648)
    at aq (.next/server/chunks/[root-of-the-server]__0tiorp-._.js:49:40531)
```

Because it happens in middleware, there is no application-level place to catch it — the exception is raised while Clerk is constructing its own request wrapper.

**Present in the latest release.** The code above is current on `main`, and the published build is the same: [`@clerk/backend@3.15.1/dist/internal.js`](https://cdn.jsdelivr.net/npm/@clerk/backend@3.15.1/dist/internal.js) lines 7007-7013.

**Possibly related, same two lines.** The regex `(%[0-9A-Z]{2})+` matches only uppercase hex digits, so lowercase escapes are skipped while uppercase ones in the same value are decoded:

```js
decodeCookieValue('x=%c3%a9')  // 'x=%c3%a9'  — left raw
decodeCookieValue('x=%C3%A9')  // 'x=é'
```

Percent-encoding is case-insensitive, so both should decode to the same value.

**Prior public report.** [ViewComfy/ViewComfy#147](https://github.com/ViewComfy/ViewComfy/issues/147) has the identical `ClerkRequest.decodeCookieValue → parseCookies → new ClerkRequest` stack, reported in July 2025 and closed with no fix and no upstream link. I searched this repo for `URI malformed`, `URIError`, `decodeCookieValue`, and malformed-cookie phrasings and did not find an existing issue.

### Environment

```shell
  System:
    OS: Linux 6.12 Debian GNU/Linux 13 (trixie)
    CPU: (5) arm64
  Binaries:
    Node: 24.18.0
    npm: 11.16.0
    pnpm: 11.11.0
  npmPackages:
    @clerk/backend: 3.15.1 => 3.15.1

  Also reproduced in production on:
    Vercel, Node.js 22.x runtime
    next: 16.2.6
    @clerk/nextjs: 7.2.1  (@clerk/backend 3.2.11)
```

brkalow:
## Summary

- parse cookie boundaries before percent-decoding values
- preserve malformed percent-encoded values instead of throwing
- add regression coverage for malformed and lowercase encodings
- include an `@clerk/backend` patch changeset

## Why

`ClerkRequest` decoded the complete `Cookie` header before parsing it. A malformed escape could throw during request construction, while encoded `;` and `=` characters could be reinterpreted as cookie boundaries. The `cookie` dependency already decodes each isolated value safely and returns the raw value when decoding fails.

Fixes #9333.

### DIFF
--- a/.changeset/tidy-cookies-smile.md
+++ b/.changeset/tidy-cookies-smile.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
+Decode request cookie values only after parsing cookie boundaries, preserving malformed percent encodings and preventing encoded delimiters from being treated as separate cookies.
--- a/packages/backend/src/tokens/__tests__/clerkRequest.test.ts
+++ b/packages/backend/src/tokens/__tests__/clerkRequest.test.ts
@@ -89,12 +89,36 @@ describe('createClerkRequest', () => {
       expect(req.cookies.get('foo')).toBe('bar');
     });
 
-    it('should parse and return cookies with special characters', () => {
+    it('should decode values after parsing cookie pairs', () => {
       const req = createClerkRequest(
-        new Request('http://localhost:3000', { headers: new Headers({ cookie: 'foo=%20bar%3B%20baz%3Dqux' }) }),
+        new Request('http://localhost:3000', {
+          headers: new Headers({ cookie: 'foo=%20bar%3B%20baz%3Dqux; after=parsed' }),
+        }),
       );
-      expect(req.cookies.get('foo')).toBe('bar');
-      expect(req.cookies.get('baz')).toBe('qux');
+
+      expect(req.cookies.get('foo')).toBe(' bar; baz=qux');
+      expect(req.cookies.get('baz')).toBeUndefined();
+      expect(req.cookies.get('after')).toBe('parsed');
+    });
+
+    it.each(['%E2%9', '%98', '%C0%80'])('should preserve a malformed encoded cookie value: %s', value => {
+      const req = createClerkRequest(
+        new Request('http://localhost:3000', {
+          headers: new Headers({ cookie: `__session=abc; analytics_id=${value}; after=parsed` }),
+        }),
+      );
+
+      expect(req.cookies.get('__session')).toBe('abc');
+      expect(req.cookies.get('analytics_id')).toBe(value);
+      expect(req.cookies.get('after')).toBe('parsed');
+    });
+
+    it('should decode lowercase percent escapes', () => {
+      const req = createClerkRequest(
+        new Request('http://localhost:3000', { headers: new Headers({ cookie: 'foo=%c3%a9' }) }),
+      );
+
+      expect(req.cookies.get('foo')).toBe('é');
     });
 
     it('should parse and return cookies even if no cookie header exists', () => {
--- a/packages/backend/src/tokens/clerkRequest.ts
+++ b/packages/backend/src/tokens/clerkRequest.ts
@@ -94,12 +94,8 @@ class ClerkRequest extends Request {
   }
 
   private parseCookies(req: Request) {
-    const cookiesRecord = parse(this.decodeCookieValue(req.headers.get('cookie') || ''));
+    const cookiesRecord = parse(req.headers.get('cookie') || '');
     return new Map(Object.entries(cookiesRecord));
-  }
-
-  private decodeCookieValue(str: string) {
-    return str ? str.replace(/(%[0-9A-Z]{2})+/g, decodeURIComponent) : str;
   }
 }
 


brkalow:
Closing in favor of #10002

PR #9404, opened 2026-08-11
feat(ui): announce the user button's in-flight action

alexcarpenter:
## Description

A row in the Mosaic user button that starts an action keeps focus while it stands down, and its spinner is decorative, so nothing reported the wait to assistive tech: `aria-disabled` says a row cannot be used, not why.

The popup now carries a polite live region from the moment it opens, and names the running action in it ("Switching to Foundry") for as long as it runs.

- The wording comes from `userButtonBase.status`, like every other string the surface renders, so it localizes with the rest.
- The label is derived from `pendingKey` and the rows' own names, keyed through `userButtonBusyKeys` rather than parsed, so the key grammar stays in one place and a container announces without passing anything extra.
- One region on the surface rather than one per affordance: `pendingKey` already allows a single action at a time.
- It lives on the popup, not in the acting row: the row often goes when the action lands (the list re-sorts, a signed-out account leaves), which would unmount the region mid-announcement.
- It is mounted while idle and empty: a region that arrives with its message already in it is not announced.

A key the surface cannot name (a row that has already left the list) leaves the region silent rather than announcing a half-filled template.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/user-button/user-button.styles.ts
```diff
@@ -3,6 +3,17 @@ import * as stylex from '@stylexjs/stylex';
...
+  // Takes the live region out of the layout without taking it out of the accessibility tree, which
+  // `display: none` and `visibility: hidden` both do.
+  visuallyHidden: {
+    overflow: 'hidden',
+    clipPath: 'inset(50%)',
+    position: 'absolute',
+    whiteSpace: 'nowrap',
+    height: '1px',
+    width: '1px',
+  },
```
alexcarpenter, line 15: probably should lift to an atom in the future to reuse.

maxyinger (approved):

alexcarpenter:
closing, gonna explore an alternate route.

PR #9411, opened 2026-08-12
refactor(ui): declare the whole Mosaic UserButton surface in one table

alexcarpenter:
## Description

feedback follow up from https://github.com/clerk/javascript/pull/9184

The Mosaic `UserButton` popup decided its shape in two places. `user-button.layout.ts` held two tables of opposite orientation (one keyed by mode, one keyed by affordance) plus a conditional type cross-checking them, and the view still hardcoded the header gear and the active-account row menu on its own.

Each mode now declares its whole surface in one table, top to bottom:

```ts
combined: {
  header: ['inviteMembers', 'manageLead'],
  organizations: { heading: ['createOrganization', 'manageAccount', 'signOut'] },
  sessions: { heading: ['addAccount'] },
  footer: ['signOutAll'],
},
```

A section the mode does not carry is `false`, so it has no array to hold actions and placing one there is impossible to express. That removed the `'none'` slot and the `ModeSlot<M>` conditional type outright.

`resolveUserButtonLayout()` runs the table against the data once and hands back which sections render and what each slot carries, in order. Every consumer loops its own list rather than asking where a given action landed:

```tsx
for (const action of data.layout.actions.header) {
  if (action === 'manageLead') { /* gear: the org where one leads, else the account */ }
}
```

Behaviour is unchanged in every reachable state. The ASCII diagram at the top of the file shows the three modes side by side.

Renames follow the data contract, so `workspace` now means only the personal-or-organization union: `ActiveAccountRow` -> `OrganizationsHeading`, `AccountsSection` -> `SessionSection`, `WorkspaceRow` -> `SwitcherRow`, and so on.

This is internal to `@clerk/ui`; the Mosaic `UserButton` is not exported yet, hence the empty changeset.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/user-button/user-button.layout.ts
```diff
@@ -1,98 +1,161 @@
...
+interface ModeLayout {
+  header: readonly UserButtonAction[];
+  /** The workspaces the active account switches between: its own, plus the organizations it is in. */
+  organizations: ListLayout | false;
+  /** The other signed-in accounts. */
+  sessions: ListLayout | false;
+  footer: readonly UserButtonAction[];
+}
+
+const modes = {
```
Ephem, line 56: This could be renamed something like `modeDefaults` right? Get's merged with data and produces a `UserButtonLayout` which is the "public" shape outside of this file?

Renaming is very much a NIT, mostly checking my understanding.

packages/ui/src/mosaic/user-button/user-button.view.tsx
```diff
@@ -721,16 +711,19 @@ function WorkspaceListLoadingRow() {
...
 
-/** The active account and everything it can switch to. This is the group that scrolls. */
-function WorkspaceSection() {
+/**
+ * The workspaces the active account can switch between, under the account's own heading. This is
+ * the group that scrolls.
+ *
+ * The heading is not the list's to withhold: an account with no organizations still needs somewhere
+ * to manage and sign out of itself. The other two modes name the account in their header instead,
+ * or are not about it at all.
```
Ephem:
```suggestion
 * The heading can render without the list: an account with no organizations still needs somewhere
 * to manage and sign out of itself.
```

From the context here it's hard to know what "The other two modes name the account in their header instead, or are not about it at all." is referring to, so I find that bit more confusing than it helps.

packages/ui/src/mosaic/user-button/user-button.view.tsx
```diff
@@ -741,13 +734,13 @@ function WorkspaceSection() {
           overflows, so short lists would sit their avatars and icons off the edge the header and
           footer align to. */}
       <Item.Group {...stylex.props(...scrollAreaViewport('auto'), styles.scroll)}>
-        {accountRow}
+        {showOrganizationsHeading ? <OrganizationsHeading /> : null}
         {/* Memberships, invitations and suggestions are three separate requests landing at three
             different moments. Rendering each as it arrives walks the list in in stages, so the
             placeholder stands in for all of them until the last one is in. */}
```
Ephem, line 739: Tangent: Oh how I want suspense. 🙃

Ephem (approved):
This is so much clearer, I ❤️ it! 🙏

packages/ui/src/mosaic/user-button/user-button.layout.ts
alexcarpenter: Your understanding is exactly right: `modes` is the static per-mode declaration, `resolveUserButtonLayout()` runs it against `UserButtonData`, and `UserButtonLayout` is the only shape that leaves the file (`ListLayout`, `ModeLayout` and `modes` are all private).

I skipped `modeDefaults` though, because "defaults" implies something overrides them and nothing does. The data can only drop an action (`offered()` removes `inviteMembers` when no org is active) or relocate one (the accounts heading's actions fall to the footer when there is no other account). Nothing ever supplies a competing value that wins, so the name would advertise an override mechanism that isn't there.

But you were pointing at something real. The collision wasn't the table, it was that `ModeLayout` (static input) and `UserButtonLayout` (resolved output) both say "layout" — and inside the resolver there was a local literally named `layout` that was *not* what the function returns:

```ts
const layout: ModeLayout = modes[mode];  // not the returned layout
```

Renamed that to `declared`, so `UserButtonLayout` is now the only "layout" in the file. Thanks for the poke.

PR #9412, opened 2026-08-12
refactor(ui): rename Mosaic Item.Title to Item.Label

alexcarpenter:
## Description

`Item.Title` becomes `Item.Label`.

That name was already taken by a second part: the quiet text on a bare action row (`Add account`, `Sign out of all accounts`) and on a group heading. Rather than give it a name of its own, it folds into the new `Item.Label` as `variant='secondary'`.

```tsx
<Item.Label>Clerk</Item.Label>                            // cl-item-label[data-variant='primary']
<Item.Label variant='secondary'>Add account</Item.Label>  // cl-item-label[data-variant='secondary']
```

The two parts only ever differed by how loud they read, so an axis on one part models that better than two parts do. `variant` matches `Button`, and the value reflects as `data-variant` like every other Mosaic axis, so consumers can scope overrides on it.

`secondary` declares no color at all. The reset sets `color: inherit` and StyleX keeps only the last atom that declares a property, so `primary` displaces it and holds its own color while `secondary` leaves it standing. That is what puts the secondary label on the row's color and carries it through the row's hover promotion, which a fixed color would freeze. It is also why `label.base` cannot hold a color.

Slot classes change: `cl-item-title` is now `cl-item-label`, and the old `cl-item-label` is now `cl-item-label[data-variant='secondary']`. Mosaic is not exported from a public `@clerk/ui` entry, so no consumer API moves.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


Ephem (approved):
Makes sense to me!

PR #9433, opened 2026-08-13
feat(ui): add Mosaic AlertDialog

maxyinger:
https://swingset-git-max-mosaic-alert-dialog.clerkstage.dev/components/alert-dialog

## Description
- Adds the Mosaic `AlertDialog`
  - extends mosaic `Dialog` with some baked in opinions
  - new `AlertDialog.Actions` api surface on top of `Dialog`
  - required title and description
  

**Baked in behavior on dialog**
| prop | value |
| --- | --- |
| `role` | `'alertdialog'` |
| `closedBy` | `'closerequest'` |
| `size` | `'prompt'` |

## Preview
Screenshot 2026-08-13 at 4 36 49 PM

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/components/alert-dialog/alert-dialog.styles.ts
```diff
@@ -0,0 +1,36 @@
...
+   * One layout at every width, because that is the convention for a `prompt` generally rather than
+   * a rule about alert dialogs: its buttons span the surface, one full width or two at even halves.
+   * A right-aligned pair sized to its labels was tried first and is what the designs do not do.
+   *
+   * A GRID rather than a flex row, and that is what makes both cases the same declaration. Filling
+   * the row needs `flex: 1` on each CHILD, which a parent cannot set — StyleX has no child
+   * selector, and reaching into the children would mean every call site remembering to pass
+   * something. `grid-auto-flow: column` with `grid-auto-columns: 1fr` puts it on the container
+   * instead: every button takes an equal share of the row, so one fills it and two split it, with
+   * no branch and nothing for a third to break.
```
maxyinger: fuggin opus.

All these styles are more comments than style. I kinda want to strip them out, but also maybe they're useful context for future agents editing 🤷 

any thoughts?

packages/ui/src/mosaic/components/alert-dialog/alert-dialog.tsx
```diff
@@ -0,0 +1,222 @@
...
+      />
+    </DialogPartNameContext.Provider>
+  );
+});
+
+/**
+ * The row holding the answer. Render the cancel first — see `alert-dialog.styles.ts` for why that
+ * ordering is what focuses it on open.
+ */
+const Actions = React.forwardRef<HTMLDivElement, AlertDialogActionsProps>(function AlertDialogActions(
```
maxyinger, line 108: almost wonder if this should just live in `Dialog` since we'll use in on other `prompt` sized dialogs for inputs that will have the same styled actions 🤔 . easy swap if so

packages/ui/src/mosaic/components/alert-dialog/alert-dialog.styles.ts
austincalvelage: I wonder if this being apart of an `alert-dialog.md`?  Makes more sense not sure tbh

packages/ui/src/mosaic/components/alert-dialog/alert-dialog.tsx
austincalvelage: Yeah I could see this living in Dialog since it isn't specific to Alert dialog

austincalvelage (approved):

.changeset/mosaic-alert-dialog.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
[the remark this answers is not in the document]
maxyinger: Keeping this changeset empty deliberately.

`@clerk/ui`'s `exports` map has no `./mosaic` subpath — the published package exposes `.`, `./no-rhc`, `./entry`, `./internal`, `./themes`, `./experimental`, `./styles.css` and `./register`, and none of them re-export Mosaic. `AlertDialog` is therefore not reachable from the published package, and the Mosaic org views are imported only by swingset, which is private. There is nothing here for someone upgrading `@clerk/ui` to act on, so a minor bump would announce an API they cannot import.

This matches how the rest of the Mosaic dialog work has been released: the changesets on the earlier PRs in this stack were emptied at review request for the same reason.

That lines up with your own criterion — empty changesets apply to non-published changes; Mosaic is not published yet. When Mosaic gets a public entry point, that change is the one that warrants the release entry.

PR #9448, opened 2026-08-13
feat(expo): add custom native profile pages

swolfand:
Adds custom user profile pages to Expo's native `UserProfileView` and `UserButton` APIs. Bridges row metadata and React Native destinations to Clerk's Android and iOS profile navigation, including URL rows and programmatic navigation. Includes unit coverage and a minor changeset.

Docs: [clerk/clerk#3158](https://github.com/clerk/clerk/pull/3158)

https://github.com/user-attachments/assets/7f60b724-6d11-40bd-a798-fbc669b432ce

https://github.com/user-attachments/assets/761daa43-ae9c-4379-ab70-faf2e8b8dce4

### DIFF
--- a/packages/expo/android/src/main/java/expo/modules/clerk/ClerkUserProfileViewModule.kt
+++ b/packages/expo/android/src/main/java/expo/modules/clerk/ClerkUserProfileViewModule.kt
@@ -4,17 +4,33 @@ package expo.modules.clerk
 
 import android.content.Context
 import android.util.Log
+import android.view.View
+import android.view.ViewGroup
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.viewinterop.AndroidView
 import androidx.lifecycle.ViewModelStore
 import androidx.lifecycle.ViewModelStoreOwner
 import com.clerk.api.Clerk
 import com.clerk.api.FrameworkIntegrationApi
+import com.clerk.ui.R
 import com.clerk.ui.navigation.ClerkHostBackActionProvider
 import com.clerk.ui.userprofile.UserProfileView
+import com.clerk.ui.userprofile.custom.LocalUserProfileCustomNavigator
+import com.clerk.ui.userprofile.custom.UserProfileCustomRow
+import com.clerk.ui.userprofile.custom.UserProfileCustomRowPlacement
+import com.clerk.ui.userprofile.custom.UserProfileRow
+import com.clerk.ui.userprofile.custom.UserProfileRowIcon
+import com.clerk.ui.userprofile.custom.UserProfileSection
 import expo.modules.kotlin.AppContext
 import expo.modules.kotlin.modules.Module
 import expo.modules.kotlin.modules.ModuleDefinition
 import expo.modules.kotlin.viewevent.EventDispatcher
+import org.json.JSONArray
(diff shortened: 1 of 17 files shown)


packages/expo/android/src/main/java/expo/modules/clerk/ClerkUserProfileViewModule.kt
```diff
@@ -24,11 +40,75 @@ private fun debugLog(tag: String, message: String) {
...
+    "email" -> R.drawable.ic_email
+    "phone" -> R.drawable.ic_phone
+    "add" -> R.drawable.ic_plus
+    "switch" -> R.drawable.ic_switch
+    "users" -> R.drawable.ic_users
+    "warning" -> R.drawable.ic_warning
+    "info" -> R.drawable.ic_information_circle
+    "globe" -> R.drawable.ic_globe
+    "folder" -> R.drawable.ic_folder
+    "book" -> R.drawable.ic_folder
```
wobsoriano, line 77: book maps to ic_folder here but to the SF Symbol "book" on iOS, is that intentional?

packages/expo/src/native/UserProfileCustomPages.tsx
```diff
@@ -0,0 +1,206 @@
...
+  return navigation;
+}
+
+export function serializeUserProfileCustomPages(customPages: UserProfileCustomPage[]): string {
+  const paths = new Set<string>();
+
+  for (const { path } of customPages) {
+    if (paths.has(path)) {
+      throw new Error(`User profile custom page path "${path}" must be unique.`);
+    }
```
wobsoriano, line 104: nit: similar to web, we just console.warn and drop the bad row in prod. Might also be good to do that

somethjiong like

```ts
if (__DEV__) {
  console.warn(`Duplicate user profile custom page path "${path}". Ignoring the duplicate.`);
}
continue;
```

packages/expo/src/native/UserProfileCustomPages.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+export type UserProfileSection = 'profile' | 'account';
+
+export type UserProfileRow = 'manageAccount' | 'security' | 'switchAccount' | 'addAccount' | 'signOut';
+
+export type UserProfileCustomPagePlacement =
+  | { type: 'sectionStart'; section: UserProfileSection }
+  | { type: 'sectionEnd'; section: UserProfileSection }
+  | { type: 'before'; row: UserProfileRow }
+  | { type: 'after'; row: UserProfileRow };
```
wobsoriano, line 33: FYI web went with a flat `pageOrder` in https://github.com/clerk/javascript/pull/9343 and we have `placement` here. I think placement is right for native since the built-in rows are fixed correct?

wobsoriano (commented):
Left a few minor comments but this is looking good overall.

Any thoughts on `icon` eventually accepting a React element? If that's not possible I think we should tell it to customer ping as well (not a blocker tho, we can follow up)

packages/expo/ios/ClerkNativeBridge.swift
```diff
@@ -662,3 +818,44 @@ struct ClerkInlineProfileWrapperView: View {
...
+
+  var body: some View {
+    Group {
+      if let index = rows.firstIndex(where: { $0.path == path }),
+         state.views.indices.contains(index)
+      {
+        ClerkReactCustomPageView(view: state.views[index])
+      }
+    }
+    .frame(maxWidth: .infinity, maxHeight: .infinity)
```
seanperez29, line 838: Should we apply the matching row’s `label` as the native navigation title here?

wobsoriano (approved):

PR #9449, opened 2026-08-14
fix(ui): populate enterprise SSO redirect urls when continuing a sign-up

zourzouvillys:
## Description

Sign-ups that continue into an enterprise SSO connection can fail with `invalid_redirect_url` ("Redirect url invalid") instead of redirecting to the identity provider, leaving the user with a flow that cannot be completed and reproduces on every retry.

### What happens

`completeSignUpFlow` routes a sign-up that is still in `missing_requirements`. Its first branch hands off to the identity provider when `enterprise_sso` is a missing field:

```ts
if (signUp.missingFields.some(mf => mf === 'enterprise_sso')) {
  return signUp.authenticateWithRedirect({ strategy: 'enterprise_sso', redirectUrl, redirectUrlComplete, ... });
}
```

`redirectUrl` and `redirectUrlComplete` defaulted to `''`, so a caller that did not pass them sent:

```json
{ "strategy": "enterprise_sso", "redirect_url": "", "action_complete_redirect_url": "" }
```

An empty string is not a valid redirect URL, so the request is rejected and the sign-up dead-ends. Because both values were optional with a silent default, omitting them was invisible at the call site — there was no type error and no local failure, only a rejected request at runtime.

### Why so many steps could reach it

A sign-up usually does not know it requires `enterprise_sso` when the form is first submitted. The requirement appears once the identity behind the sign-up is resolved, which can be several steps later. Whichever step is active at that moment performs the hand-off — and most of them did not pass the redirect URLs:

| Call site | Passes `redirectUrl` | Passes `oidcPrompt` |
| --- | --- | --- |
| `SignUpStart` (×2) | yes | yes |
| `handleCombinedFlowTransfer` | yes | no |
| `SignUpContinue` | no | yes |
| `SignUpEmailLinkCard` | no | no |
| `SignUpVerificationCodeForm` | no | no |
| `SignUpProtectCheck` | no | no |
| `EmailLinkVerify` | no | no |

Every combination of the two is represented, which points at parameters that are easy to forget rather than at one call site being wrong. All three values are derived identically from the sign-up context, which every one of these components already reads — so each call site was independently re-deriving values that are the same everywhere.

Only sign-ups that reached SSO directly from the first form were unaffected, which is why this did not show up on every SSO sign-up.

`navigateToNextStepSignUp` also omits the URLs, but it returns early unless `missingFields` is empty, so it cannot reach the branch. It is left as is.

### The change

- **`useCompleteSignUpFlow`** binds `redirectUrl`, `redirectUrlComplete` and `oidcPrompt` from the sign-up context once, along with `navigate` and the default `handleComplete` that six components were each repeating. Call sites now pass only the routing paths they genuinely differ on, so there is nothing left to forget.
- **`EmailLinkVerify`** takes an explicit `ssoCallbackUrl`, supplied by the two sign-up render sites. It is shared with the sign-in flow, so it has no sign-up context of its own to bind — and its existing `redirectUrl` prop is not the SSO callback (`SignInEmailLinkVerify` passes `../factor-two`), so that one cannot be reused.
- **`completeSignUpFlow`** no longer defaults the two URLs to `''`. If the SSO branch is reached without them it throws, so a future call site fails loudly and locally instead of producing a rejected request that surfaces to the user as a dead end. With the call sites bound this is unreachable from the UI; it exists to keep it that way.

`SignUpContinue` additionally passed a `handleComplete` that called `removeClerkQueryParam` for the ticket and invitation params — `completeSignUpFlow` already does exactly that in its `complete` branch immediately before invoking `handleComplete`, so the bound default is equivalent and those duplicated calls are dropped.

The branch ordering is unchanged: `enterprise_sso` is still evaluated before the protect-check routing, which is deliberate and already covered by `prioritizes enterprise_sso over protect_check`.

### Bundle budget

The `ui-common` bundlewatch budget moves 132KB → 134KB, regenerated with `pnpm bundlewatch:fix` (its own formula — measured size plus 1KB, rounded up to the next KB). No other entry changed.

That chunk was already sitting within about 50 bytes of its ceiling. Building `packages/ui` at this branch and at `origin/main` puts the difference at **72 bytes gzipped**, and CI reports the result 23 bytes over the 132KB limit. The added bytes are in `common/EmailLinkVerify.tsx`, which belongs to that chunk; the components simplified here are in the `signup` chunk, which is well under its own budget.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.changeset/signup-enterprise-sso-redirect-urls.md
```diff
@@ -0,0 +1,10 @@
+---
+'@clerk/shared': patch
+'@clerk/ui': patch
+---
+
+Fix sign-ups that continue into an enterprise SSO connection failing with `invalid_redirect_url` ("Redirect url invalid") instead of redirecting to the identity provider.
+
+A sign-up does not always know it requires `enterprise_sso` when the form is first submitted — the requirement appears once the identity behind the sign-up is resolved, which can happen several steps later. Whichever step was active at that point performed the hand-off to the identity provider, and most of them did so without the redirect URLs it requires, so the request was rejected and the sign-up dead-ended with no way to continue. Retrying reproduced it every time. Flows that reached SSO directly from the first sign-up form were unaffected, which is why this only showed up on some sign-ups.
+
+The redirect URLs are now derived from the sign-up context wherever the flow continues, so the hand-off works from every step: the continue form, email-link and code verification, and the verification step that precedes them.
```
dstaley, line 10:
```suggestion
Fix an issue where sign-ups that used an enterprise SSO connection did not correctly forward redirect URLs.
```

packages/shared/src/internal/clerk-js/completeSignUpFlow.ts
```diff
@@ -32,6 +32,13 @@ export const completeSignUpFlow = ({
     return handleComplete && handleComplete();
   } else if (signUp.status === 'missing_requirements') {
     if (signUp.missingFields.some(mf => mf === 'enterprise_sso')) {
+      // FAPI rejects an empty redirect url, which reaches the user as a dead end rather than the caller as a bug.
```
dstaley, line 35:
```suggestion
```

packages/ui/src/common/EmailLinkVerify.tsx
```diff
@@ -14,14 +14,16 @@ import { EmailLinkStatusCard } from './EmailLinkStatusCard';
 export type EmailLinkVerifyProps = {
   redirectUrlComplete?: string;
   redirectUrl?: string;
+  /** SSO callback url, required only when the verified sign-up still has to hand off to an enterprise connection. */
```
dstaley, line 17:
```suggestion
```

packages/ui/src/components/SignUp/useCompleteSignUpFlow.ts
```diff
@@ -0,0 +1,52 @@
...
+  handleComplete?: () => Promise<void>;
+};
+
+/**
+ * Binds the sign-up context to `completeSignUpFlow` so callers only pass the routing paths.
+ *
+ * `redirectUrl`, `redirectUrlComplete` and `oidcPrompt` are needed by the `enterprise_sso` branch
+ * but are derived identically at every call site, so binding them here is what keeps a caller from
+ * silently omitting them.
+ */
```
dstaley, line 24:
```suggestion
```

dstaley (approved):
looks straightforward! love that it consolidates some behavior into a shared module. made some suggestions to simplify some wordy stuff and remove some comments that didn't seem necessary, but lgtm otherwise.

PR #9453, opened 2026-08-14
fix(ui): render a message for password errors the UI does not map

dmoerner:
We pushed a new password error from the backend with the intention that it would be passed through to our SDKs. However, password fields have special behavior that override the generic error builder and do not fallback to the apierror from the backend.

Two layers to the fix:

1. Specific: handle `form_password_matches_identifier`, and add the `unstable__errors.form_password_matches_identifier` key with an en-US string mirroring the server copy.
2. General: filter to codes present in `mapComplexityErrors` and, when none remain, fall back to `longMessage || message`. This restores the passthrough for any future unrecognized code, and stops an unmapped code arriving alongside a mapped one from contributing a blank list item.

CleanShot 2026-08-14 at 11 08 20@2x

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/great-pots-tickle.md
+++ b/.changeset/great-pots-tickle.md
@@ -1,0 +1,7 @@
+---
+'@clerk/localizations': patch
+'@clerk/shared': patch
+'@clerk/ui': patch
+---
+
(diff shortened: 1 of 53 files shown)


Ephem (approved):
Looks good to me, nice find!

PR #9455, opened 2026-08-14
feat(e2e): use PLAPI to dynamically create instances

dstaley:
## Description

This PR leverages Clerk's Platform API to dynamically create instances based on specific configuration files instead of relying on manually created instances shared between multiple test runs. This eliminates the concern of exhausting the limits on a particular applications (such as too many users, too many organizations, etc). This also allows us to more easily create instances with specific configurations, instead of relying on creating instances and persisting their secret keys in a shared credential. This means that a single developer is now able to configure an instance directly from their PR.

## How it works

For now, the way this works is by hijacking the existing `withInstanceKeys` method. Previously, the method would lookup the provided `keyName` from either a local keys file or an environment variable. Now, if the provided `keyName` matches that of an on-disk configuration file, the method will use the configuration file to create an application with the specified configuration, and return the keys from that created instance.

Each instance is tagged with a unique hash which allows us to clean up all instances created during the e2e run. This also prevents instances from being reused across tests.

## Notes for reviewers

The main focus of review should be the configuration files found within `integration/configs`. This is the definition format for configuring an instance. The `defineConfig` method accepts an object with two properties:

1. `config` - This is an object that's sent to PLAPI's `PATCH /config` endpoint. It can be either a complete instance configuration as returned by `clerk config pull` or an incremental configuration that is merged into the default instance configuration. The two configs in this PR are of the latter type, but either is supported.
2. `setup` - An optional post-instance creation setup function that's provided with a configured `ClerkClient` BAPI client as well as a `patchConfig` function that can be used to make subsequent PLAPI requests. This is useful in instances where you want to configure settings not exposed by PLAPI, or you need access to information not available at the time of the initial request such as the instance's frontend API endpoint. (Note: while writing this I realize we can likely provide information like the application's instances to the `defineConfig` since creating the application and configuring it are two separate API requests. Maybe in a future PR!)

The remainder of the PR is the plumbing necessary for application creation and cleanup, and likely isn't worth much scrutiny. In order to make the `withInstanceKeys` method asynchronous, I had to update the `integration` folder to be interpreted as ESM to support top-level await; this caused a few files to need to be changed from CJS to ESM syntax for things like file paths.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


integration/configs/with-email-codes.js
```diff
@@ -0,0 +1,152 @@
...
+  setup: async ({ applicationName, clerkClient, publishableKey, patchConfig }) => {
+    // setup allowed origins for the electron tests
+    await clerkClient.instance.update({ allowedOrigins: ['clerk://app'] });
+
+    const parsedPublishableKey = parsePublishableKey(publishableKey);
+    if (!parsedPublishableKey) {
+      throw new Error('The created application has an invalid publishable key.');
+    }
+
+    const registrationResponse = await fetch(`${oauthProviderUrl}/oauth/register`, {
```
wobsoriano, line 111: should we delete this dynamically registered OAuth client during cleanup? We delete the generated platform app, but that does not appear to delete the OAuth client

wobsoriano (approved):
I love this direction and the move away from long-lived shared test instances. I left one question about cleaning up dynamically registered OAuth clients.

Overall this looks good to me, approved! Thank you for working on this 🙏🏼

integration/configs/with-email-codes.js
dstaley: yeah, good call. I checked that there wasn't a limit, but it's easy to just delete them. https://github.com/clerk/javascript/pull/9455/commits/ba3291bb5f6a6f66a2fbe5aa0de729125e139feb

brkalow (approved):

PR #9463, opened 2026-08-15
feat(expo): add custom user profile destinations

seanperez29:
## Description

This PR improves custom user profile pages in `@clerk/expo` and adds support for push-only custom destinations.

### Changes

- Preserves embedded user profile navigation when switching tabs on iOS.
- Keeps React Native custom page content mounted while another custom destination is pushed on top, preserving local React state.
- Correctly unmounts custom pages when they are actually removed from the navigation stack.
- Invalidates retained navigation when:
  - the active user changes;
  - any retained custom route is removed or renamed.
- Uses the custom page `label` as the native iOS navigation title.
- Adds `customDestinations` to `UserProfileView` and `UserButton`.
  - Destinations can be opened with `useUserProfileCustomPageNavigation().push(path)`.
  - Destinations do not create rows on the root profile screen.
  - Existing `customPages` behavior remains unchanged.
- Rejects duplicate paths across custom pages and destinations.
- Adds a changeset and lifecycle regression coverage for these behaviors.

### Testing

The change can be tested with a profile containing a visible custom page and a push-only destination:

1. Open the visible custom page.
2. Update some local state on that page.
3. Push the custom destination using `push(path)`.
4. Confirm the destination has a native title and back button on iOS.
5. Navigate back and confirm the original page and its local state are preserved.
6. Switch to another Expo Router tab and return to the profile.
7. Confirm the current profile route and React state are still preserved.
8. Remove an earlier custom route while a later destination is open.
9. Confirm the profile returns to its root without displaying an empty or orphaned destination.
10. Confirm push-only destinations do not appear as rows on the profile root.

Validation performed:

- All 189 `@clerk/expo` tests pass.
- Type declaration generation passes.
- The full monorepo build passes.
- The native iOS app builds and the navigation lifecycle was verified in the Expo quickstart.
- Android Kotlin compilation passes.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/fresh-tabs-remember.md
+++ b/.changeset/fresh-tabs-remember.md
@@ -1,0 +1,5 @@
+---
+'@clerk/expo': patch
+---
+
+Preserve native user profile custom pages across tab switches and display their labels as native navigation titles on iOS.
--- a/.changeset/tidy-destinations-push.md
+++ b/.changeset/tidy-destinations-push.md
@@ -1,0 +1,5 @@
+---
+'@clerk/expo': minor
+---
+
+Add push-only `customDestinations` to the native `UserProfileView` and `UserButton`, allowing custom profile pages to navigate forward without adding another row to the profile root.
--- a/packages/expo/android/build.gradle
+++ b/packages/expo/android/build.gradle
@@ -102,6 +102,8 @@ try {
 }
 
 dependencies {
+  testImplementation "junit:junit:4.13.2"
+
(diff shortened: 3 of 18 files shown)


wobsoriano (approved):
Looks good on the Expo side and thanks for the clear description!

PR #9469, opened 2026-08-17
fix(backend): Reject non-session JWT categories as session tokens

dominic-clerk:
## Description

Session tokens, handshake tokens, and JWT-template tokens are all signed with the same instance key, and only the `cat` protected-header tag distinguishes them. Nothing in the auth paths checked it, so a JWT-template token was accepted anywhere a session or handshake token was expected.

authenticateRequest() now rejects a non-session category in the Authorization header and the __session cookie with token-type-mismatch (SEC-340), and verifyHandshakeJwt rejects one before signature verification (AISEC-85).

Tokens with no `cat` and instances configured to omit the category are still accepted, so this is behaviour-preserving for anything minted before the category rollout.

Fixes AISEC-85

Fixes SEC-340

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/reject-non-session-jwt-categories.md
+++ b/.changeset/reject-non-session-jwt-categories.md
@@ -1,0 +1,5 @@
+---
+'@clerk/backend': patch
+---
+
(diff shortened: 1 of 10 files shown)


wobsoriano (approved):
thanks!

PR #9471, opened 2026-08-17
fix(ui): update Mosaic input focus styles

austincalvelage:
https://swingset-git-austin-core-3619-update-input-focus-styles.clerkstage.dev/components/input

## Description

image

image

image

image

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/components/input/input.styles.ts
```diff
@@ -1,45 +1,52 @@
...
-    transitionProperty: 'color, background-color, border-color',
+    outlineOffset: {
+      default: null,
+      '@media (forced-colors: active)': {
+        default: null,
+        ':focus-visible': '2px',
+      },
+    },
+    transitionDuration: '0.2s',
+    transitionProperty: 'color, background-color, border-color, box-shadow',
```
maxyinger, line 61: might be nice to use the swift-out easing on just the box-shadow just so the shadow has a hair of bounce when it expands. I have this in the transition easing vars somewhere:

https://www.easings.dev/swift-out

packages/ui/src/mosaic/components/input/input.styles.ts
```diff
@@ -1,45 +1,52 @@
...
-    transitionDuration: durationVars['--cl-duration-base'],
-    transitionProperty: 'color, background-color, border-color',
+    outlineOffset: {
+      default: null,
+      '@media (forced-colors: active)': {
+        default: null,
+        ':focus-visible': '2px',
+      },
+    },
+    transitionDuration: '0.2s',
```
maxyinger: mind using our duration tokens here? I've been hot on assymetric easings that are snappier in than out. maybe worth playing with and seeing if they feel good here.

Can probably do the fast duration on active here, then something a little slower when it's transitioning out of the state

maxyinger (approved):

PR #9482, opened 2026-08-18
fix(ui): User Panel rough in

austincalvelage:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/user-profile/user-profile-account-section.view.tsx
```diff
@@ -74,83 +74,103 @@ export function UserProfileAccountSectionView({
     .toUpperCase();
   const updateName = onNameChange ? () => onNameChange(name) : undefined;
   const updateUsername = onUsernameChange ? () => onUsernameChange(username) : undefined;
+  const shouldBreakOutContacts = emails.length > 1 || phones.length > 1;
```
maxyinger: Can come back to this, but this might be passed in as a prop depending on behavior?

maxyinger: feel like we should move to this being a prop passed in and controlled by account type in the future, but cool with just doing the split off of layout logic for now so ppl can get a feel

packages/ui/src/mosaic/components/section/section.styles.ts
```diff
@@ -36,32 +37,49 @@ export const styles = stylex.create({
...
   item: {
+    paddingBlock: {
+      default: null,
+      [stylex.when.ancestor(':where(*)', sectionItemsMarker)]: space['4'],
+    },
     alignItems: 'center',
+    backgroundColor: {
+      default: null,
+      [stylex.when.ancestor(':where(*)', sectionItemsMarker)]: colorVars['--cl-color-card'],
+    },
```
maxyinger: dumb question, but does `Items` need to exist anymore? or can everything just be a `Row` now with the current design?

why do the 1px gap with the parent bg painted the color of the background? this just like the simplest approach with stylex selectors?

packages/ui/src/mosaic/user-profile/user-profile-security-icon.tsx
```diff
@@ -0,0 +1,34 @@
...
+  mobile: 'device-phone',
+  passkey: 'security-passkey',
+  sms: 'security-phone',
+} as const;
+
+export function UserProfileSecurityIcon({ name }: { name: UserProfileSecurityIconName }) {
+  return (
+    <Section.Media
+      size='lg'
+      {...mergeStyleProps(stylex.props(styles.media), { style: { height: space['9'], width: space['9'] } })}
```
maxyinger, line 24: all of these are just 4px smaller than other icon frames?

maxyinger (approved):
looks good to me. just some general Qs, but know we're just trying to cast a wide net and refine later, so not blockers on my end

packages/ui/src/mosaic/user-profile/user-profile-security-icon.tsx
austincalvelage: Yeah prolly makes sense to bump those to 40 for consistency

PR #9497, opened 2026-08-18
fix(clerk-js,react,shared,ui): resume an OAuth transfer after a verification challenge

zourzouvillys:
## Description

Signing up with a social provider from the **sign-in** page works by *transfer*: the sign-in comes back with a `transferable` first-factor verification carrying `external_account_not_found`, and the client is expected to complete it as a sign-up with `signUp.create({ transfer: true })`.

That conversion lives in exactly one place — the linear branch list in `_handleRedirectCallback`. The verification-challenge check sits **above** the two transfer branches and returns early, so a challenged sign-in never evaluates them; and because the callback route is navigated away from, nothing re-runs the list.

`SignInProtectCheck` then routed onward with its own private `switch`, which enumerates the statuses an *interactive* sign-in can be in. A sign-in awaiting transfer is `needs_identifier`, which is not one of them, so it fell to `default:` and returned to the start of sign-in.

That is not merely a wrong destination. `SignInStart` displays `external_account_not_found` and then calls `signIn.create({})` to stop the error surviving a page reload — which **replaces the attempt** and discards the transferable verification, the only reference to the pending transfer. The user is stranded permanently, and every retry reproduces the whole sequence.

The sign-**up** challenge card never had this problem: it delegates to `completeSignUpFlow`, the same router the callback itself uses, so it cannot drift from it. The sign-**in** card had a private copy.

**One router, two entry points.** `_handleRedirectCallback` takes a `resuming` flag that skips the two challenge short-circuits, exposed as `clerk.__internal_resumeAfterProtectCheck(...)`. The challenge card's `default:` arm hands back to it rather than guessing a destination. No transfer routing is duplicated: `transferable: false`, the gated-transfer result, `unsafeMetadata` and finalisation on the after-**sign-up** URL are all still owned by the one router. The resumed continuation is re-authorised server-side, so the flag changes where the browser is sent, not what is permitted.

**The pending transfer is latched at mount, before the challenge runs.** `SignIn.fromJSON` replaces `firstFactorVerification` wholesale on every write, so a response that re-serializes the sign-in without it would erase the only marker distinguishing "a sign-up is in progress" from "an ordinary challenged sign-in". Re-reading it after the challenge would silently fall back to the broken path — there is a test for exactly that. The latch is defensive: `fromJSON` provably replaces the field, but a response that actually omits it was not observed, so this guards a shape the resource permits rather than one seen in the wild.

**`navigateNext` moved into `handleProtectCheck.ts` as `resumeSignInAfterProtectCheck`**, next to the helper that routes *into* the challenge. The gate has an entry choke point and an exit choke point; both now live in one file, and the exit is testable on its own. A new caller needs both — a card that enters through the helper and then hand-rolls its exit is the shape that caused this.

Two smaller fixes on the same path:

- **A stale or direct visit to `/sign-in#/protect-check`** returned an empty card shell forever. It now returns to the start of the flow, which is what the sign-up card has always done.
- **`SSOCallback`'s error handler could throw out of its own `catch`.** `handleError` re-throws anything it does not recognise, and the callback's "did not complete" guards throw a plain `Error` — which it does not. That skipped both the message and the 4-second recovery, so the page sat on its spinner and the failure appeared only as an unhandled rejection. Every dead-end on this route was invisible for that reason.

### Review

Reviewed by CodeRabbit, `codex review`, and a four-lens panel (challenge bypass, state-machine lifecycle, API surface and release, and a contrarian arguing the change is over-engineered), then a second `codex review` over the commits written in response to the first round. Every finding is either fixed below or declined with a reason. Each new guard was verified by breaking the code it protects and watching it go red, then restoring.

**Fixed**

- **The `IsomorphicClerk` wrapper threw on a runtime that predates the method** (CodeRabbit). The wrapper is always defined, so it cannot itself signal whether the loaded clerk-js supports the call, and calling straight through raised a `TypeError` at a host caller rather than doing nothing. Presence-checked now, like `__internal_windowNavigate` beside it. The prebuilt UI is unaffected either way — clerk-js constructs it with the real `Clerk`, so its own feature check tests the object whose version can actually skew.
- **Two navigation assertions could never fail** (CodeRabbit). `navigate` is called with a single argument on this path, and `expect.anything()` does not match a missing one, so `not.toHaveBeenCalledWith(url, expect.anything())` was vacuously true whatever the code did — it would have passed through the regression it was written to catch. Both sites now inspect the first argument across the recorded calls.
- **A queued resume call could become an unhandled rejection** (CodeRabbit). `replayInterceptedInvocations` invokes queued callbacks and discards what they return, so a rejection on the pre-load path had no caller left to reach. The handler now sits inside the shared `callback`, so the loaded and queued arms carry identical handling rather than two copies that can drift.
- **A newer `@clerk/ui` against an older runtime threw and stranded the transfer** (codex, P2). `@clerk/ui` reaches apps independently of `clerk-js`, so the card can meet a runtime with no `__internal_resumeAfterProtectCheck`; the call was unconditional. It is now feature-detected and degrades to the previous destination instead of throwing.
- **`IsomorphicClerk` declared `customNavigate` but forwarded only `params`** (codex P2; independently raised by the API panellist, who rated it major and silent rather than a compile error). A call through `ClerkProvider` fell back to `Clerk.navigate`, which resolves component-relative destinations against the origin — the hazard this PR avoided `__internal_handleResourceCallback` for in the first place. It is accepted and forwarded, as `__internal_handleResourceCallback` does.
- **A failed continuation left the card permanently inert** (state-machine panellist, blocker). `handleError` re-throws what it does not recognise, and the continuation can raise a plain `Error` — a transient fetch failure, or a callback that did not complete. That throw escaped the void-invoked challenge run, leaving no spinner, no message and no retry. Both reporting sites in `useProtectCheckRunner` are now guarded at the chokepoint, which also covers the sign-up card and every other error `onResolved` can raise.
- **A superseded `SSOCallback` run could schedule a stale bounce** (state-machine panellist and CodeRabbit independently, then codex on the fix). The cleanup closed over a local assigned later inside the async `catch`, so it was always `undefined` at cleanup time. Holding the id in a ref was not enough — cleanup runs while `handleRedirectCallback` is still pending, so the stale timer does not exist yet and there is nothing to clear. The run now learns it was superseded and declines to schedule at all, and stops overwriting the newer run's card state. Not covered by a test: `SSOCallback` has no harness, and driving effect supersession with an in-flight promise costs more than the fix.
- **The `complete` and `default` arms of `resumeSignInAfterProtectCheck` were byte-identical** (contrarian). Deduplicated, removing a place for the two to drift.
- **A test named for an assertion it did not make** (contrarian). `completes the transfer as a SIGN-UP and finalizes on the after-sign-up url` replaced `setActive` with a bare mock, so no URL was ever asserted. Renamed to what it checks.
- **`__internal_navigateOnSetActive` was fixed but never pinned** (contrarian). Both transfer assertions now require it; dropping it fails them.
- **Only the sign-in half of the stale-gate short-circuit was tested** (contrarian). `resuming` skips a second short-circuit keyed on the sign-up resource — the arm that diverts to a different card rather than back to the same one — and that half now has its own test.
- **The latch's justification rested on an unstated observation** (contrarian). Stated below: it is defensive.
- **Bundle ceiling was looser than the measurement warranted** (API panellist). bundlewatch measures the largest signin chunk at 17,533B gzipped at this head, against a `17.5KB` (17,920B) ceiling — 387B of headroom, where `18KB` would have left 899B unaudited. That is bundlewatch's own measurement of the merge commit rather than a local build of the branch, since bundlewatch is what enforces the ceiling.
- **`ResumeAfterProtectCheckParams` stayed out of the docs only incidentally** (API panellist), via typedoc's `excludeNotDocumented`. It carries `@internal` now, like its sibling.
- **A test fixture widened `status` to `string`** (API panellist) in a file `tsconfig` excludes from typechecking, so a typo'd status would compile. Typed from `SignInJSON['status']`.
- **Changeset bump level** (API panellist; also flagged by the Break Check bot). `@clerk/clerk-js` and `@clerk/shared` are `minor`, not `patch` — the change adds a required member to the exported `Clerk` interface plus two exported types, and that is what the bot computes and what `__internal_handleResourceCallback` shipped as. `@clerk/react` and `@clerk/ui` stay `patch`.

**Declined**

- **Move the seam into `@clerk/shared/internal/clerk-js/` instead of the `Clerk` interface** (contrarian, its strongest objection: the `IsomorphicClerk` mirror exists only to satisfy the type system, and a public seam created the skew surface that then needed a guard). It is a fair criticism of the shape, but acting on it means re-implementing the transfer completion outside the router and giving up the single-owner property this change is built around — a redesign, not a review fix, on a path that currently strands users. The skew hole it cites is closed above. Worth revisiting when the transfer completion and `handleSignUpIfMissingTransfer` are consolidated, which is the real duplication.
- **Route out of the flow when "Try again" has nothing to retry** (state-machine panellist, major). Real, and pre-existing: once the gate is cleared, `retry` re-runs an effect that returns early and renders an empty card. It is shared with the sign-up card, and `retry`'s `useCallback` closes over its first-render params, so a correct fix needs the runner to hold its params in a ref rather than a one-line check. Out of scope here; the blocker fix above at least means the failure is now visible rather than silent. To be filed as a follow-up.
- **Clear `premountMethodCalls` after replay** (state-machine panellist). The two replay sites are mutually exclusive and each runs once, so a queued resume cannot fire twice today.
- **The isomorphic wrapper resolves before the work it queues** (state-machine panellist). True, and identical to `handleRedirectCallback` and `handleGoogleOneTapCallback` beside it; changing one of the three in isolation would be the drift this PR is about.
- **No escape hatch from a terminal challenge error** (state-machine panellist). Pre-existing and shared with the sign-up card.
- **Express the skew guard as an optional-member type shim** like `windowNavigate.ts` (API panellist). `typeof x === 'function'` is the same idiom used for the `__internal_windowNavigate` chokepoint in `isomorphicClerk.ts`, and is lint-clean.
- **`handleRedirectCallback` also drops `customNavigate`** (API panellist). Pre-existing, and a behaviour change for every caller of a much older method.
- **Release `@clerk/clerk-js` as major** (codex, P1), reading `AGENTS.md` as requiring a major for any change to the core `Clerk` class API. Declined on four counts: the rule's stated rationale is older SDKs still calling the latest `clerk-js`, and an addition cannot break them because an older SDK never calls a method that did not exist — the breaking direction the same rules name is *removing or renaming*; `break-check` classifies this diff as 0 breaking / 2 additions — the same one member, reported on the `.` and `./no-rhc` subpaths — and computes `MINOR`, 6.29.2 → 6.30.0; the `check-major-bump` CI check passes with "no major version bump detected"; and the closest precedent, the commit that added `__internal_handleResourceCallback` to this same class, shipped `@clerk/clerk-js: minor`. Happy to bump it if a maintainer reads the rule more strictly.
- **The resumed path passes `__internal_navigateOnSetActive` into `setActiveNavigate`, which can race an unmount** (contrarian, incidental). Both social-button paths pass the same param the same way on the ordinary social sign-up flow that runs in production today, so the behaviour is shared rather than introduced here — and omitting it reintroduces the wrong-destination bug it was added to fix. Not changed on that basis.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


zourzouvillys:
CI is green (50 passing, 2 skipped). Two earlier failures, for the record:

**`@clerk/expo` unit tests** — `ClerkProvider.nativeClientSync` → *"recovers again inside the cooldown window once native pushes a new device token"* → `expected vi.fn() to be called 2 times, but got 1`. A timer assertion in a package this change does not touch; the full monorepo suite passed locally (46/46 tasks). It passed on re-run with no code change, so it is flaky rather than caused by this PR — worth a look on its own if it recurs.

**bundlewatch** — real, and fixed in the last commit. The sign-in chunk's largest locale variant was already within **71 bytes** of the 17KB ceiling on `main` (17,337 gzipped); this change adds **226 bytes**, taking it to 17,563. Raised to 18KB.

That delta was measured by building `@clerk/ui` at `origin/main` and at this branch and gzipping each `dist/signin*.js`, rather than from the CI report — the locale hashes in those filenames differ between builds, so comparing them by name gives a meaningless number.

packages/react/src/isomorphicClerk.ts
```diff
@@ -1595,6 +1596,17 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
...
+    const callback = () => this.clerkjs?.__internal_resumeAfterProtectCheck(params);
+    if (this.clerkjs && this.loaded) {
+      void callback()?.catch(() => {
+        // Same React 18 strict-mode double-mount caveat as handleRedirectCallback above.
+      });
+    } else {
+      this.premountMethodCalls.set('__internal_resumeAfterProtectCheck', callback);
+    }
+  };
+
```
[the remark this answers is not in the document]
zourzouvillys: Fixed in 238451c6be, with one change from the suggested diff: the `.catch()` is hoisted into the shared `callback` rather than written once per arm, so the loaded and queued paths cannot drift apart — which is the failure this PR exists to fix, one level up.

Covered by two tests in `packages/react/src/__tests__/isomorphicClerk.test.ts`, one per arm; both were verified by removing the handler and watching them go red.

While in here, a follow-up review caught that this method declared `customNavigate` but forwarded only `params`, so a call through `ClerkProvider` silently fell back to `Clerk.navigate` and resolved component-relative destinations against the origin. Fixed in 97178d1d00 and now asserted by the same two tests.

packages/react/src/isomorphicClerk.ts
```diff
@@ -1595,6 +1596,21 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
...
+  __internal_resumeAfterProtectCheck = async (
+    params?: ResumeAfterProtectCheckParams,
+    customNavigate?: (to: string) => Promise<unknown>,
+  ): Promise<void> => {
+    // Caught inside `callback` so the queued arm carries it too: replayInterceptedInvocations
+    // discards what its callbacks return, leaving a rejection there with nobody to reach.
+    const callback = () =>
+      void this.clerkjs?.__internal_resumeAfterProtectCheck(params, customNavigate)?.catch(() => {});
+    if (this.clerkjs && this.loaded) {
+      callback();
```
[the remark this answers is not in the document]
zourzouvillys: Half right, and fixed in 79131d0c7a for the half that holds.

**Correct:** `IsomorphicClerk` always exposes this wrapper, so it cannot itself signal whether the loaded runtime supports the call, and calling straight through threw a `TypeError` rather than doing nothing. It is now presence-checked, the same way `__internal_windowNavigate` a few hundred lines above already is — same category of problem (a newer method an older runtime lacks), same idiom. Covered by a new test that points the wrapper at a `clerkjs` without the method; removing the guard fails it with exactly that `TypeError`.

**Not correct:** that the UI feature check is defeated by this. `packages/ui` never sees an `IsomorphicClerk` — clerk-js constructs the UI itself with `new ClerkUI(() => this, ...)` (`packages/clerk-js/src/core/clerk.ts:609`), so `useClerk()` inside the prebuilt UI resolves to the real `Clerk`. The `typeof` check in `SignInProtectCheck` therefore tests exactly the object whose version can skew, which is the case it was added for, and its `navigate(..)` fallback is reached rather than bypassed. `SignInProtectCheck.test.tsx` covers it by removing the method from the clerk fixture and asserting the fallback destination; without the check that test fails with `TypeError: __internal_resumeAfterProtectCheck is not a function`.

So no capability needed to be plumbed out to the caller — but the wrapper should not have been a landmine for a host caller either, and no longer is.

packages/clerk-js/src/core/__tests__/clerk.test.ts
```diff
@@ -1850,6 +1850,216 @@ describe('Clerk singleton', () => {
...
+        }
+        sut.client.signUp.create = mockSignUpCreate;
+        sut.setActive = vi.fn();
+
+        await sut.__internal_resumeAfterProtectCheck({ continuation: 'transfer_to_sign_up' });
+
+        await waitFor(() =>
+          expect(mockSignUpCreate).toHaveBeenCalledWith({ transfer: true, unsafeMetadata: undefined }),
+        );
+        expect(mockNavigate).not.toHaveBeenCalledWith(expect.stringContaining('protect-check'), expect.anything());
```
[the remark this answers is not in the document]
zourzouvillys: Correct, and worse than "can pass" — that assertion could **never** fail, on any input.

The wrapper builds `const navigate = (to: string) => ...` (`packages/clerk-js/src/core/clerk.ts:2805-2807`), so `mockNavigate` is only ever called with one argument on this path. `expect.anything()` does not match a missing argument, so `not.toHaveBeenCalledWith(str, expect.anything())` was vacuously true whatever the code did — it would have sailed through the exact regression it was written to catch.

Fixed in 79131d0c7a at both sites (the pre-existing sign-in one and the sign-up variant added in this PR), inspecting the first argument across `mockNavigate.mock.calls` as suggested.

.changeset/resume-oauth-transfer-after-protect-check.md
```diff
@@ -0,0 +1,14 @@
...
+
+The card now hands back to the redirect-callback router, which resumes from where it stopped rather than starting over. The pending transfer is latched before the challenge runs, so it survives a response that re-serializes the sign-in without it.
+
+Also fixed alongside it: a stale or direct visit to the sign-in `protect-check` route now returns to the start of the flow instead of rendering an empty card, matching the sign-up side; and a failure in the SSO callback now shows a message and recovers, where previously the error handler could throw out of its own `catch` and leave the page loading indefinitely.
```
dstaley:
```suggestion
Fixes an issue where OAuth account transfers that needed additional verification were returned to the beginning of sign-in.
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2450,15 +2451,22 @@ export class Clerk implements ClerkInterface {
...
+      resuming = false,
     }: {
       signIn: SignInResource;
       signUp: SignUpResource;
       navigate: (to: string) => Promise<unknown>;
+      /**
+       * Set when this is re-entered after a verification challenge the caller has already
+       * cleared, which skips the two gate short-circuits below. Without it the resumed flow
+       * bounces straight back into the card it was resumed from.
+       */
```
dstaley:
```suggestion
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2602,14 +2610,18 @@ export class Clerk implements ClerkInterface {
     // sign-in's challenge. We only consult `si` here unless this is explicitly a sign-up callback.
     // Transfers are unaffected: the `signIn.create({ transfer })` path below checks its own fresh
     // response for the gate.
-    if (params.reloadResource !== 'signUp' && (si.protectCheck || si.status === 'needs_protect_check')) {
+    //
+    // Both gate checks are skipped when `resuming`: the caller IS the challenge card, so
+    // re-checking would either bounce control back into it, or — through the sign-up arm
+    // below, on a stale gate — hand it to the wrong card entirely.
```
dstaley:
```suggestion
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2669,7 +2681,11 @@ export class Clerk implements ClerkInterface {
       return navigateToResetPassword();
     }
 
-    const userNeedsToBeCreated = si.firstFactorVerificationStatus === 'transferable';
+    // `SignIn.fromJSON` replaces `firstFactorVerification` wholesale on every write, so a
+    // caller that observed the pending transfer BEFORE clearing a challenge cannot rely on
+    // the marker still being here afterwards. It tells us instead of us re-reading it.
```
dstaley:
```suggestion
```

packages/clerk-js/src/core/clerk.ts
```diff
@@ -2772,6 +2788,32 @@ export class Clerk implements ClerkInterface {
...
+    if (!this.loaded || !this.environment || !this.client) {
+      return;
+    }
+    const { signIn, signUp } = this.client;
+
+    // Deliberately mirrors `handleRedirectCallback` rather than
+    // `__internal_handleResourceCallback`: the latter runs every path through
+    // `buildUrlWithAuth`, which resolves a relative path against the ORIGIN on development
+    // instances and so turns `../factor-one` into an absolute URL, losing the component
+    // router's context.
```
dstaley:
```suggestion
```

packages/react/src/isomorphicClerk.ts
```diff
@@ -1595,6 +1596,29 @@ export class IsomorphicClerk implements IsomorphicLoadedClerk {
...
 
+  __internal_resumeAfterProtectCheck = async (
+    params?: ResumeAfterProtectCheckParams,
+    customNavigate?: (to: string) => Promise<unknown>,
+  ): Promise<void> => {
+    // Caught inside `callback` so the queued arm carries it too: replayInterceptedInvocations
+    // discards what its callbacks return, leaving a rejection there with nobody to reach.
+    // Presence-checked like __internal_windowNavigate above: this wrapper always exists, but an
+    // older clerk-js has no such method, and calling straight through would throw rather than
+    // no-op. The prebuilt UI never reaches this path -- clerk-js hands it the real Clerk.
```
dstaley:
```suggestion
```

packages/ui/src/common/SSOCallback.tsx
```diff
@@ -23,18 +23,45 @@ export const SSOCallbackCard = (props: HandleOAuthCallbackParams | HandleSamlCal
...
   const card = useCardState();
 
+  const bounceTimeoutRef = React.useRef<ReturnType<typeof setTimeout> | undefined>(undefined);
+
   React.useEffect(() => {
-    let timeoutId: ReturnType<typeof setTimeout>;
+    // Cleanup runs while `handleRedirectCallback` is still pending, so a superseded run's `.catch`
+    // fires AFTER it. Clearing a stored id cannot help -- the stale timer does not exist yet. The
+    // run has to know it was superseded and decline to schedule at all, or its bounce pulls the
+    // user off the route the newer run just reached.
```
dstaley:
```suggestion
```

packages/ui/src/common/SSOCallback.tsx
```diff
@@ -23,18 +23,45 @@ export const SSOCallbackCard = (props: HandleOAuthCallbackParams | HandleSamlCal
...
+
+        // Schedule the bounce FIRST, and never let the error reporting escape this handler.
+        //
+        // `handleError` re-throws anything it does not recognise, and the callback's own
+        // "did not complete" guards throw a plain `Error` — which it does not. A throw from
+        // inside this `.catch` skipped BOTH statements below, so the user got no message and
+        // no recovery: the card sat on its spinner indefinitely while the failure surfaced
+        // only as an unhandled rejection in the console. Every callback dead-end was
+        // invisible for that reason, which is a bad property for a route whose whole job is
+        // to be the last step of somebody's sign-in.
```
dstaley:
```suggestion
```

packages/ui/src/components/SignIn/handleProtectCheck.ts
```diff
@@ -36,3 +36,76 @@ export function navigateOnSignInProtectGate(
...
+
+/**
+ * Whether this sign-in is waiting to become a sign-up.
+ *
+ * An OAuth sign-in for an identity that has no account yet comes back as a *transferable*
+ * first-factor verification: the server has recorded the account transfer and the client is
+ * expected to complete it as a sign-up. It is not a sign-in that can continue on its own,
+ * and none of the interactive sign-in steps apply to it.
+ *
+ * Read this BEFORE clearing a gate, never after — see `resumeSignInAfterProtectCheck`.
```
dstaley:
```suggestion
```

packages/ui/src/components/SignIn/handleProtectCheck.ts
```diff
@@ -36,3 +36,76 @@ export function navigateOnSignInProtectGate(
...
+ * The exit choke point, and the counterpart to `navigateOnSignInProtectGate` above.
+ *
+ * The gate has two halves and both live in this file: one for routing *into* the challenge,
+ * one for routing *out* of it. A new caller needs both — a card that enters through the
+ * helper and then hand-rolls its own exit is exactly the shape that produced the outage this
+ * function was written for.
+ *
+ * `resumeOAuthContinuation` is how the card hands back to the redirect-callback router. It is
+ * injected rather than called directly so this module stays free of the Clerk instance.
+ */
```
dstaley:
```suggestion
```

packages/ui/src/components/SignIn/handleProtectCheck.ts
```diff
@@ -36,3 +36,76 @@ export function navigateOnSignInProtectGate(
...
+    startedAsOAuthTransfer,
+  }: {
+    navigate: (to: string) => Promise<unknown>;
+    resumeOAuthContinuation: () => Promise<unknown>;
+    startedAsOAuthTransfer: boolean;
+  },
+): Promise<unknown> {
+  // Chained challenge — stay here and re-run the new challenge on next render. Both
+  // signals are checked: `protectCheck` is the authoritative field, and
+  // `'needs_protect_check'` is the SDK-version-gated status.
```
dstaley:
```suggestion
```

I think there could be a useful condition to callout here, but I don't specifically understand what. What does it mean for a sign-in to be protect gated?

packages/ui/src/components/SignIn/handleProtectCheck.ts
```diff
@@ -36,3 +36,76 @@ export function navigateOnSignInProtectGate(
...
+      // Everything above is an interactive sign-in step the user can be shown. Anything
+      // else means this sign-in cannot continue on its own, and today that is an OAuth
+      // account transfer: `needs_identifier` carrying a transferable first-factor
+      // verification, whose continuation lives in the redirect-callback router.
+      //
+      // Returning to the start form instead is not merely a wrong destination — the start
+      // card renders the transfer's `external_account_not_found` error and then calls
+      // `signIn.create({})` to clear it, which replaces the attempt and discards the only
+      // reference to the pending transfer. The user is then stranded permanently, and every
+      // retry reproduces it.
```
dstaley:
```suggestion
```

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
```diff
@@ -19,50 +20,46 @@ import {
...
+  const { navigateToFlowStart } = useNavigateToFlowStart();
+  const clerk = useClerk();
+  const { setActive, __internal_resumeAfterProtectCheck } = clerk;
+  const ctx = useSignInContext();
+  const { afterSignInUrl, navigateOnSetActive } = ctx;
+
+  // Latched at mount, BEFORE the challenge is submitted. `SignIn.fromJSON` replaces
+  // `firstFactorVerification` wholesale on every write, so the transferable marker that routed
+  // us here is not guaranteed to survive `submitProtectCheck` — and it is the only thing that
+  // distinguishes "an OAuth sign-up is in progress" from "an ordinary gated sign-in".
```
dstaley:
```suggestion
  // persist the original status of whether the sign-in is pending an OAuth transfer
```

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
```diff
@@ -19,50 +20,46 @@ import {
...
+  const startedAsOAuthTransfer = useRef(isSignInPendingOAuthTransfer(signIn));
+
+  // Latches that a protect check existed at some point, so the resolution race
+  // (submitProtectCheck clearing protectCheck mid-navigation) isn't mistaken for a stale
+  // visit. Mirrors SignUpProtectCheck, which has had this since it shipped. State adjusted
+  // during render (guarded) rather than a ref write, which React disallows in the render body.
```
dstaley:
```suggestion
  // persist that a protect check existed at some point
```

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
```diff
@@ -85,7 +82,30 @@ function SignInProtectCheckInternal(): JSX.Element {
...
       }
-      await navigateNext(updatedSignIn, navigate);
+      await resumeSignInAfterProtectCheck(updatedSignIn, {
+        navigate,
+        startedAsOAuthTransfer: startedAsOAuthTransfer.current,
+        // No isCancelled() guard around this one: completing the transfer calls setActive,
+        // which flips the withRedirectToAfterSignIn guard and blanks this card. That unmount
+        // must not abort the continuation — the router owns its navigation from here.
+        // Feature-detected: @clerk/ui ships independently of the runtime, so a newer card can meet
+        // a clerk-js without this method. Degrade to the pre-existing destination rather than throw.
```
dstaley:
```suggestion
```

packages/ui/src/components/SignIn/SignInProtectCheck.tsx
```diff
@@ -85,7 +82,30 @@ function SignInProtectCheckInternal(): JSX.Element {
...
+        resumeOAuthContinuation: () =>
+          typeof __internal_resumeAfterProtectCheck === 'function'
+            ? __internal_resumeAfterProtectCheck(
+                {
+                  ...buildSignInOAuthCallbackParams(ctx),
+                  continuation: 'transfer_to_sign_up',
+                  // Carried for the same reason the social buttons carry it: without it a
+                  // completed transfer whose session has a pending task is routed with the
+                  // component's base URL rather than its mounted route, which lands on
+                  // `#/tasks/...` instead of `#/create/tasks/...` in the combined flow.
```
dstaley:
```suggestion
```

packages/ui/src/hooks/useProtectCheckRunner.ts
```diff
@@ -68,6 +68,18 @@ export interface ProtectCheckRunner {
 export function useProtectCheckRunner<TResource>(params: ProtectCheckRunnerParams<TResource>): ProtectCheckRunner {
   const card = useCardState();
 
+  // `handleError` re-throws what it does not recognise, and this runner awaits caller code that
+  // raises plain errors (a transient fetch failure, an OAuth continuation that did not complete).
+  // A throw from inside a catch escapes the void-invoked challenge run, leaving the card with no
+  // spinner, no message and no retry -- so the reporting itself must never throw.
```
dstaley:
```suggestion
```

dstaley (approved):
lgtm, just removed a bunch of comments that seemed a little too specific to the bug being solved rather than describing what the code does

packages/ui/src/components/SignIn/handleProtectCheck.ts
zourzouvillys: Trimmed to a one-liner in 8f8aea4.

On the question: "protect gated" is `isSignInProtectGated` at the top of this same file — the sign-in response came back needing a verification challenge to be cleared before it can continue, signalled either by the `protectCheck` field or by `status === 'needs_protect_check'`. That function already carries the explanation, so the essay here was restating it.

The condition worth calling out is the one you sensed: reaching this branch means the challenge we just cleared produced *another* one, so we stay on this route and let the next render run it, rather than routing onward.

PR #9516, opened 2026-08-20
feat(ui): rough in the user-profile account section flow

maxyinger:
## Description
- Creates `User profile account flow`
- Bakes half styled dialogs into view component, props passed down from machine to control them
- placeholder inputs
- mock state machine with network latency driving flows here

> [!NOTE]
> While the dialogs are baked into the view components, they're still driven my the controller/state machine, so clicking any of the buttons in the view components will not render them. If we want to add state machine wrappers in the views we can, but I went with just keeping it clean for now

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


maxyinger:
This was a quick reference pr. working through the smaller pieces more methodically rn

PR #9518, opened 2026-08-20
feat(*): add OAuth device verification flow

jeremy-clerk:
## Description

Adds an authenticated OAuth device verification flow across React, Next.js, Vue, Nuxt, and Astro, with a custom workflow hook and OAuth application lookup and decision methods.
The flow supports manual or prefilled device codes, organization-scoped approvals, localized consent and terminal states, and shared OTP input styling.

img_09_02_eu9cGAeZ
img_09_02_tWr6OuAh

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


wobsoriano (approved):
Awesome work 👏🏼 this looks good to me

packages/astro/src/internal/mount-clerk-astro-js-components.ts
```diff
@@ -21,6 +21,7 @@ const mountAllClerkAstroJSComponents = () => {
     waitlist: 'mountWaitlist',
     'pricing-table': 'mountPricingTable',
     'api-keys': 'mountAPIKeys',
+    'oauth-device-verification': '__internal_mountOAuthDeviceVerification',
```
jescalan, line 24: Adding this mapping exposes a mixed-version failure in the shared dynamic mount call: its optional chain guards the Clerk object, but it still calls an undefined method when `@clerk/astro` runs with a pinned or preloaded ClerkJS version from before this API. That throws inside `mountAllClerkAstroJSComponents()`, and because mounting runs before function replay and listener registration, one unsupported component aborts the rest of Astro initialization. React already feature-detects this exact case. Could we guard the selected method and add a loaded-old-ClerkJS test so this component no-ops or warns without breaking the remaining Clerk state setup?


<span title="This comment was written by AI, reviewed &amp; directed by Jeff">~ 🤖</span>

packages/ui/src/components/OAuthDeviceVerification/OAuthDeviceVerification.tsx
```diff
@@ -0,0 +1,523 @@
...
+      return {
+        title: 'oauthDeviceVerification.error.expiredTitle',
+        subtitle: 'oauthDeviceVerification.error.expiredSubtitle',
+        canReset: false,
+      };
+    case 'rateLimited':
+      return {
+        title: 'oauthDeviceVerification.error.rateLimitedTitle',
+        subtitle: 'oauthDeviceVerification.error.rateLimitedSubtitle',
+        canReset: false,
```
jescalan, line 494: The rate-limit copy tells the user to wait before trying another code, but `canReset: false` permanently removes the only recovery affordance. There is no timer or automatic retry, so even after the cooldown expires the user can recover only through an undocumented page reload. Could we provide a recovery path, preferably one that respects `Retry-After` or a countdown, or at minimum return to code entry?


<span title="This comment was written by AI, reviewed &amp; directed by Jeff">~ 🤖</span>

packages/shared/src/react/hooks/useOAuthDeviceVerification.tsx
```diff
@@ -0,0 +1,198 @@
...
+      }
+
+      const requestGeneration = generation.current;
+      setError(null);
+      setData(undefined);
+      setResult(undefined);
+      setIsLoading(true);
+
+      const request = oauthApplication
+        .lookupDeviceVerification({ ...params, userCode: key })
```
jescalan, line 107: Checking only that `oauthApplication` exists is not sufficient for mixed SDK/runtime versions: an older loaded ClerkJS has this resource but not the new device-verification methods. We set `isLoading` to true and then calling the missing method throws synchronously, before a promise or `finally` handler exists, leaving the hook stuck and violating its promise-returning contract. Could we feature-detect `lookupDeviceVerification` here and `submitDeviceVerification` in `submit` before mutating state, return `createNotReadyError()`, and add a version-skew test?


<span title="This comment was written by AI, reviewed &amp; directed by Jeff">~ 🤖</span>

PR #9520, opened 2026-08-21
docs(repo): update docs anchor links for decamelized slug removal

manovotny:
## Description

clerk.com stopped decamelizing heading anchors ([clerk/clerk#3161](https://github.com/clerk/clerk/pull/3161), [DOCS-11886](https://linear.app/clerk/issue/DOCS-11886)): `OAuthStrategy` now slugs to `#oauthstrategy` instead of `#o-auth-strategy`, `getToken()` to `#gettoken` instead of `#get-token`, and so on. This updates every docs deep link in this repo to the new form:

- `.typedoc/custom-plugin.mjs` — the fragment halves of `LINK_REPLACEMENTS` values and replace-text strings. Keys are typedoc's own file-name slugs and are untouched.
- JSDoc links across `packages/` (shared, nextjs, backend, react, vue) — these surface in IDE tooltips.
- `packages/upgrade` core-2 markdown and the typedoc test snapshots (regenerated via `vitest -u`, diff is anchors-only).

Verified by running `typedoc:generate`: the output contains **zero** old-form fragments, so once this lands, the docs typedoc sync produces converged content — this is also what turns clerk/clerk#3161's intentionally red docs build green. That PR merges first; the typedoc regen dispatch comes after.


packages/headless/package.json
SarahSoutoul, line 1: @manovotny This swaps the import and types condition order for every export. This seems unrelated to the anchor-link updates and may be a merge artifact. Was this intentional, or should this file retain the version currently on `main` when resolving the merge conflict?

manovotny: You're right — it was a formatter artifact from an earlier merge, not intentional. Resolved as you suggested in c2eed79ff0: the file takes main's version and is out of the PR diff entirely.

SarahSoutoul (approved):

PR #9523, opened 2026-08-21
fix(localizations): use future tense for nl-NL and nl-BE

jorisvdz:
The email and phone verification copy implied the code had already been sent. These screens send the code after the user continues, so the Dutch and Belgian strings now say it will be sent.


wobsoriano (approved):
thanks!

PR #9527, opened 2026-08-22
fix(clerk-js,react,shared,ui): bound loading a verification, not running one

zourzouvillys:
## Description

Before this change, the UI raced both verification-module loading and challenge execution against one fixed 60-second host timeout. A valid long-running challenge could therefore be aborted and reported as timed out; retrying restarted the same work from the beginning.

This PR limits the SDK-owned deadline to the dynamic module load. Loading remains bounded and abort-aware, with a 60-second default and a 600-second ceiling that prevents oversized timer values from overflowing. Once the module has loaded, the challenge owns its execution duration and the host imposes no hard deadline.

The optional `challenge_load_timeout_ms` environment field can override the load bound for an instance or an individual loader. Among the loaders applied to the current browser, the first finite, positive loader value wins; the instance value is used when no applied loader supplies one, and the SDK default is used when neither does. The selected loader value is exposed by clerk-js and forwarded through `IsomorphicClerk`, allowing `@clerk/ui` to read it across the CDN/package boundary while safely falling back when paired with an older clerk-js version that does not provide the getter.

An invisible challenge that never settles can now leave the host spinner visible indefinitely; once a challenge exposes a widget, that widget owns the progress UI. Challenge-aware soft status and recovery UX are tracked internally. When the optional backend configuration is absent, existing instances use the SDK default.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/clerk-js/src/core/protect.ts
+++ b/packages/clerk-js/src/core/protect.ts
@@ -8,6 +8,21 @@ import type { Environment } from './resources';
 export class Protect {
   #initialized: boolean = false;
   #session?: ProtectSession;
+  #challengeLoadTimeoutMs?: number;
+
+  /**
+   * The verification-module LOAD timeout asked for by the loader this browser was assigned, or
+   * undefined when none asked for one — in which case the caller falls back to the instance-wide
+   * value and then to the SDK default.
+   *
+   * Resolved from the APPLIED loaders rather than from the config, because rollout is decided by
+   * a random draw per page load: a loader being ramped can carry its own value without changing
+   * anything for browsers still on the loader it replaces, and which one a browser got cannot be
+   * recomputed afterwards.
+   */
+  get challengeLoadTimeoutMs(): number | undefined {
+    return this.#challengeLoadTimeoutMs;
+  }
 
   load(env: Environment): void {
     const config = env?.protectConfig;
@@ -39,6 +54,13 @@ export class Protect {
     // Rollout is decided before anything else, because the session is only meaningful for the
     // loaders we are actually going to apply.
     const loaders = configured.filter(loader => isLoader(loader) && inRollout(loader));
+
+    // Read off the applied set, before any of them run: this only describes config, so a loader
+    // that later fails to be placed has still spoken for the browser it was assigned to. First
+    // one that asks for it wins.
+    this.#challengeLoadTimeoutMs = loaders.find(
+      loader => typeof loader.challenge_load_timeout_ms === 'number' && loader.challenge_load_timeout_ms > 0,
+    )?.challenge_load_timeout_ms;
 
     // Only an instance whose loaders reference the correlation id gets a session, so an instance
     // not using it is unaffected and stores nothing in the browser.
--- a/packages/clerk-js/src/core/protect.ts
+++ b/packages/clerk-js/src/core/protect.ts
@@ -55,12 +55,9 @@ export class Protect {
     // loaders we are actually going to apply.
     const loaders = configured.filter(loader => isLoader(loader) && inRollout(loader));
 
-    // Read off the applied set, before any of them run: this only describes config, so a loader
-    // that later fails to be placed has still spoken for the browser it was assigned to.
-    //
-    // First one that asks for it wins, and the instance-wide value applies only when none does —
-    // the precedence is across the applied set rather than per loader, because there is no single
-    // "the" loader once more than one is live. Documented on the field.
+    // Load timeout applies to a single event and we only need one, so we pick the first valid one.
+    // This happens before applying the loaders since we want to use the value even if the loader
+    // later fails to apply.
     this.#challengeLoadTimeoutMs = loaders.find(
       loader =>
         typeof loader.challenge_load_timeout_ms === 'number' &&
--- a/packages/clerk-js/src/core/protect.ts
+++ b/packages/clerk-js/src/core/protect.ts
@@ -56,10 +56,16 @@ export class Protect {
     const loaders = configured.filter(loader => isLoader(loader) && inRollout(loader));
 
     // Read off the applied set, before any of them run: this only describes config, so a loader
-    // that later fails to be placed has still spoken for the browser it was assigned to. First
-    // one that asks for it wins.
+    // that later fails to be placed has still spoken for the browser it was assigned to.
+    //
+    // First one that asks for it wins, and the instance-wide value applies only when none does —
+    // the precedence is across the applied set rather than per loader, because there is no single
+    // "the" loader once more than one is live. Documented on the field.
     this.#challengeLoadTimeoutMs = loaders.find(
-      loader => typeof loader.challenge_load_timeout_ms === 'number' && loader.challenge_load_timeout_ms > 0,
+      loader =>
+        typeof loader.challenge_load_timeout_ms === 'number' &&
+        Number.isFinite(loader.challenge_load_timeout_ms) &&
+        loader.challenge_load_timeout_ms > 0,
     )?.challenge_load_timeout_ms;
 
     // Only an instance whose loaders reference the correlation id gets a session, so an instance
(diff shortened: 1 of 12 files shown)


packages/clerk-js/src/core/protect.ts
```diff
@@ -40,6 +55,19 @@ export class Protect {
     // loaders we are actually going to apply.
     const loaders = configured.filter(loader => isLoader(loader) && inRollout(loader));
 
+    // Read off the applied set, before any of them run: this only describes config, so a loader
+    // that later fails to be placed has still spoken for the browser it was assigned to.
+    //
+    // First one that asks for it wins, and the instance-wide value applies only when none does —
+    // the precedence is across the applied set rather than per loader, because there is no single
+    // "the" loader once more than one is live. Documented on the field.
```
Ephem: Typical AI-comment that just makes things harder to understand. Maybe:

```suggestion
    // Load timeout applies to a single event and we only need one,
    // so we pick the first valid one.
    // This happens before applying the loaders since we want to
    // use the value even if the loader later fails to apply.
```

packages/ui/src/hooks/useProtectCheckRunner.ts
```diff
@@ -238,20 +245,19 @@ export function useProtectCheckRunner<TResource>(params: ProtectCheckRunnerParam
...
+        // server-side, per decision, long after this bundle shipped, and its work may be waiting
+        // on a person or moving a server-chosen number of bytes over a link we know nothing
+        // about. The wall that used to be here aborted valid challenges and reported them to the
+        // user as a timeout, and since a re-run restarts the work from the beginning, retrying
+        // could never win on any connection slow enough to trip it in the first place.
+        const proofToken = await executeProtectCheck(protectCheck, container, {
+          signal: abortController.signal,
+          setWidgetVisible,
+          loadTimeoutMs,
+        });
```
Ephem, line 261: What's the UX like here if the challenge is actually stalled? I'm guessing it's just an infinite loading kinda thing? In some cases that might be fine, like if it's waiting for user input that never arrives, it's clear why things are not progressing, but in some cases I'm guessing this will just feel stuck?

Fine for now if intentional, but we might want to consider some kind of soft deadline, where we pop a "Still working"->"Things seem to be taking longer than expected" kinda thing? Could include manual Retry/Cancel that reveals after a while.

Perhaps each challenge needs different behavior too?

This is more forward-thinking polish thoughts. 😄

packages/ui/src/hooks/useProtectCheckRunner.ts
Ephem, line 309: Not related to this PR:

Relying on effect timing for something as important as this is always a bit sketchy. It also leads to some smells like all the different refs. A better approach here might be an explicit state machine, letting us truly own and reason about the lifecycle fully.

---

This is AI-generated and I haven't verified the details, but it seems like this hook is doing four jobs: start a challenge, survive expiry/reload, submit a proof, and route, so something like this:

**States**

| State | What it means |
|---|---|
| `idle` | No token, or waiting to start |
| `loading` | Dynamic `import()` in flight |
| `challenging` | Module has control (widget visible is a flag, not a state) |
| `reloading` | Expired check; GET in flight |
| `submitting` | Proof token on the way to FAPI |
| `resolving` | `onResolved` / `setActive` |
| `error` | Failed; retry is legal |
| `aborted` | Unmounted or superseded |

**Events**

- `TOKEN` — new or chained challenge token
- `RETRY`
- `UNMOUNT` / `SUPERSEDED` — token changed or `protectCheck` cleared mid-run
- `LOAD_OK` / `LOAD_FAILED`
- `PROOF` — module returned a token
- `EXPIRED` → then `RELOAD_REMINTED` / `RELOAD_STILL_EXPIRED` / `RELOAD_CLEARED`
- `SUBMIT_OK` / `ALREADY_RESOLVED` / `SUBMIT_FAILED`
- `RESOLVED`
- `WIDGET` — visibility handshake only; does not change state

`TOKEN` and `RETRY` are the only things that start work. `SUPERSEDED` is what the effect cleanup is doing today when deps change. `cancelled` vs `mountedRef` goes away: `resolving` ignores `SUPERSEDED` from a cleared `protectCheck`, and only `UNMOUNT` bails.

---

Not urgent or even necessary to tackle, just wanted to note a common source of bugs (especially as code changes over time), and an possible alternative.

We haven't tended to do this in the past which has both bit us before and means there aren't many examples to look at though. The new Mosaic components are moving in this direction and Mosaic even has a `useMachine` hook, but I'm not sure if it's usable outside that context (yet?). CC @alexcarpenter

packages/ui/src/hooks/useProtectCheckRunner.ts
```diff
@@ -68,6 +67,15 @@ export interface ProtectCheckRunner {
 export function useProtectCheckRunner<TResource>(params: ProtectCheckRunnerParams<TResource>): ProtectCheckRunner {
   const card = useCardState();
 
+  // Override for the module-LOAD bound only (see `executeProtectCheck`), resolved loader first
+  // and instance second: a loader being rolled out gradually can carry its own value without
+  // changing anything for browsers still on the loader it replaces. Undefined at both levels
+  // leaves the SDK default in force. Read here rather than inside the effect so the effect keeps
+  // depending on primitives.
+  const loaderTimeoutMs = useClerk().__internal_protectChallengeLoadTimeoutMs;
```
Ephem, line 76: A newer version of UI that has this code can ship with an older version of `clerk-js` that does not provide `__internal_protectChallengeLoadTimeoutMs`/`challenge_load_timeout_ms`. This should safely resolve to `undefined`, which has a fallback, so should be fine.

The more explicit we are about these things, the more top of mind it is which helps us avoid times when it might actually lead to issues, so this might warrant a comment:

```suggestion
  // Older versions of clerk-js might not have __internal_protectChallengeLoadTimeoutMs
  // and challenge_load_timeout_ms. This would safely resolve to undefined and later fallback.
  const loaderTimeoutMs = useClerk().__internal_protectChallengeLoadTimeoutMs;
```

Ephem (approved):
Looks good to me! Left a few comments but nothing blocking.

Ephem (commented):
Hmm, I think maybe `isomorphicClerk` needs the new `__internal_protectChallengeLoadTimeoutMs` too, or else it will likely fail to read and fall back to the default?

Update: Actually, maybe not. This is just used from UI so should read directly from clerk and not the isomorphic one. If we move SignIn/SignUp to be rendered inside the host app, like we are working on for the UserButton, that might change though and since this fails silently it would be hard to catch. So not necessary, but might still be a good idea. 😄

packages/ui/src/hooks/useProtectCheckRunner.ts
zourzouvillys: Yes: if a challenge stays invisible and never settles, the host spinner remains; once it exposes a widget, that widget owns the progress UI. Keeping the lack of a host-side hard deadline intentional here, and tracking challenge-aware soft status / Retry / Cancel UX separately as an internal follow-up.

packages/ui/src/hooks/useProtectCheckRunner.ts
zourzouvillys: Done in 01fd9c8 — added a terse compatibility note at the read. Older clerk-js versions still produce `undefined`, preserving the instance/default fallback.

packages/ui/src/hooks/useProtectCheckRunner.ts
zourzouvillys: Agreed that this should be evaluated separately from the timeout fix. Tracked the explicit lifecycle/state-machine follow-up in #9535, including the unmount-vs-superseded distinction and the chained/expired/already-resolved cases.

packages/clerk-js/src/core/protect.ts
zourzouvillys: Done in 01fd9c8 — replaced the longer block with the shorter explanation.

zourzouvillys:
Followed up on the remaining non-inline review items in 01fd9c8:

- Added `IsomorphicClerk.__internal_protectChallengeLoadTimeoutMs` as a forwarding getter, with coverage before and after the inner clerk-js instance loads. `@clerk/react` is included as a patch in the changeset. This is not required by today’s lazy-mounted SignIn/SignUp path, but avoids a silent fallback if those components move into the host app.
- Kept patch versioning for the mechanically reported API additions because they are optional/internal or wire-format fields in a bug fix.
- Did not add a redundant docstring solely for the automated coverage heuristic: `resolveLoadTimeoutMs` is a small normalization helper and the surrounding option/default/ceiling documentation carries the relevant rationale.

The stalled-progress UX and lifecycle/state-machine ideas are tracked internally.

PR #9534, opened 2026-08-24
feat(ui,headless): switch accounts from a flyout in the Mosaic UserButton

alexcarpenter:
## Description

The Mosaic `UserButton` popup no longer lists the other signed-in accounts inline under an **Accounts** heading, where an account read like one more workspace in the list above it.

**Switch account** at the foot opens them as a flyout instead: every signed-in account, the active one checked, with **+ Add account** at its end. Where there is nobody to switch to — one account, or a single-session instance — that row is **Add account** on its own, which is the row the flyout would have opened onto.

`menuItemOrder` names that slot by either `switchAccount` or `addAccount`, so one order places it whichever way it resolves.

Preview: https://swingset-git-carp-mosaic-user-button-switch-accounts.clerkstage.dev/user-button/user-button

### Per component

**`Menu` (`@clerk/ui`)**

- `Menu.Content` → `Menu.Popup`, matching `Popover.Popup`. `MenuContentProps` → `MenuPopupProps`.
- Popup capped at `min(18rem, calc(100vw - 2rem))`. Uncapped it grew to its longest item and nothing truncated.
- New `Menu.Media`: square leading column, centers an icon/avatar so every row's text starts on one line. `size` is `xs` or `sm`, default `sm`.
- New `Menu.Label`: the row's text. Takes the space between the media and whatever trails it, truncates to one line.
- `Menu.Item` `children` now required. Text goes in `Menu.Label`, not straight in the item — a bare text node is not a flex child the row can size. `label` is the typeahead / a11y name only.
- `menu.styles.ts` split into one `stylex.create` per slot (`positioner`, `popup`, `item`, `media`, `label`, `separator`), like `item.styles.ts`.

**`Menu` (`@clerk/headless`)**

- `sideOffset` takes `{ x, y }` as well as a number. The flyout row is inset 8px inside the card, so the sideways gap clears that before it clears the card's edge; above the row there is nothing to clear.
- `fallbackPlacements` on `Menu.Root`. The flyout opens out of the popup, so the opposite side is the popup itself and the menu has to be given somewhere else to land. `['left-start', 'top-start', 'bottom-start']` keeps the sideways flip on a desktop and drops the menu above the row on a phone.

**`UserButton`**

- Accounts move from the inline list to the flyout.
- `⋯` menu and flyout rows composed from `Menu.Media` + `Menu.Label`.

**`UserProfile`**

- Action menu, API keys revoke, MFA add: item text wrapped in `Menu.Label`.

**Swingset**

- Menu docs and stories rewritten for the slots and the rename; new `Accounts` story.
- Headless Menu docs gain `SideOffset` and `fallbackPlacements`.

`@clerk/ui` and `@clerk/headless` only, and the Mosaic `UserButton` has no public export path yet — Swingset is its only consumer — hence the empty changeset.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


Ephem (approved):
Looks good!

PR #9540, opened 2026-08-24
feat(user): localize email timestamps by timezone

tmilewski:
## Summary

Captures a timezone during authentication, persists it using trusted-device rules, and exposes the preference through Clerk's API and SDK surfaces. Customer-facing absolute timestamps render in the recipient's stored timezone with an explicit UTC fallback, so a new device cannot choose the timezone used in its own security notification.

## Changes in this repo

Reports the browser timezone only when creating authentication attempts and exposes timezone fields through ClerkJS resources, shared types, and Backend SDK user APIs.

## Companion PRs
- **clerk-sdk-go**: https://github.com/clerk/clerk-sdk-go/pull/597


packages/clerk-js/src/utils/index.ts
```diff
@@ -19,6 +19,7 @@ export * from '@clerk/shared/internal/clerk-js/queryStateParams';
 export * from '@clerk/shared/internal/clerk-js/querystring';
 export * from '@clerk/shared/internal/clerk-js/runtime';
 export * from './tokenId';
+export * from './timezone';
```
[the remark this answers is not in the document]
tmilewski: Removed the timezone barrel re-export and changed both SignIn and SignUp to import getBrowserTimezone directly from utils/timezone. The focused ClerkJS tests pass (207 tests).


<span title="This comment was written by AI, reviewed &amp; directed by Tom">~ 🤖</span>

packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -1067,12 +1084,14 @@ class SignInFuture implements SignInFutureResource {
       const identifier = params.identifier || params.emailAddress || params.phoneNumber;
       const previousIdentifier = this.#resource.identifier;
       const locale = getBrowserLocale();
+      const timezone = this.#resource.id ? null : (params.timezone ?? getBrowserTimezone());
```
joshrowley, line 1087: When `password()` follows a `create()` call, should we reuse `this.#resource.timezone` here instead of `null`? `password()` posts to `/client/sign_ins`, creating a new attempt.

Setting `timezone` to `null` omits the timezone from that request, so we could fall back to the edge timezone instead of preserving the value captured by the first `create()` call.

This is a non-blocker. I'm not as familiar with this code, but I just wanted to flag it for you in case it might be an issue.

joshrowley (approved):
Looks good. I want to flag one non-blocking issue just for you to investigate regarding when we send `null` up as the time zone in the password function.

packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -1036,13 +1051,15 @@ class SignInFuture implements SignInFutureResource {
...
     const { captchaToken, captchaWidgetType, captchaError } = await this.getCaptchaToken(params);
+    const timezone = params.timezone ?? getBrowserTimezone();
 
     const body: Record<string, unknown> = {
       ...params,
       captchaToken,
       captchaWidgetType,
       captchaError,
       locale: getBrowserLocale() || undefined,
+      ...(timezone !== null ? { timezone } : {}),
```
Ephem, line 1062: What happens if we pass this to an API version that does not support it? My understanding is that the API will reject with `422 form_param_unknown` (but might be wrong)?

I know we would release the APIs first, but I'm thinking about if we can somehow preserve the ability to do API rollbacks. Imagine the case where we roll this out API first, then SDKs, but then for some other unrelated reason someone does an API rollback? If I'm right, that could break sign-in.

Not sure how we usually handle this. Doesn't necessarily need to be code-level, we could also release APIs well in advance, be loud about it etc as long as we are aware (if the risk actually exists that is).

packages/clerk-js/src/core/resources/SignUp.ts
```diff
@@ -186,7 +201,7 @@ export class SignUp extends BaseResource implements SignUpResource {
   prepareVerification = (params: PrepareVerificationParams): Promise<this> => {
     debugLogger.debug('SignUp.prepareVerification', { id: this.id, strategy: params.strategy });
     return this._basePost({
-      body: params,
+      body: withoutTimezone(params),
```
Ephem, line 204: Just for my understanding: The types does not allow `timezone` to be in params, so the `withoutTimezone` here isn't strictly necessary if you are conforming to those? Since you already had `withoutTimezone` here you added it as a cheap extra runtime guard?

(I'm trying to understand why it's used here and not in SignIn.)

Ephem (approved):
Left a comment/question on the ability to roll back APIs, but looks good to me, nice work!

packages/clerk-js/src/core/resources/SignIn.ts
tmilewski: You're correct here. It now keeps the value of the earlier `create()`, then falls back to the browser TZ.

packages/clerk-js/src/core/resources/SignUp.ts
tmilewski: Yes, it's a runtime guard on top of the types. The asymmetry is deliberate: `SignUpUpdateParams` is `Omit<SignUpCreateParams, 'timezone'>`, so callers often reuse one params object for `create()` and then `update()`/`prepareVerification()`, and untyped or `as any` callers can slip `timezone` through. On the FAPI side only sign-up *create* reads `timezone`. The PATCH, prepare, and attempt endpoints check their params strictly, so a leaked `timezone` would 422 with `form_param_unknown`. It also keeps the value create-only, so a later update can't overwrite it. SignIn doesn't have this problem because its non-create methods take their own param types, which never included `timezone`.

<span title="This comment was written by AI, reviewed &amp; directed by Tom">~ 🤖</span>

packages/clerk-js/src/core/resources/SignIn.ts
tmilewski: Yeah, I think releasing the API well in advance and being loud about it is the way to go at this point.

Ephem: Sounds good!

packages/clerk-js/src/core/resources/SignUp.ts
Ephem: But this uses `PrepareVerificationParams`, attempt uses `AttemptVerificationParams` and ticket uses `SignUpFutureTicketParams` neither of which has `timezone`? SignIn has all of these too but without `withoutTimezone` and that's the asymmetry that confused me.

`withoutTimezone` is also used in create, but not because it's dangerous there, but for product reasons ("so an existing sign-up's values are not overwritten on update.").

So I guess the only place it's doing real heavy lifting is in `update` if that endpoint doesn't accept `timezone` (since leaving `withoutTimezone` out would break it)?

---

The reason I wanted to understand this is that this originally started out as a comment about it seeming brittle to include `timezone` by default and rely on knowing to use `withoutTimezone` to remove it (that is, opt-out instead of opt-in to including it). The fact that we use `withoutTimezone` _for different reasons_ throughout this file, but it being unclear from just reading the code what work it does in the different places makes it extra confusing to try to understand if and where it's actually important and not.

This is very much a NIT though, changing to opt-in for including it is a bit annoying and I don't mind the current approach too much. Adding it to signIn just seems like noise. Might be worth a comment on the update path to say kinda "This one is actually important" though to make clear that one is special?

tmilewski: Thank you, I wound up removing it wholesale. We're safe at this point if it does go through, though it won't be saved. I've added JSDoc for `upsert` given `timezone` can be passed but it's a bit unclear as to when it's stored.

Ephem: Nice, looks good to me. 👍 Was never a big deal but wanted to explain why I was digging into it. 😄

PR #9542, opened 2026-08-25
feat(backend,shared): Add AgentAction resource types

wyattjoh:
## Description

Applications that let an agent act for a user need to ask, before the agent does something consequential, whether policy allows it or whether a human has to approve first. `clerk.policy` is that surface in `@clerk/backend`. It is being built as four stacked PRs, and this is the bottom one: data types only, so nothing in the package calls them yet.

`parameters` and each `parametersDisplay[].key` are copied through `fromJSON` exactly as the API sent them. The key spelling an application passes to `check()` is the same spelling its policy rules use to address that value, so camelCasing `refund_amount` here would produce a rule that quietly never matches. It is the one place the package's usual camelCase habit is wrong.

The status union lands in `@clerk/shared` rather than `@clerk/backend` because the approval review UI projects the same column and the two should not drift.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/olive-pugs-smile.md
+++ b/.changeset/olive-pugs-smile.md
@@ -1,0 +1,5 @@
+---
+'@clerk/shared': minor
+---
+
+Add the experimental `AgentActionStatus` type, the lifecycle status of an agent action awaiting a policy decision. It is declared here so that `@clerk/backend` and the approval review surface share one definition rather than each carrying their own.
--- a/.changeset/olive-pugs-smile.md
+++ b/.changeset/olive-pugs-smile.md
@@ -1,5 +1,0 @@
----
-'@clerk/shared': minor
----
-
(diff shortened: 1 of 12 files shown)


wyattjoh:
Closing without merge. This layer was built against the SDK check contract that ADR 0018 (2026-09-02) superseded, and ADRs 0019/0020 (2026-09-08) changed it further: `CheckParams` is now one shape with no party field, every party derives server-side from the exchanged token, there is no `getStatus`/`waitForApproval`, and the check response is a narrower projection. The 2026-09-08 stack-health audit found the diff no longer matches `docs/spec/impl-sdk.md` as amended, and the credential seam it depends on is still undecided (AIE-1652).

Both SDK layers will be re-cut on fresh `main` once AIE-1652 ratifies. The branch is kept as reference and is not deleted. Nothing here was reviewed, so nothing is lost.

PR #9550, opened 2026-08-25
feat(backend): add clerk.policy.check

wyattjoh:
## Description

Chunk 2 of four building `clerk.policy` in `@clerk/backend`, on top of #9542. That PR landed the resource types with no consumer; this one registers `PolicyAPI` and ships `check()`, which asks whether policy allows an agent's proposed operation and returns the `AgentAction` recording the answer.

Three things the diff doesn't explain:

- **It returns errors instead of throwing**, alone among the package's endpoint classes. A policy decision and a transport failure are different questions and a caller branches on both. The `{ data, errors }` union is restored by unwrapping the legacy wrapper inside the class — `buildRequest` is untouched.
- **`actorId` is always caller-supplied, never derived.** An inbound token's client id is ambiguous between the party that exchanged it and the party the human authorized, and only the application knows which. The `oauth_token` arm derives `subjectId` and `authorizedClientId`, which its token verifiably names, and accepts neither. There is no session arm.
- **`parameters` keys reach the API byte-for-byte.** The spelling passed to `check()` is the spelling the field registry and every policy leaf must use; a mismatch is a rule that never matches and never errors.

Reads and `waitForApproval` follow in the two PRs above.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/shaggy-melons-repeat.md
+++ b/.changeset/shaggy-melons-repeat.md
@@ -1,0 +1,20 @@
+---
+'@clerk/backend': minor
+---
+
+Add the experimental `clerk.policy.check()` method, which checks an agent's proposed operation against your instance's policy and returns the `AgentAction` recording what was decided. It never blocks and never throws on an API outcome: transport and validation failures arrive on an `errors` arm, and a policy outcome that needs a human arrives as `status: 'pending'` with an `approval.url` to send them to.
+
+```ts
+const { data, errors } = await clerkClient.policy.check({
+  auth, // an authenticated `oauth_token` auth object
+  actorId: 'https://acme.example.com/mcp',
+  operation: 'api/v1/refund',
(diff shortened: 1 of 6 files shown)


wyattjoh:
Closing without merge. This layer was built against the SDK check contract that ADR 0018 (2026-09-02) superseded, and ADRs 0019/0020 (2026-09-08) changed it further: `CheckParams` is now one shape with no party field, every party derives server-side from the exchanged token, there is no `getStatus`/`waitForApproval`, and the check response is a narrower projection. The 2026-09-08 stack-health audit found the diff no longer matches `docs/spec/impl-sdk.md` as amended, and the credential seam it depends on is still undecided (AIE-1652).

Both SDK layers will be re-cut on fresh `main` once AIE-1652 ratifies. The branch is kept as reference and is not deleted. Nothing here was reviewed, so nothing is lost.

PR #9555, opened 2026-08-26
feat(ui): Add Destructive block and wire the delete account section

alexcarpenter:
## Description

Adds a `Destructive` block: type-to-confirm dialog for actions that cannot be undone. Modeled on Kumo's delete-resource block. Controlled, like any dialog. The block owns only the typed phrase; `open`, `isDeleting` and `errorMessage` stay with the caller.

Uses it in the user profile danger zone. A state machine (`idle → confirming → deleting → deleted`) drives the dialog: resolve closes it, reject keeps it open and renders the reason under the field.

Also fixes `SubmitButton`, which erased the `aria-disabled` that `focusableWhenDisabled` sets when the button is disabled but not pending.

Swingset gets a `Blocks` group with a `Destructive` page, plus an error story for the section.

**Follow-up PR: reverification.** Deleting an account needs step-up auth. Legacy does `useReverification(() => user?.delete())`. The controller does not exist yet, so nothing calls `user.delete()` here. When it lands, reverification either resolves inside the injected promise (machine unchanged) or needs its own state.

Previews:
- https://swingset-git-carp-mosaic-user-profile-delete-account.clerkstage.dev/blocks/destructive
- https://swingset-git-carp-mosaic-user-profile-delete-account.clerkstage.dev/user-profile/user-profile-delete-section

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


austincalvelage (approved):
Looks good!

PR #9585, opened 2026-08-26
fix(shared): type JWT aud as string or string array

thiskevinwang:
Manually tested w/

```
pnpm --filter @clerk/backend test:node src/tokens/__tests__/verify.test.ts

```

## Description

Type the JWT `aud` claim as an optional `string | string[]` on `JwtPayload` and the deprecated `ClerkJWTClaims` / `JWTClaims` types, matching RFC 7519. Clerk-issued OAuth access tokens may include a single RFC 8707 resource URI as a string; the claim remains optional for legacy and non-resource-bound tokens.

`decodeJwt` and `verifyJwt` already validate string and array audiences. This change exposes that shape on the public payload type (it previously resolved to `unknown` via the index signature) and covers matching and mismatched string resource audiences in OAuth JWT machine verification.

Linear: AIE-1629

## Checklist

- [ ] (If applicable) [Documentation](https://clerk.com/docs) has been updated

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

Linear Issue: [AIE-1629](https://linear.app/clerk/issue/AIE-1629/javascript-type-rfc-8707-audience-claims-in-jwt-payloads)

<div><a href="https://cursor.com/agents/bc-14a35554-dc9c-44e9-ac20-6c10844ed864?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-14a35554-dc9c-44e9-ac20-6c10844ed864&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


packages/backend/src/api/resources/__tests__/M2MToken.test.ts
```diff
@@ -105,6 +105,21 @@ describe('M2MToken', () => {
       expect(token.scopes).toEqual(['scope1', 'scope2', 'scope3']);
     });
 
+    it('seeds scopes from a string aud claim', () => {
```
thiskevinwang, line 108: This is backfilling coverage from existing behavior

packages/backend/src/tokens/__tests__/verify.test.ts
```diff
@@ -569,6 +569,60 @@ describe('tokens.verifyMachineAuthToken(token, options)', () => {
...
+      });
+
+      const result = await verifyMachineAuthToken(oauthJwt, {
+        apiUrl: 'https://api.clerk.test',
+        secretKey: 'a-valid-key',
+        audience,
+      });
+
+      expect(result.tokenType).toBe('oauth_token');
+      expect(result.data).toBeDefined();
```
thiskevinwang: toMatchInlineSnapshot

packages/backend/src/api/resources/JSON.ts
```diff
@@ -962,6 +962,7 @@ export interface IdPOAuthAccessTokenJSON extends ClerkResourceJSON {
   expiration: number | null;
   created_at: number;
   updated_at: number;
+  aud?: string | string[];
```
thiskevinwang, line 965: 👀 This type change could use a second pair of eyes.

packages/shared/src/types/jwt.ts
```diff
@@ -77,6 +77,11 @@ export interface ClerkJWTClaims {
    */
   azp?: string;
 
+  /**
+   * JWT Audience - [RFC7519#section-4.1.3](https://tools.ietf.org/html/rfc7519#section-4.1.3).
+   */
+  aud?: string | string[];
```
dominic-clerk, line 83: **[MEDIUM] The absent-`aud` case this optionality introduces is not handled — the audience check silently no-ops**

(The actual sink is `assertAudienceClaim` in `packages/backend/src/jwt/assertions.ts:10-25`, which this PR doesn't touch, so I'm anchoring here on the `aud?` typing that makes the case explicit.) `assertAudienceClaim` is the only runtime audience comparison in the SDK — reached from `verifyJwt` → `resolveKeyAndVerifyJwt` → `verifyOAuthJwt` → `verifyMachineAuthToken`/`authenticateRequest` — and it computes `shouldVerifyAudience = audienceList.length > 0 && audList.length > 0`, then returns early when false. So a token with no `aud` passes a check the caller explicitly requested. That's reachable, not theoretical: clerk_go sets `aud` only when a `resource` param was granted (`pkg/oauth2openid/strategy.go:149-150`) and its own test asserts `assert.NotContains(t, accessClaims, "aud")` when `resource` is omitted, so *any* registered OAuth client of the instance can mint an audience-less token by simply not sending `resource`. A resource server hardened with `verifyMachineAuthToken(token, { secretKey, audience: 'https://b.example.com' })` will then accept a token authorized for a different client — the confused-deputy case RFC 8707 exists to prevent. The new tests cover matching and mismatched `aud` but not the missing case, so it isn't visible from the suite. Suggest failing closed for the machine/OAuth JWT path when the caller passes a non-empty `audience` and the token carries no `aud` (the session-token path can keep relying on `azp`).

packages/backend/src/api/resources/IdPOAuthAccessToken.ts
```diff
@@ -25,6 +25,8 @@ export class IdPOAuthAccessToken {
     readonly createdAt: number,
     /** The Unix timestamp (in milliseconds) when the access token was last updated. */
     readonly updatedAt: number,
+    /** The intended audience for the access token. */
+    readonly aud?: string | string[],
```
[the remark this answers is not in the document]
thiskevinwang: fixed. [JSON.ts](https://github.com/clerk/javascript/blob/8c62389e12f2754b5dfcd0bcbd3796315de92a2c/packages/backend/src/api/resources/JSON.ts#L965) accepts aud?: string | string[]; [fromJSON](https://github.com/clerk/javascript/blob/8c62389e12f2754b5dfcd0bcbd3796315de92a2c/packages/backend/src/api/resources/IdPOAuthAccessToken.ts#L45) forwards data.aud.

packages/shared/src/types/jwt.ts
thiskevinwang: Planning to update the old `assertAudienceClaim` behavior in another PR, https://github.com/clerk/javascript/pull/9724

Keeping this PR as type-only.

packages/backend/src/api/resources/IdPOAuthAccessToken.ts
```diff
@@ -40,6 +43,7 @@ export class IdPOAuthAccessToken {
       data.expiration,
       data.created_at,
       data.updated_at,
+      data.aud,
```
dominic-clerk, line 46: **[MEDIUM] The opaque `oat_` path surfaces `aud` but never compares it, so the `audience` option is silently a no-op for the default token format**

This line makes `aud` available on the opaque-token result, but nothing on that path checks it. `verifyMachineAuthToken` routes `oat_` tokens to `verifyOAuthToken`, which returns whatever `client.idPOAuthAccessToken.verify()` gave back and never reads `options.audience`; `IdPOAuthAccessTokenApi.verify()` sends only `access_token`, so the server cannot enforce it either. Grepping `audience` across `packages/backend/src` shows `assertAudienceClaim` in `verifyJwt` is the only runtime consumer, which is why the JWT path in this PR rejects a mismatched resource audience and the opaque path does not.

A resource server B that calls `authenticateRequest(req, { acceptsToken: 'oauth_token', audience: 'https://b.example.com' })` will accept an opaque token minted for resource server A and hand back `machineData.aud === ['https://a.example.com']` as if it had been validated. That is the confused-deputy case RFC 8707 exists to prevent, and whether it bites depends only on which token format the instance happens to emit. The new opaque-token test asserts `data.aud` round-trips without ever passing `audience`, which makes the field read as checked when it is only echoed.

Running the returned `aud` through `assertAudienceClaim` in `verifyOAuthToken` when `options.audience` is set would put both formats back on the same control.

thiskevinwang:
!snapshot

thiskevinwang:
!snapshot

packages/backend/src/api/resources/IdPOAuthAccessToken.ts
```diff
@@ -19,12 +20,14 @@ export class IdPOAuthAccessToken {
...
     readonly revocationReason: string | null,
     readonly expired: boolean,
-    /** The Unix timestamp (in milliseconds) when the access token expires. */
+    /** The Unix timestamp (in seconds) when the access token expires. */
     readonly expiration: number | null,
-    /** The Unix timestamp (in milliseconds) when the access token was created. */
+    /** The Unix timestamp (in seconds) when the access token was created. */
     readonly createdAt: number,
-    /** The Unix timestamp (in milliseconds) when the access token was last updated. */
+    /** The Unix timestamp (in seconds) when the access token was last updated. */
```
thiskevinwang, line 27: These are indeed `seconds`

```ts
// import { createClerkClient, type ClerkClient } from "@clerk/backend";
      const res = await this.clerk.idPOAuthAccessToken.verify(token);
      console.log(res);
      // _IdPOAuthAccessToken {
      //   id: 'oat_3Je6nsrhV3q0O0iaa1XoIZQaG8j',
      //   clientId: 'Ggdsp1aBiyjnuXmO',
      //   type: undefined,
      //   subject: 'user_35foINd28ksQr65MyKacPf0APEZ',
      //   scopes: [ 'email', 'offline_access', 'openid', 'profile' ],
      //   revoked: false,
      //   revocationReason: null,
      //   expired: false,
      //   expiration: 1790095756,
      //   createdAt: 1790009356,
      //   updatedAt: 1790009356,
      //   aud: [ 'http://localhost:8787/mcp' ]
      // }
```

packages/backend/src/api/resources/IdPOAuthAccessToken.ts
```diff
@@ -3,6 +3,7 @@ import type { JwtPayload } from '@clerk/shared/types';
 import type { IdPOAuthAccessTokenJSON } from './JSON';
 
 type OAuthJwtPayload = JwtPayload & {
+  aud?: string | string[];
```
thiskevinwang, line 6: Raw JWT access token can hold `string` or `string[]` (or omitted)

And the SDK will normalize it to `string[]` (or omitted)

packages/backend/src/api/resources/IdPOAuthAccessToken.ts
thiskevinwang: To be handled in https://github.com/clerk/javascript/pull/9724

wobsoriano (approved):

PR #9588, opened 2026-08-27
fix(electron): retain client token when persistence fails

jeremy-clerk:
## Description

Keep the latest client token available when Electron cannot persist it securely, allowing the current authentication flow to continue without writing plaintext tokens to disk.

Successful writes continue using persisted storage as the source of truth. The in-memory fallback is removed when persistence recovers or the token is cleared.

Fixes the same issue as https://github.com/clerk/javascript/pull/9584 but handles it in the main process to avoid multiple renderer window collisions & IPC races

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/bright-dogs-remember.md
+++ b/.changeset/bright-dogs-remember.md
@@ -1,0 +1,5 @@
+---
+'@clerk/electron': patch
+---
+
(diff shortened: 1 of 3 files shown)


wobsoriano (approved):
Lgtm to me

PR #9590, opened 2026-08-27
feat(shared,js): add directory sync resource and organization contract

kalafut:
## Description

Part 1 of 4 of the self-serve Directory Sync stack. Stacked on `main`; this PR carries the changeset covering the whole stack, and the stack will be squashed on merge.

Adds the `DirectorySync`/`DirectorySyncUser` types and resources, plus `getDirectorySync` and `createDirectorySync` on the `Organization` contract, hitting `.../enterprise_connections/{id}/directory`. Mutations live on the returned `DirectorySync` resource (`update`, `rotateToken`, `delete`, `getUsers`) rather than on `Organization`, following review feedback and mirroring the `OrganizationDomain` shape. The SCIM bearer token is only present on the resources returned by create and rotate, and is deliberately excluded from snapshots. Also adds the `self_serve_directory_sync` user-settings flag (absent on older backends, defaulting to `false`).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/resources/Organization.ts
```diff
@@ -274,6 +283,95 @@ export class Organization extends BaseResource implements OrganizationResource {
...
+      search: convertPageToOffsetSearchParams(params),
+    });
+
+    const payload = res?.response as unknown as ClerkPaginatedResponse<DirectorySyncUserJSON> | undefined;
+
+    return {
+      total_count: payload?.total_count ?? 0,
+      data: (payload?.data ?? []).map(row => new DirectorySyncUser(row)),
+    };
+  };
```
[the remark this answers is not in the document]
kalafut: The `directory` segment is correct. `scim_directory` was the originally planned path, but it was renamed before the FAPI side landed (clerk/clerk_go#21433) — the routes in `clerk_go` `api/fapi/v1/directories/http.go` are `.../enterprise_connections/{id}/directory` (plus `/rotate_api_key` and `/users`). The `scim_directory` reference came from a stale PR description, which has since been updated. No change needed.

packages/clerk-js/src/core/resources/Organization.ts
```diff
@@ -274,6 +278,32 @@ export class Organization extends BaseResource implements OrganizationResource {
...
+
+  createDirectorySync = async (
+    enterpriseConnectionId: string,
+    params?: CreateDirectorySyncParams,
+  ): Promise<DirectorySyncResource> => {
+    const json = (
+      await BaseResource._fetch<DirectorySyncJSON>({
+        path: `/organizations/${this.id}/enterprise_connections/${enterpriseConnectionId}/directory`,
+        method: 'POST',
+        body: (params?.name ? { name: params.name } : {}) as any,
```
mauricioabreu: is senting `params.name: undefined` acceptable? if so, I'd write it as `body: { name: params?.name }`

mauricioabreu (approved):

packages/clerk-js/src/core/resources/Organization.ts
kalafut: Updated

.changeset/dir-sync-self-serve-wiring.md
```diff
@@ -0,0 +1,8 @@
+---
+'@clerk/clerk-js': minor
+'@clerk/localizations': minor
+'@clerk/shared': minor
+'@clerk/ui': minor
+---
+
+Add self-serve Directory Sync (SCIM) setup. The `Organization` resource gains `getDirectorySync()` and `createDirectorySync()` for the directory bound to an enterprise connection, and the returned `DirectorySync` resource exposes `update()`, `rotateToken()`, `delete()`, and `getUsers()`; the SCIM bearer token is only returned by `createDirectorySync()` and `rotateToken()`. The `OrganizationProfile` Security page gains a Directory Sync section, and the internal `ConfigureDirectorySync` component walks through the setup. Both are only shown when the instance has self-serve Directory Sync enabled.
```
dstaley: isn't there a changeset at a higher PR in the stack? either way, this is a little more descriptive that we usually want in a changelog. Since this feature is shipping as one unit I'd prefer it simply say something like "Introduce self-serve SCIM capabilities and related functionality."

packages/shared/src/types/userSettings.ts
```diff
@@ -99,6 +99,8 @@ export type OAuthProviders = {
 export type EnterpriseSSOSettings = {
   enabled: boolean;
   self_serve_sso: boolean;
+  /** Whether end-users may manage Directory Sync for their enterprise connections. Absent from older backends, which means `false`. */
```
dstaley: what does "older backends" here mean? Does this mean specific API versions? If so, we should clearly state which versions it's available in.

packages/shared/src/types/organization.ts
```diff
@@ -221,6 +222,19 @@ export interface OrganizationResource extends ClerkResource, BillingPayerMethods
     enterpriseConnectionId: string,
     params?: GetEnterpriseConnectionTestRunsParams,
   ) => Promise<ClerkPaginatedResponse<EnterpriseConnectionTestRunResource>>;
+  /**
+   * Gets the Directory Sync directory bound to the given enterprise connection. The returned resource never carries
```
dstaley: I don't think we need to call out that the returned resource does not have a bearer token.

packages/shared/src/types/directorySync.ts
```diff
@@ -0,0 +1,129 @@
...
+export interface DirectorySyncUserResource extends ClerkResource {
+  id: string;
+  userId: string;
+  firstName: string | null;
+  lastName: string | null;
+  /** The user's primary email address. */
+  identifier: string | null;
+  imageUrl: string;
+  hasImage: boolean;
+  /** `false` once the identity provider has deprovisioned the user. */
```
dstaley:
```suggestion
  /** `false` if the identity provider has deprovisioned the user. */
```

packages/shared/src/types/directorySync.ts
```diff
@@ -0,0 +1,129 @@
...
+  image_url: string;
+  has_image: boolean;
+  active: boolean;
+  provisioned_at: number;
+  updated_at: number;
+}
+
+/**
+ * A user the identity provider has provisioned into the directory, in
+ * public-user-data shape.
```
dstaley: what does "public-user-data shape" mean here?

dstaley (changes requested):

.changeset/dir-sync-self-serve-wiring.md
kalafut: I'll reword. re: the position, this used to be at the end of the stack, but then CI was always failing at the lower levels so I moved it down. idk if that was the correct approach.

kalafut: Updated the description.

packages/shared/src/types/userSettings.ts
kalafut: Unneeded and removed that bit.

packages/shared/src/types/organization.ts
kalafut: Removed.

packages/shared/src/types/directorySync.ts
kalafut: Updated (locally)

packages/shared/src/types/directorySync.ts
kalafut: Reworded to say that the fields are the same as `PublicUserData`.

dstaley (approved):

PR #9591, opened 2026-08-27
feat(shared): add internal directory sync react-query hooks

kalafut:
## Description

Part 2 of 4 of the self-serve Directory Sync stack. Stacked on `jim/dir-sync-1-resource`.  This change depends on the first PR and the stack will be squashed on merge.

Adds the internal react-query hooks `__internal_useOrganizationDirectorySync` and `__internal_useOrganizationDirectorySyncUsers`. Mutations go through the loaded `DirectorySync` resource, so they resolve `undefined` until the directory has loaded. The users hook takes the `directory` resource itself and stays dormant while it is nullish; it polls while armed via `startPolling`, so the wizard's test step doubles as a recent-activity feed.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

<span title="This PR description was written by AI using employee-skills/write-pr-description">✎</span>

### DIFF
--- a/packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
+++ b/packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
@@ -8,7 +8,6 @@ import type {
   UpdateDirectorySyncParams,
 } from '../../types/directorySync';
 import { useClerkInstanceContext } from '../contexts';
-import { defineKeepPreviousDataFn } from '../query/keep-previous-data';
 import { useClerkQueryClient } from '../query/use-clerk-query-client';
 import { useClerkQuery } from '../query/useQuery';
 import { useOrganizationBase } from './base/useOrganizationBase';
@@ -18,7 +17,6 @@ import { useOrganizationDirectorySyncCacheKeys } from './useOrganizationDirector
 export type UseOrganizationDirectorySyncParams = {
   enterpriseConnectionId: string | null;
   enabled?: boolean;
-  keepPreviousData?: boolean;
 };
 
 export type UseOrganizationDirectorySyncReturn = {
@@ -45,7 +43,7 @@ export type UseOrganizationDirectorySyncReturn = {
  * @internal
  */
 function useOrganizationDirectorySync(params: UseOrganizationDirectorySyncParams): UseOrganizationDirectorySyncReturn {
-  const { enterpriseConnectionId, enabled = true, keepPreviousData = true } = params;
+  const { enterpriseConnectionId, enabled = true } = params;
   const clerk = useClerkInstanceContext();
   const organization = useOrganizationBase();
   const [queryClient] = useClerkQueryClient();
@@ -80,7 +78,7 @@ function useOrganizationDirectorySync(params: UseOrganizationDirectorySyncParams
       }
     },
     enabled: queryEnabled,
-    placeholderData: defineKeepPreviousDataFn(keepPreviousData),
+    // No placeholderData: any key change is an identity change, and the mutations act on `query.data`.
   });
 
   const revalidate = useCallback(
--- a/packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
+++ b/packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
@@ -32,6 +32,7 @@ export type UseOrganizationDirectorySyncReturn = {
   isLoading: boolean;
   isFetching: boolean;
   createDirectorySync: (params?: CreateDirectorySyncParams) => Promise<DirectorySyncResource | undefined>;
+  /** Resolves `undefined` until `data` has loaded, since the mutations act on the loaded directory. */
   updateDirectorySync: (params: UpdateDirectorySyncParams) => Promise<DirectorySyncResource | undefined>;
   rotateDirectorySyncToken: () => Promise<DirectorySyncResource | undefined>;
   deleteDirectorySync: () => Promise<DeletedObjectResource | undefined>;
@@ -99,35 +100,37 @@ function useOrganizationDirectorySync(params: UseOrganizationDirectorySyncParams
     [organization, enterpriseConnectionId, revalidate],
   );
 
+  const directory = query.data;
+
   const updateDirectorySync = useCallback(
     async (updateParams: UpdateDirectorySyncParams) => {
-      if (!enterpriseConnectionId) {
+      if (!directory) {
         return undefined;
       }
-      const updated = await organization?.updateDirectorySync(enterpriseConnectionId, updateParams);
+      const updated = await directory.update(updateParams);
       await revalidate();
       return updated;
     },
-    [organization, enterpriseConnectionId, revalidate],
+    [directory, revalidate],
   );
 
   const rotateDirectorySyncToken = useCallback(async () => {
-    if (!enterpriseConnectionId) {
+    if (!directory) {
       return undefined;
     }
-    const rotated = await organization?.rotateDirectorySyncToken(enterpriseConnectionId);
+    const rotated = await directory.rotateToken();
     await revalidate();
     return rotated;
-  }, [organization, enterpriseConnectionId, revalidate]);
+  }, [directory, revalidate]);
 
   const deleteDirectorySync = useCallback(async () => {
-    if (!enterpriseConnectionId) {
+    if (!directory) {
       return undefined;
     }
-    const deleted = await organization?.deleteDirectorySync(enterpriseConnectionId);
+    const deleted = await directory.delete();
     await revalidate();
     return deleted;
-  }, [organization, enterpriseConnectionId, revalidate]);
+  }, [directory, revalidate]);
 
   return {
     data: query.data,
--- a/packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
+++ b/packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
@@ -48,7 +48,7 @@ function useOrganizationDirectorySync(params: UseOrganizationDirectorySyncParams
   const organization = useOrganizationBase();
   const [queryClient] = useClerkQueryClient();
 
-  const { queryKey, stableKey, authenticated } = useOrganizationDirectorySyncCacheKeys({
+  const { queryKey, invalidationKey, stableKey, authenticated } = useOrganizationDirectorySyncCacheKeys({
     organizationId: organization?.id ?? null,
     enterpriseConnectionId,
   });
@@ -82,8 +82,8 @@ function useOrganizationDirectorySync(params: UseOrganizationDirectorySyncParams
   });
 
   const revalidate = useCallback(
-    () => queryClient.invalidateQueries({ queryKey: [stableKey] }),
-    [queryClient, stableKey],
+    () => queryClient.invalidateQueries({ queryKey: invalidationKey }),
+    [queryClient, invalidationKey],
   );
 
   const createDirectorySync = useCallback(
--- a/packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
+++ b/packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
@@ -1,0 +1,145 @@
+import { useCallback } from 'react';
+
+import { isClerkAPIResponseError } from '../../error';
+import type { DeletedObjectResource } from '../../types/deletedObject';
+import type {
+  CreateDirectorySyncParams,
+  DirectorySyncResource,
+  UpdateDirectorySyncParams,
+} from '../../types/directorySync';
+import { useClerkInstanceContext } from '../contexts';
+import { defineKeepPreviousDataFn } from '../query/keep-previous-data';
+import { useClerkQueryClient } from '../query/use-clerk-query-client';
+import { useClerkQuery } from '../query/useQuery';
+import { useOrganizationBase } from './base/useOrganizationBase';
+import { useClearQueriesOnSignOut } from './useClearQueriesOnSignOut';
+import { useOrganizationDirectorySyncCacheKeys } from './useOrganizationDirectorySync.shared';
+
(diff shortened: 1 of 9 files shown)


packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
```diff
@@ -0,0 +1,148 @@
...
+      } catch (err) {
+        // No directory yet is a first-class state of the setup flow, not an error.
+        if (isClerkAPIResponseError(err) && err.status === 404) {
+          return null;
+        }
+        throw err;
+      }
+    },
+    enabled: queryEnabled,
+    placeholderData: defineKeepPreviousDataFn(keepPreviousData),
```
kalafut: **[P1] Do not carry a directory resource across identity changes**

Every query-key change here means the organization or enterprise connection changed. With the default `keepPreviousData`, React Query returns the previous `DirectorySyncResource` while the next query loads—and indefinitely when the replacement query is disabled because the connection is null or `enabled` is false. Since the mutation callbacks close over `query.data`, they can then update, rotate, or delete the previously selected directory. Please avoid placeholder reuse for this resource, or ensure both the returned data and mutations reject a resource whose organization/connection no longer match the current inputs.


<span title="This comment was written by AI, reviewed &amp; directed by Jim">~ 🤖</span>

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
```diff
@@ -0,0 +1,153 @@
...
+    queryFn: () => {
+      if (!directory) {
+        throw new Error('directory is required to fetch directory users');
+      }
+      return directory.getUsers(fetchParams);
+    },
+    refetchInterval: () => (shouldPoll ? pollIntervalMs : false),
+    enabled: queryEnabled,
+    refetchIntervalInBackground: false,
+    placeholderData: defineKeepPreviousDataFn(keepPreviousData),
```
kalafut: **[P1] Scope retained users to the current directory**

This conflicts with the hook's "dormant while nullish" contract: once `directory` becomes null or the hook is disabled, the replacement query cannot run, so the previous page remains as successful placeholder data indefinitely. An organization/directory switch can likewise expose the old rows in the new context, and the key only tracks `enterpriseConnectionId`, so deleting and recreating a directory on the same connection reuses that cache. Please retain previous data only for pagination within the same organization and `directory.id`, and return `undefined` when the query is dormant or the directory identity changes.


<span title="This comment was written by AI, reviewed &amp; directed by Jim">~ 🤖</span>

packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
kalafut: Addressed in `f801038` by removing `keepPreviousData` and placeholder reuse from this hook, so identity changes immediately clear `query.data` and the mutation closures no longer target the prior resource.


<span title="This comment was written by AI, reviewed &amp; directed by Jim">~ 🤖</span>

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
kalafut: The original issue is addressed in `f801038` by adding `directoryId` to the cache and polling identity, retaining placeholder data only within the same organization and directory, and hiding rows while the query is dormant.

One related identity mismatch remains: the cache key is scoped to the active `organization.id`, but `queryFn` invokes `directory.getUsers()`, whose request path is bound to `directory.organizationId`. If the active organization changes while a caller still supplies the previous resource, `queryEnabled` remains true and old-organization users are fetched and surfaced in the new context. Please require `directory?.organizationId === organization?.id` before enabling the query, or key and fetch consistently from the directory if cross-organization behavior is intended.

There are also no regression tests for these hooks. Please cover dormant clearing, same-directory pagination retention, directory recreation on the same enterprise connection, and active-organization mismatch.


<span title="This comment was written by AI, reviewed &amp; directed by Jim">~ 🤖</span>

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  const queryEnabled = enabled && clerk.loaded && Boolean(organization) && Boolean(directory);
+
+  const [shouldPoll, setShouldPoll] = useState(false);
+
+  useEffect(() => {
+    // Polling intent is scoped to the current directory — clear it when the
+    // identity changes so a reset/recreate doesn't inherit a stale armed poll.
+    setShouldPoll(false);
+  }, [enterpriseConnectionId, directory?.id]);
```
NicolasLopes7: The reset effect clears `shouldPoll` whenever `directory?.id` changes. 

React runs child effects before parent effects, so a child that calls `startPolling()` in the same commit the directory arrives gets silently cancelled. 

Fix: drop the effect and store the armed directory id instead, so `shouldPoll` is derived as `armedForId === directory?.id`. That also removes the redundant `enterpriseConnectionId` dep. The sibling test-runs hook has the same latent bug.

packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
```diff
@@ -0,0 +1,146 @@
...
+        }
+        throw err;
+      }
+    },
+    enabled: queryEnabled,
+    // No placeholderData: any key change is an identity change, and the mutations act on `query.data`.
+  });
+
+  const revalidate = useCallback(
+    () => queryClient.invalidateQueries({ queryKey: [stableKey] }),
```
NicolasLopes7: When this hook refreshes, it throws out the cached directory for every organization and every connection, not just the one it's looking at. Harmless today, but not ideal. The users hook already does it the narrow way. Copy that.

NicolasLopes7 (approved):
Got fable looking/running this and had a couple of inline-comments. Other than that I'd suggest tests for the hooks to cover the regressions, otherwise LGTM

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
kalafut: Updated

packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
kalafut: Updated

packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
```diff
@@ -0,0 +1,146 @@
...
+export type UseOrganizationDirectorySyncParams = {
+  enterpriseConnectionId: string | null;
+  enabled?: boolean;
+};
+
+export type UseOrganizationDirectorySyncReturn = {
+  /**
+   * The connection's directory, `null` when none has been created yet, `undefined` while loading.
+   * Never carries the bearer token — that only exists on the resources resolved by
+   * `createDirectorySync` and `rotateDirectorySyncToken`.
```
dstaley: don't think we need this

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { useClerkQueryClient } from '../query/use-clerk-query-client';
+import { useClerkQuery } from '../query/useQuery';
+import { useOrganizationBase } from './base/useOrganizationBase';
+import { useClearQueriesOnSignOut } from './useClearQueriesOnSignOut';
+import { useOrganizationDirectorySyncUsersCacheKeys } from './useOrganizationDirectorySync.shared';
+
+const DEFAULT_POLL_INTERVAL_MS = 2_000;
+
+export type UseOrganizationDirectorySyncUsersParams = {
+  /** The directory to list users for, e.g. `data` from `useOrganizationDirectorySync`. Dormant while nullish. */
```
dstaley: What does "Dormant while nullish." mean?

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
```diff
@@ -0,0 +1,167 @@
...
+  isLoading: boolean;
+  isFetching: boolean;
+  /**
+   * `true` while the hook is actively polling
+   */
+  isPolling: boolean;
+  /**
+   * Start polling. Polling runs continuously (new provisions, updates, and
+   * deprovisions keep appearing) until `stopPolling` is called — callers
+   * should stop on unmount of the view that armed it.
```
dstaley: this sounds like we should encapsulate the start and stop polling into an effect that returns a cleanup callback function.

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
```diff
@@ -0,0 +1,167 @@
...
+  stopPolling: () => void;
+  /**
+   * Force a refetch.
+   */
+  revalidate: () => Promise<void>;
+};
+
+/**
+ * The users provisioned into an enterprise connection's Directory Sync
+ * directory, most recently touched first. Polls continuously while armed via
```
dstaley: what does "armed" mean?

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
```diff
@@ -0,0 +1,167 @@
...
+  }, []);
+
+  const revalidate = useCallback(async () => {
+    await queryClient.invalidateQueries({ queryKey: invalidationKey });
+  }, [queryClient, invalidationKey]);
+
+  const isPolling = queryEnabled && shouldPoll;
+
+  return {
+    // Dormant means dormant: never surface cached rows while the query cannot run.
```
dstaley: lol what

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
```diff
@@ -0,0 +1,167 @@
...
+};
+
+/**
+ * The users provisioned into an enterprise connection's Directory Sync
+ * directory, most recently touched first. Polls continuously while armed via
+ * `startPolling`, so the setup flow doubles as a recent-activity feed.
+ *
+ * @internal
+ */
+function useOrganizationDirectorySyncUsers(
```
dstaley, line 69: will this internal hook always be used with polling?

packages/shared/src/react/hooks/useOrganizationEnterpriseConnectionTestRuns.tsx
```diff
@@ -136,13 +136,10 @@ function useOrganizationEnterpriseConnectionTestRuns(
...
 
-  const [shouldPoll, setShouldPoll] = useState(false);
-
-  useEffect(() => {
-    // Polling intent is scoped to the current connection — clear it when the
-    // connection changes so a reset/recreate doesn't inherit a stale armed poll.
-    setShouldPoll(false);
-  }, [enterpriseConnectionId]);
+  // Polling is armed for a specific connection and derived, not reset in an effect: a child
+  // effect arming it in the same commit the connection arrives would otherwise be cancelled.
```
dstaley: I do not know what this means

dstaley (changes requested):

packages/shared/src/react/hooks/useOrganizationDirectorySync.tsx
kalafut: Removed

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
kalafut: Updated to English, "Nothing is fetched while `null` or `undefined`"

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
kalafut: Reworded to be: "A disabled query still exposes rows cached under its key. Report none until it can run."

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
kalafut: Ditched the "armed" wording and rephrased to saying that the polling is opt-in via `startPolling`

packages/shared/src/react/hooks/useOrganizationEnterpriseConnectionTestRuns.tsx
kalafut: Simplified to:
```
// Polling is requested for a specific connection, so a connection change stops it. This is
// derived rather than reset in an effect because a child component may call `revalidate` in
// the same commit the connection arrives, and an unconditional reset would cancel that.
```

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
kalafut: ~No. Polling is opt-in via `startPolling`. The list fetches once without it.~

Obsolete, per prev comment.

packages/shared/src/react/hooks/useOrganizationDirectorySyncUsers.tsx
kalafut: ~The provider stays mounted across all wizard steps, so I think the hook's own unmount can't be the signal to stop?~

I ended up redoing this and it's now simpler. Polling is now tied to the test step being visible, the start/stop pair became a poll flag, so the effect went away completely.

dstaley (approved):

PR #9600, opened 2026-08-27
feat(ui,shared,localizations): dedicated screen for a blocked request

zourzouvillys:
## Summary

When a sign-in or sign-up request is blocked, the user currently sees one generic sentence in the same small inline error slot that carries "incorrect password" — with nothing to act on and nothing to quote if they contact support.

A blocked request is **terminal**: there is no field to correct and no retry that helps. This gives it its own screen instead of an inline error, shows a short reference the user can quote, and renders the application's own wording when the API supplies it.

## What the user sees

```
    We couldn't verify this sign-in

    Sign-ins from this network are not permitted. If you
    are using a VPN, try turning it off and signing in
    again.

    [ Contact support ]

    Reference: 7Q8ikxgt
```

Everything above the reference comes from the application's configuration and is optional. The reference is shown whenever the API sends one.

## The contract

The API may include these on an `action_blocked` error's `meta`:

| Field | Meaning | Rendered? |
|---|---|---|
| `trace_id` | A short reference for the request. Opaque — do not parse, reformat, or assume a width. | Yes |
| `title` | A heading, configured by the application's owner. Plain text. | Yes |
| `description` | What happened. Plain text. | Yes |
| `link_url` | An `https` URL for help. | Yes |
| `link_text` | The link's label. | Yes |
| `kind` | A tag naming why the request was blocked, e.g. `vpn_detected`. | **No** |
| `data` | Arbitrary scalars the application's owner attached. | **No** |

All optional. They are parsed to camelCase alongside the existing meta fields.

`kind` and `data` are **carried, never displayed**. They exist so an application can render its own UI instead of this screen:

```js
if (err.meta?.kind === 'vpn_detected') {
  // draw your own screen; err.meta.data has whatever you configured
}
```

Rendering `data` would put an application's internal keys in front of an end user, which is the opposite of what it is for.

## Backwards compatibility

Additive, and this is the part worth checking in review:

- The error's `code` is **unchanged**, so an older client keeps working.
- A response carrying no meta renders exactly the inline error it did before — the new screen is feature-detected off the meta, never version-gated.
- Three states are covered: no meta at all, a reference with no message, and a full message.

One deliberate change: `message` and `long_message` now carry the application owner's own wording when they configured any (and the reference appended). That is so a client rendering only those — rather than this screen — shows what the owner wrote instead of the generic sentence. With nothing configured they are unchanged.

## Implementation

- **`@clerk/shared`** — the fields on `ClerkAPIErrorJSON` and `ClerkAPIError`, parsed in `clerkApiError.ts`.
- **`@clerk/ui`** — `ActionBlockedCard`, plus detection in the shared card state. That is where every error in these flows already funnels, so the form-submit path, the OAuth-callback path and a challenge submission that is then denied are all covered without any of them knowing about it.
- **`@clerk/ui`** — new appearance descriptors (`actionBlockedIconBox`, `actionBlockedIcon`, `actionBlockedLink`, `actionBlockedTraceIdBox`, `actionBlockedTraceIdLabel`, `actionBlockedTraceId`) and an `actionBlocked` flow part, so the screen is customizable like any other.
- **`@clerk/localizations`** — `actionBlocked.title`, `.subtitle` and `.traceIdLabel` as the fallbacks used when the application supplies no wording.

Two fixes folded in, both found by review rather than by the suite:

1. **`errorToJSON` dropped these fields.** It has its own exhaustive meta list and backs `Verification.__internal_toSnapshot`, so on the SSR/hydration path a verification error lost them and the screen degraded to generic wording with nothing saying why — on the OAuth and SAML route specifically, one of the two this screen is wired for. Both directions of that mapping are hand-maintained lists, so a round-trip test now fails if either side stops carrying them.
2. **A block arriving from a challenge submission did not get the screen.** `useProtectCheckRunner` routes that error through `handleError(..., card.setError)`, and interception used to live only in the two start components — so that card showed an inline error **with a Retry button**, for something that cannot succeed. Moving detection into card state fixed it for every card at once.

## Security

The title, description and link are **written by the application's owner and rendered in an end user's browser**, so:

- The text is set as **text nodes**. Never markdown, never HTML, never interpolated into markup.
- Only `https` links become an `href`. The URL is validated before it is sent, and checked again here before it reaches the DOM — the second check is what stands between a value that arrived anyway and a `javascript:` or `data:` URI. A link that fails is dropped and the rest of the card still renders.
- The link gets `rel="noopener noreferrer"`, since the destination is not necessarily under the owner's control once followed.

## Verification

- `pnpm --filter @clerk/ui type-check` — **no errors in any file this PR touches.** (The type errors and failing test files that remain are all under `src/mosaic/**`, which this PR does not touch — they fail on an unbuilt `@clerk/headless`.)
- `pnpm --filter @clerk/shared build` and `pnpm --filter @clerk/localizations build` — clean. The localizations build is the real check on the type contract, since every locale must satisfy `LocalizationResource`.
- `npx vitest run src/utils/__tests__/actionBlocked.test.ts` — **24 passed.** Covers reading the fields off an error, the "reference but no message" case, `kind`/`data` on their own counting as something to show, returning null when there is nothing (which is what makes the caller fall back rather than render a blank screen), and every rejected URL scheme: `javascript:` in two casings and with leading whitespace, `data:`, `vbscript:`, `file:`, `http:`, relative and protocol-relative.
- `npx vitest run src/__tests__/blockedRequestMeta.spec.ts` — **3 passed**, covering the snapshot round trip. I verified this test actually fails when the mapping is removed rather than trusting a green run.

**Note on CI:** the integration suite is currently failing repo-wide, on branches unrelated to this one — including a dependency-bump branch with no UI code. Those failures are not from this change.

A changeset is included.

### DIFF
--- a/.changeset/blocked-request-screen.md
+++ b/.changeset/blocked-request-screen.md
@@ -10,6 +10,8 @@ A blocked request is terminal — there is no field to correct and no retry that
 
 The screen shows a short reference for the request, which the end user can quote when contacting support. When the application supplies its own wording, the screen renders that instead of the default: `title`, `description`, and an optional `https` link with a label are read from the error's `meta`.
 
+The error's `meta` also carries `kind` — a tag naming why the request was blocked — and `data`, an arbitrary bag of scalars the application's owner attached. Neither is rendered by the built-in screen: they are there so an application can switch on `kind` and render its own UI instead.
+
 Also adds the `actionBlocked` localization keys (`title`, `subtitle`, `traceIdLabel`) and appearance descriptors for the new elements, so both the copy and the styling are customizable.
 
 This is additive and degrades safely: the error's `code`, `message` and `long_message` are unchanged, and a response without the new `meta` renders exactly as before.
--- a/.changeset/blocked-request-screen.md
+++ b/.changeset/blocked-request-screen.md
@@ -14,4 +14,4 @@ The error's `meta` also carries `kind` — a tag naming why the request was bloc
 
 Also adds the `actionBlocked` localization keys (`title`, `subtitle`, `traceIdLabel`) and appearance descriptors for the new elements, so both the copy and the styling are customizable.
 
-This is additive and degrades safely: the error's `code`, `message` and `long_message` are unchanged, and a response without the new `meta` renders exactly as before.
+This degrades safely: the error's `code` is unchanged, and a response without the new `meta` renders exactly as before. Note that `message` and `long_message` now carry the application owner's own wording when they configured any — so a client that renders only those, rather than the new screen, shows what the owner wrote instead of the generic sentence.
--- a/.changeset/blocked-request-screen.md
+++ b/.changeset/blocked-request-screen.md
@@ -1,0 +1,15 @@
+---
+'@clerk/localizations': minor
+'@clerk/shared': minor
+'@clerk/ui': minor
+---
+
+Show a dedicated screen when a sign-in or sign-up request is blocked, instead of a generic inline error.
+
+A blocked request is terminal — there is no field to correct and no retry that helps — so it now replaces the card rather than appearing as a small error beside a form the user cannot resubmit.
+
+The screen shows a short reference for the request, which the end user can quote when contacting support. When the application supplies its own wording, the screen renders that instead of the default: `title`, `description`, and an optional `https` link with a label are read from the error's `meta`.
+
+Also adds the `actionBlocked` localization keys (`title`, `subtitle`, `traceIdLabel`) and appearance descriptors for the new elements, so both the copy and the styling are customizable.
+
+This is additive and degrades safely: the error's `code`, `message` and `long_message` are unchanged, and a response without the new `meta` renders exactly as before.
--- a/packages/ui/src/common/ActionBlockedCard.tsx
+++ b/packages/ui/src/common/ActionBlockedCard.tsx
@@ -1,0 +1,206 @@
+import { ERROR_CODES } from '@clerk/shared/internal/clerk-js/constants';
+import type { ClerkAPIError } from '@clerk/shared/types';
+import React from 'react';
+
+import { Col, descriptors, Flex, Flow, Icon, localizationKeys, Text } from '../customizables';
+import { Card } from '../elements/Card';
+import { Header } from '../elements/Header';
+import { ExclamationTriangle } from '../icons';
+
+/**
+ * The details an application can attach to a blocked request. Every field is
+ * optional; when none are present the card falls back to its own wording and
+ * shows only the reference.
+ *
+ * The text fields are plain text and are rendered as text nodes. They are
+ * written by the application's owner, so they are treated as content, never as
+ * markup.
+ */
+export type ActionBlockedDetails = {
+  traceId?: string;
+  title?: string;
+  description?: string;
+  linkUrl?: string;
+  linkText?: string;
+};
+
+/**
+ * Reads the details off an API error, or returns null when the error carries
+ * none — which is also what happens against an older backend that does not send
+ * them. Callers use the null to fall back to the previous inline error, so a
+ * missing field degrades rather than rendering a blank screen.
+ */
+export const getActionBlockedDetails = (error: ClerkAPIError | undefined): ActionBlockedDetails | null => {
+  const meta = error?.meta as ActionBlockedDetails | undefined;
+  if (!meta) {
+    return null;
+  }
+  const { traceId, title, description, linkUrl, linkText } = meta;
+  if (!traceId && !title && !description && !linkUrl) {
+    return null;
+  }
+  return { traceId, title, description, linkUrl, linkText };
+};
+
+/**
+ * Only `https` links are rendered.
+ *
+ * The URL is already checked before it is sent, so this is a second, local
+ * check rather than the only one: it is what stands between a value that
+ * reached the browser anyway and a `javascript:` or `data:` URI becoming an
+ * `href`. A link that fails is dropped and the rest of the card still renders.
+ */
+export const safeHref = (url: string | undefined): string | null => {
+  if (!url) {
+    return null;
+  }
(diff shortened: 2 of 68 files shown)


.changeset/blocked-request-screen.md
```diff
@@ -0,0 +1,17 @@
...
+
+The screen shows a short reference for the request, which the end user can quote when contacting support. When the application supplies its own wording, the screen renders that instead of the default: `title`, `description`, and an optional `https` link with a label are read from the error's `meta`.
+
+The error's `meta` also carries `kind` — a tag naming why the request was blocked — and `data`, an arbitrary bag of scalars the application's owner attached. Neither is rendered by the built-in screen: they are there so an application can switch on `kind` and render its own UI instead.
+
+Also adds the `actionBlocked` localization keys (`title`, `subtitle`, `traceIdLabel`) and appearance descriptors for the new elements, so both the copy and the styling are customizable.
+
+This degrades safely: the error's `code` is unchanged, and a response without the new `meta` renders exactly as before. Note that `message` and `long_message` now carry the application owner's own wording when they configured any — so a client that renders only those, rather than the new screen, shows what the owner wrote instead of the generic sentence.
```
Ephem, line 17: This does not read like a user-facing changelog.

packages/ui/src/elements/contexts/index.tsx
```diff
@@ -44,8 +52,34 @@ export const useCardState = () => {
...
+   * and discards the meta the screen is built from. Anything that is not a
+   * blocked request, or that carries no details (an older backend), falls
+   * through unchanged.
+   */
+  const setError = (metadata: ClerkRuntimeError | ClerkAPIError | Metadata | string) => {
+    const blocked = actionBlockedDetailsFrom(metadata);
+    if (blocked) {
+      setState(s => ({ ...s, blockedDetails: blocked, error: undefined }));
+      return;
+    }
```
Ephem, line 75: This is a UI context, dealing with UI state, which seems like the wrong place to put Protect-specific logic?

Plus, I think we should do this mapping only for the flows that actually render the new screen (/consume `blockedDetails`), or else we could risk swallowing the error and not showing it at all.

Maybe that's not possible in practice, maybe the flows that use API endpoints that can return `blockedDetails` are already covered in this PR, but there's no way to reason about that locally here without going to read the API's, which is a smell.

In short, I'd love to see a more strict link between this mapping and actually consuming `blockedDetails`.

If it helps to preserve the `rawError` to parse from later instead of the translated one, I'd say that's fine to keep here though.

packages/ui/src/elements/contexts/index.tsx
Ephem, line 1: `CardStateProvider` does this: `error: translateError(window?.Clerk?.__internal_last_error || undefined),`

That `__internal_last_error` can come from `__internal_navigateWithError` so if we ever call that function with a blocked error, it would not get parsed to `blockedDetails`.

As per the other comment, I don't think we should be handling this here anyway, but still wanted to mention it.

packages/ui/src/utils/actionBlocked.ts
```diff
@@ -0,0 +1,89 @@
...
+  if (!meta) {
+    return null;
+  }
+  const { traceId, kind, title, description, linkUrl, linkText, data } = meta;
+  // `kind` and `data` count: a rule configured with only a kind is one whose
+  // application renders its own screen, and treating that as "nothing to show"
+  // would take the feature away from exactly that integration.
+  if (!traceId && !kind && !title && !description && !linkUrl && !data) {
+    return null;
+  }
```
Ephem, line 50: I'm trying to understand this.

* This function only runs for the AIO components?
* `kind` and `data` should not really be set if you are using the AIO components?
* But if you still set it, we still want to show a generic error even if it's the only thing on the meta?

If all this is correct, I'd suggest a comment along the lines of:

```
// `kind` and `data` should only be set for custom UI flows, but if they are set
// for an AIO flow they should still count as blocked errors
```

Because I have no idea what this means?

```
// a rule configured with only a kind is one whose
// application renders its own screen, and treating that as "nothing to show"
// would take the feature away from exactly that integration.
```

packages/ui/src/elements/contexts/index.tsx
zourzouvillys: reworked it the way you suggested so card state doesn't interpret anything now.

packages/ui/src/utils/actionBlocked.ts
zourzouvillys: yes to all three, although `kind` and `data` no longer count toward showing the screen at all, since the screen never renders them.

zourzouvillys:
@Ephem thanks for the time you put into reviewing this, really appreciate it.

packages/ui/src/common/ActionBlockedCard.tsx
```diff
@@ -11,15 +11,6 @@ type ActionBlockedCardProps = {
...
 
-/**
- * The screen shown when a request was blocked and there is nothing the end user
- * can do to retry it.
- *
- * A block is terminal — there is no field to correct and no second attempt that
- * helps — so it replaces the form rather than appearing as an inline error
- * beside it. The one thing the user can act on is the reference, which is why it
- * is always rendered and is selectable.
- */
```
Ephem: I noticed the latest changes removes a ton of comments, maybe because of the new repo rule we added that AI shouldn't add comments if we don't ask for it?

I do think a lot of the original comments in this PR were hard to read, but there were quite a few I liked too, like this one.

This is not a call to action to go bring a bunch of comments back, just a meta-commentary on our discussion last week that it's hard to find the right balance, not sure the _never_ make comments agent rule benefits us.

packages/ui/src/elements/contexts/index.tsx
Ephem: Thank you, changes are looking good!

Ephem (approved):
Looking great!

PR #9610, opened 2026-08-29
feat(ui,clerk-js,shared): render the setup-passkey session task

dmoerner:
> **Drafted by Claude Code**, from a design discussion with @danielmoerner. Reviewers should treat the reasoning below as a proposal to check, not a summary of settled decisions.

Clerk's backend can now offer passkey enrolment right after sign-up by returning a `pending` session carrying a new `setup-passkey` task. No SDK knows that key, and an unknown task key is not handled gracefully: `getTaskEndpoint` builds `/tasks/undefined`, the `SessionTasks` switch matches no route, and `SessionTasksStart` re-navigates to the same undefined path. The user is left on a pending session they cannot clear — unable to use the app at all.

This adds the client half: a `TaskSetupPasskey` card that registers a passkey through the existing `/v1/me/passkeys` endpoints, and `Session.skipTask` to decline the offer.

Backend: clerk/clerk_go#21706. **That must deploy first** — this moves `SUPPORTED_FAPI_VERSION` to `2026-08-20` and the FAPI middleware rejects unknown API versions.

## Points worth review

1. **`skipTask` is public API.** For `setup-mfa`, a headless flow clears the task implicitly by calling the ordinary TOTP endpoints. An optional task has no side-effect equivalent — skipping is the only way to clear it — so custom sign-up flows need this or their users are stuck in `pending` permanently. It was internal in the first draft; that was wrong.
2. **Only an explicit `required` removes the decline button.** The mode comes from `passkey_settings.prompt_at_sign_up` on the environment. Absent, empty, `off` and `optional` all stay declinable, so stale settings degrade to a 400 on a skip attempt rather than trapping someone in an unclearable task.
3. **Unsupported devices differ by mode.** In `optional`, the card silently skips and never renders — no point asking for a passkey the device cannot create. In `required` there is no skip, so it shows an explanatory dead end with sign-out reachable, rather than a spinner. This is structural: `required` returns before any skip is attempted.
4. **`required` gets its own subtitle.** "Next time, sign in with your fingerprint" reads as an invitation, which is misleading on a step the user cannot decline. Both framings are pinned from both directions in tests so they cannot silently swap.

## Note for whoever ships passkey-as-2FA

`satisfies_second_factor` does not exist anywhere in this repo — that feature's client types were never added, and it rides the same `2026-08-20` bump. More JS work outstanding there than the backend state suggests.

## Not changed

`SUPPORTED_BAPI_VERSION` in `packages/backend` is still `2026-05-12`, and `handshake.ts` sends it as `__clerk_api_version` on the handshake redirect. Out of scope here, but it means handshake requests still declare the old FAPI version — worth a decision.

## Testing

9 tests in `TaskSetupPasskey.test.tsx` (auto-skip without rendering, both refusal modes, cancelled WebAuthn dialog leaving both actions live, both copy framings) and 3 in `Session.test.ts` (endpoint URL, no body, `currentTask` advances, 400 propagates). Full suites pass: shared 1228, clerk-js 1060, ui 3200. `tsc --noEmit` clean; no new lint warnings.

Part of [CORE-3729](https://linear.app/clerk/issue/CORE-3729/passkey-sign-up)

https://claude.ai/code/session_01VdvBsr6erhXVy4edPejNGx


dmoerner:
Tabling this for now due to time.

PR #9619, opened 2026-08-31
fix(clerk-js,ui): show the challenge raised while handing off to an enterprise connection

zourzouvillys:
## Description

A sign-in that uses an enterprise connection can be asked for a verification challenge at the point it hands off to the identity provider. When that happens the server returns before it has prepared the hand-off, so there is no external URL to follow.

`SignIn.authenticateWithRedirect` treated that response as invalid and threw, and the sign-in dead-ended on an error the user cannot act on.

It now returns instead, and the caller routes to the challenge. Once resolved, the hand-off is retried with `continueSignIn` and prepared for real. If the challenge arrives on `create`, the prepare is skipped rather than issued and discarded.

**Effects and risks**

- The early return applies to every strategy that goes through this method, not only enterprise connections. For OAuth the response today carries both a pending challenge and a redirect URL, so this stops at the challenge where it previously redirected and picked it up on the callback. Both routes end at the same place; this is the shorter one.
- `authenticateWithEnterpriseSSO` now awaits rather than returning its promise, so the loading state it holds is released by the navigation instead of by the redirect.
- No change when nothing is pending: the redirect is followed exactly as before.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/enterprise-sso-hand-off-challenge.md
+++ b/.changeset/enterprise-sso-hand-off-challenge.md
@@ -1,0 +1,6 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/ui': patch
+---
+
+Fix enterprise SSO sign-ins erroring instead of showing a verification challenge raised while handing off to the identity provider.
--- a/packages/clerk-js/src/core/resources/SignIn.ts
+++ b/packages/clerk-js/src/core/resources/SignIn.ts
@@ -388,6 +388,12 @@ export class SignIn extends BaseResource implements SignInResource {
 
     const redirectUrl = SignIn.clerk.buildUrlWithAuth(params.redirectUrl);
 
+    // A pending `protect_check` leaves the hand-off unprepared: the server returns before it
+    // builds a verification, so there is no external URL to navigate to. Stop rather than
+    // reporting the response as invalid — the caller runs the challenge and calls back in with
+    // `continueSignIn`, at which point the hand-off is prepared for real.
+    const isChallengePending = () => !!this.protectCheck || this.status === 'needs_protect_check';
+
     if (!this.id || !continueSignIn) {
       await this.create({
         strategy,
@@ -395,6 +401,10 @@ export class SignIn extends BaseResource implements SignInResource {
         redirectUrl,
         actionCompleteRedirectUrl,
       });
+
+      if (isChallengePending()) {
+        return;
+      }
     }
 
     if (strategy === 'enterprise_sso') {
@@ -405,6 +415,10 @@ export class SignIn extends BaseResource implements SignInResource {
         oidcPrompt,
         enterpriseConnectionId,
       });
+
+      if (isChallengePending()) {
+        return;
+      }
     }
 
     const { status, externalVerificationRedirectURL } = this.firstFactorVerification;
--- a/packages/clerk-js/src/core/resources/SignIn.ts
+++ b/packages/clerk-js/src/core/resources/SignIn.ts
@@ -388,10 +388,8 @@ export class SignIn extends BaseResource implements SignInResource {
 
     const redirectUrl = SignIn.clerk.buildUrlWithAuth(params.redirectUrl);
 
-    // A pending `protect_check` leaves the hand-off unprepared: the server returns before it
-    // builds a verification, so there is no external URL to navigate to. Stop rather than
-    // reporting the response as invalid — the caller runs the challenge and calls back in with
-    // `continueSignIn`, at which point the hand-off is prepared for real.
+    // Defer external navigation while a challenge is pending: the caller resolves it and calls
+    // back in with `continueSignIn`.
     const isChallengePending = () => !!this.protectCheck || this.status === 'needs_protect_check';
 
     if (!this.id || !continueSignIn) {
--- a/packages/ui/src/components/SignIn/SignInStart.tsx
+++ b/packages/ui/src/components/SignIn/SignInStart.tsx
@@ -454,13 +454,17 @@ function SignInStartInternal(): JSX.Element {
     const redirectUrl = ctx.ssoCallbackUrl;
     const redirectUrlComplete = ctx.afterSignInUrl || '/';
 
-    return signIn.authenticateWithRedirect({
+    await signIn.authenticateWithRedirect({
       strategy: 'enterprise_sso',
       redirectUrl,
       redirectUrlComplete,
       oidcPrompt: ctx.oidcPrompt,
       continueSignIn: true,
     });
+
+    // Preparing the hand-off can itself raise a challenge, in which case no redirect was issued
+    // and the sign-in is sitting on the gate instead.
+    navigateOnSignInProtectGate(signIn, navigate, 'protect-check');
   };
 
   const attemptToRecoverFromSignInError = async (e: any) => {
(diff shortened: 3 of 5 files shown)


Ephem (commented):
~~I think this makes sense to me~~ (Update: maybe not?), but I think we should group it with PR2 in this stack? This merging on its own creates a new gap, the one commented on:

```
// Defer external navigation while a challenge is pending: the caller resolves it and calls
// back in with `continueSignIn`.
```

There might technically be an issue with `clerk-js`<->`ui` version compatibility here too. If someone uses this new `clerk-js` version without the `ui` version from the next PR, `authenticateWithRedirectOrPopup` returns without navigation and everything stalls. Given this is a closed beta and we already have a bug, I think that's probably fine though.

packages/ui/src/components/SignIn/SignInStart.tsx
```diff
@@ -454,13 +454,17 @@ function SignInStartInternal(): JSX.Element {
...
       strategy: 'enterprise_sso',
       redirectUrl,
       redirectUrlComplete,
       oidcPrompt: ctx.oidcPrompt,
       continueSignIn: true,
     });
+
+    // Preparing the hand-off can itself raise a challenge, in which case no redirect was issued
+    // and the sign-in is sitting on the gate instead.
+    navigateOnSignInProtectGate(signIn, navigate, 'protect-check');
```
Ephem, line 467: Does SignInFactorOneEnterpriseConnections also need this treatment?

https://github.com/clerk/javascript/blob/31e46ed8c8af4d50d59b62de9fa9f56812a2f129/packages/ui/src/components/SignIn/SignInFactorOneEnterpriseConnections.tsx#L31-L42

packages/clerk-js/src/core/resources/SignIn.ts
```diff
@@ -388,13 +388,21 @@ export class SignIn extends BaseResource implements SignInResource {
...
       await this.create({
         strategy,
         identifier,
         redirectUrl,
         actionCompleteRedirectUrl,
       });
+
+      if (isChallengePending()) {
+        return;
+      }
```
Ephem, line 405: Since this is part of a public API, this _is_ technically a breaking change right? Before, it always failed or navigated, now it can get stuck.

We might be fine with that since it only affects a private beta, but wanted to be clear about it.

That this is a hidden behavior you need to know about and compensate for at the callsite also kind of irks me. Doesn't feel like a great long term API?

packages/clerk-js/src/core/resources/SignIn.ts
Ephem, line 381: This gets called by `authenticateWithPopup`, which passes it into `authenticateWithTransport`. When this returns nothing, I think that is going to throw a `oauth_transport_missing_verification_url` error?

https://github.com/clerk/javascript/blob/131edec6fe84830ea76f2f0a1a21cf5a0618ff6c/packages/clerk-js/src/utils/authenticateWithTransport.ts#L73-L84

Ephem, line 431: `SocialSignInButtons` calls this without handling the new case:

https://github.com/clerk/javascript/blob/131edec6fe84830ea76f2f0a1a21cf5a0618ff6c/packages/ui/src/components/SignIn/SignInSocialButtons.tsx

Ephem, line 431: This also doesn't handle the new case:

https://github.com/clerk/javascript/blob/e371555936444ee991e1079b71e5c4a4baeda2e3/packages/shared/src/internal/clerk-js/completeSignUpFlow.ts#L41-L47

zourzouvillys:
Thanks for the feedback, @Ephem. yup, agreed - i'll merge this and #9620 together rather than letting this one land on its own. 

🤖 - on the clerk-js <-> ui skew: that's mostly gone now. `authenticateWithRedirect` throws a `ClerkRuntimeError` with `code protect_check_required` when a challenge stops it (b89fd34), instead of returning quietly. so a new `clerk-js` with an older ui gets an error, same as before this PR, rather than stalling. still not great, but it fails loudly instead of silently.

packages/ui/src/components/SignIn/SignInStart.tsx
zourzouvillys: yes, it does and fixed in 7fcfb6d on #9620, and 97d6941 moves it onto the new coded error.  main also gained the SSO bypass card since, which starts the same hand-off, so that one is handled too (9dea496).

packages/clerk-js/src/core/resources/SignIn.ts
zourzouvillys: yes on both counts. changed it so that it now throws a `ClerkRuntimeError` with `code protect_check_required` instead. it still needs to have handling at the call site to route to the challenge, so agree it isn't a long term solution either but at least it fails visibly now rather than hanging.

packages/clerk-js/src/core/resources/SignIn.ts
zourzouvillys: yes. it now gets `protect_check_required`. and for a OAuth challenged create it just follows its redirect so the transport behaves the same as main.

zourzouvillys: yup - a challenge on OAuth left the button sitting there but now a challenged oauth create follows the url then handles challenge on return so behave exactluy as they do today.

zourzouvillys: it still behaves exactly as on main so it isn't affected by the change here. i left SignUp alone on purpose as it can hand off with a challenge still outstanding. Signup ordering is #9622.

if handoff fails after sign-up challenge is solved, the card would previously spin forever. #9620 fixes that for signup too.

.changeset/enterprise-sso-hand-off-challenge.md
```diff
@@ -0,0 +1,10 @@
+---
+'@clerk/clerk-js': patch
+'@clerk/localizations': patch
+'@clerk/shared': patch
+'@clerk/ui': patch
+---
+
+Fix enterprise SSO sign-ins erroring instead of showing a verification challenge raised while handing off to the identity provider.
+
+When a verification challenge has to be completed before `signIn.authenticateWithRedirect()` or `signIn.authenticateWithPopup()` can redirect, they now throw a `ClerkRuntimeError` with code `protect_check_required` instead of a generic "not supported" error. The sign-in is gated when this happens (`signIn.protectCheck` is set, or its status is `needs_protect_check`). For enterprise SSO, run the challenge and call `authenticateWithRedirect()` again with `continueSignIn: true`. If the server has already prepared the redirect, the sign-in continues to the identity provider and the challenge runs when it returns.
```
Ephem: This is a great description and all the details deserve to be in the public changelog. The one thing that could be slightly clearer is that you only have to care about these details if you are building a custom flow, have protect enabled and you are calling these functions directly, but it's a nit, fine as is.

The PR description could use an update to reflect the new approach btw.

packages/clerk-js/src/core/resources/SignIn.ts
Ephem: Nice! I really like where this ended up.

I don't mind too much that we have to handle this at the call sites now that it's a properly documented behavior. The only time this leaks out to customers is if they both have protect enabled and are doing a custom flow, and they are already taking on a ton of complexity in that case.

So yeah, would be nice not to have to handle this at the call sites, but not a huge deal.

Ephem (approved):
Looking great, I like the changes!

PR #9620, opened 2026-08-31
fix(ui): continue to the identity provider after a challenge

zourzouvillys:
## Description

Stacked on #9619 — review the delta only; that one merges first.

A sign-in whose only supported first factor is an enterprise connection has no first factor to render. The start page knows this and hands off to the identity provider instead of routing to the factor-one card.

Resuming a sign-in after a verification challenge did not know it, and routed to factor-one regardless. With nothing there to render, the card falls through to "Use another method", which lists social providers only — so the user is stranded with no way to reach their provider and no way back. It affects every entry point that resumes through this path, including an organization invitation ticket.

The rule for when a sign-in hands off to a connection now lives in one place, and the resume follows it as the start page already did.

**Effects and risks**

- Instances with more than one connection still route to factor-one, which presents the choice — the resume path there is unchanged. Their picker did have a separate gap, found in review and fixed here: `SignInFactorOneEnterpriseConnections` never routed a challenge at all. It called `authenticateWithRedirect` and returned it, and a gated `prepareFirstFactor` resolves that call without issuing a redirect — so the user clicked their connection and the card sat there, no challenge and no error. Every other first- and second-factor call site funnels through `navigateOnSignInProtectGate`; this one now does too.
- Not covered: resuming the *specific* connection a user picked before a challenge. Nothing carries that choice across the gate — `VerificationResource` has `strategy` but no `enterpriseConnectionId` — so the multi-connection case still re-presents the picker and the user picks again. Left as a follow-up rather than growing this fix into a state-carrying change.
- The predicates move out of `shared.ts` into a module neither file imports, so the resume can use them without a cycle. `hasMultipleEnterpriseConnections` is re-exported from its old path, so existing importers are untouched.
- The start page's two copies of the condition now call the shared helper. Same behaviour, one definition.
- If the retried hand-off raises a further challenge, the flow stays on the challenge route rather than redirecting.

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/enterprise-sso-hand-off-challenge.md
+++ b/.changeset/enterprise-sso-hand-off-challenge.md
@@ -3,4 +3,4 @@
 '@clerk/ui': patch
 ---
 
-Fix enterprise SSO sign-ins erroring instead of showing a verification challenge raised while handing off to the identity provider.
+Fix enterprise SSO sign-ins erroring, or appearing to do nothing, instead of showing a verification challenge raised while handing off to the identity provider. This covers the card for choosing between multiple enterprise connections, where clicking a connection left the user on an unchanged card.
--- a/.changeset/resume-enterprise-sso-after-challenge.md
+++ b/.changeset/resume-enterprise-sso-after-challenge.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Fix sign-ins that use an enterprise connection stranding on "Use another method" after a verification challenge, instead of continuing to the identity provider.
--- a/packages/ui/src/components/SignIn/SignInFactorOneEnterpriseConnections.tsx
+++ b/packages/ui/src/components/SignIn/SignInFactorOneEnterpriseConnections.tsx
@@ -8,6 +8,8 @@ import { Flow, localizationKeys } from '@/ui/customizables';
 import { withCardStateProvider } from '@/ui/elements/contexts';
 import type { AvailableComponentProps } from '@/ui/types';
 
+import { useRouter } from '../../router';
+import { navigateOnSignInProtectGate } from './handleProtectCheck';
 import { hasMultipleEnterpriseConnections } from './shared';
 
 /**
@@ -16,6 +18,7 @@ import { hasMultipleEnterpriseConnections } from './shared';
 const SignInFactorOneEnterpriseConnectionsInternal = () => {
   const ctx = useSignInContext();
   const clerk = useClerk();
+  const { navigate } = useRouter();
   const signIn = clerk.client.signIn;
 
   if (!hasMultipleEnterpriseConnections(signIn.supportedFirstFactors)) {
@@ -28,11 +31,11 @@ const SignInFactorOneEnterpriseConnectionsInternal = () => {
     name: ff.enterpriseConnectionName,
   }));
 
-  const handleEnterpriseSSO = (enterpriseConnectionId: string) => {
+  const handleEnterpriseSSO = async (enterpriseConnectionId: string) => {
     const redirectUrl = ctx.ssoCallbackUrl;
     const redirectUrlComplete = ctx.afterSignInUrl || '/';
 
-    return signIn.authenticateWithRedirect({
+    await signIn.authenticateWithRedirect({
       strategy: 'enterprise_sso',
       redirectUrl,
       redirectUrlComplete,
@@ -40,6 +43,11 @@ const SignInFactorOneEnterpriseConnectionsInternal = () => {
       continueSignIn: true,
       enterpriseConnectionId,
     });
+
+    // Preparing the hand-off can itself raise a challenge, in which case no redirect was issued
+    // and the sign-in is sitting on the gate instead. Without this the picker looks inert: the
+    // user clicks their connection and nothing happens.
+    navigateOnSignInProtectGate(signIn, navigate, '../protect-check');
   };
 
   return (
--- a/packages/ui/src/components/SignIn/enterpriseSSOFactors.ts
+++ b/packages/ui/src/components/SignIn/enterpriseSSOFactors.ts
@@ -1,0 +1,54 @@
+import type { EnterpriseSSOFactor, SignInFirstFactor, SignInResource } from '@clerk/shared/types';
+
+/**
+ * Whether every supported first factor hands off to an enterprise connection, i.e. there is no
+ * factor the sign-in card could render instead.
+ */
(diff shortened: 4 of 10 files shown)


packages/ui/src/components/SignIn/enterpriseSSOFactors.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Every place that continues a sign-in has to ask this — an SSO-only sign-in has no first factor
+ * to render, so routing it to the factor-one card leaves the user on alternative methods with no
+ * way to reach their identity provider. More than one connection is the exception: that is a
+ * choice, and the factor-one card presents it.
+ */
+function shouldHandOffToEnterpriseConnection(signIn: SignInResource): boolean {
+  return (
+    hasOnlyEnterpriseSSOFirstFactors(signIn) && !hasMultipleEnterpriseConnections(signIn.supportedFirstFactors ?? null)
+  );
+}
```
Ephem, line 52: What's the intended behavior here? If the user has multiple factors, picks one enterprise SSO one, gets challenged and end up here to determine whether to continue, shouldn't we continue _that specific factor which they had already chosen_?

zourzouvillys: Good catch, and the honest answer is "we don't, and for the multi-connection case we probably should" — with a caveat about which scenario actually reaches this line.

Taking the literal case first: with a mix of factors (say password + one enterprise connection), `hasOnlyEnterpriseSSOFirstFactors` is false, so this predicate returns false and we fall through to the pre-existing `navigate('../factor-one')`. That path is unchanged by this PR.

But the underlying point stands. This predicate is lifted verbatim out of `SignInStart`, where it's asked immediately after the identifier is submitted — a moment when the user has expressed no preference at all, so "is SSO the only way in?" is the only question there is to ask. Extracting it and reusing it here quietly carries that assumption into a place where the user *may* already have chosen. It's a configuration question standing in for an intent question, and the two only coincide when there's exactly one option.

The scenario that does reach it is more than one connection: `hasMultipleEnterpriseConnections` makes the predicate false, we route to `../factor-one`, and `SignInFactorOne` re-renders the connections picker — so the user picks the same connection a second time. Recoverable, but it is exactly what you're describing.

Chasing that turned up something worse in the same path, which I've fixed here in 27b73b1: `SignInFactorOneEnterpriseConnections` never routed the challenge at all. It called `authenticateWithRedirect` and returned it, and a gated `prepareFirstFactor` resolves that call *without* issuing a redirect — so the user clicked their connection and the card just sat there. No challenge, no error, nothing. Every other first- and second-factor call site funnels through `navigateOnSignInProtectGate`; this one didn't. It now does, with a test (and I checked it fails when the call is removed).

Resuming *that specific connection* I've deliberately left out. Nothing carries the choice across the challenge: `VerificationResource` has `strategy` but no `enterpriseConnectionId` (that lives on the factor), so it can't be recovered from the resource — it would need the id threaded through the protect-check navigation, or a new field on the verification. Happy to do it as a follow-up if you think the extra click is worth it; I didn't want to grow a bug fix into a state-carrying change.

.changeset/enterprise-sso-hand-off-challenge.md
```diff
@@ -3,4 +3,4 @@
 '@clerk/ui': patch
 ---
 
-Fix enterprise SSO sign-ins erroring instead of showing a verification challenge raised while handing off to the identity provider.
+Fix enterprise SSO sign-ins erroring, or appearing to do nothing, instead of showing a verification challenge raised while handing off to the identity provider. This covers the card for choosing between multiple enterprise connections, where clicking a connection left the user on an unchanged card.
```
Ephem: "This covers the card for choosing between multiple enterprise connections, where clicking a connection left the user on an unchanged card."

I'm not sure what this means? Unless it's carrying some extra information above "or appearing to do nothing" that's helpful to users when reading the changelog I think we can just scratch it?

Also noting this is editing the changeset from the first PR, which is probably fine if they intend to merge together and otherwise there will probably be a git conflict anyway.

.changeset/resume-enterprise-sso-after-challenge.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Fix sign-ins that use an enterprise connection stranding on "Use another method" after a verification challenge, instead of continuing to the identity provider.
```
Ephem, line 5: Perfect!

packages/ui/src/components/SignIn/SignInFactorOneEnterpriseConnections.tsx
```diff
@@ -28,18 +31,23 @@ const SignInFactorOneEnterpriseConnectionsInternal = () => {
...
       redirectUrlComplete,
       oidcPrompt: ctx.oidcPrompt,
       continueSignIn: true,
       enterpriseConnectionId,
     });
+
+    // Preparing the hand-off can itself raise a challenge, in which case no redirect was issued
+    // and the sign-in is sitting on the gate instead. Without this the picker looks inert: the
+    // user clicks their connection and nothing happens.
+    navigateOnSignInProtectGate(signIn, navigate, '../protect-check');
```
Ephem: This is a nice find, and a good fix for now. I can't help feeling something is a bit smelly to start with here though. This requires implicit knowledge that `authenticateWithRedirect` can silently fail, and the resulting `signIn` state might require a protect check.

Not sure what a better API would look like, and it's not something to tackle in this PR, but wanted to highlight that it feels slightly rough.

Ephem (approved):
Sorry it took some time to get back here, I knew I hadn't answered but I thought I had originally approved with comment so didn't realize you were fully blocked.

Nice extra finding and fix!

I agree with not growing scope to fix resuming the exact connection. I would treat that as a bug to follow up on, having to pick again doesn't feel like the gold standard we chase, but I'd consider it a minor and not urgent one.

.changeset/enterprise-sso-hand-off-challenge.md
zourzouvillys: agree - scratched!

zourzouvillys:
thanks @Ephem. few changes landed after your approval since they're in the same path:

- challenge card spin forever if the hand-off failed afyer the challenge was solved. it now shows the error and "try again" retries the handoff in the shared `useProtectCheckRunner`.
- start page also hands off to SSO from needs_identifier.
- sso bypass from main

Ephem (approved):
Changes are looking good, nice work here. 😄

PR #9626, opened 2026-08-31
feat(ui): add a Back action to the authenticator app verification step

paulogdm:
## Description

After continuing past the QR code when setting up an authenticator app for two-step verification in `<UserProfile />`, the only action was **Cancel**, which abandons MFA setup entirely. A user who wanted another look at the QR code — because they missed the scan, or picked the wrong entry in their authenticator — had to start the whole flow over.

The verification step now has a **Back** action that returns to the QR code. It reuses the existing top-level `backButton` localization key, so no new string is needed.

Back reuses the TOTP secret already issued rather than requesting a new one. This matters because `Wizard` renders a single step at a time, so navigating back remounts the QR screen and re-runs its `createTOTP()` effect; since `createTOTP()` unconditionally `POST`s `/me/totp` and mints a fresh secret, a naive Back would have silently invalidated the code the user had already scanned and left verification failing for no visible reason.

### Before

Only "Cancel" — no way back to the QR code.

CleanShot 2026-09-01 at 16 20 39

### After

Note the "back" button.

CleanShot 2026-09-01 at 16 35 49

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


justinwilloughby:
We should make sure the localization is filled out before merging. Showing as undefined for everything but EN.

packages/ui/src/components/UserProfile/AddAuthenticatorApp.tsx
```diff
@@ -67,10 +81,23 @@ export const AddAuthenticatorApp = withCardStateProvider((props: AddAuthenticato
...
-              <QRCode
-                justify='center'
-                url={totp.uri || ''}
-              />
+              <>
+                <QRCode
+                  justify='center'
+                  url={totp.uri || ''}
+                />
+                {isMobileDevice() && totp.uri && (
```
jescalan: This mobile gate excludes full-size iPads. [WebKit documents](https://webkit.org/blog/9674/new-webkit-features-in-safari-13/) that Safari on iPad normally presents a macOS user agent; at a viewport wider than 760px, that misses both the media query and the mobile-UA regex, while the touch fallback still requires `Mobi`. I reproduced `isMobileDevice()` returning `false` for that configuration, so this action—and the SessionTasks copy—never renders even though an iPad has the same same-device QR limitation. Can we use iPadOS-aware/capability-oriented gating (or show the action more broadly) and add regression coverage for desktop-class iPad Safari?


<span title="This comment was written by AI, reviewed &amp; directed by Jeff">~ 🤖</span>

paulogdm: Checking. However, the upstream fix should be shipped in `isMobileDevice()` given the argument.
This would be the wrong PR for the comment.

paulogdm: Confirmed. Unfortunately, I would say this is out of the scope for this specific PR.

Ephem (changes requested):
I agree this is an area that could use some improvements, but I have some concerns with this approach that I posted elsewhere for discussion.

packages/ui/src/components/UserProfile/AddAuthenticatorApp.tsx
```diff
@@ -38,8 +39,18 @@ export const AddAuthenticatorApp = withCardStateProvider((props: AddAuthenticato
...
+    // issued this session — otherwise navigating back here invalidates the QR the
+    // user has scanned.
+    if (resourceRef.current) {
+      return;
+    }
+
     void createTOTP()
-      .then(totp => setTOTP(totp))
+      .then(totp => {
+        resourceRef.current = totp;
```
Ephem: NIT: This ref used to save the result of the verified enrollment, but now it can _also_ mean the pending TOTP code. That technically _works_ but I find it a bit confusing.

Better, split into two:

* `pendingTotpRef`
* `verifiedTotpRef`

I'm not sure I see the point of these beings refs though, I would probably have created a single state in the parent that represents the full totp process and used that instead, but that's not necessarily in scope for this PR.

paulogdm:
PS: Failing for Astro due to upstream, unrelated issues.

Ephem (approved):
I left a nitpick about reusing the same ref, but not a blocker. This is looking good to me and seems to work well, thanks for tackling it!

packages/ui/src/components/UserProfile/AddAuthenticatorApp.tsx
paulogdm: Ah, great nitpick! Will adjust, as it makes sense.

Ephem (approved):

PR #9636, opened 2026-09-02
fix(backend,fastify): Skip handshake resolution for non-navigation requests when opted out

wobsoriano:
## Description

After merging #8560, I realized the better fix belong to `@clerk/backend`, and most of what fastify added can go. The nonce exchange in `authenticateRequest` runs before the handshake eligibility check, so a stale `__clerk_handshake_nonce` cookie on a plain API request always cost a failing Backend API call. Stripping the cookie in fastify only  hid that from backend. 

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/backend/src/tokens/request.ts
```diff
@@ -461,8 +461,14 @@ export const authenticateRequest: AuthenticateRequest = (async (
...
      * If we have a handshakeToken, resolve the handshake and attempt to return a definitive signed in or signed out state.
+     * With `__internal_resolveHandshakeOnlyForNavigation`, requests that are not eligible for a handshake redirect
+     * skip this, so a stale handshake cookie does not trigger a failing payload exchange.
      */
-    if (authenticateContext.handshakeNonce || authenticateContext.handshakeToken) {
+    const hasHandshakeToken = authenticateContext.handshakeNonce || authenticateContext.handshakeToken;
+    const canResolveHandshake =
+      !authenticateContext.__internal_resolveHandshakeOnlyForNavigation ||
+      handshakeService.isRequestEligibleForHandshake();
+    if (hasHandshakeToken && canResolveHandshake) {
```
wobsoriano, line 471: This is backwards compatible. Only fastify passes a value here, mapped from its existing `__internal_enableHandshake: false`, so the customer  already on that flag keeps working after upgrading.

PR #9650, opened 2026-09-03
feat(ui): wire up Mosaic Reverification

Ephem:
## Description

This PR wires up the Mosaic Reverification logic.

* Moves the Reverification block into a new `/features` folder
  * Maybe `/blocks` are already meant to also include logic? If so I'll move it back, I moved it because everything in there seems to be pure UI right now, but might just be because we haven't wired things up yet.
  * Didn't want to place it in the root, that doesn't scale
* A new `useReverificationWithState` hook - Simple wrapper around `useReverification` as per the custom UI part of the [useReverification docs](https://clerk.com/docs/nextjs/reference/hooks/use-reverification)
  * Also handles resetting verification state when session changes
* Model
  * Business logic, handles a bunch of logic around first/second factor, different strategies etc
  * Exposes the following actions:
    * start
    * prepare
    * attempt
    * finish 
    * cancel
  * The result of these actions are parsed into a format that is accessible to the controller
    * For example it picks a starting method, handles filtering out `passkey` if webAuthn is not accessible etc
* Controller - Consists of a few parts:
  * Semi-complex state machine
  * Derives data from model+machine for the view layer
  * Wraps functions for the view layer
  * There are a few different things that mainly drives the UI
    * `controller.status` - The overall status - `idle`, `unavailable`, `loading`, `ready` - Represents "full card states"
    * When that's `ready` - `controller.step` represents which part of the flow we are in, `method-picker`, `password` etc
    * `controller.isPending` - Progress state for the current card, if status if a full page loading state, this is a inline spinner

The way this works is:

* First call the hook:
  * `const [deleteAction, reverificationState] = useReverificationWithState(deleteMutation);`
* When `deleteAction` is called, and the API responds that it needs reverification, `reverificationState.isActive` goes `true`
* Render `<Reverification {...reverificationState}>` - This will drive the process and when resolved: 
  * `reverificationState.isActive` is going to go `false` again
  * The `deleteMutation` gets retried

So if you pass that `deleteAction` into another machine, when it gets called and needs reverification, it's going to stay pending until resolved. That means we can render `<Reverification>` inside another flow, without that machine having to know about it.

Still missing, saving for follow ups:

* Loading and Unavailable UI

I have stacked a temporary PR on top of this one that you can use to test the flow inside swingset: https://github.com/clerk/javascript/pull/9671

Here's the live preview from that PR: [https://swingset-git-fredrik-swingset-live.clerkstage.dev/live/reverification](https://swingset-git-fredrik-swingset-live.clerkstage.dev/live/reverification)

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/features/reverification/reverification.controller.ts
```diff
@@ -0,0 +1,405 @@
...
+        onError: {
+          target: 'verifying',
+          actions: assign((_, event) => ({ errorMessage: errorMessage(event.error) })),
+        },
+      }),
+    },
+
+    methodPicker: {
+      on: {
+        SELECT_METHOD: [
```
alexcarpenter, line 295: the choose method panel supports pendingMethodId for showing a spinner but the view never passes it. should we stay on the picker with the row spinning until prepare resolves?

heres the difference I am seeing in swingset

BEFORE

https://github.com/user-attachments/assets/728af1f2-86e3-49c1-92b7-2186d0f84c40

AFTER

https://github.com/user-attachments/assets/c8ac5062-70e1-4118-9634-08c3ff30738e

packages/ui/src/mosaic/features/reverification/reverification.model.tsx
```diff
@@ -0,0 +1,223 @@
...
+  const webAuthnSupported = isWebAuthnSupported();
+  const preferredSignInStrategy = environment.displayConfig.preferredSignInStrategy;
+
+  const handleResponse = (resource: SessionVerificationResource) =>
+    toResult(resource, preferredSignInStrategy, webAuthnSupported);
+
+  return {
+    status: 'ready',
+    isActive,
+    supportEmail: environment.displayConfig.supportEmail ?? '',
```
alexcarpenter: should we use the `useSupportEmail` hook here? there are a few things it does to ensure there is always an email so we don't fall back to nothing.

`packages/ui/src/hooks/useSupportEmail.ts`

packages/ui/src/mosaic/features/reverification/reverification.controller.ts
```diff
@@ -0,0 +1,405 @@
...
+  }
+  return method.strategy === 'passkey' ? ctx.deps.verifyPasskey() : ctx.deps.attempt(method, ctx.inputValue);
+}
+
+const applyResult = assign<DoneInvokeEvent<ReverificationResult>>((_, event) => ({
+  methods: event.output.methods,
+  activeMethod: event.output.startingMethod,
+  inputValue: '',
+  errorMessage: undefined,
+  canResend: true,
```
alexcarpenter:
```suggestion
  canResend: true,
  direction: 1,
```

new result should always moves forward otherwise a prior BACK leaks into the next animation

alexcarpenter (commented):
this is looking good!

---

here are a few non-blocking nits that the 🤖's surfaced 

- verifyPasskey (model line 143): Looks like attempt already covers passkey, so could we drop this one? Same in the view, onVerifyPasskey and onSubmit both just send SUBMIT.
- prepare (model line 175): Optional. The silent default: return made me pause. If prepare only accepted the OTP methods, the needsPrepare guards in the controller could become a plain type guard.
- viewStep (controller line 294): The loading check, pendingStates, and this function all list state names by hand. Worth folding into one lookup so they can't drift?
- identifier (types line 35): I don't think the view reads this anymore.
- onBack (types line 30): Could this just be required? BACK is a no-op outside picker and help anyway, and it'd drop the ?? (() => {}) in the view.
- fallback error string (controller line 64): Third copy of this string. The delete section controller has one, and reverificationBase.unstable__errors__generic already defines it but isn't used. Worth a shared helper that reads from messages?

packages/ui/src/mosaic/features/reverification/reverification.controller.ts
Ephem: Thanks, good catch, done!

packages/ui/src/mosaic/features/reverification/reverification.model.tsx
Ephem: Good catch, we can't use the existing `useSupportEmail` hook since it relies on contexts we don't have and creates a coupling to `UI`, but I created a new `useMosaicSupportEmail` (mirroring the environment one). Might not be the scaleable approach, but solves it for now.

packages/ui/src/mosaic/features/reverification/reverification.controller.ts
Ephem: Another nice catch and together with the thing you found in Slack this led me down a rabbit hole. I ended up refactoring things quite a bit.

This card now does have a correct pending state during preparing, rows and Back button are disabled during that. If prepare errors, we go to the factor card to display the error there. I did this via a new `methodPickerPreparing` machine state.

If you reach a factor card that needs prepare right from the start however, we now don't have a pending state. Instead we fake that the prepare has completed and all actions are available from the start. If prepare fails, we still show the error. If the user submits during this state, we queue that submit until after prepare comes back. IF we later want some kind of explicit pending state here like a small "Sending code ...", we can easily implement that.

I also folded the resending state into the preparing one. These two now both have the optimistic UI approach with no pending state.

I also noticed both that resending wasn't locked like it should be and we did not have a visible resending countdown, so I added that. We can implement that a few different ways, but I put it in the controller.

Ephem:
@alexcarpenter Great feedback! I did all except the `viewStep` mapping (not against it, just didn't feel worth it right now but happy for pushback), and the error fallback. The latter feels like a bigger i18n lift so while we should I'm kicking the can.

alexcarpenter (approved):

PR #9652, opened 2026-09-03
feat(ui): formalize the Mosaic Profile surface and add a Drawer

maxyinger:
## Description

Turns the user page's ad-hoc sidebar into a generic Mosaic `Profile` surface, for the user profile now and the organization profile (and their merge) later. Adds the styled `Drawer` it uses for compact navigation.

**`Profile`** (`components/profile/`), compositional; `UserProfileView` is the prop-driven shell on top (`activePage`, `pages`, `customPages`, `pageOrder`, `onPageChange`, `elevation`).

| Part | Renders | Notes |
| --- | --- | --- |
| `Root` | `div`, the `cl-profile` container | `value`, `onValueChange`, `renderBranding`, `elevation: 'card' \| 'flush'` |
| `Title` | visually hidden `h2` | names the nav, the sheet, and the dialog it sits in |
| `Nav` / `NavItem` | `nav` > tablist > tabs | a column; compact, the same tablist inside a `Drawer` |
| `PageTitle` | `h3` | compact, the whole headline is the sheet's trigger |
| `Content` | `div` | the scroll region; flush, the page scrolls |
| `TabPanel` | tab panel | `shouldForceMount` for CSS page transitions |

- **Elevation**: `card` is framed (border, radius, fixed `45rem`, pages scroll inside); `flush` is the page's own content, columns `2.5rem` apart. An `inline` dialog implies `flush`; a `profile` dialog fills the popup.
- **Compact** (`<= 48rem` of its own width, measured with `useMeasure`, since where the tablist renders is a DOM decision): full-screen, no frame, nav in a two-thirds sheet. The in-place column is also hidden by the container query, so nothing stacks over the page before hydration.
- Dialog size `panel` → `profile`. `Branding` hoisted out of `Card`. New tokens `--cl-radius-2xl`, `--cl-shadow-card`.

**`Drawer`** (`components/drawer/`): first styled consumer of the headless drawer. Prompt surface, flush sides and bottom, grip, bottom bleed for the rubber band, nested scrim inside a modal dialog. Always two thirds tall.

**Headless fixes**

- `useRender` spreads state attributes first, so a `[data-selected]`-anchored mark transitions both ways (Chrome flushes style on the `tabindex` write).
- Drawer drag: the inner-scroll walk stops at the document; upward drag at rest rubber-bands unless inner content scrolls; swipe vars mirror onto the backdrop; `window`-level release ends the gesture; close resets it; `data-swiping` lands on the first committed move.
- `useFocusTarget` shared by Dialog and Drawer; the function form resolves at restore time.

Swingset: `Profile` and `Drawer` pages, `user-page` renamed `user-profile`.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

https://claude.ai/code/session_01UNzajPkZEavBH31fpxG7sx


packages/ui/src/mosaic/components/profile/profile.tsx
```diff
@@ -0,0 +1,526 @@
...
+  const { titleId, renderBranding, compact, navOpen, closeNav, root, inline } = profile;
+  // The headline that opened the sheet belongs to the page a choice just left, so the sheet's own
+  // return-focus would land on nothing. The headline of the page now showing is the same control,
+  // on the destination.
+  const finalFocus = React.useCallback(
+    () =>
+      root?.querySelector<HTMLElement>('.cl-profile-tab-panel:not([hidden]):not([inert]) .cl-profile-nav-trigger') ??
+      null,
+    [root],
+  );
```
alexcarpenter: I think this approach is causing a timing issue with when the page becomes inert and returning focus causing the focus to move back to the body.

Can we switch to using a registry like we do in tabs with (registerTab / getTabElement)?

- TabPanelContext carries the page's identity ({ titleId, value }) rather than just the title id
- ProfileContext gains value, registerPageTitle and pageTitleFor, and drops root
- Profile.PageTitle registers its own button under its page's value — a no-op outside a Profile, so a page rendered standalone is unaffected
- finalFocus becomes () => pageTitleFor(value)

packages/headless/src/utils/use-render.tsx
```diff
@@ -215,7 +215,13 @@ export function useRender<
...
   }
 
-  const computedProps = { ...props, ...dataAttrs };
+  // State attributes lead, and still win: a key's position is fixed by its first spread, its value
+  // by its last. React writes changed attributes in key order, and Chrome flushes style when
+  // `tabindex` changes on the focused element, so a state marker written after a roving `tabindex`
+  // would be missing from that flush — an anchor named on `[data-selected]` resolves to nothing for
+  // one recalc and its transition snaps. Ahead of the rest, the marker is in place before any write
+  // that can flush.
+  const computedProps = { ...dataAttrs, ...props, ...dataAttrs };
```
alexcarpenter: this this looks like maybe a constraint that only tabs has atm. Can it stay in tabs-tab.tsx? that file already controls its own merge order.

packages/ui/src/mosaic/components/profile/profile.styles.ts
```diff
@@ -0,0 +1,329 @@
...
+} from '../../tokens.stylex';
+import { scrollAreaRoot, scrollAreaViewport } from '../scroll-area';
+
+/**
+ * The compact layout — navigation on top, as a row — queried against the profile's OWN width
+ * rather than the window's, so the same surface collapses in a narrow layout slot, an inline
+ * dialog, or a phone alike. A container cannot query itself, which is why the grid lives on an
+ * inner element: the root is the container, the layout inside it is what the query reshapes.
+ */
+const compact = '@container cl-profile (max-width: 48rem)' as const;
```
alexcarpenter: can we move `48rem` into a breakpoints.stylex.ts file with defineConst usage since we are using this across multiple files now?

packages/ui/src/mosaic/components/profile/profile.tsx
```diff
@@ -0,0 +1,526 @@
...
+const COMPACT_MAX_WIDTH_REM = 48;
+
+/** `width <= 48rem`, the way the container query reads it. Unmeasured, or not laid out, is wide. */
+function isCompact(width: number | null): boolean {
+  if (width === null || width === 0) {
+    return false;
+  }
+  const rem = parseFloat(getComputedStyle(document.documentElement).fontSize) || 16;
+  return width <= COMPACT_MAX_WIDTH_REM * rem;
+}
```
alexcarpenter: Is there a way to avoid this maybe? Could we add a sentinel that is 1px by default then 2px when the compact breakpoint is hit, then you only need to check if this element is 1px or 2px instead of computing the rem values like this?

```
sentinel: {
  position: 'absolute',
  visibility: 'hidden',
  pointerEvents: 'none',
  inlineSize: '1px',
  [compact]: { inlineSize: '2px' },
}
```

packages/ui/src/mosaic/components/profile/profile.tsx
```diff
@@ -0,0 +1,526 @@
...
+    children,
+    render,
+    className,
+    style,
+    ...rest
+  },
+  ref,
+) {
+  const dialog = React.useContext(DialogContext);
+  const inline = elevation === 'flush' || (dialog?.inline ?? false);
```
alexcarpenter, line 123: this same type of condition happens a handful of times, can we bundle it up into a hook?

something like 

```
export function isOverlayDialog(surface: DialogContextValue | null): boolean {
  return surface !== null && !surface.inline;
}
```

packages/ui/src/mosaic/components/profile/profile.tsx
```diff
@@ -0,0 +1,526 @@
...
+  );
+  const element = useRender({
+    defaultTagName: 'nav',
+    render,
+    ref,
+    props: {
+      'aria-labelledby': titleId,
+      ...mergeStyleProps(
+        themeProps('profile-nav', { compact }),
+        stylex.props(reset.base, styles.nav, inline && styles.navInline, compact && styles.navInSheet),
```
alexcarpenter: navInline and navInSheet look to be identical. Should we collapse into to one navFlush and (inline || compact) && styles.navFlush? The two comments give different reasons for the same rule.

packages/ui/src/mosaic/components/drawer/drawer.tsx
```diff
@@ -0,0 +1,120 @@
...
+const Description = React.forwardRef<HTMLParagraphElement, DrawerDescriptionProps>(
+  function DrawerDescription(props, ref) {
+    return (
+      <Primitive.Description
+        ref={ref}
+        {...props}
+      />
+    );
+  },
+);
```
alexcarpenter: instead of redefining these, can we just pass them through 

```
export const Drawer = {
  Root,
  Popup,
  Trigger: Primitive.Trigger,
  Title: Primitive.Title,
  Description: Primitive.Description,
  Close: Primitive.Close,
};
```

packages/ui/src/mosaic/utils/apply-order.ts
```diff
@@ -0,0 +1,22 @@
...
+  idOf: (item: T) => string,
+): T[] {
+  const unique = items.filter((item, index, all) => all.findIndex(other => idOf(other) === idOf(item)) === index);
+  if (!order?.length) {
+    return unique;
+  }
+
+  const named = [...new Set(order)].flatMap(id => unique.filter(item => idOf(item) === id));
+  return [...named, ...unique.filter(item => !named.includes(item))];
+}
```
alexcarpenter, line 22: applyOrder and the profile types are in better homes now. could the two shims go with them?

apply-order.ts is a shared util now, maybe should move utils/apply-order.test.ts would match how the rest of utils/ is laid out.

packages/ui/src/mosaic/user-profile/user-profile.layout.ts
```diff
@@ -0,0 +1,46 @@
+import type { IconName } from '../icons/registry';
+import { applyOrder } from '../utils/apply-order';
+import type { CustomProfilePage, UserProfilePageId, UserProfilePages } from './user-profile.types';
+
+/** The built-in pages in the order the profile lists them before a consumer reorders anything. */
+export const USER_PROFILE_PAGE_IDS: readonly UserProfilePageId[] = ['account', 'security', 'billing', 'apiKeys'];
```
alexcarpenter, line 6: USER_PROFILE_PAGE_IDS in user-profile.layout.ts is meant to be the canonical order, but this hook builds the same list by hand and never imports it. reorder the constant and this quietly disagrees, which the comment above says would break things.

could it filter the constant instead? then it's the same shape as getAvailableUserProfilePages, just a different check.

packages/ui/src/mosaic/user-profile/user-profile.layout.ts
```diff
@@ -0,0 +1,46 @@
...
+  order?: readonly string[],
+): UserProfileNavEntry[] {
+  const customById = new Map(customPages.map(page => [page.path, page]));
+  const entries: UserProfileNavEntry[] = [
+    ...builtIn.map((id): UserProfileNavEntry => {
+      const custom = customById.get(id);
+      return custom ? { id, custom } : { id };
+    }),
+    ...customPages
+      .filter(page => !builtIn.includes(page.path as UserProfilePageId))
```
alexcarpenter: page.path as UserProfilePageId is here to get past .includes. a new Set<string>(builtIn) above and !builtInIds.has(page.path) does the same thing without the cast, and there's already a Map being built on the line above it

packages/headless/src/primitives/drawer/use-drawer-drag.ts
```diff
@@ -72,6 +72,10 @@ export function useDrawerDrag(opts: UseDrawerDragOptions): UseDrawerDragReturn {
   // Removes the current iOS `touchend` fallback listener (see `onPointerDown`),
   // so it never outlives its gesture or piles up across gestures.
   const removeTouchEnd = useRef<(() => void) | null>(null);
+  // Removes the window-level release listeners armed for the current gesture (see `onPointerDown`).
+  const removeWindowRelease = useRef<(() => void) | null>(null);
+  // `onRelease` is defined after `onPointerDown`, which needs to arm it; read through a ref.
+  const onReleaseRef = useRef<((e: { clientY: number }) => void) | null>(null);
```
alexcarpenter: can we move `onRelease` above `onPointerDown` and let the window
listener close over it directly. deletes the ref, the render-phase
`onReleaseRef.current = onRelease` assignment on line 381, and the `?.` in
`onWindowRelease`.

worth avoiding the render-phase write on its own. assigning to a ref
during render isn't safe under concurrent rendering

packages/headless/src/primitives/drawer/use-drawer-drag.ts
```diff
@@ -130,11 +169,27 @@ export function useDrawerDrag(opts: UseDrawerDragOptions): UseDrawerDragReturn {
...
+      for (let el: HTMLElement | null = target; el; el = el.parentElement) {
+        if (el.scrollHeight > el.clientHeight && el.scrollTop + el.clientHeight < el.scrollHeight - 1) {
+          return false;
+        }
+        // Nothing above the sheet is inner content — its box may well be taller than the screen.
+        if (el === sheet) {
+          break;
+        }
+      }
+      return true;
```
alexcarpenter: this loop and the one below it are the same check in opposite directions:
is there room left to scroll the way the content would move. could be one
loop with a `down` param instead of two.

if you merge them, the down loop's `role="dialog"` check on 198 can go
away too, since `sheet` here is already that element.

packages/ui/src/mosaic/components/profile/profile.tsx
```diff
@@ -0,0 +1,526 @@
...
+ *   <Profile.Nav>
+ *     <Profile.NavItem value='account' icon={<Icon name='user-circle' size='sm' />}>Account</Profile.NavItem>
+ *   </Profile.Nav>
+ *   <Profile.Content>
+ *     <Profile.TabPanel value='account'>…</Profile.TabPanel>
+ *   </Profile.Content>
+ * </Profile.Root>
+ * ```
+ */
+export const Profile = { Root, Title, Nav, NavItem, PageTitle, Content, TabPanel };
```
alexcarpenter: small nit, but we have Nav, NavItem, but then have Content, TabPanel.

should we have `ContentPanel` vs `TabPanel`?

maxyinger: yea `ContentPanel` makes sense

packages/ui/src/mosaic/components/profile/profile.tsx
maxyinger: ooo. clever.

packages/ui/src/mosaic/components/profile/profile.styles.ts
maxyinger: can do this as a follow up. extracting the breakpoint out for the container queries was breaking builds and I didn't want to tack on postcss config updates with everything else going on in here

alexcarpenter (approved):

PR #9654, opened 2026-09-04
feat(ui): add input group

austincalvelage:
## Description

https://swingset-git-austin-input-group.clerkstage.dev/components/input-group

CleanShot 2026-09-03 at 19 30 10@2x

Adds Mosaic `InputGroup` for composing text and controls inside shared field chrome. It coordinates sizing, field state, and focus behavior with a headless `Input`.


packages/ui/src/mosaic/components/input/input.tsx
```diff
@@ -32,8 +33,9 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(function Mos
     'aria-describedby': ariaDescribedBy,
     ...rest
   },
-  ref,
+  forwardedRef,
 ) {
+  const inputGroup = useOptionalInputGroupContext();
```
maxyinger, line 39: can we just use presence of this to determine if it should be the headless variant?

packages/ui/src/mosaic/components/input-group/input-group.tsx
```diff
@@ -0,0 +1,116 @@
...
+  ref,
+) {
+  const field = useOptionalFieldContext();
+  const disabled = disabledProp ?? field?.disabled ?? false;
+  const invalid = invalidProp ?? field?.invalid ?? false;
+  const inputRef = React.useRef<HTMLInputElement | null>(null);
+  const focusInput = React.useCallback(() => inputRef.current?.focus(), []);
+  const setInput = React.useCallback((node: HTMLInputElement | null) => {
+    inputRef.current = node;
+  }, []);
```
maxyinger: idc too much about semantics, but since this is just a standard callback ref, would be cool with it just being like `inputRef` or something. the set makes me feel like it's holding it in state somewhere

packages/ui/src/mosaic/components/input/input.tsx
```diff
@@ -42,18 +44,32 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(function Mos
...
+    (node: HTMLInputElement | null) => {
+      setGroupInput?.(node);
+      if (typeof forwardedRef === 'function') {
+        forwardedRef(node);
+      } else if (forwardedRef) {
+        forwardedRef.current = node;
+      }
+    },
+    [forwardedRef, setGroupInput],
+  );
```
maxyinger: do we have a combine refs util? feels like it'd be worth adding one if we don't already so we can just do:
```
ref: combineRefs(forwardedRef, inputGroup?.inputRef),
```

packages/ui/src/mosaic/components/input-group/input-group.tsx
```diff
@@ -0,0 +1,116 @@
...
+        className,
+        style,
+      ),
+      onPointerDown: (event: React.PointerEvent<HTMLSpanElement>) => {
+        onPointerDown?.(event);
+        if (!event.defaultPrevented) {
+          event.preventDefault();
+          group.focusInput();
+        }
+      },
```
maxyinger: wdyt about just making this `pointer-events: none` and adding an onclick to the group that focuses the input?

packages/ui/src/mosaic/components/input-group/input-group.styles.ts
```diff
@@ -0,0 +1,82 @@
...
+  root: {
+    overflow: 'hidden',
+    alignItems: 'center',
+    display: 'flex',
+    minHeight: { default: null, '@media (pointer: coarse)': targetVars['--cl-target-coarse'] },
+    width: '100%',
+  },
+  text: {
+    alignItems: 'center',
+    color: colorVars['--cl-color-neutral-faded'],
```
maxyinger: almost like the color of this more in the disabled state 🤔 . This almost feels too emphasized, or at least a little off with the placeholders in dark mode.

packages/ui/src/mosaic/components/input-group/input-group.tsx
```diff
@@ -0,0 +1,116 @@
...
+        stylex.props(size === 'xs' && compactActionInsets[group.size]),
+        className,
+        style,
+      )}
+      {...otherProps}
+    />
+  );
+});
+
+export const InputGroup = { Root, Text, Action };
```
maxyinger: wdyt about adding like an addon container to this like shadcn?

https://ui.shadcn.com/docs/components/base/input-group#button

packages/ui/src/mosaic/components/input-group/input-group.styles.ts
maxyinger: maybe we just match the color of the placeholder for this?

maxyinger (commented):
I know we don't do concentic border radii on this project a lot, but wondering if we use this size button anywhere else and if it makes sense to just make its radius concentric with the input

packages/ui/src/mosaic/components/input/input.tsx
austincalvelage: Yeah! This defaults to 'ghost' now from within the context

packages/ui/src/mosaic/components/input-group/input-group.tsx
austincalvelage: Good call renamed

packages/ui/src/mosaic/components/input/input.tsx
austincalvelage: Good catch. `useRender` supports this switched to `ref: [forwardedRef, inputGroup?.inputRef]`

packages/ui/src/mosaic/components/input-group/input-group.tsx
austincalvelage: [Text ignores pointer events](https://github.com/clerk/javascript/blob/b57eeadbfdaa6d28239cf04b644c37c670213292/packages/ui/src/mosaic/components/input-group/input-group.styles.ts#L36) and [the group handles focusing the input](https://github.com/clerk/javascript/blob/b57eeadbfdaa6d28239cf04b644c37c670213292/packages/ui/src/mosaic/components/input-group/input-group.tsx#L48-L64)

packages/ui/src/mosaic/components/input-group/input-group.styles.ts
austincalvelage: Using same color as placeholder

packages/ui/src/mosaic/components/input-group/input-group.tsx
austincalvelage: Added InputGroup.Start and InputGroup.End containers

packages/ui/src/mosaic/components/input-group/input-group.tsx
```diff
@@ -0,0 +1,119 @@
...
+      }) as const,
+    [group.disabled, group.size],
+  );
+  const element = useRender({
+    defaultTagName: 'span',
+    render,
+    ref,
+    props: {
+      ...mergeStyleProps(
+        themeProps(`input-group-${side}`, { size: group.size, disabled: group.disabled }),
```
alexcarpenter: small thing, but I think we should avoid code like `input-group-${side}`.

when working with agents, its harder to find things that are concatenated like this.

ref: https://morizbuesing.com/blog/greppability-code-metric/

alexcarpenter (approved):

packages/ui/src/mosaic/components/input-group/input-group.tsx
austincalvelage: Totally fair and valid. Let me update

alexcarpenter: mainly considering for skills where they'd be able to scan the components to find these classNames easily.

PR #9655, opened 2026-09-04
feat(ui): add combobox

austincalvelage:
## Description

Adds a styled Mosaic `Combobox` with searchable options in floating and inline list layouts. Its input supports the headless variant for use inside `InputGroup`.

https://swingset-git-austin-combobox.clerkstage.dev/components/combobox


packages/ui/src/mosaic/components/combobox/combobox.tsx
```diff
@@ -0,0 +1,167 @@
...
+            stylex.props(reset.base, scrollAreaRoot, styles.popup),
+            className,
+            style,
+          )}
+          {...rest}
+        >
+          <div
+            {...mergeStyleProps(
+              themeProps('combobox-viewport'),
+              stylex.props(reset.base, ...scrollAreaViewport(), styles.viewport),
```
maxyinger, line 122: keep goin back n forth on if we should have a scroll bar on here at all. I like the fade regardless tho, almost want to make it a bit smaller tho. think it's customizeable through css vars, but can't remember

packages/ui/src/mosaic/components/combobox/combobox.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { Input, type InputVariant } from '../input';
+import { useOptionalInputGroupContext } from '../input-group/input-group.context';
+import { scrollAreaRoot, scrollAreaViewport } from '../scroll-area';
+import { styles } from './combobox.styles';
+
+export type ComboboxRootProps = AutocompleteProps;
+export type ComboboxSize = 'sm' | 'md' | 'lg';
+export type ComboboxTriggerProps = MosaicComponentProps<'button'>;
+
+export const ComboboxTrigger = React.forwardRef<HTMLButtonElement, ComboboxTriggerProps>(function MosaicComboboxTrigger(
```
maxyinger: what are thoughts on making the whole thing a trigger so when you click into the input, the popup appears?

packages/ui/src/mosaic/components/combobox/combobox.tsx
```diff
@@ -0,0 +1,167 @@
...
+export const ComboboxPopup = React.forwardRef<HTMLDivElement, ComboboxPopupProps>(function MosaicComboboxPopup(
+  { portalRoot, className, style, children, ...rest },
+  ref,
+) {
+  return (
+    <Primitive.Portal root={portalRoot}>
+      <Primitive.Positioner
+        {...mergeStyleProps(themeProps('combobox-positioner'), stylex.props(reset.base, styles.positioner))}
+      >
+        <Primitive.Popup
```
maxyinger: We might need to add a `Freeze` on close for the popup primitive. if it doesn't make sense in the primitive we should add one here so it doesn't flash when we select and it exits

packages/ui/src/mosaic/components/combobox/combobox.tsx
```diff
@@ -0,0 +1,167 @@
+'use client';
+
+import type { AutocompleteProps } from '@clerk/headless/autocomplete';
+import { Autocomplete as Primitive } from '@clerk/headless/autocomplete';
```
maxyinger: one thing that's interesting here is that baseui separates these out:
https://ui.shadcn.com/docs/components/aria/combobox
https://base-ui.com/react/components/autocomplete

might just be our nomenclature meaning different things, but I think they're supposed to have slightly different functionality. this is the interesting blurb there on the autocomplete:

> Use Combobox instead of Autocomplete if the selection should be remembered and the input value cannot be custom. Unlike Combobox, Autocomplete’s input can contain free-form text, as its suggestions only optionally autocomplete the text

maxyinger: I wonder if that's functionality we're kinda baking into the country code selector that should just be intrinsic to the combobox?

maxyinger (commented):
maybe make this whole thing align and give the popup some more margin so it's not touching the focus shadow?

Image

packages/ui/src/mosaic/components/combobox/combobox.tsx
austincalvelage: Added primitive

packages/swingset/src/stories/combobox.mdx
```diff
@@ -0,0 +1,77 @@
...
+Deleting all text clears the selection and calls `onValueChange(null)`.
+Opening with the trigger or arrow keys keeps the selected label visible and shows all options until you edit it.
+
+## Example
+
+Use the trigger to show every option, or type to open and filter the list. Use the arrow keys and
+Enter to select an option.
+
+`OptionIndicator` shows a check beside the selected option, independently of the hover or keyboard highlight.
+
```
maxyinger: wonder how much of this we can remove. Kyle went through yesterday and removed most of these descriptions here and just led with the example

packages/ui/src/mosaic/components/combobox/combobox.tsx
maxyinger: I'm liking how it feels rn. ignore this

maxyinger (approved):

PR #9656, opened 2026-09-04
feat(ui): Add phone input

austincalvelage:
## Description

https://swingset-git-austin-phone-input.clerkstage.dev/components/phone-input

Adds a Mosaic `PhoneInput` with a searchable country picker, formatted number entry, and normalized E.164 values. It integrates with Mosaic `Field` and supports controlled and uncontrolled usage.


packages/ui/src/mosaic/components/phone-input/phone-input.tsx
```diff
@@ -0,0 +1,341 @@
...
+                  <Combobox.Empty>{noResultsMessage}</Combobox.Empty>
+                )}
+              </Combobox.List>
+            </Combobox.Root>
+          </Popover.Popup>
+        </Popover.Root>
+
+        <InputGroup.Text
+          aria-hidden='true'
+          {...mergeStyleProps(themeProps('phone-input-prefix'), stylex.props(styles.prefix))}
```
maxyinger, line 300: like I wonder if we use the addon api if this can just be an addon aligned to the start?

similar to http input in this example:
https://ui.shadcn.com/docs/components/aria/input-group#button

packages/ui/src/mosaic/components/phone-input/phone-input.tsx
```diff
@@ -0,0 +1,341 @@
...
+              value={country.iso}
+              onValueChange={iso => {
+                const nextCountry = countryOptions.find(option => option.iso === iso);
+                if (nextCountry) {
+                  setCountry(nextCountry);
+                }
+              }}
+            >
+              <Field.Root>
+                <InputGroup.Root
```
maxyinger, line 250: base has a good example here. it's nice to see how they compose everything:
https://base-ui.com/react/components/combobox#input-inside-popup

this is a little different cus of the main input, but figured I mention the reference

maxyinger (commented):
still needs some polish, but getting there

CleanShot 2026-09-04 at 17 00 41@2x

packages/ui/src/mosaic/components/phone-input/phone-input.tsx
```diff
@@ -0,0 +1,339 @@
...
+                  {...themeProps('phone-input-country-search')}
+                >
+                  <InputGroup.Start>
+                    <Icon
+                      name='search'
+                      size='sm'
+                      aria-hidden='true'
+                    />
+                  </InputGroup.Start>
+                  <Combobox.Input
```
maxyinger, line 261: should this be focused on open?

packages/ui/src/mosaic/components/phone-input/phone-input.tsx
```diff
@@ -0,0 +1,339 @@
...
+                      {...themeProps('phone-input-country-option')}
+                    >
+                      <span
+                        aria-hidden='true'
+                        {...stylex.props(reset.base, styles.flag)}
+                      >
+                        {getFlagEmojiFromCountryIso(option.iso)}
+                      </span>
+                      <span {...stylex.props(reset.base, styles.optionName)}>{option.name}</span>
+                      <span {...stylex.props(reset.base, styles.optionCode)}>+{option.code}</span>
```
maxyinger, line 284: might need a whitespace nowrap on the country code

Image

maxyinger (approved):
looks good. no blockers. can figure out how we want to handle it so we get nice spacing with the country trigger in a follow up

PR #9665, opened 2026-09-04
docs(nextjs): Point agents at clerk init from the exports they read

shane-kercheval:
## Description

Coding agents that add Clerk to a Next.js app learn the API from the package's type declaration files under `node_modules/@clerk/nextjs/dist/types/` — `index.d.ts`, `server/clerkMiddleware.d.ts` and `app-router/server/auth.d.ts`. None of them mention that `npx clerk@latest init` exists or that it works without a Clerk account. The only mention of the CLI in the package today is the doc comment on the removed `<SignedIn>`, `<SignedOut>` and `<Protect>` stubs, which an agent reaches only after using a component that no longer exists.

An agent that does not already know the CLI may therefore conclude a Clerk account and dashboard keys are required, ask the developer for them, and either stall or write a placeholder key to get the build through.

This change adds a note to the doc comments on `ClerkProvider`, `clerkMiddleware()` and `auth()`, the three exports every Clerk-on-Next.js app touches:

> Need Clerk keys? In your terminal run `npx clerk@latest init`. It creates a Clerk application and writes the keys to your .env file. No Clerk account or login required and the command is non-interactive. See the [Clerk CLI docs](https://clerk.com/docs/cli).

The wording is similar to the existing missing-key error in `@clerk/shared`. Because the note lives in doc comments it is emitted into the shipped `.d.ts` files, so it is in front of an agent at the moment it is deciding how to get keys, and editors show it on hover for human developers.

Separately, the README's Prerequisites listed "An existing Clerk application" with a sign-up link, which is not true on the CLI path. Prerequisites and Installation now lead with `clerk init`, keeping the Quickstart as the by-hand alternative.

No runtime change.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/nextjs/README.md
```diff
@@ -34,11 +34,15 @@
 - Next.js 15.2.8 or later
 - React 18 or later
 - Node.js `>=20.9.0` or later
-- An existing Clerk application. [Create your account for free](https://dashboard.clerk.com/sign-up?utm_source=github&utm_medium=clerk_nextjs).
+- A Clerk application. Run `npx -y clerk@latest init` to create one and write its keys to your `.env.local` file. No Clerk account required to get started — it can provision temporary development keys. Or [create your account](https://dashboard.clerk.com/sign-up?utm_source=github&utm_medium=clerk_nextjs) and copy the keys from the dashboard.
 
 ### Installation
 
-The fastest way to get started with Clerk is by following the [Next.js Quickstart](https://clerk.com/docs/quickstarts/nextjs?utm_source=github&utm_medium=clerk_nextjs).
+The fastest way to get started is the Clerk CLI. In an existing Next.js project, run `npx -y clerk@latest init`. In an empty directory, run `npx -y clerk@latest init --framework next` and it scaffolds the Next.js app as well.
```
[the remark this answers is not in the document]
SarahSoutoul: Was about to flag this when I realized CodeRabbit already had - worth specifying the package manager for this empty-directory command I think since there is no project or lockfile to detect it from @shane-kercheval

packages/nextjs/README.md
```diff
@@ -34,11 +34,15 @@
...
+
+Either way it installs `@clerk/nextjs`, creates a Clerk application, writes the keys to your `.env.local` file, and adds `<ClerkProvider>`, the middleware, and sign-in and sign-up pages. No Clerk account required to get started — it can provision temporary development keys.
+
+To set things up by hand instead, follow the [Next.js Quickstart](https://clerk.com/docs/quickstarts/nextjs?utm_source=github&utm_medium=clerk_nextjs).
```
SarahSoutoul:
```suggestion
To set things up by hand instead, follow the [Next.js Quickstart](https://clerk.com/docs/nextjs/getting-started/quickstart?utm_source=github&utm_medium=clerk_nextjs).
```
That link was outdated, and redirected to the new link dropping the UTM parameters as a result. So I'd update it to this.

SarahSoutoul:
Left two comments + pushed a small docs-review update in [c90a62c](https://github.com/clerk/javascript/commit/c90a62ce97e544810cf054df5b5d269c4c2931f1). It:

- Replaces the question-led "Need Clerk keys?" wording with a direct setup instruction
- Format `.env.local` as code and clarifies that the CLI can provision temporary development keys

One thing I think we should clarify: the PR description says these changes affect `/docs/reference/nextjs/app-router/auth` and `/docs/reference/nextjs/clerk-middleware`, but those pages do not currently consume these particular TypeDoc fragments. Their introductions are hand-authored in `clerk-docs`, so this PR updates the package JSDoc/declaration files and generated TypeDoc output, but the guidance will not appear on those live docs pages as things stand today.

Is that something we're okay with? It still directly addresses the eval behavior, since the agents opened the package's `.d.ts` files. If we also want this guidance in the docs' LLM-readable Markdown, would explicit `<If is="llm">` blocks on the `auth()`, `clerkMiddleware()`, and `<ClerkProvider>` pages be a better fit than replacing their human-facing introductions with TypeDoc? Tagging @manovotny on this to open the discussion.

shane-kercheval:
Good catch. I checked both pages and you're right, their intros are hand-authored in clerk-docs, so nothing from this PR reaches them. I've dropped that line from the description. The `.d.ts` channel was the actual goal, so I'm fine with the PR as-is.

On `<If is="llm">`: agreed that's the better fit for the docs side. `clerk-middleware.mdx` already has a block saying essentially this, so it would just be adding matching blocks to `auth()` and `<ClerkProvider>`. Happy to open a separate PR in clerk/clerk for that if you and @manovotny agree.

Thanks for the rewrite on the doc comments too. Reads better than mine, and I've left it as-is.

packages/nextjs/README.md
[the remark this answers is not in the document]
shane-kercheval: Done in 84315fd. Went with `--pm npm` rather than CodeRabbit's `pnpm` so it doesn't clash with the `npx` in the same sentence, but happy to switch if you'd rather. Also added a short note that there is no lockfile to detect a package manager from, so the flag doesn't look arbitrary.

packages/nextjs/README.md
shane-kercheval: Applied, thanks. Didn't realize the redirect was dropping the UTMs.

packages/nextjs/README.md
```diff
@@ -34,11 +34,15 @@
...
+- A Clerk application. Run `npx -y clerk@latest init` to create one and write its keys to your `.env.local` file. No Clerk account required to get started — it can provision temporary development keys. Or [create your account](https://dashboard.clerk.com/sign-up?utm_source=github&utm_medium=clerk_nextjs) and copy the keys from the dashboard.
 
 ### Installation
 
-The fastest way to get started with Clerk is by following the [Next.js Quickstart](https://clerk.com/docs/quickstarts/nextjs?utm_source=github&utm_medium=clerk_nextjs).
+The fastest way to get started is the Clerk CLI. In an existing Next.js project, run `npx -y clerk@latest init`. In an empty directory, run `npx -y clerk@latest init --framework next --pm npm` and it scaffolds the Next.js app as well; there is no lockfile to detect a package manager from, so name one.
+
+Either way it installs `@clerk/nextjs`, creates a Clerk application, writes the keys to your `.env.local` file, and adds `<ClerkProvider>`, the middleware, and sign-in and sign-up pages. No Clerk account required to get started — it can provision temporary development keys.
```
manovotny: Checked this against the CLI source (`clerk@3.3`, `pickStrategy` in `commands/init/index.ts`): the accountless path only fires for agent mode or a new-project bootstrap — `!authed && supportsKeyless && (agent || isBootstrap)`. A signed-out person running bare `npx -y clerk@latest init` in an existing project falls through to the authenticated path and hits an interactive login, so "No Clerk account required to get started" doesn't hold there. The empty-directory command above is a bootstrap, so that one's fine.

Same sentence, two smaller ones: authenticated `init` can link an existing app instead of creating one, and it writes to the first existing `.env*` file (`.env.local` is only the fallback when none exist).

Want to scope the no-account line to agents and new projects, or point the existing-project path at `--keyless`?

packages/nextjs/README.md
```diff
@@ -34,11 +34,15 @@
 - Next.js 15.2.8 or later
 - React 18 or later
 - Node.js `>=20.9.0` or later
-- An existing Clerk application. [Create your account for free](https://dashboard.clerk.com/sign-up?utm_source=github&utm_medium=clerk_nextjs).
+- A Clerk application. Run `npx -y clerk@latest init` to create one and write its keys to your `.env.local` file. No Clerk account required to get started — it can provision temporary development keys. Or [create your account](https://dashboard.clerk.com/sign-up?utm_source=github&utm_medium=clerk_nextjs) and copy the keys from the dashboard.
```
manovotny: Same `No Clerk account required` scoping as my note on line 43 — for a signed-out human in an existing project, bare `init` goes to login, not accountless.

.changeset/nextjs-cli-hint-in-types.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/nextjs': patch
+---
+
+Add a note to the `ClerkProvider`, `clerkMiddleware()` and `auth()` doc comments explaining that `npx clerk@latest init` creates a Clerk application and writes its keys with no Clerk account or login required. Update the README prerequisites and installation section to say the same, replacing the statement that an existing Clerk application and account are required.
```
manovotny: This states `init` "creates a Clerk application and writes its keys with no Clerk account or login required" as an absolute, but the JSDoc it's describing softened that to "can provision." Since this ships as the changelog entry, matching the JSDoc keeps it from promising more than the CLI does for existing-project runs.

```suggestion
Add a note to the `ClerkProvider`, `clerkMiddleware()` and `auth()` doc comments explaining that `npx clerk@latest init` can provision temporary development keys without a Clerk account. Update the README prerequisites and installation section to say the same, replacing the statement that an existing Clerk application and account are required.
```

shane-kercheval:
@manovotny, thanks for the review and feedback. Good catches.

- Both README paragraphs now say agents and new projects get temporary keys with no account, and a signed-out user in an existing project is asked to log in unless they pass `--accountless`.
- Used `--accountless` instead of `--keyless` since the CLI warns that `--keyless` is deprecated.
- Replaced `.env.local` with "your env file (the first `.env*` file that exists, or `.env.local` if none does)" in the README and in the three doc comments, since they had the same claim.
- "creates a Clerk application" is now "creates or links."
- Applied your changeset suggestion as-is.

Can you give a final review?

manovotny:
Pushed a follow-up in 4808d12.

- Reverted the env-file note in the three doc comments to `.env.local`. The accountless path (`writeKeysToEnvFile` → `detectEnvFile`) writes the framework default for Next.js, which is `.env.local`, even when a `.env` already exists — so "the first `.env*` file that exists" was describing the authenticated path, not the one these sentences are about.
- Dropped the same parenthetical from the README's "Either way" sentence, since it spans both paths, and left it as "your env file."

manovotny (approved):

PR #9685, opened 2026-09-08
feat(ui): add SSO fallback sign-in flow to `<SignIn />`

mauricioabreu:
## Description

Enterprise-routed sign-ins for users the instance has allowlisted for an SSO fallback no longer redirect straight to the identity provider. `<SignIn />` stops at a screen offering SSO — or the existing connection picker, when several connections serve the address — alongside a "Can't use SSO?" link that leads to the standard email code step, which carries a notice that the organization requires single sign-on and that the attempt is recorded.

The link renders whenever `sso_fallback_first_factors` is present, which the API returns only for allowlisted users. Users without a fallback, and instances without the feature, are unaffected.

Also adds `ssoFallbackFirstFactors` to the sign-in resource for custom flows.

Fixes ORGS-1825.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/packages/shared/src/types/signIn.ts
+++ b/packages/shared/src/types/signIn.ts
@@ -51,6 +51,16 @@ export interface SignInResource extends ClerkResource {
   supportedIdentifiers: SignInIdentifier[];
   supportedFirstFactors: SignInFirstFactor[] | null;
   supportedSecondFactors: SignInSecondFactor[] | null;
+  /**
+   * Factors an enterprise-routed user can fall back to when they cannot reach their
+   * identity provider. Kept out of `supportedFirstFactors` so consumers that assert
+   * that list is purely enterprise keep working.
+   *
+   * Its presence says the instance allows a fallback, never that this particular
+   * user may use one — the API does not disclose that, and the UI must not branch
+   * on it.
+   */
+  ssoFallbackFirstFactors: SignInFirstFactor[] | null;
   clientTrustState?: ClientTrustState;
   firstFactorVerification: VerificationResource;
   secondFactorVerification: VerificationResource;
@@ -130,6 +140,11 @@ export interface SignInJSON extends ClerkResourceJSON {
   user_data: UserDataJSON;
   supported_first_factors: SignInFirstFactorJSON[];
   supported_second_factors: SignInSecondFactorJSON[];
+  /**
+   * Present only when the instance has opted into SSO fallback and the sign-in is
+   * still owed its first factor. Absent everywhere else.
+   */
+  sso_fallback_first_factors?: SignInFirstFactorJSON[];
   first_factor_verification: VerificationJSON | null;
   second_factor_verification: VerificationJSON | null;
   created_session_id: string | null;
--- a/packages/shared/src/types/signIn.ts
+++ b/packages/shared/src/types/signIn.ts
@@ -51,15 +51,6 @@ export interface SignInResource extends ClerkResource {
   supportedIdentifiers: SignInIdentifier[];
   supportedFirstFactors: SignInFirstFactor[] | null;
   supportedSecondFactors: SignInSecondFactor[] | null;
-  /**
-   * Factors an enterprise-routed user can fall back to when they cannot reach their
-   * identity provider. Kept out of `supportedFirstFactors` so consumers that assert
-   * that list is purely enterprise keep working.
-   *
-   * The API returns this only for users the instance has allowlisted for a fallback,
-   * so its presence is itself the eligibility signal. Anything rendered from it is
-   * visible to whoever supplied the identifier.
-   */
   ssoFallbackFirstFactors: SignInFirstFactor[] | null;
   clientTrustState?: ClientTrustState;
   firstFactorVerification: VerificationResource;
@@ -140,10 +131,6 @@ export interface SignInJSON extends ClerkResourceJSON {
   user_data: UserDataJSON;
   supported_first_factors: SignInFirstFactorJSON[];
   supported_second_factors: SignInSecondFactorJSON[];
-  /**
-   * Present only when the instance has opted into SSO fallback and the sign-in is
-   * still owed its first factor. Absent everywhere else.
-   */
   sso_fallback_first_factors?: SignInFirstFactorJSON[];
   first_factor_verification: VerificationJSON | null;
   second_factor_verification: VerificationJSON | null;
--- a/packages/shared/src/types/signIn.ts
+++ b/packages/shared/src/types/signIn.ts
@@ -56,9 +56,9 @@ export interface SignInResource extends ClerkResource {
    * identity provider. Kept out of `supportedFirstFactors` so consumers that assert
    * that list is purely enterprise keep working.
    *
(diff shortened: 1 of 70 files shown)


packages/ui/src/components/SignIn/__tests__/SignInStart.test.tsx
```diff
@@ -983,6 +1003,38 @@ describe('SignInStart', () => {
...
+      } as unknown as SignInResource);
+
+      Object.defineProperty(window, 'location', {
+        writable: true,
+        value: { href: 'http://localhost/sign-in?__clerk_ticket=test_ticket' },
+      });
+      Object.defineProperty(window, 'history', {
+        writable: true,
+        value: { replaceState: vi.fn() },
+      });
```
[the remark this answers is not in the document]
mauricioabreu: already fixed.

packages/shared/src/types/signIn.ts
```diff
@@ -51,6 +51,16 @@ export interface SignInResource extends ClerkResource {
   supportedIdentifiers: SignInIdentifier[];
   supportedFirstFactors: SignInFirstFactor[] | null;
   supportedSecondFactors: SignInSecondFactor[] | null;
+  /**
```
dstaley: lets remove this comment. the only fields in this type that are annotated were done so by AI, so I don't think we need them.

packages/shared/src/types/signIn.ts
```diff
@@ -130,6 +140,11 @@ export interface SignInJSON extends ClerkResourceJSON {
   user_data: UserDataJSON;
   supported_first_factors: SignInFirstFactorJSON[];
   supported_second_factors: SignInSecondFactorJSON[];
+  /**
```
dstaley: same as above: let's remove this for now since nothing else is annotated

packages/shared/src/types/signInFuture.ts
```diff
@@ -338,6 +338,11 @@ export interface SignInFutureResource {
    */
   readonly supportedSecondFactors: SignInSecondFactor[];
 
+  /**
+   * Factors an enterprise-routed user can fall back to when they cannot reach their identity provider. Empty unless the instance has allowlisted this user for a fallback, so a non-empty list is itself the eligibility signal.
```
dstaley:
```suggestion
   * Factors an enterprise SSO user can fall back to when they cannot reach their identity provider. Empty unless the instance has allowlisted this user for a fallback.
```

packages/ui/src/components/SignIn/shared.ts
```diff
@@ -108,4 +108,19 @@ function hasMultipleEnterpriseConnections(
   );
 }
 
-export { hasMultipleEnterpriseConnections, useHandleAuthenticateWithPasskey };
+/**
+ * Returns the email code factor an enterprise-routed sign-in may fall back to, or `null`.
```
dstaley:
```suggestion
 * Returns the email code factor a sign-in may fall back to, or `null`.
```

packages/ui/src/components/SignIn/shared.ts
```diff
@@ -108,4 +108,19 @@ function hasMultipleEnterpriseConnections(
...
+ *
+ * The API populates the list only for allowlisted users, so a factor here means this user is
+ * eligible and anything rendered from it discloses that to whoever supplied the identifier.
```
dstaley: not sure we need this part of the comment

packages/ui/src/components/SignIn/SignInFactorOne.tsx
```diff
@@ -109,6 +110,10 @@ function SignInFactorOneInternal(): JSX.Element {
     supportedFirstFactors,
   });
 
+  // Frozen on mount: a later response may drop the field, which would otherwise unmount the
```
dstaley:
```suggestion
  // Frozen on mount: client-piggybacking may drop the field, which would otherwise unmount the
```

packages/ui/src/components/SignIn/SignInFactorOne.tsx
```diff
@@ -109,6 +110,10 @@ function SignInFactorOneInternal(): JSX.Element {
...
+  // fallback screens mid-flow.
+  const ssoFallbackFactor = React.useRef(getSSOFallbackFactor(signIn)).current;
```
dstaley, line 115: this makes me very nervous, but I think it's probably fine? client piggybacking makes stuff like this challenging. does the API absolutely have to drop the fallback factors during the lifecycle?

packages/ui/src/components/SignIn/SignInFactorOne.tsx
```diff
@@ -162,6 +167,15 @@ function SignInFactorOneInternal(): JSX.Element {
     }));
   };
 
+  /**
+   * An enterprise-routed sign-in on an instance that allows a fallback owns its own screens,
+   * including the enterprise connection choice, so that the fallback stays reachable from them.
+   * @experimental
+   */
```
dstaley: can delete this

packages/ui/src/components/SignIn/SignInFactorOneSSOFallback.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const { fallbackFactor } = props;
+  const card = useCardState();
+  const ctx = useSignInContext();
+  const signIn = useCoreSignIn();
+  const [step, setStep] = React.useState<Step>('sso');
+  const [isRedirecting, setIsRedirecting] = React.useState(false);
+
+  // `safe_identifier` is the address the user typed, unmasked; the design shows it obfuscated.
+  const maskedFallbackFactor = React.useMemo(
+    () => ({ ...fallbackFactor, safeIdentifier: maskEmailAddress(fallbackFactor.safeIdentifier) as string }),
```
dstaley: if the user already typed the identifier, then they know what it is. We don't need to mask it.

packages/ui/src/components/SignIn/SignInStart.tsx
```diff
@@ -241,7 +242,12 @@ function SignInStartInternal(): JSX.Element {
         }
         switch (res.status) {
           case 'needs_first_factor': {
-            if (!hasOnlyEnterpriseSSOFirstFactors(res) || hasMultipleEnterpriseConnections(res.supportedFirstFactors)) {
+            if (
+              !hasOnlyEnterpriseSSOFirstFactors(res) ||
+              hasMultipleEnterpriseConnections(res.supportedFirstFactors) ||
+              // The fallback is only reachable from a screen, so stop short of the automatic redirect.
+              !!getSSOFallbackFactor(res)
```
dstaley: let's hoist this conditional up into something that describes the condition in which it's true

dstaley (approved):
approving since all the suggestions seem minor. Overall this feels very straight forward! the only thing that's somewhat weird is how the fallback factors interact with client-piggybacking, which I assume is why we're capturing the first value of it in a useRef call.

PR #9696, opened 2026-09-09
fix(ui): merge container styles into Mosaic Icon via xstyle

alexcarpenter:
## Description

Containers that sized a Mosaic `Icon` (the `Banner` and `Field.Error` icons at `1lh`, and the user-profile security, provider, and search icons) spread their own `stylex.props(...)` output onto the icon. StyleX only resolves conflicting properties within a single `stylex.props` call, so the container's `height` and the icon's own `sizes[size]` atom ended up as two equal-priority classes and stylesheet order decided which won. In practice the icon's size atom won and the `1lh` height was dropped.

`Icon` now accepts an `xstyle` prop that is merged last in its own `stylex.props` call, so the container's atoms deterministically override the icon's defaults. All call sites that previously spread resolved atoms onto an `Icon` use `xstyle` instead. The banner icon also gets the `cl-banner-icon` slot class.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-09-09 at 3 44 42 PM | Screenshot 2026-09-09 at 3 45 10 PM | 

## Checklist

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

### DIFF
--- a/.changeset/mosaic-icon-xstyle.md
+++ b/.changeset/mosaic-icon-xstyle.md
@@ -1,5 +1,2 @@
 ---
-'@clerk/ui': patch
 ---
-
-Fix the Mosaic `Banner` and `Field.Error` icons so their line-height-based sizing reliably overrides the icon's default size, and expose the banner icon as `.cl-banner-icon`.
--- a/.changeset/mosaic-icon-xstyle.md
+++ b/.changeset/mosaic-icon-xstyle.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Fix the Mosaic `Banner` and `Field.Error` icons so their line-height-based sizing reliably overrides the icon's default size, and expose the banner icon as `.cl-banner-icon`.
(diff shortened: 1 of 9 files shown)


.changeset/mosaic-icon-xstyle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Fix the Mosaic `Banner` and `Field.Error` icons so their line-height-based sizing reliably overrides the icon's default size, and expose the banner icon as `.cl-banner-icon`.
```
maxyinger: does this need a changeset?

maxyinger (approved):

.changeset/mosaic-icon-xstyle.md
alexcarpenter:
```suggestion
---
---
```

PR #9701, opened 2026-09-10
fix(repo): flatten Pick parameters in TypeDoc output

SarahSoutoul:
## Description

`OrganizationResource.createDomain()` accepts the domain name as its first argument and only `enrollmentMode` in its optional `params` object:

```ts
createDomain(domainName: string, params?: Pick<CreateOrganizationDomainParams, 'enrollmentMode'>)
```

The TypeDoc parameter-table renderer understands inline objects, interfaces, and type aliases, but it did not resolve TypeScript's built-in `Pick<T, K>`. As a result, the generated table documented only the opaque `params?: Pick<...>` parameter and did not surface `params?.enrollmentMode?`, its allowed values, description, or default.

The current docs work around that limitation by embedding the complete `CreateOrganizationDomainParams` declaration below the method. That declaration contains both `name` and `enrollmentMode`, even though this method receives `name` separately as `domainName`. Once `enrollmentMode` is flattened into the method table, retaining the workaround would also document it twice.

This PR teaches the parameter-table renderer to resolve the built-in `Pick<T, K>` when `K` is composed of string literal keys. It filters the source declaration to exactly the selected properties and passes that shape through the existing nested-parameter renderer. The parent parameter remains in the table, followed by rows for the selected properties with their original types and JSDoc.

Because the selected shape is documented inline, the source type is rendered without a link to the full unpicked declaration. Unsupported or partially resolved `Pick` shapes retain the existing output rather than producing incomplete documentation.

Adds an end-to-end snapshot for `organization.createDomain()`, which now documents `params?.enrollmentMode?` without leaking the omitted `name` property.

Although the repository contains other source-level `Pick` usages, `createDomain()` is currently the only generated standalone parameter table that enters this code path. Tests, internal types, packages outside the TypeDoc entry points, property tables, and function-valued resource-summary properties use different paths and remain unchanged.

Required merge flow:

1. Merge this PR.
2. Create the generated TypeDoc sync PR in `clerk/clerk`.
3. Merge the companion cleanup in [clerk/clerk#3341](https://github.com/clerk/clerk/pull/3341) into the generated TypeDoc PR.
4. Merge the generated TypeDoc PR.

The Clerk cleanup should not merge directly into `main`; combining it with the generated output prevents an intermediate state with missing or duplicated parameter documentation.

Resolves [DOCS-11879](https://linear.app/clerk/issue/DOCS-11879/flatten-parameters-containing-pick-in-typedoc-generated-output).

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


manovotny:
Pushed some changes directly in c0af671c7b.

- Extended `custom-theme.test.mts` with the multi-key `Pick` case and the fail-closed branches (missing keys, non-literal keys, `Omit`), plus unit coverage for `getPickPropertyNames`
- Exported `getPickPropertyNames` so it can be tested directly

Leaving a comment for one more thing.

.typedoc/custom-theme.mjs
```diff
@@ -470,6 +514,37 @@ function getParameterObjectShapeDeclaration(t) {
...
+        (sourceRef &&
+        (sourceRef.kind === ReflectionKind.TypeAlias || sourceRef.kind === ReflectionKind.Interface) &&
+        'children' in sourceRef
+          ? /** @type {import('typedoc').DeclarationReflection} */ (sourceRef)
+          : undefined);
+      if (!sourceWithChildren?.children?.length) {
+        return undefined;
+      }
+      const selected = new Set(propertyNames);
+      const children = sourceWithChildren.children.filter(child => selected.has(child.name));
```
manovotny, line 538: For a generic source like `Pick<Box<string>, 'value'>`, this reuses `Box<T>`'s child reflections, so the flattened row documents the property's type as `T` rather than `string` — TypeDoc keeps the instantiation in the use-site `typeArguments` and doesn't rewrite the referenced declaration's children.

Nothing in the entry points hits this today (`CreateOrganizationDomainParams` isn't generic), and it's a pre-existing trait of the reference-flatten path, not something this PR introduces. The one wrinkle: this branch fails open (a wrong type) rather than closed.

Worth bailing to the opaque output when the source is a generic instantiation — returning `undefined` here when `sourceType.typeArguments?.length` — so it degrades instead of documenting `T`?

SarahSoutoul:
> Pushed some changes directly in [c0af671](https://github.com/clerk/javascript/commit/c0af671c7b64df12713e593ac7ed5c3ab4f23eec).
> 
> * Extended `custom-theme.test.mts` with the multi-key `Pick` case and the fail-closed branches (missing keys, non-literal keys, `Omit`), plus unit coverage for `getPickPropertyNames`
> * Exported `getPickPropertyNames` so it can be tested directly
> 
> Leaving a comment for one more thing.

Thanks Michael! Not seeing the comment you're referring to? @manovotny

.typedoc/custom-theme.mjs
SarahSoutoul: Good catch -`Pick<Box<string>, 'value'>` could otherwise flatten the unspecialized declaration and surface T instead of string. I added a fail-closed guard for generic source instantiations and regression coverage in [e07caea](https://github.com/clerk/javascript/commit/e07caea10a6b2c032fe6b566adae3ba172888d1c).

manovotny (approved):

PR #9702, opened 2026-09-10
feat(headless): add combobox primitive

austincalvelage:
## Description

https://swingset-git-austin-headless-combobox.clerkstage.dev/primitives/combobox

@maxyinger made a fair point about [distinguishing Combobox from Autocomplete](https://github.com/clerk/javascript/pull/9655#discussion_r3938317796). This adds a Combobox primitive for selecting from a searchable list, remembering the selected option separately from the search text.

I think we can share some common logic between the two as this is heavily based autocomplete, but wanted to keep this PR focused.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/headless/src/primitives/combobox/combobox-positioner.tsx
```diff
@@ -0,0 +1,79 @@
...
+      if (!anchor) {
+        return;
+      }
+      refs.setPositionReference(anchor);
+      return () => refs.setPositionReference(refs.domReference.current);
+    }, [anchor, refs]);
+
+    const side = placement.split('-')[0];
+
+    const floatingProps = getFloatingProps();
```
maxyinger: I think floating props might need to get moved into the popup

🤖 
```
1. role="listbox" doesn't own its options — packages/headless/src/primitives/combobox/combobox-positioner.tsx:36

Actual DOM when open:
<div id=":r1:" role="listbox" data-side="bottom">   <!-- Positioner -->
  <div data-open="">                                 <!-- Popup: generic, exposed -->
    <div style="display: contents;">                 <!-- Freeze: not exposed, fine -->
      <div role="option">Apple</div>
The Popup div is a generic node between listbox and option, which breaks ARIA required-owned-elements (and fails axe aria-required-children). Inline mode is fine — List wraps options directly.
```

maxyinger (approved):
shape looks good to me

*edgecases from 🤖 , maybe just copy and paste into your agent to validate
```
1. role="listbox" doesn't own its options — packages/headless/src/primitives/combobox/combobox-positioner.tsx:36

Actual DOM when open:
<div id=":r1:" role="listbox" data-side="bottom">   <!-- Positioner -->
  <div data-open="">                                 <!-- Popup: generic, exposed -->
    <div style="display: contents;">                 <!-- Freeze: not exposed, fine -->
      <div role="option">Apple</div>
The Popup div is a generic node between listbox and option, which breaks ARIA required-owned-elements (and fails axe aria-required-children). Inline mode is fine — List wraps options directly.

Fix: floating-ui explicitly supports the positioner/popup split. Spread getFloatingProps() on Popup and leave only ref={refs.setFloating} + style={floatingStyles} + data-side on Positioner. getFloatingFocusElement() finds the inner element via data-floating-ui-focusable and useRole resolves aria-controls to the Popup's id automatically, so the pairing keeps working and the size() middleware's data-side check is unaffected.

2. Stale aria-activedescendant when nothing matches — combobox-root.tsx:210 (handleInputChange unconditionally does setActiveIndex(0))

Typing a non-matching query leaves:
value="zzzz" aria-expanded="true" aria-activedescendant=":r4:"   <!-- :r4: no longer in the DOM -->
APG requires activedescendant to reference the focused popup element. Needs to clear when the filtered list is empty.

3. empty renders inside the listbox — combobox-collection.tsx:19

<div role="listbox"><div>No fruit found</div></div> — non-option content in a listbox, same required-children rule, and aria-expanded="true" on a popup with no options. Should render outside the listbox (or close the popup / a role="status" region).

Nits

- No autoComplete="off" on the input — browser autofill UI can cover the popup.
```

PR #9706, opened 2026-09-10
feat(ui): add profile and routing props to the Mosaic UserButton

alexcarpenter:
## Description

Adds the props the Mosaic `<UserButton />` was still missing against the legacy button, so an app can migrate without losing configuration of the modals it opens:

- `signInUrl`: where adding an account goes, and the base a pending session task is routed under. Falls back to the instance sign-in URL.
- `afterCreateOrganizationUrl` and `skipInvitationScreen`: forwarded to the create-organization modal. The URL accepts a `:slug`/`:id` template or a builder, resolved by clerk-js against the new organization.
- `afterLeaveOrganizationUrl`: forwarded to the organization profile modal.
- `userProfileProps.additionalOAuthScopes`, `userProfileProps.apiKeysProps`, `userProfileProps.appearance`: forwarded to the user profile modal.
- `organizationProfileProps` with `customPages`, `pageOrder`, and `appearance`: the organization profile gets the same custom-page bridge the user profile already had. Its built-in page ids (`general`, `members`, `billing`, `apiKeys`, `security`) are mirrored from clerk-js behind the same feature guards so ordering against them works.

`appearance` is deliberately not a top-level `UserButton` prop; it only themes the profile modals. `defaultOpen` and `customMenuItems.open` from the original TODO are left out.

The pass-through props reuse the legacy types from `@clerk/shared/types` via `Pick`, so their types and docs have one source. Only `customPages` and `pageOrder` are Mosaic-owned.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


Ephem (approved):
Looks good to me, no notes.

PR #9718, opened 2026-09-11
feat(shared,js): add Google Workspace credentials and sync to DirectorySync

gabrielmeloc22:
## Description

Directory Sync shipped assuming the push model: the identity provider holds a bearer token and pushes SCIM to Clerk. Google Workspace works the other way round. It authenticates with a service account credential that Clerk stores, and Clerk pulls from it on a schedule. None of that is expressible through the current `DirectorySync` resource, so the component sends Google connections to the Clerk Dashboard, which the organization admin reading that message has no account for.

This adds the resource surface for the pull shape. The FAPI endpoints it calls land in clerk/clerk_go#22044 (credentials) and clerk/clerk_go#22045 (sync, sync status); both are still open, so this must not ship ahead of them.

- `setCredentials({ serviceAccountJson, subjectEmail })` stores the credential and activates the directory. Calling it again replaces the stored credential, which is how a rotated key is applied.
- `sync()` starts a sync instead of waiting for the next scheduled one.
- `getSyncStatus()` returns the last sync result, all fields `null` before the first sync completes.
- `credentialsConfigured` reports whether a credential is stored. It is `null` for push providers, which have no credential rather than an unconfigured one.

The uploaded key is an input only. It is never assigned to the resource and never reachable from `__internal_toSnapshot()`, which matters because snapshots can be persisted. There is a test for that, and it fails if the key is ever retained.

One thing worth knowing when wiring the UI: the `400` from `setCredentials` carries Google's own validation message, such as a missing domain-wide delegation or a rejected admin email. Surface it verbatim. It is the only thing telling the admin what is wrong with their Workspace setup.

Part of ORGS-1842

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/clerk-js/src/core/resources/__tests__/DirectorySync.test.ts
```diff
@@ -69,6 +69,102 @@ describe('DirectorySync', () => {
...
+    BaseResource._fetch = vi.fn().mockReturnValue(
+      Promise.resolve({
+        response: { ...directoryJSON, provider: 'google', enabled: true, credentials_configured: true },
+      }),
+    );
+
+    const result = await createDirectorySync().setCredentials({
+      serviceAccountJson: '{"type":"service_account"}',
+      subjectEmail: 'admin@example.com',
+    });
```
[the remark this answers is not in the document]
gabrielmeloc22: :white_check_mark: Added in 427840d. Asserts a rejected upload reaches the caller with the message intact, using the real domain-wide-delegation wording the gateway produces.

packages/clerk-js/src/core/resources/DirectorySync.ts
```diff
@@ -83,6 +87,47 @@ export class DirectorySync extends BaseResource implements DirectorySyncResource
     return new DeletedObject(json);
   };
 
+  setCredentials = async (params: SetDirectorySyncCredentialsParams): Promise<DirectorySyncResource> => {
+    const json = (
+      await BaseResource._fetch<DirectorySyncJSON>({
+        path: `${this.directoryPath}/credentials`,
```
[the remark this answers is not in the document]
gabrielmeloc22: :no_entry: These routes do exist. `POST .../directory/credentials` is added in clerk/clerk_go#22044 and `POST .../directory/sync` plus `GET .../directory/sync_status` in clerk/clerk_go#22045, both on the same org-scoped group as the CRUD and `users` routes you found, inside the same self-serve gate chain.

:bulb: The sequencing point underneath it is fair though: those two PRs are still open, so until they merge and deploy these methods would 404. Noted on the PR description, which claimed the endpoints were already live.

packages/clerk-js/src/core/resources/DirectorySync.ts
```diff
@@ -83,6 +87,47 @@ export class DirectorySync extends BaseResource implements DirectorySyncResource
...
+        method: 'POST',
+        body: {
+          service_account_json: params.serviceAccountJson,
+          subject_email: params.subjectEmail,
+        } as any,
+      })
+    )?.response as unknown as DirectorySyncJSON;
+
+    // The credential is deliberately not kept on the resource. It is an input
+    // only; the server stores it and reports `credentials_configured` back.
```
dstaley: we don't need this comment

packages/clerk-js/src/core/resources/DirectorySync.ts
```diff
@@ -83,6 +87,47 @@ export class DirectorySync extends BaseResource implements DirectorySyncResource
...
+  sync = async (): Promise<void> => {
+    await BaseResource._fetch({
+      path: `${this.directoryPath}/sync`,
+      method: 'POST',
+    });
+  };
+
+  getSyncStatus = async (): Promise<DirectorySyncStatusResource> => {
+    // Not a Clerk resource — it has no id or object — so it is fetched
+    // untyped and cast, the same way getUsers handles its paginated payload.
```
dstaley: can remove

dstaley (approved):

packages/clerk-js/sandbox/template.html
```diff
@@ -449,7 +449,7 @@
     <script
       type="text/javascript"
       src="/<%= htmlRspackPlugin.files.js[0] %>"
-      data-clerk-publishable-key="pk_test_c3VtbWFyeS1tYW4tOTkuY2xlcmsuYWNjb3VudHNzdGFnZS5kZXYk"
+      data-clerk-publishable-key="pk_test_bm90YWJsZS1tdXR0LTk4NC5jbGVyay5hY2NvdW50cy5sY2xjbGVyay5jb20k"
```
alexcarpenter: revert before merging?

gabrielmeloc22: oh, nice catch, i think i've committed it by accident. will do!

PR #9722, opened 2026-09-11
feat(ui): set up Google Workspace directories in the sync wizard

gabrielmeloc22:
## Description

Stacked on #9720 — review that first.

**Merge order:** this stack calls FAPI endpoints added in clerk/clerk_go#22044 (credentials) and clerk/clerk_go#22045 (sync, sync status). Both are still open, so this must not ship ahead of them.

An organization admin who finishes Google SAML SSO and moves on to Directory Sync is told to configure it in the Clerk Dashboard. That is the Clerk customer's account rather than theirs, so the flow dead-ends for every Google connection. This is the change that removes the dead end.

Google directories are read by Clerk rather than pushed to, so their setup collects a credential instead of handing out an endpoint and a bearer token.

- Providers now carry a `push` or `pull` mode instead of a `supportsScim` flag, and the steps branch on that.
- The configure step renders a service account key upload and the admin email to read the directory as, with Google's own validation message shown verbatim when the credential is refused. That message is the only thing telling the admin what is wrong with their Workspace setup.
- The test step gains a sync trigger and the outcome of the last run. Without it a pull directory shows an empty user list for minutes, with no way to tell a slow sync from a broken one.
- The Directory Sync section on the Security page offers setup for Google like any other provider.

The uploaded key is read with `FileReader` rather than `Blob.text()`, which Safari only gained in 14. There is an existing precedent for this in `AvatarUploader`.

Localizations: the new strings are added to `en-US`, and the removed `warning__googleUnsupported` is stripped from all 47 other locales. Untranslated keys fall back to `en-US`, so the other locales are not otherwise touched.

## Test plan

- Two new wizard tests: a Google connection renders the credential form rather than the dead-end notice and exposes no endpoint or token, and submitting sends the key and admin email then stops holding the key.
- `OrganizationSecurityPage` had a test asserting the dead end. It is inverted rather than deleted, so the old behavior cannot come back unnoticed.
- 109 tests pass across `ConfigureDirectorySync` and `OrganizationProfile`.

Note on typecheck: `packages/ui` reports 184 errors, all in `mosaic/`. The same 184 are present on a clean tree, so none come from this change.

Part of ORGS-1843

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/components/ConfigureDirectorySync/GoogleCredentialsForm.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const { directory, setCredentials } = useConfigureDirectorySync();
+  const { t } = useLocalizations();
+
+  const [serviceAccountJson, setServiceAccountJson] = useState('');
+  const [fileName, setFileName] = useState<string | null>(null);
+  const [subjectEmail, setSubjectEmail] = useState('');
+  const [fileError, setFileError] = useState<string | null>(null);
+
+  const isConfigured = Boolean(directory?.credentialsConfigured);
+  const hasPendingCredential = Boolean(serviceAccountJson) && Boolean(subjectEmail);
```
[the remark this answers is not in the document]
gabrielmeloc22: :white_check_mark: Fixed in 52320b7. Trims and validates before enabling Continue, using the same check the package already uses in `InviteMembersForm` and `SignIn/utils`, and submits the trimmed value. Test covers whitespace, a malformed address, and a padded valid one.

packages/ui/src/components/ConfigureDirectorySync/GoogleCredentialsForm.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  const selectFile = useCallback(
+    async (file: File | undefined): Promise<void> => {
+      if (!file) {
+        return;
+      }
+      const contents = await fileToText(file);
```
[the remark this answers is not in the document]
gabrielmeloc22: :white_check_mark: Fixed in 0eaff7c and 52320b7. The read is now caught and surfaced, and the pending key is cleared before reading rather than only on failure, so a failed read cannot leave the previous one staged.

The reject also carried a `ProgressEvent` rather than an `Error`, which is what `prefer-promise-reject-errors` was warning about; that is an `Error` now.

packages/ui/src/components/ConfigureDirectorySync/SyncNowRow.tsx
```diff
@@ -0,0 +1,131 @@
...
+    }
+    setError(undefined);
+    setIsSyncing(true);
+    try {
+      await onSync();
+      onSynced();
+    } catch (err) {
+      // Includes the already-running case, which is a normal thing to hit by
+      // clicking twice rather than a failure worth alarming about.
+      handleError(err as Error, [], message => setError(typeof message === 'string' ? message : undefined));
```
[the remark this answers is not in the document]
gabrielmeloc22: :white_check_mark: Fixed in 52320b7. Confirmed the mechanism: `handleError` rethrows anything `isKnownError` does not match, and the click discards it, so a network failure left the button idle as though the sync had run. There is now a fallback message for that case.

packages/ui/src/components/ConfigureDirectorySync/steps/ConfigureStep.tsx
```diff
@@ -37,14 +38,17 @@ const FieldLabel = ({ id, localizationKey }: { id: string; localizationKey: Loca
...
   const card = useCardState();
   const [isInstructionsOpen, setIsInstructionsOpen] = useState(false);
 
-  const isGoogle = provider === 'google';
-  const canProvision = Boolean(connection) && !isGoogle;
+  // Pull providers hand Clerk a credential instead of receiving a token, so the
+  // directory still has to exist first — it is what the credential attaches to.
+  const isPull = providerMeta?.mode === 'pull';
+  const credentials = useGoogleCredentialsState();
+  const canProvision = Boolean(connection);
```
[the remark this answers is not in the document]
gabrielmeloc22: :no_entry: The FAPI routes exist. `POST .../directory/credentials` is added in clerk/clerk_go#22044, and `POST .../directory/sync` plus `GET .../directory/sync_status` in clerk/clerk_go#22045, on the same org-scoped group and inside the same self-serve gate chain as the CRUD routes.

On the gating point: there is no Google gate left in `SecurityDirectorySyncSection` either. Removing it from both hosts is the change, not an oversight. Google was previously sent to the Clerk Dashboard from both places, which the organization admin reading that message has no account for.

:bulb: The sequencing concern underneath this is real though, and worth stating plainly: those two clerk_go PRs are open, so this stack must not merge ahead of them. Called out on the PR description.

packages/ui/src/components/ConfigureDirectorySync/steps/ConfigureStep.tsx
```diff
@@ -343,8 +341,11 @@ export const ConfigureStep = (): JSX.Element => {
 
       <Step.Footer>
         <Step.Footer.Continue
-          onClick={() => goNext()}
-          isDisabled={!directory}
+          onClick={isPull ? () => run(async () => (await credentials.submit(), goNext())) : () => goNext()}
```
dstaley: let's pull this out into a separately defined function, the one liner is difficult to read

packages/ui/src/components/ConfigureDirectorySync/GoogleCredentialsForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+  Input,
+  localizationKeys,
+  Text,
+  useLocalizations,
+} from '@/customizables';
+
+import { useConfigureDirectorySync } from './ConfigureDirectorySyncContext';
+
+// FileReader rather than Blob.text(), which Safari only gained in 14 and which
+// clerk-js cannot assume. Mirrors fileToBase64 in AvatarUploader.
```
dstaley: if this is some logic we use elsewhere, let's make it a shared function rather than embedding it within the component like this

packages/ui/src/components/ConfigureDirectorySync/GoogleCredentialsForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+  new Promise((resolve, reject) => {
+    const reader = new FileReader();
+    reader.readAsText(file);
+    reader.onload = () => resolve(reader.result as string);
+    // onerror hands back a ProgressEvent, which is useless to a caller.
+    reader.onerror = () => reject(new Error('Could not read the selected file'));
+  });
+
+// Same shape as the copies in InviteMembersForm and SignIn/utils: enough to
+// stop an obviously malformed address reaching the provider, no more.
```
dstaley: if that's the case, let's share it

packages/ui/src/components/ConfigureDirectorySync/GoogleCredentialsForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+  const [subjectEmail, setSubjectEmail] = useState('');
+  const [fileError, setFileError] = useState<string | null>(null);
+
+  const isConfigured = Boolean(directory?.credentialsConfigured);
+  const trimmedSubjectEmail = subjectEmail.trim();
+  // Continue lives outside the form, so the input's type='email' never triggers
+  // native validation. Without this, whitespace alone enables it.
+  const hasPendingCredential = Boolean(serviceAccountJson) && isEmail(trimmedSubjectEmail);
+
+  const selectFile = useCallback(
```
dstaley, line 54: do we think uploading the file is the best UX versus copy/pasting?

packages/ui/src/components/ConfigureDirectorySync/GoogleCredentialsForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+    [t],
+  );
+
+  const submit = useCallback(async (): Promise<void> => {
+    if (!hasPendingCredential) {
+      return;
+    }
+    await setCredentials({ serviceAccountJson, subjectEmail: trimmedSubjectEmail });
+    // Drop the key as soon as it has been accepted. Nothing in this flow needs
+    // it again, and holding it only widens where it can leak from.
```
dstaley: don't need this

dstaley (changes requested):

packages/ui/src/components/ConfigureDirectorySync/GoogleCredentialsForm.tsx
gabrielmeloc22: i think so! when generating the key, google lets you download but doesn't give you a handle to copy it, i think probably because it's not that human-readable anyway.

packages/ui/src/components/ConfigureDirectorySync/steps/ConfigureStep.tsx
```diff
@@ -343,8 +346,11 @@ export const ConfigureStep = (): JSX.Element => {
 
       <Step.Footer>
         <Step.Footer.Continue
-          onClick={() => goNext()}
-          isDisabled={!directory}
+          onClick={isPull ? () => void run(submitCredentialsAndContinue) : () => goNext()}
```
dstaley: could we instead do something like

```
const submitCredentialsAndContinue = (): void => {
  if (isPull) {
    void run(submitCredentialsAndContinue);
    return;
  }

  goNext();
};
```

that way we don't have to mentally parse a ternary, and we have a stable identity

dstaley (approved):
approving, we don't need to fix the readability now, but feel free to do so if you think it makes it more readable!

PR #9734, opened 2026-09-11
feat(ui): add Confirmation block

alexcarpenter:
## Description

Adds a `Confirmation` block to Mosaic at `packages/ui/src/mosaic/blocks/confirmation`, alongside `Destructive`.

`Destructive` is a type-to-confirm dialog: the action stays inert until the user types a phrase back. Many destructive actions warrant a second look but not that ceremony, such as removing a connected account, revoking a session, or signing out everywhere. `Confirmation` is the dialog for those: a title, a description, Cancel, and a negative action. A failed attempt renders as a negative `Banner` between the description and the actions. The description takes markup, so a name can be emphasised.

The block has two forms, discriminated on the `handle` prop.

- **Controlled.** The same surface as `Destructive`: `open`, `isConfirming`, and `errorMessage` are owned by the caller and the block holds nothing of its own.
- **With a handle.** For a table where every row can remove something but only one dialog should exist. Create a handle with `Confirmation.createHandle<Payload>()`, mount the block once, and open it from any row with `handle.open(payload)`. `title`, `description`, and `actionLabel` accept a function of the payload, and `onConfirm` receives it. A small machine (`confirmation.controller.ts`) owns open, pending, and error: a resolved promise closes the dialog, a rejected one keeps it open showing why, and the next open starts clean. Opened from a `Menu.Item`, focus returns to the menu's trigger on close.

This is a separate block rather than an optional-field mode on `Destructive` so each stays targeted and neither grows conditional branches.

Also adds the swingset page (`/components/confirmation`) with `Default`, `WithError`, and `WithHandle` stories, marked `wip` while the block is a first draft.

Nothing consumes the block yet. It is not exported from a public entrypoint, so the changeset is empty.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


austincalvelage (approved):
Love these changes coming in this will really clean things up.

PR #9735, opened 2026-09-11
feat(shared,ui): add AgentID OAuth provider

wyattjoh:
## Description

Adds [AgentID](https://www.agentid.com) as a built-in social connection. `agentid` is now part of the `OAuthProvider` union (and therefore `OAuthStrategy` as `oauth_agentid`), and `OAUTH_PROVIDERS` carries its display data with the display name `AgentID` and a `docsUrl` pointing to Clerk's AgentID guide.

Because `<SignIn />` and `<SignUp />` only render social buttons for strategies present in `OAUTH_PROVIDERS`, this is what lets an instance with `oauth_agentid` enabled show a "Continue with AgentID" button. `<ProviderIcon />` treats the AgentID mark as a monochrome icon (mask-image, tinted with the foreground color) so it stays visible in dark mode, matching Apple, GitHub, Vercel, and X. AgentID publishes its mark as a single-fill monochrome SVG, so the mask-image treatment is the right one.

`docsUrl` points at Clerk's own AgentID guide (clerk/clerk-docs#3536). AgentID's partner-maintained page documents the **Custom provider** flow, which produces `oauth_custom_agentid` — the wrong instructions for someone sitting on the built-in AgentID connection page that this link is surfaced from. Note the URL uses the current docs path rather than the legacy `/docs/authentication/social-connections/<provider>` form used by the neighbouring entries, which only resolve via redirects.

Backend support and the Dashboard provider listing ship separately; the `agentid.svg` static asset is uploaded outside this repo.

## Checklist

- [x] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated — clerk/clerk-docs#3536

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/shared/src/types/runtime-values.ts
```diff
@@ -221,6 +221,18 @@ export const OAUTH_PROVIDERS: OAuthProviderData[] = [
     name: 'Hugging Face',
     docsUrl: 'https://clerk.com/docs/authentication/social-connections/huggingface',
   },
+  {
+    provider: 'vercel',
+    strategy: 'oauth_vercel',
+    name: 'Vercel',
+    docsUrl: 'https://clerk.com/docs/authentication/social-connections/vercel',
+  },
```
Zertsov, line 229: out of curiosity, is this needed for the agentID stuff? Or is it piggybacking here?

Zertsov (approved):
LGTM, but double check the docs link that Claude pointed out and I had one other question

packages/shared/src/types/runtime-values.ts
wyattjoh: piggybacking!

PR #9737, opened 2026-09-11
feat(ui): Add profile email verification and management flows

austincalvelage:
## Description

Adds email entry and OTP verification to Mosaic UserProfile, including resend, pending, and error states. Email addresses can be removed with confirmation or set as primary after verification.

Also adds separate email-link and enterprise SSO verification views. Email-link verification shows the destination address and resend countdown. Enterprise SSO shows the matching provider, domain, and Connect action. Both support pending and error states and cancellation. Profile email uses OTP; the views use supplied callbacks for account operations.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/user-profile/user-profile-account-section/user-profile-add-email.controller.ts
```diff
@@ -0,0 +1,159 @@
...
+        },
+      }),
+    },
+    verify: {
+      on: {
+        CANCEL: 'idle',
+        TICK: tick,
+        RESEND: {
+          target: 'resending',
+          guard: context => context.resendSeconds === 0,
```
dominic-clerk, line 91: **[MEDIUM] Resend cooldown is client-side only, and the backend has no email-code equivalent**

This `resendSeconds === 0` guard plus the 12s `setTimeout` TICK is the only thing throttling verification resends, and it runs entirely in the browser. On the backend, `EmailCodePreparer.Prepare` in `api/shared/strategies/email_code.go` has no rate limiter at all, unlike `PhoneCodePreparer` which calls `rateLimiter.Check(ctx, ratelimit.PhoneCodePreparationAttempt{...})`. That matters more than plain spam: the attempt cap is per-verification (`model/verification.go` fails at `Attempts > 2`), but `prepare_verification` repoints the identification at a brand new verification each time, so the 3-guess budget resets on every resend. Someone calling FAPI directly can brute force an unclaimed address in batches of 3 forever, and each request also mails the victim.

The gap is pre-existing in clerk_go rather than introduced here, but this PR is what makes the client countdown the email flow's resend throttle, so it should not be relied on as one. Worth adding a preparation rate limit in `EmailCodePreparer.Prepare` mirroring the phone path.

packages/ui/src/mosaic/user-profile/user-profile-account-section/user-profile-verify-email-link.dialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+                size='sm'
+                variant='link'
+                color='neutral'
+                disabled={isResending || resendSeconds > 0}
+                onClick={onResend}
+              >
+                {isResending
+                  ? m.resending
+                  : resendSeconds > 0
+                    ? fill(m.resendCountdown, { seconds: resendSeconds })
```
maxyinger, line 85: wonder if we can wrap numbers in a span here and use font-variant tabular nums to prevent the parenthesis shift on the count

packages/ui/src/mosaic/user-profile/user-profile-account-section/user-profile-remove-email.dialog.tsx
```diff
@@ -0,0 +1,50 @@
...
+      open={open}
+      onOpenChange={onOpenChange}
+    >
+      <Dialog.Popup>
+        <Dialog.Title render={<Heading size='sm' />}>{m.email.removeDialog.title}</Dialog.Title>
+        <Dialog.Description render={<Text />}>
+          {beforeEmail}
+          <strong {...stylex.props(styles.confirmationContactValue)}>{emailAddress}</strong>
+          {afterEmail}
+        </Dialog.Description>
```
maxyinger, line 37: seems to be evolving as a frequent use case. just mentioning to keep an eye on

packages/ui/src/mosaic/user-profile/user-profile-account-section/user-profile-account-section.view.tsx
```diff
@@ -100,6 +99,19 @@ export function UserProfileAccountSectionView({
...
+      allowMultipleAccounts={allowMultipleAccounts}
+      onAddEmail={onAddEmail}
+      onSendEmailCode={onSendEmailCode}
+      onVerifyEmailCode={onVerifyEmailCode}
+      onManageEmail={onManageEmail}
+      onVerifyEmail={onVerifyEmail}
+      onSetPrimaryEmail={onSetPrimaryEmail}
+      onRemoveEmail={onRemoveEmail}
+    />
+  );
```
maxyinger, line 114: depending on where we move the controller/model in the future, might make most of this self contained which would be nice. just `<UserProfileEmailRowView />`

packages/ui/src/mosaic/user-profile/user-profile-account-section/user-profile-account-section.view.tsx
```diff
@@ -100,6 +99,19 @@ export function UserProfileAccountSectionView({
       onRemovePhone={onRemovePhone}
     />
   );
+  const emailRow = (
+    <UserProfileEmailRowView
+      emails={emails}
+      allowMultipleAccounts={allowMultipleAccounts}
```
maxyinger, line 105: I wonder if there's a way to avoid branching props entirely here and still keep it dry.

I know we're doing this in the phone row too, but just feels off at the moment

packages/ui/src/mosaic/user-profile/__tests__/user-profile-add-email.dialog.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+    if (method === 'typing') {
+      await user.keyboard('12345');
+      expect(onSubmit).not.toHaveBeenCalled();
+      await user.keyboard('6');
+    } else {
+      await user.paste('123456');
+    }
+
+    expect(onSubmit).toHaveBeenCalledExactlyOnceWith('123456');
+  });
```
maxyinger, line 77: not needed atm, but this is like the same exact test as in phone. feels like it's indicating a seam

maxyinger (approved):
most of my feedback is a little more abstract and not super actionable. most are notes for later.

feel like there's a lot of room for unifying this and the phone row, but also feel like introducing model and full fledged control layers might change the shape of some of this a bit, so open till waiting until then to kinda bring these two together.

Looks like we have all the views we need in here tho

packages/ui/src/mosaic/features/user-profile/user-profile-account-section/user-profile-add-email.controller.ts
```diff
@@ -0,0 +1,159 @@
...
+        },
+      }),
+    },
+    verify: {
+      on: {
+        CANCEL: 'idle',
+        TICK: tick,
+        RESEND: {
+          target: 'resending',
+          guard: context => context.resendSeconds === 0,
```
austincalvelage: Noted here: https://linear.app/clerk/issue/CORE-3809/resend-cooldown-guard

PR #9743, opened 2026-09-12
feat(ui): move focus to the entering Flow step

alexcarpenter:
## Description

Adds `useFlowAutoFocus()` to the headless `Flow` primitive so a step can name the element that should receive focus when it enters, and wires it into every Reverification view.

Focusing an input as soon as a step mounts asks the browser to scroll the viewport to reveal an element that is still translated off-screen, which fights the slide animation. The Mosaic `Flow.Root` now uses `overflow: clip` instead of `hidden`, so the viewport is not a scroll container and cannot be scrolled by focus, `scrollIntoView`, or password managers. With that in place `Flow.Step` focuses the marked element as soon as it enters, using `preventScroll` so no other scroll ancestor moves either. The initially active step is left to the Dialog or Drawer that opened it; only a closed-to-open change moves focus, and only when focus is currently on the body or inside `Flow.Root`.

A view marks its primary control with the returned ref. When several mounted elements in one step are marked, the first in DOM order wins. Outside a `Flow.Step` the ref is a no-op, so views stay portable.

Mosaic `Otp` now forwards its ref to the first slot, following the `PhoneInput` convention, so the OTP view can target it.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.changeset/mosaic-flow-autofocus.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Reverification now moves focus to the entering step's primary control after the step transition completes, so the next input or action is ready for keyboard and screen reader users without disrupting the slide animation.
```
maxyinger: rm?

maxyinger: rm?

packages/headless/src/primitives/flow/flow-step.tsx
```diff
@@ -37,6 +57,33 @@ export const FlowStep = React.forwardRef<HTMLDivElement, FlowStepProps>(function
...
   }, [open, registerActiveStep, unregisterActiveStep]);
 
+  useEffect(() => {
+    const entering = open && !wasOpenRef.current;
+    wasOpenRef.current = open;
+    if (!entering) {
+      return;
+    }
+
+    return runOnEntered(() => {
```
maxyinger: ok dope. tested locally, using `clip` removes the need for waiting for the animation to end before focusing

packages/headless/src/primitives/flow/flow.test.tsx
```diff
@@ -233,4 +244,203 @@ describe('Flow', () => {
...
+      const { rerender } = render(<TestFlow value='password' />);
+
+      rerender(<TestFlow value='otp' />);
+      expect(screen.getByTestId('otp-input')).not.toHaveFocus();
+
+      act(() => flushRaf());
+      await act(async () => {});
+
+      expect(screen.getByTestId('otp-input')).toHaveFocus();
+    });
```
maxyinger: can alter if we remove post anim check

maxyinger (commented):
Love the hook. don't need the post animation check anymore with `clip` unless you like the look of it waiting to focus after the anim settles.

Think we should update the `Flow` story examples to use this too?

.changeset/mosaic-flow-autofocus.md
alexcarpenter: Emptied in 5dcca2f, matching the other Mosaic PRs.

packages/headless/src/primitives/flow/flow-step.tsx
alexcarpenter: Dropped the wait in f525994. The step now focuses as soon as it enters, with preventScroll, and the clip on the root keeps the viewport still.

packages/headless/src/primitives/flow/flow.test.tsx
alexcarpenter: Removed the getAnimations stubbing along with it in f525994. The timing tests became plain enter and reverse cases.

packages/headless/src/primitives/flow/flow-step.tsx
```diff
@@ -37,6 +57,33 @@ export const FlowStep = React.forwardRef<HTMLDivElement, FlowStepProps>(function
...
+      const target = firstInDocumentOrder(focusTargetsRef.current);
+      const root = rootRef.current;
+      if (target && root && focusIsWithin(root)) {
```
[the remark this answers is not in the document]
alexcarpenter: Added the root.contains(target) guard plus a portal test in 5dcca2f.

packages/ui/src/mosaic/components/otp/otp.tsx
```diff
@@ -12,20 +12,21 @@ import { styles } from './otp.styles';
 /** How the entered code currently reads back to the user. */
 export type OtpStatus = 'neutral' | 'success' | 'error';
 
-export interface OtpProps extends Omit<PrimitiveOtpProps, 'children' | 'length' | 'className' | 'style'> {
+export interface OtpProps extends Omit<PrimitiveOtpProps, 'children' | 'length' | 'className' | 'style' | 'ref'> {
```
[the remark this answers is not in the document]
alexcarpenter: Otp was a plain function component before this PR, so no ref ever reached the root. The inherited HTMLDivElement ref type was never wired to anything, and omitting it is what lets the forwarded input ref type-check. Mosaic components are not a public @clerk/ui entrypoint yet and Break Check reports no API changes on this PR.

packages/headless/src/primitives/flow/README.md
```diff
@@ -48,6 +48,23 @@ Multiple ids can select the same step. Moving between those ids updates the exis
 
 `Flow.Step` also accepts standard `<div>` attributes and the package's `render` prop.
 
+## Focus
+
+`useFlowAutoFocus()` returns a ref. Attach it to the element a step should focus when it enters:
```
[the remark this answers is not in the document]
alexcarpenter: Added the import to the example in 5dcca2f.

maxyinger (approved):

PR #9744, opened 2026-09-12
fix(ui): preserve identifier focus during sign-in submissions

nikosdouvlis:
## Description

When a sign-in request disables the focused phone or email input, iOS can close the keyboard before the verification screen appears. Keep that input active through the request so the existing OTP autofocus can take over.

The identifier still blocks edits while loading. Duplicate submissions are blocked, and other controls keep their disabled behavior. Submit preserves the existing identifier focus without calling `focus()`.

Draft pending an iPhone check with the npm snapshot. Submit with the keyboard open and check that the first OTP digit needs no extra tap. Repeat with a slow response and a failed-request retry.

## Checklist

- [x] `pnpm test` runs as expected for the affected UI flows.
- [x] `pnpm build` runs as expected for `@clerk/ui` and its dependencies.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring / dependency upgrade / documentation
- [ ] other:

### DIFF
--- a/.changeset/preserve-sign-in-keyboard.md
+++ b/.changeset/preserve-sign-in-keyboard.md
@@ -1,0 +1,5 @@
+---
+'@clerk/ui': patch
+---
+
+Fix the mobile keyboard closing during sign-in submission before the verification code input appears.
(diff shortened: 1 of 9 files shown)


nikosdouvlis:
!snapshot otp-keyboard-handoff

nikosdouvlis:
!snapshot otp-keyboard-handoff

PR #9748, opened 2026-09-14
feat(ui): open a page for each enterprise connection

NicolasLopes7:
## Description

Stacked on #9729. That PR lists one row per enterprise connection on the organization Security page, with a menu whose Edit opens the setup wizard. Editing one input through the wizard means walking steps back and forth, so this PR replaces the menu with a page per connection, modelled on the Dashboard's SAML connection page.

What changes for an organization admin:

- Clicking a connection row opens the connection page. The row keeps its provider icon, name, domains, and status badge, and ends with a chevron. The three-dots menu is gone.
- The header shows the name, provider, and status, and carries one action: **Activate** for an inactive connection, or **Continue setup** for a connection still in progress, which opens the wizard scoped to it. An active connection has no header action.
- **Name** with an inline rename, and **Domains** as read-only chips (they come from the domains selected in the wizard).
- **Service provider** lists the values to copy into the identity provider as copy-to-clipboard inputs: ACS URL, entity ID, and metadata URL for SAML, redirect URI for OIDC.
- **Identity provider** lists the current configuration as read-only rows (metadata URL, or sign-on URL and issuer, plus the certificate expiry for SAML; client ID and discovery endpoint or endpoints for OIDC). **Edit** opens the wizard's SAML or OIDC form inline with Save and Cancel.
- **Settings** is one form of checkboxes with Save and Cancel: sync user attributes, allow additional identifiers, and for SAML allow subdomains, allow IdP-initiated flow, and force authentication. Save sends one update with only the changed settings.
- **Danger zone** sits at the bottom with **Deactivate connection** (active connections only) and **Remove connection**, the latter with the same typed confirmation as before. Remove returns to the overview.
- The SSO section's info icon moves next to the section title.
- The wizard banner naming the scoped connection is removed; the wizard is now entered from the connection page or the Add connection button, which already say which connection is in play.

The page follows the existing profile page patterns (rows with an inline Edit that opens a form card, checkboxes with a Save button, a danger section at the bottom) rather than the Dashboard's layout.

Under the hood, the Security page tracks its views as one union (`overview`, `wizard`, `connection`, `directorySync`) instead of a string plus a boolean, and falls back to the overview when the opened connection disappears from the list. The back control is now one component shared by the wizard, the directory sync wizard, and the page. The SAML identity provider form no longer renders `ActiveConnectionAlert` itself; the four SAML wizard steps render it, matching the OIDC steps, so the form works outside the wizard context.

Not in this PR: attribute mapping (not readable from the resource), organization account linking (not in the update params), and test runs (still in the wizard).

New localization keys live under `organizationProfile.securityPage.connectionPage` in `en-US` only. Five checkbox ids are added to the public `FieldId` union, six section ids to `ProfileSectionId`, and the `organizationProfileSecuritySsoConnectionRow`, `organizationProfileSecuritySsoConnectionPage`, and `configureSSOVerifyDomainCardCheckbox` appearance keys; all additive. The `configureSSO.connectionScopeBanner` localization keys stay in the type since they shipped, but are no longer rendered. The `ssoSection.menuAction__*` keys are no longer rendered but stay in place.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


gabrielmeloc22 (commented):
A couple of UI nits and one main UX problem

- seems like it's not possible to select the actual domain i want to use for the new connection, and then when i try to continue no error is surfaced but it says the domain is being used. ideally, i think we should let the C2 select the domains specifically for the new connection, and probably only allow to select the unused ones, since we don't allow  two connections use the same domain, except if the feature flag is enabled.

https://github.com/user-attachments/assets/b5115d13-9ea1-44b7-95c1-5002f8bdb17c


- the left padding on the domain text looks a bit odd
Image

- the identity provider section content seems to be top-aligned, but the other sections don't
Image

kalafut:
It is a little confusing when you have data on a connection (URLs clearly shown) and the click **Continue setup** and it's like you're starting anew.

image

gabrielmeloc22 (approved):

PR #9749, opened 2026-09-14
chore(repo): gate cognitive complexity with bulk suppressions

alexcarpenter:
## Description

Adds `sonarjs/cognitive-complexity` as an error at 15 so new functions can't land over the limit.

Cognitive over ESLint's `complexity`: cyclomatic counts every `??`, `?.`, `&&`, and default param as a branch, so a component merging props from context scores like a 60-line state machine. Cognitive scores a fallback chain as one and penalizes nesting instead, so it flags functions that are genuinely hard to follow rather than ones that are just wide.

Existing violations (92 functions, 15 packages) are grandfathered via [ESLint bulk suppressions](https://eslint.org/docs/latest/use/suppressions) in each package's `eslint-suppressions.json`. No inline disables, no refactoring required to land this. Fixing one leaves a stale entry that fails lint until pruned, so the baseline only shrinks.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


AGENTS.md
```diff
@@ -10,6 +10,7 @@ Clerk's JavaScript SDK and library monorepo.
...
 - PR descriptions follow `.github/PULL_REQUEST_TEMPLATE.md` and add no sections of their own. Never add a "Testing" (or "Test plan" / "How to test") section summarizing the tests written or the checks run; the Checklist covers that and reviewers read the diff. Describe the change, not the work done on it.
+- Keep functions under a cognitive complexity of 15 (`sonarjs/cognitive-complexity`, reported as an error). Pre-existing violations are grandfathered in each package's `eslint-suppressions.json`. A change that pushes a function past the limit must split it; do not add new suppressions or `eslint-disable` comments for this rule. When a grandfathered function is simplified, run `pnpm eslint src --prune-suppressions` in that package and commit the result.
```
dstaley: lets try without this rule. I think the agent would likely do the right thing.

alexcarpenter: 👍🏼 https://github.com/clerk/javascript/pull/9749/commits/007419aed38a3a458054f4ccd9a3e7ab9748ddf4

PR #9750, opened 2026-09-14
feat(ui): add the Mosaic user profile password section

kylemac:
## TLDR from a Human
- First pass at implementing the Password Section for UserProfile. Also documents it in Swingset

## AI Description

Implements the password section of the Mosaic `UserProfileSecurityPanel`, following the edit-name and edit-username flows: a leaf section view owning a machine-backed controller and a dialog view that sets or changes the password. The dialog carries every state of the shipped `PasswordForm`: set vs. change copy, the current-password field when the account has a password and reverification is off, a confirm field that must match, the sign-out-of-other-devices checkbox, an inert read-only mode for enterprise-only accounts, pending and failure states with banner and field-scoped errors, and an eye toggle on each field via `InputGroup`. The security panel replaces `onChangePassword` with `onSubmitPassword`, `requiresCurrentPassword`, and `isPasswordReadOnly`, and renders the section whenever a password exists or can be set. Swingset gains a fixture and five stories for the section, and the security panel and profile fixtures drive the real dialog.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


austincalvelage (approved):
Approving noting, We'll follow-up with a checkbox component

PR #9751, opened 2026-09-14
refactor(ui): bring Mosaic Button and Input under the complexity limit

alexcarpenter:
## Description

Stacked on #9749. A worked example of clearing two grandfathered `complexity` suppressions, to show what a reduction looks like before deciding how far to push the rule.

**Button** (25 → 14). The stylex composition moves into a `buttonStyleProps` function alongside the two derived flags it alone uses. `MosaicButton` keeps resolving props against `ButtonContext` and rendering. No behaviour change.

**Input** (28 → 15). `useOptionalFieldControlProps` already returns every control attribute with the own prop as fallback, or `null` outside a `Field`. The five `fieldProps?.x ?? x` lookups collapse to one `fieldProps ?? ownProps` choice, and the two variant style conditionals become a lookup. No behaviour change.

Both stale entries are pruned from `packages/ui/eslint-suppressions.json`, so this is also the first run of the prune flow described in #9749.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


alexcarpenter:
Closing. The reductions here are mostly metric-chasing: cyclomatic complexity counts every `??` and `?.`, so prop-merging primitives score high without being hard to read. Evaluating cognitive complexity instead on #9749.

PR #9754, opened 2026-09-14
feat(ui): confirm Web3 wallet removal from its row

austincalvelage:
## Description

Adds a confirmation dialog before removing a Web3 wallet. Each row owns its removal state and error handling, with the dialog in a separate file using the existing menu/dialog focus pattern. Connect and Set as primary behavior are unchanged.

Try Remove wallet in the Web3 wallets Swingset example, then cancel or confirm.

## Checklist

- [ ] (If applicable) JSDoc comments have been added or updated for any package exports

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/user-profile/user-profile-web3-wallets-section.view.tsx
```diff
@@ -1,118 +1,66 @@
...
+}
+
+export interface UserProfileWeb3Wallet {
+  id: string;
+  address: string;
+  provider?: string;
+  iconUrl?: string;
   isPrimary?: boolean;
-  isVerified?: boolean;
+  isVerified: boolean;
```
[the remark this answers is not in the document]
austincalvelage: Requiring it makes the display state explicit and matches legacy, which checks verification.status === 'verified'. Defaulting missing data to verified would hide mistakes in future controller wiring.

packages/ui/src/mosaic/user-profile/user-profile-remove-web3-wallet.dialog.tsx
```diff
@@ -0,0 +1,67 @@
...
+            <Banner.Label>{errorMessage}</Banner.Label>
+          </Banner.Root>
+        ) : null}
+        <Dialog.Actions>
+          <Dialog.Close
+            disabled={isPending}
+            render={<Button variant='outline' />}
+          >
+            {m.removeDialog.cancel}
+          </Dialog.Close>
```
maxyinger, line 53: Might have missed this in connected accounts. but since `esc` still calls to close the dialog, it can be used to exit this while it's pending. I think you have to block the `setOpen(false)` when it's pending. I think this might all be sorted out in the confirmation block alex is adding tho.

also tho if we have that behavior, do you have to reload the page to get out of that pending story? maybe best to set a timeout or something on that one so it isn't stuck

packages/swingset/src/stories/fixtures/user-profile-web3-wallets.ts
```diff
@@ -0,0 +1,111 @@
...
+      }
+      setWallets(current =>
+        current.map(wallet => ({ ...wallet, isPrimary: wallet.id === id, primaryError: undefined })),
+      );
+    },
+    onRemove: (id: string) => {
+      setWallets(current => {
+        if (removalState === 'pending') {
+          return current.map(wallet => (wallet.id === id ? { ...wallet, isRemoving: true } : wallet));
+        }
```
maxyinger, line 101: maybe set a timeout here or something so it eventually exits

maxyinger (approved):
mostly same feedback as what I left in connected accounts pr, so won't repeat it here.

For the pending story might be nice to cap it with a timeout or something so the playground doesn't get caught in a stuck state. Think Alex's confirmation block addresses the `esc` behavior there

packages/swingset/src/stories/fixtures/user-profile-web3-wallets.ts
austincalvelage: Added!

packages/ui/src/mosaic/user-profile/user-profile-remove-web3-wallet.dialog.tsx
austincalvelage: Will leave for now following up Alex's changes tomorrow

PR #9726, opened 2026-09-11
fix(headless): measure flow steps without parent transforms

austincalvelage:
## Description

Using `offsetHeight` rather than `getBoundingClientRect().height`. This gives us the full height keeping the container sized correctly while the dialog animates.

Before:
image

After:
image

## Checklist

- [ ] `pnpm test` runs as expected.
- [ ] `pnpm build` runs as expected.
- [ ] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [x] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #9756, opened 2026-09-14
feat(ui): animate mosaic field messages

alexcarpenter:
## Description

Restores the height animation Mosaic field messages had in the legacy `FormFeedback`, using the headless `useTransition` data-attribute pattern the other Mosaic components follow.

- Adds `Field.Message`, a container that measures whichever message is showing (via `ResizeObserver`) and transitions its own height to match. It grows when a message arrives, follows the text when the message changes, and collapses after it clears, so callers never conditionally mount a message.
- Adds `Field.Success`, the positive counterpart to `Field.Error`, sharing one internal feedback part (icon + text, `aria-describedby` registration). When one replaces the other inside `Field.Message`, the outgoing message fades out absolutely positioned over the incoming one while the container height moves once from the old height to the new.
- `Field.Error` and `Field.Success` keep the last message through the exit transition and unmount once animations finish; both expose `data-open` / `data-closed` / `data-starting-style` / `data-ending-style` plus the `.cl-field-message`, `.cl-field-error`, and `.cl-field-success` slot classes. Under `prefers-reduced-motion` the height snaps and only the message fades.
- `Field.Message` is also the field's live region: it stays mounted with `role='status'`, so a message that arrives while the user is in the control is announced politely, and a message in its exit transition is `aria-hidden` so only the incoming one is read. Callers pass `role='alert'` when a message must interrupt. `aria-describedby` on the control always points at the open message only.
- Existing Mosaic call sites (destructive block, reverification panels, edit name and username dialogs) are wrapped in `Field.Message`.
- `Field.Message` measures with a small local hook rather than `useMeasure`. It needs a node handed in from registration (the measured child changes on every swap) and a synchronous first read so the height variable lands in the same commit as `data-starting-style`; `useMeasure` attaches its own ref and waits for the first `ResizeObserver` callback. Teaching `useMeasure` to cover both, and moving this and the headless `FlowRoot` onto it, is a follow-up.
- Swingset: the field page documents the new parts, the error example is wider so long messages no longer reflow, and a new "Success" example swaps a two-line error for a one-line success message.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/ui/src/mosaic/components/field/field.tsx
```diff
@@ -116,42 +125,160 @@ const Description = React.forwardRef<HTMLParagraphElement, FieldDescriptionProps
...
+  success: { slot: 'field-success', icon: 'check', color: styles.success },
+};
+
+interface FieldFeedbackProps extends MosaicComponentProps<'p'> {
+  kind: FieldFeedbackKind;
+}
+
+function hasMessage(children: React.ReactNode) {
+  return React.Children.toArray(children).some(child => child !== '');
+}
```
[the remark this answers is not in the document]
alexcarpenter: Leaving as is. `React.Children.toArray` already drops null, undefined, booleans, and we filter empty strings; every call site passes a string or undefined. A fragment wrapping only null is not a shape we use. If a second consumer needs a children-emptiness check we will extract a shared util with a recursive walk then, with real cases in hand.

packages/ui/src/mosaic/components/field/field.styles.ts
```diff
@@ -18,15 +20,50 @@ export const styles = stylex.create({
   description: {
     color: colorVars['--cl-color-neutral-faded'],
   },
-  error: {
+  messageRoot: {
+    overflow: 'clip',
+    alignContent: 'start',
+    display: 'grid',
```
maxyinger, line 26: any purpose for this grid? looks like we're popping exiting animation out of flow?

packages/ui/src/mosaic/components/field/field.styles.ts
```diff
@@ -18,15 +20,50 @@ export const styles = stylex.create({
...
+      '@media (prefers-reduced-motion: reduce)': 'none',
+    },
+    transitionTimingFunction: easingVars['--cl-ease-enter'],
+    height: {
+      default: 'var(--_cl-field-message-height)',
+      ':where(:not([data-open]), [data-starting-style])': 0,
+    },
+    marginTop: {
+      default: 0,
+      ':where(:not([data-open]), [data-starting-style])': `calc(-1 * ${ROOT_GAP})`,
```
maxyinger, line 40: ahh to account for the gap jump. nice.

packages/ui/src/mosaic/components/field/field.styles.ts
```diff
@@ -18,15 +20,50 @@ export const styles = stylex.create({
...
     gap: space['1'],
     alignItems: 'flex-start',
-    color: colorVars['--cl-color-negative'],
     display: 'flex',
+    opacity: { default: 1, ':where([data-starting-style], [data-ending-style])': 0 },
+    position: { default: null, ':where([data-ending-style])': 'absolute' },
     textWrap: 'pretty',
+    transitionDuration: durationVars['--cl-duration-base'],
+    transitionProperty: 'opacity',
+    transitionTimingFunction: 'linear',
```
maxyinger, line 53: Sometimes I try to avoid these frames where they cut the content with different exiting or entering easings/delays/durations or adding a slight scale anim. more of a reach goal tho/might add too much flavor to it. Already feels nice with the gap jump accounted for ❤️ 

Image

packages/ui/src/mosaic/components/field/field.tsx
```diff
@@ -116,42 +125,160 @@ const Description = React.forwardRef<HTMLParagraphElement, FieldDescriptionProps
...
+      return undefined;
+    }
+
+    const observer = new ResizeObserver(measure);
+    observer.observe(active);
+    return () => observer.disconnect();
+  }, [active]);
+
+  return height;
+}
```
maxyinger, line 152: this is basically what we're doing in the flow component right? I added a `useMeasure` in here. might have to update it to use offsetHeight. looks like I missed adopting it in the Flow component

maxyinger (approved):
looooks gooooood

packages/ui/src/mosaic/components/field/field.styles.ts
alexcarpenter: yeah, was toying around with a mask during reveal, but didn't land on anything solid. something to maybe polish in a future PR https://x.com/blvdmitry/status/2000860557298045163

packages/ui/src/mosaic/components/field/field.tsx
alexcarpenter: austin fixed that here recently https://github.com/clerk/javascript/pull/9726

packages/ui/src/mosaic/components/field/field.styles.ts
maxyinger: ooo I like that. yea def a future improvement and not a blocker

packages/ui/src/mosaic/components/field/field.tsx
maxyinger: yea I'm just wondering if we should unify into a central `useMeasure` since they all basically do the same thing?

I have this one:
https://github.com/clerk/javascript/blob/main/packages/ui/src/mosaic/hooks/useMeasure.ts

but needs to be updated to mirror austin's fix for flow and adopted everywhere

PR #9338, opened 2026-08-05
feat(nextjs): export Mosaic UserButton from an experimental subpath

alexcarpenter:
## Description

Stacked on https://github.com/clerk/javascript/pull/9185

Adds an experimental subpath so Mosaic components mount directly in the host app's tree rather than being rendered by clerk-js. `UserButton` is the first one. It reads Clerk through hooks, so a `ClerkProvider` above it is all it needs:

```tsx
import { UserButton } from '@clerk/nextjs/experimental/mosaic';
```

```css
@import '@clerk/nextjs/experimental/mosaic/styles.css' layer(clerk);
```

## Checklist

- [x] `pnpm test` runs as expected.
- [x] `pnpm build` runs as expected.
- [x] (If applicable) [JSDoc comments](https://jsdoc.app/about-getting-started.html) have been added or updated for any package exports
- [ ] (If applicable) [Documentation](https://github.com/clerk/clerk-docs) has been updated

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] 📖 Refactoring / dependency upgrade / documentation
- [ ] other:

PR #9765, opened 2026-09-15
feat(mosaic): extract mosaic into @clerk/mosaic

Ephem:
## Description

This PR moves Mosaic into its own `@clerk/mosaic` package.

> This PR is best reviewed commit-by-commit.
>
> The first commit is what sets up the new package, the second moves Mosaic and cleans up `ui`, the rest is mostly path updates.

In its current form, merging this PR will also release a first version of Mosaic, which makes it publicly available. In other words, this is also an entry-point PR replacing the strategy in https://github.com/clerk/javascript/pull/9338 (for the experimental phase at least).

This PR exports `UserButton` directly from `@clerk/mosaic` and lets the <1.0 version and readme indicate that it's experimental. An alternative is to export it from `@clerk/mosaic/experimental` to make it extra clear, but I don't think that's necessary.

When components are stable, they will move to be exported from the framework packages like `@clerk/nextjs`. This package will still be the home of the components themselves after that, and `ui` will move to import them from here and become a compatibility layer for non-React frameworks (is the current plan, until it changes).

> ~~⚠️ Before merging, we need to set up trusted publishing for `@clerk/mosaic` or else the publish will fail.~~
> ✅ Trusted publishing has been set up

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


packages/mosaic/README.md
Ephem, line 1: This was mostly AI, I did tweak a few things and I think it's pretty decent for a first small version but definitely could use extra 👀

packages/mosaic/README.md
```diff
@@ -0,0 +1,77 @@
...
+      <img src="https://images.clerk.com/static/logo-light-mode-400x400.png" height="64">
+    </picture>
+  </a>
+  <br />
+  <h1 align="center">@clerk/mosaic</h1>
+</p>
+
+<div align="center">
+
+[![Chat on Discord](https://img.shields.io/discord/856971667393609759.svg?logo=discord)](https://clerk.com/discord)
```
dstaley: I think we can remove this?

packages/mosaic/README.md
```diff
@@ -0,0 +1,77 @@
...
+·
+[Get help](https://clerk.com/contact/support?utm_source=github&utm_medium=clerk_mosaic)
+
+</div>
+
+## Getting started
+
+`@clerk/mosaic` is Clerk's experimental next-generation React components.
+
+This package let's us build and iterate on our next component refresh in public, feel free to try them out and provide feedback. If you do, know that we reserve the right to make breaking changes in minors and pin your version.
```
dstaley:
```suggestion
This package lets us build and iterate on our next component refresh in public, feel free to try them out and provide feedback. If you do, know that we reserve the right to make breaking changes in minors and pin your version.
```

packages/mosaic/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+When stable, components will move to be exported through the regular packages.
+
+### Prerequisites
+
+- Node.js `>=20.9.0` or later
+- A Clerk SDK that renders a `ClerkProvider` (`@clerk/react`, `@clerk/nextjs`, …)
```
dstaley: not sure if it's worth mentioning, but installing a version of `@clerk/mosaic` that references a different version of `@clerk/shared` than their framework SDK package will result in the "must be rendered under ClerkProvider" error since there are two different ClerkProvider types. So they would need to make sure to install the version of their framework SDK that was published at the same time as `@clerk/mosaic`.

packages/mosaic/tsdown.config.mts
```diff
@@ -0,0 +1,33 @@
+import stylexPlugin from '@stylexjs/rollup-plugin';
+import { defineConfig } from 'tsdown';
+
+import { mosaicLightningCssTargets } from './stylex-lightningcss.config.mjs';
+
+export default defineConfig({
```
dstaley, line 6: we can review this later, but I think we need to avoid actually bundling UI components since it makes it more difficult to tree-shake unused code.

packages/mosaic/tsdown.css.config.mts
```diff
@@ -0,0 +1,35 @@
...
+      useCSSLayers: true,
+      lightningcssOptions: { targets: mosaicLightningCssTargets },
+    }),
+  ],
+  onSuccess: async () => {
+    const { cp, rm } = await import('node:fs/promises');
+    await cp('./dist-css/styles.css', './dist/styles.css');
+    await rm('./dist-css', { recursive: true, force: true });
+    await rm('./dist/styles.partial.css', { force: true });
+    await rm('./dist/index.js.map', { force: true });
```
dstaley, line 33: for future investigation, this seems like we're holding the bundler wrong

dstaley (approved):

packages/mosaic/README.md
Ephem: This is very true.. Hmm, I wonder if we should let Mosaic just bundle the shared stuff it needs? It's not much. When we move exports over to the framework packages when they are stable this wont be an issue.

I'll go ahead and do that in this PR tomorrow. 😄

Ephem: I was tired yesterday when I wrote this, I had the assumption we were only using a few shared helpers that we could specifically bundle to avoid this, but ofc we are using `useClerk` etc inside Mosaic so that wont help.

Need to think through this some more. Keeping Mosaic and framework packages in sync would be very annoying and opposite to what we want, we want people to be able to pin Mosaic but not the main package. 🤔

Ephem: I think your original idea of just documenting this was right. This should normally not be a problem, if it is, you do want to resolve it so you don't have duplicate packages. I added a troubleshooting section to the PR with details and how to fix it if you see the error we already have in place for this.

packages/mosaic/README.md
Ephem: I've done a bunch more iteration here since I wrote this.

packages/mosaic/README.md
Ephem: Done!

PR #9777, opened 2026-09-15
feat(ui): view transition list re-order exploration

maxyinger:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


maxyinger:
I don't think view transitions are it here. we'll need to go with flip on the tag input so we should use that here too

PR #9787, opened 2026-09-16
refactor(ui): resolve mosaic messages through the shared helpers

alexcarpenter:
## Description

Follows up https://github.com/clerk/javascript/pull/9786 by moving every Mosaic feature onto the shared message helpers:

- Deletes the four per-feature copies of `fill` and `plural` and imports them from `src/mosaic/utils/messages.ts`.
- Replaces the inline `.replace('{provider}', …)` calls in the connected-account, web3-wallet and enterprise-account rows and dialogs with `fill`.
- Replaces the `.split('{emailAddress}')` pattern in the remove-email, remove-phone and verify-email-link dialogs with `rich`, so the bolded value is authored in the messages file as `{#strong}…{/strong}`.
- Passes the resend countdown span in the verify-email-link, add-email and add-phone dialogs as an element value, so the string stays `Resend ({seconds})` and the view owns the wrapper.
- Renames the `*Base` exports to `*Messages` so every messages file exports the same shape under the same name.

No rendered output changes.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


austincalvelage (approved):
Nice cleanup!

PR #9788, opened 2026-09-16
fix(repo): wait for npm availability before release dispatches

Ephem:
## Description

After publish, npm can still take some time before the packages are available. This fix makes the job pending until we actually see the packages live, and dispatches the dependent workflows only after that.

You can run the script locally: `node scripts/wait-for-packages-on-npm.mjs`

You can also tweak a version in e.g. `@clerk/nextjs` `package.json` to something that doesn't exist and run the script, it will retry and output something like:

```
Attempt 1/60: waiting for @clerk/nextjs@8.0.0
Attempt 2/60: waiting for @clerk/nextjs@8.0.0
```

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


scripts/wait-for-packages-on-npm.mjs
```diff
@@ -0,0 +1,97 @@
...
+
+    if (unavailable.length === 0) {
+      log(`All ${packages.length} public package versions are available on npm.`);
+      return;
+    }
+
+    const versions = unavailable.map(packageInfo => `${packageInfo.name}@${packageInfo.version}`).join(', ');
+
+    if (attempt < maxAttempts) {
+      log(`Attempt ${attempt}/${maxAttempts}: waiting for ${versions}`);
```
Ephem, line 81: This script was entirely vibed of course. I reviewed it pretty carefully and it looks good to me, but I wonder if this log will get verbose when a lot of packages takes 5 minutes to publish?

dstaley (approved):

dstaley:
thinking out loud: does it make sense for this check to live here in the publishing, rather than wherever we're attempting to consume a specific version? If the thinking is that the publishing succeeds, but the npm API needs time to report the new version, I wonder if it's better for us to poll for the specific package/version in the places where we're expecting a specific version to be available on npm. The failure mode here is in consumption of the specific package, not in its publishing.

Ephem:
@dstaley I went back and forth on this. The other argument is "Is a publish truly complete before the packages are actually available"?

Discussing it with AI was equally back-and-forth. On the one hand, the consumers should be resilient (but not dispatching until actually available is a good courtesy to have in this repo anyway). On the other hand, the above argument.

I'm struggling to form a real opinion myself here tbh (not even sure if I care a little or a lot either way), hence the "Not sure if we fix this by delaying the workflow dispatches in javascript until after the packages have actually landed, or if we update sdk-infra-workers and AP to retry for far longer." question in Slack. Do you have a strong opinion or just posing the (good) question? I'd love any input/thoughts.

Ephem:
I went ahead and merged this to unblock, but happy to continue the discussion on where the wait should live. 😄

PR #9796, opened 2026-09-16
fix(ui): position mosaic link button underline from the font metric

maxyinger:
## Description

small update to use from-font vs 2px offset for link buttons. I think skip ink auto is already the browser default, but just added in there to be more explicit.

visual change isn't much if anything, but makes this much more resilient to people using other fonts vs hardcoding `2px` for offset

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


alexcarpenter (approved):
TIL!

PR #9799, opened 2026-09-16
feat(ui): keep a sign-out row at the foot of the Mosaic user button in single session

alexcarpenter:
## Description

The Mosaic user button's foot now always ends in a sign-out row. With a second signed-in account it stays **Sign out of all accounts**; with only the one, the same row becomes **Sign out** of the active session instead of disappearing. Previously a single-session surface had no sign-out at the foot at all.

`menuItemOrder` gains a `signOut` id alongside `signOutAll`, the same way `switchAccount` and `addAccount` name the two forms of the accounts row, so one order can place the sign-out row whichever way it resolves.

The account's own `⋯` (combined mode) and the header button (user mode) still carry **Sign out** as well; that duplication is intended.

Preview: https://swingset-git-carp-mosaic-user-button-sign-out.clerkstage.dev/user-button/user-button

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


alexcarpenter:
Superseded by #9807, which carries this change on the new `packages/mosaic` paths: the foot always ends in a sign-out row, `signOutAll` with a second account and `signOut` of the one account without.

PR #9802, opened 2026-09-16
feat(ui): Add MFA section + flow

austincalvelage:
## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [ ] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


packages/swingset/src/stories/fixtures/user-profile-mfa-example.tsx
```diff
@@ -0,0 +1,113 @@
...
+                  />
+                </Flow.Step>
+                <Flow.Step ids={['backup-codes']}>
+                  <UserProfileBackupCodesView
+                    {...current.backupCodes}
+                    onCancel={() => current.setup.onOpenChange(false)}
+                  />
+                </Flow.Step>
+              </>
+            )}
```
maxyinger: think I'm missing some context here. why move to flow in the example vs the add mfa dialog?

maxyinger (commented):
just one question. Think I'm missing some context, but things are looking good

packages/swingset/src/stories/fixtures/user-profile-mfa-example.tsx
austincalvelage: I was documenting each step separately similar to Reverification, but ended up putting the combined Flow in the example instead of the reusable view. Let me fix that up.

austincalvelage: Reducded to a single example with the MFA dialog

maxyinger: dope dope

packages/mosaic/src/features/user-profile/user-profile-add-authenticator.view.tsx
```diff
@@ -0,0 +1,146 @@
...
+            invalid={Boolean(errorMessage)}
+            xstyle={styles.field}
+          >
+            <Field.Label xstyle={styles.label}>{m.codeLabel}</Field.Label>
+            <Otp
+              name='code'
+              value={code}
+              onValueChange={onCodeChange}
+              onComplete={submitCode}
+            />
```
maxyinger, line 112: should otp gain the `actionRef` from `useFlowAutoFocus` vs the submit?

packages/mosaic/src/features/user-profile/user-profile-add-authenticator.view.tsx
```diff
@@ -0,0 +1,146 @@
...
+        <Button
+          type='button'
+          variant='outline'
+          color='neutral'
+          fullWidth
+          disabled={Boolean(setup) && isPending}
+          onClick={onCancel}
+        >
+          {m.cancel}
+        </Button>
```
maxyinger, line 129: Trying to decide if this is like user testing bias on my end, but do these first views after the mfa selection step feel like they should be `back` buttons instead of `cancel` to you?

packages/mosaic/src/features/user-profile/user-profile-add-authenticator.dialog.tsx
```diff
@@ -0,0 +1,43 @@
...
+        >
+          <UserProfileAddAuthenticatorView
+            {...props}
+            onCancel={() => onOpenChange(false)}
+          />
+        </Card.Root>
+      </Dialog.Popup>
+    </Dialog.Root>
+  );
+}
```
maxyinger: When will we use these isolated mfa method dialogs?

packages/mosaic/src/features/user-profile/user-profile-authenticator-setup.view.tsx
```diff
@@ -0,0 +1,106 @@
...
+              >
+                <Banner.Label>{copyErrorMessage}</Banner.Label>
+              </Banner.Root>
+            ) : null}
+            <Text
+              role='status'
+              aria-label={m.copyFeedback}
+            >
+              {copyStatus === 'pending' ? m.copying : copyStatus === 'success' ? m.copied : null}
+            </Text>
```
maxyinger: this is feeling a bit off. maybe something worth chatting through with design?

https://github.com/user-attachments/assets/1ea0f722-66c3-40a2-858f-d546daf09f75

packages/mosaic/src/features/user-profile/user-profile-backup-codes.styles.ts
```diff
@@ -0,0 +1,52 @@
...
+    animationIterationCount: 'infinite',
+    animationName: {
+      default: pulse,
+      '@media (prefers-reduced-motion: reduce)': 'none',
+    },
+    animationTimingFunction: 'cubic-bezier(0.4, 0, 0.6, 1)',
+    backgroundColor: colorVars['--cl-color-neutral-alpha-200'],
+    height: '1lh',
+    width: space['16'],
+  },
```
maxyinger, line 22: ooo this looks fun. maybe worth adding some latency to the fixture so we can see it? having trouble finding where this plays in the playground.

for cubic beziers, I'm trying to stick to tokens. ok if none of the existing ones cover this scenario well, but if we add a new one like this maybe worth adopting as a token and mentioning when to use it in the skill

packages/mosaic/src/features/user-profile/user-profile-add-authenticator.view.tsx
```diff
@@ -0,0 +1,146 @@
...
+          {...setup}
+          onCopy={onCopy}
+          copyStatus={copyStatus}
+          copyErrorMessage={copyErrorMessage}
+        />
+      ) : (
+        <>
+          <Card.Header>
+            <Card.Title>{setupMessages.title}</Card.Title>
+          </Card.Header>
```
maxyinger, line 66: Polish follow up, but these two being just slightly different heights makes me either want to either:
- be able to temporarily disable the height animation from inside a flow
- or force them to be the same height

maxyinger (approved):
Everything looks good. This is a big boi, so in favor of just getting it in and doing polish followups.

Main question on this one is do we plan on using these dialogs anywhere?
- `user-profile-add-authenticator.dialog.tsx`
- `user-profile-add-sms.dialog.tsx`
- `user-profile-backup-codes.dialog.tsx`

Outside of that only thing that doesn't feel quite right is the copy banner in the authenticator setup view, but feels like something new we should discuss with design

Nice work!

packages/mosaic/src/features/user-profile/user-profile-add-authenticator.view.tsx
austincalvelage: Asked Preston here about restructuring in away they maintain the same height. https://www.figma.com/design/yt5qKlfNyIIdaKaMoav1v2/Clerk-Mosaic-2.0?node-id=9578-28439#1933135188

If not we can look into disabling height animations within the flow.

packages/mosaic/src/features/user-profile/user-profile-authenticator-setup.view.tsx
austincalvelage: Oof this one slipped through. Removed we have an error state if for some reason the copy fails. I think eventually we create a block/component w/ tooltip and button for copy actions

PR #9807, opened 2026-09-17
feat(ui): add a user-button-header slot and layouts to the Mosaic UserButton

alexcarpenter:
## Description

Reworks the Mosaic `UserButton` popup header to support the header layouts in the latest design, and adds slot classes to the surface's outer parts.

- The header is its own `user-button-header` slot instead of an `Item` row, with `-content`, `-title`, `-description`, and `-actions` sub-slots and a `data-layout` axis.
- New `headerLayout` prop (`'inline'` default, or `'stacked'`). Inline trails the workspace with the actions and shows the gear as an icon alone. Stacked runs them under the workspace as full-width labelled buttons, with the gear reading "Settings" and leading the row.
- The trigger and popover carry `cl-user-button-trigger` and `cl-user-button-popover` classes.
- An account that leads a combined surface is described by the organization it works in and its role there ("Clerk app · Admin"), and its avatar wears the organization's avatar in the corner (10px in the trigger, 14px in the header). `user` mode keeps the identifier. This adds `roleLabel` to `UserButtonMembership`, sourced from the membership's `roleName`.
- The organization subtitle now reads plan first, then member count.

Swingset gets `StackedHeader` and `UserPriorityStackedHeader` stories, a "Header" section, and a slot table for the new classes.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


alexcarpenter:
!snapshot

alexcarpenter:
!snapshot

dstaley:
!snapshot

dstaley:
!snapshot

dstaley:
!snapshot

maxyinger (approved):
looks good!

Keep goin back n forth on if the nested avatar should be included in the flow or not.

its so small it's not a huge deal either way, but `Alt` here keeps the subavatar x axis in layout flow and pops it out on the y. also is weird but 1px offset on y with 2px on x feels better to me than 2px on each side 🤷 . anyway that's what's goin on in second image

| Current | Alt |
| --- | --- |
| Image |  Image |

PR #9809, opened 2026-09-17
feat(ui,clerk-js,shared,localizations): Manage the SSO bypass allowlist from the orgs Security page

mauricioabreu:
## Description

Org admins had no way to manage who may sign in with an email code when their identity provider is down. Only the C1 could, from the Dashboard.

This adds an "SSO bypass" row to the Security page of `<OrganizationProfile />` showing the size of the allowlist, with a Manage action that opens a page to search, add and remove members. Both are gated on the new `org:sys_entconns_sso_bypass:manage` permission, seeded by clerk/clerk_go#21999 and enforced by the FAPI routes from clerk/clerk_go#22192. The bypass applies to every enterprise connection, so the Security page renders each section by the permissions the member holds: a member with only the bypass permission sees the connections read-only next to the allowlist.

For custom flows, `organization.ssoBypassAllowlist` exposes `getUsers()`, `addUser({ userId })` and `removeUser(userId)`, returning the new `SSOBypassAllowlistUserResource`.

Resolves ORGS-1822.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


.changeset/sso-bypass-allowlist.md
```diff
@@ -0,0 +1,12 @@
...
+'@clerk/shared': minor
+'@clerk/ui': minor
+---
+
+Organization admins can now manage the SSO bypass allowlist from the Security page of `<OrganizationProfile />`. Members on the list can sign in with an email code when the organization's identity provider is unavailable. The page shows how many members are allowlisted, and the Manage action opens a page to search the list, add a member, or remove one. Both the section and the page require the new `org:sys_entconns_sso_bypass:manage` permission, and they also appear for organizations whose enterprise connections were set up by the instance owner rather than through self-serve SSO. In that case the SSO section lists the connections without any configuration actions.
+
+For custom flows, `OrganizationResource` gains `getSsoBypassAllowlistUsers()`, `addSsoBypassAllowlistUser({ userId })` and `removeSsoBypassAllowlistUser(userId)`, returning the new `SsoBypassAllowlistUserResource`.
+
+New customization handles ship with it: the `ssoBypass` `ProfileSectionId`, and the `organizationProfileSecuritySsoBypass*` appearance elements for the section, the allow list page, its search inputs, the Add button and the member picker options.
```
dstaley:
```suggestion
Add the ability for Organization admins to manage the SSO bypass allowlist from the Security page of `<OrganizationProfile />`.
```

packages/clerk-js/src/core/resources/Organization.ts
```diff
@@ -1,5 +1,6 @@
 import type {
   AddMemberParams,
+  AddSsoBypassAllowlistUserParams,
```
dstaley: let's replace `Sso` with `SSO` to be consistent with other methods/types like `ConfigureSSO`, `CreateEnterpriseSSOLinkFlowReturn`, `selfServeSSOEnabled`, etc

packages/clerk-js/src/core/resources/SsoBypassAllowlistUser.ts
```diff
@@ -0,0 +1,52 @@
...
+  createdAt!: Date;
+  updatedAt!: Date;
+
+  constructor(data: SsoBypassAllowlistUserJSON | SsoBypassAllowlistUserJSONSnapshot) {
+    this.fromJSON(data);
+  }
+
+  reload(_?: ClerkResourceReloadParams): Promise<this> {
+    clerkUnsupportedReloadMethod('SsoBypassAllowlistUser');
+  }
```
dstaley: why is this here? is it from some interface? if so, would implementing the method not work for some reason?

packages/shared/src/react/hooks/useOrganizationSsoBypassAllowlist.tsx
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  const revalidate = useCallback(
+    () => queryClient.invalidateQueries({ queryKey: [stableKey] }),
+    [queryClient, stableKey],
+  );
+
+  const addUser = useCallback(
+    async (addParams: AddSsoBypassAllowlistUserParams) => {
+      const added = await organization?.addSsoBypassAllowlistUser(addParams);
```
dstaley: seeing stuff like this makes me wonder if it would be more appropriate to model the API like `organization.ssoBypassAllowlist.addUser(addParams)`. That way the allowlist itself has methods rather than methods on the organization itself. That's a little more aligned with how we did DirectorySync I believe.

packages/ui/src/components/OrganizationProfile/OrganizationSecurityPage.tsx
```diff
@@ -25,20 +29,35 @@ type SecurityPageView =
...
+  if (!canManageConnections) {
+    return <SsoBypassOnlySecurityPage canManageSsoBypass={canManageSsoBypass} />;
+  }
+
+  return (
+    <OrganizationSecurityPageContent
+      contentRef={contentRef}
+      canManageSsoBypass={canManageSsoBypass}
+    />
+  );
```
dstaley: can we not simply render the sections that the user has the ability to manage? why do we need to have an entirely different component that only renders the SSO bypass section?

I would have expected it to look like this:

```
function OrganizationSecurityPageContent() {
  const canManageConnections = ...
  const canManageSSOBypass = ...

  return (
		<Container>
			{canManageConnections && <EntConnSection />}
      {canManageSSOBypass && <SSOBypass />}
		</Container>
  )
}
```

packages/ui/src/components/OrganizationProfile/SecuritySsoBypassSection.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { ProfileSection } from '@/ui/elements/Section';
+import { ThreeDotsMenu } from '@/ui/elements/ThreeDotsMenu';
+
+import { Badge, Col, descriptors, Flex, localizationKeys, Spinner, Text } from '../../customizables';
+
+type SecuritySsoBypassSectionProps = {
+  onManage: () => void;
+};
+
+const isFeatureNotEnabled = (error: Error | null): boolean =>
```
dstaley: let's rename this `isFeatureNotEnabledError` to make it clear that it's checking an error rather than generically checking if the feature is enabled.

dstaley (approved):
approving now so you don't need a re-review, but the only thing I'd like to see revised in this PR is the `Sso` -> `SSO` change and the changeset.

in a future PR we can refactor the Security page to be a little cleaner in how it renders the sections.

also, if you think it would be a slightly better API to do `organization.ssoBypassAllowlist.addUser` we can do that in this PR. but if you think what you have now makes more sense I'm cool going with that!

.changeset/sso-bypass-allowlist.md
mauricioabreu: manually applied. thanks!

packages/shared/src/react/hooks/useOrganizationSsoBypassAllowlist.tsx
mauricioabreu: good idea. `organization` (`OrganizationResource`) now has ssoBypassAllowlist with all the methods it needs

packages/ui/src/components/OrganizationProfile/SecuritySsoBypassSection.tsx
mauricioabreu: renamed

packages/shared/src/react/hooks/useOrganizationSSOBypassAllowlist.tsx
```diff
@@ -0,0 +1,94 @@
...
+    isSignedOut: organization === null,
+    authenticated,
+    stableKeys: stableKey,
+  });
+
+  const query = useClerkQuery({
+    queryKey,
+    queryFn: () => organization?.ssoBypassAllowlist.getUsers(),
+    enabled: queryEnabled,
+    placeholderData: defineKeepPreviousDataFn(keepPreviousData),
```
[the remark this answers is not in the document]
mauricioabreu: valid. fixed in 7d2a1bddb1

packages/ui/src/components/OrganizationProfile/useSecurityRouteAccess.ts
```diff
@@ -0,0 +1,17 @@
...
+  const { shouldShowSelfServeSSO } = useOrganizationProfileContext();
+  const canManageConnections = useProtect({ permission: 'org:sys_entconns:manage' });
+  const canManageSSOBypass = useProtect({ permission: 'org:sys_entconns_sso_bypass:manage' });
+  const canConfigureSso = shouldShowSelfServeSSO && canManageConnections;
+
+  const { data: enterpriseConnections } = __internal_useOrganizationEnterpriseConnections({
+    enabled: canManageSSOBypass && !canConfigureSso,
+  });
+
+  return canConfigureSso || (canManageSSOBypass && (enterpriseConnections?.length ?? 0) > 0);
```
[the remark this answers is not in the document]
mauricioabreu: valid. fixed in 7d2a1bddb1

packages/clerk-js/src/core/resources/Organization.ts
mauricioabreu: good catch! fixed

PR #9815, opened 2026-09-17
feat(mosaic): add inset outline to Avatar with a bordered prop

alexcarpenter:
## Description

Adds a 1px inset outline to the Mosaic Avatar so a photo or fallback that shares the page's background still reads as a bounded shape. The line is drawn with `outline` and a `-1px` offset on `Avatar.Image` and `Avatar.Fallback`, so it never shrinks the box and never collides with the focus ring on an interactive root.

Figma's `border-overlay` has no Mosaic token, so the colour is `--cl-color-neutral-alpha-100`: the translucent twin of `--cl-color-border`, which lands at the same weight over a matching background and still shows over an image.

`Avatar.Root` gains a `bordered` prop (default `true`, reflected as `data-bordered`) so a consumer can drop the outline. Swingset gets the knob and the prop docs.

## Checklist

## Type of change

- [ ] 🐛 Bug fix
- [x] 🌟 New feature
- [ ] 🔨 Breaking change
- [ ] other:


maxyinger (commented):
80 line pr to add a 1px border. sheesh.

maxyinger (approved):

